PR #26, opened 2019-11-09
A few typos + function cleanup

donavon:


src/index.js
```diff
@@ -727,22 +727,13 @@ function isObject(a) {
...
-    return (
-      document.visibilityState === 'visible' ||
-      document.visibilityState === 'prerender'
-    )
-  }
-
-  return true
+  return document.visibilityState === undefined
+    || document.visibilityState === 'visible'
+    || document.visibilityState === 'prerender'
```
tannerlinsley: The changes here are not logically equivalent

src/index.js
```diff
@@ -727,22 +727,13 @@ function isObject(a) {
...
 }
 
 function isOnline() {
-  if (typeof navigator.onLine !== 'undefined') {
-    return navigator.onLine
-  }
-
-  return true
+  return navigator.onLine === undefined || navigator.onLine
```
tannerlinsley: I believe here too

src/index.js
donavon: they are equivalent. see this [CodeSandbox](https://codesandbox.io/s/stupefied-http-l3v0p) console output

src/index.js
donavon: also equivalent

src/index.js
tannerlinsley: Ah. Right you are!

donavon: maybe some unit tests would be in order 😉

tannerlinsley:
Indeed.

Tanner Linsley

tannerlinsley:
Merged!

tannerlinsley:
Released!

PR #29, opened 2019-11-11
API doc typo for useQuery param: 'pagination' -> 'paginated'

choznerol:
Accroding to [the src of `useQuery`](https://github.com/tannerlinsley/react-query/blob/master/src/index.js#L414) and [the infinite scroll example](https://github.com/tannerlinsley/react-query#load-more--infinite-scroll-pagination), the config key to enable pagination is `paginated:boolean` instead of `pagination:boolean`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1117,13 +1117,13 @@ const {
   failureCount,
   isLoading,
   refetch,
-  // with pagination mode enabled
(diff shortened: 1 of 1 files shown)


tannerlinsley:
Nice catch!

PR #37, opened 2019-11-15
refetchAllQueries promises error would not be raised anymore

### LINKED ISSUE #36: Error is not getting cought when coming back to the window
Hey, great library!
I noticed that when I useQuery and it fails, go out of the page and then go back - the error will not be caught in useQuery but be raised to the UI.

How to test?
1. Use useQuery to a endpoint that does not work (using async await)
2. Let it fail one time, go out of the tab and go back in.
3. You should see the Ui crashes.

Thanks!

amitripshtos:
Fixed #36

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -667,7 +667,7 @@ export async function refetchAllQueries({
         return query.fetch({ force })
       }
     })
-  );
+  )
 }
 
 export function clearQueryCache() {
--- a/src/index.js
+++ b/src/index.js
@@ -665,7 +665,9 @@ export async function refetchAllQueries({
         return query.fetch({ force })
       }
     })
-  )
+  ).catch(error => { 
+    console.error(error.message)
+  });
 }
 
 export function clearQueryCache() {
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -665,7 +665,9 @@ export async function refetchAllQueries({
         return query.fetch({ force })
       }
     })
-  )
+  ).catch(error => { 
+    console.error(error.message)
+  });
 }
```
tannerlinsley: This is great! Although, what this does now, is makes it so that if you import and use the `refetchAllQueries` function intentionally, you can't detect errors :(

tannerlinsley: This block however:

https://github.com/tannerlinsley/react-query/pull/37/files#diff-1fdf421c05c1140f6d71444ea2b27638R13-R15

is where the window-focus fetching happens. You may have better luck doing something similar there :)

tannerlinsley: https://github.com/tannerlinsley/react-query/blob/5194bbe7dee13008d08c990eed98b1a3bd8844ae/src/index.js#L13-L15

amitripshtos: I now catch the errors only for the usage of this method instead inside the method, thanks!

tannerlinsley:
Perfect!

amitripshtos:
@tannerlinsley Can you please release a new version? I want this change :)

PR #52, opened 2019-11-25
Remove fast async nodent runtime

### LINKED ISSUE #46: fast-async babel plugin modifies `Function.prototype` globally
I wonder if we can let the user decide to use `fast-async` themselves, rather than being baked in this project.

One reason could be that it can impact other parts of their app (see my PR to react-native https://github.com/facebook/react-native/pull/27264).

It currently breaks react native apps that does dynamic styling through passing a function to `StyleSheet.create`

fmnxl:
`fast-async` injects the nodent runtime, which adds `Function.prototype.$asyncbind` to the output.

This runtime method is not used in the output, so we can remove it.

Resolves #46

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "eslint-plugin-react": "7.x",
     "eslint-plugin-react-hooks": "1.5.0",
     "eslint-plugin-standard": "^4.0.0",
+    "fast-async": "^6.3.8",
     "prettier": "^1.18.2",
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -46,6 +46,7 @@
     "eslint-plugin-react": "7.x",
     "eslint-plugin-react-hooks": "1.5.0",
     "eslint-plugin-standard": "^4.0.0",
+    "fast-async": "^6.3.8",
```
purelysloth: this was missing for me

tannerlinsley:
Incredible! Great work!

tannerlinsley:
Screen Shot 2019-11-25 at 9 28 48 AM

tannerlinsley:
Of course!

PR #54, opened 2019-11-28
Fix setQueryData crash

edvinerikson:
Hello!

This PR fixes a bug in setQueryData where it would crash if one attempted to change a query which has not been created yet. 

The current change is making this a no-op. However we can probably create the query object here instead and apply the update on it, but I'm not sure what is desired.

 I also took the opportunity to write a simple test case for this using Jest. Please point me in the right direction if you find something that you don't feel comfortable with in this change. :)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -134,7 +134,10 @@ function makeQuery(options) {
         // Schedule garbage collection
         query.cacheTimeout = setTimeout(
           () => {
-            queries.splice(queries.findIndex(d => d === query), 1)
+            queries.splice(
+              queries.findIndex(d => d === query),
+              1
+            )
             globalStateListeners.forEach(d => d())
           },
           query.state.isCached ? query.config.cacheTime : 0
@@ -653,6 +656,10 @@ export function setQueryData(
   }
 
   const query = queries.find(d => d.queryHash === queryHash)
+
+  if (!query) {
+    return
+  }
 
   query.setData(updater)
 
--- a/.babelrc
+++ b/.babelrc
@@ -10,10 +10,26 @@
     "@babel/react"
   ],
   "plugins": [
-    ["module:fast-async", {
-      "compiler": {
-        "noRuntime": true
+    [
+      "module:fast-async",
+      {
+        "compiler": {
+          "noRuntime": true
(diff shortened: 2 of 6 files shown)


src/index.js
```diff
@@ -134,7 +134,10 @@ function makeQuery(options) {
         // Schedule garbage collection
         query.cacheTimeout = setTimeout(
           () => {
-            queries.splice(queries.findIndex(d => d === query), 1)
+            queries.splice(
```
edvinerikson: Seems to be prettier that is changing this.

src/index.js
```diff
@@ -654,6 +657,10 @@ export function setQueryData(
 
   const query = queries.find(d => d.queryHash === queryHash)
 
+  if (!query) {
+    return
```
edvinerikson: Here we can probably create the query object instead.

tannerlinsley: Yeah, we could create the query here. Do you have a use-case for that? Honestly, it could be the same as (or share logic from) `prefetchQuery`

edvinerikson: No I don't have a use case for it. However I worked around this bug by calling prefetchQuery prior to running setQueryData. Would you prefer doing the same thing as prefetchQuery?

edvinerikson: I tried using the same logic as prefetchQuery, but I couldn't get it to work. The tests seems to have some race-conditions and I feel that it's generally become very unpredictable how the different code paths reacts to a query without a query function provided.

I can push the code if you want to have a look.

tannerlinsley: Yeah I’ll take a look

edvinerikson:
I wasn't fast enough to push the code, do you think the current implementation is fine, or should I push my new changes to a new PR? 

The change done now is absolutely fine for me, but I can think of one case where it would be useful to prime a query prior to it being fetched client-side. If I fetch some data on the server and provide it inlined in the html markup, it can be useful to have a way to use that data instantly when booting the app.

tannerlinsley:
Does the newly added `config.initialData` work for this use case?

edvinerikson:
Hi! Yes that do seem to work.

PR #58, opened 2019-12-02
Prevent manually updating sizes-*.json

kuldeepkeshwar:


kuldeepkeshwar:
@tannerlinsley this(`/bin/sh: 1: jest: not found`) is weird. any idea ? I just wanted to add the env variable (`NODE_ENV=production`) in CI(travis)

tannerlinsley:
That Is strange. Never seen it. Maybe that’s causing yarn not to install dev deps? I would try making the env variable more specific to the build script instead of all of Travis.

Tanner Linsley

kuldeepkeshwar:
@tannerlinsley yes you were correct. yarn doesn't install dev deps in case of NODE_ENV=production 😄.

- Update `rollup-plugin-size`configuration to avoid writing files(size*.json) in non-ci environment

tannerlinsley:
So when do they get written outside of ci?

kuldeepkeshwar:
Bot automatically update them incase of any delta in sizes.


These files are source of truth for previous build stats. If we manually update them and CI runs after it then it won’t show any delta as sizes are already updated.

tannerlinsley:
Ah, that makes sense :)

PR #77, opened 2019-12-10
Fix refetchQueries readme 

### LINKED ISSUE #76: refetchQueries doesn't work in mutation call
Based on the [documentation](https://github.com/tannerlinsley/react-query#invalidate-and-refetch-queries-from-mutations), I expected passing the `refetchQueries` key in the second argument object to `mutate` would have refetched the specified queries.
e.g. `mutate(newTodo, { refetchQueries: ["todos"]})`

I am only able to get it to refetch when the key is passed in to the `useMutation` hook. 
e.g. `useMutation(addTodo, { refetchQueries: ["todos"]})`

I've modified the `EditTodo` function in the [default codesandbox app](https://codesandbox.io/s/flamboyant-water-cfph2) to recreate the issue.

Am I doing something wrong, is this a bug or does the documentation just need to be updated a bit to be consistent w/ the other examples?

Awesome library btw. Thanks!

timrbula:
Fixes #76

### DIFF
--- a/README.md
+++ b/README.md
@@ -832,7 +832,12 @@ When a successful `postTodo` mutation happens, we likely want all `todos` querie
 
 ```js
 // When this mutation succeeds, any queries with the `todos` or `reminders` query key will be refetched
-mutate(newTodo, { refetchQueries: ['todos', 'reminders'], })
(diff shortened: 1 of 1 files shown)


tannerlinsley:
Very nice! :)

PR #88, opened 2019-12-26
update guard against missing `document`

oliverwinter-vdb:
Pull request https://github.com/tannerlinsley/react-query/pull/38 added the guard for missing document on react native, this PR adds the check to the isDocumentVisible function so that the missing document error does not appear for things like intervals and other checks for visible document etc.


tannerlinsley:
Nice work :)

PR #90, opened 2019-12-28
Fix prefetchQuery link

mskelton:
The prefetchQuery link was not working due to it being camel cased instead of lower case.  This will fix that and make the link work again!


tannerlinsley:
Thanks!

PR #94, opened 2020-01-07
Fix query request cancellation

cschwebk:
Query request cancellation did not work as documented in the readme [here](https://github.com/tannerlinsley/react-query#cancelling-query-requests). Hopefully this PR makes it function in the way it was intended.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -134,9 +134,7 @@ function makeQuery(options) {
         // Cancel any side-effects
         query.cancelled = cancelledError
 
-        if (query.promise && query.promise.cancel) {
-          query.promise.cancel()
-        }
+        query.cancelQueries()
 
         // Mark as inactive
         query.setState(old => {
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -134,9 +134,7 @@ function makeQuery(options) {
         // Cancel any side-effects
         query.cancelled = cancelledError
 
-        if (query.promise && query.promise.cancel) {
```
cschwebk: `query.promise.cancel` would never get called because `query.promise` is declared within `react-query` as a promise that signifies the completion of the entire query. It is not the promise returned by the user provided `queryFn`.

src/index.js
```diff
@@ -162,9 +160,12 @@ function makeQuery(options) {
   const tryFetchQueryPages = async pageVariables => {
     try {
       // Perform the query
-      const data = await Promise.all(
-        pageVariables.map(variables => query.queryFn(variables))
-      )
+      const promises = pageVariables.map(variables => query.queryFn(variables))
+
+      query.cancelQueries = () =>
```
cschwebk: The only place to access the user provided `cancel` method on the promise returned from `queryFn` is here, and because this code will execute before `query.promise` is defined, the method cannot be stored as `query.promise.cancel`.

tannerlinsley:
Fantastic :)

src/index.js
dysfunc: 👏 👏 👏

sbogdaniuk:
Made some example for axios cancel token

https://codesandbox.io/s/react-query-axios-canceltoken-bzmdp

Issue #96, opened 2020-01-08
Unhandled rejection on error


tibotiber:
Hi and happy new year!

I've been fighting with an unhandled rejection on `useMutation` and it seems it might be coming from react-query. I am able to reproduce on your [sandbox](https://codesandbox.io/s/github/tannerlinsley/react-query/tree/master/examples/sandbox) and it seems to affect `useQuery` as well.

On the sandbox, increase the error rate to 0.95 (for example) and try adding a todo or refetching. The error is treated properly and surfaced in userland as expected, however create-react-app still catches it as unhandled and display its error overlay. 

Any idea? 

PS: Really great idea to have a sandbox playground for this kind of things 👍

tannerlinsley:
Ah, this is because there is a `mutate()` call happening in AddTodo without a try catch wrapper, thus, when the mutation fails, it throws all the way up into the renderer code (which is not async) and ends up at the top level of the app. This should be wrapped in a try catch block and handled like other errors, regardless if anything is done. If you would like to submit a PR to fix that in the example, I would accept it quickly.

tibotiber:
Thanks for explaining this. I've opened the PR you suggested.

Would you be opened to adding a `noThrow` boolean option to `mutate` that would basically avoid rethrowing here?
https://github.com/tannerlinsley/react-query/blob/6b56a14f061aaf9f515cc0c476acfff27c148cc3/src/index.js#L643

If yes, I don't mind trying to submit a PR.

PR #102, opened 2020-01-10
add try/catch in addTodo mutate call

tibotiber:
Hi @tannerlinsley, this is the quick PR you suggested in #96.


tibotiber:
made a booboo while committing, feel free to squash when merging

PR #122, opened 2020-02-07
add `reset` method to `useMutation`

vankhoawin:
This PR adds a `reset` method to the `useMutation`'s `MutationResult`. This function would be useful in cases where a React function's state should be reset to how it was originally, at initialization. As far as I know, there is only one way to update the data within a `useMutation` call: https://github.com/tannerlinsley/react-query/blob/07b0a66f39a9ce3b5b61ec720f3aebae6128bce2/src/index.js#L639

This PR also adds a test and some setup for more testing using `react-testing-library`.

And great work on the project, really been liking the API so far!

PS I'll add TS typing if this PR is approved.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,13 +1,13 @@
 {
   "dist/index.js": {
-    "bundled": 29796,
-    "minified": 14847,
-    "gzipped": 4213
+    "bundled": 29896,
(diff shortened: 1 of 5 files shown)


tannerlinsley:
I dig it. Thankyou!

PR #136, opened 2020-02-20
enhanced setQueryData to update multiple queries

kelleyvanevert:
This is a little example implementation of the requested feature of Issue https://github.com/tannerlinsley/react-query/issues/135.

_I basically just copied the logic from `refetchQuery`. Haven't tested it fully yet, will add some tests and an example usage later._

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,13 +1,13 @@
 {
   "dist/index.js": {
-    "bundled": 31777,
-    "minified": 16088,
-    "gzipped": 4619
+    "bundled": 32204,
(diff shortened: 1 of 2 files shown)


tannerlinsley:
This is nice :) I'll merge this when I get a break from the `next` branch.

tannerlinsley:
After some thought, and the decision for the same API question in v1 (in the `next` branch), I've decided to keep this api as a single-query function. You can always loop over individual queries keys to update their data and I don't believe that functionality is something I want to make possible out of the box.

Thanks for your input!

Issue #124, opened 2020-02-11
query.fetch raises unhandled error when it finishes to retry


amitripshtos:
Hey,
When you are using "useQuery" with the parameters:
```
{ refetchInterval: 1000 }
```

And the function which fetches the data raises error multiple time (and retrying is being finished), an exception will be thrown without any handling.

How to test it:
1. Open the Codesandbox of react-query
2. in line 62 change errorRate to be 1
3. In line 357 add the refetchInterval property
```
 } = useQuery(["todos", { filter }], fetchTodos, { refetchInterval: 1000 });
```
4. After few seconds the UI will show the unhandled exception

How to fix that:
Basically I think that in query.fetch the in line 366 at "index.js" the "throw error" is causing the issue. I'm not sure why we need to throw that error there so I would like a second eye for that.

Thanks,
Amit.

barrymay:
@tannerlinsley please check out my PR when you can, I believe it fixes the problem as desired.

PR #142, opened 2020-02-23
#124 Refetched query fetch should catch the promise call directly

barrymay:
@tannerlinsley The change you made for #124 doesn't seem correct (and doesn't work in my own testing of the sandbox mentioned).  

The try catch wouldn't catch the throw in the promise, whereas this would.

(you might want to just take out the try/catch entirely and catch the promise this way, up to you)

Great job btw, it's a fantastic library!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -447,15 +447,8 @@ export function useQuery(queryKey, queryFn, config = {}) {
     if (config.refetchInterval && !query.refetchInterval) {
       query.refetchInterval = setInterval(() => {
         if (isDocumentVisible() || config.refetchIntervalInBackground) {
-          try {
-            query.fetch().catch(err => {
-              Console.error(err)
-              // Swallow this error, since it is handled elsewhere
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -447,12 +447,8 @@ export function useQuery(queryKey, queryFn, config = {}) {
...
       query.refetchInterval = setInterval(() => {
         if (isDocumentVisible() || config.refetchIntervalInBackground) {
-          try {
-            query.fetch()
-          } catch (err) {
-            Console.error(err)
-            // Swallow this error, since it is handled elsewhere
-          }
+          // Swallow this error, since it is handled elsewhere
+          query.fetch().catch(Console.error)
```
tannerlinsley: It's actually even simpler than that!

tannerlinsley:
Thanks for catching this :)

tannerlinsley (approved):

PR #163, opened 2020-02-27
Stop auto fetch for manual useQuery option

oliverwinter-vdb:
I believe this will fix the manual query issue here https://github.com/tannerlinsley/react-query/issues/161


tannerlinsley:
Right you are! Thanks :)

tannerlinsley:
Right you are! Thanks :)

PR #166, opened 2020-02-27
Update useMutation memoization

owinter86:
This will fix the issue identified here https://github.com/tannerlinsley/react-query/issues/165, however I am not sure if the check on the status in the useCallback was needed for a particular reason?

This will allow the mutation function to safely be included in a dependency array of an effect, like below.
```javascript
const [mutate]  = useMutation(SOME_MUTATION);
React.useEffect(() => {
  mutate();
}, [mutate]);
```

### DIFF
--- a/src/useMutation.js
+++ b/src/useMutation.js
@@ -67,9 +67,6 @@ export function useMutation(
 
   const mutate = React.useCallback(
     async (variables, options = {}) => {
-      if (![statusIdle, statusSuccess, statusError].includes(state.status)) {
-        return
-      }
       dispatch({ type: actionMutate })
 
       const resolvedOptions = {
@@ -95,7 +92,7 @@ export function useMutation(
         }
       }
     },
-    [getConfig, getMutationFn, state.status]
+    [getConfig, getMutationFn]
   )
 
   const reset = React.useCallback(() => dispatch({ type: actionReset }), [])
--- a/src/useMutation.js
+++ b/src/useMutation.js
@@ -64,9 +64,15 @@ export function useMutation(
     ...useConfigContext(),
     ...config,
   })
+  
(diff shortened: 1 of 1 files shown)


src/useMutation.js
```diff
@@ -67,9 +67,6 @@ export function useMutation(
 
   const mutate = React.useCallback(
     async (variables, options = {}) => {
-      if (![statusIdle, statusSuccess, statusError].includes(state.status)) {
-        return
-      }
       dispatch({ type: actionMutate })
```
tannerlinsley:
```suggestion
      if (![statusIdle, statusSuccess, statusError].includes(getStatus())) {
        return
      }
      dispatch({ type: actionMutate })
```

src/useMutation.js
```diff
@@ -67,9 +67,6 @@ export function useMutation(
```
tannerlinsley:
```suggestion
  const getStatus = useGetLatest(state.status)
  
```

src/useMutation.js
```diff
@@ -95,7 +92,7 @@ export function useMutation(
         }
       }
     },
-    [getConfig, getMutationFn, state.status]
+    [getConfig, getMutationFn]
```
tannerlinsley:
```suggestion
    [getConfig, getMutationFn, getStatus]
```

tannerlinsley:
That multiline suggesting thing is trippy!

owinter86:
Yeah that works too.

tannerlinsley:
Yeah, the check is to make sure the mutation is idle or an existing mutation is settled before allowing another one to to be triggered. We may not need it, but I'd like to keep it in until a use case comes along to remove it.

PR #172, opened 2020-02-28
Change 'todos' ID to 'projects' in pagination example

thatzacdavis:
We are fetching projects in this example, so calling it 'todos' didn't make sense to me.


tannerlinsley:
Indeed :)

PR #191, opened 2020-03-04
Reset failureCount to zero

adibfirman:
Hello,
I found some problem when i am using `failureCount` it's not reset to zero as your readme say.

Thanks

### DIFF
--- a/src/queryCache.js
+++ b/src/queryCache.js
@@ -21,7 +21,6 @@ const actionFetch = {}
 const actionSuccess = {}
 const actionError = {}
 const actionSetData = {}
-const actionResetFailureCount = {}
 
 export function makeQueryCache() {
   const listeners = []
@@ -317,7 +316,6 @@ export function makeQueryCache() {
               try {
                 const data = await tryFetchData(queryFn, ...args)
                 if (query.cancelled) return reject(query.cancelled)
-                dispatch({ type: actionResetFailureCount })
                 resolve(data)
               } catch (error) {
                 if (query.cancelled) return reject(query.cancelled)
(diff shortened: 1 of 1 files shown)


src/queryCache.js
```diff
@@ -316,6 +317,7 @@ export function makeQueryCache() {
               try {
                 const data = await tryFetchData(queryFn, ...args)
                 if (query.cancelled) return reject(query.cancelled)
+                dispatch({ type: actionResetFailureCount })
```
tannerlinsley: This is unnecessary with the suggested changes below.

src/queryCache.js
```diff
@@ -21,6 +21,7 @@ const actionFetch = {}
 const actionSuccess = {}
 const actionError = {}
 const actionSetData = {}
+const actionResetFailureCount = {}
```
tannerlinsley: This is unnecessary with the suggested changes below.

src/queryCache.js
```diff
@@ -419,6 +421,11 @@ export function makeQueryCache() {
 
 export function defaultQueryReducer(state, action) {
   switch (action.type) {
+    case actionResetFailureCount:
+      return {
+        ...state,
+        failureCount: 0
+      }
```
tannerlinsley: Instead of using this new action, just updated the success action below.

tannerlinsley: The `actionSuccess` handler can look like this: 

```diff
case actionSuccess:
      return {
        ...state,
        status: statusSuccess,
        data: action.data,
        error: null,
        isStale: false,
        isFetching: false,
        canFetchMore: action.canFetchMore,
        updatedAt: Date.now(),
+        failureCount: 0
      }
```

adibfirman: that s is a neat, thanks for reviewing my code :pray:

deevus:
Love an easy fix :)

PR #201, opened 2020-03-05
Fix typo in README.md

ipanasenko:
This looks like a typo for me. In key there's `completed`, and in `fetchTodoList` it becomes `status`. How should it be correctly?

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,7 +402,7 @@ To use external props, state, or variables in a query function, it's easiest to 
 ```js
 function Todos({ completed }) {
   const { status, data, error } = useQuery(
-    ['todos', { completed, page }],
+    ['todos', { status: completed, page }],
     fetchTodoList
   )
 }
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -402,7 +402,7 @@ To use external props, state, or variables in a query function, it's easiest to
 ```js
 function Todos({ completed }) {
   const { status, data, error } = useQuery(
-    ['todos', { completed, page }],
+    ['todos', { status: completed, page }],
```
tannerlinsley:
```suggestion
    ['todos', { status, page }],
```

ipanasenko:
@tannerlinsley in your change, where is `status` coming from? I think new code will fail. `status` should come from props then, but it will result in duplicate variables

tannerlinsley:
Status is just a made up variable like `foo`. Not to be confused with the actual status variable returned by a query. If you want to change it to something like `completed` you can to avoid confusion.

PR #207, opened 2020-03-06
queryCache.prefetchQuery should return cache contents on cache hit

brumm:
Hey Tanner,

currently queryCache.prefetchQuery only explicitly returns if a fetch was actually attempted, due to `query.state.isStale || force`. This leads to it returning `undefined` for cache hits.

This PR naively fixes my failing test, let me know if there's more to be considered here.

codesandbox repro https://codesandbox.io/s/react-query-bug-repro-gi7oy

### DIFF
--- a/src/queryCache.js
+++ b/src/queryCache.js
@@ -167,9 +167,9 @@ export function makeQueryCache() {
           throw err
         }
       }
-    } else {
-      return Promise.resolve(query.state.data)
     }
+    
+    return query.state.data
   }
 
   cache.setQueryData = (queryKey, updater, { exact } = {}) => {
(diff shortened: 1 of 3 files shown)


src/queryCache.js
```diff
@@ -167,6 +167,8 @@ export function makeQueryCache() {
           throw err
         }
       }
+    } else {
+      return Promise.resolve(query.state.data)
     }
```
tannerlinsley: Because this is async already, the `Promise.resolve` is redundant along with the else statement. I suggest this:
```suggestion
    }
    
    return query.state.data
```

brumm: D'oh, I missed the fact that we're async already. Thanks!

PR #210, opened 2020-03-06
Add types to react-query package

nksaraf:
Hi, thanks for this awesome library. I was using it with typescript but found that the typings at `@types/react-query` are old and not up to date with the API. So I refreshed them and thought it would be a good idea if the library shipped with them. 

P.S: Just starting with contributing to OSS, if I missed any steps in this process, please correct me

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,10 +1,3 @@
-// Type definitions for react-query 0.3
-// Project: https://github.com/tannerlinsley/react-query
-// Definitions by: Lukasz Fiszer <https://github.com/lukaszfiszer>
-//                 Jace Hensley <https://github.com/jacehensley>
-//                 Matteo Frana <https://github.com/matteofrana>
-// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-
 import { ComponentType } from "react";
 
 export function useInfiniteQuery<TResult, TVariables extends object>(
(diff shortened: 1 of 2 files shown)


tannerlinsley:
Hey, this is great! I'm stoked that you're getting into contributing :)

So in the case of these types (which are awesome), for the time being we don't host the types directly here in this repo since the library isn't written in TS itself. However, @types/react-query could definitely use this update:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-query/index.d.ts

jarleriksen:
Hi @nksaraf - Awesome types! Are you going to create a pull request on the DefinitelyTyped repo? :)

nksaraf:
Hi @jarleriksen I want to do that. But like I said, I am new to this and contributing to DefinitelyTyped seemed a little daunting to me. I don't want to screw things up there. Any help/guidance will be amazing. I have been using the typings internally for my project but haven't got time to do all the things necessary for that repo. If you could use the typings internally, it would be give me some confidence to make the effort for the PR. Also I only use the useQuery and useMutation hooks mainly, not sure if everything is definitely correct. But here is a link to a gist https://gist.github.com/nksaraf/a00f0670c1d29e9490a1d1d856a3866d with the types file only (updated with some bug fixes against this PR)

tannerlinsley:
FYI, you're cordially invited to this discussion: https://github.com/tannerlinsley/react-query/discussions/362

PR #211, opened 2020-03-06
updateQuery

denisborovikov:
Since updateQuery was removed in 1.0.0 these parts should be removed or rewritten.


README.md
```diff
@@ -2133,8 +2133,6 @@ The `queryCache` instance is the backbone of React Query that manages all of the
   - If you want to force the query to prefetch again, you can pass the `force: true` option in the query config
 - If the query does not exist, it will be created and immediately be marked as stale. **If this created query is not utilized by a query hook in the `cacheTime` (defaults to 5 minutes), the query will be garbage collected**.
 
-> The difference between using `prefetchQuery` and `updateQuery` is that `prefetchQuery` is async and will ensure that duplicate requests for this query are not created with `useQuery` instances for the same query are rendered while the data is fetching.
```
tannerlinsley: It should actually say `setQueryData` instead of `updateQuery`. With that change, it would still be valid.

PR #224, opened 2020-03-10
Output ES module from rollup

warlo:
After v1.0.12 our rollup build broke due to namedExports no longer being available to rollup due to the `NODE_ENV` + `require()` dev/prod logic. That forces us to explicitly alias all react-query exports in our rollup-config, however building an es module makes rollup happy.

I guess the dev / prod build is not that required for es modules, and libraries like redux only provides a single minified es build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "license": "MIT",
   "repository": "tannerlinsley/react-query",
   "main": "index.js",
+  "module": "dist/react-query.min.mjs",
   "sideEffects": false,
   "scripts": {
     "test": "is-ci \"test:ci\" \"test:dev\"",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -23,13 +23,20 @@ export default [
   },
   {
     input: 'src/index.js',
-    output: {
-      name: 'ReactQuery',
-      file: 'dist/react-query.production.min.js',
-      format: 'umd',
-      sourcemap: true,
-      globals,
-    },
+    output: [
+      {
+        name: 'ReactQuery',
+        file: 'dist/react-query.production.min.js',
+        format: 'umd',
+        sourcemap: true,
+        globals,
+      },
+      {
+        file: 'dist/react-query.min.mjs',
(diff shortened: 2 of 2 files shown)


tannerlinsley:
Hmm. I thought the way it is currently set up was sufficient as we simply mimicked the way React itself was set up. If I recall correctly, React doesn't have an es module build does it?

hanswilw:
You are probably right about react, and it might be a good reason for it not having one yet. (We actually use preact that provides an es module build). I see now that you had multiple outputs, including an es module, thats probably why it worked in our build prior to v1.0.12. However when you removed the module and made the index.js entrypoint the only option, rollup (or their commonjs plugin) is no longer able to resolve what is exported. 
This does atleast give me:
`[!] Error: 'useMutation' is not exported by node_modules/react-query/index.js`
or even using the production build directly 
`[!] Error: 'useMutation' is not exported by node_modules/react-query/dist/react-query.production.min.js`

It might be some bad config on our end, but regardless of that I am not sure if there is a reason _not_ to provide an es module as an alternative. UMD is not statically analyzable afaik and generally used as fallback builds(?), which makes it harder for bundlers to leverage properties like tree-shaking etc.

tannerlinsley:
Is there any reason to use `react-query.mjs` over just `react-query.es.js`. Also, I don't think it should be minified, since that will likely happen with their own minifier at a later stage.

hanswilw:
Node went with .mjs, and mozilla recommends and refers to node as well: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#Aside_%E2%80%94_.mjs_versus_.js

Personally not important with minification and as you say it will probably be minified anyways, but would it pose any issue? Sourcemaps should be available, and it _might_ be preferable for users loading the module from a CDN or similar.

tannerlinsley:
Okay. In that case, we can keep it `mjs`. Okay, CDN usage makes sense. We can minify it until someone complains (maybe never)

PR #228, opened 2020-03-11
Fix returned value from useIsFetching

Zekneer:
useIsFetching is returning value of queryCache.isFetching in numbers (1.0.23)


tannerlinsley:
This is intentional.

PR #265, opened 2020-03-19
tuple returned by useMutation

mwmcode:
Hi again,

I could be wrong, but I think you meant to document `reset` instead of `promise` (I did a quick check and found that no `promise` is being returned)


tannerlinsley:
Mutate returns a promise. useMutation returns a tuple, and it’s documented as such.

PR #271, opened 2020-03-20
clear stale timeout when removing queries

### LINKED ISSUE #270: Recommendation on testing
Hey there :wave:

I'm getting an `act` warning when using react-query hooks because of this:

https://github.com/tannerlinsley/react-query/blob/df4bd3b54943db6acc4df612e6e85d6b09ea1a29/src/queryCache.js#L228-L237

Basically what's happening is my component finishes rendering and then react-query triggers a re-render ([thanks to this](https://github.com/tannerlinsley/react-query/blob/c66dcc9e28cf24b8469cad3af489d16ec1499aa8/src/useBaseQuery.js#L67)) which is the thing that's giving me the act warning.

Now I can add an [`await waitFor(() => {})`](https://testing-library.com/docs/dom-testing-library/api-async#waitfor) call to the end of my test and the warning goes away, but I think it would be better to have an assertion in that callback. I'm just not sure what to assert happened as a result of the state update.

Do you have any suggestions for people testing this stuff? Should I just set the `staleTime` to `Infinity`? Kinda thinking that might work except then my tests would have a different config from my app which could lead to some problems.

Any ideas?

kentcdodds:
Closes #270

### DIFF
--- a/src/queryCache.js
+++ b/src/queryCache.js
@@ -72,6 +72,7 @@ export function makeQueryCache() {
     const foundQueries = findQueries(predicate, { exact })
 
     foundQueries.forEach(query => {
+      clearTimeout(query.staleTimeout)
       delete cache.queries[query.queryHash]
     })
 
@@ -230,9 +231,7 @@ export function makeQueryCache() {
         return
       }
       query.staleTimeout = setTimeout(() => {
-        if (query) {
-          dispatch({ type: actionMarkStale })
-        }
+        dispatch({ type: actionMarkStale })
       }, query.config.staleTime)
     }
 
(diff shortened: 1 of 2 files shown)


src/queryCache.js
```diff
@@ -230,9 +231,7 @@ export function makeQueryCache() {
         return
       }
       query.staleTimeout = setTimeout(() => {
-        if (query) {
-          dispatch({ type: actionMarkStale })
-        }
+        dispatch({ type: actionMarkStale })
```
kentcdodds: I couldn't see anywhere in the code where `query` would be `falsy` and especially now that we're clearing the timeout when they're removed I think we're good to simplify this.

tannerlinsley:
Patch Released

PR #310, opened 2020-04-03
fix(prefetchQuery): Throw error when throwOnError is set

### LINKED ISSUE #309: bug: prefetchQuery does not throw error when throwOnError is true
According to the documentation, when `prefetchQuery` is called with the config option `throwOnError` set to `true`, the query should throw when encountering an error. Instead, the error is still swallowed (and since `onSuccess` / `onError` callbacks are not invoked with `prefetchQuery`, the error can not be seen).

I have a test case to reproduce this and will open a PR to propose a fix. It looks like `throw error` was commented out recently, possibly due to another issue.

kamranayub:
Fixes #309 

The previous test added for #167 still passes, so I would assume this is safe.

## Changes

- prefetchQuery: Add test case for `throwOnError` set to `true`
- Uncomment `throw error` that was commented in #167

### DIFF
--- a/src/queryCache.js
+++ b/src/queryCache.js
@@ -413,7 +413,7 @@ export function makeQueryCache() {
                   instance.onSettled && instance.onSettled(undefined, error)
               )
 
-              // throw error
+              throw error
             }
           }
         })()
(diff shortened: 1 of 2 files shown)


tannerlinsley:
Nice Catch!

PR #312, opened 2020-04-03
fix(prefetchQuery): Support force option when calling query.fetch

### LINKED ISSUE #311: bug: prefetchQuery does not respect force option when initialData is used
When using `prefetchQuery` and `initialData`, calling `prefetchQuery` will always return the `initialData`, like so:

```js
const fetchFn = () => Promise.resolve('fresh')
const first = await prefetchQuery('test', fetchFn, { initialData: 'initial' });
const second = await prefetchQuery('test', fetchFn);

expect(first).toBe('initial');
expect(second).toBe('initial');
```

According to the [documentation](https://github.com/tannerlinsley/react-query#querycacheprefetchquery), I believe this is "as designed" since the intention is for `useQuery` to not have to fetch new data.

Even so, you _should_ be able to workaround this by passing `force: true` to always fetch fresh data, as the docs suggest:

> If you want to force the query to prefetch again, you can pass the force: true option in the query config

```js
const fetchFn = () => Promise.resolve('fresh')
const first = await prefetchQuery('test', fetchFn, { initialData: 'initial' });
const second = await prefetchQuery('test', fetchFn, { force: true });

expect(first).toBe('initial'); // pass
expect(second).toBe('fresh'); // fails, actual: 'initial'
```

I have a test and fix for this `force` workaround.

kamranayub:
Fixes #311 

## Changes

- prefetchQuery: Passthru `force` option when calling `query.fetch`
- Add test coverage

### DIFF
--- a/src/queryCache.js
+++ b/src/queryCache.js
@@ -160,7 +160,7 @@ export function makeQueryCache() {
     if (query.state.isStale || force) {
       // Trigger a fetch and return the promise
       try {
-        const res = await query.fetch()
+        const res = await query.fetch({ force })
         query.wasPrefetched = true
         return res
       } catch (err) {
--- a/src/tests/queryCache.test.js
+++ b/src/tests/queryCache.test.js
@@ -24,6 +24,16 @@ describe('queryCache', () => {
     expect(second).toBe(first)
   })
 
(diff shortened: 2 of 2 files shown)


tannerlinsley:
Again, great catch!

Issue #324, opened 2020-04-06
queryFn is called twice in Suspense mode


simoncarpinter:
Per the title, when using `useQuery` with `{ suspense: true }`, the supplied query function will be called twice.

See https://github.com/simoncarpinter/react-query/commit/cc09457c85946c8097334e516ecdd5f35a0a8801 for details, opening a PR now that adds the test.  I haven't yet spotted the actual bug.

simoncarpinter:
It appears to be somewhere in `query.fetch`, maybe?  `handleSuspense` in `useQuery.js` calls `query.refetch`, which calls `query.fetch`, which -- if I'm reading things right -- *should* return the existing promise, and only start a new one if there's nothing preexisting:

```
// Create a new promise for the query cache if necessary
      if (!query.promise) {
        query.promise = (async () => {
```

And yet, if I debug/`console.log` (lazy), `query.promise` is undefined in the initial request (expected) but is *still* undefined in the second request (unexpected) -- despite the fact that debugging *afterwards* shows `query.promise` is set in both cases.

Possibly a race condition?  Or possibly the `query` instance is changing, despite the fact it shouldn't be...

simoncarpinter:
Ah.  Removing `delete query.promise` resolves the issue, but presumably will re-open #216, which is what that line closed.

tannerlinsley:
This is indeed a funky issue. Still thinking here... keep up the info, it's helping.

pjshu:
same problem😂

ofer-h:
Hey @tannerlinsley 
Seems like there is a known issue around this topic as  useState / useMemo ... , are not behaving as expected while inside a suspend comp as it may initializer multiple times.

As I'm not 100% sure this is the same case here I will just share here the issues I refer to  

[useMemo gets called twice with Suspense](https://github.com/facebook/react/issues/14413)
[useState initializer called multiple times with suspense](https://github.com/FormidableLabs/react-ssr-prepass/issues/4
)

ofer-h:
So, I was able to reproduce this thanks to @simoncarpinter fork and the cuse was this line:

https://github.com/tannerlinsley/react-query/blob/8a6ec246deaef9f4a0c1b32298874e3e74dc93c5/src/utils.js#L134
[...]

PR #325, opened 2020-04-06
test that Suspense mode calls queryFn only once

simoncarpinter:
See #324 -- this adds some test coverage.  I'd like to fix the bug too, but haven't yet spotted where it's lurking.

### DIFF
--- a/src/tests/suspense.test.js
+++ b/src/tests/suspense.test.js
@@ -1,0 +1,33 @@
+import { render, waitForElement, cleanup } from '@testing-library/react'
+import * as React from 'react'
+
+import { useQuery, queryCache } from '../index'
+import { sleep } from './utils'
+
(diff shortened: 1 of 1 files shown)


src/tests/suspense.test.js
```diff
@@ -0,0 +1,33 @@
...
+import { useQuery, queryCache } from '../index'
+import { sleep } from './utils'
+
+describe("useQuery's in Suspense mode", () => {
+  afterEach(() => {
+    queryCache.clear()
+    cleanup()
+  })
+
+  it('should not call the queryFn twice when used in Suspense mode', async () => {
```
simoncarpinter, line 13: can remove the "when used in Suspense mode" bit, woops.

```suggestion
  it('should not call the queryFn twice', async () => {
```

PR #338, opened 2020-04-09
test(useQuery): Ensure we await the rendered text we expect

kamranayub:
I noticed that CI was failing intermittently on a test and it was receiving an `act()` warning. I believe it's because we should be awaiting the find of the text to allow for async work to happen in `useQuery`.

To reproduce locally, run `yarn test` then `a` for all tests, then hit `Enter` twice to do two full test runs in sequence. The subsequent runs should fail reliably on this test.

This seemed to be introduced after my PR #332 perhaps due to fixing the GC issue, and a state update happening that isn't waited for during the test. Other tests do use `act` to do assertions as well, plus using `find` for components that do async work seems a lot more reliable.

## Changes

- Fix `act` warning that happens on subsequent test runs for test `should be in "success" state by default in manual mode`

### DIFF
--- a/src/tests/useQuery.test.js
+++ b/src/tests/useQuery.test.js
@@ -1,9 +1,4 @@
-import {
-  render,
-  act,
-  waitForElement,
-  fireEvent,
-} from '@testing-library/react'
+import { render, act, waitForElement, fireEvent } from '@testing-library/react'
 import * as React from 'react'
 
 import { useQuery, queryCache } from '../index'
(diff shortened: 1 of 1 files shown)


src/tests/useQuery.test.js
```diff
@@ -1,9 +1,4 @@
-import {
-  render,
-  act,
-  waitForElement,
-  fireEvent,
-} from '@testing-library/react'
+import { render, act, waitForElement, fireEvent } from '@testing-library/react'
```
kamranayub, line 1: Prettier did this for me

src/tests/useQuery.test.js
```diff
@@ -152,9 +147,9 @@ describe('useQuery', () => {
       )
     }
 
-    const rendered = render(<Page />)
+    const { findByText } = render(<Page />)
```
kamranayub, line 150: Suspicion: due to GC happening on subsequent tests that was fixed in #332, this may have been passing as a fluke. The test before this ensures it's in a `loading` state by default, which is what the rendered output looks like for this test when it was failing.

PR #354, opened 2020-04-15
add rick and morty example

juanprq:
this is a very similar example to the star wars app, since the star wars api
is going to be deprecated and is currently down [swapi](https://swapi.co/), my proposal is to use
this rick and morty example to replace the star wars example that is not
working right now.


tannerlinsley:
Awesome! Can you update the Readme with a link to it too?

juanprq:
sure! I'm on it

PR #385, opened 2020-04-21
Add cyberhaven to used by

zmitry:


tannerlinsley:
Do you happen to work for Cyber Haven?

zmitry:
@tannerlinsley What do you mean? I do work for cyberhaven at the moment. 

https://www.linkedin.com/in/zdmitr/
https://github.com/CyberhavenInc/graphviz-wasm

tannerlinsley:
Cool. So the way I'm structuring things is that for companies that *want* be in the readme (and thus get a prominent link to their site), will need to become a sponsor. The exception to that rule is for companies who are already sponsors, or if I personally want to choose any other large house-hold name companies that speak the reputation of the library.

zmitry:
Okay, got it. Thanks.

PR #386, opened 2020-04-21
Update CachedQuery to add type to state member

salockhart:
While using the `queryCache.refetchQueries` method with a predicate, I noticed that the `query.state` was typed as `unknown`. This blocks retrieval of the `data` member of `state`.

I have added types to the `state` member, `CachedQueryState<T>`. I decided to make it generic in order to be able to constrain the type for `data` - however, it could also be `any` just as easily.

Should this be backported back to DefinitelyTyped?


tannerlinsley:
No. The DT types need to be deprecated tbh

PR #408, opened 2020-04-26
fix: rebuild cursors on infinite query refetch

krasnedev:
This is the proposed fix for https://github.com/tannerlinsley/react-query/issues/407

The basic idea is to invalidate cursors on refetch by reloading pages sequentially, instead of in parallel.
Refer to the issue above for more details and an example of the flaw in the current implementation.
I also included a test demonstrating the validity of this approach.

### DIFF
--- a/README.md
+++ b/README.md
@@ -746,7 +746,7 @@ function Projects() {
 
 ### What happens when an infinite query needs to be refetched?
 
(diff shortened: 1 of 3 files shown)


tannerlinsley:
Looks good!!!

PR #429, opened 2020-04-30
fix(config): Restore previous original config when ReactQueryConfigProvider is unmounted

### LINKED ISSUE #428: `defaultConfigRef` is not cleared when ReactQueryConfigProvider is unmounted
Working within Storybook, stories are unmounted when switching between each other. Using "decorator" functions, this is useful because we can manage different query configs for different stories.

However, I see that `defaultConfigRef` is updated when config changes but not reset to the original defaults when the provider unmounts. This persists config across Stories even when config provider is unmounted.

I have reproduced this in a test case and can submit a PR to address this.

kamranayub:
Fixes #428 

## Changes

- Track the "starting" config that is built with each `ReactQueryConfigProvider` and restore it when the provider is unmounted. This should support restoring to starting defaults _and_ previous parent config, when nested.


.vscode/settings.json
```diff
@@ -1,8 +1,8 @@
 {
-  "workbench.colorCustomizations": {
-    "titleBar.activeBackground": "#ff4154", // change this color!
-    "titleBar.inactiveBackground": "#ff4154", // change this color!
-    "titleBar.activeForeground": "#ffffff", // change this color!
-    "titleBar.inactiveForeground": "#ffffff" // change this color!
+  "workbench.colorCustomizations": {
```
kamranayub, line 2: omg this file is gonna kill me

.vscode/settings.json
```diff
@@ -1,8 +1,2 @@
 {
```
kamranayub: @tannerlinsley This keeps being updated automatically by VS Code every time I want to contribute, is it possible maybe to ignore this?

It gets changed to:

```js
{
  "workbench.colorCustomizations": {
    "titleBar.activeBackground": "#ff4154",
    "titleBar.inactiveBackground": "#ff415499",
    "titleBar.activeForeground": "#e7e7e7",
    "titleBar.inactiveForeground": "#e7e7e799",
    "activityBar.activeBackground": "#ff7482",
    "activityBar.activeBorder": "#70ff60",
    "activityBar.background": "#ff7482",
    "activityBar.foreground": "#15202b",
    "activityBar.inactiveForeground": "#15202b99",
    "activityBarBadge.background": "#70ff60",
    "activityBarBadge.foreground": "#15202b",
    "statusBar.background": "#ff4154",
    "statusBar.border": "#ff4154",
    "statusBar.foreground": "#e7e7e7",
    "statusBarItem.hoverBackground": "#ff7482",
    "titleBar.border": "#ff4154"
  },
  "peacock.color": "#ff4154"
}
```

tannerlinsley: Yes, you can add it to gitignore

kamranayub:
Hmm, I am noticing bleeding of config across tests. Not good. Have to keep investigating.

tannerlinsley:
Hmm, anything I can do to help out here?

kamranayub:
I'll be looking at this again soon but since it's only an issue right now for us within Storybook, I'll probably look at it next week 👍

kamranayub:
Closer. One of the tests will fail without the change to config now.

One very strange thing. This line:

```
await queryCache.prefetchQuery(...)
```

Will throw an `act` warning. However, if I change it to:

```
await act(async () => {
  await queryCache.prefetchQuery(...)
})
```

Then _nothing after it will be expected_ so the test "passes". It's very strange and I am not sure why that's happening.

kamranayub:
OK, it's in a better spot now. I need to test this locally with our codebase and see if it fixes the issue I intended to fix 👍

kamranayub:
👍 It fixes the issue!

PR #452, opened 2020-05-05
fix(typescript): changed root location of `types` relative to `main` index.js to fix import autocomplete

### LINKED ISSUE #427: types autocomplete import from `react-query/types`
image

They should import from `react-query`

ilovett:
Fixes #427


tannerlinsley:
Nice!

Issue #461, opened 2020-05-06
Support per-request cache with SSR


Ephem:
This is in a sense a continuation of the now closed #70 

First of all, thanks for your hard work on this library, it's great! I'm currently designing a new somewhat complex data fetching solution at work and I would like to build it on top of react-query, but there is some functionality around SSR that I would like to discuss.

I think the call to not cache data at all in #70 was the right one given the circumstances, for security reasons there can not be a cache by default on the server since it might leak user data cross-requests. However, this has some cumbersome consequences:

* Using `queryCache.prefetchQuery` ahead of the server rendering (like in Next.js `getInitialProps`) wont automatically prime the cache
* Calling `useQuery` with `initialData` in one place wont prime the cache for use in a second place, so reading the same data from the cache in multiple places becomes impossible

A big point of react-query is to be _the_ global cache for data. Since this is not true on the server, a lot of nice patterns falls apart, and the only way to fix that is to implement your _own_ cache on the server that wraps react-query, for example using a custom `useQueryWithSSRCache` that passes some `initialData` every time it is used. This seems a bit backwards since it's re-implementing a somewhat big part of react-query.

My suggestion is to make using a cache on the server opt-in and design it around creating a new cache per request, that you place on a context provider. If a cache is available on the context, use that instead of the global one. 

The other part of the puzzle is to make that cache serialisable (and you probably want to destroy the entire cache when you do so), so you can send it to the client and hydrate it there. On the client you could either hydrate the global cache before rendering, or create a cache and place on a context like on the server (but then you can't import `queryCache` like normal).

This is also something that will very likely be needed to support Suspense server rendering on the server in the future, if you don't have a cache per request, what do you read from to see if you need to Suspend or not? I'm not sure if this is something you are interested in, or if you deem it out of scope for the library? From the Readme:

> Query caches are not written to memory during SSR. This is outside of the scope of React Query and easily leads to out-of-sync data when used with frameworks like Next.js or other SSR strategies.

(Btw, I'm curious about what you mean by out of sync here?)

I'd love to get your input on this! This is a complex area so even if this became somewhat long, there is still a lot of nuance here of course, I'm very open to discussing it further and might very well be interested in contributing to something like this as well.

tannerlinsley:
Go on, go on.. I'm listening 😉 I definitely want to figure this out.

[...]

Issue #460, opened 2020-05-06
unit testing at odds with queryCache


jackmellis:
I'm having trouble working out the best way to test using react-query. I've looked at #113 but it doesn't solve my issues.

I've got a pretty good DI pattern for the actual fetching of data so that's not an issue. The issue I have is the queryCache persisting between tests. I can explicitly clear the cache before/after each test, but that doesn't help if you've got multiple async tests running in parallel.

Rather than importing a global queryCache, would it be helpful to store it in a react context, and then in your tests you can pass in a fresh cache for each test i.e.
```tsx
mount(
  <QueryCacheProvider>
    <MyComponent/>
  <QueryCacheProvider>
)
```
or add it to the existing config provider. 

The other issue I have is that my test runner (mocha) is hanging forever after the tests have run. I assume this is because of some internal timeouts/intervals/evenListeners or something. Am I missing some obvious way to stop them?

An example of an extremely simple test that's both failing and hanging:
```tsx
      const Sink = createSink(() => {
        useAccount('foo'); // useAccount just calls useQuery
        expect(mockApiCall.called).to.be.true;
      });
      
      mount(
        <Wrapper>
          <Sink/>
        </Wrapper>
      );
```
(the reason it's failing is because a previous test has already called the useQuery hook and cached the response, if I ran a single test in isolation, it passes)

I think my bottom line of all this is I think react-query could do with some official ways to make unit testing easier (and keeping things isolated)

jackmellis:
you can ignore the "hanging" part, I forgot you have to manually unmount enzyme-mounted components :facepalm:

KerrGrah:
thanks for raising this - your suggestion to clear the cache is sufficient for me 
`beforeEach(queryCache.clear)`

would be great if was built-in

iamtekeste:
@KerrGrah how do you get access to `queryCache` from inside your test? As how do I import it?

KerrGrah:
@iamtekeste 
`import { queryCache } from 'react-query'` :)

iamtekeste:
> @iamtekeste
> `import { queryCache } from 'react-query'` :)

I can't believe it was that easy! Thank you.

adeelibr:
How can you use `queryCache` in `react-query` v3 inside tests?

TkDodo:
[...]

Issue #432, opened 2020-05-01
Improve docs to avoid 'act' warning when using react-query in tests


dustinsoftware:
This is sort of related to #270. Filing in case we can improve the docs to help others avoid this issue.

When running tests, I ran into this error:

```
console.error node_modules/react-dom/cjs/react-dom.development.js:88
    Warning: An update to _default inside a test was not wrapped in act(...).
```

Here's the test:
```tsx
test('renders the component', async () => {
  await act(async () => {
    const { getByText } = render(<Component />);
    const linkElement = getByText(/Hello/i);
    expect(linkElement).toBeInTheDocument();
  });
});
```

And here's the component (simplified):
```tsx
const getPing = async () => {
  const { data } = await axios.get<{ content: string }>(
    getRoutePrefix() + 'ping'
  );
  return data;
};

export default function () {
  const { status, data, error, isFetching } = useQuery('ping', getPing);

  return (
    <>
      <div>Hello!</div>

      <div>{JSON.stringify(status)}</div>
      <div data-testid="testdata">{JSON.stringify(data)}</div>
      <div>Err: {JSON.stringify(error)}</div>
      <div>{JSON.stringify(isFetching)}</div>
    </>
  );
}
```

In #271 there was some cleanup logic added, however the error in my case appears to have been thrown before the cleanup call was fired.

I can fix the issue by adding to my tests at the end of each act call:
```diff
- import { render, wait } from '@testing-library/react';
+ import { render, wait, cleanup } from '@testing-library/react';

test('renders the component', async () => {
  await act(async () => {
    const { getByText } = render(<Component />);
    const linkElement = getByText(/Hello/i);
    expect(linkElement).toBeInTheDocument();
+   cleanup();
  });
});
```

Doing the cleanup _outside_ the act does not help; that appears to be too late.

Is it necessary to add the `cleanup` call inside every `act` call? If so, can we update the react-query docs on how to test with jest + react-testing-library? I bet others will run into this issue as well and hoping to save some time as more users adopt the library.

Here's some screenshots with a debugger attached indicating the warning is originally caused by a call to `refetch`:
![Screen Shot 2020-05-01 at 12 07 49](https://user-images.githubusercontent.com/942358/80820912-c473a780-8ba5-11ea-83c4-bc186cad3768.png)
[...]

PR #476, opened 2020-05-10
feat: query cache provider

jackmellis:
the query cache now comes from a context, meaning users can easily override and control the cache being used inside their application

see https://github.com/tannerlinsley/react-query/issues/460

this is a basic use case of creating an isolated cache for testing

in component:
```jsx
const MyComponent = () => {
  const queryCache = useQueryCache();

  const [ mutate ] = useMutation(someFn, {
    onSettled() {
      queryCache.refetchQueries('basket');
    }
  });
  return ( ... );
};
```

in unit test:
```jsx
import { makeQueryCache, QueryCacheProvider } from 'react-query';

const cache = makeQueryCache();
sinon.spy(cache, 'refetchQueries');


mount(
  <QueryCacheProvider queryCache={cache}>
    <MyComponent/>
  </QueryCacheProvider>
);

// ...

await invokeSomeMutation(); // after mutating, we should refetch the basket

expect(cache.refetchQueries).to.be.calledWith('basket');
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -2598,7 +2598,7 @@ function App() {
 
 ## `ReactQueryCacheProvider`
 
-`ReactQueryCacheProvider` is an optional provider component for explicitly setting the query cache used by `useQuery`. This is useful for creating component-level caches that are not completely global, as well as making truly isolated unit tests.
+`ReactQueryCacheProvider` is an optional provider component for explicitly setting the query cache used by React Query. This is useful for creating component-level caches that are not completely global, as well as making truly isolated unit tests.
 
 ```js
 import { ReactQueryCacheProvider, makeQueryCache } from 'react-query';
--- a/README.md
+++ b/README.md
@@ -256,8 +256,10 @@ This library is being built and maintained by me, @tannerlinsley and I am always
   - [`queryCache.isFetching`](#querycacheisfetching)
   - [`queryCache.subscribe`](#querycachesubscribe)
   - [`queryCache.clear`](#querycacheclear)
+  - [`useQueryCache`](#usequerycache)
   - [`useIsFetching`](#useisfetching)
   - [`ReactQueryConfigProvider`](#reactqueryconfigprovider)
+  - [`ReactQueryCacheProvider`](#reactquerycacheprovider)
   - [`setConsole`](#setconsole)
 - [Contributors ✨](#contributors-)
 
@@ -2522,6 +2524,18 @@ queryCache.clear()
 - `queries: Array<Query>`
   - This will be an array containing the queries that were found.
 
+## `useQueryCache`
+
+The `useQueryCache` hook returns the current queryCache instance.
+
+```js
+import { useQueryCache } from 'react-query';
+
+const queryCache = useQueryCache()
+```
+
+If you are using the `ReactQueryCacheProvider` to set a custom cache, you cannot simply import `{ queryCache }` any more. This hook will ensure you're getting the correct instance.
+
 ## `useIsFetching`
 
 `useIsFetching` is an optional hook that returns the `number` of the queries that your application is loading or fetching in the background (useful for app-wide loading indicators).
@@ -2581,6 +2595,29 @@ function App() {
 - `config: Object`
   - Must be **stable** or **memoized**. Do not create an inline object!
   - For non-global properties please see their usage in both the [`useQuery` hook](#usequery) and the [`useMutation` hook](#usemutation).
+
+## `ReactQueryCacheProvider`
+
+`ReactQueryCacheProvider` is an optional provider component for explicitly setting the query cache used by `useQuery`. This is useful for creating component-level caches that are not completely global, as well as making truly isolated unit tests.
+
+```js
+import { ReactQueryCacheProvider, makeQueryCache } from 'react-query';
+
+const queryCache = makeQueryCache()
+
+function App() {
+  return (
+    <ReactQueryCacheProvider queryCache={queryCache}>
+      ...
+    </ReactQueryCacheProvider>
+  )
+}
+```
+
+### Options
+- `queryCache: Object`
+  - In instance of queryCache, you can use the `makeQueryCache` factory to create this.
+  - If not provided, a new cache will be generated.
 
 ## `setConsole`
 
--- a/src/queryCache.js
+++ b/src/queryCache.js
@@ -15,20 +15,37 @@ import { defaultConfigRef } from './config'
 
 export const queryCache = makeQueryCache()
 
-export const defaultQueryCacheRef = {
-  current: queryCache
-}
-
 export const queryCacheContext = React.createContext(queryCache)
+
+export const queryCaches = [queryCache]
 
 export function useQueryCache() {
   return React.useContext(queryCacheContext)
 }
 
 export function ReactQueryCacheProvider({ queryCache, children }) {
-  defaultQueryCacheRef.current = queryCache || makeQueryCache()
+  const cache = React.useMemo(() => queryCache || makeQueryCache(), [
+    queryCache,
+  ])
+
+  React.useEffect(() => {
+    queryCaches.push(cache)
+
+    return () => {
+      // remove the cache from the active list
+      const i = queryCaches.indexOf(cache)
+      if (i >= 0) {
+        queryCaches.splice(i, 1)
+      }
+      // if the cache was created by us, we need to tear it down
+      if (queryCache == null) {
+        cache.clear()
+      }
+    }
+  }, [cache, queryCache])
+
   return (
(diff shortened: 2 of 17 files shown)


tannerlinsley:
Nice! I would love to see the tests converted to use this. Thoughts?

jackmellis:
Sure I can update them if you like

tannerlinsley:
If it’s not too much trouble! 👍

jackmellis:
I've updated the tests. I sometimes get a warning about overlapping `act()` calls. I can't figure out why and I'm not sure if it was happening before or if I've uncovered something...

src/setFocusHandler.js
```diff
@@ -9,7 +9,7 @@ const onWindowFocus = () => {
   const { refetchAllOnWindowFocus } = defaultConfigRef.current
 
   if (isDocumentVisible() && isOnline()) {
-    queryCache
+    defaultQueryCacheRef.current
```
tannerlinsley, line 12: Hmmm. Does this feel funny to you? What if a user had multiple query caches in their app? How would they hook up to this? Should multiple caches be allowed?

jackmellis: yes it definitely felt funny! But as you can set up the focus handler outside of any sort of react context, it's very hard to determine which cache to use. I used the same pattern as the config where there is a `defaultConfigRef` that the provider writes over. the focus handler actually uses the config ref as well so it's already a bit of a grey area...

jackmellis: for reference https://github.com/tannerlinsley/react-query/blob/master/src/setFocusHandler.js#L9

Definitely open to cleaner solutions!

Ephem:
Nice work! This also relates to server rendering as per #461, if you want I can take a closer look and test it from that perspective tomorrow. 😄

tannerlinsley:
@Ephem Awesome, since this is clearly cross-cutting, I'll let you test before proceeding.

Also, just to confirm, this will all still be backwards compatible right? With the main goals being:
- Truly isolated tests
- SSR isolation

Confirm?

src/setFocusHandler.js
Ephem: Since this specific config value is "global", it kind of do make sense to use the latest one. For caches, maybe we should keep track of _all active_ caches and update them all? I think that should make it possible to get rid of `defaultQueryCacheRef`?

src/queryCache.js
```diff
@@ -14,6 +15,25 @@ import { defaultConfigRef } from './config'
...
+  current: queryCache
+}
+
+export const queryCacheContext = React.createContext(queryCache)
+
+export function useQueryCache() {
+  return React.useContext(queryCacheContext)
+}
+
+export function ReactQueryCacheProvider({ queryCache, children }) {
```
Ephem, line 28: What happens if `queryCache` changes here?

* If the old prop value was a `queryCache`, it's probably up to the parent to clean it up if necessary, it would be unsafe for this Provider to do so since it might still be in use
* If the old value was undefined, the old cache was created in this Provider, so this provider needs to clean it up

This relates to the more general issue of cleanup which I'll add a separate comment for.

Ephem (changes requested):
First of all, very nice work on this, I like the approach you are taking!

Just so my view on this PR is clear I think the goal should be to support explicit caches for testing, using different caches for different parts of the application/separate React roots and laying the foundation to build improved SSR support on top, but not change SSR behaviour in this PR.

**Tests**

* All internal tests are now updated to the new approach, which is great, but this also means the old approach is currently untested. Maybe add a few key tests targeting using the global cache, without a Provider?
* Maybe add a few tests using multiple caches/providers?

**Query cleanup**

I think an existing major issue right now is cache cleanup. I think the issues #432 and #460 are both related to this and also the act warnings in this PR. I think this could and should be fixed in this PR since the problem might actually become worse otherwise.

EDIT: This could just as well be done in a separate PR before this is merged.

Let me expand:

I'm pretty sure these problems stem from query timeouts not being cleared.

```js
cache.clear = () => {
  cache.queries = {}
  notifyGlobalListeners()
}
```

The old cache clear function does not in any way remove the pending query timeouts for staleness etc. To get rid of these problems, I think calling `cache.clear()` should also synchronously loop through all queries and remove _all_ timeouts related to that query.

* If a user creates a cache manually with `makeQueryCache`, they own the cache and are responsible for calling `cache.clear()`.
* If `react-query` creates a cache, the library owns that cache and needs to clean it up if it is no longer used (for example, in `ReactQueryCacheProvider`).

**Serverside rendering**

This approach looks great from a SSR-perspective and this PR implements a key part of what is needed to improve that story. There is more work that is needed on top, but that should be done in a follow up and I see nothing that wouldn't be compatible!

src/queryCache.js
```diff
@@ -14,6 +15,25 @@ import { defaultConfigRef } from './config'
...
+  defaultQueryCacheRef.current = queryCache || makeQueryCache()
```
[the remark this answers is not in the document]
tannerlinsley: It should only create it once. That looks like it will be created on every render (if the queryCache isn't passed)

src/setFocusHandler.js
jackmellis: I've removed the `defaultQueryCacheRef` in favour of a list of active caches

jackmellis:
> First of all, very nice work on this, I like the approach you are taking!
> 
> Just so my view on this PR is clear I think the goal should be to support explicit caches for testing, using different caches for different parts of the application/separate React roots and laying the foundation to build improved SSR support on top, but not change SSR behaviour in this PR.
> 
> **Tests**
> 
> * All internal tests are now updated to the new approach, which is great, but this also means the old approach is currently untested. Maybe add a few key tests targeting using the global cache, without a Provider?
> * Maybe add a few tests using multiple caches/providers?
> 
> **Query cleanup**
> 
> I think an existing major issue right now is cache cleanup. I think the issues #432 and #460 are both related to this and also the act warnings in this PR. I think this could and should be fixed in this PR since the problem might actually become worse otherwise.
> 
> EDIT: This could just as well be done in a separate PR before this is merged.
> 
> Let me expand:
> 
> I'm pretty sure these problems stem from query timeouts not being cleared.
> 
> ```js
> cache.clear = () => {
>   cache.queries = {}
>   notifyGlobalListeners()
> }
> ```
> 
> The old cache clear function does not in any way remove the pending query timeouts for staleness etc. To get rid of these problems, I think calling `cache.clear()` should also synchronously loop through all queries and remove _all_ timeouts related to that query.
> 
> * If a user creates a cache manually with `makeQueryCache`, they own the cache and are responsible for calling `cache.clear()`.
> * If `react-query` creates a cache, the library owns that cache and needs to clean it up if it is no longer used (for example, in `ReactQueryCacheProvider`).
> 
> **Serverside rendering**
> 
> This approach looks great from a SSR-perspective and this PR implements a key part of what is needed to improve that story. There is more work that is needed on top, but that should be done in a follow up and I see nothing that wouldn't be compatible!

I've added tests to show the global/default cache still works and a few tests for different caching scenarios

The provider now tracks its queryCache value and if it's a self-created one, it cleans it up and removes it on unmount. If the cache is manually created, it's up to the creator to manage cleanup.

I haven't touched on the in-depth cache cleanup (cancelling timeouts etc.) as I don't want to lose the focus of this PR. Happy to take this on separately though if nobody is working on it...

src/queryCache.js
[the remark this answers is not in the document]
jackmellis: I've cleaned this logic up now so it tracks queryCache instances between renders

src/queryCache.js
jackmellis: yup I've addressed this now (per my other comments)

Ephem:
Changes and tests looks great to me, really great work!

I totally agree with addressing cache cleanup in a separate PR. Since it's possible that using a lot of Providers in tests could worsen the existing problems (more timeouts) I think it should probably be addressed before a release.

README.md
```diff
@@ -2582,6 +2596,29 @@ function App() {
   - Must be **stable** or **memoized**. Do not create an inline object!
   - For non-global properties please see their usage in both the [`useQuery` hook](#usequery) and the [`useMutation` hook](#usemutation).
 
+## `ReactQueryCacheProvider`
+
+`ReactQueryCacheProvider` is an optional provider component for explicitly setting the query cache used by `useQuery`. This is useful for creating component-level caches that are not completely global, as well as making truly isolated unit tests.
```
tannerlinsley, line 2601:
```suggestion
`ReactQueryCacheProvider` is an optional provider component for explicitly setting the query cache used by React Query. This is useful for creating component-level caches that are not completely global, as well as making truly isolated unit tests.
```

tannerlinsley (changes requested):

tannerlinsley (approved):

PR #482, opened 2020-05-12
Feature proposal: clearOnNoQuery option for usePaginatedQuery

JaminFarr:
This PR adds a clearOnNoQuery option to usePaginatedQuery. When clearOnNoQuery is true resolvedData will clear if the query key is falsy.

I'm using usePaginatedQuery to provide the data for a search page with paginated results.

The issue I currently have is I want the results to clear when the search is cleared while still being able to paginate through the results. I set the query key to null when the search form is reset but the last searches results still in resolvedData.

The 2 workarounds that I've figured out before this PR are a bit hacky.
1. A special query key that immediately resolves to a special no results value
```js
const { resolvedData } = usePaginatedQuery(
  searchTerm ? ['search', searchTerm] : ['NO-SEARCH'],
  (queryKey) => {
    if (queryKey === 'NO-SEARCH') return Promise.resolve('NO-RESULTS')
  }
  // return real request
)
```
2. Perform the request in a results component that is only mounted when there a search value. I feel this mixes the logic with the display too much.

### DIFF
--- a/src/tests/usePaginatedQuery.test.js
+++ b/src/tests/usePaginatedQuery.test.js
@@ -122,7 +122,7 @@ describe('usePaginatedQuery', () => {
     expect(rendered.getByTestId('status').textContent).not.toBe('loading')
   })
 
-  it('should clear resolvedData data while query is falsy and clearOnNoQuery option is set', async () => {
+  it('should clear resolvedData data when query is falsy', async () => {
     function Page() {
       const [searchTerm, setSearchTerm] = React.useState('')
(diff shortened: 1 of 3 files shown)


tannerlinsley:
Is there any reason why this shouldn't be the default?

JaminFarr:
Not that I know of.
I didn't want to break anyone else's use case.

JaminFarr:
Updated to the default behavior

tannerlinsley:
Nice! Yeah I think that's a good default.

PR #485, opened 2020-05-13
docs(queryCache): illustrate how to use the `getQueries` function to find most recent queries

antgonzales:
Provide a way to get the latest query by a partial query key match. Get the latest data when you don't know the full query key due to a large parameter object, i.e. complicated filters.

### DIFF
--- a/src/queryCache.js
+++ b/src/queryCache.js
@@ -70,6 +70,16 @@ export function makeQueryCache() {
 
   cache.getQueryData = queryKey => cache.getQuery(queryKey)?.state.data
 
+  cache.getLatestQueryData = partialKey => {
+    const query = findQueries(partialKey).reduce(
+      function(prev, current) {
+        return prev.state.updatedAt > current.state.updatedAt ? prev : current
+      },
+      { state: { updatedAt: 0 } }
+    )
+    return query.state.data
+  }
+
   cache.removeQueries = (predicate, { exact } = {}) => {
     const foundQueries = findQueries(predicate, { exact })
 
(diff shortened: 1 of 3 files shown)


antgonzales:
hiya @tannerlinsley 👋 our team is currently using this functionality in order to get the latest query by the simple string part of the cache key. We have a lot of filters influencing a single API call for a search app and thought this might be helpful. Let me know what you think, happy to also add to the README if you don't think this belongs in the API itself 👍

tannerlinsley:
Have you tried using the getQuery method with a predicate? Seems like this is the perfect use case for it?

antgonzales:
@tannerlinsley can you explain using the predicate? not sure I follow

tannerlinsley:
You can pass a function that behaves like an Array.prototype.filter call. You can inspect each query and return true if you want to return it in the array.

tannerlinsley:
Here is a better example:

```js
import { queryCache, deepIncludes } from 'react-query'

const queries = queryCache.getQueries(query => {
  return deepIncludes(query.queryKey, 'data')
})

const [latest] = queries.sort((a, b) => a.state.updatedAt > b.state.updatedAt ? 1 : -1)
``

**Comment by antgonzales** [comment]
Ended up going with this based on your example:
```
  const [latest] = queryCache.getQueries('data').sort((a, b) => {
    return b.state.updatedAt - a.state.updatedAt;
  });
  return latest?.state.data;
```

A few questions/thoughts:

1. Does the `getQueries` function handle the deep includes due to this line?
https://github.com/tannerlinsley/react-query/blob/master/src/queryCache.js#L61
2. I think the tests set up with `prefetchQuery` make timestamps the same in jest. I ended up using a delay in our test suite. 
3. Would you be open to me changing this PR to use this code as an example of how to use the `getQueries` method?

tannerlinsley:
Yeah I think that’s a better direction :)

tannerlinsley:
As this is a doc change, let's close this PR and open a new one when we get to it. :)

PR #495, opened 2020-05-15
fix: clear cache timeouts

jackmellis:
calling queryCache.clear() was empting the queries object but not tearing any of them down
this commit adds a clear method to the individual queries, which cancels timeouts and promises
queryCache.clear now loops through the queries and clears each one
I've also update the return type of queryCache.clear as it doesn't actually return anything

I've not added any tests as this is a very intermittent problem and I wasn't entirely sure if I could reliably reproduce the warnings. All existing tests still pass

this should solve https://github.com/tannerlinsley/react-query/issues/432 and unblock https://github.com/tannerlinsley/react-query/pull/476

### DIFF
--- a/src/queryCache.js
+++ b/src/queryCache.js
@@ -520,10 +520,7 @@ export function makeQueryCache() {
     query.clear = () => {
       clearTimeout(query.staleTimeout)
       clearTimeout(query.cacheTimeout)
-      if (query.cancelQueries) {
-        query.cancelQueries()
-      }
-      query.cancelled = cancelledError
+      query.cancel()
     }
 
     return query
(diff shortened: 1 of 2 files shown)


src/queryCache.js
```diff
@@ -474,6 +475,15 @@ export function makeQueryCache() {
       query.scheduleStaleTimeout()
     }
 
+    query.clear = () => {
+      clearTimeout(query.staleTimeout)
+      clearTimeout(query.cacheTimeout)
+      if (query.cancelQueries) {
```
tannerlinsley, line 481: We could use the new `query.cancel()` method here :)

tannerlinsley (changes requested):

src/queryCache.js
jackmellis: added. I notcied that `query.cancel` isn't in the readme api or in the `.d.ts` definition, which is probably why I missed it...

tannerlinsley:
Nice work!

PR #523, opened 2020-05-28
Dependent query relying on context in suspense mode regression

kradical:
Hello!

I am a huge fan of this library.

Recently while performing a minor version upgrade I hit a bug where my components weren't properly re-rendering.

The regression is here: https://github.com/tannerlinsley/react-query/compare/v1.4.1...v1.4.2.

I added a minimal example and I think the appropriate fix to this PR. I believe when you stopped using "placeholders" for suspended query instances there is nothing that "force renders" on state update. Which afaik is fine in most cases.. but when using dependent queries and the dependency comes out of a react context things don't re-render. 

NOTE: This bug does not appear to be reproducible on experimental react.. only latest (16.13.1). But because we all write real production code we are using a pinned version of react.

If suspense mode is only designed to be used with experimental React we should probably fail horrifically (throw an error) on version mismatch and have a giant disclaimer in the docs or else small bugs like this could continue to crop up.

Let me know if you have any further questions/comments/suggestions.

Edit: Before merging I can remove the example, just wanted to throw it in here so you can reproduce it for yourself.


examples/suspense-context-bug/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "new",
+  "version": "1.0.0",
+  "description": "",
+  "keywords": [],
+  "main": "src/index.js",
+  "dependencies": {
+    "axios": "0.19.2",
+    "react": "16.13.1",
+    "react-dom": "16.13.1",
```
kradical, line 10: NOTE! This is a suspense mode example based on the latest version of react, NOT an experimental version with concurrent mode.

examples/suspense-context-bug/src/index.js
```diff
@@ -0,0 +1,51 @@
...
+    <Suspense fallback="Loading...">
+      <Provider>
+        <Inner />
+      </Provider>
+    </Suspense>
+  </ReactQueryConfigProvider>
+);
+
+const rootElement = document.getElementById("root");
+ReactDOM.render(<Test />, rootElement);
```
kradical, line 51: Here is the minimum example I could get to reproduce the bug.

kradical:
@tannerlinsley when you have time could you (or another maintainer) take a look at this? I would be interested to hear if this is considered a bug, or if we should migrate away from suspense mode on react `16.x`

tannerlinsley:
I plan on looking at this soon. I'm in a sprint for my startup right now, so hang tight :)

tannerlinsley:
Could you perhaps test this out in the next branch? There’s a few breaking changes to the config you’ll want to take into account. Go off the suspense example changes and you should be Good to go.

kradical:
Hey @tannerlinsley it seems like its fixed on the next branch 🎉 . BUT I had to change how I was doing dependent queries because of all the invalidQueryKey() stuff that was added.

I changed:
```javascript
const queryKey = test && ["test"];
const { data } = useQuery(queryKey, getData);
```

Which crashes to:

```javascript
const queryKey = test ? [] : ["test"];
const { data } = useQuery(queryKey, getData);
```

Does that seem correct? or is there a new way to do dependent queries? I checked the docs and looks like they haven't been updated yet.

tannerlinsley:
Yep. There is :) Try this:

```js
const { data } = useQuery('my-key', getData, { enabled: condition });
```

kradical:
Cool! With the new `enabled` syntax things still work 🎉 . I added some logging to double check things were actually changing and stuff like that. I'm convinced this bug is no longer an issue. Looking forward to 2.0!

PR #540, opened 2020-06-02
fix(types): adapt type definition of dataOrUpdater in setQueryData

### LINKED ISSUE #506: typeDefinition of updater function in setQueryData
Hi,

First of all: thanks for this 😎 library. I'm having a quick typescript question:

`setQueryData` takes an `updater` in the form of:

```
((oldData: T | undefined) => T)
```

The question is: if `oldData` is possibly undefined, how would be we able to construct `T`?

suppose we have a simple type:

```
type Person = {
    firstName: string,
    lastName: string
}
```

and we mutate that persons firstName, which we want to optimistically update (I've taken the [optimistic update example](https://github.com/tannerlinsley/react-query/blob/master/examples/optimistic-updates/pages/index.js#L13-L40) as a base):

```
useMutation(
    firstName => axios.patch('/api/person', { firstName }),
    {
      onMutate: firstName => {
        queryCache.cancelQueries('person')

        const previousValue = queryCache.getQueryData('person')

        queryCache.setQueryData('person', old => ({
          ...old,
          firstName
        }))

        return previousValue
      }
  )
```

but if `oldData` is possibly `undefined`, the resulting object can never be of type `Person` - it might miss the `lastName`.

We are currently working around this by doing:

```
const previousValue = queryCache.getQueryData('person')

if (previousValue) {
    queryCache.setQueryData('person', {
        ...previousValue,
        firstName
    })
}
```

The question is: can `old` really be `undefined` ? If so, wouldn't this be a runtime error in the example [here](https://github.com/tannerlinsley/react-query/blob/master/examples/optimistic-updates/pages/index.js#L26) when accessing `old.items` ? Or is this maybe just an issue with the type definitions and `old` can actually never be undefined?

Thanks for clarifying

TkDodo:
closes #506

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -604,7 +604,7 @@ export interface CachedQuery<T, TError = unknown> {
   config: QueryOptions<unknown, TError>
   state: CachedQueryState<T>
   setData(
-    dataOrUpdater: unknown | ((oldData: unknown | undefined) => unknown)
+    dataOrUpdater: unknown | undefined | ((oldData: unknown | undefined) => unknown | undefined)
   ): void
   clear(): void
 }
@@ -686,7 +686,7 @@ export interface QueryCache {
   getQueryData<T = unknown>(key: AnyQueryKey | string): T | undefined
   setQueryData<T = unknown>(
     key: AnyQueryKey | string,
-    dataOrUpdater: T | ((oldData: T | undefined) => T)
+    dataOrUpdater: T | undefined | ((oldData: T | undefined) => T | undefined)
   ): void
   refetchQueries<TResult>(
     queryKeyOrPredicateFn:


TkDodo:
Not sure what I should do against the broken link checker, this seems unrelated:

```
HTTPSConnectionPool(host='upc.ch', port=443): Max retries exceeded with url: / (Caused by SSLError(SSLError("bad handshake: Error([('SSL routines', 'tls_process_server_certificate', 'certificate verify failed')])")))
https://upc.ch
```

Also, some more questions please:
1. is `dtslint` part of the PR checks? I am getting a bunch of errors locally:

<details>
  <summary>see `dtslint` errors</summary>

```
ERROR: 621:44  no-unnecessary-generics  Type parameter TError is used only once. See: https://github.com/Microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md
ERROR: 632:44  no-unnecessary-generics  Type parameter TError is used only once. See: https://github.com/Microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md
ERROR: 649:44  no-unnecessary-generics  Type parameter TError is used only once. See: https://github.com/Microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md
ERROR: 661:44  no-unnecessary-generics  Type parameter TError is used only once. See: https://github.com/Microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md
ERROR: 683:44  no-unnecessary-generics  Type parameter TError is used only once. See: https://github.com/Microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md
ERROR: 686:57  no-unnecessary-generics  Type parameter T is used only once. See: https://github.com/Microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md
ERROR: 701:14  no-unnecessary-generics  Type parameter TResult is used only once. See: https://github.com/Microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md
```
</details>

2. Should I add some tests to `types/test.ts` for the type amendments I have made?

3. the `format` script doesn't format `.ts` files (and doesn't run on the `types` directory in general), which is why the `types/test.ts` file is not formatted with it. Is this worth addressing or is this maybe on purpose?

Thanks

tannerlinsley:
1. It's not, but I'd like to get them fixed anyway. Low priority.
2. If you'd like! More tests can't hurt.
3. Again, I would love to update that, but it's low priority.

PR #569, opened 2020-06-09
Small fix for the Basic sample

weyert:
I have made a small change to the basic example so it's not using a third-party data retrieval library Axios as it's quite confusing as looking at the example you would think using `fetch` can be used and `useQuery` would return the data of the request. This caught me off guard and I think as first example I think it's worth to keep it simple by using the `fetch`-function instead.

As running the example failed because it's missing the `eslint-config-prettier`-package I have added that too

### DIFF
--- a/examples/basic/package.json
+++ b/examples/basic/package.json
@@ -20,7 +20,8 @@
     "@rescripts/cli": "^0.0.11",
     "@rescripts/rescript-use-babel-config": "^0.0.8",
     "@rescripts/rescript-use-eslint-config": "^0.0.9",
-    "babel-eslint": "10.0.1"
+    "babel-eslint": "10.0.1",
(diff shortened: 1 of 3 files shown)


tannerlinsley:
If we move to fetch these examples, we'll also need to illustrate how to properly throw errors using fetch (one of the initial reasons why axios is nice, since it throws automatically on 404-like error codes). Check this out: https://stackoverflow.com/questions/39297345/fetch-resolves-even-if-404#:~:text=3%20Answers&text=A%20fetch()%20call%20is,the%20promise%20point%20of%20view.

weyert:
Hmm, okay, I have never used Axios before so I found the code a bit confusing. Maybe it's better then that I update the README to make it more clear how the hooks work without needing a deep dive :)

PR #570, opened 2020-06-10
WIP: Improve support for server side rendering

### LINKED ISSUE #461: Support per-request cache with SSR

Ephem:
Closes #461

> This PR should be complete, but I marked it as WIP since I haven't had time to verify everything a 100% myself and will soon be on a weekisch long vacation. Since it's a largeish change in scope and I expect it will take some time to review and think about I wanted to get it up as soon as possible.

This PR adds:

- `makeServerQueryCache()` - Create a `queryCache` that caches queries in a server environment
- `queryCache.dehydrate()` - A method that outputs a serializable version of the cache
-  A `initialQueries`-option to `makeQueryCache` that creates a pre-warmed cache from dehydrated queries
- Tests, docs and an example for the above

Together, these new APIs form the missing pieces for proper SSR-support. Docs and the example is a good place to start reviewing.

Caching on the server isn't a huge change conceptually now that caches can be provided via `ReactQueryCacheProvider`, but I think the de/rehydration parts require some close scrutiny and thinking about.

**Caveats**

- `initialQueries` are scheduled for stale timeout from the time of client cache creation + `staleTime`, not from the time they were fetched on the server. Using the server time is tricky since server and client time might be out of sync.

**Breaking vs non-breaking and an alternative approach**

This PR is not a breaking change since only caches created by the new `makeServerQueryCache` caches data on the server. An alternative approach that would technically be a breaking change is to enable query caches returned by the regular `makeQueryCache` to also cache data on the server (but still not the global one).

The reason this would be breaking is that it is today technically possible to create such a cache on the server and do `queryCache.prefetchQueries` on that cache, using the returned result, but not expecting it to end up in the cache.

**Missing/Future/Follow up?**

- Types
- Improved devTools-support?
  - Seeing which queries were rehydrated via `initialQueries`?
  - I haven't used the devtools myself, so there might be opportunities I'm missing
- More examples (Next, Gatsby?)
- Improve SSR-docs with different scenarios/cases?
  - For example, when caching the server output, you can call `queryCache.refetchQueries` _after_ hydration to trigger a refetch for up to date data
- Provide a way to _configure_ automatically refetching data after hydration?
- Add server Suspense support (currently not supported by the official renderer, only with experimental things like [react-ssr-prepass](https://github.com/FormidableLabs/react-ssr-prepass) or [react-lightyear](https://github.com/Ephem/react-lightyear))


src/queryCache.js
```diff
@@ -526,6 +581,34 @@ export function makeQueryCache() {
...
+      })
+
+      cache.queries[queryHash].wasPrefetched = true
+
+      if (!isServer) {
+        cache.queries[queryHash].scheduleStaleTimeout()
+        // Schedule for garbage collection in case
+        // nothing subscribes to this query
+        cache.queries[queryHash].scheduleGarbageCollection()
+      }
```
Ephem, line 608: Upon reflection, this block is a mistake. Side effects when creating things isn't very good practice and for Next for example we want `makeQueryCache` to be able to run inside of render.

One solution would be to move this into an init-function and let `ReactQueryCacheProvider` check if the cache has been/needs to be initialized, in that case that happens in an effect. A custom cache should always be paired with a cache provider, so this should provide an equally simple API for users, but be a safe way to do it.

Another option would be to add the prop `initialQueries` directly to the cache provider to avoid having to create a store at all on the client, which could be an ever nicer API?

tannerlinsley:
What if all of the server side stuff was in its own import? react-query/ssr?

Ephem:
> What if all of the server side stuff was in its own import? react-query/ssr?

Again a great question, I think so and it got me thinking about a slightly alternative approach. This PR in essence does two things:

- Adds an extra "serverEnabled" conditional for "should I save this query to cache or not"

After doing some thinking, maybe the concept of "frozen" vs "unfrozen" caches makes more sense? All caches start as frozen on the server by default, but you can `.unfreeze()` them (and even `.freeze()` them again?). This part needs to stay in the `queryCache` to avoid a breaking change, but is very minimal, non-intrusive and doesn't really add complexity.

- The big one: Adds de/rehydrate functionality

I think this one could probably use a functional approach instead (naming could need work):

```js
import { makeQueryCache } from 'react-query'
import { dehydrateCache, rehydrateCache } from 'react-query/some-scope' 

// Next line would look a little different without a breaking change:
const serverCache = makeQueryCache()
const dehydratedQueries = dehydrateCache(serverCache)

const clientCache = makeQueryCache()
rehydrateCache(clientCache, dehydratedQueries)
```

One might start questioning why this should even be part of React Query in that case, but this would rely on internals, so it would be unsafe to implement in user-space.

A version closer to what you are suggesting is:

```js
import { makeQueryCache } from 'react-query/ssr'

const serverCache = makeQueryCache()
const dehydratedQueries = serverCache.dehydrate()

// Note that this needs to use the same import since rehydration is also included there:
const clientCache = makeQueryCache({ initialQueries: dehydratedQueries })
```

Both have merits and I like both better than the existing PR both to avoid increasing payload for the default use case unnecessarily and to separate concerns. What do you think?

tannerlinsley:
The functional approach is really cool actually. Imagine if someone wanted to dehydrate a client side cache to store in local storage? I totally agree on the bundle savings by keeping it separate. Can’t rely on tree shaking in every environment so this would draw a clear boundary between the core and such a utility.

tannerlinsley:
Also, with the future in mind, we can really focus on improving the size and integration even more when we choose to do 3.0

Ephem:
> The functional approach is really cool actually. Imagine if someone wanted to dehydrate a client side cache to store in local storage?

Yeah, this is another use case I have been thinking about, this could help solve scroll restoration on reload among other things. :)

If you approve, I think I'll go ahead and split this PR when I have time:

* Add frozen/unfrozen concept
* Add separate entry with de/rehydration
* Examples

tannerlinsley:
Sounds like a good plan :)

tannerlinsley:
Also, if there are things you think would be better as a breaking change, I would be fine talking about them. I would rather do things right than avoid backwards compatibility at the cost of strangeness.

tannerlinsley:
Surprise! The `next` branch is ready for testing and I think we should push this change to the next branch and also make any breaking changes to it to make this all work even better.

Ephem:
Great! I created a new PR with the breaking change I mentioned. It still uses the `frozen` concept internally, but that doesn't need to be exposed to users now that we are targeting a major. 🎉 

I'll go ahead and close this PR now that it's being broken apart into pieces instead. I added a summary of the next steps to the issue. 😄

PR #577, opened 2020-06-12
fix(types): enable more types for first QueryKey

CreativeTechGuy:
This change enables more types beyond just `string` for the first QueryKey but still ensures that it's not `null`/`undefined`.

* The current type is too strict
* The `queryKeySerializerFn` doesn't play well with TypeScript when removing the first key so TypeScript users will benefit from being able to move their string key to the end of the list instead of having to add an additional argument to their functions.

This change will enable the following TypeScript code:

```ts
function doAdd(num: number): Promise<number> {
    return Promise.resolve(num + 1);
}
useQuery<number, [number, string]>([10, "adder"], doAdd);
```

Before this change you'd have to modify your function to accept an additional argument which was unnecessary:

```ts
function doAdd(key: string, num: number): Promise<number>{
    return Promise.resolve(num + 1);
}
useQuery<number, [string, number]>(["adder", 10], doAdd);
```

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -339,17 +339,23 @@ export function useInfiniteQuery<
   config?: InfiniteQueryOptions<TResult, TMoreVariable, TError>
 ): InfiniteQueryResult<TResult, TMoreVariable, TError>
 
+export type DefinedQueryKeyPart = 
+  | string
+  | object
+  | boolean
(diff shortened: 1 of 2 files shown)


CreativeTechGuy:
Updated tests file to reflect these changes

PR #583, opened 2020-06-14
chore: upgrade dependencies

CreativeTechGuy:
For the past several weeks CI has been failing at the installation stage with the error:

```
[!] (plugin babel) Error: Cannot find module '@babel/compat-data/corejs3-shipped-proposals'
```

In an attempt to fix this, I opted to update the dependencies to see if that solves the issue. All dependencies are updated to the latest versions. This mostly affects the tests which needed to be updated as it now used deprecated features. **I diffed the output bundles before and after the change and there was no meaningful difference.**

Note, there is one failing `useQuery` test (`should reset error state if new component instances are mounted`) which I presume is failing due to a timing issue with this [setTimeout](https://github.com/tannerlinsley/react-query/blob/master/src/utils.js#L132-L134). I was not able to fix this test but at least now we should be able to see accurate CI results that it's failing and can follow up with fixing the test or the library. This test error was introduced with this upgrade, but only in the unit tests as the output bundle has no change.


CreativeTechGuy:
Interesting, the preactjs/compressed-size-action still fails with the same error during the build step but the node 12 ubuntu-latest test successfully installs and fails at the `useQuery` test as expected (see above).

Still needs work, but I think it's closer. @tannerlinsley where do we go from here?

tannerlinsley:
Let’s delete the size workflow. Useless tbh

CreativeTechGuy:
I'm glad you said that because I didn't understand it either haha. But do you have any idea how to fix the suspense hook? I wasn't able to reproduce the problem in a codesandbox so my best guess is that it's a issue with one of the test libraries and possibly re-rendering twice before the `setTimeout` is able to fire.

The best I was able to do was:

```
fireEvent.click(rendered.getByText('retry'))
await sleep(100)
fireEvent.click(rendered.getByText('retry'))
```

But this seems like a really bad solution as it's no longer a representative test of reality.

tannerlinsley:
Yeah, there’s something tricky going on for sure. I think that’s fine for now. We know it works in the wild

CreativeTechGuy:
I don't think we should leave tests failing that we expect to fail. That's a really bad impression to give off.

I'd vote to mark the suspense features in the docs as experimental (because suspense for data fetching itself is experimental) and disable (or rewrite) the tests for now. If someone wants to use those features it's buyer beware. The concern with this approach is that there may be people using the feature in production and losing confidence by removing tests isn't good.

tannerlinsley:
Let’s just do what we need to do to make it pass for now

CreativeTechGuy:
Removed workflow, added workaround for test with explanation.

The CI tests now succeed!

tannerlinsley:
This is great! Would you mind applying this to the next branch too?

CreativeTechGuy:
What is the `next` branch? I couldn't find the process for that documented of what goes there.

tannerlinsley:
It’s the next major version, but under construction.

PR #584, opened 2020-06-14
Cache data on the server with manually created caches

Ephem:
This is a breakout from #570 and replaces `makeServerQueryCache` in that PR. More background can be found there.

Before this PR, queries wont ever get cached in a server environment to avoid leaking sensitive information between requests (#70). This PR makes it so any cache that has been manually created via `makeQueryCache` will cache data, even on the server. 

This change wont improve SSR (much) by itself but is a prerequisite. I'm pretty sure this PR is the only breaking change needed and that de/rehydration can be achieved without breaking changes.

This also adds some extra guards for `scheduleStale` and `scheduleGarbage`, since I don't believe these should ever be called on the server. Lifetime on the server should be manually managed (usually lifetime should be a single request).

I saw `defaultConfig` was added to `makeQueryCache`. I needed an extra `frozen`-option that I don't think makes sense as a cache config, but rather an option to `makeQueryCache` itself, so I took the liberty of changing the signature into an `options`-object instead if that's ok?

I'm not sure if the `frozen`-option should be documented and public or not and it would also be trivial to add `.freeze()` and `.unfreeze()`-methods, but I'm not sure if I see any use-cases for exposing this?

Just a heads up: I'm AFK for about a week from tomorrow, so while I can read and respond, actually fixing comments will take some time. 😄  I'll tackle next steps after that.


tannerlinsley:
Alright, the next branch is pretty darn stable. If you could would you mind resolving any conflicts and updating any uses of `makeQueryCache` in the repo to take it into account?

tannerlinsley:
NM, I got it merged :)

PR #587, opened 2020-06-17
Add config parameter to `setQueryData` call

PepijnSenders:
The `setQueryData` call also accepts a config object. This was not defined in the types yet.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -424,6 +424,10 @@ export interface PrefetchQueryOptions<TResult, TError = Error> extends QueryOpti
   throwOnError?: boolean
 }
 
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -690,9 +694,10 @@ export interface QueryCache {
   }): Promise<TResult>
 
   getQueryData<T = unknown>(key: AnyQueryKey | string): T | undefined
-  setQueryData<T = unknown>(
```
PepijnSenders, line 693: Not sure if changing the `<T = unknown>` to `<TResult, TError>` could be problematic here.

PR #589, opened 2020-06-18
Refetch interval fix

relayrlukaszmakuch:
Hi! 👋 

Auto refetching is great! 👍

This PR contains a fix for the timers used to handle auto refetching.

As you can see in the added test case, if 2 components were using the same resource, hiding one of them was breaking auto refetching for the other component.

P.S. I've just seen this PR https://github.com/tannerlinsley/react-query/pull/585 and now I'm not sure if mine makes sense. 😅

Please let me know what do you think.

Cheers! 🙂

### DIFF
--- a/src/queryCache.js
+++ b/src/queryCache.js
@@ -10,6 +10,7 @@ import {
   getQueryArgs,
   deepIncludes,
   noop,
+  Console,
 } from './utils'
 import { defaultConfigRef } from './config'
 
@@ -324,6 +325,52 @@ export function makeQueryCache() {
       query.cancelled = null
     }
 
+    // These values are used to handle auto refetching.
+    
+    // The result of calling setInterval.
+    let currentRefetchIntervalId = null;
+    // The current interval in milliseconds .
+    let currentRefetchInterval = null;
+    // A map like { instanceId: true, anotherInstanceId: true }.
+    // If an instance ID is present among the keys,
+    // it means that instance expects auto refetching to work
+    // in the background.
+    let instancesRequestingBackgroundRefetch = {}
+    // Each instance writes here its desired interval, like
+    // { instanceId: 300, anotherInstanceId: 500 }
+    let refetchIntervalsRequestedByInstances = {}
+
+    // This function should be called every time some preferences
+    // regarding auto refetching change.
+    // It controls the timing and behavior of auto refetching.
+    const adjustRefreshInterval = () => {
+
+      // The shortest requested interval wins.
+      const requestedInterval = (
+        Object.values(refetchIntervalsRequestedByInstances).length > 0
+          ? Math.min(...Object.values(refetchIntervalsRequestedByInstances))
+          : null
+      );
+
+      // If the interval has changed since the last time we checked.
+      if (requestedInterval !== currentRefetchInterval) {
+        // Stop the old auto refetching.
+        clearInterval(currentRefetchIntervalId);
+
+        // If the new interval is truthy, enable the new auto refetching.
+        if (requestedInterval) {
+          currentRefetchIntervalId = setInterval(() => {
+            // At least one instance expects auto refetching to work in the background.
+            const shouldRefetchInBackground = Object.keys(instancesRequestingBackgroundRefetch).length > 0;
+            if (isDocumentVisible() || shouldRefetchInBackground) {
+              query.refetch().catch(Console.error)
+            }
+          }, requestedInterval);
+          currentRefetchInterval = requestedInterval;
+        }
+      }
+    }
+
     query.cancel = () => {
       query.cancelled = cancelledError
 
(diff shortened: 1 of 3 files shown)


tannerlinsley:
This is quite a bit of code. I’m wondering if you can confirm if this is fixed for you in 2.0? If it is, then I would rather backport that fix as it’s much smaller.

relayrlukaszmakuch:
> This is quite a bit of code. I’m wondering if you can confirm if this is fixed for you in 2.0? If it is, then I would rather backport that fix as it’s much smaller.

I've just run this test https://github.com/tannerlinsley/react-query/pull/589/files#diff-79cafc188e2c2ae00a832f0ce03a41e6 against https://github.com/tannerlinsley/react-query/pull/585 and it seems that the shortest interval is not respected. 

One of the consumers is requesting auto refetching at the rate of 300 ms:

```js
    function NumberConsumer1() {
      const { data: number } = useQuery(
        ['number'],
        fetchNumber,
        { refetchInterval: 300 }
      );
      return `consumer 1: ${number}`
    }
```

Sadly, nothing is happening

```
  ● useQuery › respects the shortest active refetch interval

    expect(received).toMatch(expected)

    Expected pattern: /consumer 1: 1/
    Received string:  "consumer 1: 0consumer 2: 0hide consumer 1"

      824 |     await sleep(301);
      825 | 
    > 826 |     expect(container.textContent).toMatch(/consumer 1: 1/)
          |                                   ^
      827 |     expect(container.textContent).toMatch(/consumer 2: 1/)
      828 | 
      829 |     // await sleep(301);
```

When I extend the sleep to over 1 second, this particular assertion passes. That suggest the longer (last?) is the one taken into account:
```js
    function NumberConsumer2() {
      const { data: number } = useQuery(
        ['number'],
        fetchNumber,
        { refetchInterval: 1000 }
      );
      return `consumer 2: ${number}`
    }
```

To sum it up, at the moment the _next_ branch introduces a breaking change regarding the most important interval.

I'll tweak the test to check if the problem I was describing is fixed.

relayrlukaszmakuch:
I've checked and in https://github.com/tannerlinsley/react-query/pull/585 removing a component subscribing to a query doesn't break the interval! 👍 

Sadly, adding one component may break it.

Here's an example of a test with one component. It passes:
```js
  it.only('respects the shortest active refetch interval', async () => {
    let currentNumber = 0;
    const fetchNumber = () => currentNumber++;

    const NumberConsumer1 = function () {
      const { data: number } = useQuery(
        ['number'],
        fetchNumber,
        { refetchInterval: 300 }
      );
      return `consumer 1: ${number}`
    }

    function Page() {
      return (<div>
        <NumberConsumer1 />
      </div>);
    }

    const { container } = render(
      <ReactQueryConfigProvider config={{ suspense: true }}>
        <React.Suspense fallback={"loading"}>
          <Page />
        </React.Suspense>
      </ReactQueryConfigProvider>
    );

    await waitFor(() => {
      expect(container.textContent).toMatch(/consumer 1: 0/)
    });

    await waitFor(() => {
      expect(container.textContent).toMatch(/consumer 1: 1/)
    });
  })
```

But this test with two components fails:
```js
  it.only('respects the shortest active refetch interval', async () => {
    let currentNumber = 0;
    const fetchNumber = () => currentNumber++;

    const NumberConsumer1 = function () {
      const { data: number } = useQuery(
        ['number'],
        fetchNumber,
        { refetchInterval: 300 }
      );
      return `consumer 1: ${number}`
    }

    const NumberConsumer2 = function () {
      const { data: number } = useQuery(
        ['number'],
        fetchNumber
      );
      return `consumer 2: ${number}`
    }

    function Page() {
      return (<div>
        <NumberConsumer1 />
        <NumberConsumer2 />
      </div>);
    }

    const { container } = render(
      <ReactQueryConfigProvider config={{ suspense: true }}>
        <React.Suspense fallback={"loading"}>
          <Page />
        </React.Suspense>
      </ReactQueryConfigProvider>
    );

    await waitFor(() => {
      expect(container.textContent).toMatch(/consumer 1: 0/)
      expect(container.textContent).toMatch(/consumer 2: 0/)
    });

    await waitFor(() => {
      expect(container.textContent).toMatch(/consumer 1: 1/)
      expect(container.textContent).toMatch(/consumer 2: 1/)
    });
  })
```

It fails, because `NumberConsumer2` doesn't ask for auto refetching. And because it appears after `NumberConsumer1`, it overrides `NumberConsumer1`'s preferences to enable auto refetching.

The same issue may exhibit itself in a way described in the message above, when the interval is not properly updated as components appear and disappear from the screen.

The way I approached it in this PR is to keep track of the interval preferences of every instance. That way, when one component which asked for refetches every second unmounts, another component's preferences become active.

tannerlinsley:
Hmmm. I think what I would like to do is replicate this PR's behavior, but in the new system for 2.0

tannerlinsley:
I just pushed a fix for all of this to the next branch. Please try it out! :)

tannerlinsley:
https://github.com/tannerlinsley/react-query/commit/b9168b83f1c10763d29d730a6f3527bcb930ebad

tannerlinsley:
If your tests pass, feel free to PR them to the next branch :)

relayrlukaszmakuch:
I think it's going in the right direction. 👍 We just need a couple more adjustments to make it work. 

I've posted some comments in https://github.com/tannerlinsley/react-query/commit/b9168b83f1c10763d29d730a6f3527bcb930ebad

tannerlinsley:
The latest changes in 286be394d03a07a1ea8d1175885f709142cd6b5e should take care of the background fetching issue. Confirm?

tannerlinsley:
Let's track more conversations in #585

PR #597, opened 2020-06-20
RFC - define query types

JakeGinnivan:
This is just an idea so haven't taken it very far. 

## The goal

To allow you to define the available keys and data types once and be enforced throughout the API

## Example

By putting this type in your code:

```
declare module 'react-query' {
  export interface QueryTypes {
    'typed-todo': TodoItem
  }
}
```

useQuery is now typed and the return type is enforced.

image

image

## Current issues

I have not yet figured out how to keep compatibility, the current approach will require the user to either define their types, or opt into the existing behavior by adding the following to their code.

```
declare module 'react-query' {
  export interface QueryTypes extends Record<string, any>
}
```

And also going through the entire public API to use the QueryTypes, we could use it to type the cache functions too

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -7,6 +7,19 @@
 import * as React from 'react'
 import * as _ from 'ts-toolbelt'
 
+  /**
+   * This allows global typing of React query functions
+   * @example
+   *
+   * declare module 'react-query' {
+   *   export interface QueryTypes {
+   *    todos: TodoItem[]
+   *   }
+   * }
+   **/
+  export interface QueryTypes {
+  }
+
 // overloaded useQuery function
 export function useQuery<TResult, TKey extends AnyQueryKey, TError = Error>({
   queryKey,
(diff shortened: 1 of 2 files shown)


tannerlinsley:
This is a cool idea. I don't use types a lot, so this would just need to be vetted by a bunch of people that we trust and a decision made. 👍

JakeGinnivan:
What is the best way to make that happen?

Jump into discord or could you tag a few people here?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

tannerlinsley:
Woah... I legit don't remember closing this. We should reopen it.

tannerlinsley:
And yes, let's tag a few people here.

tannerlinsley:
I think another way (if my TS knowledge is sufficient) is that you could maybe define your own query type for `useQuery` that extends the original useQueryType, but with a specific key? That way you could reference that query type in your code instead of the generic one. I realize that doesn't restrict people from using the open-ended useQuery though. Does that sound remotely right?

PR #607, opened 2020-06-23
Update Table of Contents

Xadeka:
The "Manual Querying" link seems to have been left over from the previous major version. I believe this has been replaced with the `enabled` config option which has a good example under [Disabling or Pausing a Query](https://github.com/tannerlinsley/react-query#disabling-or-pausing-a-query). Because of that, I don't think this section is still needed.


tannerlinsley:
Hmm actually, can you run `yarn doctoc`?

Xadeka:
Sorry! I overlooked the "DO NOT EDIT" warning 😅  

Looks like that fixed a few others too!

tannerlinsley:
Nice!

PR #627, opened 2020-06-24
don't schedule garbage collection unnecessarily

kentcdodds:
This doesn't actually fix the issue I'm looking into, but I did notice that this timeout is scheduled even when the query is cleared and I *think* it shouldn't be. If we do need to garbage collect something, that should happen in `clear` I think.

Feel free to ignore this PR. This is just a drive-by note more than a real contribution.


tannerlinsley:
I like it.

PR #633, opened 2020-06-25
fix(types): prefetchQuery

kizivat:


wolverineks:
its not valid to have 2 optional parameters in a row

wolverineks:
my initial thought (just get it done), copy and paste all the function definitions and manually define each possible combination
but there are a bunch now, even the query function is optional, only the queryKey is required

types/index.d.ts
```diff
@@ -600,7 +602,8 @@ export interface QueryCache {
       | undefined
       | (() => TKey | false | null | undefined),
```
wolverineks, line 603:
```suggestion
```

types/index.d.ts
```diff
@@ -611,7 +614,8 @@ export interface QueryCache {
...
```
wolverineks, line 615:
```suggestion
```

types/index.d.ts
```diff
@@ -627,7 +631,8 @@ export interface QueryCache {
       | undefined
       | (() => TKey | false | null | undefined)
```
wolverineks, line 632:
```suggestion
```

types/index.d.ts
```diff
@@ -578,7 +578,8 @@ export interface QueryCache {
       | undefined
       | (() => TKey | false | null | undefined),
```
wolverineks, line 579:
```suggestion
```

types/index.d.ts
```diff
@@ -589,7 +590,8 @@ export interface QueryCache {
...
```
wolverineks, line 591:
```suggestion
```

wolverineks:
i think 2 of the function overloads might be duplicates

kizivat:
> its not valid to have 2 optional parameters in a row

Wait. It isn't?

This looks OK to me: https://www.typescriptlang.org/play/?ssl=10&ssc=18&pln=1&pc=1#code/GYVwdgxgLglg9mABMMAKAhgLkWEBbAIwFMAnAGkQIH5tdDSKIad9iSBKbANzhgBMA3ACghoSLATI0WRAGcoJGGADmFarVYNETDfQ7degkWOjwkKVADpr6EstnYA2vMUrEAHxZ6KdNlR+aJFQAupyIPPyIAN5CiHHaCLJwADZElslwyhh2so4ADMEUtvaOAIyFiMW5AEyhwgC+xmilFNUUAMzsomgA5AAWRMkZPa0d7EA

Can you, please, point me to some docs that elaborate this?

wolverineks:
if you dont pass in the 1st optional argument, but do pass in the 2nd, then the 2nd argument is in the wrong place

wolverineks:
can you move these into the other PR
https://github.blog/2018-10-16-future-of-software/#suggested-changes-public-beta

kizivat:
> can you move these into the other PR
> https://github.blog/2018-10-16-future-of-software/#suggested-changes-public-beta

I tried to but it doesn't seem I can comment or make suggestions outside of a very narrow context of the lines that were changed in the PR

kizivat:
> plus, tanner already put in the code needed to make it work without the placeholder
> https://discord.com/channels/719702312431386674/719702452861009930/725511890196889721

What server is this please? I'm not in it.

kizivat:
@wolverineks As for the two optionals:

I see what you're saying and definitely agree. However, as I'm looking into the implementation it doesn’t look like you could get away with passing the prefetch options (e.g. the `{ force, throwOnError }` obj, as the third argument.

https://github.com/tannerlinsley/react-query/blob/ce25ffa5dac431ca4cf965b3f99c5efb519a97b2/src/queryCache.js#L239

In the above line the `queryArgs(args)` would return the obj into the `config` var, thus it wouldn't work as expected. At least that's how I see it. Correct me if I'm wrong.

tannerlinsley:
You'll need to dive into the `getQueryArgs` function to see how it remaps some things.

kizivat:
Yeah, I see now. Now that that's clear to me and there are some changes around those lines in the original PR, I'll do the suggestions there and close this PR afterwards.

PR #640, opened 2020-06-26
fix(query-cache): Prevent re-run of query for refetch interval if update is disabled

bishwenduk029:
This PR resolves the issue: https://github.com/tannerlinsley/react-query/issues/635.

### DIFF
--- a/src/core/queryCache.js
+++ b/src/core/queryCache.js
@@ -398,8 +398,7 @@ export function makeQueryCache({ frozen = isServer, defaultConfig } = {}) {
                 isDocumentVisible() ||
                 query.instances.some(
                   instance =>
-                    instance.config.refetchIntervalInBackground &&
-                    instance.config.enabled
+                    instance.config.refetchIntervalInBackground
                 )
               ) {
                 query.fetch()
--- a/src/core/queryCache.js
+++ b/src/core/queryCache.js
@@ -394,6 +394,7 @@ export function makeQueryCache({ frozen = isServer, defaultConfig } = {}) {
           ) {
             instance.refetchIntervalId = setInterval(() => {
               if (
+                query.instances.some(instance => instance.config.enabled) &&
                 isDocumentVisible() ||
                 query.instances.some(
                   instance =>
--- a/src/core/queryCache.js
+++ b/src/core/queryCache.js
@@ -396,7 +396,9 @@ export function makeQueryCache({ frozen = isServer, defaultConfig } = {}) {
               if (
                 isDocumentVisible() ||
                 query.instances.some(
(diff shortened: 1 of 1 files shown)


src/core/queryCache.js
```diff
@@ -396,7 +396,9 @@ export function makeQueryCache({ frozen = isServer, defaultConfig } = {}) {
               if (
                 isDocumentVisible() ||
                 query.instances.some(
-                  instance => instance.config.refetchIntervalInBackground
+                  instance =>
+                    instance.config.refetchIntervalInBackground &&
+                    instance.config.enabled
```
tannerlinsley, line 401: This check needs to happen before the is document visible. If the query doesn’t have an enabled instance, then we should never even reach these other conditions.

tannerlinsley (changes requested):

src/core/queryCache.js
```diff
@@ -396,7 +396,9 @@ export function makeQueryCache({ frozen = isServer, defaultConfig } = {}) {
               if (
                 isDocumentVisible() ||
```
tannerlinsley, line 397:
```suggestion
                query.instances.some(instance => instance.config.enabled) &&
                isDocumentVisible() ||
```

src/core/queryCache.js
```diff
@@ -394,9 +394,12 @@ export function makeQueryCache({ frozen = isServer, defaultConfig } = {}) {
           ) {
             instance.refetchIntervalId = setInterval(() => {
               if (
+                query.instances.some(instance => instance.config.enabled) &&
                 isDocumentVisible() ||
                 query.instances.some(
-                  instance => instance.config.refetchIntervalInBackground
+                  instance =>
+                    instance.config.refetchIntervalInBackground &&
+                    instance.config.enabled
```
tannerlinsley, line 402:
```suggestion
                    instance.config.enabled
```
```suggestion
                    instance.config.refetchIntervalInBackground
```

src/core/queryCache.js
```diff
@@ -394,9 +394,11 @@ export function makeQueryCache({ frozen = isServer, defaultConfig } = {}) {
           ) {
             instance.refetchIntervalId = setInterval(() => {
               if (
+                query.instances.some(instance => instance.config.enabled) &&
                 isDocumentVisible() ||
                 query.instances.some(
-                  instance => instance.config.refetchIntervalInBackground
+                  instance =>
+                    instance.config.refetchIntervalInBackground
```
tannerlinsley, line 401:
```suggestion
                  instance => instance.config.refetchIntervalInBackground
```

tannerlinsley (approved):

PR #641, opened 2020-06-26
fix: multiple mutations not calling callbacks each time

wtanna:
By having `isLatest` be a function it was only setting the `data` if it
was the last mutation being run, which then called all the callbacks with that 
one set of data. With this adjustment every mutation will call every callback.

This change fixes this issue: https://github.com/tannerlinsley/react-query/issues/241


tannerlinsley:
I fixed this in a bit of a different way, but the test here is great. Can you isolate the test and drop the rest of the changes and then we can merge this?

Issue #642, opened 2020-06-27
enabled flag doesn't seem to stop query from happening when suspense: true


bugzpodder:
```
const url = `/api/user/posts?user_id=${userId}`;
  const { data } = useQuery(url, () => ky.get(url).json(), { enabled: userId });
```
I also tried enabled: false and it doesn't seem to stop the query from going off.

Edit: suspense: true needs to be enabled.

tannerlinsley:
I’ll take a closer look

bugzpodder:
```
if (queryInfo.query.config.suspense && queryInfo.status !== statusSuccess) {
      queryInfo.query.wasSuspended = true;
      throw queryInfo.query.fetch();
    }
```
I think this piece of code needs to check queryInfo.query.config.enabled?

bugzpodder:
@all-contributors please add @bugzpodder for code, bug

PR #643, opened 2020-06-27
fix disabling query in suspense mode

bugzpodder:
#642


tannerlinsley:
Bingo da dingo! This is great.

Issue #639, opened 2020-06-26
useMutation's onSuccess promise never resolves due to setQueryData and refetchQuery edge case


franleplant:
**Describe the bug**

I have created a very contrived example reproduction code that you can find here:
https://github.com/franleplant/react-query-unresolved-promises/blob/master/src/App.tsx

I say contrived because it doesn't represent real use case but I first found this error
in a closed source company app where, as you probably know, things started to get messy
and there were some interactions between mutation's onSuccess and refetchQueries.

To be a bit more concrete, when we save `EntityA` we refetch `EntityA` (because of the way our particular backend works) but since we weren't using `exact: true` it was causing it to refetch `EntityA1` via a query that wasn't initialized, so by a semi complex interactions of the query of `EntityA` we also did `setQueryData` of `EntityA1` which caused the problem I explain below. 

(Sorry, it ended up not being that much more concrete)

I found this bug in react-query 1.x but I have noticed the same code in 2.x

**Expected behavior**

A mutation promise always resolves to something, either a value or an error.

**Desktop (please complete the following information):**
 - OS: macos
 - Browser all

**Additional context**

I have isolated the bug to an edge case interaction between `queryCache.setQueryData` and `queryCache.refetchQuery`

(extracted from the repro repo)

```typescript

export function useSomeMutation() {
  return useMutation<void, undefined>(
    async () => {
      await delay();
    },
    {
      onSuccess: async () => {
        console.log("onSuccess start");
        queryCache.setQueryData("notInstantiated", "hello"); //BANG
        queryCache.setQueryData("notInstantiated2", "hello"); // BANG
        const a = queryCache.refetchQueries("notInstantiated", { force: true });
        const b = queryCache.refetchQueries("notInstantiated2", {
          force: true,
        });
        console.log("promises are never resolved", a, b);
        await Promise.all([a, b]);
        // This never reaches
        console.log("onSuccess end");
      },
    }
  );
}
```

Whenever you try to `setQueryData` on a query that hasn't been instantiated react-query gives the `queryFn` a value of `new Promise(noop)` which is a promise that never resolves.
So when you later do `refetchQuery` of that same query you get a promise that never returns and since `useMutation` waits for the `onSuccess` async function to resolve then it never resolves.

We are virtually saying this 

```typescript
onSuccess: () => new Promise(noop)
``` 

- check the 1.x code for [setQueryData](https://github.com/tannerlinsley/react-query/blob/1.x/src/queryCache.js#L234-L247)
[...]

PR #645, opened 2020-06-27
fix(setQueryData) unresolved promises

franleplant:
setQueryData(notInstantiatedQueryKey) now assigns Promise.resolve() instead of new Promise(noop) that caused a never ending promise en certain scenarios. Ref bug #639

EDIT
Forgot to mention that I went with `Promise.resolve` because in some other place we do `const promise = fn(...)` and if we simply make `fn` a `noop` then `promise` will be undefined and when we later do `promise.cancel` it will break. (typescript would help us in this case)

Here is where we access `promise.cancel` https://github.com/tannerlinsley/react-query/blob/1.x/src/queryCache.js#L375

### DIFF
--- a/src/queryCache.js
+++ b/src/queryCache.js
@@ -236,7 +236,7 @@ export function makeQueryCache() {
 
     if (!queries.length && typeof queryKey !== 'function') {
       queries = [
-        cache._buildQuery(queryKey, undefined, () => new Promise(noop), {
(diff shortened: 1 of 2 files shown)


franleplant:
thanks @tannerlinsley , sorry to bug you, is this going to be published automatically by github actions or do we need to do it manually?

tannerlinsley:
It should have been released. I’ll check it out

PR #656, opened 2020-06-29
TS make arg of mutation->onError rollback optional

mwmcode:
Hi Tanner,

I was gonna open an issue, but then i thought maybe I send PR since it's a small change.

Following the optimistic UI update section [Updating a list of todos when adding a new todo](https://github.com/tannerlinsley/react-query#updating-a-list-of-todos-when-adding-a-new-todo) in TS app, it seems that the type definition for the `rollback` function expects a mandatory parameter (variable) to be passed when it, _I think_, it should be optional (since the example in the docs doesn't pass anything to `rollback()`).

**To Reproduce**

- Check the linting error on line 26 in this [codesandbox](https://codesandbox.io/s/hardcore-greider-bydvq?file=/src/useApi.ts)

**Expected behavior**
No linting error about missing required 1 argument.

**Screenshots**

Screen Shot 2020-06-29 at 3 33 58 pm

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -405,7 +405,7 @@ export interface MutateOptions<TResult, TVariables, TError = Error> {
   onError?: (
     error: TError,
     snapshotValue: unknown,
-    onMutateValue: (variable: TVariables) => Promise<unknown> | unknown
+    onMutateValue: (variable?: TVariables) => Promise<unknown> | unknown
   ) => Promise<void> | void
   onSettled?: (
     data: undefined | TResult,


tannerlinsley:
You’ll have to run this by the TS people here. I’m not sure myself on this one TBH

jelteliekens:
@mustafawm Do you still have the same issue with v2.4.5 which fixes some types?

mustafawm:
@tannerlinsley the TS people have spoken :D @jelteliekens you're right, I just pulled 2.4.5 and the issue is fixed 👍

Gracias todos 🙏

PR #657, opened 2020-06-29
Fix types for MutationOptions

smashercosmo:


smashercosmo:
Ok, haven't seen #655

PR #663, opened 2020-06-29
New MutationResultPair type

justincy:
This change just assigns a generic type to the array returned by `useMutation`. This is helpful when creating hooks which wrap `useMutation` because it helps avoid such verbose types (repeating `TResult` and `TError` twice).

**Before**:

```ts
export default function useUpdateAccount(): [
  MutateFunction<AccountResult, Account, Error>,
  MutationResult<AccountResult, Error>
] {
    return useMutation<AccountResult, Account, Error>(...);
}
```

**After**:

```ts
export default function useUpdateAccount(): MutationResultPair<AccountResult, Account, Error> {
    return useMutation<AccountResult, Account, Error>(...);
}
```

I'm not convinced that I chose the best name, especially since `MutationResult` already exists, but I really couldn't think of anything else.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -393,7 +393,9 @@ export function useMutation<
 >(
   mutationFn: MutationFunction<TResult, TVariables, TError, TSnapshot>,
   mutationOptions?: MutationOptions<TResult, TVariables, TError, TSnapshot>
-): [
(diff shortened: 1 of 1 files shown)


MichaelDeBoey:
@all-contributors Please add @justincy for code

PR #666, opened 2020-06-29
Use funding key in package.json instead of postinstall script

MichaelDeBoey:


tannerlinsley:
Why not both?! :)

MichaelDeBoey:
@tannerlinsley npm is discouraging it

MichaelDeBoey:
@opencollective too: https://github.com/opencollective/opencollective/issues/2391

tannerlinsley:
Hmmm. Fair enough :)

PR #673, opened 2020-06-30
fix: add force option for typescript definition of refetch

mek-yt:


tannerlinsley:
`query.refetch` is already "forced". So this is not needed.

rshov:
You might just need to remove it from the docs. A "force" option is listed for the `refetch` method on this page:
[https://react-query.tanstack.com/docs/api#usequery](https://react-query.tanstack.com/docs/api#usequery)

> To bypass the stale check, you can pass the force: true option and refetch it regardless of it's freshness

PR #674, opened 2020-06-30
fix(types): fix MutateFunction

### LINKED ISSUE #661: useMutation expects incorrect mutateOptions as argument on mutate
**Describe the bug**
Typescript complains about the argument I provide to my mutate function. This is since upgrade to the 2.4.4 version today

**To Reproduce**
```
 const [remove, { status }] = useMutation(removeSomething, {
    onSuccess: () => {},
    onError: () => {},
  });

```
When I call the mutate method like so:
`remove('someId'); `

I get **'Type string has no properties in common with type 'MutateOptions''**

The removeSomething function looks like:

```
export async function removeSomething(idToRemove: string): Promise<void> {
  // fetch
}
```

jelteliekens:
TVariables now extends undefined to determine the return type of MutateFunction.

Fixes #661. When disableling [`--strictNullChecks`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#--strictnullchecks) `undefined` is always in every type's domain. `undefined extends TVariables` would always be true.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -443,7 +443,7 @@ export type MutateFunction<
   TResult,
   TVariables,
   TError = Error
-> = undefined extends TVariables
+> = TVariables extends undefined
   ? (options?: MutateOptions<TResult, TVariables, TError>) => Promise<TResult>
   : (
       variables: TVariables,


MichaelDeBoey:
@all-contributors Please add @jelteliekens for code

basslagter:
As I said in my issue (https://github.com/tannerlinsley/react-query/issues/661), this fix doesn't seem to be helpfull

jelteliekens:
@basslagter It is merged with master but not yet release. v2.4.6 doesn't include this fix yet. Hopefully it does solve your issue once it get released.

basslagter:
Works! Thanks

justincy:
I think that this change is making `variables` required when calling the mutate function, whereas previously you were able to call the mutate function without any arguments. Is that intended? The docs say arguments are optional.

jelteliekens:
Can you give a code example? If you expect variabels in the mutate function to do your API call, isn’t it required to pass variables when calling the mutation function?

justincy:
Yes, but not all API calls require variables and thus mutate doesn't require it either: https://github.com/tannerlinsley/react-query#returns-3

```js
const [mutate] = useMutation(() => {
    return axiox.post('/some/url');
});
```

justincy:
I don't know what the right fix is. I'm not claiming this needs to be reverted. But before this change I was able to call mutate with no args, then after the change the compiler is telling me that at least one arg is required.

PR #679, opened 2020-07-01
fix(types): useQuery enabled can be any type

dpyzo0o:
`enabled` can also be used for dependent queries, thus it's not restricted to `boolean`

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,7 +285,6 @@ $ npm i --save react-query
 $ yarn add react-query
 ```
 
-
 # Defaults to keep in mind
 
 Out of the box, React Query is configured with **aggressive but sane** defaults. **Sometimes these defaults can catch new users off guard or make learning/debugging difficult if they are unknown by the user.** Keep them in mind as you continue to learn and use React Query:
@@ -1041,7 +1040,7 @@ import { ReactQueryConfigProvider } from 'react-query'
 const queryConfig = {
   shared: {
     suspense: true,
-  }
+  },
 }
 
 function App() {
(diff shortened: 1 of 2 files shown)


MichaelDeBoey:
@dpyzo0o Can you give a little bit more explanation please?

dpyzo0o:
@MichaelDeBoey As described [here](https://github.com/tannerlinsley/react-query#dependent-queries) in the docs, `enabled` is used as a dep to run the query, it could literally be any type. Also see the source code here 
https://github.com/tannerlinsley/react-query/blob/ef1d0d512f3de769b71487650c98a3d56881ce3c/src/react/useBaseQuery.js#L40
whenever `enabled` changes, the instance will rerun the query.

tannerlinsley:
I'm not sure what it should be in TS, but technically it *can* be anything that can be used to pass a boolean condition

types/index.d.ts
```diff
@@ -222,7 +222,7 @@ export interface BaseQueryOptions<TResult = unknown, TError = Error> {
    * Set this to `false` to disable automatic refetching when the query mounts or changes query keys.
    * To refetch the query, use the `refetch` method returned from the `useQuery` instance.
    */
-  enabled?: boolean
+  enabled?: any
```
MichaelDeBoey, line 225: I would make it `boolean | unknown` instead

MichaelDeBoey (changes requested):

types/index.d.ts
dpyzo0o: done. I also add some explanation in the readme. Feel free to change the wording if necessary.

dpyzo0o:
I have no idea why the checks failed

README.md
```diff
@@ -1772,8 +1771,9 @@ const queryInfo = useQuery({
   - Receives the following variables in the order that they are provided:
     - Query Key Variables
   - Must return a promise that will either resolves data or throws an error.
-- `enabled: Boolean`
+- `enabled: Boolean | any`
```
MichaelDeBoey, line 1774:
```suggestion
- `enabled: Boolean | unknown`
```

MichaelDeBoey (changes requested):

MichaelDeBoey:
@dpyzo0o CI seems to fail.
Please take a look at it 🙂

dpyzo0o:
> @dpyzo0o CI seems to fail.
> Please take a look at it 🙂

I have no idea why it failed. The type change should not break the tests. I have run `yarn test:ci` locally and it passed.

MichaelDeBoey:
@tannerlinsley I'm not confident enough to merge this with failing CI.
I'll leave this one up to you

PR #688, opened 2020-07-01
fix(types): queryCache.getQueries

jgettings:
There were some missing types in the getQueries method's type definitions. I copied the parameter types from the other methods that allow `AnyQueryKey | string | function`, and added `boolean` as it looks like that's also a possible parameter type, looking at the implementation.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -614,7 +614,14 @@ export interface QueryCache {
     { exact }?: { exact?: boolean }
   ): void
   getQuery(queryKey: AnyQueryKey): CachedQuery<unknown> | undefined
-  getQueries(queryKey: AnyQueryKey): Array<CachedQuery<unknown>>
+  getQueries(
+    queryKeyOrPredicateFn:
+      | AnyQueryKey
+      | string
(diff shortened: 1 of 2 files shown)


smashercosmo:
```
queryKeyOrPredicateFn:
      | AnyQueryKey
      | string
      | boolean
      | ((query: CachedQuery<unknown>) => boolean)
```
this type should be added to almost all queryCache methods: getQueryData, setQueryData, getQuery, getQueries
make sense to extract it into separate type:
```js
type QueryKeyOrPredicateFn:
      | AnyQueryKey
      | string
      | boolean
      | ((query: CachedQuery<unknown>) => boolean)
```

jgettings:
@smashercosmo Sounds good! I wasn't sure how far I should go with it. I just pushed an update.

jgettings:
Sorry about all of the weird extraneous commits, I was just trying to get the checks into a good state, there seems to be a preexisting race condition in a test or two 😢

MichaelDeBoey:
@all-contributors Please add @jgettings for code

PR #694, opened 2020-07-02
Update global suspense configuration in example

deodad:
This updates the Suspense example in the readme to use the version 2 global configuration schema where suspense is no longer a top level config option.


tannerlinsley:
Nice catch!

adaddeo:
Sure thing! Just so happened to upgrade and reference that section in quick succession :)

shaohaolin:
@tannerlinsley  what is the typing for queryConfig?

PR #701, opened 2020-07-03
docs: update return for useQuery

MarceloAlves:
Adds documentation for `clear()` on `useQuery`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1860,6 +1860,8 @@ const queryInfo = useQuery({
   - A function to manually refetch the query if it is stale.
   - To bypass the stale check, you can pass the `force: true` option and refetch it regardless of it's freshness
   - If the query errors, the error will only be logged. If you want an error to be thrown, pass the `throwOnError: true` option
+- `clear: Function() => void`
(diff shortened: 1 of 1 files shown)


MarceloAlves:
There are a handful of other methods returned but not documented/typed that come from the `queryCache` (`heal()`, `cancel()`, `invalidate()`, etc). Should those be added to the docs/types as well?

MichaelDeBoey:
@MarceloAlves If you want, you can add them too 🙂
Same for typings (in a separate PR) 😉

tannerlinsley:
Query methods themselves are not that public... but I guess people are finding them anyway.

MarceloAlves:
> Query methods themselves are not that public

@tannerlinsley yeah thats what I was wondering. Sure they're available but are they really meant to be documented and used

tannerlinsley:
At this moment, no. If we do decide to expose them properly, then I’ll want to look through them and their signatures to make sure they’re good to go.

MarceloAlves:
Sounds good!

PR #717, opened 2020-07-08
fix: (queryInstance.clearInterval) is not called on unmount

### LINKED ISSUE #711: queryInstance.clearInterval() is not called on Unmount
When a component unmounts `queryInstance.unsubscribe()` tries to clear intervals here:
https://github.com/tannerlinsley/react-query/blob/master/src/core/queryInstance.js#L73
Note that `query.instances.length===0`

BUG:
`query.clearIntervals()` does no clear any intervals because the array it iterates over is empty.
https://github.com/tannerlinsley/react-query/blob/master/src/core/query.js#L138

otaciliolacerda:
Fix #711


johnsetzer:
Impressive turn around time. Nice work.

PR #720, opened 2020-07-08
Docs: Hide <kbd> Shortcuts for mobile/tablet breakpoints in Search

MrLeebo:
These devices don't can't take advantage of shortcuts


MrLeebo:
There's already a PR about it

PR #721, opened 2020-07-08
fix(examples): Import ReactQueryDevtools from correct package

keichinger:
While browsing through the docs I've noticed that the `pagination` example is broken (can't be build): https://codesandbox.io/s/github/tannerlinsley/react-query/tree/master/examples/pagination

The `load-more-infinite-scroll` was affected by the very same issue, which have been fixed with this PR too: https://codesandbox.io/s/github/tannerlinsley/react-query/tree/master/examples/load-more-infinite-scroll

As far as I could tell (coming off a very, very long day...) these were the only examples that were broken.

### DIFF
--- a/examples/load-more-infinite-scroll/pages/index.js
+++ b/examples/load-more-infinite-scroll/pages/index.js
@@ -3,7 +3,7 @@ import Link from 'next/link'
 import axios from 'axios'
 
 import { useInfiniteQuery } from 'react-query'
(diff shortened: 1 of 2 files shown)


keichinger:
I'm a bit confused if the failing CI is related to my changes. Did I miss an important step that caused the CI to fail? I've never worked with jest, nor jest-haste-map before. Can someone point me in the right direction? :)

PR #725, opened 2020-07-09
docs(guides/mutations): add 'throwOnError' option

corbecker:
I was working off the examples for useMutation trying to get a form submission to throw with errors from the back-end.
The rejected Promise from mutate wasn't getting caught in the try/catch. I now see in the API an option for throwOnError which is defaulted to false, so I added it to the useMutation example with the try/catch block.


andrewmclagan:
dam, this really got me...

PR #727, opened 2020-07-09
Fix incorrect statement about Apollo

fizwidget:
Apollo does a deep comparison of the variables object when deciding whether or not the query needs to be re-executed. 

I can’t find this explicitly stated in the documentation unfortunately, but I know it to be true from experience. A quick squiz at the source code seems to indicate this is where it’s implemented:

https://github.com/trojanowski/react-apollo-hooks/blob/master/src/useQuery.ts#L119

(‘objToKey’ serialises the variables object.)


tannerlinsley:
Awesome update. So you think it’s worth separating into two rows? One for query function and one for variables?

PR #728, opened 2020-07-09
Add queryCache de/rehydration

Ephem:
This is a continuation of things described in issue #461. It's a breakout from the closed PR #570 and builds on the merged PRs #584 and #917.

This PR adds a couple of new public APIs:

* `dehydrate(queryCache, dehydrateConfig)`
* `hydrate(queryCache, dehydratedQueries)`
* A new `hydration/ReactQueryCacheProvider`-component with the additional `dehydratedState`-prop
* `useHydrate(dehydratedQueries, hydrateConfig)` - A hook that does `hydrate` for you in a React-compatible way

Together, this provides a way to dehydrate a cache, pass its serialized form over the wire or persist it somehow, and later hydrate those queries back into an active cache. Main goals are to improve support for server rendering and help with things like persisting to localstorage or other storage.

An important feature of these new APIs is that the shape of the dehydrated cache is meant to be a private implementation detail that consumers should not rely on, which needs to be emphasized in the docs. This means that the de/rehydrate functionality can more easily be built upon in the future without breaking changes.

This PR is missing API-docs for now (which is why it's WIP).

### SSR Example with Next.js

> ❗ This example previously looked different, but has been updated for `useHydrate`

A minimal example (not tested/verified to work, but should):

```jsx
// _app.jsx
import { ReactQueryCacheProvider } from 'react-query/hydration'

export default function MyApp({ Component, pageProps }) {
  return (
    <ReactQueryCacheProvider initialQueries={pageProps.initialQueries}>
      <Component {...pageProps} />
    </ReactQueryCacheProvider>
  )
}

// pages/posts.jsx
import { makeQueryCache } from 'react-query'
import { dehydrate } from 'react-query/hydration'

export async function getStaticProps() {
  const queryCache = makeQueryCache()

  await queryCache.prefetchQuery('posts', getPosts)

  return {
    props: {
      initialQueries: dehydrate(queryCache)
    }
  }
}

function Posts() {
  // This useQuery could just as well happen in some deeper child to
  // the "Posts"-page, data will be available immediately either way
  const { data } = useQuery('posts', getPosts)

  // This query was not prefetched on the server and will not start
  // fetching until on the client, both patterns are fine to mix
  const { data: otherData } = useQuery('posts-2', getPosts)

  // ...
}
```

### Config options

There are a few config options that you can set when dehydrating and hydrating. Most notably is the `shouldDehydrate` and `shouldHydrate`-functions, that can filter the queryCache for queries to de/rehydrate. Usecase when dehydrating is to filter out only the queries you want to persist to localstorage for example. Usecase for hydration is when you don't want to hydrate certain queries that are too old for example.

_With or without shouldDehydrate, only successful queries are dehydrated from the cache_

**dehydrateConfig**

```ts
type ShouldDehydrateFunction = <TResult, TError = unknown>(
  query: Query<TResult, TError>
) => boolean

interface DehydrateConfig {
  shouldDehydrate?: ShouldDehydrateFunction
}
```

### Questions

These are a few questions I'm not totally certain about myself and might be worth an extra look when reviewing.

#### ~~Should hydration be its own entry point?~~

> Solved, is now its own entry point

#### ~~Should `dehydrateQuery` be included in the official APIs?~~~

> Solved, was removed in favour of a `shouldDehydrate`-function to keep the API lean. Could be exposed publically later on if there is need for it.

#### ~~How should we handle `updatedAt`?~~

> Solved, updatedAt is now hydrated and also used for determining how far into the future staleness should be scheduled.

#### Can naming be improved?

There are a bunch of new public APIs, so let's get the naming juust right!

---

I'm sure I've missed a bunch of stuff I wanted to highlight and/or ask, but hopefully this is a good start for getting feedback and questions. 😅


tannerlinsley:
I'm loving this. I think something we should consider here is making this its own bundle via rollup, then proxying it with a nested import:

- Rollup creates similar packages for both the core and the hydration utilities. The existing stuff doesn't need to change, but we can add a new configuration to rollup for hydration and have it place the files in `dist/hydration/...`
- Create a file in the root of the repo called `hydration.js`, which is essentially the same as `/index.js`, but points the hydration files instead.

This should keep the bundle size down for the default use case.

tannerlinsley:
As for updatedAt, technically it *is* UTC because it uses `Date.now()`, right? Wouldn't this mean that it would be okay to included in the dehydration? Then really what should happen on hydration is that stale and garbage collection times should be based off of that time (or immediately, if it's too far in the past).

Ephem:
Sorry for the radio silence, busy summer. 😄

* I'll go ahead and separate this out into it's own entry point.
* I'm not sure why I brainfarted on the `updatedAt`, you are right of course. Requiring server time to be set correctly for staleness to work correctly isn't such a big ask so I think we should definitely base the stale time off of that when hydrating! I'll fix that too.

Another thing I've realized is that the Next-example in the description wont work since setting up the cache per page would blow away the entire cache on page transitions. This _might_ mean we want to separate out the hydration-specific functionality from `ReactQueryCacheProvider`, possibly to a hook instead. This would also move that to the separate entry point as a bonus.

I'll start working on this now, but might take a little more time figuring out the details (and tests).

What do you think about releasing this as `unstable_` or something similar at the start before we have docs, examples etc in place, as well as giving it some time to mature?

src/hydration/tests/useHydrate.test.js
```diff
@@ -0,0 +1,179 @@
...
+
+  beforeAll(async () => {
+    const serverQueryCache = makeQueryCache()
+    await serverQueryCache.prefetchQuery('string', dataQuery)
+    const dehydrated = dehydrate(serverQueryCache)
+    stringifiedQueries = JSON.stringify(dehydrated)
+    serverQueryCache.clear({ notify: false })
+  })
+
+  test('should handle global cache case', async () => {
```
Ephem, line 29: I'm not sure how to handle this case properly. Right now it works on the client, but kind of fails silently on the server (since queries wont actually get added to the cache there). Using the global cache with hydration could be nice in some cases with custom SSR with a separate client entry-point, but is a footgun in Next.

Since v3 will remove the global cache, maybe this behavior should be changed to throw an error if no custom cache has been provided via a `ReactQueryCacheProvider` instead?

Ephem:
Okay, this PR has been updated with a bunch of stuff.

* Hydration is now its own entry point `react-query/hydration`
* `initialQueries` was removed again from `ReactQueryCacheProvider`, instead a new `useHydrate`-hook was introduced
* `updatedAt` is now included when de/rehydrating
* When scheduling staleness on hydration, it is now measured from the point in time the data was fetched on the server (`updatedAt`), instead of the time of hydration.

I also updated the description above, including the `Next.js`-example. Except for missing types, docs and examples I hope this is in a pretty good place now and ready for review. I still need to verify a few things so I'll leave the WIP on a while longer.

Two questions:

* Should we make this an unstable feature at first to make sure it works as we've intended and for all use-cases before we commit to the design fully?
* Which of types, docs and examples would you like to see in this PR as opposed to in follow-ups?

src/hydration/useHydrate.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
```
Ephem, line 1: This hook is included in the `hydration`-bundle which makes the bundle React-specific. I think this makes sense for now to keep down default payload, but that this should probably move to the React-bundle if `core` and `react` are split to separate bundles in the future to keep `hydration` framework agnostic.

kamranayub:
I am not quite sure where to put this (maybe in v3 issue) but the approach we took in our platform was wrapping react-query with our own thin layer that takes care of SSR preloading as well as named query support.

For SSR, I took a pretty naive approach where we use a `Map` store on the request context (through Next.js) and that store gets serialized to the page on render. I then override `initialData` in react-query to hydrate from that global storage when available. This lets us achieve most of what we want by using `getInitialProps` to execute and preload query data and get it hydrated on the client.

## Pseudo usage

**Homepage.js**

```js
import { pageHeading } from 'data/page-heading'

// pageHeadingData would be a query descriptor
// containing its name, queryFn, default options, etc.

export const Homepage = () => {
  const { data: heading } = pageHeading.useQuery(/* vars, options */);
  
  return (
    <h1>{heading}</h1>
  )
}

Homepage.getInitialProps = (ctx) => {
  await ctx.preloadInitialQuery(pageHeading/*, vars, options */)
  return {};
}
```

The `preloadInitialQuery` function inserts the query result into the context request cache:

```js
async function preloadInitialQuery(ctx /* next.js context */, queryDescriptor, vars, options) {
  const { data, error } = await executeQuery(queryDescriptor, vars, options);
  if (!error) {

    // set the result for this query key
    // in the per-request context injection store
    ctx.queryStore.set(getQueryKey(queryDescriptor.id, vars), data);
  }
}
```

On render, the our app server would take that `Map` and inject it into the page like:

```html
<script>
window.__QUERIES__ = {
  "<QUERY_KEY>": "<QUERY_RESULT>"
};
```

And our `initialData` would check for that:

```js
initialData() {
  const queryKey = getQueryKey(name, vars);

  if (window.__QUERIES__[queryKey]) {
    return window.__QUERIES__[queryKey]
  }
  return undefined;
}
```

By using a `Map` as the generic store on the request, we were able to use this both within Next.js and outside Next.js (we have two different app servers that use our shared library).

Just wanted to share this approach for future reference/ideas! It works pretty well mainly due to our ability to use named/registered queries so we can always look them up and not have to pass `queryFns` around.

Ephem:
@kamranayub That's a neat solution when using the current React Query, thanks for sharing! What you are doing is essentially implementing a separate cache to use when server rendering, this PR is all about implementing support for this directly in the library so that you don't need to have that separate cache anymore. Besides hopefully reducing complexity for you, this also means:

* We can transfer some config between server and client (staleTime, cacheTime, updatedAt for now)
* We can set the query to become stale a certain amount of time _after it was fetched on the server_
* We can support Suspense with SSR better in the future (or with third party solutions like [`react-ssr-prepass`](https://github.com/FormidableLabs/react-ssr-prepass))
* We better support things like persisting the cache to things like localstorage

You probably got this from the description already, just wanted to be clear on the motivations for implementing this as a first class thing. 😄

I'm curious, does it look like this PR would make things easier for you or is it missing something?

Btw, I think the named query registry is still a great idea for other reasons, you probably wanna keep that around either way. 💯

kamranayub:
Yes, we were looking through this and I think it would help make it easier to integrate with! Since we have two app servers, a Next.js-based one and a custom one, the approach has to end up being able to support a "generic" integration.

Ephem:
Glad to hear that! I built this for use in a custom SSR solution myself, so it should absolutely support that.

* Server
  * Create a cache manually for every request using `makeQueryCache`
  * Use `yourQueryCache.prefetchQuery` to prime the cache
  * Place cache on context with `ReactQueryCacheProvider`
  * Render app
  * Call `dehydrate(yourQueryCache)`
  * Serialize and embed result in payload like you are doing with `window.__QUERIES__`
  * Send payload
* Client
  * Create a cache manually
  * Get dehydrated queries from `window.__QUERIES__` (or whatever you called it)
  * Either:
    * Call `hydrate(queryCache, dehydratedQueries)` and the resulting `initializeQueries()` directly (low level)
    * Render (with `ReactQueryCacheProvider`)
  * Or:
    * Render (with `ReactQueryCacheProvider`)
    * Do `useHydrate(dehydratedQueries)` in child

Typing out a full example is a bit more verbose with custom SSR of course which is why I left it out of this PR. It will definitely be covered by docs and examples though!

pseudo-su:
`useHydrate ` seems like a strange API choice instead of something like 

```tsx
import React from "react";
import {render} from "react-dom";
import {hydrateCache} from "react-query/hydrate";

const hydratedQueryCache = hydrateCache(window.__QUERIES__);

render(
  <ReactQueryCacheProvider cache={hydratedQueryCache} />
    <App />
  </ReactQueryCacheProvider>
)
```

Is there a reason? Is it to keep things related to hydration out of the main entry point? triggering hydration of the cache using the global data from a child component and effectively passing it *up* the component tree seems wrong. IMO it would make more sense to provide it at the highest point in the component tree so parent state isn't updated by a child rendering

Ephem:
@pseudo-su Great question, this is something I definitely want feedback on so I'm glad you asked. 

In Next.js, since we receive the dehydrated queries as props we need to hydrate the queries _inside of render_. This isn't necessarily great, but I know no other way to solve it currently. Hydrate consists of two parts, first we need to put the queries into the cache and then we need to schedule timeouts for staleness and garbage collection. The first one might not be great to do in render since it could be considered a side effect to add queries to an external cache, but at least it's idempotent, meaning that we can run it several times and get the same outcome, queries already in the cache wont be written over. Scheduling timeouts however is something that should probably happen in an effect instead. (Scheduling timeouts in render is something the other hooks like `useQuery` do today though, this PR also paves the way to change that if desirable)

When using custom SSR none of this is a problem and you can just do both before even rendering and in fact, the `hydrate`-function you are suggesting already exists in this PR! The only difference is that in order to separate putting queries into the cache and the scheduling of timeouts, the `hydrate` in this PR returns a function `initializeQueries` which you need to call to schedule the timeouts. This feels a bit wonky, and maybe the public API should just be a `hydrate`-function that does both that you can use in custom SSR, and a `useHydrate` that you can use inside of render, but if we want to support other frameworks in the future (or just usecases we haven't thought about), it might make sense to provide a low level API as well that lets you separate these. I do worry that it's easy to forget to call the function returned from `hydrate` though since the API is a bit unusual..

An earlier version of this PR had an extra `initialQueries`-argument to `ReactQueryCacheProvider`, but you might want to put hydration and the cache provider in different components. Separating it also means we can move hydration to a separate entry point to avoid increasing the JS-payload for people who don't use hydration. An alternative/complementary API to `useHydrate` could be to have a `ReactQueryHydrator`-component or something similar:

```jsx
function ReactQueryHydrator({ queries, children }) {
  useHydrate(queries)
  return children
}

function App() {
  return (
    <ReactQueryCacheProvider>
      <ReactQueryHydrator queries={dehydratedQueries}>
        <Content />
      </ReactQueryHydrator>
    </ReactQueryCacheProvider>
  )
}
```

Hydration is purely behaviour though and the rest of React Query is hooks-based, so a hook seemed like a good fit. I'm very open to including such a component as well for conveniency though!

Another difference from your example is that in that, the `hydrateCache`-function seems to be the one actually creating the cache. This was considered in an earlier version of this PR (#570), but as an argument to `makeQueryCache` instead, similar to how libraries like Redux does it. I realized that we don't need to constrain ourselves to only hydrating at creation though, it might very well make sense to hydrate several different dehydrated queries and/or at different points in time, for example from both server rendering and localstorage.

To address what seems to be a main concern though, `useHydrate` isn't actually updating a top components state, only an external cache so the question is, will that update trigger a re-render somewhere? First of all, the `queryCache` itself is still the same, so `ReactQueryCacheProvider` wont re-render. Second, if a component is already listening to one of the queries that exists in the `dehydratedQueries`, this means that query has already been created in the cache (without data) and therefor wont be overwritten by the hydration, so even in that case no re-render should happen. This means `useHydrate` should only ever affect components further down in the tree which makes sense as you pointed out, you need to hydrate it before you use it.

This became somewhat of a wall of text, but I wanted to try to be as clear as I could about some of the tradeoffs and motivations about the current API design-choices. 😄 Even so, I think there is still room for improvement, so I'm very grateful for all feedback!

pseudo-su:
> This became somewhat of a wall of text

As someone prone to writing accidental walls of text as well I appreciate the detail 😄. I'll try go through things point by point and then propose/elaborate some thoughts all at once at the end.

> In Next.js, since we receive the dehydrated queries as props we need to hydrate the queries inside of render

That does explain why `useHydrate` is required. My knee-jerk response to that would be why impose hydrating the cache inside of rendering if/when people are doing basic/vanilla SSR without that limitation?

> maybe the public API should just be a hydrate-function that does both that you can use in custom SSR, and a useHydrate that you can use inside of render

Wouldn't it be possible to expose the multiple hydration functions and just provide people with the `useHydrate` and a thing that coordinates them? from what I understand (which might not be accurate) something like:

```
const queryCache = makeQueryCache();

hydrate(queryCache).initQueries();
```


> but if we want to support other frameworks in the future [...] it might make sense to provide a low level API as well that lets you separate these

I'm not exactly clear on which one of these are being considered as the "low level" vs "high level" API. personally I would consider the "low level" API to be the one that has no knowledge about react and is purely dealing with the `queryCache`. Which I think I would prefer to use.

> An alternative/complementary API to `useHydrate` could be to have a `ReactQueryHydrator` component or something similar

If it was necessary to do hydration *during* the react render I'd personally find it preferable to have the `<ReactQueryHydrator />` API (because otherwise everyone doing vanilla SSR has to create it themselves anyway), but I'm not clear on why anyone doing basic SSR would prefer that over just setting up the `queryCache` outside of react.

> Separating it also means we can move hydration to a separate entry point to avoid increasing the JS-payload for people who don't use hydration

Makes sense wanting to keep the entrypoints seperate

> Another difference from your example is that in that, the hydrateCache-function seems to be the one actually creating the cache

yeah that was just for terseness (and I wasn't aware of the need to hydrate and also have the seperate `initializeQueries`), I don't really think it's necessary to have an API that create+hydrates+initializes all in one, but I do think exposing a set of functions to do this would be preferable for anyone doing basic/vanilla SSR.

> To address what seems to be a main concern though, useHydrate isn't actually updating a top components state

I did have a quick skim through the code and I saw that it's technically not updating a parent components state, but my first impression was that's what it's function was which made the API feel strange to me. It makes sense that if/when using certain SSR frameworks it's a limitation/requirement to do the hydration as part of the react render. For what it's worth it might just be the name, `<ReactQueryHydrator queries={dehydratedQueries}>` seemed clearer to me (all in all probably not a big deal tho).

pseudo-su:
Is something like this viable?
```tsx
import React from "react";
import {render} from "react-dom";
import {makeQueryCache} from "react-query";
import {hydrateCache} from "react-query/hydrate";
const queryCache = makeQueryCache();
hydrate(queryCache, window.__QUERIES__).initQueries();
render(
  <ReactQueryCacheProvider cache={queryCache} />
     <App />
  </ReactQueryCacheProvider>
)
```

I think there's a real potential benefit of providing an API to use that's totally independent of the `react` render (not a provider and not a hook) and I think I can provide an example:

Currently I heavily use `import()` expression syntax in my client entrypoint which allows webpack to do automatic bundle splitting, it looks a bit like this

```tsx
import "core-js/stable";
import "regenerator-runtime/runtime";

async function loadDependenciesAsync() {
  const loadReact = import(/* webpackPreload: true */ "react").then((m) => ({
    React: m.default,
  }));
  const loadApp = import(/* webpackPreload: true */ "./app");
  const loadLibComponents = import(
    /* webpackPreload: true */ "@mysite/lib-design-system"
  );
  const loadSsrToolbox = import(
    /* webpackPreload: true */ "@mysite/lib-ssr-toolbox/client"
  );

  // Not used directly but still critical-path can good to preload.
  // When making changes to this check the build/webpack.report.html to
  // see how it affects the resulting bundle.
  // prettier-ignore
  {
    void import(/* webpackPreload: true */ "react-dom");
    void import(/* webpackPreload: true */ "react-router");
    void import(/* webpackPreload: true */ "react-router-dom");
  }

  const allLoaded = await Promise.all([
    loadReact,
    loadApp,
    loadLibComponents,
    loadSsrToolbox,
  ]);

  const combined = Object.assign({}, ...allLoaded);

  return combined;
}

void loadDependenciesAsync().then((asyncDeps) => {
  const { React, hydrate, ThemeProvider, App, DefaultTheme } = asyncDeps;
  hydrate({
    render() {
      return (
        <ThemeProvider theme={DefaultTheme}>
          <App />
        </ThemeProvider>
      );
    },
  });
});

```

If the only API I have is one that runs during the react render then I have to wait for react (and potentially a whole bunch of other libraries) to load before being able to create and hydrate the queryCache. Alternatively with an API that allowed me to hydrate the cache outside of the react render I could do something like this and initialize the queryCache before/while the react JS bundle is downloading:

```tsx

async function loadDependenciesAsync() {
  /* etc */
}

async function initQueryCache() {
  const {makeQueryCache} = await import("react-query");
  const {hydrateCache} = await import("react-query/hydrate");
  cosnt queryCache = makeQueryCache();

  // maybe this should fail in dev mode but not in prod mode
  await hydrate(queryCache, window.__QUERIES__).initQueries()

  return queryCache;
}

void Promise.all(loadDependenciesAsync, initQueryCache).then(([asyncDeps, queryCache]) => {
  /* etc */
})
```

Ephem:
@pseudo-su  ❤️ for the detailed discussion!

> That does explain why useHydrate is required. My knee-jerk response to that would be why impose hydrating the cache inside of rendering if/when people are doing basic/vanilla SSR without that limitation?

Just to clear up what seems to be a misunderstanding, we aren't imposing that, the API you are asking for is already included in this PR! It's _slightly_ different (and I'm very open to changing the details), but this works:

```jsx
import React from "react";
import {render} from "react-dom";
import {makeQueryCache} from "react-query";
import {hydrate} from "react-query/hydrate";

const queryCache = makeQueryCache();

const initQueries = hydrate(queryCache, window.__QUERIES__);
initQueries();

render(
  <ReactQueryCacheProvider cache={queryCache} />
     <App />
  </ReactQueryCacheProvider>
)
```

An extra detail to note here which is also pretty confusing to communicate is that queries should never be stale or garbage collected before the first render. This would mean different things could render on the server and first render, leading to hydration mismatches.

The example above is safe only because `render` happens immediately after `initQueries`, or at least in the same tick. If `render` would happen in a later tick because of some asynchronous event between them, this could mean trouble.. So `await`-ing `initQueries` is actually the opposite of what we want.. So in practice, `initQueries` should probably be run _after_ the first render:

```jsx
const initQueries = hydrate(queryCache, window.__QUERIES__);

render(/* ... */);

initQueries();
```

This is pretty confusing and I don't like this level of detail leaking into the public API, so this is a bit of a sore point currently, I'd love any suggestions to make this less confusing (maybe better naming would help somewhat?). An alternative safer API, but possibly less flexible, could be:

```jsx
hydrate(queryCache, window.__QUERIES__).then(() => {
  render(/* ... */);
});
```

This way we could make sure to run the init behind the scenes _after_ calling the thenable? Note that the return wouldn't be a Promise, but rather a synchronous thenable. We could also call it something other than `.then` to make this clearer. I'm just brainstorming.

Without a doubt we'll support hydrating both inside and outside of render! _How_ we do that and how we best communicate it in docs and examples is very much up for debate though.

> Note: `render` in this entire comment should actually be `hydrate` when server rendering.. This made me realize calling the API in React Query just `hydrate` as well is.. not good. `hydrateCache` makes a lot more sense, I'll change that!

pseudo-su:
> Just to clear up what seems to be a misunderstanding, we aren't imposing that, the API you are asking for is already included in this PR!

Right 😅. I'm slowly getting onto the same page.

> Queries should never be stale or garbage collected before the first render

Ah right yep that makes sense.

I'm curious, what would be the result of never querying `initQueries`? cache timeouts are never set so cache items hydrated from SSR are never invalidated?

I can see why it would be a good idea to perform the second step of hydration during the render as a side-effect 🤔. It sounds like what I would want then is something that allows me to `hydrateCache` outside of react and then `initQueries` inside of react. I'm not quite sure if the following is a good idea/API but I'm going to throw it out there anyway:

```tsx
import React from "react";
import { hydrate } from "react-dom";
import { makeQueryCache } from "react-query";
import { hydrateCache } from "react-query/hydrate";

const queryCache = makeQueryCache();

// HydratedQueryInvalidator is probably not a great name but it's my
// attempt to think of a name that will communicate to consumers the
// reason they they have to render it into their component tree
// (presuming what I assume in my question above is accurate)
const { HydratedQueryInvalidator } = hydrateCache(queryCache, window.__QUERIES__);
// const { QueryInvalidation } = hydrateCache(queryCache, window.__QUERIES__);
// const { HydrationInvalidator } = hydrateCache(queryCache, window.__QUERIES__);
// const { HydrateFreshness } = hydrateCache(queryCache, window.__QUERIES__);
// const { HydrateStaleness } = hydrateCache(queryCache, window.__QUERIES__);

hydrate(
  <ReactQueryCacheProvider cache={queryCache} />
     <HydratedQueryInvalidator />
     <App />
  </ReactQueryCacheProvider>
)
```

Maybe `hydrateCache` could provide multiple options as ways to run the `initQueries()`; maybe an option to call a function or a component to render into your component tree. (As a side note did find the naming of `initQueries` slightly confusing, it isn't/wasn't clear to me *why*/*what* needs to be initialized).

```tsx
const { initQueries } = hydrateCache(queryCache, window.__QUERIES__);
// maybe a different name?
initQueries()

// OR

const { HydrationInvalidator } = hydrateCache(queryCache, window.__QUERIES__);;
<HydrationInvalidator />
```

tannerlinsley:
Looks like this is going to need to be updated for the new types :)

Ephem:
I've already talked to Tanner about this, but I wanted to update the PR as well for anyone following along. I'm back from vacations and plan to work on this for three full days during next week to hopefully finish it up! That includes refactoring to TS, docs and some extra functionality to better support dehydration/rehydration to/from localstorage.

This should have no breaking changes, so should be able to land it in a minor before v3.

Ephem:
This PR is starting to come together and I think only API-docs are missing which I'll work on today. I've updated the description to reflect the current state of things, here are the latest changes:

* Basing the stale timeout on `updatedAt` and refactoring how timeouts work was moved to #917 which has been merged
* Refactored the PR to TS (I really need help reviewing the types, I'm not a TS-expert, but they do pass, maybe @boschni wants to take a look?)
* Added a bunch of new SSR integration tests around hydration
* Removed `dehydrateQuery` from the public API
* `hydrate` now schedules timeouts by default to be less confusing, but supports a `activateTimeoutsManually`-option for low level use (for example in `useHydrate`) (ping @pseudo-su)
* `dehydrate` and `hydrate` now takes optional config-objects, most notably supporting `shouldDehydrate` and `shouldHydrate` filter-functions

There is an interesting caveat with the current API that I want to shine some extra light on, namely that only successful queries are dehydrated (ping @boschni). This has some consequences for SSR:

* Any errors that happen on the serverside will show up as loading and be retried automatically on the client
* To avoid React hydration mismatches when the above happens, you need to have a separate queryCache for prefetching and for rendering on the server and dehydrate between them before rendering.
  - This sounds funky, but the actual patterns aren't that weird when you look at them (check out the Next example in the description, that one works for example, or look at the SSR hydration integration tests). This is a potential footgun that needs to be highlighted in docs though and can be somewhat mitigated by recommending `useHydrate` over `hydrate`.
* When you want to avoid a retry on the client or set a correct status code on the server because some critical piece of data failed, you want to pass `throwOnError: true` to the `queryCache.prefetchQuery`-function and handle that manually

src/hydration/hydration.ts
```diff
@@ -0,0 +1,155 @@
...
+        continue
+      }
+      dehydratedQueries[queryHash] = dehydrateQuery(query)
+    }
+  }
+
+  return dehydratedQueries
+}
+
+export function hydrate<TResult>(
```
pseudo-su, line 93: Not sure if this is even particularly helpful/desireable (and I'm no TS expert either) but I think you *could* make `hydrate` only return the `activateTimeouts` function when `activateTimeoutsManually=true`. I think the types to support it would be something like this using constant overloads:

```tsx
function hydrate<TResult>(
  queryCache: QueryCache,
  dehydratedQueries: unknown;
  hydrateConfig?: HydrateConfig;
): void | () => void;

function hydrate<TResult>(
  queryCache: QueryCache,
  dehydratedQueries: unknown;
  hydrateConfig: Omit<HydrateConfig, "activateTimeoutsManually"> & { activateTimeoutsManually: true };
): () => void;

function hydrate<TResult>(
  queryCache: QueryCache,
  dehydratedQueries: unknown;
  hydrateConfig: Omit<HydrateConfig, "activateTimeoutsManually"> & { activateTimeoutsManually?: false };
): void;
```

https://stackoverflow.com/questions/25671381/can-i-overload-function-on-constant-of-number-type

Ephem: Nice! I thought it was probably possible but didn't have the TS-skills for it myself. 😄  I'm not sure what's desirable either, but I also thought of this kind of like "whatever the config, if you call the function returned by hydrate you can be sure that timeouts have been activated".

If you change the config option from manual to automatic and forget to remove the `activateTimeouts()` call you'll get an error with your version, which is good because you find a now pretty useless call to `activateTimeouts()`, but bad if this happens at runtime in prod because you are not using TS. 😅  Should be unlikely though, I don't have strong feelings on this either way really. 🤷

Ephem:
API-docs and a new SSR-guide is done, I think this is ready for review now!

This should be fully backwards compatible and not introduce any bugs in existing functionality. Hydration is nuanced, so there might very well be edge cases I've missed thinking about, but I'm pretty confident in this as a first version.

What I'm not super confident about is the TS-typings, and the build setup could use a careful review too since having a new entry-point is a first for the library. And of course just the design overall.. 😅

boschni:
Looking good! The need for separate caches when server side rendering does sound a bit confusing. Would be nice if there is a way to make both the server side and the client side straightforward to use. Think something like this would be the ideal way of working?

```js
// Server
const cache = makeQueryCache()
cache.prefetchQuery('key', fn)
const state = dehydrate(cache)
const html = ReactDOM.renderToString(
  <ReactQueryCacheProvider queryCache={cache}>
    <App />
  </ReactQueryCacheProvider>
)
res.send(`
<html>
  <body>
    <div id="app">${html}</div>
    <script>window.__STATE__ = ${JSON.stringify(state)};</script>
  </body>
</html>
`)

// Client
const cache = makeQueryCache()
const state = JSON.parse(window.__STATE__)
hydrate(cache, state)
ReactDOM.hydrate(
  <ReactQueryCacheProvider queryCache={cache}>
    <App />
  </ReactQueryCacheProvider>
)
```
The main issue here is that if timers start when hydrating the cache, then the render could end up to be different between the server and the client. But maybe there are ways around that? The cache timer should not really matter because this is most likely in the range of minutes and queries do not GC immediately anymore. The one that is most important is the stale time, which I think is only relevant when there are actual observers. If the stale timer would only start when observers subscribe (which happens in a `useEffect`) then I guess the first render should be the same?

Ephem:
@boschni Thanks for checking it out!

I do agree it's a bit cumbersome in some setups and your example was the original plan. Not dehydrating errors really messes with using the same cache for prefetching and rendering on the server though, since to implement that we would have to have a separate "mode" somehow that doesn't persist errors to the cache. Not only would this have to be implemented in the core package which breaks separation of concerns, it is confusing to users in its own right.

Say we add a `excludeErrors` to `prefetchQuery` for example that you always need to call if you intend to dehydrate, or if you need to configure the `queryCache` itself for that when you create it? What about when you want to persist to localstorage? In that case you want the errors in the cache, but not dehydrate them.

Dehydrating the errors too would be another approach, but this is hard since errors are usually not vanilla JS and therefor hard to serialize (or at least hydrate on the other end).

So I ended up with the thinking that the current version is the lesser of two evils. This also creates an interface where we can have other logic in the future without causing mismatches, say for example that we set the default `staleTime` to a higher amount for dehydrated queries in the future? Ensuring queries on the server and the client go through the same process of dehydration before rendering also creates future opportunities.

About `hydrate` vs `useHydrate`, using `hydrate` the way you are using it in the example works and is safe. If you add some async process between that call and render however, that can break unless you add `activateTimeoutsManually: true`, so that is one reason I'm pushing `useHydrate` as the preferred API. The other reason is that pushing that hydration into the React tree, that encourages a pattern where the above problem of having separate caches for prefetching and rendering on the server becomes less of a problem and more part of the pattern. :)

If we look at a snippet from the SSR integration-tests:

```jsx
    const serverPrefetchCache = makeQueryCache()
    await serverPrefetchCache.prefetchQuery('success', () =>
      fetchDataSuccess('success')
    )
    const dehydratedQueriesServer = dehydrate(serverPrefetchCache)
    const markup = ReactDOMServer.renderToString(
      <ReactQueryCacheProvider>
        <Page dehydratedQueries={dehydratedQueriesServer} />
      </ReactQueryCacheProvider>
    )
```

I think needing two separate caches for prefetching and rendering is troublesome on paper (and somewhat in practice), but when looking at the code like this, it's kind of hidden away by the fact that `<ReactQueryCacheProvider>` creates the render-cache for you behind the scenes. :)

So I totally agree this is a pain point, but the alternatives have some tradeoffs I considered to be worse.

About activating timeouts only in a `useBaseQuery` effect, this would work if all queries that you prefetch on the server were actually used on first render, but a valid pattern is also to prefetch some queries on the server where the latency is low and de/rehydrate it, but not rendering the corresponding React components until lazily on the client. Those would end up never getting scheduled for staleness or GC unless they were actually rendered at some point.

tannerlinsley:
Now that I can see it on paper, I really do prefer @boschni's first suggested syntax: https://github.com/tannerlinsley/react-query/pull/728#issuecomment-680104238

Ephem:
@tannerlinsley Well, I do too, but it doesn't work. It will lead to hydration mismatch on errors unless:

* We dehydrate errors too (which I don't know how to solve)
* We introduce something like this:

```jsx
// Server
const cache = makeQueryCache()
cache.prefetchQuery('key', fn, { ejectQueryFromCacheOnError: true }) // <-- This is new
const state = dehydrate(cache)
const html = ReactDOM.renderToString(
  <ReactQueryCacheProvider queryCache={cache}>
    <App />
  </ReactQueryCacheProvider>
)
res.send(`...`)
```

`ejectQueryFromCacheOnError` could be named better of course, but either way, this would be _required_ to avoid hydration mismatches. Another approach could be to re-introduce the concept of `makeServerQueryCache` with slightly separate behaviour from `makeQueryCache` (not saving errors). This was in my first PR, but you wanted this contained in a separate entry point.. 😅

To guarantee that the cache we are rendering with on the server and the client are the same and doesn't cause React hydration mismatches on error, we _somehow_ need to remove queries that error out on the server or solve de/rehydration of them (which also has an effect on UX that we would need to discuss). Doing this by using the same `dehydrate`-interface in both places still seems like the lesser evil to me.

I still agree this is a pain point and I'm very open to suggestions on how to solve this.

Ephem:
Btw, none of this has any bearing on Next.js btw, there you _have_ to use a separate cache for prefetching and rendering on the server anyway, so whatever extra API we introduce to solve this for _some_ custom SSR cases will probably be an unnecessary extra burden there.

boschni:
The serialization part is indeed interesting as both the query results and errors could contain JSON incompatible values. GraphQL does not have this issue because all values are JSON compatible and Redux stores could have this issue but it is generally recommended to only have JSON compatible values in the store. There are a few options how we could deal with this:

1. Require users to make sure their results and errors are JSON compatible when using SSR.
2. Provide a query option to automatically serialize results and errors before storing them in the query state.
3. Provide the possibility to serialize and deserialize non-JSON compatible values in results and errors when hydrating and dehydrating. This could be done by providing a serialize function to `hydrate` and a deserialize function to `dehydrate` (personally I like this one most).

As for sending queries to the client which are not rendered: I think these don't need to have a stale timeout as this only makes sense for queries which are being observed. The GC timeout could always be set (even on query creation) because this will most likely only trigger after a few minutes?

tannerlinsley:
Yes, I like the sound of that.

- Assume that all store values are JSON compatible. 99% of people are probably already doing this regardless of SSR. If they are using SSR, they don't really have any other choice (even with Next)
- Dehydrate and hydrate errors (probably even reinstantiate the err eg. `new Error(serializedError)` on hydration). This is one that might change over time, but as long as we do the following 👇 , we'll leave ourselves plenty of room to improve
- Provide dehydrate({serialize}) and hydrate({deserialize}) callbacks with sane defaults
- Simplify the API as discussed above

Ephem:
I have thought along those lines as well, but decided to leave custom serialization out in this initial PR for a couple of reasons, which I'd love to talk through.

First of all, we should consider what we want the default UX to look like. Should we present loading states to the users and refetch queries that had errors by default, or should we show the errors by default? My thinking has been that refetching those on the client is a sane default as a means of trying to recover, especially since a lot of SSR is cached for a period of time. I'm not at all certain if this is the best approach, but since this affects UX I think it's something we should consider first. (We always need a way to opt out of course)

> Assume that all store values are JSON compatible. 99% of people are probably already doing this regardless of SSR. If they are using SSR, they don't really have any other choice (even with Next)

I absolutely agree with this when it comes to data, but from the examples I've seen, people aren't really doing that or dehydrating/rehydrating errors today. Consider the approach in the docs right now, it only talks about `initialData`. Often with Next or Redux, people aren't really saving _the errors themselves_ in a store or de/rehydrating them, only the error state, but with React Query this isn't really true since it saves the actual error in the cache by default.

A thing to consider here is that a lot of errors can come from node/the network layer. I think a lot of default errors that users have not introduced themselves can easily propagate through into the cache? Some of these might not exist on the client and therefor not be truly deserializeable. Users would have to catch every single error and convert them into something serializable.

Even with sane defaults (which I'm not sure how or if we could implement well), I think requiring custom serialization/deserialization is a huge burden to add to even get up and running, compared to the API in this PR, slightly awkward as it might be for custom SSR setups.

I think what I'm arguing is that both for users and developers, only dehydrating successful queries might be a good default?

Either way, I agree we should add support for custom serializers and deserializers and an option to include errors when dehydrating. For personal reasons I don't think I can find time to implement that in the next month or so because I will be mostly afk, it was a stretch getting this PR together in time. If this is a blocker for the PR I would love it if someone else could take over that part, if not, I would love it if we made that part of a follow up.

Also, this is still an early API, we can also tweak things that doesn't work well in the v3 release?

Ephem:
Oh and if it's the `useHydrate` itself that is awkward, does introducing a `<ReactQueryHydrator queries={dehydratedQueries}>` help with that? I kind of think it does in the custom SSR case, and having a declarative API for it makes it a bit clearer that it's _the app itself_ that needs a set of `dehydratedQueries` to render properly?

**Activation of timeouts**

> As for sending queries to the client which are not rendered: I think these don't need to have a stale timeout as this only makes sense for queries which are being observed. The GC timeout could always be set (even on query creation) because this will most likely only trigger after a few minutes?

This is a good point! So is your suggestion then to remove the `activateTimeoutsManually` and `activateTimeouts` entirely from the public API, do that automatically behind the scenes for GC in `hydrate` and just let the `useBaseQuery` activate the staleness? Would require some reshuffling of the earlier refactor in `query`, but should be simple enough.

tannerlinsley:
I'm 100% on board with just skipping (and not dehydrating failed queries) by default. And I'm even okay not offering a way to do that until it comes up as a strong use case (if ever)/

tannerlinsley:
As for the `useHydrate` API, I  personally don't like it, nor do I want to add another API. I would rather just have it be:
```js
const cache = makeQueryCache()
const state = JSON.parse(window.__STATE__)
hydrate(cache, state)
ReactDOM.hydrate(
  <ReactQueryCacheProvider queryCache={cache}>
    <App />
  </ReactQueryCacheProvider>
)
```

And whatever we need to do to get to that point, I'm willing discuss. Keeping the API surface area the same size if possible is my intuition here.

Ephem:
I agree, but for Next we don't have control over the app at that point in time, so we sadly need to do it in render somehow for that case. :( This is a big whole in the API of Next in my opinion, a lot of examples for how to integrate Redux and other libraries are pretty ugly because of this too.

A big goal here was to hide that fact for users so that Next-users get a nice API to work with out of the box instead of having to implement this themselves. Because we need it anyway I decided to kind of embrace it for custom SSR as well to document a single consistent way of doing it instead of trying to explain when to use what if that makes sense?

Ephem:
Also, if you don't want to use `useHydrate` on the server with custom SSR, you can also do this, but I think it looks a bit wonky which is another reason I've recommended the `useHydrate`-approach:

```jsx
// Prefetch
const prefetchCache = makeQueryCache()
prefetchCache.prefetchQuery('key', fn)
const state = dehydrate(prefetchCache)

// Render
const renderCache = makeQueryCache()
hydrate(renderCache, state)
const html = ReactDOM.renderToString(
  <ReactQueryCacheProvider queryCache={renderCache}>
    <App />
  </ReactQueryCacheProvider>
)
res.send(`...`)
```

boschni:
**Timers**
Yes exactly! I think we can just schedule the GC in the query constructor as previously because this would be in the order of minutes. And if we schedule the stale timeout only whenever an observer subscribes, then there shouldn’t be any timing issues?

**Errors serialization**
While it is indeed true that there are a lot of different type of errors which could occur, only the ones that the user is actually handling in the UI are relevant. Which means the user should know exactly which errors could affect the UI. Which also means they should be able to write a serializer for it.

The serialization and deserialization could be done like this:  

```js
function dehydrate(cache, config) {
  const state = JSON.parse(JSON.stringify(cache.queries, config.serialize))
}

function hydrate(cache, state, config) {
  const queries = JSON.parse(state, config.deserialize)
}

dehydrate(cache, {
  serialize: (_key, value) => {
    if (value instanceof Error) {
      return { $type: "Error", message: value.message }
    }
    return value
  }
})

hydrate(cache, state, {
  deserialize: (_key, value) => {
    if (value?.$type === 'Error') {
      return new Error(value.message)
    }
    return value
  }
})
```

Or someone could choose to handle this without serialization in the query functions:

```js
 useQuery('key', async () => {
  try {
    return await api()
  } catch (error) {
    if (error instanceof Error) {
      throw error.message
    }
    throw error
  }
})
```

**Removing failed queries**
Could this be an option if someone wanted to do that?

```js
const cache = makeQueryCache()
cache.prefetchQuery('key', fn)
cache.removeQueries(query => query.state.isError)
const state = dehydrate(cache)
const html = ReactDOM.renderToString(
  <ReactQueryCacheProvider queryCache={cache}>
    <App />
  </ReactQueryCacheProvider>
)
```

Ephem:
**Timers**

Purely code-wise I still think we should keep the scheduling out of the constructor for different reasons unrelated to this PR, but I do agree we could and should schedule GC immediately on hydrate and let the observer schedule staleness which lets us remove the whole `activateTimeoutsManually` thing. I've updated the PR with this!

Ephem:
**Error serialization**

I talked to Tanner about this at length yesterday and we discussed very similar APIs to that. It took me a good nights sleep to realize that this isn't even a problem for React Query at all!

I even wrote this in the documentation yesterday:

> - This result is not in serialized form, you need to do that yourself if desired

So since `dehydrate` returns an un-serialized form, serialization is a user-level concern, not a framework one. So what's left to discuss is if we should include errors or not by default, and how to let the user toggle that.

I still think we should remove errors by default because I think that should be the majority use case, and the alternative requires a bunch of work on the users part. When it comes to how to toggle this we discussed a `includeErrors: true`-option yesterday, but now I realized that `shouldDehydrate` already covers this use case!

I've pushed an update that does this:

* Instead of having a hardcoded `query.state.status === 'success'` in `dehydrate`, this has been moved to be a default `shouldDehydrate`-function
* If you want to include all queries, provide your own `shouldDehydrate: () => true` or `shouldDehydrate: query => query.state.status !== 'loading'` if you want to exclude loading

Jordan-Gilliam:
@Ephem This is a killer release, not to mention 0 breaking changes. Thank you much!!

boschni:
Getting there!

**GC**
It might be good to active the GC timer in the constructor though, because if for some reason the timer is not activated, it'll stay in the cache forever. Although preferably the QueryCache would be responsible of GC instead of the Query itself.

**Serialization**
While I definitely agree the actual (de)serialization should be the responsibility of the user, I do think it would be nice if we would provide an easy and straightforward way for them to do so.

**Query filtering**
Would it be possible to only have one cache on the server? Similar to the example in https://github.com/tannerlinsley/react-query/pull/728#issuecomment-680697850 ?

**useHydrate**
Is this hook still needed?

Ephem:
**GC**

Yeah, we need to make sure to always schedule this internally. There is a caveat around prefetching that makes me want to avoid doing it in the constructor that I can explain elsewhere, but in its current form this is all internal implementation details and we can tweak this after the PR. Okay if we move this discussion elsewhere?

**Serialization**

I agree! We can always add custom `JSON.stringify`-replacers to the library in the future.

**New `hydration/ReactQueryCacheProvider`**

This hopefully addresses both your points about filtering and `useHydrate` at the same time. 😄  Yes, we do need `useHydrate` and/or some other way to ergonomically hydrate inside of render because of how Next works. In Next, we only have access to the dehydrated queries via props, which we only have access to in render.

Tanner came up with the idea of adding a new `ReactQueryCacheProvider` that you can import directly from the hydration entrypoint. This one has two additional props to the one you can import from the core package:

* initialQueries
* hydrationConfig

So instead of exposing a `<Hydrator />`-component, we bake that into CacheProvider instead. This means your custom SSR-example would become:

```jsx
// Server
const prefetchCache = makeQueryCache()
await prefetchCache.prefetchQuery('key', fn)
const initialQueries = dehydrate(prefetchCache)

const html = ReactDOM.renderToString(
  <ReactQueryCacheProvider initialQueries={initialQueries}>
    <App />
  </ReactQueryCacheProvider>
)
res.send(`
<html>
  <body>
    <div id="app">${html}</div>
    <script>window.__REACT_QUERY_INITIAL_QUERIES__ = ${JSON.stringify(initialQueries)};</script>
  </body>
</html>
`)

// Client
const initialQueries = JSON.parse(window.__REACT_QUERY_INITIAL_QUERIES__)
ReactDOM.hydrate(
  <ReactQueryCacheProvider initialQueries={initialQueries}>
    <App />
  </ReactQueryCacheProvider>
)
```

(I've also updated the Next example in the PR description)

Hopefully this looks okay? I'm open to adding a more generic `queryCache.filter`-function in the future that would when used right let you use the same cache on the server, but I don't think that should be the recommended approach mostly because I want to recommend a single way of doing it for Next and custom SSR.

I've pushed the `hydration/ReactQueryCacheProvider` as well as new docs.

@tannerlinsley This PR should now be up to date with what we talked about yesterday, but with the addition that you can now include errors in dehydration by using `shouldDehydrate: () => true` 🎉  (see [comment above](https://github.com/tannerlinsley/react-query/pull/728#issuecomment-680742084)). I've also added the `initialData`-approach back to the docs.

src/hydration/hydration.ts
```diff
@@ -0,0 +1,127 @@
+import { DEFAULT_STALE_TIME, DEFAULT_CACHE_TIME } from '../core/config'
+
+import type { Query, QueryCache, QueryKey, QueryConfig } from 'react-query'
+
+export interface DehydratedQueryConfig<TResult> {
```
boschni, line 5: Maybe also include the query key? It might not be possible to convert a hash back into a query key and each query can have it's own `queryKeySerializerFn`.

src/hydration/hydration.ts
```diff
@@ -0,0 +1,127 @@
...
+  staleTime?: number
+  cacheTime?: number
+  initialData?: TResult
+}
+
+export interface DehydratedQuery<TResult> {
```
boschni, line 11: Think the `<TResult>` is not needed anywhere in the hydration types as we don't really care about the actual data within the queries.

src/hydration/hydration.ts
```diff
@@ -0,0 +1,127 @@
...
+  config: DehydratedQueryConfig<TResult>
+  updatedAt: number
+}
+
+export interface DehydratedQueries<TResult> {
+  [hash: string]: DehydratedQuery<TResult>
+}
+
+export type QueryKeyParserFunction = (queryHash: string) => QueryKey
+export type ShouldHydrateFunction = <TResult>({
```
boschni, line 21: Should we provide this functionality? Maybe better to wait until someone has a request or use-case for it, otherwise we might need to support functionality which nobody uses

src/hydration/react.tsx
```diff
@@ -0,0 +1,61 @@
...
+  hydrationConfig,
+  children,
+}) => {
+  useHydrate(initialQueries, hydrationConfig)
+  return children as React.ReactElement<any>
+}
+
+export interface HydrationCacheProviderProps
+  extends ReactQueryCacheProviderProps {
+  initialQueries?: unknown
```
boschni, line 41: Might be good to have a more generic name for this like `dehydratedState`. The content should be opaque to the user because we might add other things besides just queries

docs/src/pages/docs/guides/ssr.md
```diff
@@ -5,59 +5,175 @@ title: SSR & Next.js
...
+> Note: The global `queryCache` you can import directly from 'react-query' does not cache queries on the server to avoid leaking sensitive information between requests.
+
+- Prefetch data
+  - Create a `prefetchQueryCache` specifically for prefetching by calling `const prefetchQueryCache = makeQueryCache()`
+  - Call `prefetchQueryCache.prefetchQuery(...)` to prefetch queries
+  - Dehydrate by using `const initialQueries = dehydrate(prefetchQueryCache)`
+- Render
+  - Wrap the app in `<ReactQueryCacheProvider>` from `'react-query/hydration'` and pass in `initialQueries`
+    - This makes sure a separate `queryCache` is created specifically for rendering
+    - **Do not** pass in the `prefetchQueryCache` from the last step, the server and client both needs to render from the dehydrated data to avoid React hydration mismatches. This is because queries with errors are excluded from dehydration by default.
```
boschni, line 124: This still feels a bit shaky. We could require users to filter themselves with:
```
prefetchCache.removeQueries(query => query.state.isError)
```
Or do it automatically in the `dehydrate` function?

docs/src/pages/docs/guides/ssr.md
```diff
@@ -5,59 +5,175 @@ title: SSR & Next.js
...
+### Integrating with Next.js
+
+To support caching queries on the server and set up hydration, you start with wrapping your application with `<ReactQueryCacheProvider>` in `_app.js`.
+
+> Note: You need to import `ReactQueryCacheProvider` from `'react-query/hydration'` for it to support hydration!
+
+```jsx
+// _app.jsx
+import { ReactQueryCacheProvider } from 'react-query/hydration'
+
```
boschni, line 65: Might be good to explicitly create the query cache here as the default one is going to be removed.

src/hydration/react.tsx
```diff
@@ -0,0 +1,61 @@
...
+import {
+  useQueryCache,
+  ReactQueryCacheProvider as CacheProvider,
+} from 'react-query'
+import { hydrate } from './hydration'
+
+import type { HydrateConfig } from './hydration'
+import type { ReactQueryCacheProviderProps } from '../react'
+
+export function useHydrate(queries: unknown, hydrateConfig?: HydrateConfig) {
```
boschni, line 11: With the `ReactQueryCacheProvider` component, think we can make this one private?

src/hydration/hydration.ts
Ephem: I like that, it simplifies things. So what would that make the data? Unknown?

src/hydration/hydration.ts
Ephem: Great question! I think we do, this came up in a discussion with Tanner and the use case is avoiding hydrating queries that have become "extra stale" kind of, that is, they are too old to even show the data while refetching in the background. We only know that at hydration time. Makes sense?

This is there for persisting to and from localstorage, not SSR, since using it with SSR would mean hydration mismatches.

src/hydration/react.tsx
Ephem: That's a good point. I'll rename it and will rename the `DehydratedQueries`-type and anywhere in the docs that `dehydratedQueries` is used as a variable name as well.

docs/src/pages/docs/guides/ssr.md
Ephem: My understanding is that the global cache is being removed, but that `ReactQueryCacheProvider` will still create a cache for you by default if you don't provide one yourself?

src/hydration/react.tsx
Ephem: I thought about that, but I think there are still usecases for this one, for example hydrating into an existing cache.

The current Next example is implemented entirely in `_app.js`, but there are also cases where you might want to hydrate in the page components instead (to not rely on a magical prop name for example).

docs/src/pages/docs/guides/ssr.md
Ephem: Even with that approach, the one documented now will still work too, so this comes down to what we should recommend as an approach in the docs. I think requiring `.removeQueries` in some cases but not others is equally shaky and confusing and would rather just recommend a single approach for Next and custom SSR, even with the above caveat.

I'm very open to adding `.removeQueries` later though, as well as changing the docs if we see this becoming too confusing for users.

As for making `dehydrate` remove it automatically, I really don't want to make `dehydrate` mutative and destructive for several reasons, main one being it can also be used with localstorage where you wouldn't want that.

src/hydration/hydration.ts
Ephem: This is a very good point and would let us remove the `queryHash` from the payload instead, as well as get rid of `queryKeyParserFn` which I've never liked. It requires changing the format around a little bit, I'll get on it!

docs/src/pages/docs/guides/ssr.md
boschni: Yes good point, the `dehydrate` function should indeed not be mutative. But people may also expect the `dehydrate` function to dehydrate everything by default. The `removeQueries` method already exists and using it would make it very clear to users on what is happening:
1. Prefetch
2. Filter out anything you don't want
3. Dehydrate
4. Render

There would be no implicit behaviour or juggling needed.

src/hydration/hydration.ts
boschni: I would say `unknown` if possible and otherwise just `any`.

docs/src/pages/docs/guides/ssr.md
Ephem: Oh, look at that! I had totally missed that the function already existed, nice! :)

There is definitely the case where you don't want to remove ahead of time, but only on the dehydrate step (for localstorage), but there is `shouldDehydrate` for that.

I am starting to warm to this a little bit, I still think there is a big case to be made for `dehydrate` filtering by default since that's the major usecase and we want to support that as easily as possible, but there really a big point to be made for no implicit behaviour as well.

For custom SSR, you might be able to set it up in a single place. The sad part though is that in Next, you would have to call `removeQueries` or provide a `shouldDehydrate` in every single page where you fetch data, and if you forget, you can quite possibly get errors in production (because no queries failed in dev). I think that's a major drawback.

src/hydration/hydration.ts
boschni: Maybe this use case is not hydration specific? For example: if the user is on a page and then after a while navigates to another page, then the data might also be too old to show there. The existence of a query in the cache should not really matter.

Ephem: Isn't `cacheTime` there for that on the client? Since we don't want to make that based on `updatedAt` like `staleTime`, we need something else for hydration.

boschni: The GC timer would not have started because the previous page was still showing it, but the data might be too old to show on the next page. You do bring up a good point though, should queries be hydrated when the cache time exceeded?

Ephem: Ah yes, I think there are probably other good arguments to be made that `cacheTime` is slightly different as well.

We do need to hydrate queries even if `cacheTime` is exceeded, or we will get React hydration mismatch warnings with SSR. :)

docs/src/pages/docs/guides/ssr.md
Ephem: Since `removeQueries` already exists and docs can be changed, what we need to decide is ultimately if `dehydrate` should only hydrate successful queries by default.

src/hydration/hydration.ts
Ephem: Fixed.

src/hydration/hydration.ts
Ephem: Fixed.

src/hydration/react.tsx
Ephem: Fixed.

src/hydration/react.tsx
boschni: Would it then be possible to only have `useHydrate` instead of `ReactQueryCacheProvider`? Then people can either do `hydrate(queryCache, state, config)` or `useHydrate(queryCache, state, config)`? Just trying to limit the number of possibilities if possible

Ephem: Most people are going to want to use `<ReactQueryCacheProvider dehydratedState={...}>`, so they'll end up implementing it themselves anyway. See earlier discussions in this PR and also, check out the Next implementation example in the docs: https://react-query-j5pum9ymn.vercel.app/docs/guides/ssr#integrating-with-nextjs

You can't `useHydrate` in `_app.js` because you don't have a `queryCache` on context yet, which is why `hydration/ReactQueryCacheProvider` was born. :)

docs/src/pages/docs/guides/ssr.md
boschni: Ah I'm actually not sure, maybe it was just my assumption it would become more explicit like with `ApolloProvider`. But maybe we should keep it like this then

src/hydration/hydration.ts
Ephem: I removed shouldHydrate for now. We can always add it back, or add the more generic mechanism later.

docs/src/pages/docs/guides/ssr.md
Ephem: The more I think about this the more I think we should filter unsuccessful queries by default. It's easier to get started with and is the safe default. Often when you develop you only have successful queries, so it's easy to end up with pretty bad errors in prod if you forget to add removeQueries somewhere (as in, the server might crash).

I'll go ahead and resolve this.

Ephem:
Thanks for the excellent comments, they have really helped improve this PR! From my point of view this is ready now.

PR #733, opened 2020-07-10
fix: docsearch is rendered twice with keyboard shortcuts

raoufswe:
the search modal appears twice when a user opens it with cmd + k. 

Screenshot 2020-07-10 at 9 46 43 PM


tannerlinsley:
Nice! Would mind doing the same for react table?

PR #742, opened 2020-07-11
fix: use generic Error type in ReactQueryProviderConfig.queries

### LINKED ISSUE #718: [Types] ReactQueryProviderConfig doesn't pass provided error type to `onError` callback
`ReactQueryProviderConfig` doesn't pass provided error type to `onError` callback

**Expected behavior**
`
const config: ReactQueryProviderConfig<IError> = {
 queries: {
    onError: (error) => {
      // error is type Error not IError
    },
  },
};
`
Screen Shot 2020-07-08 at 16 39 23

Bene8493:
Fixes #718

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -668,7 +668,7 @@ export function ReactQueryConfigProvider<TError = Error>(props: {
 }): React.ReactElement
 
 export interface ReactQueryProviderConfig<TError = Error> {
-  queries?: BaseQueryOptions & {
+  queries?: BaseQueryOptions<unknown, TError> & {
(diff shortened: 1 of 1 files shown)


SinanBaymammadli:
Wouldn't it be better to also add `TResult = unknown` to `ReactQueryProviderConfig` instead of just unknown?

Bene8493:
> Wouldn't it be better to also add `TResult = unknown` to `ReactQueryProviderConfig` instead of just `unknown`?

I think since there are usually multiple queries and mutations using a single config, unknown is ok.

PR #746, opened 2020-07-14
fix(types): remove promise from useMutation

wolverineks:


tannerlinsley:
To be fair I don’t think this was ever there.

PR #749, opened 2020-07-14
Fix makeQueryCache default caching behaviour in node

### LINKED ISSUE #706: Default makeQueryCache server behaviour does not match Readme
**Describe the bug**

This is either a bug or an error in the Readme.

**To Reproduce**

Readme says:

> If you create a queryCache manually with `makeQueryCache`, queries will be cached during SSR. Make sure you create a separate cache per request to avoid leaking data.

This is not true however, you have to explicitly call `makeQueryCache({ frozen: false })` to enable caching on the server.

**Expected behavior**

One of:

* Update Readme to match actual behaviour
* Update default behaviour to match Readme

**Additional context**

Release notes says:

> - `makeQueryCache` now accepts an optional configuration object. The `defaultConfig` object is used to override the default query configuration config to use inside of this cache. The `frozen` option if set to `true` will simulate React Query being run on the server, where no queries are cached for privacy and safety concerns. This is the default when using React Query on the server and is optional. You can also set it to `true` on the server and have it work as it would on the client. More information on this coming soon!

I'm guessing this behaviour was changed between #584 being merged and the release, but the info in the Readme was unintentionally not updated?

I don't have any strong feelings on this and would be happy to fix it either way, but I'd argue that creating a cache manually on the server only makes sense if you intend to use it for caching, making unfrozen a good default (but not for the global cache of course). I also don't see a security risk in this, since the developer is intentionally creating the cache.

Changing the default behaviour would kind of be a small breaking change though (intended for v2), but could also be considered a fix to match the Readme, depending. It would not affect people using `{ frozen: true/false }` explicitly, so should be fine.

I'd be happy to just update the Readme as well of course. 😄

Ephem:
Closes #706

This PR makes it so caches returned from `makeQueryCache` will always be unfrozen, and therefor cache data, by default. This now includes on the server.

The global cache on the server is still frozen by default and does not cache data, this is to avoid leaking data across requests as per #70.

This makes it so the behaviour of `makeQueryCache` matches the one described in the Readme.

I also made the tests around this more explicit.

### DIFF
--- a/src/core/queryCache.js
+++ b/src/core/queryCache.js
@@ -8,16 +8,16 @@ import {
 import { defaultConfigRef } from './config'
 import { makeQuery } from './query'
 
-export const queryCache = makeQueryCache()
+export const queryCache = makeQueryCache({ frozen: isServer })
 
 export const queryCaches = [queryCache]
 
-export function makeQueryCache({ frozen = isServer, defaultConfig } = {}) {
+export function makeQueryCache({ frozen = false, defaultConfig } = {}) {
   // A frozen cache does not add new queries to the cache
   const globalListeners = []
 
   const configRef = defaultConfig
-    ? { current: { ...defaultConfigRef.current, ...defaultConfig }}
+    ? { current: { ...defaultConfigRef.current, ...defaultConfig } }
     : defaultConfigRef
 
   const queryCache = {
--- a/src/react/tests/ssr.test.js
+++ b/src/react/tests/ssr.test.js
@@ -8,19 +8,44 @@ import {
   usePaginatedQuery,
   ReactQueryCacheProvider,
   queryCache,
-  queryCaches,
   makeQueryCache,
   useQuery,
 } from '../index'
 import { sleep } from './utils'
 
 describe('Server Side Rendering', () => {
-  describe('global cache', () => {
-    afterEach(() => {
-      queryCaches.forEach(cache => cache.clear({ notify: false }))
-    })
+  // A frozen cache does not cache any data. This is the default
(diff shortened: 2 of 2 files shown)


src/core/queryCache.js
```diff
@@ -8,16 +8,16 @@ import {
 import { defaultConfigRef } from './config'
 import { makeQuery } from './query'
 
-export const queryCache = makeQueryCache()
+export const queryCache = makeQueryCache({ frozen: isServer })
```
tannerlinsley, line 11: Hmm. I had someone trying to use ReactQuery in the CLI the other day. This would make that default to frozen, correct?

Ephem: Yes, the global cache will still be frozen by default in all server environments in this PR since changing that behavior would regress #70 and in my eyes be a breaking change.

I would suggest not only that we change this default behavior in the next major, but actually that we get rid of the global cache entirely (at least on the server side, possibly client too). So instead of making the global cache available for CLI-development, we push _everyone_ to create their own cache.

Creating a `queryCache` via `makeQueryCache` is trivial and you can easily just export that as a global cache yourself if that is what you want. What we would gain from this is that developers would explicitly have to create a cache somewhere, making it clear that whatever you fetch into this cache will be shared. This would let us get rid of the frozen/unfrozen concept, streamline the behavior and still make it hard to/avoid accidentally leaking data cross-request.

tannerlinsley: While I totally wish I had avoided the singleton from the start, it's going to be very difficult to get away from in the future. Even IMO, it's very convenient. I'd like to avoid it as long as possible ;)

PR #754, opened 2020-07-15
fix: reduce the number of renders

boschni:
Hi there! This PR includes the following changes:

- Make sure the hook return values match the public APIs.
- Make sure the hooks only re-render when something has changed.
- Make sure the hook return values are stable / memoized.
- Batch the isStale and succes update together when possible.
- Batch the failureCount and error update together when possible.

With these changes, the number of renders for successful and unsuccessful queries can be reduced from 4 to 2.

PS: While working on the tests, I noticed the `canFetchMore` and `isFetchingMore` properties are sometimes set to `undefined`. I guess this is not correct? For now I matched the assertions with the current behaviour.

### DIFF
--- a/src/core/utils.js
+++ b/src/core/utils.js
@@ -90,18 +90,33 @@ export function getQueryArgs(args) {
   return [queryKey, queryFn ? { ...config, queryFn } : config, ...rest]
 }
 
+export function deepEqual(a, b) {
+  return equal(a, b, true)
+}
+
+export function shallowEqual(a, b) {
+  return equal(a, b, false)
+}
+
 // This deep-equal is directly based on https://github.com/epoberezkin/fast-deep-equal.
 // The parts for comparing any non-JSON-supported values has been removed
-export function deepEqual(a, b) {
+function equal(a, b, deep, depth = 0) {
   if (a === b) return true
 
-  if (a && b && typeof a == 'object' && typeof b == 'object') {
+  if (
+    (deep || !depth) &&
+    a &&
+    b &&
+    typeof a == 'object' &&
+    typeof b == 'object'
+  ) {
     var length, i, keys
     if (Array.isArray(a)) {
       length = a.length
       // eslint-disable-next-line eqeqeq
       if (length != b.length) return false
-      for (i = length; i-- !== 0; ) if (!deepEqual(a[i], b[i])) return false
+      for (i = length; i-- !== 0; )
+        if (!equal(a[i], b[i], deep, depth + 1)) return false
       return true
     }
 
@@ -118,7 +133,7 @@ export function deepEqual(a, b) {
     for (i = length; i-- !== 0; ) {
       var key = keys[i]
 
-      if (!deepEqual(a[key], b[key])) return false
+      if (!equal(a[key], b[key], deep, depth + 1)) return false
     }
 
     return true
--- a/src/react/useBaseQuery.js
+++ b/src/react/useBaseQuery.js
@@ -3,11 +3,11 @@ import React from 'react'
 //
 
 import { useQueryCache } from './ReactQueryCacheProvider'
-import { useMountedCallback } from './utils'
+import { useRerenderer } from './utils'
 
 export function useBaseQuery(queryKey, config = {}) {
   // Make a rerender function
-  const rerender = useMountedCallback(React.useState()[1])
+  const rerender = useRerenderer()
 
   // Get the query cache
   const queryCache = useQueryCache()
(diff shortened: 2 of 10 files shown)


tannerlinsley:
Wow, this is a lot to take in all at once. I'll try and get to this asap.

boschni:
Happy to help! I have adjusted the PR with these changes:

- Removed the API normalisation in `useBaseQuery`. It would indeed be better to do this in a major version.
- Removed the state equality checks in `useBaseQuery` and moved them into `query` to make sure the state is only updated and dispatched when something has changed.

With these changes the number of renders is still reduced to 2 and I think this is actually a simpler and more performant solution :) But if you have any doubts about the implementation then I don’t mind splitting it up.

PS: I removed the `queryInfoRef` because with this implementation the reference would not get updated when a value inside the query info object changes. The idea with memoization is that the reference remains the same unless something has changed. This is more of a nice to have anyway because in most cases consumers will destructure the query info object instead of doing checks on the object itself.

tannerlinsley:
It’s looking better! So do you really think we need the shallow equal? If we are using a reducer we should be able to do an identity check instead, right?

boschni:
Currently the state always changes when an action is dispatched, regardless of whether some value has actually changed or not.

We could add manual checks in the reducer (or before dispatching actions) to prevent this.

For example:
```
case actionFetch:
  const status = typeof state.data !== 'undefined' ? statusSuccess : statusLoading
  const isFetching = true
  const failureCount = 0

  if (state.status === status && state.isFetching === isFetching && state.failureCount === failureCount) {
    return state
  }

  return {  ...state, status, isFetching, failureCount }
```
But this would need to be done in multiple places and is a bit prone to mistakes.

Doing a shallow check after the state update is probably the simplest and most bullet-proof solution.

tannerlinsley:
Alright, thats fair. How about we do this instead and just get rid of the shallow equal?
```js
// Do nothing if the state did not change
    if (deepIncludes(query.state, newState)) {
      return
    }
```

It passes your tests! ;)

boschni:
The `deepIncludes` function would be a bit slower because it does a deep comparison (including response data). Also, because it checks if `b` is a subset of `a` it does not really check if the objects are equal. One way to solve this would be to run the check two times like `deepIncludes(query.state, newState) && deepIncludes(newState, query.state)` but that would slow it down even more. Think the `shallowEqual` function is more suited for the job :)

tannerlinsley:
Okay. The primary motivation for that suggestion was just bundle size. So I’m good if that’s the only way to do it right.

tannerlinsley:
Can you pull in the latest from master?

boschni:
Updated 👍

src/react/useBaseQuery.js
```diff
@@ -41,15 +45,19 @@ export function useBaseQuery(queryKey, config = {}) {
...
+  const queryInfo = getQueryInfo(query)
+
+  // Handle suspense
+  handleSuspense(queryInfo)
 
-  const queryInfoRef = React.useRef({})
+  return queryInfo
+}
 
-  Object.assign(queryInfoRef.current, {
```
tannerlinsley, line 48: I made a few core changes around here that may make some of this unnecessary. Would you like to see if those changes you pulled in solve the issue without these changes?

boschni: Which changes are you referring to? PS: I combined the deepEqual and shallowEqual functions to reduce file size.

tannerlinsley: Master has a change that uses a Ref right here to stabilize the return value of useBaseQuery. What I'm getting as is that I think a lot of changes with regards to getBaseQueryInfo and all of the implementation to memoize the return values of the hooks might be able to be simplified. But I'm keeping an open mind. I need to look through the code again today.

src/core/utils.js
```diff
@@ -118,7 +133,7 @@ export function deepEqual(a, b) {
     for (i = length; i-- !== 0; ) {
       var key = keys[i]
 
-      if (!deepEqual(a[key], b[key])) return false
```
tannerlinsley, line 121: This is fantastic!

src/react/useBaseQuery.js
boschni: Ah yes I reverted that change (you can find the reasoning for that in https://github.com/tannerlinsley/react-query/pull/754#issuecomment-659252841). But I can also remove the `getBaseQueryInfo` part if you want. It is not needed for memoization but it did allow to move the handling of suspense into the base query hook. Maybe it would be an idea to move the pagination behaviour into the core and enable it with `config.paginated = true` for example just like the infinite query? Then the specific hooks would only do some configuration and all of the behaviour would be in the core or base query hook.

boschni:
Updated the branch again and removed the `getBaseQueryInfo` part

kamranayub:
@tannerlinsley @boschni Hey folks! I am suspecting this PR caused some regressions in our test suite regarding when `isStale` is set, to help us avoid re-fetching a query if it isn't stale. I still have to dig into why the test is failing, but the suite was passing with 2.5.4. It could be I need to adjust our logic that digs into the cache to check `isStale` but I'll let you know what I find.

PR #767, opened 2020-07-17
feat: add typescript support

boschni:
Hi there! While working on the project as a newcomer I noticed it was challenging at times to know which type of arguments some functions accept, which properties should be available on certain objects or which type a certain value should have.

Some of this information can be gathered from the docs, some of it from the tests, and some of it from the type definition file. But the docs only document the public APIs and are (together with the types) not always in sync with the code. Which means spending time on reading and debugging the code itself, while not knowing for sure what the intended behaviour exactly is.

I think adding types to the code will help people understand the project more easily and provide more confidence when making changes because mistakes can be caught early. It would also reduce the number of type related issues and knowing the project is written in TypeScript can be a selling point for TypeScript users.

This PR is mostly intended to get the discussion going and to see what would be needed to add support for it. At first glance not much is needed, but there would be some work involved in typing the code and it could potentially uncover some bugs here and there.

Would you be interested in exploring this direction?


tannerlinsley:
Welcome to the long line of people that have and will continue to try and convince me to use TS for OSS libraries :) I hope it works this time! Here's my standard thoughts when it comes to TS and more or less my requirements for migrating to TS:

- I don't want any breaking changes to the code or removal of the type defs for the first attempt. It can and should be a gradual and augmentative process. As the internal types get better with time, we can decide when the right time is to replace the type definitions with the actual generated types from the library.
- I will always prefer `any`, `unknown` or something flexible (and maybe not 100% type-safe) before changing the dynamism of the code in a way that would make it less flexible than it already is. TS is notorious for putting restrictions on API design. Usually this is a good thing, but not always.
- I don't want it to be intimidating or difficult for newcomers to contribute to the library. I realize that if you know TS, then technically it makes it easier to contribute and not mess things ups, but at the same time, people who don't know TS are now a bit less enthused to contribute. If we can ease that feeling by making the types as clean and as inferred as possible, I may be swayed.
- I am by no means a TS-first developer, but I feel comfortable saying that if I can't understand a type signature or typing strategy in TS, I won't want to use it, nonetheless expose it to my users and expect them to understand it.
- I have tried converting both React Query and React Table to TS a few times and get stuck pretty easily due to the dynamic nature of both libraries and their flexible APIs that, while being logically sound and well tested, are very very difficult to statically type.

With all that said, I am open to it. I can't give any guarantees that I won't smash my computer in a TS rage and put it off for another year... but I'm willing to try.

Thoughts?

boschni:
Hope it works out too let’s see ;) I think your requirements make sense and I actually agree to all of them.

It should definitely be a gradual process while keeping everything backwards compatible, including the current type definitions. This gives us time to work on the internal typings without affecting anyone. At some point the new types could be published as breaking change, or removed if it does turn out to be too difficult. Because transpilation is still done with Babel, and TypeScript is only used as advanced linter, it should be pretty easy to revert if needed. This setup at least gives us the possibility to try without much risk.

How would you like to proceed with this?

tannerlinsley:
I think the best way forward would be to start "migrating" the types from the type definitions into the primitive-ish interfaces of the library. Queries, instances, caches, etc. Not really focus on typing all of the functions and what not yet, but just working on all of the different versions of the objects and interfaces, all with the primary goal of just not touching any of the original source code other than simply adding type references to the objects going in and out. Sure, there will be tons of errors at that point. But once it's done, that will be a good point to know if and how much the library actually needs to change logically to accomodate the types. If it's only a little, we'll make the changes and stay non-breaking :) If it's a lot, then we'll just have to discuss on how to go forward (complex types, a breaking rewrite, etc). But I think ingesting as much of the type definitions as possible and seeing what's left will be a good start.

src/core/query.ts
```diff
@@ -13,6 +13,18 @@ import {
...
 import { makeQueryInstance } from './queryInstance'
 
+export interface QueryState<T, TError = Error> {
+  data?: T
+  error?: TError | null
+  failureCount: number
+  isFetching: boolean
+  canFetchMore?: boolean
+  isStale: boolean
+  status: 'idle' | 'loading' | 'error' | 'success'
```
ayush987goyal, line 23: Instead of inlining, we should probably make a explicit `type` for `status` which will be easier for consumers and can potentially be reused across.

tannerlinsley: Yes. Make all the things reusable. In fact you’ll see soon that status is an enum that has constants in the utils file.

tannerlinsley:
Run into anything tricky yet?

boschni:
I managed to get almost everything typed while fixing some bugs along the way. The tricky parts are these:
- Different configurations are merged and augmented at various stages, this makes is a bit unclear which configuration belongs to which stage.
- `Query` accepts a single configuration object, but there are multiple type of configurations (regular/paginated/infinity). This makes it a bit hard to know which properties are available at what time.
- In `Query`, the `data` property is `T[]` in `infinite` mode, but `T` in regular mode. This is a bit hard to type. Also the returned values are different in each mode.

A possible solution would be to create specific queries (like we do with the hooks), which encapsulate different behaviours instead of having them all in Query.

What do you think of the MR? The types might not be 100% yet, but they could be refined along the way as they are not exposed yet.

tannerlinsley:
Nothing jumped out at me on my initial pass through. Moving to classes was interesting... but I realize that they might be easier to type than object closures. I really dislike classes, but if that's a necessary change, then I think I can live with it.

boschni:
Yeah because classes have a certain structure and behaviour they are a lot easier to type. I have simplified the typings a bit and I believe the PR is now ready to be merged. The only conflicts left are in `query.ts` because of the `data` type mismatch between infinite and regular queries. This can be solved in multiple ways which I would like to explore in an additional PR :)

tannerlinsley:
You'll need to ingest the changes done in #754. They were very very important for reducing renders.

boschni:
Updated 👍

tannerlinsley:
So what's the status on this?

boschni:
Everything is typed so I think this one is ready to be merged. There are some type casts here and there to cope with the different type of queries (as they have different configurations and different type of results) but this probably needs to be solved with some refactoring.

Some ideas on this:
- Splitting `Query` up into specific queries like `InfiniteQuery` and `PaginatedQuery`. These can then have their own configurations, behaviours and result types.
- Keep using `Query` for all type of queries but add a different property for infinite query data like `infiniteData` (similar to `resolvedData` with pagination), which could then be typed as array. Plus adding a discriminator property to the configuration object which could then be used to check what type of configuration is passed (like `type: 'infinite'`). The results would still need to be casted though.

tannerlinsley:
Would you to hop on a call and review this PR?

boschni:
Yeah sure just let me know how to contact you

tannerlinsley:
Jump on here and let's chat: https://discord.com/invite/WrRKjPJ

src/core/types.ts
```diff
@@ -0,0 +1,187 @@
+import { Query, FetchMoreOptions } from './query'
+
+export type QueryKey = string | ArrayQueryKey
```
tannerlinsley, line 3: JSON Serializable

Valid keys:
'posts'
{ type: 'posts' }
['posts']
['posts', { filter: 'published' }]

Invalid keys:
() => {}
[() => {}]
['todos', { fn: () => {} }]

src/core/config.ts
```diff
@@ -1,6 +1,41 @@
...
+    ? queryKey
+    : [queryKey]
+
+  if (
+    arrayQueryKey[0] === '' ||
+    typeof arrayQueryKey[0] !== 'string' ||
+    arrayQueryKey.some(d => typeof d === 'function')
+  ) {
+    throw new Error('A valid query key is required!')
+  }
```
tannerlinsley, line 30: We can catch bad keys like this, I think?

`JSON.stringify(key, (key, value) => { if (typeof value === 'function') { throw new Error('Bad Key') }})`

src/react/useMutation.ts
```diff
@@ -13,19 +11,64 @@ import {
...
 })
 
-const actionReset = {}
-const actionLoading = {}
-const actionResolve = {}
-const actionReject = {}
+const actionReset = 'Reset'
+const actionLoading = 'Loading'
+const actionResolve = 'Resolve'
+const actionReject = 'Reject'
```
ayush987goyal, line 66: Let's have an `enum` for these types and use them in above interfaces and in the below comparisons. It would help in consistency. It could be something like:
```ts
enum ActionType {
	RESET = 'Reset',
	LOADING = 'Loading',
	...
}

interface ResetAction {
	type: ActionType.RESET
}

...

if (action.type === ActionType.RESET) {
	...
}
```

src/core/types.ts
boschni: Fixed

src/react/useMutation.ts
boschni: Good suggestion, done!

boschni:
I have processed the comments and added additional tests to check if the typings are correct with `expectType` and `@ts-expect-error` assertions.

The PR should be backwards compatible except for these changes:

Functional changes:
- The query key can now be any JSON serializable value.
- The hook results now match with the public API and do not spread all query properties.

Types changes to keep them simple:
- The query function arguments are not inferred from the query key anymore, which means developers are responsible to keep them in sync.
- The hook results are not discriminated unions anymore, which means the `data` and `error` properties need to be checked.

The new types can be published with `npm run gen:types` but I think it would be good if some people first try them out in their projects to get some feedback. The `dtslint` tool can be removed after publishing.

From my perspective this PR is ready to be merged.

src/react/ReactQueryConfigProvider.tsx
```diff
@@ -1,19 +1,30 @@
...
+)
 
 export function useConfigContext() {
   return React.useContext(configContext) || defaultConfigRef.current
 }
 
-export function ReactQueryConfigProvider({ config, children }) {
-  let configContextValueOrDefault = useConfigContext()
-  let configContextValue = React.useContext(configContext)
+export interface ReactQueryProviderConfig extends ReactQueryConfig {}
```
ayush987goyal, line 14: Why is this needed? We should ideally be using `ReactQueryConfig` itself at other places right?

boschni: I mostly extended here to keep the exported type name the same as the current one. The provider could accept different options in the future though, but I don't think it is likely

src/react/useMutation.ts
```diff
@@ -0,0 +1,196 @@
...
+        status: QueryStatus.Success,
+        data: action.data,
+      }
+    case ActionType.Reject:
+      return {
+        status: QueryStatus.Error,
+        error: action.error,
+      }
+    default:
+      throw new Error()
```
ayush987goyal, line 84: Is this case reachable? Does TS complain about not having it?

src/core/queryInstance.ts
```diff
@@ -0,0 +1,104 @@
...
+    onStateUpdate: OnStateUpdateFunction<TResult, TError>
+  ) {
+    this.id = uid()
+    this.onStateUpdate = onStateUpdate
+    this.query = query
+    this.config = {}
+  }
+
+  clearInterval(): void {
+    clearInterval(this.refetchIntervalId)
```
ayush987goyal, line 32: Do we need a check wether `refetchIntervalId` is present before we clear it?

src/core/queryInstance.ts
```diff
@@ -0,0 +1,104 @@
...
+    delete this.refetchIntervalId
```
ayush987goyal, line 33: Do you think instead of using `delete`, we should use the type of `refetchIntervalId` as `number | null` and just set it as `null` on init and clear?

src/core/query.ts
```diff
@@ -0,0 +1,653 @@
...
+}
+
+export interface QueryState<TResult, TError> {
+  canFetchMore?: boolean
+  data?: TResult
+  error: TError | null
+  failureCount: number
+  isError: boolean
+  isFetching: boolean
+  isFetchingMore?: false | 'previous' | 'next'
```
ayush987goyal, line 40: If we are using this type at multiple places (I think in Infinite query), let's have them as explicit type in core types.

ayush987goyal:
> The query function arguments are not inferred from the query key anymore, which means developers are responsible for keeping them in sync.

Is there any specific reason for this change? I have used this at quite a few places since most often then not, the params that you use in your queryFunction are the ones that would need to go in the queryKey.

src/react/useMutation.ts
boschni: The case should not be reachable but TS does not complain about it, changed to `return state` though to keep it in line with regular reducers

src/core/query.ts
boschni: Yes good one, created a type for it

src/core/queryInstance.ts
boschni: I don't have a strong opinion on `delete` vs `null` but this could be a separate PR as it is not really related to TS

boschni:
> Is there any specific reason for this change? I have used this at quite a few places since most often then not, the params that you use in your queryFunction are the ones that would need to go in the queryKey.

It is mostly to keep the types simple for other contributors but also for library consumers that want to create their own hooks. Personally I don't have a strong opinion on it though it is open for discussion. The impact on keeping this functionality can be seen here: https://github.com/boschni/react-query/commit/0ff7972680476a62301e2cefaa1dfce080764012

ayush987goyal:
Hi @boschni , I was wondering if we want to rename the types etc. in core which are prefixed with React (eg. `ReactQueryConfig`) to *not* have it? I am saying this because we are going to migrate `core` to a separate package and then consume it at multiple places (eg. `react-query`, `vue-query` etc.). If we have React specific types and exports in core, it would be confusing for other consumers. I suggest we remove those prefixes and have a generic names for them (eg. `CoreQueryConfig`). You might not have to worry about breaking current users because we are anyway targeting TS migration to v3. Thoughts? @tannerlinsley @boschni

boschni:
Hi @ayush987goyal , I think eventually the `react-query` package should still expose a `ReactQueryConfig` type because there might be configuration options specific to React (like for example the `suspense` and `useErrorBoundary` options). Other frameworks might have different ones.

When creating a separate `QueryCache` or `QueryClient` library, it would be nice if it was truly framework agnostic. Meaning it should be generic and flexible enough to support these kind of framework specific functionalities while not knowing about them, provide APIs for things like pagination and infinite queries, and it should be straightforward to use outside of any view layer.

The actual logic around suspense could then hopefully be implemented within `react-query`. What I mean is: does it make sense to separate these types now, even though we don’t really know yet how the new library types would look like? Maybe some things like extracting any React specific functionality from the query cache should be explored first in the current repo?

src/core/queryInstance.ts
boschni: Added checks before clearing

src/core/queryInstance.ts
boschni: Removed `delete`, they are now set to `undefined`

johnnyjeannatasha:
> Hi @boschni , I was wondering if we want to rename the types etc. in core which are prefixed with React (eg. `ReactQueryConfig`) to *not* have it? I am saying this because we are going to migrate `core` to a separate package and then consume it at multiple places (eg. `react-query`, `vue-query` etc.). If we have React specific types and exports in core, it would be confusing for other consumers. I suggest we remove those prefixes and have a generic names for them (eg. `CoreQueryConfig`). You might not have to worry about breaking current users because we are anyway targeting TS migration to v3. Thoughts? @tannerlinsley @boschni 

Yes. I mentioned that before in a text... Reply.

PR #773, opened 2020-07-17
docs: refactor search modal

MrLeebo:
This PR refactors the Search logic to allow you to launch a search from the landing page and cleans up the Search component generally.

  - state/Modal has been moved into a SearchProvider and passed down via context
  - `<Search />` is basically just the UI "button" and doesn't render the modal anymore
  - keyboard shortcut works everywhere (even without rendering `<Search />`)
  - rendering `<Search />` twice won't cause the keyboard shortcut to render 2 modals
  - removed props, states, and variables that were unused/misspelled
  - renders the Search "button" in the nav for larger displays (not changed on tablet/mobile)

Screen Shot 2020-07-17 at 10 34 43 AM

### DIFF
--- a/docs/src/components/Nav.js
+++ b/docs/src/components/Nav.js
@@ -2,12 +2,13 @@ import * as React from 'react'
 import Link from 'next/link'
 import logoSrc from '../images/logo.svg'
 import { siteConfig } from 'siteConfig'
+import { Search } from './Search'
 
 export const Nav = () => (
   <div className="bg-white border-b border-gray-200">
     <div className="container mx-auto">
-      <div className="grid grid-cols-1 md:grid-cols-12 md:gap-6">
-        <div className="w-60 col-span-3 flex items-center h-16 pt-4 md:pt-0">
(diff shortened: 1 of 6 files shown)


tannerlinsley:
WOW! Love this change.

I did notice some weird layout squish when you get smaller with the menu:

Screen Shot 2020-07-17 at 10 23 17 AM

With that fixed, I love it! Would you be willing to apply the same changes to the React Table documentation? (It uses the exact clone of these docs)

MrLeebo:
Interesting, I'll take a look. I tested each of the device breakpoints in chrome but forgot to try resizing the window itself. Also, looks like a force pushed commit changed some commit hashes in master

MrLeebo:
@tannerlinsley I fixed the layout issue and made the same change in react-table, which was torturous because I'm still on v6 so none of those shiny new docs are any good to me 😾

tannerlinsley:
Yeah, there was a bad commit message in the history that was causing a major version increment, so I had to go in and clobber some things. Thanks for updating! :)

PR #774, opened 2020-07-17
feat(queryCache): merge defaultConfig options

TuckerWhitehouse:
Providing a `defaultConfig` to `makeQueryCache` will merge `shared`, `queries`, and `mutations`
objects instead of overwriting. This allows you to configure only some options.

### DIFF
--- a/src/core/tests/queryCache.test.js
+++ b/src/core/tests/queryCache.test.js
@@ -1,5 +1,6 @@
 import { sleep } from './utils'
 import { queryCache, queryCaches } from '../'
+import { makeQueryCache } from '../queryCache'
 
 describe('queryCache', () => {
   afterEach(() => {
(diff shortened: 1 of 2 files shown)


src/core/tests/queryCache.test.js
```diff
@@ -224,4 +228,15 @@ describe('queryCache', () => {
...
+
+  test('makeQueryCache merges defaultConfig so providing a queryFn does not overwrite the default queryKeySerializerFn', async () => {
+    const queryFn = () => 'data'
+    const queryCache = makeQueryCache({
+      defaultConfig: { queries: { queryFn } },
+    })
+
+    expect(() => queryCache.buildQuery('test')).not.toThrow(
+      'config.queryKeySerializerFn is not a function'
+    )
```
TuckerWhitehouse, line 240: Open to suggestions on a better way to validate this -- I started by writing this test expecting the exception because that is the current behavior, switched to `.not.toThrow` and applied my code changes.

tannerlinsley: I'm not sure either. But we can make it better in the future if we need to. If tests pass and you think this is good enough, then let me know and I'll merge :)

TuckerWhitehouse: If you're happy with it, the tests pass and it resolves the issue I was having :)
I had also considered exposing `configRef` on the `queryCache` but wasn't sure if increasing the public api to support testing was a good idea 🤔

tannerlinsley: I'd rather not expose more api lol. I'm doing too much of that as it is.

PR #808, opened 2020-07-29
fix(QueryCache): Use default configuration used for QueryCache instance for context

kamranayub:
## Context

See test cases but when using `useQuery` with no config provider and a custom query cache provider, the `defaultConfig` was not being merged correctly in `queryCache.buildQuery`.

A previously failing test:

```js
test('uses defaultConfig for queries when they don\'t provide their own config', async () => {
    const cache = makeQueryCache({
      defaultConfig: {
        queries: {
          staleTime: Infinity,
        },
      },
    })

    function Page() {
      const { data } = useQuery('test', async () => {
        await sleep(10)
        return 'test'
      })

      return (
        <div>
          <h1>{data}</h1>
        </div>
      )
    }

    const rendered = render(
      <ReactQueryCacheProvider queryCache={cache}>
        <Page />
      </ReactQueryCacheProvider>
    )

    await waitFor(() => rendered.getByText('test'))

    expect(cache.getQuery('test')).toBeDefined()
    expect(cache.getQuery('test').config.staleTime).toBe(Infinity)
    cache.clear({ notify: false })
  })
```

`useConfigContext` was falling back to `defaultConfigRef` but when using `makeQueryCache`, the ref is not updated (on purpose) when merging in `defaultConfig`.

Now the actual merged `configRef` for the query cache is exposed so that we can access it in `useConfigContext`.

## Wrong `queryCache` variable being passed

I fixed another two issues with `buildQuery` and `notifyGlobalListeners` that was passing the `queryCache` module variable instead of `this` in the new refactored TS code.

## Exposing the config ref

~I'm not _sure_ whether you want it exposed off the instance like that or through some other means (since this would also consumers to mess with it?). However, it does address the issue.~

I am exposing it through `QueryCache.getDefaultConfig` which could be used to freeze or clone the config to prevent mutation.

Exposing it like this would mean:

- It's in the public API
- It should be added to API docs
- People could mess with it

I can document this as a new API, I could freeze it, whatever we think would make sense. We just need to get a reference to the merged queryCache-specific config.

## Changes

- Use `useQueryCache` in `useConfigContext` to fallback to queryCache's `configRef`
- Expose `QueryCache.getDefaultConfig()` method
- Fix passing default `queryCache` ref in `notifyGlobalListeners`
- Fix passing default `queryCache` ref in `buildQuery`


src/react/ReactQueryConfigProvider.js
```diff
@@ -1,12 +1,14 @@
...
+import { useQueryCache } from './ReactQueryCacheProvider'
 
 //
 
 const configContext = React.createContext()
 
 export function useConfigContext() {
-  return React.useContext(configContext) || defaultConfigRef.current
+  const queryCache = useQueryCache()
+  return React.useContext(configContext) || queryCache.configRef.current
```
kamranayub, line 11: The default `queryCache.configRef` will be `defaultConfigRef` out-of-the-box unless you're using `makeQueryCache`.

src/core/queryCache.js
```diff
@@ -49,6 +49,8 @@ export function makeQueryCache({ frozen = false, defaultConfig } = {}) {
     globalListeners.forEach(d => d(queryCache, query))
   }
 
+  queryCache.configRef = configRef;
```
kamranayub, line 52: We just need access to this. We could hide behind an accessor.

```suggestion
  queryCache.getDefaultConfig = () => configRef.current; // could now clone, freeze, etc. to prevent mutation
```

kamranayub:
I am currently "integration" testing this in our repo.

src/core/queryCache.js
kamranayub: I've switched to `getDefaultConfig()` accessor which should allow some flexibility in the future to freeze/clone/etc.

kamranayub:
Just seeing weird issues locally when testing where the query cache I am creating and providing via `ReactQueryCacheProvider` isn't actually being used in the component. Not sure yet what is causing that, whether it's a package linking issue or an actual issue with the code.

kamranayub:
OK, the issue is that there were two major bugs with `makeQueryCache` and `QueryCache` usage because they were unknowingly passing `queryCache` module variable in a few places accidentally. I will fix those issues in this too.

kamranayub:
Alrighty, this is working now as I would expect!

tannerlinsley:
Fantastic!

PR #817, opened 2020-07-30
fix: check if the document global exists before usage

boschni:
Closes https://github.com/tannerlinsley/react-query/issues/816


renatobenks:
great catch @boschni, it's a nice a patch

PR #825, opened 2020-08-03
refactor: remove gc flag

boschni:


tannerlinsley:
Very nice. This was added by a PR a while back and I can't remember why it even needed to be there.

boschni:
Yeah I could not see why the flag is needed and it also triggers a re-render. So I thought, why not remove it :)

PR #827, opened 2020-08-03
fix: fetch query when any instance is enabled

### LINKED ISSUE #823: Query not refetching when enabled becomes 'true' and different instance is disabled
**Describe the bug**
I have 2 Components using the same query. At one point one instance changes the enabled flag to true while the other stays false. The query is not executed.

**To Reproduce**
I created a CodeSandbox: https://codesandbox.io/s/zen-almeida-g2o5z
Click the button to enable one of the queries => nothing happens. If you remove the <DisabledQuery /> component it works as expected.

**Expected behavior**
A query fetches when **any** instance is enabled.

benediktviebahn:
This PR fixes #823.
In updateConfig some() was used in a way that doesn't make sense to me. I think this is the way it should be, but let me know if there is anything special about it.

### DIFF
--- a/src/core/queryInstance.ts
+++ b/src/core/queryInstance.ts
@@ -59,10 +59,10 @@ export class QueryInstance<TResult, TError> {
       ) {
         this.refetchIntervalId = setInterval(() => {
           if (
-            this.query.instances.some(_ => this.config.enabled) &&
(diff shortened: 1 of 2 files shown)


tannerlinsley:
Wonderful!

Issue #830, opened 2020-08-04
useMutation types are off


probablyup:
I'm seeing this type:

```ts
export type MutateFunction<
  TResult,
  TVariables,
  TError = Error
> = TVariables extends undefined
  ? (options?: MutateOptions<TResult, TVariables, TError>) => Promise<TResult>
  : (
      variables: TVariables,
      options?: MutateOptions<TResult, TVariables, TError>
    ) => Promise<TResult>
```

Which implies that you can do this:

```ts
mutate({
  onSuccess() { }
})
```

But that functionality to assume that the first argument is options and not variables doesn't appear to be implemented in code. I think it's a mistake and the type should be adjusted to:

```ts
export type MutateFunction<
  TResult,
  TVariables,
  TError = Error
> = (
      variables: TVariables,
      options?: MutateOptions<TResult, TVariables, TError>
    ) => Promise<TResult>
```

tannerlinsley:
The new types have been released! Huzzah!

cgarrovillo:
Which version fixed these types?

PR #833, opened 2020-08-05
Update index.d.ts

probablyup:
Per #830


tannerlinsley:
The new native types are coming soon, so with that said, is this change still necessary?

probablyup:
I don't know, where are they so I can test them?

tannerlinsley:
https://react-query.tanstack.com/docs/typescript

OR, you can build the types in the repo with `yarn gen:types`

Tanner Linsley

PR #840, opened 2020-08-06
feat: add notify flag for controlling re-renders

boschni:
Some applications might not have indicators in the UI to indicate if data is stale or if some query is refetching/revalidating. In these cases, the UI also does not need to re-render if any of these states change. It would be sufficient to only re-render when the actual `data` or `error` properties change. This flag resembles the `notifyOnNetworkStatusChange` flag from Apollo.


tannerlinsley:
Conflicts ;)

boschni:
Resolved :)

docs/src/pages/docs/comparison.md
```diff
@@ -38,6 +37,8 @@ Feature/Capability Key:
 | Prefetching APIs                             | ✅                                     | 🔶                         | ✅                                  |
 | Query Cancellation                           | ✅                                     | 🛑                         | 🛑                                  |
 | Partial Query Matching<sup>2</sup>           | ✅                                     | 🛑                         | 🛑                                  |
+| Stale While Revalidate                       | ✅                                     | ✅                         | 🛑                                  |
+| Stale Time Configuration                     | ✅                                     | 🛑                         | 🛑                                  |
```
tannerlinsley, line 41: I do believe swr has this. No?

PR #871, opened 2020-08-14
refactor: define configuration merging strategy

boschni:
The current config merging looked a bit confusing so I tried to simplify this a bit and to define a formal strategy:

When using hooks the config will be merged in the following order:

1. Defaults.
2. Defaults from the hook query cache.
3. Combined defaults from any config providers in the tree.
4. Query/mutation config provided to the hook.

When using a query cache directly the config will be merged in the following order:

1. Defaults.
2. Defaults from the query cache.
3. Query/mutation config provided to the query cache method.


tannerlinsley:
Ready?

boschni:
Yes this one is ready to be merged

PR #897, opened 2020-08-20
feat: publish new types

boschni:
This MR includes the following changes:
- Remove custom types.
- Build generated types on release.
- Remove type-only imports and exports from the declaration files to support TypeScript versions below 3.8.
- Remove the type connection between the query key and the query function parameters because this does not match with the API. Parameters can be filtered with `queryFnParamsFilter` or added with for example the `fetchMoreVariable` variable. This also simplifies the typings for less experienced TS users.


tannerlinsley:
Just needs a rebase ;)

robbtraister:
This appears to be a breaking change wrt typescript. In attempting to upgrade, I've noticed at least the following issues (with the latter being the more problematic of the two).

- `QueryOptions` has been renamed `QueryConfig`
- `queryResult.isSuccess` no longer implies that `queryResult.data` is defined

Should I open a new issue?

PR #919, opened 2020-08-24
feat: add refetch on reconnect functionality

boschni:


src/core/index.ts
```diff
@@ -1,13 +1,7 @@
 export { queryCache, queryCaches, makeQueryCache } from './queryCache'
 export { setFocusHandler } from './setFocusHandler'
-export {
-  CancelledError,
-  deepIncludes,
-  isCancelledError,
-  isError,
-  setConsole,
-  stableStringify,
```
9still, line 9: @boschni could you explain why `stableStringify` is no longer exposed?   I realize this util was not documented, it *was* a part of the public interface exposed by the library and it ended up being a breaking change for us... It was convenient for key comparison...

boschni: Hi! It has been removed because it is an undocumented internal function which happened to be used for key serialization (it was also not exposed as a key serialization function). Could you explain a bit more about your use case? Then we might be able to come up with a proper API for it.

9still: @boshini Thanks for the reply! Our use-case is simple.  We have some generic functionality in our app, which wraps react query, and simply wants to detect if the query key has changed.  Because keys can be heterogeneous in structure, there is no good way to compare them.  We could deep compare the objects on our own, but that requires pulling in a separate implementation, and more importantly risks not fully aligning with the comparison logic used by `react-query`.  We assumed that since the serialization routine was exported, it was effectively the supported if undocumented approach @tannerlinsley had exposed to do just that. 

I realize that you can rightly view these utils as a leaky abstraction, and truly appreciate all the work you guys are doing, but just a kind request as a semver consideration for the future:

Since the removed utils *were* a part of the external interface, it seems more appropriate to reserve the cleanup for a major version, with at least a comment on the PR explaining the change.

boschni: Added `getDefaultReactQueryConfig()`in https://github.com/tannerlinsley/react-query/pull/1001 which can be to get the default `queryKeySerializerFn` . Would that be sufficient?

9still: That sounds awesome!  Will check it out on Monday.  Thanks so much for adding it so quickly!!!

9still: @boschni We've switched using the new method you added & it works great!  Thanks again!

PR #920, opened 2020-08-24
Update comparison.md

shuding:
This part is not true:

> SWR uses component styling/visibility to only prefetch the next page and does not provide the capability to "lag" previous query results while new ones load without much non-trivial hackery.

SWR has a pagination API that supports the capability ([docs](https://swr.vercel.app/docs/pagination#use-swr-infinite)). "component styling/visibility" is just a code pattern and it's never required by SWR.

### DIFF
--- a/docs/src/pages/docs/comparison.md
+++ b/docs/src/pages/docs/comparison.md
@@ -26,7 +26,7 @@ Feature/Capability Key:
 | Polling/Intervals                            | ✅                                     | ✅                                             | ✅                                                                        |
 | Parallel Queries                             | ✅                                     | ✅                                             | ✅                                                                        |
 | Dependent Queries                            | ✅                                     | ✅                                             | ✅                                                                        |
-| Paginated Queries                            | ✅                                     | 🛑<sup>1</sup>                                 | ✅                                                                        |
+| Paginated Queries                            | ✅                                     | ✅                                             | ✅                                                                        |
 | Infinite Queries                             | ✅                                     | ✅                                             | ✅                                                                        |
 | Initial Data                                 | ✅                                     | ✅                                             | ✅                                                                        |
 | Scroll Recovery                              | ✅                                     | ✅                                             | ✅                                                                        |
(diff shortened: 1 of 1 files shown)


tannerlinsley:
I believe the docs are still very accurate. It's important to understand that paginated or "lagged" queries are not the same as infinite queries. You can clearly see that there is a ✅ next to `Infinite Queries` for SWR as per your documentation link. As far as I know, SWR does not have the capability to lag query results in either the `useSWR` or `useSWRInfinite` hooks.

shuding:
If that's your definition, you should rename "Paginated Queries" to "Lagged Queries". Otherwise it will be misleading and people will think that SWR doesn't support pagination at all. It's not necessary to have that "lagged data" for pagination. I probably want to see the loading screen of the next page immediately.

tannerlinsley:
We can for sure talk about a better name. Lagged queries isn’t half bad actually. Regardless of the name, I wouldn’t down play the functionality. That very type of UX is what is inspiring and driving Suspense and CM (but for an entire app). Its always better not to clobber data UI to show a loader when it’s avoidable.

styfle:
Why not use something like:

✅ Paginated Queries<sup>1</sup>

Then its clear it supports Paginated Queries but the description below could mention the lagged queries?

tannerlinsley:
I've updated the docs to reflect that SWR is capable of pagination (which is really just being able to change the query key to reflect a specific page), and added an additional row for "Lagged Queries" and an explanation behind it. This further revealed that Apollo also does not have the capability for lagged queries and also that lagged queries apply to infinitely loaded queries as well. So win win!

The important thing to note here (for posterity) is that query lagging essentially gives the same user experience as suspense will in the future. It allows you to persist your current view while the next view loads without the jarring experience of seeing a loading state. Of course, this is partially avoidable via prefetching, but obviously you wouldn't want to disable a button to load the next page until it's been prefetched. Hence, the awesomeness of this feature.

Thanks @styfle and @shuding!

PR #933, opened 2020-08-27
feat: determine staleness locally instead of globally

boschni:
This MR allows users to define different stale times for different situations. Instead of marking a whole query as stale, the staleness of a query is determined when consuming the query. The rationale for this can be found here: https://github.com/tannerlinsley/react-query/discussions/770#discussioncomment-42352 . This also resolves some of the confusion about what happens if different hooks define different stale times.


tannerlinsley:
This is genius.

tannerlinsley:
Do you know if this is something we should highlight in the comparison docs? Or does everyone else have this too...?

tannerlinsley:
This good to merge?

boschni:
Added a comparison to the docs and integrated the hydration changes. Should be good to go now!

PR #942, opened 2020-08-31
fix: always use config from last query execution

boschni:
Currently the query configuration is dependent on the order in which hooks are rendered and this could result in unexpected behaviour. This MR will make sure the query configuration is only changed on execution. Related to https://github.com/tannerlinsley/react-query/discussions/932


tannerlinsley:
Excellent

tannerlinsley:
It's weird that tests are failing randomly. Any ideas here?

PR #943, opened 2020-08-31
feat: remove ReactQueryCacheProvider from hydration

boschni:
This MR removes the `hydration/ReactQueryCacheProvider` component and replaces it with a `Hydrate` component. The `hydration/ReactQueryCacheProvider` component has some mixed responsibilities and having two of them could be a bit confusing. Exposing a `Hydrate` component simplifies the API because we only export three things which all basically do the same thing: `hydrate()`, `<Hydrate />` and `useHydrate()`. @Ephem

### DIFF
--- a/docs/src/pages/docs/api.md
+++ b/docs/src/pages/docs/api.md
@@ -792,13 +792,13 @@ setConsole({
 
 ## `hydration/dehydrate`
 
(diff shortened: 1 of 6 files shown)


Ephem:
Hey! I wont have time to really review this PR because of personal stuff happening, but I'm absolutely behind the idea. `<Hydrate />` was my first choice for the component API as well and I still prefer that to what got merged in the end.

tannerlinsley:
Yep yep yep. I'll get this out asap.

PR #944, opened 2020-08-31
fix: make sure initial data is used when switching queries

boschni:
Fixes https://github.com/tannerlinsley/react-query/issues/927


tannerlinsley:
Beautiful!

PR #949, opened 2020-09-01
fix: make sure queries with only inactive observers are also invalidated

boschni:
Fixes https://github.com/tannerlinsley/react-query/issues/947


jsangaritar:
Hi, im using `"react-query": "^3.24.3"` and this issue is still happening, i have a query defined with `enable: false` and when i want to invalidate it does not work with that config.

PR #954, opened 2020-09-02
feat: add forceFetchOnMount flag

boschni:


tannerlinsley:
So in v3, is this just going to become `refetchOnMount: false` and the new default is to only refetch when it's stale?

boschni:
I'm not 100% sure on the naming but this was my reasoning so far:
- `fetchOnMount`: could be confusing because if set to `false` it would still fetch if a query has not been fetched yet or when the query is stale.
- `refetchOnMount`: could be confusing because if set to `false` it would still refetch if a query is stale.
- `forceFetchOnMount`: if set to `false` it will not force a fetch. Think this one is the least ambiguous, but it's not the prettiest name.

tannerlinsley:
At least in v3 (because it would be a breaking change), and coming from https://github.com/tannerlinsley/react-query/pull/789, I would like to consolidate into a single option called `refetchOnMount` that defaults to `false`. This way, it's very clear when a query is refetched by default:
- when the query is stale and an "interaction" event happens like mounting, window focus, network online

Thoughts?

boschni:
Sounds like a plan! Think ‘forceFetchOnMount’ is a little bit less ambiguous, but ‘refetchOnMount’ definitely looks nicer and with some documentation it should be clear enough on what it does.

tannerlinsley:
Do you want to implement that here and we can close #789?

boschni:
Version 3 might take a while, maybe get this one in and then rename the property to refetchOnMount in v3?

tannerlinsley:
Yep. I'm good with that. Add one more, delete 3 later ;)

PR #958, opened 2020-09-02
docs(react-native): add solution for fullscreen error

strdr4605:


donni106:
newer React Native versions does not throw errors in full screen anymore. is this somehow important to know on this page?
I think the info for ability of setting the consoles can be still useful.

PR #966, opened 2020-09-05
fix: make sure setQueryData is not considered as initial data

boschni:


kentcdodds:
This fixed: https://github.com/tannerlinsley/react-query/discussions/962

PR #1009, opened 2020-09-13
refactor: remove query status bools

boschni:


tannerlinsley:
Why did you remove the Bools?

boschni:
I guess they are not really needed as they are also added in the observer?

tannerlinsley:
Ohhhh so they’re still there? Sorry I’m on my phone lol. Didn’t see the observer

PR #1021, opened 2020-09-16
docs: remove shared config

boschni:


src/core/config.ts
```diff
@@ -68,7 +68,6 @@ export const DEFAULT_CONFIG: ReactQueryConfig = {
 
 export function getDefaultReactQueryConfig() {
   return {
-    shared: { ...DEFAULT_CONFIG.shared },
```
tannerlinsley, line 71: Is this a breaking change?

boschni: The function returns the default config and shared is typed optional so this should not be breaking. Also this function has just been introduced in the previous release ;)

PR #1022, opened 2020-09-16
feat: add remove method and deprecate clear

boschni:


src/core/query.ts
```diff
@@ -206,7 +206,14 @@ export class Query<TResult, TError> {
     })
   }
 
+  /**
+   * @deprecated
+   */
   clear(): void {
```
tannerlinsley, line 212: Can we add a console log here to warn?

boschni: Added

renatobenks:
shouldn't it be a major release instead? 🤔 it's a breaking change, isn't it?

tannerlinsley:
The functionality is exactly the same for now. The only difference is you will see a deprecation notice until you change to the new function name.

PR #1068, opened 2020-09-21
feat: add new types

amen-souissi:
This MR includes the following changes:

- Add overloaded **getQueryArgs** function (TS: with typed args)
- Use typed config for the  **isResolvedQueryConfig** function, use 'queryHash' in config instead of Boolean(...)
- **DehydratedQuery** now is dynamic


tannerlinsley:
These types will no longer be required in v3, so going to close this in favor of the WIP branch for v3 :) Thanks!

PR #1079, opened 2020-09-22
feat: initial v3 changes

boschni:


src/core/query.ts
```diff
@@ -13,145 +14,140 @@ import {
...
-  Invalidate,
-}
-
 interface SetDataOptions {
   updatedAt?: number
 }
 
 interface FailedAction {
-  type: ActionType.Failed
+  type: 'failed'
```
tannerlinsley, line 65: I think this is a good move.

docs/src/pages/docs/guides/migrating-to-react-query-3.md
```diff
@@ -0,0 +1,241 @@
...
+useQuery({
+  queryKey: 'posts',
+  queryFn: fetchPosts,
+  staleTime: Infinity,
+})
+```
+
+### queryCache.invalidateQueries()
+
+The `client.invalidateQueries()` method will now only invalidate queries.
```
tannerlinsley, line 106: I thought `invalidateQueries` was still going to be "smart", eg. by refetching active queries and marking inactive ones as stale. Is there no way to do that now?

tannerlinsley: The idea behind this was to allow users a very efficient way to do:
```js
client.invalidateQueries('posts', { active: false, inactive: true })
client.refetchQueries('posts', { active: true, inactive: false })
```

boschni: What I like about the separation is that it is really clear what each method does. The `invalidateQueries()` method marks queries as invalid and the `refetchQueries()` method refetches queries. Most people probably just want to refetch?

tannerlinsley: I thought so too. But then you have a long season of cache behind you and trigger refetches for a query key prefix for example, you run the risk of triggering a massive amount of requests from all of the inactive cache entries. However you also want to mark those as stale.

tannerlinsley: But I see your point. Essentially you would suggest using both with the correct inactive and active filters, right?

boschni: In this implementation `refetchQueries()` only fetches active queries by default to prevent over-fetching. But it does not also invalidate inactive queries when specifying a prefix. Think the `refetchQueries()` concept is simpler and more easy to grasp for users, but they might "forget" that when specifying a prefix, there could be inactive queries that might need to be marked as invalid. Hmmm

tannerlinsley: I love it when you go "Hmmmm..." Good things usually come soon after lol.

PR #1089, opened 2020-09-24
Adjust link into overview page

iaurg:
I've adjusted the link to the installation page, I think that's the right page.


tannerlinsley:
Remove the .md and it will work

iaurg:
@tannerlinsley I've removed the md and separate the text into 2 links: guides -> instalation and api -> api docs. Thanks for help.

PR #1092, opened 2020-09-24
docs: fix typo in Important Defaults

evanfreeze:
I was reading the docs and noticed this wasn't quite correct. I assume "it makes your app more performant" is what's intended here, but "it will make your app more performant" could work as well.


tannerlinsley:
Nice catch!

Issue #772, opened 2020-07-17
V3 Umbrella Issue


tannerlinsley:
Below are the official breaking changes and features for v3. If you have a feature request or idea, please use the Discussions tab to open a new discussion about that idea.

Global:
- [x] Remove usePaginatedQuery.
- [x] Remove ReactQueryConfig.shared as it only contains one value.
- [x] Change QueryConfig.enabled type to boolean but keep behavior.
- [x] Change query key type to match with documentation (string | any[]).
- [x] Remove react-query.mjs in favor of unbundled ESM files.
- [x] QueryCache and Query async methods should throw and hook methods should not.

Query:
- [x] Rename clear() to remove().
- [x] Remove refetch() in favor of QueryObserver.refetch(). The refetch() method on Query is redundant as the functionality is the same as fetch().
- [x] Remove subscribe() as it is only used in the tests.
- [x] Remove fetchMoreVariable argument from fetchMore() in favor of FetchMoreOptions.

QueryCache:
- [x] Remove global query cache.
- [x] Remove makeQueryCache in favor of new QueryCache().
- [x] Remove resetErrorBoundaries().
- [x] Remove prefetch() options (force/throwOnError).
- [x] Rename prefetchQuery() to fetchQuery().
- [x] Change isFetching to isFetching() and do not store state.
- [x] Remove true predicate option from getQueries() in favor of undefined.

ReactQueryCacheProvider
- [x] Make usage required. (Non-breaking, documented)
- [x] Do not create query cache automatically.

QueryObserver:
- [x] Rename clear() to remove(). (Non-breaking deprecation)

Query Config:
- [x] Remove forceFetchOnMount.
- [x] Change refetchOnMount false behavior to disable refetch on mount for that specific hook

smashercosmo:
Would be nice to address this somehow in v3 https://github.com/tannerlinsley/react-query/discussions/669

Danscap:
@tannerlinsley is there any support for class components for react? not directly but rendering a functional component with query hooks into a class component.

kamranayub:
For V3, I'd propose allowing extending the `QueryCache` class when calling `makeQueryCache` (or providing a custom cache).

Use case: I am writing test utilities that allow you to mock named queries and right now the `QueryCache` methods for getting/setting only work with exact query key matches but when testing, we want to support partial matching.

To handle this, I have to mutate the queryCache that comes back to override methods when it would be nicer to extend `QueryCache` instead.

```js
[...]

PR #1093, opened 2020-09-24
refactor(types): expose types for `react-query/es/*`

buschtoens:
By using `typesVersions` instead of `types` pointing to a single `index` declaration file, we enable consumers of `react-query` to also import directly from nested files, like `react-query/core`. These files are de-facto importable, but TypeScript does not "see" them currently.

I understand that in regular React day to day usage, this is irrelevant (or potentially even undesirable), but I am currently working on building a `react-query` clone for Ember.js and would love to use the rock-solid `react-query/core` as a foundation for this.

For `react-query` v3 I'd like to suggest splitting out the `core` lib into a standalone package to make code-sharing easier. It could still reside in this repo by switching to a yarn/lerna monorepo.

Thank you for considering! 😊

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,14 @@
   },
   "main": "lib/index.js",
   "unpkg": "dist/react-query.development.js",
-  "types": "types/index.d.ts",
+  "typesVersions": {
+    "*": {
+      "*": [
+        "types/*",
+        "types/*/index"
+      ]
+    }
+  },
   "module": "es/index.js",
   "scripts": {
     "test": "is-ci \"test:ci\" \"test:dev\"",
--- a/package.json
+++ b/package.json
@@ -12,11 +12,12 @@
   },
   "main": "lib/index.js",
   "unpkg": "dist/react-query.development.js",
+  "types": "types/index.d.ts",
   "typesVersions": {
     "*": {
-      "*": [
+      "es/*": [
         "types/*",
-        "types/*/index"
+        "types/*/index.d.ts"
       ]
     }
   },


boschni:
Tested this change in a project but it's seems to break the regular `react-query` import. The current package also does not have an entry point for `react-query/core` because the index file resides at `react-query/lib/core` or `react-query/es/core`. With this change TypeScript would think there is an entry point while there is none.

package.json
```diff
@@ -12,7 +12,14 @@
...
   "unpkg": "dist/react-query.development.js",
-  "types": "types/index.d.ts",
+  "typesVersions": {
+    "*": {
+      "*": [
+        "types/*",
+        "types/*/index"
+      ]
+    }
+  },
```
buschtoens, line 22:
```suggestion
  "types": "types/index.d.ts",
  "typesVersions": {
    "*": {
      "es/*": [
        "types/*",
        "types/*/index.d.ts"
      ]
    }
  },
```

buschtoens:
@boschni Thank you for looking into this, and sorry, that I broke it. 😓
I've restored the original `types` export, so that the standard pathless import from `react-query` works again. `react-query/hydration` also works, since the `.d.ts` file is located alongside the `.js` file in standard location. I've also corrected the nested imports, so that `react-query/es/*` works correctly as well.

I've also verified that all these imports resolve correctly:

```ts
import { useInfiniteQuery, useQuery } from 'react-query';
import type { Query } from 'react-query/es/core';
import type { QueryCache } from 'react-query/es/core/queryCache';
import type { ReactQueryConfig } from 'react-query/es/core/types';
import type { HydrateProps } from 'react-query/hydration';
```

boschni:
No problem! I made some changes in the V3 branch to create multiple entry points: https://github.com/tannerlinsley/react-query/pull/1079. With these changes it should be possible to do the following, while supporting multiple export formats and types:
```js
import { Query } from 'react-query/core'
import { useQuery } from 'react-query/react'
import { HydrateProps } from 'react-query/hydration'
```

buschtoens:
Are you implying that you would rather hold back with this until v3 (#772) lands and close this PR or do you think that we could still ship this improvement for v2? 😊 
I'd greatly appreciate that, but I can understand, if you would like to avoid the `/es/` churn. If so, do you a rough ETA when the first pre-release versions will be released?

boschni:
Yes I would rather hold this one off in favor of the v3 release, I hope we can push out a pre-release this week

buschtoens:
Awesome! Looking forward to that. Thank you.

Issue #1115, opened 2020-09-30
infiniteQuery:getFetchMore() - continues to fetch data even if `false` is returned


pvomacka:
**Describe the bug**
Prerequisities: Having list of 7 items, fetched page by page, where page has 3 items. 

In case infiniteQuery fetches whole list (containing 3 pages) and then on another page one item from the list is removed. Then after returning back on page with the list, it sends 3 requests where the last one contains `false` as param, as `false` was returned from `getFetchMore()` function, because after second page there is no more data to load. Returning `false` from `getFetchMore()` should stop sending requests. The issue is reproduced in codesandbox.io - link in the steps to reproduce section. 

**To Reproduce**
1. Go to https://codesandbox.io/s/ancient-snow-0pqdp?file=/pages/index.js (Modified example from your documentation)
2. Turn on Developers Console -> Network Tab
3. Scroll down until  'Project 6' is loaded (which is the last item in the list)
4. Click on the link "Go to another page and remove one project"
5. Wait till deleting is completed
6. Click on "Go back and show the list without the last item"
7. Requests with `?page=false` are visible in Developers Console -> Network Tab

**Expected behavior**
After returning `false` from getFetchMore(), no more requests are sent

**Additional context**
Related discussion link: https://github.com/tannerlinsley/react-query/discussions/1114

p-siriphanthong:
I think this problem is not `getFetchMore` problem but it is the refreshing problem.

When `refetch` is called, there are always `n` fetch requests where `n` is the number of pages in the cache.

In your case.

1. Scroll down until 'Project 6' is loaded -> There are 3 pages in the cache.
2. Go to removing page and wait till deleting is completed -> There are only 2 pages left.
3. Go back to list page -> The `refetch` will be called and request fetching data 3 times (number of pages in the cache).

I already create a pull request (#1124) to fix this problem 😄

PR #1124, opened 2020-10-01
fix: refetching on infinite query when a page was deleted

p-siriphanthong:
Issue: #1115

### DIFF
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -474,17 +474,18 @@ export class Query<TResult, TError> {
     }
 
     // Create function to fetch the data
-    const fetchData = () => {
+    const fetchData = async () => {
       if (isFetchingMore) {
         return fetchPage(prevPages, previous, fetchMoreVariable)
       } else if (!prevPages.length) {
         return fetchPage([])
       } else {
-        let promise = fetchPage([])
-        for (let i = 1; i < prevPages.length; i++) {
-          promise = promise.then(fetchPage)
+        let pages: TResult[] = []
+        for (let i = 0; i < prevPages.length; i++) {
+          pages = await fetchPage(pages)
+          if (!hasMorePages(config, pages)) break
         }
-        return promise
+        return pages
       }
     }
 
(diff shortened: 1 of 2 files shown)


src/core/query.ts
```diff
@@ -474,17 +474,18 @@ export class Query<TResult, TError> {
...
         return fetchPage(prevPages, previous, fetchMoreVariable)
       } else if (!prevPages.length) {
         return fetchPage([])
       } else {
-        let promise = fetchPage([])
-        for (let i = 1; i < prevPages.length; i++) {
-          promise = promise.then(fetchPage)
+        let pages: TResult[] = []
+        for (let i = 0; i < prevPages.length; i++) {
+          pages = await fetchPage(pages)
```
boschni, line 485: Thanks for the PR! The for loop with async await will add quite a bit of runtime code to the bundle. You should also be able to fix this in the `fetchPage` function as the result of that function is the final list of pages.

src/core/query.ts
```diff
@@ -452,39 +452,47 @@ export class Query<TResult, TError> {
...
         typeof lastPage !== 'undefined' &&
         config.getFetchMore
       ) {
-        cursor = config.getFetchMore(lastPage, pages)
+        cursor = config.getFetchMore(lastPage, newPages)
       }
 
-      const page = await config.queryFn(...params, cursor)
+      while (i < numberOfPages) {
+        lastPage = await config.queryFn(...params, cursor)
```
boschni, line 476: This will still generate a lot of polyfill code because of the async await usage in iterations.

But I think the only change that is really needed in the original code is:
```js
cursor = config.getFetchMore(lastPage, pages)
```
to:
```js
cursor = config.getFetchMore(lastPage, pages)

if (typeof cursor === 'undefined' || cursor === false) {
  return pages
}
```

p-siriphanthong: That's a nice idea, I will update the code. 👍

tannerlinsley:
Excellent!!

Issue #1100, opened 2020-09-26
build does not work on windows terminal


jauniuse:
**Describe the bug**
`yarn build` does not work on windows terminal. It's because `rm -rf` does not exist on windows terminal - suggestion is to use [rimraf](https://github.com/isaacs/rimraf). Assignment of environment variables do not work on windows terminal as well - suggestion is to use [cross-env](https://github.com/kentcdodds/cross-env). Tried and fixed this locally.

**To Reproduce**
1. Run `yarn build` from windows terminal.

**Expected behavior**
`yarn build` should work on windows terminal.

PR #1129, opened 2020-10-04
build: add building support on Windows terminal

jueinin:
replace `rm -rf` to rimraf, and use `cross-env` to set env variable. so that we can build on Windows terminal~


jueinin:
#1100

PR #1130, opened 2020-10-04
feat: split mutate into mutate and mutateAsync

boschni:
Currently the `mutate` function returns a promise which by default returns `undefined` on error. This behavior seems to confuse some people as they initially expect the promise to behave like a regular promise. There is a `throwOnError` option to make the promise behave like a regular promise, but this option seems to be overlooked and does not work well with the types. Returning a different type based on an option is possible (although a bit cumbersome), but it's impossible to guarantee the correctness because it can also be defined globally. This PR splits the functionality into `mutate` and `mutateAsync`. The `mutate` function returns nothing and the `mutateAsync` function returns a regular promise.

Issue https://github.com/tannerlinsley/react-query/issues/1111

### DIFF
--- a/docs/src/pages/guides/mutations.md
+++ b/docs/src/pages/guides/mutations.md
@@ -130,7 +130,7 @@ useMutation(addTodo, {
   onError: (error, variables, context) => {
     // An error happened!
     if (context.rollback) {
-      rollbac()
+      context.rollback()
     }
   },
   onSuccess: (data, variables, context) => {
--- a/docs/src/pages/guides/mutations.md
+++ b/docs/src/pages/guides/mutations.md
@@ -9,22 +9,25 @@ Here's an example of a mutation that adds a new todo the server:
 
 ```js
 function App() {
-  const [
-    mutate,
-    { isLoading, isError, isSuccess, data, error },
-  ] = useMutation(newTodo => axios.post('/todods', newTodo))
+  const mutation = useMutation(newTodo => axios.post('/todods', newTodo))
 
   return (
     <div>
-      {isLoading ? (
+      {mutation.isLoading ? (
         'Adding todo...'
       ) : (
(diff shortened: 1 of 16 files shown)


tannerlinsley:
Sounds like we should deprecate the original and switch to the async version in v3?

tannerlinsley:
Not a fan of the combined returned variable. The naming of the actual function is nice as opposed to requiring the use of thing.mutate() syntax

boschni:
The downside of returning a regular promise by default is that it would require users to always wrap the mutate function:
```js
const [mutate] = useMutation()
return (
  <button onClick={async () => {
    try {
      await mutate()
    } catch {}
  }} />
)
```
Instead of:
```js
const [mutate] = useMutation()
return (
  <button onClick={mutate} />
)
```
Was also a bit in doubt on the object return value, but it does align more with `useQuery` and destructuring is still possible:
```js
const { mutate } = useMutation()
return (
  <button onClick={mutate} />
)
```
Or:
```js
const { mutate: addTodo } = useMutation()
return (
  <button onClick={addTodo} />
)
```

docs/src/pages/guides/mutations.md
```diff
@@ -111,73 +116,88 @@ const CreateTodo = () => {
...
+      },
     }
-  }
-  onError: (error, variables, rollbackVariable) => {
+  },
+  onError: (error, variables, context) => {
     // An error happened!
-    if (rollbackVariable) rollbackVariable()
+    if (context.rollback) {
+      rollbac()
```
tannerlinsley, line 133:
```suggestion
      context.rollback()
```

tannerlinsley:
Alright then, I'm on board. :) This is going to be a big breaking change for docs, examples, and all related courses on the library, but I think it's a good one.

PR #1133, opened 2020-10-04
docs: Elaborate on usage of queries

cseas:
There's a bit of a learning curve to react-query, especially when it comes to understanding the difference between queries and mutations and how to use react-query for a POST method.

I've added some explanation to the docs based on the discussion in #801 

Currently, the algolia search for "post" leads to results from the "Invalidation from Mutations" document which might not exactly be relevant for someone just trying to make a basic POST call. 

This change will hopefully make the "Queries" section show up in results for "post" too since a simple query might be ideal to use when the POST function isn't modifying server data. The docs currently don't mention clearly that queries can work fine for POST calls.

Please feel free to edit the words to a better-suited language if possible but it'd really be great if the docs explicitly mentioned that queries can be used with POST calls.

### DIFF
--- a/docs/src/pages/docs/guides/queries.md
+++ b/docs/src/pages/docs/guides/queries.md
@@ -5,7 +5,7 @@ title: Queries
 
 ## Query Basics
 
-In React Query, a query is a declarative dependency on some asynchronous source of data.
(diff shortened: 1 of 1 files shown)


tannerlinsley:
Please make this same PR to the next branch?

cseas:
Same PR made on next branch. Thanks for merging!

PR #1135, opened 2020-10-04
feat: add bi-directional infinite query support

boschni:
PR for discussion:

Adds ability to automatically support both directions:
```js
const {
  fetchNextPage,
  fetchPreviousPage,
  hasNextPage,
  hasPreviousPage,
  isFetchingNextPage,
  isFetchingPreviousPage
} = useInfiniteQuery(key, fn, {
  getNextPageParam: lastPage => lastPage.nextId,
  getPreviousPageParam: firstPage => firstPage.prevId
})

return (
  <>
    <button onClick={fetchPreviousPage} disabled={!hasPreviousPage || isFetchingPreviousPage}>Load previous</button>
     ...
    <button onClick={fetchNextPage} disabled={!hasNextPage || isFetchingNextPage}>Load next</button>
  </>
)
```


tannerlinsley:
I like it. Though, personally the append/prepend option is strange to think about. Technically this means that you can fetchNextPage(), but it would get prepended, correct? (and visa versa)

I realize that might be helpful for like... a news feed concept. But wouldn't that just end up being a visual reversal at render time and not at the data model?

boschni:
Yeah the append mode is indeed for those kind of scenarios, where new pages appear at the top instead of at the bottom. Reversing the list when rendering would have the same effect. Mostly added it for convenience and because the reporter of https://github.com/tannerlinsley/react-query/issues/1086 was trying to load previous pages instead of reversing the list. Don’t have a strong opinion on this option though

tannerlinsley:
The amount of code to support the option is tiny and it also won’t carry a lot of opinion with it into the future. So I think it’s fine.

boschni:
This one should be ready. Also includes support for refetching when manual page params are provided

tannerlinsley:
Excellent usage of the select option to avoid the reversal option.

PR #1141, opened 2020-10-06
docs: add Relay to comparison + added extra bullet points

RichardLindhout:


RichardLindhout:
Scroll Recovery -> I can't find the build in functionality for this in Apollo Client
Network Status Refetching -> I can't find the build in functionality for this in Apollo Client

tannerlinsley:
Looks like a good start. @boschni , thoughts? I feel like we could tick a yellow box at the very least on offline?

maraisr:
I think Relay would green tick "Generated type safety for input & response", as it can output both TS and Flow types.

docs/src/pages/docs/comparison.md
```diff
@@ -12,40 +12,47 @@ Feature/Capability Key:
...
+| GraphQL Fragments                            | 🛑                                     | 🛑                         | ✅                                    | ✅                                    |
+| GraphQL Subscriptions                        | 🛑                                     | 🛑                         | ✅                                    | ✅                                    |
+| Persisted Queries (APQ)                      | 🛑                                     | 🛑                         | ✅                                    | ✅                                    |
+| GraphQL Connections                          | 🛑                                     | 🛑                         | ✅                                    | ✅                                    |
+| Data masking                                 | 🛑                                     | 🛑                         | 🛑                                    | ✅                                    |
+| Generated type safety for input & response   | 🛑                                     | 🛑                         | 🟡                                    | 🟡                                    |
```
alloy, line 54:
```suggestion
| Generated type safety for input & response   | 🛑                                     | 🛑                         | 🟡                                    | ✅                                    |
```

docs/src/pages/docs/comparison.md
```diff
@@ -12,40 +12,47 @@ Feature/Capability Key:
...
+| Lagged / "Lazy" Queries<sup>1</sup>          | ✅                                     | 🛑                         | 🛑                                    | ✅                                    |
+| Initial Data                                 | ✅                                     | ✅                         | ✅                                    | ✅                                    |
+| Scroll Recovery                              | ✅                                     | ✅                         | ✅                                    | 🛑                                    |
+| Cache Manipulation                           | ✅                                     | ✅                         | ✅                                    | ✅                                    |
+| Outdated Query Dismissal                     | ✅                                     | ✅                         | ✅                                    | ✅                                    |
+| Render Optimization<sup>2</sup>              | ✅                                     | 🛑                         | 🛑                                    | ✅                                    |
```
alloy, line 35: The footnote for this doesn't really explain to me if it will only update a tree of components when a query has changed or if, like Relay, only the individual components that have subscribed to pieces of the result set get re-rendered when their data changes. Which is it?

docs/src/pages/docs/comparison.md
```diff
@@ -12,40 +12,47 @@ Feature/Capability Key:
...
+| Dependent Queries                            | ✅                                     | ✅                         | ✅                                    | ✅                                    |
+| Paginated Queries                            | ✅                                     | ✅                         | ✅                                    | ✅                                    |
+| Infinite Queries                             | ✅                                     | ✅                         | ✅                                    | ✅                                    |
+| Lagged / "Lazy" Queries<sup>1</sup>          | ✅                                     | 🛑                         | 🛑                                    | ✅                                    |
+| Initial Data                                 | ✅                                     | ✅                         | ✅                                    | ✅                                    |
+| Scroll Recovery                              | ✅                                     | ✅                         | ✅                                    | 🛑                                    |
```
alloy, line 32: What does scroll recovery mean in the case of data fetching libraries? Isn’t this a routing concern?

docs/src/pages/docs/comparison.md
```diff
@@ -12,40 +12,47 @@ Feature/Capability Key:
...
+| GraphQL Subscriptions                        | 🛑                                     | 🛑                         | ✅                                    | ✅                                    |
+| Persisted Queries (APQ)                      | 🛑                                     | 🛑                         | ✅                                    | ✅                                    |
+| GraphQL Connections                          | 🛑                                     | 🛑                         | ✅                                    | ✅                                    |
+| Data masking                                 | 🛑                                     | 🛑                         | 🛑                                    | ✅                                    |
+| Generated type safety for input & response   | 🛑                                     | 🛑                         | 🟡                                    | 🟡                                    |
+| Offline support                              | 🛑                                     | 🛑                         | 🛑                                    | 🟡                                    |
```
alloy, line 55: Does it make sense to add a row about moving runtime overhead to compile time? Which is one of the reasons people both like and dislike Relay.

boschni:
Think a lot of these points are somewhat difficult to compare because the "GraphQL" support React Query provides basically relies on the capabilities of the used GraphQL client. A lightweight client like graphql-request can be used, but it's also possible to use a heavier one like Apollo. Apart from that, think fragments and connections should be supported. The type safety is dependent on the technology used as types can be generated from Swagger, GraphQL schemas or Protocol buffer definitions. Offline support sounds a bit broad as there are multiple levels of support.

docs/src/pages/docs/comparison.md
boschni: It will only re-render individual components which subscribed to pieces of a changed result. Updates to multiple components are also batched together into one render, not sure if Relay does that?

alloy: I believe it does, but that would need confirmation. My point is more that the footnote isn’t clear enough to me as to what’s being compared here.

docs/src/pages/docs/comparison.md
RichardLindhout: I think Apollo does not check here green either

docs/src/pages/docs/comparison.md
RichardLindhout: Yes that's a good idea but I don't a good for it. Optimized Runtime maybe?

alloy: Perhaps ‘Compile-time optimisations’ ?

RichardLindhout: I think that's perfect

docs/src/pages/docs/comparison.md
RichardLindhout: Based on this documentation Relay checks here green too since it loads the old cache till something newer is loaded. E.g because of the Lagged / "Lazy" Queries<sup>1</sup> part

RichardLindhout:
> Think a lot of these points are somewhat difficult to compare because the "GraphQL" support React Query provides basically relies on the capabilities of the used GraphQL client. A lightweight client like graphql-request can be used, but it's also possible to use a heavier one like Apollo. 

I think in a lot of comparison this is the case but I think react-query could maybe provide a footnote here and document some alternatives as it's very interesting use case for REST api's to have types generated e.g. via openapi.

> Apart from that, think fragments and connections should be supported. The type safety is dependent on the technology used as types can be generated from Swagger, GraphQL schemas or Protocol buffer definitions. \
I've changed these to 
🟡  Supported, but as an unofficial 3rd party or community library/contribution
| GraphQL Fragments                            | 🟡                                    
| GraphQL Subscriptions                        | 🟡

>Offline support sounds a bit broad as there are multiple levels of support.
I think it means the application still has data if the user has visited the app once and if it comes online it van syncs the mutations and/or put/posts.

RichardLindhout:
I'm not sure about the Query Key Change Detection if its unstable or stable serialization?

docs/src/pages/docs/comparison.md
```diff
@@ -8,44 +8,53 @@ title: Comparison
...
+| Polling/Intervals                            | ✅                                     | ✅                         | ✅                                    | ✅                                    |
+| Parallel Queries                             | ✅                                     | ✅                         | ✅                                    | ✅                                    |
+| Dependent Queries                            | ✅                                     | ✅                         | ✅                                    | ✅                                    |
+| Paginated Queries                            | ✅                                     | ✅                         | ✅                                    | ✅                                    |
+| Infinite Queries                             | ✅                                     | ✅                         | ✅                                    | ✅                                    |
+| Lagged / "Lazy" Queries<sup>1</sup>          | ✅                                     | 🛑                         | 🛑                                    | ✅                                    |
```
tannerlinsley, line 30: In the beta version of the comparison, this has been changed to `Lagged Query Data`, which has nothing to do with what the world understands as "lazy" querying. This essentially means that the previous query results can be lagged or kept in the query as new ones are fetched. Imagine paging through some results. Normally, with a new page, you would see a hard loading state since the query and/or variables have changed. This allows you to continue to display the latest successful data while a completely. new query is fetched in the background to replace it.

Can you verify this?

tannerlinsley:
Just so we can reference the new changes in beta: https://github.com/tannerlinsley/react-query/blob/beta/docs/src/pages/comparison.md

docs/src/pages/docs/comparison.md
```diff
@@ -8,44 +8,53 @@ title: Comparison
...
+| Devtools                                     | ✅                                     | 🟡                         | ✅                                    | ✅                                    |
+| Polling/Intervals                            | ✅                                     | ✅                         | ✅                                    | ✅                                    |
+| Parallel Queries                             | ✅                                     | ✅                         | ✅                                    | ✅                                    |
+| Dependent Queries                            | ✅                                     | ✅                         | ✅                                    | ✅                                    |
+| Paginated Queries                            | ✅                                     | ✅                         | ✅                                    | ✅                                    |
+| Infinite Queries                             | ✅                                     | ✅                         | ✅                                    | ✅                                    |
```
tannerlinsley, line 29: Another instance where, in the new beta comparison, we also have "Bi-directional Infinite Queries", which from what I understand is usually *possible* in most libraries, but not officially supported and thus, one I'd like to verify on Relay's end.

docs/src/pages/docs/comparison.md
RichardLindhout: I think Relay does this with the "store-and-network" fetchPolicy.
```
fetchPolicy: Optional. Determines if cached data should be used, and when to send a network request based on the cached data that is currently available in the Relay store (for more details, see our Fetch Policies and Garbage Collection guides):
"store-or-network": (default) will reuse locally cached data and will only send a network request if any data for the query is missing. If the query is fully cached, a network request will not be made.
"store-and-network": will reuse locally cached data and will always send a network request, regardless of whether any data was missing from the local cache or not.
"network-only": will not reuse locally cached data, and will always send a network request to fetch the query, ignoring any data that might be locally cached in Relay.
```

docs/src/pages/docs/comparison.md
tannerlinsley, line 29: Also in the new beta comparison is "Infinite Query Refetching". To explain better, if you have added more data to an infinite list, let's say "next, next, previous, next, previous" or any order/permutation, the ability to refresh the data would need to take into account that exact order and use/update the cursors and/or variables used to fetch each action, then ultimately "replay" those actions to refresh the final window of data.

I have not yet heard of or seen another library that does this out of the box, and while I wouldn't be surprised if Relay supported it, I am doubtful it does currently. Let's verify this one as well.

RichardLindhout: Interesting! I'm not sure, I think thats where GraphQL connections are for, but I only see refetchConnection in Relay as whole not partly for 1 page.

RichardLindhout: I'm not sure what Bi-directional Infinite Queries mean

docs/src/pages/docs/comparison.md
tannerlinsley: That doesn't sound like what I described. A better test question might be:

If you change the query or variables to an *uncached or new* query permutation, can you configure it to continue to display the previous query's data while the new/uncached one fetches?

docs/src/pages/docs/comparison.md
tannerlinsley: Hmmm. Honestly, I think we could definitely use a ❔ icon in the key for things that still need to be verified. I don't want any of these icons to be guessing TBH.

tannerlinsley: Its a term for describing an infinite query that you can cursor/window in both directions. Imagine you start with a window that shows the data for page 6. From there you could scroll-up and load in more results from page 5,4,3,... or scroll down and load in more results from 7,8,9... and so forth. This is bidirectional infinite querying, where you can technically go in both directions from a single root cursor/index of data view.

tannerlinsley: Another good example would be something like a log window. New logs are coming in above you, but only get loaded as you scroll up and old logs are available below you if you scroll down. This would presumably use cursors to "paginate" the data in chunks in either direction you go.

docs/src/pages/docs/comparison.md
tannerlinsley: Let's lead with a ❔ then?

tannerlinsley:
As for the offline support, I would say we need to make it specific. The definition I would give would be that the data must be able to survive a a hard page refresh and upon refresh, be available in the cache without or before a synchronization occurs.

docs/src/pages/docs/comparison.md
sergiodxa: I think it's not possible with another lib, maybe they are waiting for useTransition to be available

RichardLindhout: Only in the experimental version with the new concurrent api's like usePending (??) but there's no icon for that and we're waiting for concurrent mode like forever :-D

docs/src/pages/docs/comparison.md
RichardLindhout: Relay can do this with the paginationContainer or in experimental with the usePaginationFragment but only forward and backward not batches between it as far as I know.

RichardLindhout: https://relay.dev/docs/en/pagination-container#createpaginationcontainer

boschni:
Is there any client code needed for fragment and connection support? Also think the generated types can be marked yellow for React Query and SWR as types can be generated for all protocols.

tannerlinsley:
Where are we on this?

docs/src/pages/docs/comparison.md
RichardLindhout: Relay has great render optimisation but I'm not sure about the batched updates. Maybe @kassens from the Relay team could answer this question.

RichardLindhout:
I'll try to update the feedback next week

MFredrich (approved):
- [ ]

tannerlinsley:
Where are we at on this?

redbar0n:
Would it be a good idea to include version number on the comparisons?
For anyone coming to the site at a point in the future, it would be hard to know if the comparison is stale or not.

redbar0n:
Also, **Relay v.10.1.1** bundle size is: **36.4 kB** (the comparison shows an error currently: `[![null][bp-react-relay]][bpl-react-relay]`). [https://bundlephobia.com/result?p=react-relay@10.1.1](https://bundlephobia.com/result?p=react-relay@10.1.1)

tannerlinsley:
Marking as stale and closing for now, but would really like to see this merged eventually. Comment here if you are interested in finishing this PR :)

PR #1144, opened 2020-10-08
GQL Blog Example

JimmyMcBride:
Using React Query to manage the data coming from a GQL back end to serve content


tannerlinsley:
Like like this is using a git submodule?

JimmyMcBride:
Fixed it! Forgot to remove the .git folder.

PR #1148, opened 2020-10-08
feat: add support for plugins

boschni:
PR to explore support for plugins/middleware. The implementation supports chaining and short circuiting for queries and mutations. Some initial questions:
1. How useful would this be?
2. To also support mutations they had to be moved into the core. Which might be a good thing because the mutation logic can then be re-used in different frameworks. But on the other hand, are mutations really a concept?
3. If mutations are actually a thing, do we then also want to show them in the devtools? And if so, how would they be identified?

### DIFF
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -9,5 +9,12 @@ export { hashQueryKey, isCancelledError, isError } from './utils'
 // Types
 export * from './types'
 export type { CancelledError } from './utils'
-export type { Query } from './query'
 export type { Logger } from './logger'
+export type {
+  OnMutateContext,
+  OnMutatePluginFunction,
+  OnQueryContext,
+  OnQueryPluginFunction,
+  Plugin,
+} from './plugins'
+export type { Query } from './query'
--- a/src/core/mutation.ts
+++ b/src/core/mutation.ts
@@ -1,0 +1,211 @@
+import { getLogger } from './logger'
+import { OnMutateContext, composeOnMutate } from './plugins'
+import { notifyManager } from './notifyManager'
+import { getStatusProps } from './utils'
+import type { MutateOptions, MutationOptions, MutationStatus } from './types'
+import type { MutationObserver } from './mutationObserver'
+import type { QueryClient } from './queryClient'
+
+// TYPES
+
+interface MutationConfig<TData, TError, TVariables, TContext> {
+  options: MutationOptions<TData, TError, TVariables, TContext>
+}
+
+export interface MutationState<TData, TError, TVariables> {
+  data: TData | undefined
+  error: TError | null
+  isError: boolean
+  isIdle: boolean
+  isLoading: boolean
+  isSuccess: boolean
+  status: MutationStatus
+  variables: TVariables | undefined
+}
+
+interface ResetAction {
+  type: 'reset'
+}
+
+interface LoadingAction<TVariables> {
+  type: 'loading'
+  variables: TVariables
+}
+
+interface SuccessAction<TData> {
+  type: 'success'
+  data: TData
+}
+
+interface ErrorAction<TError> {
+  type: 'error'
+  error: TError
+}
+
+type Action<TData, TError, TVariables> =
+  | ErrorAction<TError>
+  | LoadingAction<TVariables>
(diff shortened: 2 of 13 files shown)


zmitry:
Are mutations responsibility of the react query at all? It feels to me that mutation hook is kind of separate from react-query responsibility it's in main package just for convenience. As for query intercepting I find it really useful. But I'd just add single "OnQuery" which allows you to intercept query and add some extra logic like local storage caching or adding extra parameters. As for composition I'd left it in user space but save some kb for library size.

tannerlinsley:
Holy moly. I think this is actually pretty darn cool. What came to mind first was that the devtools still need to be ported to the main `react-query` library and maybe we should try and refactor them to use this plugin system? That would help us dogfood it heavily and even lay the groundwork to show mutation information in the devtools as well. And even further down the road, showing timelines, profiling, etc.

Beyond that, the next thing that came to mind was an offline plugin. It would be pretty easy to throw something together for that using this system. Taking that to the next level would also mean being able to queue up "offline" mutations, which could also use this. 

Either way, I think this could unlock some great features, as long as we take it slow and keep the control relatively limited and scoped to what *we* need it for. Devtools and offline support would be a great and safe start.

tannerlinsley:
Thinking more about mutations. While the general concept is *simple* and easily replicable, the real value I think comes from the integration that we're exploring here. In the vein the offline plugin idea, with mutations being an official part of the core, we can have more control over queueing, suspending, replaying, etc those mutations over time.

boschni:
Not sure if the plugin system would be usable for the devtools because currently it only intercepts query requests. Think the devtools only needs to be able to get the current state and get notified if something has changed?

Including mutations into the core is a separate topic but definitely interesting. Think there is value in showing previous and current mutations in the devtools and also in providing offline support. What do you think of the following draft API?

```js
class Mutation {
  client: QueryClient
  context: TContext
  data: TData
  error: TError
  status: 'idle' | 'loading' | 'success' | 'error'
  failureCount: number
  variables: TVariables
  mutationKey: string | unknown[]
  options: MutationOptions
  constructor(config: MutationConfig) {}
  mutate(variables: TVariables) => Promise<TData>
}

interface MutationConfig {
  client: QueryClient
  options: MutationOptions
}

interface MutationOptions {
  mutationKey: string | unknown[]
  mutationFn: (variables: TVariables) => TData | Promise<TData>
  onMutate: (mutation: Mutation) => void | Promise<void>
  onSuccess: (mutation: Mutation) => void | Promise<void>
  onError: (mutation: Mutation) => void | Promise<void>
  onSettled: (mutation: Mutation) => void | Promise<void>
  retry: boolean | number | ((mutation: Mutation) => boolean)
  retryDelay?: number | ((mutation: Mutation) => number)
}

const promise = client.mutate({
  mutationKey: ['addTodo'],
  mutationFn: variables => todoApi.create(variables),
  onMutate: async ({ client, variables, context }) => {
    await client.cancelQueries('todos')
    const todo = { id: uuid(), ...variables }
    // add optimistic todo
    client.setQueryData('todos', todos => [...todos, todo])
    context.todo = todo
  },
  onSuccess: ({ client, context, data }) => {
    // replace optimistic todo with server todo
    client.setQueryData('todos', todos =>
      todos.map(todo => (todo.id === context.todo.id ? data : todo))
    )
  },
  onError: ({ client, context }) => {
    // remove optimistic todo
    client.setQueryData('todos', todos =>
      todos.filter(todo => todo.id === context.todo.id)
    )
  },
  retry: ({ failureCount, error }) =>
    failureCount < 3 && error.response.status === 500,
  retryDelay: 1000,
})

const mutations = client.findMutations()

mutations.forEach(mutation => {
  console.log(mutation.mutationKey, mutation.variables, mutation.status)
})

client.cancelMutations('addTodo')
```
The `useMutation` hook would create and watch a mutation and call `mutate`. Any subsequent `mutate` calls would then create new mutations?

tannerlinsley:
This looks great!

tannerlinsley:
Probably want to back off on this until v4 eh?

boschni:
Yes it can be added at a later time, this one can be closed

Issue #1158, opened 2020-10-11
initialData has no effect if query is in flight


tannerlinsley:
Prepping for a workshop and wanted to use this paradigm, but it feels broken and we should talk about it:

- https://codesandbox.io/s/agitated-colden-0v0mp?file=/src/index.js
- Hover over item (it will start prefetching, which artificially takes 3 seconds)
- Click on item before prefetching completes
- Post view is shown in the "hard" `isLoading` state.

Expected:
- Click on item before prefetching completes
- Post view is shown using the `initialData` (imagine a preview/partial version of the data)
- Full post is loaded as the original fetch from the prefetch call completes

Thoughts?

PS. This is also the behavior in 2.x

tannerlinsley:
I think this is obviously working exactly how it's designed where if a query is already in flight, it is considered in a loading state from the beginning, which means it won't try and use initialData. I'm wondering how we could model this enhancement though. Maybe something like:

```js
if (isLoading && initialData) {
  // do initial data stuff
}
```

boschni:
What if you move the initialData into the prefetch?

tannerlinsley:
Yeah that’s the current workaround that works. Just a weird location. It would be awesome to be able to just configure the initial data in a single place in my app.

Tanner Linsley

tannerlinsley:
This is where i wonder about just using destructured default data instead of initial data. But then it kinda messes with the whole `status` rendering concept. I would then have to *also* check if `data` exists, even though it's in a hard loading state.... Hmmmm

tannerlinsley:
WHAT IF.... `initialData` didn't set the cache, but it just overrides the observer state/data.

So in pseudo code:

```js
// observer
if (isLoading && initialData) {
  isSuccess = true
  data = initialData()
}
```

tannerlinsley:
I remember there were still reasons to allow `initialData` to set the data in the cache, so maybe this could even be a new option like `placeholderData`. 🤷

sergiodxa:
In SWR, the initial data is local and not saved in the cache, this causes a lot of issues when hydrating from SSR, but with the hydrate feature of RQ this should not be a problem.

boschni:
Placeholder data has been added

PR #1161, opened 2020-10-11
feat: add placeholderData to queryObserver

tannerlinsley:
Trying my hand at a bit of Typescript! I feel so out of my element here... but hopefully not for long. 😉 

This PR is clearly incomplete, but actually works as a solution in my codesandbox from #1158.

Thoughts here?


boschni:
Great to see you are digging into it! I kind of like the `placeholderData` but do have two questions:

1. What are the benefits over using:
```js
const { data = 'initial' } = useQuery()
```

2. The `success` status does not feel correct to me as the query has never been completed. It removes a bit of insight into the actual status of the query. Maybe it should be `loading` while having data?

tannerlinsley:
### 1
- You don't have to destructure to get the ease of setting the default.
- You can hoist the placeholder data into a custom hook or configuration object, which is way more 
- composable/portable than scripting into the default value at every usage
- It does not take care of the added state overrides needed to display the data.

### 2
- Keep in mind, we're not changing the status of the query, just the observers interpretation of the query, much like `isStale`.
- The data you're providing should be roughly compatible with any actual data you're displaying, so `isSuccess` is a pretty safe bet, especially when you consider that the user is opting into the placeholder data. Similar to `initialData` they can and should expect it to be treated like normal data while it's being filled.
- Adding yet another state to the already complicates existing template flows for almost every user. `isIdle` was already difficult enough for most people to handle when we introduced it, and I think this would similarly add a weird step for users where it's unnecessary. eg.

```js
return query.isLoading ? 'Loading' : query.isError ? 'Error' : 'Show Data JSX'

// Becomes

return query.isPlaceholder
  ? 'Show Successful JSX'
  : query.isLoading
  ? 'Loading'
  : query.isError
  ? 'Error'
  : 'Show Successful JSX'
```

Empirically, I can see how adding another state would make technical sense, but where successful and placeholder states are (and should be) almost identical in templating logic, I see more negatives (breaking everyone's templates who uses it) with introducing the new state.

boschni:
Yeah a `placeholderData` option can indeed provide some additional convenience. I'm in favor of adding it. My main doubt is about the `status` property, which does not really seem to indicate the query has executed successfully anymore. It's becoming more of a flag to indicate `data` is not `undefined` instead of reflecting the current query status. It's one of the reasons why I think why value checking is more flexible and correct:
```js
const { data } = useQuery()

if (data) {
  return <div>{data}</div>
}
```
If you also want to know what kind of data it is:
```js
const { data, isPlaceholderData } = useQuery()

if (data) {
  return <div className={isPlaceholderData ? 'loading' : ''}>{data}</div>
}
```

tannerlinsley:
Hmmm. That would create a lot of upheaval from the current documentation, but if it's the right thing to do, we should discuss it thoroughly to make sure we're making the right move. It would be a relatively large change conceptually from what it's been in the past.

tannerlinsley:
I think the whole data checking approach fits this situation pretty well, but it's when we get into showing/persisting errors and data that the approach rubs me the wrong way.

tannerlinsley:
From a maintainer perspective, it costs us very little to just keep things the way they are and add the feature. It doesn't break anyone in v2 or v3 and it's pretty clear what the option is doing in the docs. We can always reformat things in the future, but my guess is that no one would even bat an eyelash at the placeholder/isSuccess combo for quite a long time. 

In contrast, changing everything to be data/error/loading checking as a primary means of templating would be a ton of work  today and probably create a lot of churn for little payoff IMO.

src/core/queryObserver.ts
```diff
@@ -256,6 +262,18 @@ export class QueryObserver<TResult, TError> {
       isPreviousData = true
     }
 
+    if (status === 'loading' && this.config.placeholderData) {
+      const placeholderData =
+        typeof this.config.placeholderData === 'function'
+          ? (this.config.placeholderData as PlaceholderDataFunction<TResult>)()
+          : (this.config.placeholderData as TResult)
```
boschni, line 269: This typecast should not be needed

src/react/tests/useQuery.test.tsx
```diff
@@ -2361,4 +2361,29 @@ describe('useQuery', () => {
...
+        <div>
+          <h2>Data: {first.data}</h2>
+          <div>Status: {first.status}</div>
+        </div>
+      )
+    }
+
+    const rendered = render(<Page />)
+
+    rendered.getByText('Data: placeholder')
```
boschni, line 2383: Maybe use the states checking pattern here for consistency?

boschni:
Yeah I agree let's keep the value/status checking discussion in https://github.com/tannerlinsley/react-query/pull/1108. It should not be blocking this feature. Wouldn't a `isPlaceholderData` flag be handy if you want to show a loader or maybe a bit different design while rendering placeholder data?

tannerlinsley:
Added

docs/src/pages/docs/api.md
```diff
@@ -26,6 +26,7 @@ const {
   enabled,
   initialData,
   initialStale,
+  placeholderData,
```
boschni, line 29: Maybe also add the `isPlaceholderData` to the result?

tannerlinsley: Oh, it should be there now, just forgot to document it.

tannerlinsley:
I'm guessing this will need to be manually added to v3 as well.

jxe:
It's not clear from the docs if setting placeholderData means the hook won't suspend even when isLoading and suspense are both true.

PR #1166, opened 2020-10-13
docs(api): change manual to isEnabled

jpangelle:
## Description

From my understand, the `manual` property was changed to `enabled`. This pull request updates the documentation to reflect that change.


tannerlinsley:
Nice! Could you also do this for the beta branch?

PR #1172, opened 2020-10-14
fix: add descriptive error when missing queryFn

chrisvasz:
The empty `Promise.reject()` call in `config.ts` causes meaningless errors to appear in our bug tracker that look like this:
image

I'm pretty sure we are hitting this error due to [this issue](https://github.com/tannerlinsley/react-query/issues/1088) in react-query v2 (fixed in v3). It took quite a while to figure out what was causing those bug reports though. A meaningful error message could save hours and hours of debugging.

### DIFF
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -54,7 +54,7 @@ export const DEFAULT_CONFIG: ReactQueryConfig = {
     cacheTime: 5 * 60 * 1000,
     enabled: true,
     notifyOnStatusChange: true,
-    queryFn: () => Promise.reject(),
+    queryFn: () => Promise.reject(new Error('react-query: missing queryFn')),
     queryKeySerializerFn: defaultQueryKeySerializerFn,
     refetchOnMount: true,
     refetchOnReconnect: true,


tannerlinsley:
This is great. It would be even better if it only displayed this in development though (gotta keep that bundle size down 😉 ). Think you make that change?

chrisvasz:
Happy to made a change – but what would you recommend it say in production? The production error message needs to be meaningful enough to point to the root cause. Those bug reports with `unknown/undefined` are being generated by the production build right now, so if we only generate the error in dev, our problem still remains.

tannerlinsley:
In production you should be able to stack trace into the error cause without an issue, right?

chrisvasz:
The stack trace leads me to a strange location when I trigger the error in Chrome. I'm taken to the `Console.error` line in this block of code: https://github.com/tannerlinsley/react-query/blob/6f2ae92ee1b1478078cca9ab59af04b5427d4090/src/core/query.ts#L405-L421

The `Promise.reject()` line does not appear in the stack trace. I had to use Chrome's "pause on caught exceptions" option to eventually find it:

image

This has been my experience with error boundaries (and more generally, code that catches and re-throws exceptions in JavaScript). Sometimes re-throwing exceptions appears to corrupt the stack trace – particularly when you add in aggressive transpiling of async functions down to ES5. For this error, our bug tracker doesn't even attempt to show a stack trace even though it does for others.

Setting the stack trace problem aside, a bug report of `undefined` is still not very helpful. Let's say a helpful stack trace pointed to the `Promise.reject()` line. Even then, how would I know what to do? I eventually figured it out by reading the source code, but wouldn't an actionable error message be more helpful?

tannerlinsley:
Yes you’re right. Let’s just keep it as short as possible.

chrisvasz:
Sounds good! Will it work like it is? I'm struggling to think of a way to shorten it any further without dropping meaningful information. Happy to change it if there is a better idea though!

Here's the current proposal: `react-query: missing queryFn`

tannerlinsley:
Sounds good to me.

boschni:
A more descriptive message has been added in v3

PR #1181, opened 2020-10-18
feat: tree shakeable api

boschni:
The bundle size of React Query has been increasing lately as more features are being added. While the size increase is inevitable, it would be nice to only "pay" for what you use.

In the current V3 API, most functionality is tied together with `QueryClient`. This allows for easy discoverability and adds some convenience because everything is in one place. The downside of this approach is that it cannot be tree-shaked.

With the current set of features, this PR reduces the bundle size from 9.6 kB to 6.07 kB (gzipped) when only importing `useQuery`. This is only 1kB more than for example SWR which is 5.1kB.


tannerlinsley:
I love the savings! I love the new environment concept too, but not the name. I really liked client better. I really do not like the new syntax for interacting with the client where you have to import two things. This doesn’t feel like enough of an improvement on the size to justify upheaval of the entire api syntax. The client change was cool because it was just a rename.

tannerlinsley:
Is there any way to keep the client.method syntax but also get the savings?

boschni:
Not binding everything together into a client and providing separate functions is the only way to enable tree shaking and to achieve these file size reductions. It's also pretty similar to the Relay approach: https://relay.dev/docs/en/fetch-query . Personally I think the reductions are quite significant and they will only get larger when more features are being added.

It's also pretty easy to create a client similar to the previous version:
```js
function useClient() {
  const environment = useEnvironment()
  return {
    refetchQueries: refetchQueries.bind(null, environment),
    invalidateQueries: invalidateQueries.bind(null, environment),
  }
}
```

tannerlinsley:
I see. That’s a good point. So would you recommend leading in the docs with the raw approach?

tannerlinsley:
Also, when should we rip out the core to the other repo?

boschni:
Think we should provide one recommended way of working to prevent any confusion. This would either be with a `Client` or with individual functions. The `Client` approach is syntactically a bit nicer, but it does tie everything together and prevents bundlers from tree shaking.

**Client**
```js
const environment = new Environment({
  queryCache: new QueryCache
})

const client = new Client(environment)
client.watchQuery({ queryKey: 'posts', queryFn: fetchPosts })
client.prefetchQuery({ queryKey: 'posts', queryFn: fetchPosts })
client.runMutation({ mutationFn: addPost, variables: { title: 'title' }})
```

**Functions**
```js
const environment = new Environment({
  queryCache: new QueryCache
})

watchQuery(environment, { queryKey: 'posts', queryFn: fetchPosts })
prefetchQuery(environment, { queryKey: 'posts', queryFn: fetchPosts })
runMutation(environment, { mutationFn: addPost, variables: { title: 'title' }})
```
**Comparison**
Just did some additional comparisons to check the rollup bundle size when importing `useQuery`:

Client: 32 kB (8.01 kB gzipped)
Functions: 24 kB (5.9 kB gzipped)

Which approach to take mostly depends on which aspect we think is more important. The functional approach is a bigger change to the current API, but it also allows us to keep and maintain a competitive edge over the other libraries. Maybe it would be good to also involve other people into this discussion?

**Moving the core**
Let's create a separate issue for this and define a plan of approach there :)

boschni:
Another way of splitting things up and still provide a similar api would be to provide specific clients. It is less granular than providing individual functions, but it does allow to split some functionality by domain:
```js
const environment = new Environment({
  queryCache: new QueryCache
})

const client = new QueryClient(environment)
client.fetchQuery()
client.prefetchQuery()
client.cancelQueries()
client.removeQueries()
client.invalidateQueries()
client.refetchQueries()
client.setQueryData()
client.getQueryData()
client.findQuery()
client.findQueries()

const client = new InfiniteQueryClient(environment)
client.fetchInfiniteQuery()
client.prefetchInfiniteQuery()
client.prefetchInfiniteQueryNextPage()
client.prefetchInfiniteQueryPreviousPage()

const client = new MutationClient(environment)
client.runMutation()
client.cancelMutations()
client.findMutations()

const observer = new QueryObserver(environment)
const observer = new InfiniteQueryObserver(environment)
const observer = new QueriesObserver(environment)
const observer = new MutationObserver(environment)

const client = useQueryClient()
```

tannerlinsley:
Hmm. I think you’ve convinced me to go with the functional versions from before.

tannerlinsley:
I got some feedback from some people that I know use React Query extensively (educators, big-co's, streamers, and some random twitter friends that love RQ) and after discussing all of the pros and cons to this approach, these are their collective, but summarized opinions:

- When asked about the current size of v3, it was unanimous that ~8-10 kb is perfectly fine for the features that v3 will bringing to the table
- No one I spoke to *only* uses `useQuery` OR `useMutation` and a vast majority of them use `useInfiniteQuery` or plan to.
- When asked if the 1-3 kb of overhead would ever stop them from using it for just one of those utilities, everyone said no.
- The existing move to `new QueryClient({ cache: new QueryCache() }` was really well received by every person and everyone was pretty excited about it. They also stated that it will be easy to migrate to.
- Every person I showed the functional style to was not excited about the proposed change. Despite explaining that's how Relay currently does their API and that it would allow for tree shaking and smaller bundle sizes, they said the savings didn't justify the API change for them stating that it felt "clunkier" and "too verbose".

A really nice tidbit that I got from one individual:

> The size of React Query is already super small and the purpose is very focused. Worrying about 2 or 3 kb here or there to add or improve the core is a wonderful problem to have. Most libraries in the ecosystem are stuck in much higher bundle size ranges because they try to include every single feature (something I admire that React Query doesn't do), and as for SWR, I think it's pretty clear that they don't include *enough* in their API to make it friendly to use like React Query. It's all about balance and React Query is very well balanced. I trust that if something makes it into the React Query bundle, it's there for good reason.

So after getting this feedback and having a few days to stew on it, my gut is telling me that we should just stick with the current v3 `new QueryClient({ cache: new QueryCache() }` setup.

tannerlinsley:
Another piece of input I got:

> Relay is awesome, so it's great you're looking to it for inspiration(we use it in one of our other producst), but if we're talking about API style and bundle tradeoffs, I would rather use React Query's queryCache any day over Relay's API. I just don't like the amount of imports and composition that I have to do on my own with it. 🤷‍♂️

PR #1213, opened 2020-10-25
feat: add data, error and fetch origins

boschni:
This PR adds some properties to the query to indicate where a fetch, data or error originated from.

This could be useful for:
- Knowing if the data in the cache came from the server or from some local update.
- Not showing loading indicators when refetching because of a window focus for example.
- Not showing errors if they originated from a window focus for example.
- Debugging

It's an idea, would love to to get some feedback on it.

### DIFF
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -33,18 +33,35 @@ interface QueryConfig<TData, TError, TQueryFnData> {
 
 export interface QueryState<TData = unknown, TError = unknown> {
   data: TData | undefined
+  dataOrigin?: QueryDataOrigin
   dataUpdateCount: number
   dataUpdatedAt: number
   error: TError | null
+  errorOrigin?: QueryErrorOrigin
   errorUpdateCount: number
   errorUpdatedAt: number
   fetchFailureCount: number
   fetchMeta: any
+  fetchOrigin?: QueryFetchOrigin
   isFetching: boolean
   isInvalidated: boolean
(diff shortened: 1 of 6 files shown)


tannerlinsley:
I think this is a good addition and solves some open issues I think.

src/core/query.ts
```diff
@@ -33,18 +33,36 @@ interface QueryConfig<TData, TError, TQueryFnData> {
...
+  | 'clientFetch'
+  | 'clientRefetch'
+  | 'observerFetch'
+  | 'observerRefetch'
+  | 'observerRefetchInterval'
+  | 'observerRefetchOnMount'
+  | 'observerRefetchOnReconnect'
+  | 'observerRefetchOnWindowFocus'
+
+export type QueryDataOrigin = QueryFetchOrigin | 'initialData' | 'setData'
```
TkDodo, line 62: quick question: what about `placeholderData`? Wouldn’t that also be a possible origin like `initialData`?

boschni: Not on the query level at least as the placeholder data will never be persisted in the query. Additional origins could be added on the observer level though...

tikotzky: should hydration count as its own origin, or should it just use the origin that was used to load it on the server?

boschni: Think it should hydrate the original origins, otherwise the UI might change between hydrations.

PR #1244, opened 2020-11-02
Guide page for testing - first draft

sazzer:
I can't promise it's the best written documentation page, but it's a starting point for people to read to understand how they can test their code that makes use of React Query :)

Cheers

### DIFF
--- a/docs/src/pages/docs/guides/testing.md
+++ b/docs/src/pages/docs/guides/testing.md
@@ -1,0 +1,88 @@
+---
+id: testing
+title: Testing
+---
+
+React Query works by means of hooks - either the ones we offer or custom ones that wrap around them.
+
+Writing unit tests for these custom hooks can be done by means of the [React Hooks Testing Library](https://react-hooks-testing-library.com/) library.
+
+Install this by running:
+
+```sh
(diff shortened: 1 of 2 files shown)


tannerlinsley:
It looks great! Maybe we could use the isLoading/isSuccess from the query instead of waiting on the nock status?

sazzer:
Hi there,

To be honest, the code I've got here was lifted almost as-is from my own
project, and because I'm using *Suspense* I didn't make use fo the
isLoading/isSuccess stuff.
I'll have a look at re-working it to use that for you tomorrow though.

Cheers

docs/src/pages/docs/guides/testing.md
```diff
@@ -0,0 +1,88 @@
...
+const expectation = nock('http://example.com')
+  .get('/api/data')
+  .reply(200, {
+    answer: 42
+  });
+
+const { result, waitFor } = renderHook(() => useFetchData(), { wrapper });
+
+await waitFor(() => {
+  return expectation.isDone();
```
tannerlinsley, line 82:
```suggestion
  return result.current.isSuccess();
```

tannerlinsley (changes requested):

tannerlinsley (approved):

PR #1247, opened 2020-11-02
feat: add discriminated unions for query results

boschni:
Related PR: https://github.com/tannerlinsley/react-query/pull/1108


tannerlinsley:
Heck yes. This is beautiful @boschni.

TkDodo:
@boschni I played around with this, and it's great. even the `isSuccess` and `isError` booleans now narrow the type 🚀 .

I have two more questions, not sure if I should rather open a new discussion?

I converted the simple example to TS: https://codesandbox.io/s/dark-feather-bvigq

1) switching on the status, I have to include `idle` or add a `default` case somewhere along the lines, even though I think the query can only ever be `idle` if we use the `enabled` prop?
2) There is no easy way to narrow the type of `error`. This has also recently come up on Discord. I'm aware that we cannot know the error because anything can be thrown, so we have to:
  - check for `instanceOf Error` or so to narrow it down (not so nice with the exhaustive matching)
  - provide generics to useQuery: `useQuery<MyData, Error>`, which is okay, but in order to define the type for Error, I have to provide the type for Data as well, and that would actually be nicely inferred by the queryFn.

Not sure what we could do to make 2) a bit more usable? Turning the order of Error and Data around maybe, but I'm not sure if that would make it better :/

FYI, I want to contribute with a TS example or two to the docs because it has come up a lot, so I would like to get the best way right in this regard. Thanks for your input!

tannerlinsley:
We could convert the generic to a mapped generic with defaults, Conceptually, this is `useQuery<{Data?: MyData, Error?: MyError} = defaults>`. It would still be able to use inference for either type and you could only pass it one like `useQuery<{Error: MyError}>`. Thoughts?

TkDodo:
> We could convert the generic to a mapped generic with defaults, Conceptually, this is `useQuery<{Data?: MyData, Error?: MyError} = defaults>`. It would still be able to use inference for either type and you could only pass it one like `useQuery<{Error: MyError}>`. Thoughts?

Ah, I guess this is similar to what you’ve been doing with react-table lately. I think it would be cool, also for `useMutation`, see this discussion for example where TVariables goes back to undefined I think because you to provide all the generics right now, or let them all be inferred: https://github.com/tannerlinsley/react-query/discussions/1307

tannerlinsley:
Yeah when each generic has equal priority, but can all be inferred *sometimes*, a generic map can be really nice.

PR #1261, opened 2020-11-05
feat: add query function context

boschni:
Closes https://github.com/tannerlinsley/react-query/discussions/1246

### DIFF
--- a/docs/src/pages/guides/migrating-to-react-query-3.md
+++ b/docs/src/pages/guides/migrating-to-react-query-3.md
@@ -65,6 +65,32 @@ const queryClient = new QueryClient({
 })
 ```
 
+### Query function parameters
+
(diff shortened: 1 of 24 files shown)


docs/src/pages/guides/migrating-to-react-query-3.md
```diff
@@ -65,6 +65,32 @@ const queryClient = new QueryClient({
 })
 ```
```
tannerlinsley, line 67: We probably need to mention the removal of `queryFnParamsFilter` here as well.

tannerlinsley (changes requested):

docs/src/pages/guides/migrating-to-react-query-3.md
boschni: Good one, added!

PR #1302, opened 2020-11-19
feat(useQuery): Add refetchNever config option to always use cache wh…

eamonn:
This features allows a config option to always use cache to serve data and never hit the API if a cache exists for the given query key.  This is useful when, for example, loading static files with react-query as there is no point in revalidating the cache as it will never change. `refetchOnMount` is not always sufficient as sometimes user activity may unmount a component and when it is remounted the developer may want to avoid revalidating the cache. If the file is several megabytes and we know the cache is good, why bother revalidating it?

I have a use case where I load static json files (so they never change) they have WebGL vertex plots in them and they can get big (several megs) I tried `refetchOnMount` but sometimes the component with the useQuery Hook unmount. I could have an invisible copy of a component with the hook always in the DOM but that is a bit of a hack so I thought this feature may be useful.

**tl;dr this option prevents cache revalidation in all circumstances.**

Looking forward to your feedback.

I'm hoping this feature can merge go in as I don't want to maintain this fork forever 🤣 . Go easy on me I'm fairly new to Open Source. Thanks for the great Lib.

### DIFF
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -57,6 +57,7 @@ export const DEFAULT_CONFIG: ReactQueryConfig = {
     queryFn: () => Promise.reject(),
     queryKeySerializerFn: defaultQueryKeySerializerFn,
     refetchOnMount: true,
+    refetchNever: false,
     refetchOnReconnect: true,
     refetchOnWindowFocus: true,
     retry: 3,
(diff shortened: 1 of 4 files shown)


TkDodo:
quick question: doesn't `staleTime: Infinity` solve most of the problems that this flag would also solve?

eamonn:
@TkDodo oooh you are right, sorry I thought that even when the query was stale the data was still being revalidated on new mounts. I will show myself the door...

TkDodo:
@eamonn no worries. each `refetchOn...` prop always respects staleTime :)

PR #1308, opened 2020-11-21
feat: Add snapshotValue to use mutation success paths

JakeGinnivan:
I raised a discussion a while back, my scenario is around optimistic concurrency and being able to remove the _in progress_ items on success.

https://github.com/tannerlinsley/react-query/discussions/1063

I have chosen to add the pending items outside react query's query cache, meaning invalidating the cache will not help me. Without having access to the snapshot value there is no way to pass state between the use mutation action and the success methods.

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -233,7 +233,7 @@ export interface MutateConfig<
   TVariables = unknown,
   TSnapshot = unknown
 > {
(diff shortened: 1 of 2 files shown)


tannerlinsley:
Document the change, port it to the v3 branch and you’ve got yourself a deal

JakeGinnivan:
Looks like someone has already made this change on the beta branch. 🥳 

Will give that a spin instead.

PR #1309, opened 2020-11-21
Throw error when QueryClientProvider has not been used to set a query client

JakeGinnivan:
After migrating to v3, I had a number of tests fail because I was not using an explicit queryCache before the migration. Another option is to create a default query client, but I prefer having to do this step yourself personally. 

The error I got was `Error: Uncaught [TypeError: Cannot read property 'defaultMutationOptions' of undefined]`

Added a error message so this is now the error my tests give:

image


boschni:
Great thnx!

PR #1311, opened 2020-11-22
Test which reproduces warn when mutate called and component unmounts …

JakeGinnivan:
…immediately

I have not looked into fixing yet, this scenario did not warn in 2.x so I just thought i'd throw up a test.

The test passes, unsure how to make it actually fail without patching console.error and checking to see if it warns.

image


boschni:
Thanks! Fixed with https://github.com/tannerlinsley/react-query/commit/9d66f43d0dc89213112d5aa90076037a9fff6933

PR #1317, opened 2020-11-23
docs: add isFetched docs

yordis:


yordis:
I would like to get some support around documenting `isFetched` variable.

Is `isFetched` true when the request either succeed or failed?

yordis:
@tannerlinsley you may know the answer to my question 🤷🏻

TkDodo:
> Is isFetched true when the request either succeed or failed?

I think it is. I took a quick look at the tests: https://github.com/tannerlinsley/react-query/blob/7535383eba20fcfb0e476dbf199823d1423cd7e1/src/react/tests/useQuery.test.tsx#L103

it seems that `isFetched` is false initially, and then transitions to `true` for successful queries and for erroneous queries alike, but it will stay `false` during retries.

it also looks like it is `true` if we successfully prefetch the query, which seems to be the difference to `isFetchedAfterMount`, which is only true after you have also mounted an observer.

yordis:
@TkDodo I guess the current wording is fine, just that we need to be clear on what `being fetched` means in this context (across this package btw, since that is what is not clear to me until now I guess)

PR #1330, opened 2020-11-26
fix(types): change the order of types so that correct type inference is picked for **Object syntax**

udayvunnam:
> change the order of overridden functions so that correct type inference is picked for **Object syntax**

```

// **A)** Parameter syntax with optional config
export function useQuery<TResult = unknown, TError = unknown>(
  queryKey: QueryKey,
  queryConfig?: QueryConfig<TResult, TError>
): QueryResult<TResult, TError>

// **B)** Object syntax
export function useQuery<TResult = unknown, TError = unknown>(
  config: UseQueryObjectConfig<TResult, TError>
): QueryResult<TResult, TError>
```

I am using a query builder that creates a typed structure of **config** with generics.

when invoked with useQuery(configThatIsAlreadyTyped), it is picking syntax **A** instead of syntax **B** with Object syntax. 

This is because of the order of overridden methods. By changing the order the correct type is picked

### DIFF
--- a/src/react/useInfiniteQuery.ts
+++ b/src/react/useInfiniteQuery.ts
@@ -19,6 +19,11 @@ export interface UseInfiniteQueryObjectConfig<TResult, TError> {
 
 // HOOK
 
+// Object syntax
+export function useInfiniteQuery<TResult = unknown, TError = unknown>(
+  config: UseInfiniteQueryObjectConfig<TResult, TError>
+): InfiniteQueryResult<TResult, TError>
+  
 // Parameter syntax with optional config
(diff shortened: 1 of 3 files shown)


src/react/useInfiniteQuery.ts
```diff
@@ -19,6 +19,11 @@ export interface UseInfiniteQueryObjectConfig<TResult, TError> {
 
 // HOOK
 
+// Object syntax
```
udayvunnam, line 22: @tannerlinsley changed the order of overridden function types, so that correct type inference is picked for Object syntax

PR #1359, opened 2020-12-06
fix: cancellations for useInfiniteQuery

Tommos0:
Currently queries aren't canceled when using useInfiniteQuery.


tannerlinsley:
Excellent!

PR #1366, opened 2020-12-08
(docs): add typescript example for optimistic updates

TkDodo:
per popular demand, I added a typescript example for v3

maybe it's not so good to have the same example twice? But how to get types right for optimistic updates has been asked a lot lately... I can try to come up with something else if you want, or maybe we can replace the javascript one with this one or so...

### DIFF
--- a/examples/optimistic-updates-typescript/pages/index.tsx
+++ b/examples/optimistic-updates-typescript/pages/index.tsx
@@ -23,7 +23,7 @@ export default function App() {
   )
 }
 
-type TodoData = {
+type Todos = {
   items: readonly {
     id: string
     text: string
@@ -31,18 +31,20 @@ type TodoData = {
   ts: number
 }
 
-async function fetchTodos(): Promise<TodoData> {
+async function fetchTodos(): Promise<Todos> {
   const res = await axios.get('/api/data')
   return res.data
 }
 
-function useTodos<TData = TodoData>(
-  options?: UseQueryOptions<TData, AxiosError, TodoData>
+function useTodos<TData = Todos>(
+  options?: UseQueryOptions<TData, AxiosError, Todos>
 ) {
   return useQuery('todos', fetchTodos, options)
 }
 
 function TodoCounter() {
+  // subscribe only to changes in the 'data' prop, which will be the
+  // amount of todos because of the select function
   const counterQuery = useTodos({
     select: data => data.items.length,
     notifyOnChangeProps: ['data'],
@@ -61,34 +63,37 @@ function Example() {
   const { isFetching, ...queryInfo } = useTodos()
 
   const addTodoMutation = useMutation(
-    (newTodo: string) => axios.post('/api/data', { text: newTodo }),
+    newTodo => axios.post('/api/data', { text: newTodo }),
     {
-      // Optimistically update the cache value on mutate, but store
-      // the old value and return it so that it's accessible in case of
-      // an error
-      onMutate: async newTodo => {
+      // When mutate is called:
+      onMutate: async (newTodo: string) => {
         setText('')
+        // Cancel any outgoing refetches (so they don't overwrite our optimistic update)
         await queryClient.cancelQueries('todos')
 
-        const previousValue = queryClient.getQueryData<TodoData>('todos')
+        // Snapshot the previous value
+        const previousTodos = queryClient.getQueryData<Todos>('todos')
 
-        if (previousValue) {
-          queryClient.setQueryData<TodoData>('todos', {
-            ...previousValue,
+        // Optimistically update to the new value
+        if (previousTodos) {
+          queryClient.setQueryData<Todos>('todos', {
+            ...previousTodos,
             items: [
-              ...previousValue.items,
+              ...previousTodos.items,
               { id: Math.random().toString(), text: newTodo },
             ],
           })
         }
 
-        return previousValue
+        return { previousTodos }
       },
-      // On failure, roll back to the previous value
-      onError: (err, variables, previousValue) => {
-        queryClient.setQueryData('todos', previousValue)
+      // If the mutation fails, use the context returned from onMutate to roll back
+      onError: (err, variables, context) => {
+        if (context?.previousTodos) {
+          queryClient.setQueryData<Todos>('todos', context.previousTodos)
+        }
       },
-      // After success or failure, refetch the todos query
+      // Always refetch after error or success:
       onSettled: () => {
         queryClient.invalidateQueries('todos')
       },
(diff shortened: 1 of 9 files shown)


examples/optimistic-updates-typescript/pages/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+          <ul>
+            {queryInfo.data.items.map(todo => (
+              <li key={todo.id}>{todo.text}</li>
+            ))}
+          </ul>
+          {isFetching && <div>Updating in background...</div>}
+        </>
+      )}
+      {queryInfo.isLoading && 'Loading'}
+      {queryInfo.error instanceof Error && queryInfo.error.message}
```
TkDodo, line 116: do we want to advocate this behaviour (instanceOf check at runtime?) Otherwise, I would have to add generics to the `useQuery` call, which I think is not ideal either ....

examples/optimistic-updates-typescript/pages/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+          onChange={event => setText(event.target.value)}
+          value={text}
+        />
+        <button disabled={addTodoMutation.isLoading}>Create</button>
+      </form>
+      <br />
+      {queryInfo.isSuccess && (
+        <>
+          <div>
+            Updated At: {new Date(queryInfo.data.ts).toLocaleTimeString()}
```
TkDodo, line 105: I'm showing off the new type-narrowing here (`data` is defined if we check for `isSuccess`). Should I add a comment for that?

examples/optimistic-updates-typescript/pages/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+            ],
+          })
+        }
+
+        return previousValue
+      },
+      // On failure, roll back to the previous value
+      onError: (err, variables, previousValue) => {
+        queryClient.setQueryData('todos', previousValue)
+      },
```
TkDodo, line 70: @boschni I couldn't get type inference to work properly here. It seems that `TContext` is not inferred by the return value of `onMutate` - it's always `unknown` for me. I know that we can add generics to the `useMutation` call, but it would have to be all 4 of them 😅 .

It works here coincidentally because `unknown` is valid for `setQueryData` if no generic is specified. What I really wanted to do is return a `rollback` function:

 ```
        return () => queryClient.setQueryData('todos', previousValue)
      },
      // On failure, roll back to the previous value
      onError: (err, variables, rollback) => {
        rollback()
      },
```

but this is then not possible ... :(

examples/optimistic-updates-typescript/pages/index.tsx
TkDodo: ah, it actually mentions this in the docs:

https://react-query-beta.tanstack.com/typescript#defining-custom-hooks

so I can change that if you want. Also, maybe we should amend that section a bit? If you use the `select` option, you have to actually define 3 generics.

examples/optimistic-updates-typescript/pages/index.tsx
TkDodo: added a comment: 68d2f81

examples/optimistic-updates-typescript/pages/index.tsx
TkDodo: I made up my mind and converted this to a custom hook with a second usage that uses the `select` option to only display the length of todos in a counter, which only renders if the length of the array changes. Now I need the generics anyways 😅 . I hope its not too complicated now 🙃

examples/optimistic-updates-typescript/pages/index.tsx
boschni: Yeah it seems like TypeScript is not able to infer the context correctly anymore when the context is actually used in any of the onError/onSuccess/onSettled functions. It does infer it correctly when only specifying onMutate. I'll need to figure out why this happens and how this can be fixed

boschni: It only infers correctly if you do one of the following things:
- explicitly type the variables argument in `onMutate`.
- omit the variables argument in `onMutate`.
- explicitly type the context argument in `onError/onSuccess/onSettled`.

This is probably related to https://github.com/microsoft/TypeScript/issues/26242. Not sure what we can do to fix it.

examples/optimistic-updates-typescript/pages/index.tsx
bartzy: @TkDodo What exactly in `react-query` is causing that type-narrowing?

examples/optimistic-updates-typescript/pages/index.tsx
TkDodo: so interestingly, if I define the type of `variables` only on `onMutate`, but not on the `mutationFn`, the variables are still correctly inferred everywhere, including `mutationFn`.

rollback then has the correct type, but it will be potentially undefined:

Screenshot 2020-12-11 at 19 46 15

do you know where the union with undefined is coming from? Because if we can solve that, I think thats a "good enough" approach because we only have to define variables once (`onMutate`). If not, I would just stick with the current approach because it works and people can always annotate their types manually. what do you think?

examples/optimistic-updates-typescript/pages/index.tsx
TkDodo: it was added in this PR by @boschni : https://github.com/tannerlinsley/react-query/pull/1247

the `status` field and their derived boolean flags (isSuccess, isError, ...) now perform a type narrowing of the `data` resp. the `error` field

examples/optimistic-updates-typescript/pages/index.tsx
boschni: In case of `onError` the context could be `undefined` if the `onMutate` function errored. At `onSuccess` it should always be available though (will fix the types for that). Think defining the variables on `onMutate` is a good compromise.

Apart from that, maybe better to use actual data as context because a rollback function cannot be serialized and persisted: https://react-query-beta.tanstack.com/guides/optimistic-updates

TkDodo: Cool, I’ll update to that syntax later today, thanks

TkDodo: done: 855b202f

TkDodo:
@tannerlinsley do you want me to open a new PR to the new master branch, or was there a different reason for closing it 🤔 ?

tannerlinsley:
Oh dang. Didn’t realize it removed the branch lol. Just reopen to master :) so sorry.

TkDodo:
Hehe sure, no worries 👍

PR #1371, opened 2020-12-08
fix(hydration): set fetchMeta to null by default to make it serializable

### LINKED ISSUE #1370: [V3]: SSR with NextJs - undefined cannot be serialized as JSON
Hey, 

I've been trying to update a NextJs repo which uses hydration to V3 and got the following error:

image

Quickly looking at the code and there's indeed a `fetchMeta` state property set by default to `undefined`.

https://github.com/tannerlinsley/react-query/blob/4d01e1fd543d304d2b07defcd4cda55e089db88b/src/core/query.ts#L442

tatchi:
Fixes #1370 

Although it seems to fix the issue, nothing makes sure that this value is not set to `undefined` elsewhere in the code (or prevent to set it in the future) since it's typed as `any`.

What do you think about introducing a [JsonValue](https://github.com/sindresorhus/type-fest/blob/master/source/basic.d.ts#L53) type for values that must be serializable?

### DIFF
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -439,7 +439,7 @@ export class Query<TData = unknown, TError = unknown, TQueryFnData = TData> {
       errorUpdateCount: 0,
       errorUpdatedAt: 0,
       fetchFailureCount: 0,
-      fetchMeta: undefined,
+      fetchMeta: null,
       isFetching: false,
       isInvalidated: false,
       isPaused: false,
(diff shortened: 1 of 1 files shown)


tannerlinsley:
Is there a Typescript serializable primitive type?

tatchi:
> Is there a Typescript serializable primitive type?

I don't think so unfortunately. See [this issue](https://github.com/microsoft/TypeScript/issues/1897)

PR #1375, opened 2020-12-08
feat: reset query utils

tannerlinsley:
- [x] Documentation
- [x] Tests


src/core/queryClient.ts
```diff
@@ -132,6 +132,18 @@ export class QueryClient {
...
   }
 
+  resetQueries(filters?: QueryFilters): void
+  resetQueries(queryKey?: QueryKey, filters?: QueryFilters): void
+  resetQueries(arg1?: QueryKey | QueryFilters, arg2?: QueryFilters): void {
+    const [filters] = parseFilterArgs(arg1, arg2)
+    const queryCache = this.queryCache
+    notifyManager.batch(() => {
+      queryCache.findAll(filters).forEach(query => {
+        queryCache.reset(query)
```
boschni, line 142: This can also be `query.reset()`

src/core/queryCache.ts
```diff
@@ -80,6 +80,15 @@ export class QueryCache extends Subscribable<QueryCacheListener> {
     }
   }
 
+  reset(query: Query<any, any>): void {
+    const queryInMap = this.queriesMap[query.queryHash]
+
+    if (queryInMap) {
+      query.reset()
+      this.notify(query)
```
boschni, line 88: The query cache will already be notified when calling `query.reset()`. But maybe we don't even need these methods on the query cache?

aaronjensen:
@tannerlinsley thank you for your work on this. I'll see if I can help out a bit.

aaronjensen:
@tannerlinsley I added a few tests and some documentation. Let me know if you'd like me to make any tweaks. I'm also testing it out here:

https://codesandbox.io/s/react-query-3-refetch-on-error-ux360?file=/src/App.js

One thing that seems odd about it is that it does not automatically trigger a refetch. This is especially odd when using suspense because you expect the data to be present unless you set enabled: false. 

Should it refetch automatically if enabled is not false?

PR #1396, opened 2020-12-13
chore(Docs): Add script to install with Yarn

EfstathiadisD:
The Docs, only had instructions that contained `npm` script. Yarn is now added as an alternative.


tannerlinsley:
Unnecessary now :)

PR #1397, opened 2020-12-13
feat: resetQueries refetches active queries

aaronjensen:
This is a follow up to #1375 that, I think, makes `resetQueries` more useful and less surprising.

### DIFF
--- a/src/core/queryClient.ts
+++ b/src/core/queryClient.ts
@@ -20,6 +20,7 @@ import type {
   QueryObserverOptions,
   QueryOptions,
   RefetchOptions,
+  ResetOptions,
 } from './types'
 import type { QueryState, SetDataOptions } from './query'
 import { QueryCache } from './queryCache'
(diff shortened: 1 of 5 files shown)


src/core/queryClient.ts
```diff
@@ -132,15 +133,22 @@ export class QueryClient {
...
+    const refetchFilters: QueryFilters = {
+      ...filters,
+      active: true,
+    }
+
+    return notifyManager.batch(() => {
       queryCache.findAll(filters).forEach(query => {
         query.reset()
       })
+      return this.refetchQueries(refetchFilters, options)
```
aaronjensen, line 151: Would it be better to have `query.reset()` trigger the fetch if the query active? That seems like better encapsulation, but it seems like the client coordinating things is more typical than the query doing it in this project, so far.

tannerlinsley:
Excellent :)

PR #1416, opened 2020-12-14
docs: fix example in defaultQueryFn guide page

TimoVink:
Unfortunately the example on the `defaultQueryFn` guide page doesn't appear to work as-written for `v3.2.0`, but the example in `examples/default-query-function` _does_ work. 

This PR aims to fix this minor inconsistency in the docs.

Thanks for your work on this library -- it's extremely useful!


tannerlinsley:
Nice catch!

PR #1424, opened 2020-12-15
feat: list of queries and mutations to be prevented from being persisted

aminerol:
should be able to pass an array of QueryKeys and MutationKeys to be prevented from persisting, a use case that this should be handy its when we querying an access token, this token shouldn't be stored locally since it's not encrypted

### DIFF
--- a/src/persist-localstorage-experimental/index.ts
+++ b/src/persist-localstorage-experimental/index.ts
@@ -1,5 +1,6 @@
 import { QueryClient } from '../core'
 import { dehydrate, hydrate } from '../hydration'
+import deepEqual from "deep-equal";
 
 interface LocalStorageCache {
   timestamp: number
@@ -20,6 +21,12 @@ interface Options {
   /** A unique string that can be used to forcefully
    * invalidate existing caches if they do not share the same buster string */
   buster?: string
+  /** A list of QueryKeys to be blacklisted and prevented from persisted
+   * Example: you dont wanna persist an access token here especially we persisting without
+   * any sort of encryption */
+  blacklistQueries?: QueryKey[];
+  /** A list of MutationKeys to be blacklisted and prevented from persisted */
+  blacklistMutations?: MutationKey[];
 }
 
 export function persistWithLocalStorage(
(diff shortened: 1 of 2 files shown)


src/persist-localstorage-experimental/index.ts
```diff
@@ -29,15 +36,32 @@ export function persistWithLocalStorage(
...
   }: Options = {}
 ) {
   if (typeof window !== 'undefined') {
     // Subscribe to changes
     const saveCache = throttle(() => {
       const storageCache: LocalStorageCache = {
+        const cacheState = dehydrate(queryClient, {
+            shouldDehydrateQuery: (query) => {
+                const isExist =
+                    blacklistQueries.filter((bl) => deepEqual(bl, query.queryKey))
```
TkDodo, line 50: I’m pretty sure there are Util functions already to compare query keys without the need of an extra dependency :)

tannerlinsley: Indeed. We should use those utils

src/persist-localstorage-experimental/index.ts
```diff
@@ -20,6 +21,12 @@ interface Options {
   /** A unique string that can be used to forcefully
    * invalidate existing caches if they do not share the same buster string */
   buster?: string
+  /** A list of QueryKeys to be blacklisted and prevented from persisted
+   * Example: you dont wanna persist an access token here especially we persisting without
+   * any sort of encryption */
+  blacklistQueries?: QueryKey[];
+  /** A list of MutationKeys to be blacklisted and prevented from persisted */
+  blacklistMutations?: MutationKey[];
```
tannerlinsley, line 29: We should also be avoiding charged terminology like blacklist. A good alternative might be disallow Queries, filter Queries etc.

TkDodo: I think `blocklist` / `allowlist` ist the new `blacklist` / `whitelist`

tannerlinsley:
This PR is approaching stale status. Would you like us to keep this open @aminerol?

charlesrollin:
Hi all, I don't know what the status of this PR is, but wouldn't it make more sense to enrich the `options` of `persistWithLocalStorage` by adding the 2 following keys :
```ts
{
  hydrate: HydrateOptions,
  dehydrate: DehydrateOptions,
}
```
and using them when calling `hydrate` and `dehydrate` respectively.

This would allow full customization for the user, and could meet the needs of @aminerol if the utility functions @TkDodo  talks about in a thread are exposed by the package.

ericbiewener:
Throwing in a request to add an `allowList` option in this PR as well as the `blockList`.

tannerlinsley:
We're going to be changing the API of the persist plugin soon that will make these changes obsolete. Thanks for your input!

jspizziri:
FWIW, it looks like the option suggested by @charlesrollin was added in PR https://github.com/tannerlinsley/react-query/pull/2131

PR #1470, opened 2020-12-17
fix: infer result data type from query fn return type instead of the reverse

boschni:
Fix for https://github.com/tannerlinsley/react-query/issues/1445 .

This PR swaps the query result type and query function result type inference. The first generic will now define the query function result type, and then the default query result type is inferred from this instead of the other way around.

Might be good to do a thorough review on this one as it is quite a big change.


tannerlinsley:
After trying this out, it seems to work without a hitch. I'm good to merge if you are!

tannerlinsley (approved):

PR #1476, opened 2020-12-18
fix: make sure typed query fn contexts are accepted

boschni:
Fixes https://github.com/tannerlinsley/react-query/issues/1462


jon301:
Thank you guys ;) ❤️

PR #1489, opened 2020-12-19
fix: make sure queries are retried after an error boundary reset

boschni:
This PR contains two fixes:
- Always set query in `loading` state when fetching and no data is cached.
- Make sure components suspend and error when needed.

Fixes https://github.com/tannerlinsley/react-query/issues/1471


alexandernanberg:
This would fix #1253 afaict

PR #1493, opened 2020-12-19
feat: options.initialDataUpdatedAt

tannerlinsley:
Adds an `initialDataUpdatedAt` option to the query options. This allows initialData to be "tagged" with an updatedAt timestamp. Previously, it was possible for old initialData to be considered fresh if a `staleTime` was also provided to the query. Now, you can easily use both `initialData` and `staleTime` together without this edge case.

- [x] Docs
- [x] Tests


boschni:
Nice! Think this is a good addition to `initialData`

PR #1507, opened 2020-12-22
Changing the new invalidate button color to have more contrast

Mzaien:
Hello it is my pleasure to help in enhancing React-Query,
the button it turned to be more of  an orange but it have almost double contrast ratio

### DIFF
--- a/src/devtools/devtools.tsx
+++ b/src/devtools/devtools.tsx
@@ -85,7 +85,7 @@ const theme = {
   success: '#00ab52',
   danger: '#ff0085',
   active: '#006bff',
-  warning: '#ff6200',
+  warning: '#ffb200',
 }
 
 export function ReactQueryDevtools({
@@ -658,6 +658,7 @@ export const ReactQueryDevtoolsPanel = React.forwardRef(
                   onClick={() => queryClient.invalidateQueries(activeQuery)}
                   style={{
                     background: theme.warning,
+                    color: theme.inputTextColor,
                   }}
                 >
                   Invalidate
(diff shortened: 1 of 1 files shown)


src/devtools/devtools.tsx
```diff
@@ -85,7 +85,7 @@ const theme = {
   success: '#00ab52',
   danger: '#ff0085',
   active: '#006bff',
-  warning: '#ffb200',
+  warning: '#ff6200',
```
tannerlinsley, line 88: This works! However, I think we could probably just make the text black for that button like the rest of the tools.

tannerlinsley (changes requested):

src/devtools/devtools.tsx
Mzaien: For All other buttons? or just for the new one,
because it is kinda off having this button with different color (inconsistent)

tannerlinsley: Anything with the yellow background should have black text I think.

Mzaien: Got it 👍🏻, What do you thing if it was moved to this place to make it more consistent 👇 
image

tannerlinsley: The order before was from most useful to least. What is this order inspired from?

Mzaien: White text buttons then the black one 
if it makes bad DX it shall remain as is

Mzaien: Sorry for making this PR as long as this, i wanted to make a good one 😅

tannerlinsley: I personally like the order as is. And no worries on the PR! These are the interactions that bring us all closer :)

PR #1539, opened 2020-12-29
docs: fix initialDataUpdatedAt timestamp information

VanTanev:
The documentation of `initialDataUpdatedAt` added in #1493 erroneously specifies the option as taking a unix timestamp, while it actually takes a JS timestamp.


tannerlinsley:
Good catch

PR #1556, opened 2021-01-02
fix: :bug: Make sure queries which have dataUpdatedAt of 0 will be run

meesvandongen:
When initialDataUpdatedAt is a falsy numeric value (e.g. 0), it will cause it to be replaced with Date.now(). This leads to unexpected behaviour, as one would expect the query to be invalidated instantly, if staleTime is set to an amount less than the result of Date.now(). This change will check if the value is numeric, and if so, always use this numeric value.

BREAKING: This changes the behaviour of the plugin for those users that have specified initialDataUpdated at <= 0.

### DIFF
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -475,7 +475,11 @@ export class Query<
     return {
       data,
       dataUpdateCount: 0,
-      dataUpdatedAt: hasData ? initialDataUpdatedAt || Date.now() : 0,
+      dataUpdatedAt: !hasData
+        ? 0
+        : typeof initialDataUpdatedAt !== 'number'
+        ? Date.now()
+        : initialDataUpdatedAt,
       error: null,
       errorUpdateCount: 0,
       errorUpdatedAt: 0,
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -464,22 +464,18 @@ export class Query<
 
     const hasInitialData = typeof options.initialData !== 'undefined'
 
-    const initialDataUpdatedAt =
-      hasInitialData &&
-      (typeof options.initialDataUpdatedAt === 'function'
+    const initialDataUpdatedAt = hasInitialData
+      ? typeof options.initialDataUpdatedAt === 'function'
         ? (options.initialDataUpdatedAt as () => number | undefined)()
-        : options.initialDataUpdatedAt)
+        : options.initialDataUpdatedAt
+      : 0
 
     const hasData = typeof data !== 'undefined'
 
     return {
       data,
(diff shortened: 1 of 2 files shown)


src/core/query.ts
```diff
@@ -475,7 +475,11 @@ export class Query<
     return {
       data,
       dataUpdateCount: 0,
-      dataUpdatedAt: hasData ? initialDataUpdatedAt || Date.now() : 0,
```
TkDodo, line 478: Wouldn’t using nullish coalesece solve the same purpose?

```diff
- hasData ? initialDataUpdatedAt || Date.now()
+ hasData ? initialDataUpdatedAt ?? Date.now()
```

meesvandongen: It would perhaps be the same as the current code on master, but not as the numeric type check.

TkDodo: that's true, but what else should it be? The types clearly state that it should be `number | undefined`, and the docs say that it has to be _a JS timestamp in milliseconds_.

I don't think that we need to check every option for every possible input - that's what TypeScript is for. Yes, there are also (lots of) JS consumers, but if I put in a string (`"1608412420052"`) - would it not be weird if that were converted to `Date.now()` ?

meesvandongen:
The problem is that a valid timestamp, 0 (number, 1970) is falsy, and
therefore gets converted to date now, the current timestamp (2020)

Op za 2 jan. 2021 18:33 schreef Dominik Dorfmeister <
notifications@github.com>:

TkDodo:
> The problem is that a valid timestamp, 0 (number, 1970) is falsy, and
therefore gets converted to date now, the current timestamp (2020)

Yes, I think I got the problem, and using `??` instead of `||` would solve it with minimal changes. This is one of the prime reasons why nullish coalesce was introduced - so that you can easily get "fallback values" without trapping into the js falsy-ness of the number 0 / empty string. It will only convert `undefined` or `null` to `new Date()` and nothing else, which I think was the intention of the original code.

meesvandongen:
Ah sorry, I misunderstood nullish coalescing, and therefore your comment. I see now, it makes a lot of sense. I will change the PR accordingly.

meesvandongen:
Alright, I updated the PR, was thinking whether I should have `0` or `Date.now()` on line 471, but ultimately this value would never be used anyway in the current implementation and is only used to satisfy typescript.

boschni:
Nice! Could you add a test to cover this bug?

meesvandongen:
Test added

Vasujkes (approved):

PR #1578, opened 2021-01-05
Feature/use tracked query

TkDodo:
I made a first POC of what we discussed here: #1554

I tried it out in the simple example and it worked fine - just looks too easy to be true. Am I missing something obvious? Apart from:

- [x] tests (no idea how to test the proxy tracking, ideas welcome
- [x] documentation

just wanted to ask upfront if a new hook is wanted / the way to go before I put any more effort into this :)


boschni:
Nice! Couple of thoughts:
- Think we don't necessarily need a separate hook for this, maybe a query option to disable/enable this behavior like `notifyOnChangeTracked: false`.
- The proxy could be replaced with `Object.defineProperty`? In the descriptor you can define a `get` function which can be used to track if a property is accessed.

TkDodo:
> Think we don't necessarily need a separate hook for this, maybe a query option to disable/enable this behavior like notifyOnChangeTracked: false.

yeah, I’ve been thinking that a separate hook might be suboptimal, as you can’t combine it easily with `useInfiniteQuery`. The option might make more sense, and you can always set it globally to enable it everywhere. Let me see how that looks :)

> The proxy could be replaced with Object.defineProperty?

Hmm i honestly don’t know. I thought proxies are the idiomatic way to do “interceptors”. Do you know of any example code with `defineProperty` I could take a look at?

dai-shi:
I suppose you can use `defineProperty` to define object getters. It should work because there are only predefined set of properties.

For example:
```js
const target = { isFetching: true, data: {} }
const trackingObject = {
  get isFething() {
    addNotifyOnChangeProps('isFetching')
    return target.isFetching
  },
  get data() {
    addNotifyOnChangeProps('data')
    return target.data
  },
}
```

`defineProperty` is a way do this dynamically.

boschni:
Example with `Object.defineProperty`:
```js
const target = { isFetching: true, data: {} }
const result = {}

Object.keys(target).forEach(key => {
  Object.defineProperty(result, key, {
    configurable: false,
    enumerable: true,
    get: () => {
      addNotifyOnChangeProps(key)
      return target[key]
    }
  });
});
```

TkDodo:
@boschni I changed the implementation from a new hook to a new option, and it also uses `Object.defineProperty` - works like before! Have a look please: 761a1a76

TkDodo:
I did some documentation, is that enough or do we want:
- a separate docs page?
- an example?

src/core/queryObserver.ts
```diff
@@ -148,6 +148,37 @@ export class QueryObserver<
     this.currentQuery.removeObserver(this)
   }
 
+  createQueryResult(
```
boschni, line 151: What if `getCurrentResult` would either return a regular or a tracked version? Then it would be a bit more transparent

src/core/queryObserver.ts
```diff
@@ -148,6 +148,37 @@ export class QueryObserver<
...
+    queryObserverResult: QueryObserverResult<TData, TError>
+  ): QueryObserverResult<TData, TError> {
+    if (this.options.notifyOnChangeTracked) {
+      const trackedResult = {}
```
boschni, line 155: Think we should store the tracked result object on the instance and update it only when the result object changes. This will make sure the reference also stays the same between renders for the tracked version

src/core/queryObserver.ts
```diff
@@ -148,6 +148,37 @@ export class QueryObserver<
...
+      const addNotifyOnChangeProps = (prop: keyof QueryObserverResult) => {
+        if (!this.options.notifyOnChangeProps) {
+          this.options.notifyOnChangeProps = []
+        }
+
+        if (!this.options.notifyOnChangeProps.includes(prop)) {
+          this.options.notifyOnChangeProps.push(prop)
```
boschni, line 162: Not sure if it is a good idea to mutate the `notifyOnChangeProps` option. What about storing the accessed properties in a separate array and then merge it with `notifyOnChangeProps` in `shouldNotifyListeners`?

docs/src/pages/reference/useQuery.md
```diff
@@ -117,6 +118,9 @@ const result = useQuery({
   - Optional
   - If set, the component will not re-render if any of the listed properties change.
   - If set to `['isStale']` for example, the component will not re-render when the `isStale` property changes.
+- `notifyOnChangeTracked`
```
boschni, line 121: Maybe `notifyOnChangeTrackedProps` would be a bit more correct. And should the option be enabled by default? It could prevent a lot of re-renders out of the box.

src/react/tests/useQuery.test.tsx
```diff
@@ -737,6 +737,41 @@ describe('useQuery', () => {
     expect(states[1].dataUpdatedAt).not.toBe(states[2].dataUpdatedAt)
   })
 
+  it('should track properties and only re-render when a tracked property changes', async () => {
```
boschni, line 740: Might be good to add some additional tests to see if the referential integrity stays the same and to see how it interacts with the other `notifyOnChange*` props. And we also need to define and test the behavior: will the hook remember each accessed property during its lifespan? Or will it reset on each render?

src/core/queryObserver.ts
TkDodo: yes, I think this was my first attempt, but we then access the result of that method (`currentResult`) internally, which would lead to some tracking that we don't want, e.g. here:

or here:

https://github.com/tannerlinsley/react-query/blob/361935a12cec6f36d0bd6ba12e84136c405047c5/src/react/useBaseQuery.ts#L83

https://github.com/tannerlinsley/react-query/blob/361935a12cec6f36d0bd6ba12e84136c405047c5/src/react/useBaseQuery.ts#L89

we really only want to track what the consumer of _useQuery_ uses, so I thought the only safe way is to create the tracked version right at the return of `useBaseQuery`

docs/src/pages/reference/useQuery.md
TkDodo: So, I've been thinking: I don't really like having 3 fields that tackle the same area, especially if we enable tracking per default. One possibility would be to _merge_ this new feature with `notifyOnChangeProps`:

```
notifyOnChangeProps?: Array<keyof InfiniteQueryObserverResult> | 'tracked'
```

and have it default to `'tracked'` (or a different, better name 😅 ). That way, everyone who has now already optimized manually via e.g. `notifyOnChangeProps: ['data', 'error']` would keep the same behaviour, but per default, we would track the changes, and we wouldn't need to add a new option field ....

src/react/tests/useQuery.test.tsx
TkDodo: > add some additional tests

will definitely do that.

> will the hook remember each accessed property during its lifespan?

I think it needs to reset and start new tracking with every render. it might be minor, but consider the following example:

```
const queryInfo = useQuery(...)

if (queryInfo.isLoading) return "Loading...";

if (queryInfo.isError)
  return "An error has occurred: " + queryInfo.error.message + queryInfo.dataUpdatedAt;

return <div>query.data ...</div>

```

in the first render, we would only track `isLoading` because of the early return. if we then get an error, we would track `isLoading, isError, error, dataUpdatedAt`. and if we then get data on the next refetch, we would want to track `isLoading, isError, data`, but if we don't reset, we would also keep tracking `error, dataUpdatedAt` properties, which doesn't seem right since we are not currently observing them

src/core/queryObserver.ts
TkDodo: did that here: ba1bb199

I think code will be a bit easier if we decide to go with this proposal though: https://github.com/tannerlinsley/react-query/pull/1578#discussion_r554466089

src/core/queryObserver.ts
TkDodo: yes, good catch. did that here: 978fb758

i had to adapt the test because capturing the state now led to capturing the same instance, and the get interceptor would return the same data for both captured states. I hope this is just a problem in the way we write these tests...

/edit oof, I think this might lead to stale closure issues when accessing queryInfo in an async way:

consider:

```
useEffect(() => {
  console.log(queryInfo.data)
});
```

this logs `undefined` (while we are loading), then the `data` object, but this:

```
useEffect(() => {
  setTimeout(() => console.log(queryInfo.data), 500);
});
```

will log the data object twice, because at the time we invoke the getter, we have already updated the single instance...

is the referential equality important for the returned queryInfo?

TkDodo: aah, I think a better approach might be to create the `trackedResult` new whenever we also create a new `currentResult`, which would be here:

https://github.com/tannerlinsley/react-query/blob/b0c2f6795292d1580e112794d18b27adff77bf48/src/core/queryObserver.ts#L479-L481

That should get rid of the aforementioned issue and also make the code a bit more transparent - writing to currentResult / trackedResult at the same place only. I'll try to do that tomorrow

TkDodo: I think I got it: 4ea4eaa3

I'll add tests for referential equality, too.

src/react/tests/useQuery.test.tsx
TkDodo: > Might be good to add some additional tests to see if the referential integrity stays the same

fe0b4dc7

tannerlinsley:
Where are we sitting with this awesome feature?

TkDodo:
> Where are we sitting with this awesome feature?

basically this discussion on api design needs to be decided: https://github.com/tannerlinsley/react-query/pull/1578#discussion_r554466089

- to we want to turn tracking on per default?
- how should it interact with other `notifyOnChange*` props.

what would be your opinion / preference?

once this is decided I can add more tests for that interaction and then we are good to go I believe :)

tannerlinsley:
I think it would be awesome to have it enabled by default, as long as it’s backwards compatible for existing tests and use cases.  Curious, does it still work with destructuring?

TkDodo:
> as long as it’s backwards compatible 

I have found one edge case where this might lead to stale data: You need to access a field _during render_ to subscribe to it. Accessing it in an effect only is also fine, but if you subscribe to one field, but access a different field during a callback only, you might not see the latest data. Example:

```
  const queryInfo = useQuery("repoData", () =>
    fetch(
      "https://api.github.com/repos/tannerlinsley/react-query"
    ).then((res) => res.json())
  );

return (
    <div>
      {queryInfo.data?.name}
      <button
        onClick={() => {
          alert(queryInfo.dataUpdatedAt);
        }}
      >
        When did I update?
      </button>
    </div>
  );
```

Here, we only subscribe to `data`, so the component will only re-render when `data` changes. If we re-focus the window and get the same data from the request, we won't re-render due to structural sharing. Awesome. But internally, `dataUpdatedAt` was updated. If I now click the button, I will see the old `dataUpdatedAt`, because I didn't access that property during render, so I'm not subscribed to it. Hope that makes sense.

> Curious, does it still work with destructuring?

yes, it does. destructuring does count as property access. And actually, destructuring would solve the above problem, because you would subscribe to both:

```
const { data, dataUpdatedAt } = useQuery(...)
```

---

Regarding API design: I don't think we should do an extra flag, but rather combine it with `notifyOnChangeProps`, see my proposal above.

- `notifyOnChangeProps`: undefined --> re-render always
- `notifyOnChangeProps: ['data', 'error']` --> re-render if data or error changes
- `notifyOnChangeProps: 'tracked'` --> re-render if a tracked property changes 

Because of the edge case (and maybe there I more I haven't considered), I would maybe keep defaulting it to undefined and let people opt-in (easy with a default config), and maybe we can set it to `'tracked'` per default in the next major version. How does that sound?

src/core/queryObserver.ts
```diff
@@ -208,6 +215,10 @@ export class QueryObserver<
     return this.currentResult
   }
 
+  getTrackedResult(): QueryObserverResult<TData, TError> {
```
boschni, line 218: Maybe name these `getTrackedCurrentResult` as it is the current result but then tracked?

src/core/queryObserver.ts
```diff
@@ -478,6 +500,23 @@ export class QueryObserver<
     // Only update if something has changed
     if (!shallowEqualObjects(result, this.currentResult)) {
       this.currentResult = result
+      const addTrackedProps = (prop: keyof QueryObserverResult) => {
+        if (!this.trackedProps.includes(prop)) {
+          this.trackedProps.push(prop)
+        }
+      }
+      this.trackedResult = {} as QueryObserverResult<TData, TError>
```
boschni, line 508: Can we somehow only construct this object when the option is enabled? The enumerations can add up if there are a lot of observers on a page.

docs/src/pages/reference/useQuery.md
boschni: Sounds good, it does discard the option to specify some properties manually in addition to the tracked properties, but I'm not sure if people would needs this.

src/react/useBaseQuery.ts
```diff
@@ -91,5 +91,9 @@ export function useBaseQuery<TQueryFnData, TError, TData, TQueryData>(
     }
   }
 
-  return currentResult
+  observer.resetTrackedProps()
```
boschni, line 94: While I do like the dynamic tracking, would it not "forget" tracked properties in `useEffect` for example?
```js
const result = useQuery()

useEffect(() => {
  if (result.error) {
    // do something
  }
}, [result])

return <>{result.data}</>
```

docs/src/pages/reference/useQuery.md
TkDodo: I also don't think that's a use-case, so I'll go ahead with the proposal 👍

src/react/useBaseQuery.ts
TkDodo: You're right, this  is similar to the callback case I was showing above. You'd need to change the dependencies to include `error`:

```diff
useEffect(() => {
  if (result.error) {
    // do something
  }
-}, [result])
+}, [result.error])
```

this is also what the linter suggests when you add an empty array at the let the auto-fix add it.

Do you think it's worth _not_ resetting because of this?

boschni: Think it's at least a good reason to not enable it by default as this could break applications. Apart from that: not resetting would probably be the safest option because if we would reset, bugs like this could easily sneak in. Maybe the behavior should be configurable with proper documentation so people know the implications?

TkDodo: Yes, I agree on not enabling by default. The callback case is something we can't solve with this I believe. I'm happy to change it to not resetting for now, it's the "safer" case because it means an unnecessary re-render at worst rather than a bug :)

lucalves (approved):

src/core/queryObserver.ts
TkDodo: yup, did that here: c653de1f394f16d0ba23bbfc66312a6b1fefb109

are there any drawbacks, like if we switch from `'tracked'` to not-tracked? I don't think so, and also doesn't sound like a use case :)

src/core/queryObserver.ts
TkDodo: see: 6a93c803

src/react/tests/useQuery.test.tsx
TkDodo: here is another test for a combination of tracking props and excluding props: exclusion takes precedence, which I think is good: 5d923ce

so even is you track props and use one, if you exclude it explicitly, it will not re-render.

docs/src/pages/reference/useQuery.md
TkDodo: here it is: ad540e2d

src/react/useBaseQuery.ts
TkDodo: auto resetting removed here: 14940909

TkDodo:
@boschni could you have another look please, i think i addressed all points 🙂

boschni:
> @boschni could you have another look please, i think i addressed all points 🙂

lgtm 👍

TkDodo:
> > @boschni could you have another look please, i think i addressed all points 🙂
> 
> lgtm 👍

cool! @tannerlinsley if you're also happy I think we could merge this. Would you want more docs or examples?

faradaytrs:
@tannerlinsley why not having function to decide whether component should update? I would be similar to shouldComponentUpdate,
we could provide all the information to this function and depending on the result (true,false) rerender the component.

TkDodo:
do you have a specific use case in mind where you would want to use such a function?

faradaytrs:
> do you have a specific use case in mind where you would want to use such a function?

I can't say for a now, but maybe for example if I have a request that often fails, I make few requests before I get one successful and I kinda don't need to rerender component until I got a successful result. I know this example is quite shallow, but I think such functionality is very useful anyway. It would pair well with select, sometimes you get the data you don't need to use and it's a good idea to decide whether to re-render the component. Of course, now you can, for example, wrap request function with another one that would throw an error on an undesirable result and configure query not to rerender on error, but the simple function would be much cleaner.

TkDodo:
> and I kinda don't need to rerender component until I got a successful result.

so if you use `notifyOnChangeProps: 'tracked'`, and you only use `data` in your component, but not error or any loading flag, you will only re-render if `data` changes. If you start do observe errors as well, they will be "implicitly added" to notifyOnChangeProps. Unless there is reason that you are reading a value from the queryResult in you render function, but you don't want to render that component if that value changes, thus choosing to deliberately display outdated data, this should cover most bases.

what you cannot do at the moment is subscribing to parts of data with this. say you only access `data.foo`, and `data.bar` changes - your component will re-render with this approach.

that's where `select` comes in. with `select`, you can transform the data to what you need in your component, and only subscribe to those changes. So if you do: `select: data => data.foo`, you have just subscribed to `foo` only, and your component will no longer re-render if `bar` changes.

together, those two props (select and tracked queries) give you a very high amount of render optimization.

the problem with a "manual" function is the same as with `shouldComponentUpdate`. It is quite error prone to get right, leaving out things that you are actually using, but forget to add to that function will lead to stale data etc etc.

faradaytrs:
> > and I kinda don't need to rerender component until I got a successful result.
> 
> so if you use `notifyOnChangeProps: 'tracked'`, and you only use `data` in your component, but not error or any loading flag, you will only re-render if `data` changes. If you start do observe errors as well, they will be "implicitly added" to notifyOnChangeProps. Unless there is reason that you are reading a value from the queryResult in you render function, but you don't want to render that component if that value changes, thus choosing to deliberately display outdated data, this should cover most bases.
> 
> what you cannot do at the moment is subscribing to parts of data with this. say you only access `data.foo`, and `data.bar` changes - your component will re-render with this approach.
> 
> that's where `select` comes in. with `select`, you can transform the data to what you need in your component, and only subscribe to those changes. So if you do: `select: data => data.foo`, you have just subscribed to `foo` only, and your component will no longer re-render if `bar` changes.
> 
> together, those two props (select and tracked queries) give you a very high amount of render optimization.
> 
> the problem with a "manual" function is the same as with `shouldComponentUpdate`. It is quite error prone to get right, leaving out things that you are actually using, but forget to add to that function will lead to stale data etc etc.

Well using "tracked" with select has some obvious disadvantages. Sometimes you one want to track data.foo, but what you actually want to use is foo.bar. Select is nice for getting the necessary data from the response. If you make "tracked" rely on select, it makes select unuseful in some situations.
Another problem (maybe just for me). I don't really understand how "tracked" works. Data comes from JSON and I don't know how it detects changed. (Maybe it's problem that docs are not clear). I understand your position too, I will try to think of some situation where the current solution might cause inconvenience.

TkDodo:
> I don't really understand how "tracked" works. Data comes from JSON and I don't know how it detects changes.

so tracked doesn't really do much 😅 . It only tracks access on the "top level" of the `queryInfo` to determine: Is the consumer interested in `isLoading` ? Or in `data`, or in `error`. It doesn't go "deep" into data.

where it plays nicely together with select is the "structural sharing" feature of react-query. If you have a json structure in `data` like this:

```
foo: { hello: 'world' },
bar: 1,
```

and the next response comes from the backend incrementing `bar` to `2`, the structural sharing feature will make sure that `data.foo` stays referentially ident to the one from the "previous" render (I hope I'm not wrong on this 😅 ).

so if you only `select: data => data.foo`, your component will not re-render in this example _if_ you set your subscriber to only be notified when `data` changes (either via setting nofiyOnChangeProps manually or by tracking and only using `data` in your component).

faradaytrs:
> so tracked doesn't really do much 😅 . It only tracks access on the "top level" of the `queryInfo` to determine: Is the consumer interested in `isLoading` ? Or in `data`, or in `error`. It doesn't go "deep" into data.

I actually got it, it's different from what I thought it is... I believe "tracked" should be the default option. 
What I offered appears to be a different feature.

TkDodo:
> I believe "tracked" should be the default option.

yes, we discussed that in the PR, but for some edge cases (observing values in callback functions like event handlers), it was "too risky" for a minor release, so we made it opt-in. it might become default in the next major.

> What I offered appears to be a different feature.

how about you open a new discussion with a feature request and we can discuss it further there :)

faradaytrs:
> > I believe "tracked" should be the default option.
> 
> yes, we discussed that in the PR, but for some edge cases (observing values in callback functions like event handlers), it was "too risky" for a minor release, so we made it opt-in. it might become default in the next major.
> 
> > What I offered appears to be a different feature.
> 
> how about you open a new discussion with a feature request and we can discuss it further there :)

well let me think of how it might look like and then i will.

PR #1585, opened 2021-01-06
docs(useMutation): onMutate should be async

hardfire:
Since we await in the function body, onMutate should be an async function.


boschni:
Thanks!

PR #1586, opened 2021-01-06
Binded class methods so its possible to destruct queryClient object w…

### LINKED ISSUE #1575: Explode useQueryClient losing this reference
**Describe the bug**
explode `const {invalidateQuery} = useQueryClient()` makes the `invalidateQuery` lose `this` context

**To Reproduce**
`const {invalidateQuery} = useQueryClient()`
`invalidateQuery()`
https://codesandbox.io/s/clever-keldysh-704g7?fontsize=14&hidenavigation=1&theme=dark&file=/src/index.js

**Expected behavior**
If exploded the `useQueryClient`  object, invalidateQuery should not lost `this` reference

 - OS: Linux Ubuntu
 - Browser: Firefox
 - Version: Latest

lrdass:
closes #1575 
Simple fix so that the  `queryClient` object is now destructable.
Just binded every class method

### DIFF
--- a/src/core/queryClient.ts
+++ b/src/core/queryClient.ts
@@ -68,6 +68,36 @@ export class QueryClient {
     this.defaultOptions = config.defaultOptions || {}
     this.queryDefaults = []
     this.mutationDefaults = []
+
+		this.invalidateQueries = this.invalidateQueries.bind(this)
+		this.fetchQuery = this.fetchQuery.bind(this)
+		this.isFetching = this.isFetching.bind(this)
+		this.getQueryData = this.getQueryData.bind(this)
+		this.setQueryData = this.setQueryData.bind(this)
+		this.getQueryState = this.getQueryState.bind(this)
+		this.removeQueries = this.removeQueries.bind(this)
+		this.resetQueries = this.resetQueries.bind(this)
+		this.cancelQueries = this.cancelQueries.bind(this)
+		this.refetchQueries = this.refetchQueries.bind(this)
+		this.fetchQuery = this.fetchQuery.bind(this)
+		this.prefetchQuery = this.prefetchQuery.bind(this)
+		this.fetchInfiniteQuery = this.fetchInfiniteQuery.bind(this)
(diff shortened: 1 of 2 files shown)


tannerlinsley:
What does this do to bundle size?

tannerlinsley:
@boschni What are your thoughts here?

lrdass:
> @boschni What are your thoughts here?

To be honest, I don't know the "standart" way to measure this precisily.
So I've done a `du -h` of `/lib` after building it.

results are:
before
 ```
   8,0K    lib/persist-localstorage-experimental
   72K lib/
    68K lib/react
   16K lib/hydration
   132K    lib/core
   304K    lib/
```

and  after the changes
```
8,0K	lib/persist-localstorage-experimental
72K	lib/devtools
68K	lib/react
16K	lib/hydration
136K	lib/core
308K	lib/

```

**Review by tannerlinsley — CHANGES_REQUESTED** [review]
If my memory serves me well, shouldn't we be able to auto-bind these class methods with the `method = () => {}` syntax?

```js
class Tanner {
  posts: Array<any>

  constructor () {
    this.posts = []
  }

  getPosts = () => {
    return this.posts
  }
}

const tanner = new Tanner()

const {
  getPosts
} = tanner

getPosts()
```

lrdass:
> If my memory serves me well, shouldn't we be able to auto-bind these class methods with the `method = () => {}` syntax?
> 
> ```js
> class Tanner {
>   posts: Array<any>
> 
>   constructor () {
>     this.posts = []
>   }
> 
>   getPosts = () => {
>     return this.posts
>   }
> }
> 
> const tanner = new Tanner()
> 
> const {
>   getPosts
> } = tanner
> 
> getPosts()
> ```

Yea, it would! Going to work on it.

boschni:
Public class fields will eventually generate the same amount of code. Usually class methods are not bound, which means you cannot do something like `const { toISOString }  = new Date()`. I'm fine with that behavior, but if we really want to bind them you can create a `bindAll` utility function to automatically bind all methods.

PR #1599, opened 2021-01-07
shorter code on doc about mutations

kud:


TkDodo:
Hi @kud , thank you for your contribution.

My personal opinion is that, with consistency in mind, I don't see why we should change that in one place only, when all the other places use the ternaries.

It's quite a code-style related topic that differs from project to project, but I think this read also helps: https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

kud:
Interesting! I'll keep it mind!

Well, in this case, we've got change this so?  https://react-query.tanstack.com/guides/mutations#resetting-mutation-state

TkDodo:
I've also used `&&` in one of the examples I made 😅  - I think this is just because different people have different code styles and thus, code looks a bit different. Not sure if we have a strict rule for one or the other - I'm generally un-opinionated in this regard :)

tannerlinsley:
IMO, it's become very standard to use ternaries + `null` instead of `&&`, mostly because of the edge cases around rendering `undefined` and `false` (which render to strings)

PR #1609, opened 2021-01-09
docs: mention query options strictly for useQuery

waptik:
This is in accordance with issues faced when trying use certain query options in `prefetchQuery` and `prefetchInfiniteQuery`. I opened a [discussion](https://github.com/tannerlinsley/react-query/discussions/1596) and was [suggested](https://github.com/tannerlinsley/react-query/discussions/1596#discussioncomment-267856) to open a PR to mention/describe the options available.

### DIFF
--- a/docs/src/pages/reference/QueryClient.md
+++ b/docs/src/pages/reference/QueryClient.md
@@ -88,7 +88,7 @@ try {
 
 **Options**
 
-The options for `fetchQuery` are exactly the same as those of [`useQuery`](#usequery).
+The options for `fetchQuery` are exactly the same as those of [`useQuery`](#usequery), except the following: `enabled, refetchInterval, refetchIntervalInBackground, refetchOnWindowFocus, refetchOnReconnect, notifyOnChangeProps, notifyOnChangePropsExclusions, onSuccess, onError, onSettled, useErrorBoundary, select, suspense, keepPreviousData, placeholderData`; which are stictly for useQuery and useInfiniteQuery. You can check the [source code](https://github.com/tannerlinsley/react-query/blob/361935a12cec6f36d0bd6ba12e84136c405047c5/src/core/types.ts#L83) for more clarity.
 
 **Returns**
 
@@ -109,7 +109,7 @@ try {
 
 **Options**
 
-The options for `fetchInfiniteQuery` are exactly the same as those of [`useInfiniteQuery`](#useinfinitequery).
+The options for `fetchInfiniteQuery` are exactly the same as those of [`fetchQuery`](#queryclientfetchquery).
 
 **Returns**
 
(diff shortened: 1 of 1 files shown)


docs/src/pages/reference/QueryClient.md
```diff
@@ -131,7 +131,7 @@ await queryClient.prefetchQuery(queryKey)
 
 **Options**
 
-The options for `prefetchQuery` are exactly the same as those of [`useQuery`](#usequery).
+The options for `prefetchQuery` are exactly the same as those of [`useQuery`](#usequery) except the following: `enabled, refetchInterval, refetchIntervalInBackground, refetchOnWindowFocus, refetchOnReconnect, notifyOnChangeProps, notifyOnChangePropsExclusions, onSuccess, onError, onSettled, useErrorBoundary, select, suspense, keepPreviousData, placeholderData`; which are stictly for useQuery and useInfiniteQuery. You can check the [source code](https://github.com/tannerlinsley/react-query/blob/361935a12cec6f36d0bd6ba12e84136c405047c5/src/core/types.ts#L83) for more clarity.
```
TkDodo, line 134: what about `fetchQuery?` there, it still says:

> The options for `fetchQuery` are exactly the same as those of [`useQuery`](#usequery).

which is also not true. My suggestion was to add this exceptions to `fetchQuery` and then have all others reference that:

```diff
- The options for `prefetchQuery` are exactly the same as those of [`useQuery`](#usequery).
+ The options for `prefetchQuery` are exactly the same as those of [`fetchQuery`](#queryclientfetchquery).

Comment (tannerlinsley) [reply docs/src/pages/reference/QueryClient.md:134]: Yeah, I like this idea, too 👍

File: docs/src/pages/reference/QueryClient.md [hunk 2]
```diff
@@ -88,7 +88,7 @@ try {
 
 **Options**
 
-The options for `fetchQuery` are exactly the same as those of [`useQuery`](#usequery).
+The options for `fetchQuery` are exactly the same as those of [`useQuery`](#usequery), except the following: `enabled, refetchInterval, refetchIntervalInBackground, refetchOnWindowFocus, refetchOnReconnect, notifyOnChangeProps, notifyOnChangePropsExclusions, onSuccess, onError, onSettled, useErrorBoundary, select, suspense, keepPreviousData, placeholderData`; which are stictly for useQuery and useInfiniteQuery. You can check the [source code](https://github.com/tannerlinsley/react-query/blob/361935a12cec6f36d0bd6ba12e84136c405047c5/src/core/types.ts#L83) for more clarity.
```
TkDodo, line 91: - the link to `useQuery` is not working for me - can you maybe fix that, too?

- if we are linking to the source, should we link to the `master` branch? I usually like permalinks, but it will be outdated if we add more things

waptik: I thought of linking to master but what if the content gets outdated as you say? That's why i linked it to a recent commit.

waptik: I made the change and it looks good https://react-query-ifwq37wlh.vercel.app/reference/QueryClient#queryclientfetchquery

TkDodo (approved):

PR #1643, opened 2021-01-13
Fix typo: useMemo -> useEffect

AjaxSolutions:


AjaxSolutions:
See [discussion](https://github.com/tannerlinsley/react-query/discussions/1641).

tannerlinsley:
My intuition is that this might have adverse effects on things that we possibly have not tested yet, but I'm going to trust the tests here and merge. Be prepared for a possible issue to arise and for this PR to be rolled back.... or it will go unnoticed forever because everything is fine 😉 We'll see! 🥂

lynxtaa:
I think this PR should be reverted: it breaks SSR in Next.js.  See https://github.com/tannerlinsley/react-query/issues/1661

AjaxSolutions:
> I think this PR should be reverted: it breaks SSR in Next.js. See #1661

You might be right and thanks for checking.  My intent was to point out that the useMemo hook is misused when it runs a side effect.

useMemo should be used to memoize the return value of a function, which is not what what this hook is doing.

Also, see how the Apollo Client uses the [ssr](https://www.apollographql.com/docs/react/api/react/hooks/) option.

ssr - boolean - Pass in false to skip your query during server-side rendering.

tannerlinsley:
Reverted with eb3d700dc522b86f6e9cc42d08e3b6197766c53b

PR #1663, opened 2021-01-17
feat(types): Add optional cancelRefetch field to ResultOptions interface

### LINKED ISSUE #1662: The refetch function returned from useQuery should have a cancelRefetch option
**Describe the bug**
This is related to a question I had asked [previously](https://github.com/tannerlinsley/react-query/discussions/1636). The `refetch` function returned from `useQuery` should have a `cancelRefetch` option exposed in the typescript typings. Currently this option is hard coded to `true` for the `fetchNextPage` and `fetchPreviousPage` methods in `InfiniteQueryObserver`. I think this should be exposed to users because it will allow users to tell `react-query` to cancel the previous query when refetching.

**Current Behavior**
Subsequent calls to the `refetch()` function will wait until the promise is settled to make a new network request.

**Expected behavior**
Subsequent calls to the `refetch({cancelRefetch: true})` function will cancel the previous in-flight request and immediately refetch again. To fix this, the `RefetchOptions` typescript interface should have an additional property: `cancelRefetch?: boolean;`

ashwinGokhale:
Closes #1662

### DIFF
--- a/docs/src/pages/reference/useQuery.md
+++ b/docs/src/pages/reference/useQuery.md
@@ -195,8 +195,9 @@ const result = useQuery({
   - The failure count for the query.
   - Incremented every time the query fails.
   - Reset to `0` when the query succeeds.
-- `refetch: (options: { throwOnError: boolean }) => Promise<UseQueryResult>`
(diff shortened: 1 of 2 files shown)


TkDodo (approved):

boschni:
Awesome!

PR #1665, opened 2021-01-17
Bugfix/keep previous data error

### LINKED ISSUE #1659: Possible inconsistency with `isError` and `error` when using `keepPreviousData: true`
- Turn on `keepPreviousData: true`
- Template using `isError`, not `error`, to detect an error state
- Fire a query that resolves and displays normally
- Fire a query that fails and introduces an error state
- The `isError` state is `false`, but there is clearly an `error` present.

I *think* the desired effect is that if an `error` is encountered in the *new/latest* query, the status should not be `success` with the previous data, but `error` with the new query.

Codesandbox: https://codesandbox.io/s/optimistic-hooks-n5v12?file=/src/index.js

TkDodo:
closes #1659


src/react/tests/useQuery.test.tsx
```diff
@@ -1123,6 +1123,112 @@ describe('useQuery', () => {
...
+      isPreviousData: true,
+    })
+    // Hook state update
+    expect(states[3]).toMatchObject({
+      data: 0,
+      isFetching: true,
+      status: 'success',
+      error: null,
+      isPreviousData: true,
+    })
```
TkDodo, line 1195: I am not really sure why we get two re-renders with the same state, but this is the same for the already existing `keepPreviousData` tests 🤔

PR #1676, opened 2021-01-21
docs: add v2 docs links

babycourageous:
Since folks are still on v2 and asking about those docs this adds a couple front-facing spots to find them: 

1. main repo README
2. main RQ docs nav

Screen Shot 2021-01-21 at 2 37 19 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ Enjoy this library? Try the entire [TanStack](https://tanstack.com)! [React Tabl
 
 ## Visit [react-query.tanstack.com](https://react-query.tanstack.com) for docs, guides, API and more!
 
+Still on **React Query v2**? No problem! Check out the v2 docs here: https://react-query-v2.tanstack.com/.
 ## Quick Features
 
 - Transport/protocol/backend agnostic data fetching (REST, GraphQL, promises, whatever!)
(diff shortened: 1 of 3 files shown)


TkDodo:
Awesome. Minor thing: it doesn’t look so good on mobile, and it’s hard to know on which docs I am because v3 is not “highlighted”:
7CEB8CBA-3326-4BE3-AA42-AB36FEC9592C

babycourageous:
Ah - you're totally right. My initial preview wasn't small enough to catch that. 

I like what Tailwind does - 
image

What do you think of keeping NAV as originally was - but putting a line on the homepage (like I did for README) that basically says "Still on v2? Check out the v2 docs here: LINK"

TkDodo:
Yep, that sounds good

babycourageous:
B4 pushing - how's this?

sm: 

image

md: 

image

lg:

image

TkDodo:
I'd put the link under `here` because now it says `v2` 3 times in the same sentence 😅 . Or even like this:

Still using v2? No problem! [Find the v2 docs here](https://react-query-v2.tanstack.com/).

babycourageous:
Went with the "Still using v2? No problem! Find the v2 docs here." option. Also - apologies for the accidental period in my commit msg. Didn't mean to do that.

TkDodo (approved):

PR #1698, opened 2021-01-27
fix(docs): fix testing example

flextremedev:
The first testing example didn't work because useCustomHook returns undefined when first rendering.
Now we're waiting for useCustomHook to succeed and expect afterwards.

### DIFF
--- a/docs/src/pages/guides/testing.md
+++ b/docs/src/pages/guides/testing.md
@@ -21,8 +21,7 @@ Once installed, a simple test can be written. Given the following custom hook:
 
 ```
 export function useCustomHook() {
-  const { data } = useQuery('customHook', () => 'Hello');
-  return data;
+  return useQuery('customHook', () => 'Hello');
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/src/pages/guides/testing.md
```diff
@@ -36,9 +36,11 @@ const wrapper = ({ children }) => (
   </QueryClientProvider>
 );
 
-const { result } = renderHook(() => useCustomHook(), { wrapper });
+const { result, waitFor } = renderHook(() => useCustomHook(), { wrapper });
 
-expect(result.current).toEqual('Hello');
+await waitFor(() => result.current.isSuccess);
+
+expect(result.current.data).toEqual("Hello");
```
boschni, line 43: Shouldn't `useCustomHook` return the full result object for this to work?

flextremedev: Yes definitely! Sorry, must have been late...

PR #1707, opened 2021-01-28
docs (useQuery): add some missing items to api

babycourageous:
The useQuery reference was missing some return value documentation as well as some option values.

**Options**
`initialDataUpdatedAt` was added to the list of options accepted and the api explanation

**Returns**
`dataUpdatedAt`, `errorUpdatedAt`, `isLoadingError`, and `isRefetchError` was added to the list of return values
`dataUpdatedAt` and `errorUpdatedAt` was added to the api explanation

### DIFF
--- a/docs/src/pages/reference/useQuery.md
+++ b/docs/src/pages/reference/useQuery.md
@@ -6,15 +6,19 @@ title: useQuery
 ```js
 const {
   data,
+  dataUpdatedAt,
   error,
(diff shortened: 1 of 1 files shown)


babycourageous:
I think this covers all the return values 
The missing `initialDataUpdatedAt` option popped out for me, but there may be other options missing still. Let me know and I'll be happy to add them to the PR.

PR #1717, opened 2021-01-31
fix(queryObserver): Load currentResult for placeholder data after first update

### LINKED ISSUE #1522: placeholderData called twice
[Open in CodeSandbox](https://codesandbox.io/s/github/tannerlinsley/react-query/tree/master/examples/basic)

add this code to the example, you will see it's called twice
```js
 { placeholderData: () => { console.log(Date.now()); } }
```

The doc says "This function will be executed only once"

DennyScott:
Only the `this.currentQuery` data is checked for stored data when updating results, where placeholder data is stored only in `this.currentResult` on update. So if we hit the placeholder conditional, we check to see if there is data already in the `currentResult` before running the placeholder fn.

It could be argued that this logic could be pulled out of the placeholder conditional, since if any other flow ever does this, it would also need this data, but unsure of the effect of this change.

Fixes #1522

### DIFF
--- a/src/core/queryObserver.ts
+++ b/src/core/queryObserver.ts
@@ -406,10 +406,18 @@ export class QueryObserver<
       typeof data === 'undefined' &&
       status === 'loading'
     ) {
-      const placeholderData =
+      let placeholderData : TData | undefined;
+
+      // Use the previous result data if exists instead of readding placeholder data
+      if(this.currentResult && typeof this.currentResult.data !== 'undefined') {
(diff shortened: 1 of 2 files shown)


boschni:
Will be fixed by https://github.com/tannerlinsley/react-query/pull/1775 . It cannot be guaranteed the function will only trigger once though, but you can provide a memoized function (or the result of a memoized function) as placeholder data to reduce resource usage if needed.

PR #1728, opened 2021-02-02
chore: set exact as true by default

PepijnSenders:
Closes https://github.com/tannerlinsley/react-query/issues/1725


TkDodo:
Not sure what the reason was to change this behavior in v3. Let’s wait for the input of @tannerlinsley here. 

Maybe we should only change it for `getQueryData` and not for `.find` ?

boschni:
Followed up by https://github.com/tannerlinsley/react-query/pull/1756

PR #1739, opened 2021-02-04
refactor react-query core react integration to observer hooks

nksaraf:
Refactor the integration between the core and react-query to use `useQueryObserver` instead of `useBaseQuery` and add `useMutationObserver` and `useQueriesObserver` hooks. These hooks allow the consumer to pass a custom QueryObserver/MutationObserver/etc to customize these hooks where you want to synchronize the data from another external source outside of the react lifecycle.

The purpose was to allow overriden QueryObserver classes to be passed as the

### DIFF
--- a/src/core/queriesObserver.ts
+++ b/src/core/queriesObserver.ts
@@ -5,17 +5,26 @@ import type { QueryClient } from './queryClient'
 import { QueryObserver } from './queryObserver'
 import { Subscribable } from './subscribable'
 
-type QueriesObserverListener = (result: QueryObserverResult[]) => void
+type QueriesObserverListener<TData, TError> = (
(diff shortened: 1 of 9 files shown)


boschni:
Think observers are not the right level for this as they are tightly coupled with the hooks and they are meant for watching a query. The QueryCache or Query would probably be better suited for this. You could even do something like this right now:
```js
const queryCache = new QueryCache()

// Listen for query changes
queryCache.subscribe(() => {
  // Iterate all queries and check if the data in Apollo is different from the data in RQ
  queryCache.getAll().forEach(query => {
    // If the data is different, copy the data from Apollo to RQ
    if (condition) {
      query.setData(newData)
    }
  })
})
```

tannerlinsley:
Marking as stale

PR #1754, opened 2021-02-06
WIP: feat: add suspense to useQueries

boschni:
Fixes https://github.com/tannerlinsley/react-query/issues/1523

### DIFF
--- a/src/react/useQueries.ts
+++ b/src/react/useQueries.ts
@@ -3,31 +3,114 @@ import React from 'react'
 import { notifyManager } from '../core/notifyManager'
 import { QueriesObserver } from '../core/queriesObserver'
 import { useQueryClient } from './QueryClientProvider'
+import { useQueryErrorResetBoundary } from './QueryErrorResetBoundary'
 import { UseQueryOptions, UseQueryResult } from './types'
 
-export function useQueries(queries: UseQueryOptions[]): UseQueryResult[] {
+export function useQueries<TData = unknown, TError = unknown>(
+  queries: UseQueryOptions[]
+): UseQueryResult<TData, TError>[] {
   const queryClient = useQueryClient()
+  const errorResetBoundary = useQueryErrorResetBoundary()
+  const defaultedQueries: UseQueryOptions[] = []
+
+  let someSuspense = false
+  let someUseErrorBoundary = false
+
+  queries.forEach(options => {
+    const defaulted = queryClient.defaultQueryObserverOptions(options)
+
+    // Include callbacks in batch renders
+    if (defaulted.onError) {
+      defaulted.onError = notifyManager.batchCalls(defaulted.onError)
+    }
+
+    if (defaulted.onSuccess) {
+      defaulted.onSuccess = notifyManager.batchCalls(defaulted.onSuccess)
+    }
+
+    if (defaulted.onSettled) {
+      defaulted.onSettled = notifyManager.batchCalls(defaulted.onSettled)
+    }
+
+    if (defaulted.suspense) {
+      someSuspense = true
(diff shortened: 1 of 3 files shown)


tannerlinsley:
Anything I can do to help out here? Any questions?

src/react/useQueries.ts
```diff
@@ -3,31 +3,114 @@ import React from 'react'
 import { notifyManager } from '../core/notifyManager'
 import { QueriesObserver } from '../core/queriesObserver'
 import { useQueryClient } from './QueryClientProvider'
+import { useQueryErrorResetBoundary } from './QueryErrorResetBoundary'
 import { UseQueryOptions, UseQueryResult } from './types'
 
-export function useQueries(queries: UseQueryOptions[]): UseQueryResult[] {
+export function useQueries<TData = unknown, TError = unknown>(
```
johnnyreilly, line 9: Nice work!

I have a question. This adds the ability to specify generics for `useQueries`. Reading this I think that this may fall into the ["generic that is only used in the return position" anti-pattern](https://github.com/microsoft/TypeScript/issues/36062#issuecomment-571799283) category.

It's possible I'm reading it wrong so I wanted to ask a clarifying question: Does the approach here allow a user to specify types using the generic parameters that would differ from the return type of the query or of the `select` function?

If that's possible then the generics would not be adding a great deal of value as incorrectly typed data could still flow through.

Is that possible in this case? 

[As discussed](https://github.com/tannerlinsley/react-query/pull/1527#issuecomment-761564193), I'm generally keen on making use of type inference. But in the absence of that, avoiding type definitions that can possibly let incorrect types through is still worth aiming for.

escaton:
@boschni do you need any help with this PR? I'm really looking forward to this feature and ready to help you finish it.

tibi77:
> @boschni do you need any help with this PR? I'm really looking forward to this feature and ready to help you finish it.

I would like to give a hand as well if there is a need.
I need this feature as well :D

steevsachs:
@boschni We would also benefit from this fix, and I'd be happy to contribute if that'd be helpful.

@johnnyreilly It appears the original review just raised a question about the new type annotation.  Is that the only outstanding change requested on this PR (besides conflict resolutions at this point : ))?  

If no other changes are requested, I'd be happy to resolve the conflicts and revert the generics type change.  Then @boschni can open a future change to re-add generics pending that discussion.  Would that be ok?

johnnyreilly:
Go for it!

steevsachs:
Thanks for confirming.  This merge turns out to be a bit of a doozy, but I should be able to get through it over this weekend.

steevsachs:
@boschni @johnnyreilly Quick question as I work through what's amounting more to refactor of the fix than conflict resolution (given the amount of change to the target logic that's occurred since this was opened):  

The fix here assumes `useQueries` should throw as a whole with the first error encountered, is that the expected behavior?

escaton:
Talking about errors, my thought is that it should be configured with `useErrorBoundary`, not `useSuspense`.
But for consistency with `useQuery` i guess: yes it should throw with the first error.
https://github.com/tannerlinsley/react-query/blob/master/src/react/useBaseQuery.ts#L132

steevsachs:
I've opened the new PR as https://github.com/tannerlinsley/react-query/pull/2109 due to the significant amount of change from this original PR.  The commit in this PR is included in that PR.  Please let me know if you'd prefer this be handled differently git-wise.

Requesting a very thorough review of this one, as it was significantly more complicated than I expected and my context as a contributor going in was basically nil. Please be brutal : ) I did confirm it working with a basic CRA and yarn link.

PR #1756, opened 2021-02-06
fix: single query methods should be exact by default

boschni:
Fixes https://github.com/tannerlinsley/react-query/issues/1725


TkDodo (approved):

tannerlinsley:
👍

PR #1775, opened 2021-02-10
feat: add support for concurrent mode

boschni:
This PR adds support for concurrent mode rendering.

Render functions will now only have these side effects:
- Creating a cache entry (should be allowed)
- Dispatching a GET request (should be allowed)

When calling a render function, it will return an optimistic result, without actually changing the observer state. The observer state will only be changed when new options are set/committed in an effect.

Unfortunately I was not able to keep the referential integrity of the result object between renders because we do not know which optimistic result will eventually get committed.

Fixes https://github.com/tannerlinsley/react-query/issues/1536


klarstrup:
I don't have a minimal repro at present, but I believe this change has broken `useInfiniteQuery` SSR in my app. Worked on 3.8.3, simply renders with `data: undefined` on 3.9.0.

boschni:
Could you provide some more information? Does it only apply to infinite queries? What options are you using? Is your usage similar to https://github.com/tannerlinsley/react-query/tree/master/examples/nextjs ?

klarstrup:
My specific setup is using `react-ssr-prepass` but that example is similar enough I believe. I'm working on a minimal repro.

klarstrup:
And yeah, only useInfiniteQuery is misbehaving for SSR.

klarstrup:
Digging a bit `useInfiniteQuery` now behaves like `useQuery` during SSR: It does not use the `data.pages` pattern as it does client-side.

klarstrup:
Alright @boschni here's reproduction based on the given example: https://github.com/klarstrup/react-query/commit/c2434e563038a7adf9013b30fc916276c3efda3c

klarstrup:
Sorry for spamming here, I'll make an issue,

PR #1776, opened 2021-02-10
add typescript example for basic mutation

SirwanAfifi:
Looking at the documentation I couldn't find an example of using `useMutation` hook. so I have put together a simple example showing the usage of the hook in a React/TypeScript app


TkDodo:
we have this, which does cover `useMutation`, typescript and optimistic updates: https://react-query.tanstack.com/examples/optimistic-updates-typescript

tannerlinsley:
To keep the maintenance burden as low as possible, I think we should instead direct our efforts to improving the existing optimistic updates typescript example (if needed at all).

PR #1781, opened 2021-02-11
fix: queries should not end up in an error state when a component unmounts

boschni:
Follow up on: https://github.com/tannerlinsley/react-query/issues/1412


dantman:
I'm having a persistent bug in an application I'm currently working on due to this bug. I've tested this PR in my codebase and it fixes the bug. Would love to see it released.

jakepeyser:
Any chance this can be shipped in a v2 patch? I'd be happy to put up a PR based off v2.26.4

PR #1792, opened 2021-02-13
feat(subscriptions): add event types and metadata to subscriptions

tannerlinsley:


src/core/queryCache.ts
```diff
@@ -20,7 +20,18 @@ interface QueryHashMap {
   [hash: string]: Query<any, any>
 }
 
-type QueryCacheListener = (query?: Query) => void
+interface QueryCacheNotifyMeta {
```
boschni, line 23: If you define the event separately you can specify which additional properties it has, like:
```ts
interface QueryAddedEvent {
  type: 'queryAdded'
  query: Query
}

interface QueryUpdatedEvent {
  type: 'queryUpdated'
  query: Query
  action: Action
}

type Events = QueryAddedEvent | QueryUpdatedEvent
```
This way TS can check if the correct properties are set when defining an event, but it can also be used when using an event:
```ts
if (event.type === 'queryUpdated') {
  console.log(event.action) // TS knows action is there
}
```

src/core/queryCache.ts
```diff
@@ -65,7 +76,9 @@ export class QueryCache extends Subscribable<QueryCacheListener> {
     if (!this.queriesMap[query.queryHash]) {
       this.queriesMap[query.queryHash] = query
       this.queries.push(query)
-      this.notify(query)
+      this.notify(query, {
```
boschni, line 79: Usually only messages are dispatched on event busses. As `notify` is not a public API we could change it to:
```ts
this.notify({ type: 'queryAdded', query })
```

src/core/queryCache.ts
tannerlinsley: Thanks for the tip. I'll implement this right now :)

src/core/queryCache.ts
tannerlinsley: Done

src/core/query.ts
```diff
@@ -106,6 +106,7 @@ interface ContinueAction {
 interface SetStateAction<TData, TError> {
   type: 'setState'
   state: QueryState<TData, TError>
+  setStateOptions?: SetStateOptions
```
boschni, line 109: Think this change is not needed anymore?

src/core/queryCache.ts
```diff
@@ -20,7 +20,46 @@ interface QueryHashMap {
...
+  action: Action<any, any>
+}
+
+type QueryCacheNotifyEvent =
+  | NotifyEventQueryAdded
+  | NotifyEventQueryRemoved
+  | NotifyEventObserverAdded
+  | NotifyEventObserverRemoved
+  | NotifyEventUpdateResults
+  | NotifyEventDispatch
```
boschni, line 60: Dispatch seems a bit implementation specific, what about `NotifyEventQueryUpdate` and `NotifyEventQueryObserverUpdate`? Maybe also an idea to add the observer to the observer events?

tannerlinsley: Done and done.

src/core/query.ts
tannerlinsley: Done.

src/core/query.ts
```diff
@@ -118,6 +118,10 @@ export type Action<TData, TError> =
   | SetStateAction<TData, TError>
   | SuccessAction<TData>
 
+export interface SetStateOptions {
```
boschni, line 121: Leftover ;)

src/core/queryCache.ts
```diff
@@ -20,7 +21,48 @@ interface QueryHashMap {
...
+  query: Query<any, any>
+  action: Action<any, any>
+}
+
+type QueryCacheNotifyEvent =
+  | NotifyEventQueryAdded
+  | NotifyEventQueryRemoved
+  | NotifyEventObserverAdded
+  | NotifyEventObserverRemoved
+  | NotifyEventUpdateResults
```
boschni, line 62: Might be good to add `Observer` to the name like `NotifyEventQueryObserverUpdated` as this event is observer specific?

src/core/queryCache.ts
```diff
@@ -20,7 +21,48 @@ interface QueryHashMap {
...
+  observer: QueryObserver<any, any, any, any>
+}
+
+interface NotifyEventUpdateResults {
+  type: 'updateResults'
+  query: Query<any, any>
+}
+
+interface NotifyEventQueryUpdated {
+  type: 'dispatch'
```
boschni, line 52: Think it would be good to keep the event name and type in sync like `queryUpdated`

boschni (approved):

PR #1793, opened 2021-02-13
feat(broadcastQueryClient): experimental support for tab/window syncing

tannerlinsley:


tannerlinsley:
This PR relies on #1792

src/broadcastQueryClient-experimental/index.ts
```diff
@@ -0,0 +1,84 @@
...
+      return
+    }
+
+    const { type, queryHash, queryKey, state } = action
+
+    if (type === 'queryUpdated') {
+      const query = queryCache.get(queryHash)
+
+      if (query) {
+        query.setState(state, {
```
boschni, line 60: We might also be able to use a flag to indicate a transaction:
```ts
function broadcastQueryClient() {
  let transaction = false
  
  queryClient.getQueryCache().subscribe((query, event) => {
    if (transaction) {
      return
    }
  })
  
  channel.onmessage = action => {
    transaction = true
    query.setState(state)
    transaction = false
  }
}
```

docs/src/pages/plugins/broadcastQueryClient.md
```diff
@@ -0,0 +1,59 @@
...
+broadcastQueryClient({ queryClient, broadcastChannel })
+```
+
+### `Options`
+
+An object of options:
+
+```ts
+interface PersistQueryClientOptions {
+  /** The QueryClient to persist */
```
TkDodo, line 45: I think this is still copy-pated from the persist plugin and should be `BroadcastQueryClientOptions` here :)

TkDodo (commented):
General question: It seems that we are not syncing query-clients, but queries inside the query-cache? If that's the case, can we make this a bit more obvious in the docs? Settings like `queryDefaults` on the client would _not_ be synced as far as I can tell.

Also: Does this work if I have two apps that use the same `broadcastChannel`? If so, do both apps need to call `broadcastQueryClient`, or is this meant for the same app opened in multiple browsers only?

src/broadcastQueryClient-experimental/index.ts
tannerlinsley: Great idea, done.

docs/src/pages/plugins/broadcastQueryClient.md
tannerlinsley: Fixed 👍

tannerlinsley: To answer your questions:
- **For now**, we are only syncing the cache within the clients, but similar to the persistence plugins, we could include other things as well like mutations and settings like you mentioned.
- Broadcast channels (and all of its fallback polyfills) are restricted by CORS, so this plugin is meant to sync between similar instance of the same app on the same subdomain/CORS. Anything more than that would require websockets and a common server connection, but that's a whole other topic :)

docs/src/pages/plugins/broadcastQueryClient.md
```diff
@@ -0,0 +1,59 @@
...
+```ts
+broadcastQueryClient({ queryClient, broadcastChannel })
+```
+
+### `Options`
+
+An object of options:
+
+```ts
+interface PersistQueryClientOptions {
```
TkDodo, line 44: it still says `PersistQueryClientOptions` here 😅 

```suggestion
interface BroadcastQueryClientOptions {
```

TkDodo (approved):

PR #1798, opened 2021-02-13
fix: respect the enabled flag

robbtraister:
The 3.9 update seems to have broken the `enabled` logic in certain circumstances. I think this fixes it (I tested in my app locally).

### DIFF
--- a/src/core/queryObserver.ts
+++ b/src/core/queryObserver.ts
@@ -666,8 +666,8 @@ function shouldFetchOptionally(
   prevOptions: QueryObserverOptions<any, any>
 ): boolean {
   return (
-    (query !== prevQuery ||
-      (options.enabled !== false && prevOptions.enabled === false)) &&
+    (options.enabled !== false && 
+      (query !== prevQuery || prevOptions.enabled === false)) &&
     isStale(query, options)
   )
 }


tannerlinsley:
Tests are failing so it seems it might need some more work.

robbtraister:
@tannerlinsley I'm happy to look into this further, but I'm not sure I agree with the premise of the failing test. If `enabled` is false, I would expect `isFetching` to never be true.

I guess there's an inconsistency where the consumer has both specified `enabled` _and_ called `refetch()`. I still expected `enabled` to take precedence.

Could you explain the rationale for this behavior?

boschni:
Thanks for the PR! Fixed in 3.9.4 (also added a test to cover this scenario)

TkDodo:
> I guess there's an inconsistency where the consumer has both specified enabled: false and called refetch(). I still expected enabled to take precedence.

refetch bypasses the enabled flag, i think that’s on purpose. It’s one way of escaping the declarative react-query world and say: i only want to fetch once when user does X, so you disable the query in general and call refetch on this one instance imperatively.

robbtraister:
I just updated my app to 3.9.5 and it's working as expected. Thanks so much for the quick turn-around!

PR #1804, opened 2021-02-15
refactor(QueryClientProvider): validation for context being overwritten

JobenM:
I had an issue within our application which basically meant we were overwriting our `ReactQueryClientContext` with the default because of some unexpected rerendering. This meant even having set `<QueryClientProvider>` at basically root we were still getting the error `No Query set...` error. I think it would be great to supply a correct error message in this scenario as I'm sure I won't be the only one to come across this in the future.

First time contributing here, let me know if I need to change anything :)

### DIFF
--- a/src/react/QueryClientProvider.tsx
+++ b/src/react/QueryClientProvider.tsx
@@ -6,6 +6,10 @@ const QueryClientContext = (() => {
   const context = React.createContext<QueryClient | undefined>(undefined)
   if (typeof window !== 'undefined') {
     // @ts-ignore
+    if (typeof window.ReactQueryClientContext !== 'undefined') {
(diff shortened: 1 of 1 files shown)


TkDodo:
Please have a look at this issue: https://github.com/tannerlinsley/react-query/issues/1764

i fear this PR might worsen the situation if you mount two independent microfrontends, since the second one will just get the Error :/

tannerlinsley:
I’m working on a fix for all of this right now. Stay tuned.

Tanner Linsley

tannerlinsley:
I've got a separate fix which also addresses a few other related concerns in #1805

JobenM:
Awesome. Cheers @tannerlinsley 👍

PR #1810, opened 2021-02-16
docs: add isFetched to useQuery

bjarkehs:
I could not tell if this was left out intentionally or simply an oversight.


Fhiltyphil88 (changes requested):
#Trigger

PR #1838, opened 2021-02-21
Feature/error for retry delay

TkDodo:
pass error as second parameter to retryDelay function

this streamlines the interfaces of `retry` and `retryDelay` functions.

One use-case would be to have a retry that depends on the `Retry-After` header sent from the backend, see also: https://twitter.com/SleeplessByte/status/1362946642513719296


src/core/retryer.ts
```diff
@@ -163,7 +166,10 @@ export class Retryer<TData = unknown, TError = unknown> {
           // Do we need to retry the request?
           const retry = config.retry ?? 3
           const retryDelay = config.retryDelay ?? defaultRetryDelay
-          const delay = functionalUpdate(retryDelay, this.failureCount) || 0
+          const delay =
+            typeof retryDelay === 'function'
+              ? retryDelay(this.failureCount, error)
+              : retryDelay
```
TkDodo, line 172: I am not sure why `|| 0` was needed here before. On type level, the function was only allowed to return a `number`, and we also use the `defaultRetryDelay` in case `retryDelay` was `undefined`

tannerlinsley (approved):
Looks great! Derk will be happy about this one.

SleeplessByte:
Oooooh yeah! I only have praise. Thanks @TkDodo!

TkDodo:
Can we merge this?

PR #1854, opened 2021-02-23
Prefer ReadonlyArray for query keys

pluma:
Mutable arrays can be passed to functions expecting readonly arrays but readonly arrays can't be passed to functions expecting mutable arrays. Since `react-query` doesn't actually mutate query keys, it should use the more permissive `ReadonlyArray` for `QueryKey` (and `MutationKey`) instead.

This allows users to do things like this without having to strip `readonly` from the inferred types:

```
const query = [["my.query.key", "abc"], () => fetch("/my/resource")] as const
```

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -3,7 +3,7 @@ import type { QueryBehavior } from './query'
 import type { RetryValue, RetryDelayValue } from './retryer'
 import type { QueryFilters } from './utils'
 
-export type QueryKey = string | ReadonlyArray<unknown>
+export type QueryKey = string | readonly unknown[]
 
 export type QueryFunction<T = unknown> = (
   context: QueryFunctionContext<any>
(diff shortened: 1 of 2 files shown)


src/core/types.ts
```diff
@@ -3,7 +3,7 @@ import type { QueryBehavior } from './query'
 import type { RetryValue, RetryDelayValue } from './retryer'
 import type { QueryFilters } from './utils'
 
-export type QueryKey = string | unknown[]
+export type QueryKey = string | ReadonlyArray<unknown>
```
TkDodo, line 6: minor nitpick since we use `[]` everywhere else would be to change it to `readonly unknown[]`

pluma: done

TkDodo (approved):

tannerlinsley (approved):

pluma:
@tannerlinsley Do I need to do anything for this to be merged?

PR #1921, opened 2021-03-08
Add throwOnError to prefetchQuery()

PepijnSenders:
## Description

This PR adds a `throwOnError` option to `prefetchQuery()`. With this new option, devs can opt to have `prefetchQuery()` throw an error when one happens inside the `queryFn`.

Example:

```ts
try {
  await queryClient.prefetchQuery('test', () => throwError(), { throwOnError: true })
} catch (err) {
  console.error('error happened', err);
}
```

### DIFF
--- a/docs/src/pages/guides/prefetching.md
+++ b/docs/src/pages/guides/prefetching.md
@@ -14,6 +14,7 @@ const prefetchTodos = async () => {
 
 - If data for this query is already in the cache and **not invalidated**, the data will not be fetched
 - If a `staleTime` is passed eg. `prefetchQuery('todos', fn, { staleTime: 5000 })` and the data is older than the specified staleTime, the query will be fetched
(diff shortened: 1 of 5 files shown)


tannerlinsley:
I believe this is what fetchQuery is for, right?

PepijnSenders:
I feel like the semantics for `fetchQuery()` aren't the same as `prefetchQuery()`. In cases where I want to prefetch something and not care about the response, I feel like having a `prefetchQuery()` call that could throw an error would be semantically better.

tannerlinsley:
Other than the name, you just described fetchQuery, which returns a promise or throws an error which must be handled as opposed to prefetchQuery which does exactly the same thing, but returns void.

Tanner Linsley

TkDodo:
I agree that the difference to `fetchQuery` is very minor. You would get an error, but not data? If that's what you want, you can also use `fetchQuery`, `await` it and discard the result ...

PR #1934, opened 2021-03-09
fix: Do not refetch active queries if "refetchActive" is set to "false"

DerJacques:
This PR fixes the issue reported here: https://github.com/tannerlinsley/react-query/issues/1931

When a query was invalidated and `refetchActive` was set to `false`, the query would still be refetched, unless `refetchInactive` was set to `true`. 

The PR fixes this by not letting the `inactive` option in the query filters default to `false` if it was actually `undefined`.


tannerlinsley (approved):

DerJacques:
@tannerlinsley Should I do anything else to have this PR merged? 😊

Issue #1462, opened 2020-12-17
TypeScript: Not possible to pass the type of the QueryKey to `useQuery`(react-query v3.3.0)


jon301:
**Describe the bug**
Current typings of `useQuery` (also `prefetchQuery` and `fetchQuery`) support string-only query keys.
It seems not possible to pass a custom type (for array keys).

**To Reproduce**
https://codesandbox.io/s/objective-nobel-8incc?file=/src/App.tsx

**Expected behavior**
It should be possible to pass a generic type for query keys.

**Desktop**
 - Version: 3.3.0

TkDodo:
not sure what you mean. You can pass anything serializable as a queryKey. Have you read the migration guide, especially:

https://react-query.tanstack.com/guides/migrating-to-react-query-3#query-key-partspieces-are-no-longer-automatically-spread-to-the-query-function

react-query will no longer "inject" the queryKey to your fetchFn. It is recommended to use inline functions, especially with typeScript to get typesafety:

```
const fetchContentById = async (id: number) => ...
...
const id = 1;
const query = useQuery(["contentById", id], () => fetchContentById(id));
```

jon301:
Hi @TkDodo , thanks for your answer.

![image](https://user-images.githubusercontent.com/122108/102461548-d2007380-4048-11eb-9abf-67c32a920edc.png)

This is basically what I want to achieve with TypeScript:
https://react-query.tanstack.com/guides/query-functions#query-function-variables

TkDodo:
Ah yes, I think you are right. I haven't actually used the context before, the types are defined like this:

```
export declare type QueryKey = string | unknown[];
export declare type QueryFunction<T = unknown> = (context: QueryFunctionContext) => T | Promise<T>;
export interface QueryFunctionContext<TQueryKey extends QueryKey = QueryKey, TPageParam = any> {
    queryKey: TQueryKey;
    pageParam?: TPageParam;
}
```

so the QueryFunction only takes a QueryFunctionContext without generic, I think we might need to change that to:

```diff
export declare type QueryKey = string | unknown[];
- export declare type QueryFunction<T = unknown> = (context: QueryFunctionContext) => T | Promise<T>;
+ export declare type QueryFunction<T = unknown, TQueryKey extends QueryKey = QueryKey> = (context: QueryFunctionContext<QueryKey>) => T | Promise<T>;
export interface QueryFunctionContext<TQueryKey extends QueryKey = QueryKey, TPageParam = any> {
[...]

PR #1955, opened 2021-03-12
fix(useQuery): let context infer queryKey

onursagir:
Allows for the queryKey in the queryFn to be inferred as discussed in  #1462 

For example
image

### DIFF
--- a/src/react/types.ts
+++ b/src/react/types.ts
@@ -4,6 +4,8 @@ import {
   InfiniteQueryObserverResult,
   MutateOptions,
   MutationStatus,
+  QueryFunction,
+  QueryKey,
   QueryObserverOptions,
   QueryObserverResult,
 } from '../core/types'
@@ -18,8 +20,16 @@ export interface UseBaseQueryOptions<
 export interface UseQueryOptions<
   TQueryFnData = unknown,
   TError = unknown,
-  TData = TQueryFnData
-> extends UseBaseQueryOptions<TQueryFnData, TError, TData> {}
+  TData = TQueryFnData,
+  TQueryKey extends QueryKey = QueryKey
+>
+  extends Omit<
+    UseBaseQueryOptions<TQueryFnData, TError, TData>,
+    'queryFn' | 'queryKey'
+  > {
+  queryKey?: TQueryKey
+  queryFn?: QueryFunction<TQueryFnData, TQueryKey>
+}
 
 export interface UseInfiniteQueryOptions<
   TQueryFnData = unknown,
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -5,9 +5,10 @@ import type { QueryFilters } from './utils'
 
 export type QueryKey = string | readonly unknown[]
 
-export type QueryFunction<T = unknown> = (
-  context: QueryFunctionContext<any>
-) => T | Promise<T>
+export type QueryFunction<
+  T = unknown,
+  TQueryKey extends QueryKey = QueryKey
+> = (context: QueryFunctionContext<TQueryKey>) => T | Promise<T>
 
 export interface QueryFunctionContext<
   TQueryKey extends QueryKey = QueryKey,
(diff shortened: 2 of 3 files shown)


src/react/types.ts
```diff
@@ -18,8 +20,16 @@ export interface UseBaseQueryOptions<
...
   TError = unknown,
-  TData = TQueryFnData
-> extends UseBaseQueryOptions<TQueryFnData, TError, TData> {}
+  TData = TQueryFnData,
+  TQueryKey extends QueryKey = QueryKey
+>
+  extends Omit<
+    UseBaseQueryOptions<TQueryFnData, TError, TData>,
+    'queryFn' | 'queryKey'
+  > {
```
TkDodo, line 29: why not change this on `UseBaseQueryOptions` directly? As far as I can see, this is the only interface that extends `UseBaseQueryOptions`....

src/react/types.ts
```diff
@@ -18,8 +20,16 @@ export interface UseBaseQueryOptions<
...
+  queryKey?: TQueryKey
+  queryFn?: QueryFunction<TQueryFnData, TQueryKey>
+}
 
 export interface UseInfiniteQueryOptions<
```
TkDodo, line 34: could you maybe also add the support for infinite queries as well?

src/react/types.ts
onursagir: there is a function in `src/react/useBaseQuery.ts:9` that uses that uses the UseBaseQueryOptions interface, I mainly did it this way to keep the impact minimal

src/react/types.ts
onursagir: I could in the future, but I think I would prefer to do that on a separate branch (and perhaps another one for useMutation?).

src/react/types.ts
TkDodo: I think this goes hand-in-hand with the request for infinite query support, because `src/react/useBaseQuery.ts:9` is only used for `useQuery` - which you have already adapted, and `useInfiniteQuery`. With regards to the queryKey, they are identical, so if you add useInfiniteQuery support to this PR, you could also get rid of the Omit I think.

onursagir: Yes I see what you mean now. I will update it accordingly.

onursagir: I had a check and I don't think I can make it work like that the interfaces are put together like this

```
interface UseQueryOptions extends UseBaseQueryOptions extends QueryObserverOptions {}
interface UseInfiniteQueryOptions extends InfiniteQueryObserverOptions extends QueryObserverOptions {}
```

While it is true that useQuery and useInfiniteQuery do both use the useBaseQuery function, I don't think that matters when it comes to inferring the queryKey (which is either inferred through `QueryFunction` or `UseQueryOptions`).

TkDodo: @boschni what do you think?

boschni: Think we should try to keep the API consistent and implement things like this library wide instead of in one hook

onursagir: @boschni This pr includes no API changes, only a minor change to an interface to resolve a minor TypeScript inconvenience. I am not saying the interfaces for ouseInfiniteQuery can't be changed in order for inference to work. I'm saying it can't be done as @TkDodo [describes](https://github.com/tannerlinsley/react-query/pull/1955#discussion_r593250990)

boschni: If someone is opting in to use this feature, they most likely want to use it throughout their codebase, and not only when using `useQuery`. At this moment you might only use `useQuery`, but others are also using `queryClient.fetchQuery` or `useInfiniteQuery`. If we provide support for this pattern, users should be able to expect that they can use it anywhere.

TkDodo: I’ll try to take another stab at this if i find the time - i think we’re already on the right track here 👍

TkDodo:
@onursagir I picked up from where this PR was closed and I think I got it working for useQuery and useInfiniteQuery a like: https://github.com/tannerlinsley/react-query/pull/2047

Issue #1298, opened 2020-11-19
useIsFetching does not count for useMutation


ilkerceng:
**Describe the bug**
I use useIsFetching for global loading component. It works weel with useQuery, but does not triggered with service call using useMutation. I use the latest version 2.26.2.

lucbpz:
Quoting the docs:

> This isFetching property is an integer representing how many queries, if any, in the cache are currently fetching (including background-fetching, loading new pages, or loading more infinite query results)

Mutations are supposed to represent individual actions that change some data, so when you mutate you don't retrieve data or update from the cache, so I would not expect the `isFetching` to represent that state (it is not being represented in the `useMutation` hook).

tannerlinsley:
In the future, we *may* introduce a `useIsMutating` hook or something similar, but for now their current functionality is intentional.

ilkerceng:
@lucbpz @tannerlinsley Thanks 👍

kachkaev:
@tannerlinsley WDYT of `useIsMutating` in some 3.x release? I wonder if the new architecture allows that with relative ease.

tannerlinsley:
I’d be good with that

kripod:
@tannerlinsley is there a way to implement `useIsMutating` as a userland feature? For instance, by creating a context for tracking the amount of pending mutations and then adding a wrapper for `useMutation` to manage the provider-based state.

I would love to hear your recommendations on how to implement this.

TkDodo:
There is a `MutationObserver` exported, but it doesn’t have any documentation as far as I can see.

kripod:
I’ve just created a custom `useIsMutating` hook which looks as follows:

```ts
import * as React from "react";
import { useQueryClient } from "react-query";

export function useIsMutating() {
	const queryClient = useQueryClient();
	const [pendingMutations, setPendingMutations] = React.useState(0);

	React.useEffect(
		() =>
			queryClient.getMutationCache().subscribe(() => {
				setPendingMutations(
					queryClient
						.getMutationCache()
						.getAll()
						.filter((mutation) => mutation.state.status === "loading").length,
				);
			}),
		[queryClient],
	);

	return pendingMutations;
}
```

I tried my best to keep it as simple as possible.

TkDodo:
I think you need to unsubscribe in the effect cleanup function, but looks good otherwise!

kripod:
@TkDodo the `subscribe` method returns with a cleanup function, so it should do that exactly 😊

[...]

PR #1971, opened 2021-03-15
feat: add useIsMutating hook

benediktviebahn:
This PR adds a useIsMutating hook as discussed in #1298.

Let me know if it needs any changes.

### DIFF
--- a/docs/src/manifests/manifest.json
+++ b/docs/src/manifests/manifest.json
@@ -166,9 +166,9 @@
           "editUrl": "/guides/scroll-restoration.md"
         },
         {
-          "title": "Query Filters",
-          "path": "/guides/query-filters",
-          "editUrl": "/guides/query-filters.md"
+          "title": "Filters",
+          "path": "/guides/filters",
+          "editUrl": "/guides/filters.md"
         },
         {
           "title": "SSR & Next.js",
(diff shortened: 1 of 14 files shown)


benediktviebahn:
Any idea why the tests are failing? I saw some lint errors in VS Code but I assumed I had something incorrectly set up.

TkDodo:
looks like a cyclic import:

```
/home/runner/work/react-query/react-query/src/core/retryer.ts
Error:   1:1  error  Dependency cycle via ./utils:2=>./mutation:1  import/no-cycle
Error:   2:1  error  Dependency cycle via ./utils:2=>./mutation:1  import/no-cycle
Error:   3:1  error  Dependency cycle via ./mutation:1             import/no-cycle
```

benediktviebahn:
Oh right. Got it.

TkDodo (commented):
I think this looks good! Two requests:
- can you please add tests for the `useIsMutating` hook as well
- I think we should also document `queryClient.isMutating` (next to [isFetching](https://react-query.tanstack.com/reference/QueryClient#queryclientisfetching))

benediktviebahn:
Thanks for reviewing. Sure I will do that tomorrow.

TkDodo (approved):

TkDodo:
I think this looks splendid! Let's wait for an additional review and then 🚀

tannerlinsley (approved):

PR #1983, opened 2021-03-17
adds onQuery option which is called everytime before a fetch happens

Jeffrey-Zutt:
### Description

This PR adds an `onQuery` option to `useQuery()`. 
Devs can opt to add a callback to handle side-effects related to the query in a similar fashion as `useMutation()` `onMutate`.

Example:
```tsx
const result = useQuery(
   queryKey,
  () => 'data',
  { onQuery: () => console.log('onQuery') }
)
```

Also:
Great job on react-query! Love using it :-)

### DIFF
--- a/docs/src/pages/reference/useQuery.md
+++ b/docs/src/pages/reference/useQuery.md
@@ -34,6 +34,7 @@ const {
   keepPreviousData,
   notifyOnChangeProps,
   notifyOnChangePropsExclusions,
+  onFetch,
   onError,
   onSettled,
   onSuccess,
@@ -127,6 +128,12 @@ const result = useQuery({
   - Optional
   - If set, the component will not re-render if any of the listed properties change.
   - If set to `['isStale']` for example, the component will not re-render when the `isStale` property changes.
+- `onFetch: (context: QueryFunctionContext) => void`
+  - Optional
+  - This function will fire will any time right before the query tries to fetch new data.
+  - Receives a `QueryFunctionContext` object with the following variables:
+    - `queryKey: QueryKey`
+  - Please note: `onFetch` is called for every retry. With default `retry: 3`, this function is called 4 times before `useQuery()` reaches an error state.
 - `onSuccess: (data: TData) => void`
   - Optional
   - This function will fire any time the query successfully fetches new data.
--- a/src/core/infiniteQueryBehavior.ts
+++ b/src/core/infiniteQueryBehavior.ts
@@ -43,6 +43,7 @@ export function infiniteQueryBehavior<
             pageParam: param,
           }
 
+          context.preFetchFn(queryFnContext)
           const queryFnResult = queryFn(queryFnContext)
 
           const promise = Promise.resolve(queryFnResult).then(page => {
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -144,6 +144,11 @@ export interface QueryObserverOptions<
    */
   notifyOnChangePropsExclusions?: Array<keyof InfiniteQueryObserverResult>
   /**
+   * This callback will fire right before the query will start fetching new data.
+   * Useful for side-effects that are specific to this query.
+   */
+  onFetch?: (context:QueryFunctionContext) => void
+  /**
    * This callback will fire any time the query successfully fetches new data.
    */
   onSuccess?: (data: TData) => void
--- a/docs/src/pages/reference/useInfiniteQuery.md
+++ b/docs/src/pages/reference/useInfiniteQuery.md
@@ -31,6 +31,13 @@ The options for `useInfiniteQuery` are identical to the [`useQuery` hook](/refer
     - `pageParam: unknown | undefined`
   - Must return a promise that will either resolves data or throws an error.
   - Make sure you return the data *and* the `pageParam` if needed for use in the props below.
+- `onFetch: (context: QueryFunctionContext) => void`
+  - Optional
+  - This function will fire will any time right before the query tries to fetch new data.
+  - Receives a `QueryFunctionContext` object with the following variables:
+    - `queryKey: QueryKey`
+    - `pageParam: unknown | undefined`
+  - Please note: `onFetch` is called for every retry. With default `retry: 3`, this function is called 4 times before `useInfiniteQuery()` reaches an error state.
 - `getNextPageParam: (lastPage, allPages) => unknown | undefined`
   - When new data is received for this query, this function receives both the last page of the infinite list of data and the full array of all pages.
   - It should return a **single variable** that will be passed as the last optional parameter to your query function.
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -47,6 +47,7 @@ export interface QueryState<TData = unknown, TError = unknown> {
 }
 
 export interface FetchContext<TQueryFnData, TError, TData> {
+  preFetchFn: (context:QueryFunctionContext) => void
   fetchFn: () => unknown | Promise<unknown>
   fetchOptions?: FetchOptions
   options: QueryOptions<TQueryFnData, TError, TData>
@@ -78,6 +79,11 @@ interface FailedAction {
 interface FetchAction {
   type: 'fetch'
   meta?: any
+}
+
+interface PreFetchAction {
+  type: 'preFetch'
+  context: QueryFunctionContext
 }
 
 interface SuccessAction<TData> {
@@ -114,6 +120,7 @@ export type Action<TData, TError> =
   | ErrorAction<TError>
   | FailedAction
   | FetchAction
+  | PreFetchAction
   | InvalidateAction
   | PauseAction
   | SetStateAction<TData, TError>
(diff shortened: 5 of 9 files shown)


docs/src/pages/reference/useQuery.md
```diff
@@ -127,6 +128,9 @@ const result = useQuery({
   - Optional
   - If set, the component will not re-render if any of the listed properties change.
   - If set to `['isStale']` for example, the component will not re-render when the `isStale` property changes.
+- `onQuery: () => void`
+  - Optional
+  - This function fire will any time right before the query tries to fetch new data.
```
daraselia, line 133: `...will fire...`

src/core/retryer.ts
```diff
@@ -6,6 +6,7 @@ import { sleep } from './utils'
 
 interface RetryerConfig<TData = unknown, TError = unknown> {
   fn: () => TData | Promise<TData>
+  onQuery?: () => void
```
daraselia, line 9: Would it be useful to pass query time variables as an argument to callback function?

docs/src/pages/reference/useQuery.md
Jeffrey-Zutt: Thanks :-)

TkDodo:
How will this work for infinite queries please? When a refetch happens, will it be called once, or once for each page that is refetched?

Jeffrey-Zutt:
Hi TkoDodo, it will be called for every page that is fetched

TkDodo:
> Hi TkoDodo, it will be called for every page that is fetched

cool, can you add a test for this maybe? Also, should we get access to the `pageParams` or the `index` or so ? it also needs documentation please :)

Jeffrey-Zutt:
Yes, of course :-).

- Added a test for `useInfiniteQuery`
- onQuery is now called with a `QueryFunctionContext`.
    - updated tests accordingly  
- Updated documentation

Please let me know if you want anything changed!

src/core/types.ts
```diff
@@ -62,6 +62,11 @@ export interface QueryOptions<
   initialData?: TData | InitialDataFunction<TData>
   initialDataUpdatedAt?: number | (() => number | undefined)
   behavior?: QueryBehavior<TQueryFnData, TError, TData>
+  /**
+   * This callback will fire right before the query will start fetching new data.
+   * Useful for side-effects that are specific to this query.
+   */
+  onQuery?: (context:QueryFunctionContext) => void
```
TkDodo, line 69: should this not be defined on the `QueryObserverOptions`, like the other callbacks (`onSuccess`, `onError`, `onSettled`) ? We want to be able to have different callbacks per observer

src/core/infiniteQueryBehavior.ts
```diff
@@ -8,7 +8,43 @@ export function infiniteQueryBehavior<
...
+            ? pageParam
+            : getPreviousPageParam(context.options, oldPages)
+        }
+
+        const queryFnContext: QueryFunctionContext = {
+          queryKey: context.queryKey,
+          pageParam: param
+        }
+
+        context.options.onQuery?.(queryFnContext)
```
TkDodo, line 46: - none of this needs to be done if `context.options.onQuery` is `undefined`
- i see a bit of an overlap with what we are doing in `context.fetchFn`. not sure if it's good to call `getNextPageParam` / `getPreviousPageParam` twice - once for `onQuery` and once for the actual fetch. I'm not sure here so I'd like to wait on more feedback from either @tannerlinsley or @boschni

TkDodo (changes requested):
I think it's good 👍 , but need some more feedback. I would also appreciate another test that shows that `onQuery` is called even when a `retry` happens, maybe we should even mention this in the docs? Because if an error happens, with default of `retry: 3`, the function would be called 4 times before we get to an error state (unless I misinterpreted the code 😅 )

src/core/infiniteQueryBehavior.ts
boschni: Hi Jeffrey! Think this can also be implemented in the `QueryObserver.onQueryUpdate` method. There you should be able to listen for the fetch action and trigger the callback. Then it is consistent with the other callbacks and it also works outside of the hooks. Regarding the naming, maybe `onFetch` would be a bit more descriptive? Then it also corresponds more with the `fetchFn` option.

src/core/types.ts
Jeffrey-Zutt: Thank you! These are valid points.

- I moved `onQuery` to `QueryObserverOptions`.
- When `onQuery` exists,  we're calling it from `queryFn()` now, instead of the retryer.
    - It makes the code a lot simpler: no need to modify infiniteQueryBehavior anymore.

src/core/infiniteQueryBehavior.ts
Jeffrey-Zutt: Please see previous comment, this whole code-block is unnecessary because `onQuery` is now called from `queryFn` .

Jeffrey-Zutt:
> I would also appreciate another test that shows that `onQuery` is called even when a `retry` happens, maybe we should even mention this in the docs? Because if an error happens, with default of `retry: 3`, the function would be called 4 times before we get to an error state (unless I misinterpreted the code 😅 )

I added the test and updated the docs.
Please let me know what you think :-)

src/core/infiniteQueryBehavior.ts
boschni: My comment was not really related to this piece of code, just placed it here because I got tagged ;)

TkDodo: > maybe onFetch would be a bit more descriptive?

I like both `onQuery` as well as `onFetch` 🤔 . `onQuery` for `useQuery` corresponds well with `onMutate` for `useMutation`.`onFetch` would also make sense because we have the `isFetching` flag 🤷

boschni: Yeah it also relates to the `refetch()` method in `useQuery`. When doing mutations you call `mutate()` so it makes sense to use `onMutate` there.

TkDodo: if you prefer `onFetch`, then let's rename 👍

Jeffrey-Zutt: Thanks guys! I'll rename to `onFetch` and move the call to `QueryObserver.onQueryUpdate`.

Jeffrey-Zutt: @TkDodo @boschni 
Update:

- I renamed the function `onFetch`.
- The function is called from the `queryObserver`

Let me know what you think :-).

boschni:
Nice! Two question that come to mind:

1. Should the `onFetch` function be called once, or also on each retry? What is its purpose?

2. The functionality can also be achieved with `useEffect`, do we need a separate callback?

```js
const { isFetching } = useQuery()

useEffect(() => {
  if (isFetching) {
    // Run code...
  }
}, [isFetching])
```

src/core/infiniteQueryBehavior.ts
Jeffrey-Zutt: @TkDodo @boschni 
- I renamed the function `onFetch`
- I moved the call to `QueryObserver.onQueryUpdate`

Let me know what you think :-)

TkDodo:
> The functionality can also be achieved with useEffect, do we need a separate callback?

can't all of the callbacks theoretically be achieved with useEffect?

if we include retries and call it once per retry, I don't think that we can actually do it with useEffect :)

Jeffrey-Zutt:
@boschni 

1. We'd like onFetch to be called on every retry as well. For instance, we'd like to able to log query-duration for client-side analytics. As @TkDodo says, that's currently not possible using useEffect with isFetching.

2. For consistency/simplicity I think it's nice to be able to configure these callbacks in a similar way as `useMutation`.
I think it's easier to configure them 'from top', instead of relying on its result to hook into it.

E.g.

```tsx
const useMyQuery = (enableLogging:boolean) => {
   return useQuery(
      key, 
      () => { /* fetch */ }, 
      enableLogging 
            ? configForLogging
            : undefined
)}
```

Instead of:

```tsx
const useMyQuery = (enableLogging:boolean) => {
   const { isFetching, ...rest } = useQuery(
      key, 
      () => { /* fetch */ }
   )
   
   // I don't like this as tightly coupled here.
   React.useEffect(() => {
      if (enableLogging) {
         if (isFetching) {
            startClock()
         } else {
            stopClock()
         }
      }
      
   }, [ isFetching, enableLogging ])
		
   return {
      isFetching,
      ...rest
   }
)}
```

I think you're right though. We could probably solve it using some well chosen `useEffect()`, or some other way.
I also think the onFetch is a nice addition for consistency :-).

boschni:
I'm not sure if this would be the best approach for request logging because if you have multiple components using the same query then it would be registered multiple times. Usually I place functionality like this (authorization/logging) in an API client:

```js
class SomeApi {
  fetchSomething() {
     this.request("/something")
  }
  request(url) {
    log(url)
  }
}

const api = new SomeApi()

function Component() {
  const result = useQuery(() => api.fetchSomething())
}
```

TkDodo (approved):

tannerlinsley (commented):
I personally don’t find this addition helpful considering this functionality is easily achieved in user land by either augmenting or proxying the query function with the desired additional fetch logic. It doesn’t add any control or configuration to the internal lifecycle that is unique beyond being a “hook” into the fetch function, which already exist in user land anyway.

tannerlinsley:
Just to make sure none of this was in vain, I would gladly accept a PR to improve the documentation on this point. It's much easier to educate people that they can write their own "onFetch" functionality by simply using their own function that's passed to useQuery. This pattern offers more control, is an API decision that inverts control and power back to the user instead of drawing it away into abstraction, and will arguably help any dev that implements the pattern become more skilled in utilizing the general pattern of inversion of control to achieve domain-specific results.

PR #2047, opened 2021-03-29
fix(useQuery): let type of queryKey for the QueryContext be inferred from the key passed to useQuery

TkDodo:
This PR picks up where https://github.com/tannerlinsley/react-query/pull/1955 left off and infers the QueryKey of the `QueryContext` passed to the queryFn from the actual QueryKey


src/react/useBaseQuery.ts
```diff
@@ -96,7 +117,7 @@ export function useBaseQuery<TQueryFnData, TError, TData, TQueryData>(
     throw obsRef.current
       .fetchOptimistic(defaultedOptions)
       .then(({ data }) => {
-        defaultedOptions.onSuccess?.(data)
+        defaultedOptions.onSuccess?.(data as TData)
```
TkDodo, line 120: @boschni I went a bit down the rabbithole with that last commit (395a5099), but I think types are now more exact than before. here, for example, `data` is inferred to `TData | undefined`, which is correct, and it was `any` before. But `onSuccess` expects `TData`, which is why I added the type assertion. I think this is the only real change - everything else is just drilling generics :)

src/core/query.ts
```diff
@@ -363,7 +379,7 @@ export class Query<
 
     // Create query function context
     const queryKey = ensureArray(this.queryKey)
-    const queryFnContext: QueryFunctionContext = {
+    const queryFnContext: QueryFunctionContext<unknown[]> = {
```
TkDodo, line 382: I needed to use `unknown[]` here because after `ensureArray`, `TQueryKey` is no longer true, as that contains strings as well. I think it's not a big deal, but if you know how to make `ensureArray` return `TQueryKey` if the input is an array and `[string]` is not, we could improve this further.

tannerlinsley (approved):
This looks good to me, but I'd like @boschni's approval as well before merging.

src/react/useBaseQuery.ts
TkDodo: can we merge it then? I'd like to fix types of `initialData` but I don't want to start until this huge type monster is merged :)

PR #2071, opened 2021-04-01
docs(useQuery): change isFetching default to false

gornyyvladimir:
Based on this question https://github.com/tannerlinsley/react-query/issues/624#issuecomment-778182411
Now isFetching is `false` when enabled is `false`

### DIFF
--- a/docs/src/pages/reference/useQuery.md
+++ b/docs/src/pages/reference/useQuery.md
@@ -211,7 +211,7 @@ const result = useQuery({
   - Will be `true` if the query has been fetched after the component mounted.
   - This property can be used to not show any previously cached data.
 - `isFetching: boolean`
-  - Defaults to `true` so long as `enabled` is set to `false`
+  - Defaults to `false` so long as `enabled` is set to `false`
   - Will be `true` if the query is currently fetching, including background fetching.
 - `failureCount: number`
   - The failure count for the query.


docs/src/pages/reference/useQuery.md
```diff
@@ -211,7 +211,7 @@ const result = useQuery({
   - Will be `true` if the query has been fetched after the component mounted.
   - This property can be used to not show any previously cached data.
 - `isFetching: boolean`
-  - Defaults to `true` so long as `enabled` is set to `false`
+  - Defaults to `false` so long as `enabled` is set to `false`
```
TkDodo, line 214: I don’t understand that sentence at all. isFetching is returned from useQuery - it doesn’t "default" to anything. In my understanding, this flag will always be true whenever a request for this query is in-flight. This includes initial loading as well as background refetches.

TkDodo (changes requested):

docs/src/pages/reference/useQuery.md
gornyyvladimir: Ok, maybe better to remove this sentence from docs?

gornyyvladimir: Because we can not fetch data while enabled is `false`

gornyyvladimir: But sentence in docs is:

> Defaults to `true` so long as `enabled` is set to `false`

And from this information, I expect isFetching to always be `true` while enabled `false`.

TkDodo: I would just change the sentence to:

```
- `isFetching: boolean`
  - this flag is `true` whenever a request is in-flight, which includes initial `loading` as well as background refetches.
```

you can also change that sentence semantically, but that is what this flag is doing.

PR #2073, opened 2021-04-02
docs(QueryStatus): change from enum to union type

luisadame:
### Description

In version 2 the `QueryStatus` typing was:
```typescript
export enum QueryStatus {
  Idle = 'idle',
  Loading = 'loading',
  Error = 'error',
  Success = 'success',
}
```
https://github.com/tannerlinsley/react-query/blame/8aeb209c3b9c84382a88059afbc398be93a69b33/src/core/types.ts#L183

And now in version 3 is:
```typescript
export type QueryStatus = 'idle' | 'loading' | 'error' | 'success'
```
https://github.com/tannerlinsley/react-query/blame/7b14ef519a544d725f2cff99574f0a01f92b8c94/src/core/types.ts#L257

### Changes
- documented the change described above
- added underline to links that are children of headings
  image
  image
  this affects anchor links but these already have an underline property set, so I think there's no issue with that

### DIFF
--- a/docs/src/pages/guides/migrating-to-react-query-3.md
+++ b/docs/src/pages/guides/migrating-to-react-query-3.md
@@ -116,6 +116,46 @@ Together, these provide the same experience as before, but with added control to
 
 **Notice that it's now a function instead of a property**
 
+### `QueryStatus` has been changed from an [enum](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) to a [union type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types)
+
+So, if you were checking the status property of a query or mutation against a QueryStatus enum property you will have to check it now against the string literal the enum previously held for each property.
+
+Therefore you have to change the enum properties to their equivalent string literal, like this:
+- `QueryStatus.Idle` -> `'idle'`
+- `QueryStatus.Loading` -> `'loading'`
+- `QueryStatus.Error` -> `'error'`
+- `QueryStatus.Success` -> `'success'`
+
+This is an example of how you would have done it in the previous version:
+
+```js
+import { QueryStatus } from 'react-query';
+
+const { data, status } = useQuery(['post', id], (_key, id) => fetchPost(id))
+
+if (status === QueryStatus.Loading) {
+  ...
+}
+
+if (status === QueryStatus.Error) {
+  ...
+}
+```
+
+And this is how it is now in version 3:
+
+```js
+const { data, status } = useQuery(['post', id], () => fetchPost(id))
+
+if (status === 'loading') {
+  ...
+}
+
+if (status === 'error') {
+  ...
+}
+```
+
 ### The `useQueryCache` hook has been replaced by the `useQueryClient` hook.
 
 It returns the provided `queryClient` for its component tree and shouldn't need much tweaking beyond a rename.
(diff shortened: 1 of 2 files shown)


docs/src/pages/guides/migrating-to-react-query-3.md
```diff
@@ -116,6 +116,46 @@ Together, these provide the same experience as before, but with added control to
 
 **Notice that it's now a function instead of a property**
 
+### `QueryStatus` has been changed from an [enum](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) to a [union type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types)
```
TkDodo, line 119: Can you please mention somewhere that this applies to `TypeScript` only - maybe a dedicated _TypeScript_ sub-section?

docs/src/pages/guides/migrating-to-react-query-3.md
```diff
@@ -116,6 +116,46 @@ Together, these provide the same experience as before, but with added control to
...
+const { data, status } = useQuery(['post', id], () => fetchPost(id))
+
+if (status === 'loading') {
+  ...
+}
+
+if (status === 'error') {
+  ...
+}
+```
```
TkDodo, line 157: maybe we can show a `diff`:

```diff
- import { useQuery, QueryStatus } from 'react-query'
+ import { useQuery } from 'react-query'

const { data, status } = useQuery(['post', id], (_key, id) => fetchPost(id))

- if (status === QueryStatus.Loading) {
+ if (status === 'loading') {
  ...
}

- if (status === QueryStatus.Error) {
+ if (status === 'error') {
  ...
}
```

TkDodo (changes requested):

docs/src/pages/guides/migrating-to-react-query-3.md
luisadame: Sure, we will have to show the diff on the query function too right? now it's given a context as a parameter not the query keys

docs/src/pages/guides/migrating-to-react-query-3.md
luisadame: yup :+1:

docs/src/pages/guides/migrating-to-react-query-3.md
TkDodo: ah yes, I missed that!

docs/src/pages/guides/migrating-to-react-query-3.md
luisadame: done!

docs/src/pages/guides/migrating-to-react-query-3.md
luisadame: I'm a bit hesitant about the diff, I don't know if it seems a bit cluttered the example or if it's just me, it looks fine anyways
image

TkDodo: right, maybe we could get rid of the difference for the actual useQuery call? the `(_key, id)` change is not actually part of the union type change, so we can just use the inline function for both version? Then it looks good I'd say :) I updated the diff here: https://github.com/tannerlinsley/react-query/pull/2073#discussion_r606296742

luisadame: Done :+1:

TkDodo (approved):

PR #2074, opened 2021-04-02
Add TQueryKey to QueryFunction generic type in QueryOptions, remove ensureArray from Query

kachkaev:
This PR fixes the following edge case issue in typings:

```ts
import { QueryFunction, useQuery } from "react-query";

type MyData = number;
type MyQueryKey = readonly ["my-data", number];

const getMyData: QueryFunction<MyData, MyQueryKey> = async ({
  queryKey: [, n],
}) => {
  return n + 42;
};

// ...

const myData = useQuery({
  queryKey: ["my-data", 100],
  queryFn: getMyData,
});
```

**Before**

```
  Overload 1 of 3, '(options: UseQueryOptions<number, unknown, number, (string | number)[]>): UseQueryResult<number, unknown>', gave the following error.
    Type 'QueryFunction<number, MyQueryKey>' is not assignable to type 'QueryFunction<number, QueryKey>'.
      Type 'QueryKey' is not assignable to type 'MyQueryKey'.
        Type 'string' is not assignable to type 'readonly ["my-data", number]'.
  Overload 2 of 3, '(queryKey: QueryKey, options?: UseQueryOptions<unknown, unknown, unknown, QueryKey> | undefined): UseQueryResult<unknown, unknown>', gave the following error.
    Argument of type '{ queryKey: (string | number)[]; queryFn: QueryFunction<number, MyQueryKey>; }' is not assignable to parameter of type 'QueryKey'.
      Object literal may only specify known properties, and 'queryKey' does not exist in type 'readonly unknown[]'.ts(2769)
types.d.ts(32, 5): The expected type comes from property 'queryFn' which is declared here on type 'UseQueryOptions<number, unknown, number, (string | number)[]>'
```

**After**

No TS error

### DIFF
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -379,7 +379,7 @@ export class Query<
 
     // Create query function context
     const queryKey = ensureArray(this.queryKey)
-    const queryFnContext: QueryFunctionContext<unknown[]> = {
+    const queryFnContext: QueryFunctionContext<TQueryKey> = {
       queryKey,
       pageParam: undefined,
     }
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -378,7 +378,7 @@ export class Query<
     }
 
     // Create query function context
-    const queryKey = ensureArray(this.queryKey)
+    const queryKey = (ensureArray(this.queryKey) as unknown) as TQueryKey
     const queryFnContext: QueryFunctionContext<TQueryKey> = {
       queryKey,
       pageParam: undefined,
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -1,6 +1,5 @@
 import {
   Updater,
-  ensureArray,
   functionalUpdate,
   isValidTimeout,
   noop,
@@ -393,7 +392,7 @@ export class Query<
     const context: FetchContext<TQueryFnData, TError, TData, any> = {
       fetchOptions,
       options: this.options,
-      queryKey: ensureArray(this.queryKey),
+      queryKey: this.queryKey,
       state: this.state,
       fetchFn,
     }
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -378,9 +378,8 @@ export class Query<
     }
 
     // Create query function context
-    const queryKey = (ensureArray(this.queryKey) as unknown) as TQueryKey
     const queryFnContext: QueryFunctionContext<TQueryKey> = {
-      queryKey,
+      queryKey: this.queryKey,
       pageParam: undefined,
     }
 
@@ -394,7 +393,7 @@ export class Query<
     const context: FetchContext<TQueryFnData, TError, TData, any> = {
       fetchOptions,
       options: this.options,
-      queryKey,
+      queryKey: ensureArray(this.queryKey),
       state: this.state,
       fetchFn,
     }
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -59,7 +59,7 @@ export interface QueryOptions<
   retryDelay?: RetryDelayValue<TError>
   cacheTime?: number
   isDataEqual?: (oldData: TData | undefined, newData: TData) => boolean
-  queryFn?: QueryFunction<TQueryFnData>
+  queryFn?: QueryFunction<TQueryFnData, TQueryKey>
   queryHash?: string
   queryKey?: TQueryKey
   queryKeyHashFn?: QueryKeyHashFunction<TQueryKey>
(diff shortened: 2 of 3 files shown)


src/core/query.ts
```diff
@@ -378,8 +378,8 @@ export class Query<
     }
 
     // Create query function context
-    const queryKey = ensureArray(this.queryKey)
-    const queryFnContext: QueryFunctionContext<unknown[]> = {
+    const queryKey = (ensureArray(this.queryKey) as unknown) as TQueryKey
+    const queryFnContext: QueryFunctionContext<TQueryKey> = {
```
kachkaev, line 382: Could not find good alternatives to `as unknown`


```ts
    const queryKey = ensureArray(this.queryKey)
    const queryFnContext: QueryFunctionContext<TQueryKey> = {
      queryKey, // Type 'unknown[]' is not assignable to type 'TQueryKey'.
      pageParam: undefined,
    }
```


```ts
const queryKey: TQueryKey = ensureArray(this.queryKey) // Type 'unknown[]' is not assignable to type 'TQueryKey'.
```

TkDodo: The problem seems to be that `ensureArray` narrows the type to `Array`, but `QueryFunctionContext` still wants us to give it a `TQueryKey`. `TQueryKey` can, on type level, also be e.g. a `string`, but it cannot be that after `ensureArray`.

I would also question why we do the `ensureArray` here in the first place, maybe @boschni knows? Because it means that the `queryKey` we get into our context is actually not the same as the queryKey we pass into the queryFn. example:

```
  const queryInfo= useQuery(
    "test",
    async (context) => {
      console.log(context.queryKey.toUpperCase());
      return 5;
    }
  )
```

here, context.queryKey will be an Array: `["test"]` at runtime, but if we do the type assertion like we do it here, it will be of type `"test"`. This means that typescript is fine calling `toUpperCase` on the queryKey, but it will fail at runtime.

---

so maybe we should just remove the `ensureArray` call? Then we'll have no problems with the types, and no problems at runtime - the key will just be what the key is. it might break apps that rely on this conversion, but I highly doubt it since it's not mentioned.

kachkaev: That’s a really good point @TkDodo! I looked at the code again and managed to find a solution in which we don’t don’™ use typecasting, but all tests still pass.

`QueryFunctionContext` gets the original `TQueryKey`, while `FetchContext` gets the one with `ensureArray` around it.

kachkaev: To be honest, all test still pass even if I remove `ensureArray` in `FetchContext`. Not sure if this is good or bad 😁

TkDodo: as far as i can see, `context` is mainly used for `onFetch` of the `infiniteQueryBehavior`. But the type of the queryKey is still `TQueyKey`, which could also be a string on type level. so if you want to use `context.queryKey`, you cannot rely on it being an Array, which is why I think we don't need _ensureArray_ at all, too :)

`context.queryKey` is also only used to pass it to the `queryFn` in `infiniteQueryBehavior`, which is the same use-case as passing it here, so we would have the same runtime problem with infinite queries like I described above.

---

So my vote would go towards removing the `ensureArray` call here completely, but I'd like @tannerlinsley or @boschni to confirm before we do this please :)

kachkaev: Interesting – I did not know about `infiniteQueryBehavior`. Makes sense! I removed `ensureArray` in 415e4a41d2b74e44c01300bfe4b7a3d86a3d7d35 to demonstrate passing tests. Happy to revert if that was not needed!

kachkaev: @tannerlinsley or @boschni WDYT?

boschni: Think it would indeed be better to remove it as this will make sure the query key type is consistent throughout the application.

TkDodo (approved):

src/core/query.ts
TkDodo: it means we'd have a small runtime change as well, but I think it would still qualify as a patch release. transforming the query key to an array when passing it to the queryFn via queryContext was not documented, so I think this is good 👍 .

kachkaev: Thanks for your replies folks! Please let me know if you’d like me to make any further edits 🙂

TkDodo: @boschni are you fine with a patch release?

boschni: Yes I'm fine with a patch release. The current types should already prevent users from using it in such a way.

PR #2081, opened 2021-04-04
Add a minor comment related to typescript

arfanliaqat:
Hi, this may seem insignificant but i just spent hours trying to make useMutation in typescript project, after much debugging and github issues reading i realized only typing my mutationFn should have fixed it. :man_shrugging: I saw this during my debugging but didn't realize it is important, just typing it here so it saves someone trouble.

### DIFF
--- a/examples/optimistic-updates-typescript/pages/index.tsx
+++ b/examples/optimistic-updates-typescript/pages/index.tsx
@@ -21,6 +21,9 @@ type Todos = {
   ts: number
 }
 
+// IMPORTANT!! make sure that your mutationFn (fetchTodos) is property typed as in example below, 
+// otherwise you may run into weird typescript errors
+
 async function fetchTodos(): Promise<Todos> {
   const res = await axios.get('/api/data')
   return res.data


TkDodo (commented):
I don’t think a comment about “weird errors” is going to give much information to users. Here, if you don’t type it, it will be implicitly typed to `Promise<AxiosResponse<any>>`

While not very exact, this is still valid and shouldn’t yield errors. It’s also not related to mutations at all...

arfanliaqat:
Well said @TkDodo 

```
  The last overload gave the following error.
    Argument of type '(data: { id: string; flag: boolean; }) => void' is not assignable to parameter of type 'MutationKey'.
      Type '(data: { id: string; flag: boolean; }) => void' is missing the following properties from type 'readonly unknown[]': concat, join, slice, indexOf, and 17 more.  TS2769

    24 | 
    25 |   const { mutate } = useMutation(
  > 26 |     (data: { id: string; flag: boolean }) => {
       |     ^
    27 |       const { id, flag } = data;
    28 |       return updateFlag(id, flag);
    29 |     }
```

here is the weird error i was facing, i realized another reason for this was that i was not returning promise :facepalm: silly mistake, closing this since i tried and it doesn't throw weird errors if promise is returned :sweat_smile:

PR #2084, opened 2021-04-04
docs(useQuery): add the useErrorBoundary option

ohansemmanuel:
`useErrorBoundary` is listed as a valid option in the useQuery [API reference](https://react-query.tanstack.com/reference/useQuery), however, an actual explanation is missing. This PR adds one. 

CleanShot 2021-04-04 at 18 28 33@2x

### DIFF
--- a/docs/src/pages/reference/useQuery.md
+++ b/docs/src/pages/reference/useQuery.md
@@ -165,6 +165,9 @@ const result = useQuery({
   - Optional
   - Defaults to `true`
   - If set to `false`, structural sharing between query results will be disabled.
+- `useErrorBoundary: boolean`
+  - Defaults to the global query config's `useErrorBoundary` value, which is false
+  - Set this to true if you want mutation errors to be thrown in the render phase and propagate to the nearest error boundary
 
 **Returns**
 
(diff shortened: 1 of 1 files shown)


docs/src/pages/reference/useQuery.md
```diff
@@ -165,6 +165,9 @@ const result = useQuery({
   - Optional
   - Defaults to `true`
   - If set to `false`, structural sharing between query results will be disabled.
+- `useErrorBoundary: boolean`
+  - Defaults to the global query config's `useErrorBoundary` value, which is false
+  - Set this to true if you want mutation errors to be thrown in the render phase and propagate to the nearest error boundary
```
TkDodo, line 170: Why mutation errors? This is for useQuery, isn’t it?

TkDodo (changes requested):

docs/src/pages/reference/useQuery.md
ohansemmanuel: My bad. That's absolutely correct. Since errors may still be thrown, I'll kill the "mutation" i.e ... 

> Set this to true if you want ~~mutation~~ errors to be thrown in the render phase and propagate to the nearest error boundary

TkDodo (approved):

PR #2089, opened 2021-04-06
Add `undefined` return type in onMutate 

eomttt:
[useMutation](https://react-query.tanstack.com/reference/useMutation) has void return type but in code has not

In my code

```js
  const onMutate = () => {
    const previous = queryClient.getQueryData<CartProductsDataType>(queryKey);
    return previous;
  };
  const onError = (error: any, variable: AddOptionItem, context: CartProductsDataType | undefined) => {
    queryClient.setQueryData(queryKey, context);
  };

  return useMutation(QueryKey.CARTS, mutateFn, {
    onMutate,
    onError,
  });
```

onMutate return `CartProductsDataType or undefind` because of `getQueryData` but in useMutation error!!!
because useMutation onMutate function type only return `Promise<Context> or Context` not has `undefined`
so i try to add `undefined` type in onMutate return type

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -472,7 +472,7 @@ export interface MutationOptions<
   mutationFn?: MutationFunction<TData, TVariables>
   mutationKey?: MutationKey
   variables?: TVariables
-  onMutate?: (variables: TVariables) => Promise<TContext> | TContext | undefined
+  onMutate?: (variables: TVariables) => Promise<TContext> | Promise<undefined> | TContext | undefined
   onSuccess?: (
     data: TData,
     variables: TVariables,
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -472,7 +472,7 @@ export interface MutationOptions<
   mutationFn?: MutationFunction<TData, TVariables>
   mutationKey?: MutationKey
   variables?: TVariables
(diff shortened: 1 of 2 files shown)


src/core/types.ts
```diff
@@ -472,7 +472,7 @@ export interface MutationOptions<
   mutationFn?: MutationFunction<TData, TVariables>
   mutationKey?: MutationKey
   variables?: TVariables
-  onMutate?: (variables: TVariables) => Promise<TContext> | TContext
+  onMutate?: (variables: TVariables) => Promise<TContext> | TContext | undefined
```
TkDodo, line 475: what about `Promise<undefined>` like in the docs?

eomttt: you mean like this???
```suggestion
  onMutate?: (variables: TVariables) => Promise<TContext | undefined> | TContext | undefined
```

TkDodo: not sure if it isn't rather:

`Promise<TContext> | Promise<undefined> | TContext | undefined`

is there any difference 😅 ?

eomttt: i'm not sure. but your answer is good to me

TkDodo (approved):

PR #2108, opened 2021-04-11
Feature/1824 transform placeholder data with select

### LINKED ISSUE #1824: placeholderData is not transformed by select
**Describe the bug**

`placeholderData` passed to `useQuery`'s option is not transformed by `select` when it exists.

**To Reproduce**

https://codesandbox.io/s/react-query-placeholderdata-with-select-dsvxl

`data.toString()` renders `2020-02-18T15:29:59.326Z` at first.
After 1 second, it renders formatted date like `Fri Feb 19 2021 00:29:59 GMT+0900` as `data` is transformed to Date object by `select` function.

Then replace `placeholderData` to `initialData`. It shows `Wed Feb 19 2020 00:29:59 GMT+0900` and  1 second later `Fri Feb 19 2021 00:29:59 GMT+0900` as both are Date object.

**Expected behavior**

It seems `placeholderData` should be transformed by `select` to keep return type of `data` like `initialData`.

TkDodo:
closes #1824


src/core/queryObserver.ts
```diff
@@ -507,13 +507,31 @@ export class QueryObserver<
...
+            errorUpdatedAt = Date.now()
+            status = 'error'
+          }
+        }
       }
 
       if (typeof placeholderData !== 'undefined') {
         status = 'success'
-        data = placeholderData
+        data = placeholderData as TData
```
TkDodo, line 534: not sure why I need the type assertion here - not really happy with it. `placeholderData` is of type `TData | TQueryFnData` - maybe I have the typings wrong somewhere @boschni?

boschni: Think there is no way around this. If there is no select function or behavior then placeholderData should be of type TData but TS does not know if this is the case upfront

PR #2109, opened 2021-04-11
Feature/use queries suspense

### LINKED ISSUE #1523: suspense is not working for useQueries
**Describe the bug**

**To Reproduce**

```ts

const DemoA = () => {
  return (
    <Suspense fallback={'loading'}>
      <Works />
    </Suspense>
  );
};
const DemoB = () => {
  return (
    <Suspense fallback={'loading'}>
      <NotWork />
    </Suspense>
  );
};
const Works: React.FC = () => {
  // status success success success
  console.log(
    'status',
    [1, 2, 3]
      .map(
        (v) =>
          ({
            queryKey: ['users', v],
            queryFn: fetchApi,
            suspense: true,
          } as UseQueryOptions),
      )
      .map(useQuery)
      .map(({ status }) => status),
  );

  return null;
};
const NotWork: React.FC = () => {
  // status loading loading loading
  // status success loading loading
  // status success success loading
  // status success success success
  console.log(
    'status',
    useQueries(
      [1, 2, 3].map(
        (v) =>
          ({
            queryKey: ['users', v],
            queryFn: fetchApi,
            suspense: true,
          } as UseQueryOptions),
      ),
    ).map(({ status }) => status),
  );
  return null;
};
```

**Expected behavior**
useQueries should works like the useQuery loop, the result status should be success

**Screenshots**

 - OS: macOS
 - Browser: chrome
 - Version: react-query@3.5.5

**Additional context**

steevsachs:
fixes #1523

Attempting to resolve merge conflicts and carry forward the fix from https://github.com/tannerlinsley/react-query/pull/1754

This is my first contribution to the repo and I have low context (both on the repo and the PR, since I didn't open it originally), so I would appreciate someone taking a close look at this change to ensure it does not introduce new issues.  

Notes:
* Running format picked up some unrelated changes to `queryClient.test.tsx`
* I removed the new test counting number of renders because the number changed from the original PR, and seemed brittle and relatively arbitrary in terms of future changes.  Please let me know if that test was desirable and I'll get it back in.


src/react/tests/useQueries.test.tsx
```diff
@@ -35,11 +43,129 @@ describe('useQueries', () => {
...
+      queryClient,
+      <React.Suspense fallback={'test'}>
+        <Page />
+      </React.Suspense>
+    )
+
+    await sleep(10)
+
+    expect(results.length).toBe(1)
+    expect(results[0]).toMatchObject([{ data: 1 }, { data: 2 }])
```
escaton, line 91: i'd expect more from this test. If you are playing with `suspense: true/false(undefined)` queries why not check intermediate values like `[{ data: 1}, { loading: true }]` in this case?

escaton: And actually suspending on first query :)

steevsachs: Thank you for the review!

What I think is happening here is that it _is_ suspending, so fallback is rendering until the hook is no longer throwing, at which point it does its first render with actual data.  Are there intermediate states I would expect to see on first render?  Would a refetch to show new states after original suspense render help to make this case more robust?

escaton: What i mean is: the component suspends on first render, but 5s later it render again and push `[{ data: 1}, { loading: true }]` in results. This should happen because first query suspends since it has `suspense: true` but the second query behaves like regular `useQuery(...{ suspense: false })` doesn't it?

steevsachs: Ah, I see what you mean.  The initial PR suspended when any of the queries was configured for suspense and any of the queries was loading. In other words, if anything had suspense, everything had suspense.  

This could be revised so that it only suspends when any of the queries that are configured for suspense are loading--is that the expected functionality? I'm honestly not sure what a client would expect as behavior here. I feel the same about the current summary error handling.

tannerlinsley: Yeah, the test should go something like:
- Suspends on first load
- Data comes in
- query becomes stale
- query gets refetched
- `isFetching === true`

RoyGrosz:
Any news on this pr?

src/react/tests/useQueries.test.tsx
steevsachs: Revisiting this, apologies for the delay. I do feel a bit out of context on this one : )

Is the desired behavior that all queries have aggregate suspense and error state _until_ the first settlement of all queries, and then queries behave independently? 

My concern with this approach is that the individual queries still accept a fully query config, including options like suspense, but they aren't guaranteed to honor that config (and won't individually honor certain configs like `suspense`).

I wonder if this is a case where `useQueries` and `useQueriesSuspense` should exist as distinct hooks that can explicitly define their option subsets and be opinionated for suspense vs no suspense.

steevsachs:
Apologies for letting this drag @RoyGrosz. The problem turned out to be more complex than expected, and I question the implementation proposed by this PR. Just left another comment for direction going forward.  

Not confident I should be the one owning this issue. For our org we've implemented some more opinionated workarounds, but we would still benefit from built-in support of suspense for a dynamic array of queries.

src/react/tests/useQueries.test.tsx
TkDodo: > I wonder if this is a case where useQueries and useQueriesSuspense should exist as distinct hooks that can explicitly define their option subsets and be opinionated for suspense vs no suspense.

Hmm, I wouldn't be a big fan of this api wise. For example, because we just got #2395 reported: `useErrorBoundary` alone also doesn't work with `useQueries`, and I think the reasons are similar. I tried the reproduction out with this PR and it seems to work.

Now we have to think about if we even want different behaviour for multiple queries in the same `useQueries` batch. E.g. one of them having suspense, while others don't. Or one has useErrorBoundary and another one doesn't.

I would say this shouldn't be allowed, kind of like an invalid state. The current api makes this possible because it's just an array of useQuery configs. If the structure were a bit different, it would be more obvious, something like:

```
useQueries({
    queries: [{ queryKey1, queryFn1, options1 }, { queryKey2, queryFn2, options2 }],
    suspense: boolean,
    useErrorBoundary: boolean
})
```

with that, we would lift "global" options out of the array, so you can only set them once for all queries. The options on each query would then _not_ allow what we have at the top level, e.g. suspense or useErrorBoundary.

---

I don't to propose a breaking change to the useQueries api though, just some food for thought. I am personally fine with doing what we do now, which is:

```
const someSuspense = defaultedQueries.some((q) => q.suspense);
const someUseErrorBoundary = defaultedQueries.some((q) => q.useErrorBoundary);
```

But we'd need to document that setting `suspense` or `useErrorBoundary` to true on _any_ of the queries will affect all queries! It's not that intuitive, but we can revisit it for v4 maybe.

steevsachs:
@escaton Based on continued interest in the comments, I went back through and updated this and improved the test as suggested.  

GH is not allowing me to re-request your review, but this is ready for another look if there's desire to move forward. 

Note: Looks like workflows need approval to run remaining checks.

its-miller-time:
Hey all, I dont want to be a nag, but am also checking daily to see if this feature has been included! We are very excited about the possibilities it opens with our app architecture. If there is something that needs to be worked on, I would be glad to pull this repo down and contribute, however it looks like perhaps its just the testing and review left?

steevsachs:
From my perspective this is just awaiting review, approval. @escaton you'd done the previous review, are you available for another look?

TkDodo:
Has this been addressed (document which settings will be "on" if they are on for one of the queries): https://github.com/tannerlinsley/react-query/pull/2109#discussion_r659152561?

also, please solve the conflicts

steevsachs:
@TkDodo I've resolved the conflict and updated the docs. 

I'm reproducing inconsistent test behavior locally for `useInfiniteQuries` and the new `keepPreviousData` test cases that may be related to timings rather than invalid behavior, do the passed checks here indicate that flakiness is on my end?

TkDodo:
> I'm reproducing inconsistent test behavior locally for useInfiniteQuries and the new keepPreviousData test cases that may be related to timings rather than invalid behavior, do the passed checks here indicate that flakiness is on my end?

hmm, let's see. I've also had some failing tests on node10, they were timing related and I think I fixed them (they were also showing up on CI). Which node version are you using locally?

steevsachs:
@TkDodo Thanks for running the remaining tests, I'm on 14.17.1 locally. Appears the failures do repro in CI. 

Looks from blame like you might have context on the `keepPreviousData` changes--any thoughts on what change here is affecting those test cases?

TkDodo:
> Looks from blame like you might have context on the keepPreviousData changes--any thoughts on what change here is affecting those test cases?

yeah I wrote those tests, and it was merged about a week ago. I'll have to investigate 🤔

src/core/queriesObserver.ts
```diff
@@ -63,6 +71,16 @@ export class QueriesObserver extends Subscribable<QueriesObserverListener> {
     return this.result
   }
 
+  refetch(options?: QueriesRefetchOptions): Promise<QueryObserverResult[]> {
+    let observers = this.observers
+
+    if (options?.filter) {
+      observers = this.observers.filter(options.filter)
+    }
```
TkDodo, line 79: could this just be:

```
const observers = options?.filter ? this.observers.filter(options.filter) : this.observers
```

src/react/tests/useQueries.test.tsx
```diff
@@ -25,7 +40,7 @@ describe('useQueries', () => {
         {
           queryKey: key2,
           queryFn: async () => {
-            await sleep(10)
+            await sleep(7)
```
TkDodo, line 43: is this just to make the tests faster?

src/react/useQueries.ts
```diff
@@ -20,18 +22,30 @@ export function useQueries(queries: UseQueryOptions[]): UseQueryResult[] {
     return defaultedOptions
   })
 
-  const [observer] = React.useState(
-    () => new QueriesObserver(queryClient, defaultedQueries)
+  const observerRef = React.useRef(
+    new QueriesObserver(queryClient, defaultedQueries)
```
TkDodo, line 26: why are we moving away from state to refs please? I explicitly made that change not too long ago :) It's the preferred way of doing one-time initializations -  I don't see us _writing_ to `observerRef.current` anyhwere?

src/react/useQueries.ts
```diff
@@ -20,18 +22,30 @@ export function useQueries(queries: UseQueryOptions[]): UseQueryResult[] {
...
+  const someIsLoading = result.some(r => r.isLoading)
 
   React.useEffect(() => {
     mountedRef.current = true
 
-    const unsubscribe = observer.subscribe(
+    const unsubscribe = observerRef.current.subscribe(
       notifyManager.batchCalls(() => {
-        if (mountedRef.current) {
+        if (mountedRef.current && someIsLoading) {
```
TkDodo, line 48: why do we need this check for `someIsLoading` here please? This looks like we're never force-updating if nothing is loading, which doesn't seem right 🤔

src/react/tests/useQueries.test.tsx
```diff
@@ -36,14 +51,145 @@ describe('useQueries', () => {
...
-    await sleep(30)
+    await sleep(10)
 
     expect(results.length).toBe(3)
     expect(results[0]).toMatchObject([{ data: undefined }, { data: undefined }])
     expect(results[1]).toMatchObject([{ data: 1 }, { data: undefined }])
     expect(results[2]).toMatchObject([{ data: 1 }, { data: 2 }])
   })
 
+  it('should return the correct states with suspense', async () => {
```
TkDodo, line 62: can you maybe add a separate test for only `useErrorBoundary` without `suspense` ?

src/react/tests/useQueries.test.tsx
```diff
@@ -36,14 +51,145 @@ describe('useQueries', () => {
...
+    await waitFor(() => rendered.getByText('retry'))
+    succeed = true
+    fireEvent.click(rendered.getByText('retry'))
+    await waitFor(() => rendered.getByText('data1: data1'))
+    await waitFor(() => rendered.getByText('data2: data2'))
+
+    consoleMock.mockRestore()
+  })
+
   it('should keep previous data if amount of queries is the same', async () => {
```
TkDodo, line 193: regarding the test failures: could it be that the tests somehow influence each other? I see no failures on master, neither locally nor in CI 🤔

src/react/useQueries.ts
```diff
@@ -41,13 +55,29 @@ export function useQueries(queries: UseQueryOptions[]): UseQueryResult[] {
...
-  }, [observer])
+  }, [someIsLoading])
 
-  React.useEffect(() => {
-    // Do not notify on updates because of changes in the options because
-    // these changes should already be reflected in the optimistic result.
-    observer.setQueries(defaultedQueries, { listeners: false })
-  }, [defaultedQueries, observer])
+  const handleReset = React.useCallback(() => {
+    errorResetBoundary.clearReset()
```
TkDodo, line 61: I don't think we can just call this _during_ render, as this has a side-effect. In `useBaseQuery.ts`, we call this in the `catch` of the `throw observer.fetchOptimistic`.

src/react/useQueries.ts
```diff
@@ -41,13 +55,29 @@ export function useQueries(queries: UseQueryOptions[]): UseQueryResult[] {
...
+    const unsubscribe = observerRef.current.subscribe()
+    throw observerRef.current
+      .refetch({ filter: x => x.getCurrentResult().isLoading })
+      .finally(unsubscribe)
```
TkDodo, line 65: should we not do something like `fetchOptimistic` does on the QueryObserver, just for multiple queries? We're also not calling `onSuccess / onError / onSettled` like we do in `useBaseQuery` ....

lubieowoce:
Hi! it's been a while since this PR saw any activity, and I'd like to offer some help in getting this over the finish line -- i *really* want to use `useQueries` with Suspense ;) @steevsachs would you be okay with me picking this up where you left off? Also, @TkDodo - assuming I could get this PR the implementation to a state everyone's happy with, do you see this as mergeable anytime soon? I've seen your comments in the [uSES thread](https://github.com/reactwg/react-18/discussions/86#discussioncomment-1561066), so i assume there's going to be a bunch of changes to the codebase in the near future. and I wouldn't want to go in and figure this one out only to have to re-do everything in a month :D

steevsachs:
@lubieowoce I appreciate the offer, I would really appreciate it if you could take this the rest of the way.

TkDodo:
@lubieowoce thank you for offering to take this over.

> do you see this as mergeable anytime soon? I've been keeping up with the React 18 stuff over at reactwg and saw your comments in the uSES thread indicating that you've got some changes in the pipeline (probably significant ones!).

from what I've seen so far, the uSES changes will _not_ interfere with what `suspense` is doing / trying to do. We will mainly switch out the the custom forceUpdate in useEffects with uSES. I'm working on a draft PR and can link it here once I have it.

also, I suspect that it will take quite some time until React 18 will be released, so yes, I think I would definitely merge this before 🚀

lubieowoce:
Alright, I'll get on it! :)

I've got one question re: intended behavior. Based on previous discussions i've seen:
- [useQueries should throw an error as soon as one of its queries errors](https://github.com/tannerlinsley/react-query/pull/1754#issuecomment-817265795)
- [[one query having suspense, while others don't] shouldn't be allowed](https://github.com/tannerlinsley/react-query/pull/2109#discussion_r659152561)

It seems as though, for a static array of queries (no adding new queries, no reordering), this should work as a "reference implementation" (tldr: call `useQuery` for each query, see if it errors, suspends, or returns, and either throw, suspend with a `Promise.all`, or return all the results):

```js
export const useQueriesSuspense = (queries) => {
  // Execute the queries "in parallel", see if any of them suspend or throw an error
  // (kind of like Promise.allSettled)
  const queryAttempts = queries.map((query) => {
    try {
      const value = useQuery({ ...query, suspense: true, useErrorBoundary: true });
      return { type: 'ready', value }
    } catch (thrown) {
      if (thrown instanceof Promise) {
        return { type: 'suspended', promise: thrown }
      } else {
        return { type: 'error', error: thrown }
      }
    }
  });

  const didError = (a) => a.type === 'error';
  const didSuspend = (a) => a.type === 'suspended';

  if (queryAttempts.some(didError)) {
    // Throw the first error
    throw queryAttempts.find(didError).error;

  } else if (queryAttempts.some(didSuspend)) {
    // Suspend until all the queries that suspended are ready
    throw Promise.all(
      queryAttempts.filter(didSuspend).map((a) => a.promise)
    );
      
  } else {
    // Everything's ready, return!
    return queryAttempts.map((a) => a.value: result);
  }
}
```

<details>
<summary>Note about concurrent rendering</summary>
Based on my (limited 😅) understanding this *may* not be concurrent-rendering safe:

  > If data for the requested key is already being fetched, the same *exact same Promise must be thrown*.
  https://github.com/facebook/react/issues/17526#issuecomment-769151686

so my thinking is, if a render got interrupted and this hook was called again in a concurrent render,
we'd throw a different promise and maybe (?) break something. But that's proabably not going to be a problem for quite a while.
</details>

@TkDodo  It'd be great to hear if this makes sense for you -- having a reference implementation to compare against would be great for testing/development.

TkDodo:
> for each query, do a separate useQuery call in a try-catch & see if it errors, suspends, or returns

I don't think this is how useQueries is implemented. There is a separate `QueriesObserver` that takes care of everything - `useQueries` doesn't call `useQuery` at all.

What you have stated about the intended behaviour is correct though. We can't really have one query suspend while others should _not_ suspend. This is why we've agreed that we'll just check if any of the queries uses suspense to see if all of them should suspend.

lubieowoce:
oh yeah, i know it's not implemented that way :D i've already played around with a real implementation (based on this PR), i just wanted to make sure that's behaviorally equivalent to what we want (for the subset of cases where it works)

TkDodo:
@lubieowoce we have a proposal for a potential re-design of the `useQueries` api with v4 that might make the things we've discussed here (suspending when any of the queries suspends) more explicit. Please have  a look here
- https://github.com/tannerlinsley/react-query/issues/2923

steevsachs:
This is very stale, and afaik this feature has been fixed in latest versions.  Closing : )

PR #2110, opened 2021-04-11
fix(query): add isStale to QueryState type

DamianOsipiuk:
While writing Vue binding for this library I stumbled upon typing error.
Devtools use a property that is not available on types exported by the lib:
https://github.com/tannerlinsley/react-query/blob/master/src/devtools/utils.ts#L13

This generates an error:
https://github.com/DamianOsipiuk/vue-react-query/blob/main/src/devtools/utils.ts#L15

Therefore adding this property to the types

### DIFF
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -48,6 +48,7 @@ export interface QueryState<TData = unknown, TError = unknown> {
   isFetching: boolean
   isInvalidated: boolean
   isPaused: boolean
+  isStale: boolean
   status: QueryStatus
 }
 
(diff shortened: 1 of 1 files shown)


TkDodo:
@DamianOsipiuk I think this is just an issue with the devtools - I don't think that flag really exists at runtime, and just adding it to the types won't magically make it work 😅 . See here:

Screenshot 2021-04-12 at 08 06 34

I also don't know why `isStale` _shouldn't_  be a function on a query, because it definitely is, see here: https://github.com/tannerlinsley/react-query/blob/e42cbc32dfcd9add24cadd06135e42af1dbbc8ad/src/core/query.ts#L271-L277

so the better fix would be to remove the check if `query.isStale` is really a function. The problem is that the devtools were migrated from js and are not really type-checked at the moment.

DamianOsipiuk:
Oh, in that case, it has to be some migration leftover. I will open a PR to remove it.
Thanks :wink:

PR #2113, opened 2021-04-12
Feature/1936 no queryclient set in devtools

### LINKED ISSUE #1936: "No QueryClient set" when upgrading from 3.12.0 to 3.12.1
**Describe the bug**

When upgrading from `react-query@3.12.0` to `react-query@3.12.1`, getting the following error in the browser console:

```
QueryClientProvider.js:36 Uncaught (in promise) Error: No QueryClient set, use QueryClientProvider to set one
    at useQueryClient2 (QueryClientProvider.js:36)
    at ReactQueryDevtoolsPanel2 (devtools.js:267)
```

**To Reproduce**

```
git clone git@gist.github.com:5ae9c1e18b571bafe17da22775b28329.git react-query-1936
cd react-query-1936
npm install
npm run dev
```

Open the dev server in the browser. You should see a blank page and observe the following error in the browser console:
```
QueryClientProvider.js:36 Uncaught (in promise) Error: No QueryClient set, use QueryClientProvider to set one
    at useQueryClient2 (QueryClientProvider.js:36)
    at ReactQueryDevtoolsPanel2 (devtools.js:267)
```

Now change the version of `react-query` in `package.json` to `3.12.0`, run `npm install` and re-start the dev server (interrupt `npm run dev` and run it again). You should now see `App` on the page and no error in the console.

TkDodo:
closes #1936


Ephem:
Nice!

I have no idea if there is any extra build config that is needed to support this as well. When I went back to my original PR for hydration, I saw there was some external and globals-config stuff in [the rollup config](https://github.com/tannerlinsley/react-query/pull/728/files#diff-6814bf77564b4f1c92f5861e184e28fe217c080a047fefa8b73a728f755ec45cR11) that might be relevant.

I haven't checked if that's the case, but might be worth double checking if this PR doesn't bundle the entire react-query into the devtools package. 😄

TkDodo:
so I downloaded: https://pkg.csb.dev/tannerlinsley/react-query/commit/6a704898/react-query

and it doesn't look like the devtools package contains a copy of react-query - it is also the same size as before.

The rollup config changed significantly with v3 and there is now no globals-config anymore

do you think we can merge this? Is the `eslint-disable` acceptable?

Ephem:
That's good! This warning when building doesn't look great though:

image

According to [this article (now pretty old)](https://www.mixmax.com/engineering/rollup-externals/) Rollup tries to add external and globals for you automatically for unresolved dependencies, but prints a warning. In this case, it seems as if it's guessing the wrong name for the global name though (`reactQuery` instead of `ReactQuery`). I think this means this PR can break in some cases (when loading React Query/devtools from a CDN?).

Since it's a one line change and needs the exact same fixes, maybe you'd like to fix the relative import in the hydration package in this PR as well? In that case, I could verify that it works, otherwise I'll open a separate PR for that. 😄

Ephem:
I think maybe a rollup config like this would work and be flexible to fix the relative imports that exists in other packages as well?

```js
import babel from 'rollup-plugin-babel'
import { terser } from 'rollup-plugin-terser'
import size from 'rollup-plugin-size'
import externalDeps from 'rollup-plugin-peer-deps-external'
import resolve from 'rollup-plugin-node-resolve'
import commonJS from 'rollup-plugin-commonjs'
import visualizer from 'rollup-plugin-visualizer'
import replace from '@rollup/plugin-replace'

const inputSrcs = [
  ['src/index.ts', 'ReactQuery', 'react-query'],
  ['src/core/index.ts', 'ReactQueryCore', 'react-query-core'],
  [
    'src/devtools/index.ts',
    'ReactQueryDevtools',
    'react-query-devtools',
    {
      extraExternals: ['react-query'],
      extraGlobals: { 'react-query': 'ReactQuery' },
    },
  ],
  [
    'src/hydration/index.ts',
    'ReactQueryHydration',
    'react-query-hydration',
    {
      extraExternals: ['react-query'],
      extraGlobals: { 'react-query': 'ReactQuery' },
    },
  ],
  [
    'src/persistQueryClient-experimental/index.ts',
    'ReactQueryPersistQueryClientExperimental',
    'persistQueryClient-experimental',
  ],
  [
    'src/createLocalStoragePersistor-experimental/index.ts',
    'ReactQueryCreateLocalStoragePersistorExperimental',
    'createLocalStoragePersistor-experimental',
  ],
  [
    'src/broadcastQueryClient-experimental/index.ts',
    'ReactQueryBroadcastQueryClientExperimental',
    'broadcastQueryClient-experimental',
  ],
]

const extensions = ['.js', '.jsx', '.es6', '.es', '.mjs', '.ts', '.tsx']
const babelConfig = { extensions, runtimeHelpers: true }
const resolveConfig = { extensions }

export default inputSrcs
  .map(
    ([input, name, file, { extraExternals = [], extraGlobals = {} } = {}]) => {
      const external = ['react', 'react-dom', ...extraExternals]

      const globals = {
        react: 'React',
        'react-dom': 'ReactDOM',
        ...extraGlobals,
      }

      return [
        {
          input: input,
          output: {
            name,
            file: `dist/${file}.development.js`,
            format: 'umd',
            sourcemap: true,
            globals,
          },
          external,
          plugins: [
            resolve(resolveConfig),
            babel(babelConfig),
            commonJS(),
            externalDeps(),
          ],
        },
        {
          input: input,
          output: {
            name,
            file: `dist/${file}.production.min.js`,
            format: 'umd',
            sourcemap: true,
            globals,
          },
          external,
          plugins: [
            replace({
              'process.env.NODE_ENV': `"production"`,
              delimiters: ['', ''],
            }),
            resolve(resolveConfig),
            babel(babelConfig),
            commonJS(),
            externalDeps(),
            terser(),
            size(),
            visualizer({
              filename: 'stats-react.json',
              json: true,
            }),
          ],
        },
      ]
    }
  )
  .flat()
```

TkDodo:
> Since it's a one line change and needs the exact same fixes, maybe you'd like to fix the relative import in the hydration package in this PR as well? In that case, I could verify that it works, otherwise I'll open a separate PR for that. 😄

let's do a separate PR as a follow up please. from what I can see, hydration imports multiple things relatively ...

Ephem:
> > Since it's a one line change and needs the exact same fixes, maybe you'd like to fix the relative import in the hydration package in this PR as well? In that case, I could verify that it works, otherwise I'll open a separate PR for that. 😄
> 
> let's do a separate PR as a follow up please. from what I can see, hydration imports multiple things relatively ...

Of course! I'll take care of that tomorrow. 😄

.eslintrc
```diff
@@ -32,5 +32,13 @@
     "import/no-unresolved": "error",
     "import/no-unused-modules": ["off", { "unusedExports": true }],
     "no-redeclare": "off"
-  }
+  },
+  "overrides": [
+    {
+      "files": ["src/devtools/**/*.{ts,tsx}"],
+      "rules": {
+        "import/no-unresolved": ["error", { "ignore": ["react-query"] }]
```
Ephem, line 40: NIT: Since we added `react-query` as an external and global for all entry points, I think it makes sense to configure this ignore to be global instead of a devtools-specific override as well? I definitely don't have strong feelings on it though.

Ephem (approved):
I'm not familiar enough with the devtools to verify the fix itself but saw that was done in the issue, so this looks good to me! 💯

.eslintrc
TkDodo: I just thought that maybe it's  a good idea to avoid wrong auto-imports or so, but being in line with the rollup config is also valuable: https://github.com/tannerlinsley/react-query/pull/2113/commits/7398b24819eaf263aae42f2eeaaa2aebe16ec8b8

uff, and the missing comma 🤦 
https://github.com/tannerlinsley/react-query/pull/2113/commits/1e18f1cab0bb9945ef9ddbd069f8d75bfb689a78

PR #2119, opened 2021-04-13
Fix useInfiniteQuery suspense bug

flybayer:
Closes: https://github.com/tannerlinsley/react-query/issues/2095

This fixes the useInfiniteQuery suspense bug where `options.behavior` was missing and so `infiniteQueryBehavior.onFetch` was not being called, resulting incorrect query `data`.

### DIFF
--- a/src/react/tests/suspense.test.tsx
+++ b/src/react/tests/suspense.test.tsx
@@ -72,19 +72,20 @@ describe("useQuery's in Suspense mode", () => {
     const states: UseInfiniteQueryResult<number>[] = []
 
     function Page() {
+      const [multiplier, setMultiplier] = React.useState(1)
       const state = useInfiniteQuery(
-        key,
-        ({ pageParam = 0 }) => Number(pageParam),
+        `${key}_${multiplier}`,
+        ({ pageParam = 1 }) => Number(pageParam * multiplier),
         {
           suspense: true,
           getNextPageParam: lastPage => lastPage + 1,
         }
       )
       states.push(state)
-      return null
+      return <button onClick={() => setMultiplier(2)}>next</button>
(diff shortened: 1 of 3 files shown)


src/react/useBaseQuery.ts
```diff
@@ -81,7 +81,7 @@ export function useBaseQuery<
     >(queryClient, defaultedOptions)
   }
 
-  let result = obsRef.current.getOptimisticResult(defaultedOptions)
+  let result = obsRef.current.getOptimisticResult(obsRef.current.options)
```
flybayer, line 84: This is the fix

src/react/tests/suspense.test.tsx
```diff
@@ -95,7 +96,16 @@ describe("useQuery's in Suspense mode", () => {
...
+      data: { pages: [1], pageParams: [undefined] },
+      status: 'success',
+    })
+
+    fireEvent.click(rendered.getByText('next'))
+    await sleep(10)
+
+    expect(states.length).toBe(3)
+    expect(states[2]).toMatchObject({
+      data: { pages: [2], pageParams: [undefined] },
```
flybayer, line 108: It was surprising to me that states here is 3 instead of 2, but I'm assuming this is expected.

boschni:
Awesome! The behavior was indeed missing but should be added in InfiniteQueryObserver.getOptimisticResult so it will also be fixed for other frameworks

flybayer:
@boschni thanks! I pushed a commit that makes that change. Tests pass, but I had to do a type cast. Not sure if that's ok? If not, can you give some guidance on how to fix?

boschni:
lgtm!

PR #2121, opened 2021-04-14
Fix "No queryClient set" when hydrating

### LINKED ISSUE #2112: Hydration is broken, "No QueryClient set"
**Describe the bug**

Hydration is broken. This can also be seen in the Next.js-example:

> No QueryClient set, use QueryClientProvider to set one

The `/hydration`-package has at some point (v3 release?) switched from `import { useQueryClient } from 'react-query'` to `import { useQueryClient } from '../react'`. This bundles the `useQueryClient` code into the hydration package itself, which also means a _separate_ `QueryClientSharingContext` gets created there. This is line 124 in `dist/react-query-hydration.development.js`:

```
var QueryClientSharingContext = /*#__PURE__*/React.createContext(false);
```

Because this context is different from the one `QueryClientProvider` uses, the hydration package can't find a queryClient to work with.

I plan to fix this but wanted to document the bug as soon as I found it.

Ephem:
Closes #2112 

This PR changes the relative import of `useQueryClient` to an absolute one (`from 'react-query'`). As described in the issue, this is important to avoid bundling parts of the main package into the hydration entry point and fixes a bug caused by the `QueryClientProvider`-context not lining up between the different entry points.

I've verified this fixes the broken NextJS-example as well as our code base at work.

Also added TS-resolves for this path so TS knows where to look for the types.

There are more relative imports in the hydration package, but they are all related to types which is fine.

### DIFF
--- a/src/hydration/react.tsx
+++ b/src/hydration/react.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 
-import { useQueryClient } from '../react'
+import { useQueryClient } from 'react-query'
 import { hydrate, HydrateOptions } from './hydration'
 
 export function useHydrate(state: unknown, options?: HydrateOptions) {
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -13,7 +13,10 @@
     "noUnusedParameters": true,
     "skipLibCheck": true,
     "strict": true,
-    "types": ["jest"]
+    "types": ["jest"],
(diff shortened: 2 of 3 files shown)


TkDodo (approved):

Ephem:
Doing some extra verification on our project at work before I merge this.

PR #2130, opened 2021-04-15
feat(query): add 'getObserversCount' util

DamianOsipiuk:
Adding 'getObserversCount' util

Reson for making this change:

The [observers](https://github.com/tannerlinsley/react-query/blob/master/src/core/query.ts#L159) on the query is private property, therefore Typescript is complaining when trying to use it.

DevTools is using the length of the observers in a couple of places like [here](https://github.com/tannerlinsley/react-query/blob/master/src/devtools/utils.ts#L23)

Since this property should probably stay as private, I have added utility to get the observers count as a getter method.

### DIFF
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -342,6 +342,10 @@ export class Query<
     }
   }
 
+  getObserversCount(): number {
+    return this.observers.length
+  }
+
   invalidate(): void {
     if (!this.state.isInvalidated) {
       this.dispatch({ type: 'invalidate' })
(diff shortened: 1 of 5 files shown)


DamianOsipiuk:
@TkDodo Hi, any chance this could be merged? :wink:

TkDodo:
> DevTools is using the length of the observers in a couple of places like here

sure, can you maybe use new function then in the devtools and replace the private accesses with it?

DamianOsipiuk:
@TkDodo Sure :+1:

TkDodo:
it's also used in tests with ts-expects-error as far as I can see :)

TkDodo (approved):

PR #2131, opened 2021-04-15
Added hydrate and dehydrate options to the persistQueryClient function

gabrielbull:
The reason to add this is to be able to customize the hydrate/dehydrate behaviour on the persistQueryClient. For instance, I don't want a network error to remove a cache entry from my persisted state, I want the client cache to be equivalent to the persisted cache at all time, with dehydrate options, I can achieve this behaviour.


makirby:
Was just about to come and add this exact PR!

We're running in prod with our own modified persistQueryClient at the moment would be great to see this merged.

TkDodo (approved):

PR #2152, opened 2021-04-20
fix: focus handler typings

fmauquie:
Typings did not allow passing a function taking `setFocused` to addEventListener

Fixing by using the same typings as in `onlineManager.ts`, tweaked to allow default Window.Event behavior

This allows setting the React Native focus handler like this:
```ts
focusManager.setEventListener((setFocused) => {
  const listener = (status: AppStateStatus) => setFocused(status === "active");
  AppState.addEventListener("change", listener);
  return () => AppState.removeEventListener("change", listener);
});
```

### DIFF
--- a/src/core/focusManager.ts
+++ b/src/core/focusManager.ts
@@ -12,12 +12,12 @@ class FocusManager extends Subscribable {
   }
 
   setEventListener(
-    setup: (onFocus: () => void) => (focused?: boolean) => void
+    setup: (setFocused: (focused?: boolean) => void) => () => void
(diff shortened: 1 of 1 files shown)


TkDodo:
seems like I fixed this for the `onlineManager` here: https://github.com/tannerlinsley/react-query/pull/2053/files

but forgot about this one 🤦

fmauquie:
Your approach is slightly different, it better captures the intention of the method IMO. I'm modifying my commit to match.

PR #2155, opened 2021-04-20
Highlighting change of the default config key.

jonathonhawkins92:
While migrating I failed to spot this change and almost caused a prod bug, so I'm proposing this change to avoid anyone else also doing so 👍


TkDodo:
> almost caused a prod bug

can't help but say "TypeScript" :)

jonathonhawkins92:
> > almost caused a prod bug
> 
> can't help but say "TypeScript" :)

I know I know, we are in the process of migrating 😄

PR #2183, opened 2021-04-23
feat(queryClient): return QueryClient instance from mount function

DamianOsipiuk:
Return QueryClient from mount function to allow to chain them
This is a convenience solution for a vue-query package, where user has to call mount manually

### DIFF
--- a/src/core/queryClient.ts
+++ b/src/core/queryClient.ts
@@ -72,7 +72,7 @@ export class QueryClient {
     this.mutationDefaults = []
   }
 
-  mount(): void {
+  mount(): QueryClient {
     this.unsubscribeFocus = focusManager.subscribe(() => {
       if (focusManager.isFocused() && onlineManager.isOnline()) {
         this.mutationCache.onFocus()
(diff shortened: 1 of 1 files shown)


DamianOsipiuk:
@TkDodo What do you think?

wintorez (approved):

DamianOsipiuk:
For example instead of this
```ts
import { createApp } from "vue";
import { QueryClient, VUE_QUERY_CLIENT } from "vue-query";
import App from "./App.vue";

queryClient.mount();

createApp(App).provide(VUE_QUERY_CLIENT, queryClient).mount("#app");
```

You will be able to do something like this:
```ts
import { createApp } from "vue";
import { QueryClient, VUE_QUERY_CLIENT } from "vue-query";
import App from "./App.vue";

createApp(App).provide(VUE_QUERY_CLIENT, new QueryClient().mount()).mount("#app");
```

But as mentioned it's only for convinience. Nothing important.

tannerlinsley:
I think for simplicity's sake here, I'm not going to merge this in. I get why it would be more convenient, but it would lock in the return type of mount for every one else without really any benefit. Instead, I would rather reserve the return type for something that may come in the future that would affect most users. 

Thanks for the input though! :)

PR #2186, opened 2021-04-23
feat(useQuery): Add option for strict query keys evaluation

rametta:
Related to this discussion: https://github.com/tannerlinsley/react-query/discussions/2169

### DIFF
--- a/src/core/queryObserver.ts
+++ b/src/core/queryObserver.ts
@@ -141,6 +141,13 @@ export class QueryObserver<
     const prevOptions = this.options
     const prevQuery = this.currentQuery
 
+    if (options?.strict && Array.isArray(options?.queryKey)) {
+      const isInvalidKey = options.queryKey.some(key => key === undefined || key === null)
+      if (isInvalidKey) {
+        options.enabled = false
+      }
+    }
+
     this.options = this.client.defaultQueryObserverOptions(options)
 
     if (
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -98,6 +98,13 @@ export interface QueryObserverOptions<
    */
   enabled?: boolean
   /**
+   * Set this to `true` to have the queryKey checked for null and undefined values.
+   * If there are any null or undefined values in the queryKey, the `enabled` option
+   * will automatically be set to `false`.
+   * Defaults to `false`.
+   */
+  strict?: boolean
+  /**
    * The time in milliseconds after data is considered stale.
    * If set to `Infinity`, the data will never be considered stale.
    */
--- a/src/react/tests/useQuery.test.tsx
+++ b/src/react/tests/useQuery.test.tsx
@@ -2226,6 +2226,39 @@ describe('useQuery', () => {
     await waitFor(() => rendered.getByText('new'))
   })
 
+  it('should start with status idle if strict is strue and queryKey contains null or undefined', async () => {
+    const key1 = queryKey()
+    const key2 = queryKey()
+    const key3 = queryKey()
+
+    function Page() {
+      const first = useQuery([key1, null], () => 'data', {
+        strict: true,
+      })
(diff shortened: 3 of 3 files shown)


markmssd (approved):
Much needed! 🙌🏼

TkDodo:
Personally, I don't see advantage of this flag, especially because overriding _another_ flag is not a good api imo. It leads to impossible states:

```
strict: true,
enabled: true
```

You now have to know that `strict` takes precedence over `enabled`. A better api would be to set `enabled: 'strict' `, but given that it's a one-liner in user-land, I'd not add this:

```
enabled: queryKey.every(Boolean)
```

(not 100% the same, I know, but you get the picture)

I think the explicitness here would be preferred the "magic", and a custom `useStrictQuery` hook seems like a good solution if you want this.

Happy to wait fo other maintainer's opinions. If we feel that it doesn't do any harm and is a good abstraction, we can also merge it, I have no hard opinion here :)

rametta:
Hey @TkDodo,

Thanks for the feedback! I definitely agree having one flag override another flag is probably not the most elegant solution.

I think the ideal solution for me would be able to set a global option in the queryClient to have `strict: true`. The `enabled: 'strict'` is a good idea on a single query call but I don't think would work if it was a global flag in the queryClient.

If anyone else agrees with the global option idea, I will be happy to change up my PR to support that instead

TkDodo:
> I think the ideal solution for me would be able to set a global option in the queryClient to have strict: true. The enabled: 'strict' is a good idea on a single query call but I don't think would work if it was a global flag in the queryClient.

why would it not work on a global level? There is basically no difference between the settings on global level, the settings set via `queryClient.setQueryDefaults` and the options set on an observer level. They have the same type, and react-query resolves options from the bottom-up.

if you set `strict:true` on a global level, the api becomes even more weird, because what happens if I say `enabled: someCondition` on an observer then? The strict still overrides my condition, and unless I know that `strict` is set somewhere higher up the tree, I will probably wonder why this isn't working as expected...

rametta:
You're right! so it already works as a global flag, that's great!
I wonder if there's a way we can name it so it doesn't cause any confusion

rametta:
Since the options are resolved bottom-up, I think it makes sense that you can override the enabled option from the query call itself. That logic sort of applies for any option right?

TkDodo:
yes, you can. I was referring to the following example:

```
const queryClient = new QueryClient({ queries: { defaultOptions: { strict: true }}})

....

useQuery(key, fn, { enabled: someCondition }
```

now I have strict: true globally, and if my condition evaluates to true, strict will still take precedence.

You will likely not use strict and enabled together, so from an api design standpoint, having two separate options seems wrong.

rametta:
Yeah I agree. So maybe the `enabled: 'strict'` route like you suggested would be a better solution. It would be more clear for users when their options are getting overridden or not

rametta:
What do you think about this idea: when the `enabled: 'strict'` option set, we can `throw` when they queryKey is invalid.
Like:
```ts
if (options.enabled === 'strict' && isInavlidQueryKey) {
  throw new Error('Enabled option is set to strict. Query key must not contain null or undefined')
}
```

rametta:
@tannerlinsley what do you think about this proposal?

TkDodo:
> What do you think about this idea: when the enabled: 'strict' option set, we can throw when they queryKey is invalid.

what would be the advantage of that? I thought your idea was to disable queries that have `null` in the query key, now the new idea seems to be to throw at runtime if that happens. Is having `null` in the queryKey an invariant for you, like a "this-should-never-happen" case?

I'm gonna close this since I'm not seeing the value of this feature and there isn't any additional input from other maintainers. It can also be easily achieved in user-land. We can continue to discuss it of course, and if there's further input we can also re-open it.

PR #2200, opened 2021-04-27
fix(useQuery): make sure queryKeys are always an array at runtime when passed to the queryFn

### LINKED ISSUE #2149: Can't pass a string value as the first parameter of `useQuery()`
**Describe the bug**
When passing a string as the first parameter of `useQuery()`, it doesn't get passed into the `defaultQueryFn` correctly.

**To Reproduce**
1. Create a `defaultQueryFn` for your `QueryClient`
2. Provider the `QueryClient` using `QueryClientProvider`.
3. Try and run a query with `useQuery('my-endpoint')`
4. See incorrect params being passed to `defaultQueryFn`

**Expected behavior**
The `queryKey` passed to the `defaultQueryFn` should be in the format `[key, params]`, no matter if `useQuery()` has been passed an array, string or object as the first parameter.

```js
const defaultQueryFn = async ({ queryKey }) => {
  const [key, params = {}] = queryKey;

  // should equal { key: 'my-endpoint', params: {} }
  // actually equals { key: 'm', params: y} }
  console.log({ key, params })
});

//

const query = useQuery('my-endpoint');
```

TkDodo:
fixes #2149 

this is a regression from https://github.com/tannerlinsley/react-query/pull/2074

the real culprit we tried to fix was typings being still just TQueryKey in the FetchContext, which is not what it is at runtime, because if your QueryKey is a string, it will be [string] there.

See: https://github.com/tannerlinsley/react-query/pull/2074/files#r606362656 

this fix re-adds the `ensureArray` call, but makes it narrow the type (compromised type assertion) so that the FetchContext will receive an `EnsuredQueryKey`, which is always an Array


src/core/utils.ts
```diff
@@ -88,8 +89,12 @@ export function isValidTimeout(value: any): value is number {
...
 }
 
-export function ensureArray<T>(value: T | T[]): T[] {
-  return Array.isArray(value) ? value : [value]
+export function ensureQueryKeyArray<T extends QueryKey>(
+  value: T
+): EnsuredQueryKey<T> {
+  return (Array.isArray(value)
+    ? value
+    : ([value] as unknown)) as EnsuredQueryKey<T>
```
TkDodo, line 97: couldn't get the type to work any other way, I think `Array.isArray` is not smart enough. Do you know a better way @boschni ?

tannerlinsley (approved):

PR #2204, opened 2021-04-28
feat(queryClient): add setQueriesData utility

TkDodo:


src/core/queryClient.ts
```diff
@@ -123,6 +123,31 @@ export class QueryClient {
...
+    queryKey: QueryKey,
+    updater: Updater<TData | undefined, TData>,
+    options?: SetDataOptions
+  ): [QueryKey, TData][]
+
+  setQueriesData<TData>(
+    filters: QueryFilters,
+    updater: Updater<TData | undefined, TData>,
+    options?: SetDataOptions
+  ): [QueryKey, TData][]
```
TkDodo, line 136: not sure if this is the best return type - wanted to provide a way to show which key actually had data set, as calling `setQueriesData` on a filter that doesn't match anything will not set anything

src/core/queryClient.ts
```diff
@@ -123,6 +123,31 @@ export class QueryClient {
...
+  setQueriesData<TData>(
+    queryKeyOrFilters: QueryKey | QueryFilters,
+    updater: Updater<TData | undefined, TData>,
+    options?: SetDataOptions
+  ): [QueryKey, TData][] {
+    return this.getQueryCache()
+      .findAll(queryKeyOrFilters)
+      .map(({ queryKey }) => [
+        queryKey,
+        this.setQueryData<TData>(queryKey, updater, options),
```
TkDodo, line 147: @boschni I'm a bit worried that calling `setQueryData` multiple times in a row triggers multiple re-renders, and it does. see this example, where onButtonClick, I call `setQueryData` three times for three queries, and even though it's synchronous, it will trigger 3 re-renders:

https://codesandbox.io/s/tannerlinsleyreact-query-basic-forked-5c6rs?file=/src/index.js

I tried to wrap it in `notifyManager.batchCalls`, but it didn't change anything.

This is not much to do with this PR, but because I'm now calling setQueryData in a loop, it might show :)

Do you have any idea how to batch these calls maybe?

src/core/queryClient.ts
```diff
@@ -335,8 +402,15 @@ export class QueryClient {
...
-    arg1: TQueryKey | FetchInfiniteQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+  fetchInfiniteQuery<
+    TQueryFnData,
+    TError,
+    TData = TQueryFnData,
+    TQueryKey extends QueryKey = QueryKey
+  >(
+    arg1:
+      | TQueryKey
+      | FetchInfiniteQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
```
TkDodo, line 413: not sure why prettier re-formatted all of this all of a sudden. looks like we are not checking prettier formatting as part of the pipeline 🤔 .

I will make a second PR that adds prettier formatting to the pipeline to avoid unnecessary diffs.

src/core/queryClient.ts
```diff
@@ -123,6 +123,31 @@ export class QueryClient {
       .setData(updater, options)
   }
 
+  setQueriesData<TData>(
```
TkDodo, line 126: since we are not creating queries like with `setQueryData`, because we only update matching keys, would `updateQueries` or `updateQueriesData` be a better name?

this was also mentioned by @tannerlinsley here: https://github.com/tannerlinsley/react-query/issues/135#issuecomment-589710815

but I don't think it made it?

src/core/queryClient.ts
TkDodo: couldn't find this out, but is also not so important I guess. if we get the api of that function right, we can always optimize it internally later :)

src/core/queryClient.ts
```diff
@@ -123,6 +123,31 @@ export class QueryClient {
...
+    queryKey: QueryKey,
+    updater: Updater<TData | undefined, TData>,
+    options?: SetDataOptions
+  ): [QueryKey, TData][]
+
+  setQueriesData<TData>(
+    filters: QueryFilters,
```
TkDodo, line 133: I didn't create an overload that accepts queryKey _and_ filters, even though some other functions have this, because the `updater` is required so the signature would be very weird. I also don't think it's important because the `filters` also have the option to pass a `queryKey`.

src/core/queryClient.ts
boschni: Hi @TkDodo! The entire function body should be wrapped in `notifyManager.batch`, this should prevent redundant renders

TkDodo: thank you 🙏 , seems like I messed up `batchCalls` and `batch` 😅 . fixed it here: 95bed800

src/core/queryClient.ts
tannerlinsley: I think the current name is fine as long as people realize that it will not create queries.

PR #2213, opened 2021-04-29
refactor(devtools): move default theme to theme file, fix types

DamianOsipiuk:
Move the default theme to the proper file.
Fix types for `theme.tsx`
No functional changes.

### DIFF
--- a/src/devtools/devtools.tsx
+++ b/src/devtools/devtools.tsx
@@ -17,7 +17,7 @@ import {
   Select,
   ActiveQueryPanel,
 } from './styledComponents'
-import { ThemeProvider } from './theme'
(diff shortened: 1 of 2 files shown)


TkDodo:
have a look at the auto-deployed codesandboxes please - I can't see the devtools anymore :/

https://codesandbox.io/s/tannerlinsleyreact-query-basic-y5jg9

DamianOsipiuk:
That is very weird, as it worked fine when I was checking locally.
I will double-check locally, and if I would not be able to find the issue, then maybe deployment to codesandbox somehow got broken.

DamianOsipiuk:
@TkDodo Should be fixed now

PR #2237, opened 2021-05-04
Update types.ts

lew:
provide type string for first element in QueryKey

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -3,7 +3,7 @@ import type { QueryBehavior } from './query'
 import type { RetryValue, RetryDelayValue } from './retryer'
 import type { QueryFilters } from './utils'
 
-export type QueryKey = string | readonly unknown[]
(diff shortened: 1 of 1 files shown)


TkDodo (commented):
The first element in the array does not need to be a string

lew:
https://react-query.tanstack.com/guides/query-keys
> The simplest form of a key is actually not an array, but an individual string. When a string query key is passed, it is converted to an array internally with the string as the only item in the query key. […] When a query needs more information to uniquely describe its data, you can use an array **with a string** and any number of serializable objects to describe it.

So if I have a single non-string query key I can just make it an one element array?

`const result = useQuery(3)`
vs
`const result = useQuery([3])`

TkDodo:
seems like we should rather update the docs to match the types rather than the other way around. I see no technical reason why, when using an array, that the first element needs to be a string. It might be good practice (but so is _not_ using bare strings as keys imho), but we shouldn't enforce it on type level for no good reason

PR #2250, opened 2021-05-07
Update query-keys.md

chloetedder:


TkDodo:
The idea here is to show that different object insertion order does not change the query key serialization.

PR #2257, opened 2021-05-10
feat: resetQueries disable refetch option

verycosy:
I used resetQueries function when user logged out, but this function always refetching queries.
Someone like me, maybe don't want to refetch when call resetQueries.
 
So I added new option for resetQueries.

### DIFF
--- a/src/core/queryClient.ts
+++ b/src/core/queryClient.ts
@@ -179,7 +179,7 @@ export class QueryClient {
     arg1?: QueryKey | QueryFilters,
     arg2?: QueryFilters | ResetOptions,
     arg3?: ResetOptions
-  ): Promise<void> {
+  ): Promise<void> | void {
     const [filters, options] = parseFilterArgs(arg1, arg2, arg3)
     const queryCache = this.queryCache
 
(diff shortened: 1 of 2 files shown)


TkDodo:
> I used resetQueries function when user logged out

Are you sure you are not looking for `queryClient.removeQueries` ?

verycosy:
> > I used resetQueries function when user logged out
> 
> Are you sure you are not looking for `queryClient.removeQueries` ?

Yes. React-Query documents are very useful, so I know removeQueries also and I tried. 

I'm using this library in react-native with react-navigation.
I made a mistake react-navigation part. After fixed routing code, removeQueries works like a charm !
Before fixing, it seems didn't work. but my mistakes. sorry for that.

Do you think resetQueries don't need disable refetch option?
If so, Can I close this PR?

TkDodo:
> Do you think resetQueries don't need disable refetch option?
If so, Can I close this PR?

I think every additional option comes at the price of complexity, not only for maintainers but also for users. boolean options are always hard to grasp and can introduce impossible states quite easily. I've written about this here: https://tkdodo.eu/blog/no-love-for-boolean-parameters

If your problem is solved with `removeQueries`, I'd rather not merge this :)

PR #2275, opened 2021-05-13
fix: update matchQuery logic to cater for all permutations

### LINKED ISSUE #2271: invalidateQueries with refetchActive: false on inactive query refetches it
**Describe the bug**
Calling `queryClient.invalidateQueries` with `{ refetchActive: false }` on inactive query actually refetches it.

**To Reproduce**
Codesandbox: https://codesandbox.io/s/react-query-refetchactive-bug-n3kwc

With the following query cache state:
```
['queryA'] // inactive
['queryB'] // active
```

When I call 
```
// ❌ Incorrect, refetches 😵
queryClient.invalidateQueries(['queryA'], {
  refetchActive: false
});

// ✅ Correct, doesn't refetch
queryClient.invalidateQueries(['queryB'], {
  refetchActive: false
});
```

**Expected behavior**
Option `{ refetchActive: false }` on inactive query shouldn't trigger refetch

**Screenshots**

 - OS: macOS
 - Browser: chrome
 - Version: 90

**Additional context**
N.A

donysukardi:
Fixes #2271

### DIFF
--- a/src/core/queryClient.ts
+++ b/src/core/queryClient.ts
@@ -239,7 +239,7 @@ export class QueryClient {
     const refetchFilters: QueryFilters = {
       ...filters,
       active: filters.refetchActive ?? true,
-      inactive: filters.refetchInactive,
+      inactive: filters.refetchInactive ?? false,
     }
 
     return notifyManager.batch(() => {
--- a/src/core/queryClient.ts
+++ b/src/core/queryClient.ts
@@ -238,7 +238,9 @@ export class QueryClient {
 
     const refetchFilters: QueryFilters = {
       ...filters,
-      active: filters.refetchActive ?? true,
+      // if filters.refetchActive is not provided and filters.active is explicitly false,
+      // e.g. invalidateQueries({ active: false }), we don't want to refetch active queries
+      active: filters.refetchActive ?? filters.active ?? true,
       inactive: filters.refetchInactive ?? false,
     }
 
--- a/src/core/utils.ts
+++ b/src/core/utils.ts
@@ -68,6 +68,8 @@ export type Updater<TInput, TOutput> =
   | TOutput
   | DataUpdateFunction<TInput, TOutput>
 
+export type QueryStatusFilter = 'all' | 'active' | 'inactive' | 'none'
+
 // UTILS
 
 export const isServer = typeof window === 'undefined'
@@ -164,6 +166,25 @@ export function parseFilterArgs<
     : [arg1 || {}, arg2]) as [TFilters, TOptions]
 }
 
+export function mapQueryStatusFilter(
+  active?: boolean,
+  inactive?: boolean
+): QueryStatusFilter {
+  if (
+    (active === true && inactive === true) ||
+    (active == null && inactive == null)
+  ) {
+    return 'all'
+  } else if (active === false && inactive === false) {
+    return 'none'
+  } else {
+    // At this point, active|inactive can only be true|false or false|true
+    // so, when only one value is provided, the missing one has to be the negated value
+    const isActive = active ?? !inactive
+    return isActive ? 'active' : 'inactive'
+  }
+}
+
 export function matchQuery(
   filters: QueryFilters,
   query: Query<any, any, any, any>
@@ -188,16 +209,18 @@ export function matchQuery(
     }
   }
 
-  let isActive
+  const queryStatusFilter = mapQueryStatusFilter(active, inactive)
 
-  if (inactive === false || (active && !inactive)) {
-    isActive = true
-  } else if (active === false || (inactive && !active)) {
-    isActive = false
-  }
-
-  if (typeof isActive === 'boolean' && query.isActive() !== isActive) {
+  if (queryStatusFilter === 'none') {
     return false
+  } else if (queryStatusFilter !== 'all') {
+    const isActive = query.isActive()
+    if (queryStatusFilter === 'active' && !isActive) {
+      return false
+    }
+    if (queryStatusFilter === 'inactive' && isActive) {
+      return false
+    }
   }
 
   if (typeof stale === 'boolean' && query.isStale() !== stale) {
--- a/src/core/tests/queryClient.test.tsx
+++ b/src/core/tests/queryClient.test.tsx
@@ -789,6 +789,27 @@ describe('queryClient', () => {
       expect(queryFn1).toHaveBeenCalledTimes(2)
       expect(queryFn2).toHaveBeenCalledTimes(1)
     })
+
+    test('should not refetch active queries when "refetchActive" is not provided and "active" is false', async () => {
+      const key1 = queryKey()
+      const key2 = queryKey()
+      const queryFn1 = jest.fn()
+      const queryFn2 = jest.fn()
+      await queryClient.fetchQuery(key1, queryFn1)
+      await queryClient.fetchQuery(key2, queryFn2)
+      const observer = new QueryObserver(queryClient, {
+        queryKey: key1,
+        queryFn: queryFn1,
+        staleTime: Infinity,
+      })
+      const unsubscribe = observer.subscribe()
+      queryClient.invalidateQueries(key1, {
+        active: false,
+      })
+      unsubscribe()
+      expect(queryFn1).toHaveBeenCalledTimes(1)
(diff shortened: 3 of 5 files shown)


src/core/utils.ts
```diff
@@ -164,6 +166,25 @@ export function parseFilterArgs<
...
+  } else if (active === false && inactive === false) {
+    return 'none'
+  } else {
+    // At this point, active|inactive can only be true|false or false|true
+    // so, when only one value is provided, the missing one has to be the negated value
+    const isActive = active ?? !inactive
+    return isActive ? 'active' : 'inactive'
+  }
+}
+
```
donysukardi, line 187: @TkDodo here it is. This should cater for all the different permutations outlined in this test - https://github.com/tannerlinsley/react-query/pull/2275/files#diff-edc0f64b1261687e43194a6b6ec3881b7ece03d3d2c1236b2e087deae220fc82R311-R329

TkDodo: wow, thanks a lot 🚀 

what I find interesting is that it seems that the actual fix only involved coercing the default value to `false` for `invalidateQueries` here:

https://github.com/tannerlinsley/react-query/pull/2275/files#diff-8e0c93e2ffd9983864be13d7b233dfc6d757e2fc95dd8c7fbc83a1a71aec6cbeR242

does that mean the issue was only present for `invalidateQueries`, but not for other methods like `refetchQueries` , because they don't have this mapping from `refetchActive -> active` and `refetchInactive -> inactive` ?

if so, can we maybe add another test especially for `invalidateQueries` - possibly the one that would fail before, but works now ? (The tests you've added on the queryClient are all with `refetchQueries` as far as I can see)

src/core/queryClient.ts
```diff
@@ -239,7 +239,7 @@ export class QueryClient {
     const refetchFilters: QueryFilters = {
       ...filters,
       active: filters.refetchActive ?? true,
-      inactive: filters.refetchInactive,
+      inactive: filters.refetchInactive ?? false,
```
TkDodo, line 242: I found something slightly unrelated: `InvalidateQueryFilters` is defined as:

https://github.com/tannerlinsley/react-query/blob/3b2c4088e103e4b4ca990662c03fd6e67330b477/src/core/types.ts#L243-L246

`extends QueryFilters` means we can also pass `active` and `inactive` to it, but those inputs seem invalid, because they will actually be overridden with `refetchInactive` and `refetchInactive` here. I think a more correct type definition would be:

```diff
- export interface InvalidateQueryFilters extends QueryFilters {
+ export interface InvalidateQueryFilters extends Omit<QueryFilters, 'active' | 'inactive'> {
  refetchActive?: boolean
  refetchInactive?: boolean
}
```

what do you think?

src/core/utils.ts
donysukardi: Yes, that's one of the issues. `refetchInactive` should be defaulted to false.

Another issue is that the original code in `matchQuery` doesn't cater for scenario when both `active` and `inactive` are false, i.e. all queries shouldn't be matched.

Sure, I can add more test to cover for the failing scenario.

src/core/queryClient.ts
donysukardi: You're right. The new type is definitely more appropriate as the two values don't currently have any effect.

src/core/utils.ts
donysukardi: Added more tests.

Actually this newly refactored QueryFilterStatus can be introduced as breaking change in the future. If we do, we will need to exclude `none` option as it doesn't make sense to run a filter function expecting nothing to be returned at all time.

TkDodo: > In the interim, do you think we should bail out earlier from refetch when both active and inactive are false? Otherwise, it makes no sense to loop through this.queries.filter(query => matchQuery(filters, query)) when the result are always going to be []

while the assumption of having both active and inactive set to false _should_ yield empty array, it didn't do so with the previous implementation:

```
  let isActive

  if (inactive === false || (active && !inactive)) {
    isActive = true
  } else if (active === false || (inactive && !active)) {
    isActive = false
  }
```

here, `inactive` is checked first, and the `active` check was in an `else if`, so if you pass `{ inactive: false, active: false }`, you would still be matching all active queries 😅 

I still think this is a good change and can be seen as a fix 👍 . I would not do an early bailout because I consider this combination an invariant anyways. If we refactor this away from two boolean values later, we'll have less code to delete :)

donysukardi: I'm referring to invalidateQueries, when both `refetchActive` and `refetchInactive` are set to `false` (likely as the result of setting `refetchActive` to false, because the other one is defaulted to false anyway).

This is how it looks like if we bails out from refetching,

```ts
// In invalidateQueries

return notifyManager.batch(() => {
  this.queryCache.findAll(filters).forEach(query => {
    query.invalidate()
  })
  
  const { refetchActive = true, refetchedInactive = false } = filters
  if(!refetchActive && !refetchInactive) {
    return;
  } else {
    const refetchFilters: QueryFilters = {
      ...filters,
      active: filters.refetchActive,
      inactive: filters.refetchInactive,
    }
    return this.refetchQueries(refetchFilters, options)
  }
})

```

src/core/queryClient.ts
TkDodo: I think what I said here is actually wrong. The `refetchFilters` we are building here are only used for the call to `refetchQueries` - for the actual invalidation, we use the `filters` passed to `invalidateQueries`.

That means the previous interface is actually correct, because you can pass `active` / `inactive`, and it will affect the invalidation, and you can pass `refetchActive` / `refetchInactive`, which will only affect the invalidation - sorry for that. It's curious that we don't have test for this behaviour of `invalidateQueries` :/

What I'm just wondering now too is this:

If I call `queryClient.invalidateQueries({ active: false })`, active queries will not be invalidated. But they will still be refetched because `refetchInactive` is not passed, and it falls back to `true`. So shouldn't this mapping rather read something like:

```diff
- active: filters.refetchActive ?? true,
+ active: filters.refetchActive ?? filters.active ?? true, 
```

?

src/core/utils.ts
TkDodo: please see my other comment here: https://github.com/tannerlinsley/react-query/pull/2275#discussion_r659211920

I don't think this would work if we also choose to fallback to `filters.active` / `filters.inactive` for the `refetchFilters`.

src/core/queryClient.ts
donysukardi: I see. So, we'll make it such that if active queries don't get invalidated, they don't get refetched either. Correct?

donysukardi: would this be better? explicitly checking if `filters.active === false` better conveys the intention of the code.

```ts
const refetchFilters: QueryFilters = {
  ...filters,
  // if filters.refetchActive is not provided and filters.active is explicitly false,
  // we don't want to refetch active queries
  active: filters.refetchActive ?? (filters.active === false ? false : true),
  inactive: filters.refetchInactive ?? false,
}
```

TkDodo: I think this is the same as my above code with the double ??:

```
active: filters.refetchActive ?? filters.active ?? true,
```

- first, we take `refetchActive` if provided.
- if not, we take `active` if provided
- if this is still `undefined`, we take `true`

> So, we'll make it such that if active queries don't get invalidated, they don't get refetched either. Correct?

correct. if you call `queryClient.invalidateQueries({ active: false })`, maybe with some more filters, then we would now currently still refetch those active queries, while with the new code, we would take `active:false` for `refetchActive` because it was supplied. with this, you could even _not_ invalidate, _but still_ refetch, with:

```
queryClient.refetchQueries({ active: false, refetchActive: true })
```

not sure if that makes any sense, but it's what the api allows :)

donysukardi: yes, it achieves the same effect, but i thought making more explicit assertion might be better. i'm ok with your suggestion as well, since i'll also be adding some comment.

i'll add some test cases to cover the scenario too.

TkDodo: nice, thank you 🙏

donysukardi: done 😄

TkDodo:
I think this looks good 👍

PR #2282, opened 2021-05-14
feat: session storage persistor

CurtisHughes:
Adds session storage persistor as an experimental feature (mirrors the logic for the local storage persistor).


TkDodo:
Since `localstorage` and `sessionStorage` have literally the same api - thee [Web Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API), maybe we should just rename `localStoragePersistor` to `webStoragePersistor` and allow to pass in if you want `window.localStorage` or `window.sessionStorage` ?

What do you think @tannerlinsley ?

CurtisHughes:
That feels like a clean solution. I can attempt those changes if that is the direction you all want to go.

TkDodo:
it's experimental so I think we can justify the rename. I think we should add a `storage: Storage` prop to the createPersistor function, which could then be either `window.localStorage`, `window.sessionStorage` or maybe something else that I am not even aware of.

CurtisHughes:
Ok @TkDodo, I took a stab at it. Changed `createLocalStoragePersitor` -> `createWebStoragePersistor` and added a required property `storage: Storage`. Let me know if that looks like what you had in mind.

TkDodo:
I think this looks good 👍 . Let's wait if @tannerlinsley supports the re-name.

tannerlinsley (approved):
💯 % approve.

PR #2329, opened 2021-06-01
docs: mutations example for custom typescript hooks

RipDevil:
docs: mutations example for custom typescript hooks

There should be also an example for Mutations custom hook definition.

### DIFF
--- a/docs/src/pages/typescript.md
+++ b/docs/src/pages/typescript.md
@@ -28,6 +28,6 @@ function useGroups() {
 
 ```js
 function useAddGroup() {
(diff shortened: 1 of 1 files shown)


TkDodo:
Thanks for the PR. I plan on re-writing the typescript docs soon to be more extensive. Generally, I'd like to emphasize that passing Generics is not the best possible solution and mostly only needed if you plan to make your own generics abstractions over useQuery / useMutation. Letting TypeScript do it's thing with type inference is the better approach, and I'll try to reflect that in the re-write. So not going to merge this as of now.

PR #2342, opened 2021-06-04
Use higher order function (array) in code example

41y08h:


docs/src/pages/guides/infinite-queries.md
```diff
@@ -158,11 +158,10 @@ queryClient.setQueryData('projects', data => ({
...
 
 ```js
-const newPagesArray = []
-oldPagesArray?.pages.forEach(page => {
-  const newData = page.filter(val => val.id !== updatedId)
-  newPagesArray.push(newData)
-})
+const newPagesArray = oldPagesArray?.pages.map((page) =>
+  page.filter((val) => val.id !== updatedId)
+)
```
TkDodo, line 163: if we do that and `oldPagesArray` is `undefined`, `newPagesArray` will be `undefined` as well, while in the previous example, it was an empty array. So we'd need `?? []` please

41y08h: Sorry, my bad 😄. I added it.

PR #2376, opened 2021-06-14
Hide devtools when closed, prevent focus

### LINKED ISSUE #2375: Devtools elements accept focus even when closed
**Describe the bug**
When tabbing through my site, I noticed that there were several hidden elements that took focus.  I traced them down to the react-query devtools, which, even though it was closed, took focus.  

**To Reproduce**
1. Go to a page where the devtools is included, but not open.
2. Open and close the devtools to set focus on the devtools button.
3. Press shift-tab to focus the previous element.
4. Run `document.activeElement` in the devtools and confirm that the `Remove` button of the devtools is focused.

**Expected behavior**
I expect to be able to tab through elements on the page, including the button to open and close the query devtools, but not tab through the devtools itself when it is closed.

IanVS:
Fixes #2375.

Previously, opening and closing the devtools only changed it opacity (and subscribed/unsubscribed to queries).  That meant that it was still possible to tab through the elements in the panel, even when it was closed.

With this change, when the panel starts out closed, we set a `visibility: hidden` style, so that it is gone from both the accessibility tree and the tab order.  I've also set up transition listeners so that before a transition starts, if the devtools are opening, we turn visibility to visible, so that the opacity fade-in works correctly.  Likewise, on a transition end, we check if the panel is closed, and turn visibility to hidden.

I tested this by compiling, copying the built files into my project's node_modules, and running my app, and it worked as I expected.  The fade still occurs (both in and out), focus does not move into a closed devtools, and the devtools are hidden in the accessibility tree. 

This probably would have been more straightforward with a keyframe animation, but I am not sure how to do that using inline styles, or whether its even possible.

### DIFF
--- a/src/devtools/devtools.tsx
+++ b/src/devtools/devtools.tsx
@@ -407,9 +407,16 @@ export const ReactQueryDevtoolsPanel = React.forwardRef(
 
     React.useEffect(() => {
       if (isOpen) {
-        return queryCache.subscribe(() => {
+        const unsubscribe = queryCache.subscribe(() => {
           setUnsortedQueries(Object.values(queryCache.getAll()))
         })
+        // re-subscribing after the panel is closed and re-opened won't trigger the callback,
+        // So we'll manually populate our state
+        setUnsortedQueries(Object.values(queryCache.getAll()))
+
+        return () => {
+          unsubscribe()
+        }
       }
       return undefined
     }, [isOpen, sort, sortFn, sortDesc, setUnsortedQueries, queryCache])
--- a/src/devtools/devtools.tsx
+++ b/src/devtools/devtools.tsx
@@ -214,7 +214,7 @@ export function ReactQueryDevtools({
             borderTop: `1px solid ${theme.gray}`,
             transformOrigin: 'top',
             // visibility will be toggled after transitions, but set initial state here
-            visibility: initialIsOpen ? 'visible' : 'hidden',
+            visibility: isOpen ? 'visible' : 'hidden',
             ...panelStyle,
             ...(isResizing
               ? {
--- a/src/devtools/devtools.tsx
+++ b/src/devtools/devtools.tsx
@@ -133,6 +133,33 @@ export function ReactQueryDevtools({
     setIsResolvedOpen(isOpen)
   }, [isOpen, isResolvedOpen, setIsResolvedOpen])
 
+  const handlePanelTransitionStart = React.useCallback(() => {
+    if (panelRef.current && isResolvedOpen) {
+      panelRef.current.style.visibility = 'visible'
+    }
+  }, [isResolvedOpen])
+
+  const handlePanelTransitionEnd = React.useCallback(() => {
+    if (panelRef.current && !isResolvedOpen) {
+      panelRef.current.style.visibility = 'hidden'
+    }
+  }, [isResolvedOpen])
+
+  // Toggle panel visibility before/after transition (depending on direction).
+  // Prevents focusing in a closed panel.
+  React.useEffect(() => {
+    if (panelRef.current) {
+      const ref = panelRef.current
+      ref.addEventListener('transitionstart', handlePanelTransitionStart)
+      ref.addEventListener('transitionend', handlePanelTransitionEnd)
+
+      return () => {
+        ref.removeEventListener('transitionstart', handlePanelTransitionStart)
+        ref.removeEventListener('transitionend', handlePanelTransitionEnd)
+      }
+    }
+  }, [handlePanelTransitionStart, handlePanelTransitionEnd])
+
   React[isServer ? 'useEffect' : 'useLayoutEffect'](() => {
     if (isResolvedOpen) {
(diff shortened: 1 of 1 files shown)


IanVS:
Oh cool, you can see it working in the codesandbox as well.  I haven't seen a github codesandbox bot before, that's pretty nifty!

TkDodo:
One of the best bots :)

TkDodo:
Thank you 👍 . I've tested this in the codesandbox, and it looks good. While testing, I found a semi-related issue that I probably introduced when I made the devtools unsubscribe when they are hidden:

- go to the codesandbox
- open the devtools
- see that you have the `posts` query there
- close the devtools
- navigate to one post
- open the devtools again
--> observe that you still only have one query (the `posts`) query in the panel
- trigger a refetch (e.g. window focus)
- second query appears

I think this happens because we put the queries into local state: 

https://github.com/tannerlinsley/react-query/blob/ff830db831f89ba7cebe66cf8c23ca218bdb3643/src/devtools/devtools.tsx#L339-L341

and then update in the effect when we are subscribed:

https://github.com/tannerlinsley/react-query/blob/ff830db831f89ba7cebe66cf8c23ca218bdb3643/src/devtools/devtools.tsx#L382

but since the update happens when we are closed, the component only re-renders when we open it and thus it still has the old queries in the local state.

---

one way to fix it is to really unmount when we close and mount again, because it would pick up the initial state of `useState` again. Another one would be to sync the state in a separate effect, calling `setUnsortedQueries` with the current value of what's in the query cache whenever `isOpen` changes. We could also pass `isOpen` as a `key` to the `ReactQueryDevtoolsPanel` which would make it unmount whenever that value changes. Not sure how that would work with the animation though, we'd have to try.

---

I understand if you don't want to tackle this at all, and we can open a separate issue as well. But if we decide to go for the unmount/remount solution because of this finding, I would rather not add the `visibility` change now just to remove it then. What do you think?

IanVS:
If we unmount/mount, we would lose other local state like filtering and such.  Is that acceptable?

If we want to keep that state, it sounds like your suggested approach would be to:

> sync the state in a separate effect, calling setUnsortedQueries with the current value of what's in the query cache whenever isOpen changes.

Are there any downsides to that approach?  It sounds relatively straightforward, and I can try it out here if you'd like.

TkDodo:
> It sounds relatively straightforward, and I can try it out here if you'd like.

cool, go ahead 😊

IanVS:
I believe I've fixed the stale queries, but just realized I've introduced a regression.  Opening the devtools and refreshing the page results in a `close` button but a missing panel.  I'll work on fixing that now...

IanVS:
Alright, I think I've taken care of that too.  Should be ready for another set of 👀 .  Thanks!

src/devtools/devtools.tsx
```diff
@@ -378,9 +407,16 @@ export const ReactQueryDevtoolsPanel = React.forwardRef(
...
+        const unsubscribe = queryCache.subscribe(() => {
           setUnsortedQueries(Object.values(queryCache.getAll()))
         })
+        // re-subscribing after the panel is closed and re-opened won't trigger the callback,
+        // So we'll manually populate our state
+        setUnsortedQueries(Object.values(queryCache.getAll()))
+
+        return () => {
+          unsubscribe()
+        }
```
TkDodo, line 419: would this also work?

```suggestion
        return unsubscribe
```

src/devtools/devtools.tsx
```diff
@@ -133,6 +133,33 @@ export function ReactQueryDevtools({
...
+    if (panelRef.current && isResolvedOpen) {
+      panelRef.current.style.visibility = 'visible'
+    }
+  }, [isResolvedOpen])
+
+  const handlePanelTransitionEnd = React.useCallback(() => {
+    if (panelRef.current && !isResolvedOpen) {
+      panelRef.current.style.visibility = 'hidden'
+    }
+  }, [isResolvedOpen])
```
TkDodo, line 146: If we move those function definitions _into_ the `useEffect` where they are used, we don't need to wrap them in `useCallback. What do you think?

src/devtools/devtools.tsx
IanVS: Yes, but it would do the same thing at the cost of another function declaration, right?

src/devtools/devtools.tsx
IanVS: Sure, I can do that.

src/devtools/devtools.tsx
TkDodo: quite the contrary. The previous solution declares a new inline arrow function and returns that, while the proposed solution just returns the already declared function _unsubscribe_ for react to invoke. It doesn't matter perf wise so we can also leave it as is :)

IanVS: Oh sorry, I was reading this on my phone and it looked backwards for some reason, lol

TkDodo (approved):
great work, thanks 🚀

PR #2378, opened 2021-06-15
updating rtk-query links in Normalized Caching topic in the docs

Luccasoli:
I just noticed this detail while I was reading the comparison.


TkDodo:
Thanks. Can you maybe fix the link to the rtk query comparison page? I think it has changed recently

Luccasoli:
Seems to be they closed the rtk-query repository after alpha, i'll update the link to star the redux-toolkit repository

PR #2384, opened 2021-06-17
docs: Fix isMutatingPosts example

Sakarisson:
Previously the example was using an array as argument, which doesn't match the newest API.

### DIFF
--- a/docs/src/pages/reference/useIsMutating.md
+++ b/docs/src/pages/reference/useIsMutating.md
@@ -10,7 +10,7 @@ import { useIsMutating } from 'react-query'
 // How many mutations are fetching?
 const isMutating = useIsMutating()
 // How many mutations matching the posts prefix are fetching?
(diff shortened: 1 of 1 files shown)


TkDodo:
right, I think we missed that `useIsMutating` should actually have the same api as `useIsFetching` (and all the other methods that take queryFilters), which includes 2 overloads. It would be appreciated if you could maybe streamline the interface - then we could keep the example, because it would be working. How does that sound?

Sakarisson:
Thanks for the comment, @TkDodo! Unfortunately I'm not very familiar with React Query, as I only recently started using it. As such, I don't feel I am qualified to do the additional changes requested.

TkDodo:
no worries - let me close this and open a PR that adds overloads 👍

TkDodo:
https://github.com/tannerlinsley/react-query/pull/2451

PR #2394, opened 2021-06-22
docs: update query function usage examples

br0p0p:
[Per the v3 breaking changes, query key parts are not automatically spread to query functions.](https://react-query.tanstack.com/guides/migrating-to-react-query-3#query-key-partspieces-are-no-longer-automatically-spread-to-the-query-function)

I've removed an old usage example that did it this way and added an example using `context.queryKey` instead.

### DIFF
--- a/docs/src/pages/guides/query-functions.md
+++ b/docs/src/pages/guides/query-functions.md
@@ -8,12 +8,12 @@ A query function can be literally any function that **returns a promise**. The p
 All of the following are valid query function configurations:
 
 ```js
-useQuery(['todos', todoId], fetchTodoById)
 useQuery(['todos', todoId], () => fetchTodoById(todoId))
 useQuery(['todos', todoId], async () => {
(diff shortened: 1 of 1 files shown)


docs/src/pages/guides/query-functions.md
```diff
@@ -8,12 +8,12 @@ A query function can be literally any function that **returns a promise**. The p
 All of the following are valid query function configurations:
 
 ```js
-useQuery(['todos', todoId], fetchTodoById)
```
TkDodo, line 11: I would please keep one example that shows that the `queryFn` doesn't need to be an inline function. Maybe:

```
useQuery(['todos'], fetchAllTodos)
```

?

br0p0p: good thinking — updated with your suggestion

TkDodo (approved):
awesome, thx

PR #2410, opened 2021-06-27
Add Uber

joey-colon:
We're big fans of react-query over at ubereats.com

Thanks @tannerlinsley & team 😃


tannerlinsley:
Nice! Let’s talk about a sponsorship! 😉

PR #2421, opened 2021-06-30
fix: rename asyncStoragePersistor to createAsyncStoragePersistor

brunohkbx:
## Why
This function was being called as `createAsyncStoragePersistor` in the docs but was exported as `asyncStoragePersistor`

 Screenshot from 2021-06-30 16-20-02

### DIFF
--- a/docs/src/pages/plugins/createAsyncStoragePersistor.md
+++ b/docs/src/pages/plugins/createAsyncStoragePersistor.md
@@ -17,7 +17,7 @@ This utility comes packaged with `react-query` and is available under the `react
 - Pass it to the [`persistQueryClient`](../persistQueryClient) function
 
 ```ts
-import AsyncStorage from '@react-native-community/async-storage'
+import AsyncStorage from '@react-native-async-storage/async-storage'
 import { persistQueryClient } from 'react-query/persistQueryClient-experimental'
 import { createAsyncStoragePersistor } from 'react-query/createAsyncStoragePersistor-experimental'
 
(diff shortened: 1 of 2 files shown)


Aung-Myint-Thein:
Yes, I found out that just today and also `asyncStorageKey` storage persistor options is just `key` in the codes. So, might need to fix the doc or fix the codes. I think changing the doc is easier approach but the overall naming will be a bit off. Any take on this?

maylinmedina:
I found another issue. The `storage `parameter in `CreateAsyncStoragePersistorOptions` has `AsyncStorage` type, but the `AsyncStorage` object imported from` @react-native-async-storage/async-storage` has `AsyncStorageStatic ` type, so there is an incompatibility when building the persistor by following the documentation:
```
import AsyncStorage from '@react-native-async-storage/async-storage';
import { createAsyncStoragePersistor } from 'react-query/createAsyncStoragePersistor-experimental';

const asyncStoragePersistor = createAsyncStoragePersistor({
  storage: AsyncStorage,
});
```

brunohkbx:
Thanks @Aung-Myint-Thein @maylinmedina for pointing it out. I think it should be good now. can you guys take a look again?

maylinmedina:
@brunohkbx  Looks good to me. Thanks.

TkDodo (approved):

bombillazo:
I'm eagerly awaiting these changes :D wanna test it out in my build.

Aung-Myint-Thein:
> I'm eagerly awaiting these changes :D wanna test it out in my build.

Oh you can actually test out this in the codes. The discussion here is just for naming consistency :D Please let us know where you have problems and I can help out.

PR #2433, opened 2021-07-04
test(devtools): Add base tests for react-query devtools

### LINKED ISSUE #2371: Add tests for react-query devtools
**Describe the bug**
I saw that there aren't any tests as of now for react-query devtools. 

If there is no one working on it right now, and there are plans to add tests to devtools, then I would love to work on it.

prateek3255:
- ✅.  Added the following tests for devtools (except query details explorer)
    - Opening and closing the devtools panel
    - Displaying the correct states for queries
    - Displaying all the query hash in the rows and opening the query details explorer upon clicking it
    - Searching the queries
    - Sorting the queries via different sorting filters available
- ♿️    Added accessibility labels to buttons and inputs, which not only helped in writing good tests but also improved the overall accessibility of devtools
- 🐛.  Fixed a bug that sorted queries incorrectly. While adding the test cases for sorting queries, I found that the sorting was not working as expected because of an incorrect key name being used (`updatedAt` instead of `dataUpdatedAt`)

Fixes #2371

### DIFF
--- a/src/devtools/tests/devtools.test.tsx
+++ b/src/devtools/tests/devtools.test.tsx
@@ -282,10 +282,10 @@ describe('ReactQueryDevtools', () => {
     // to just the number, with the order being -> query-1, query-2, query-3
     fireEvent.change(sortSelect, { target: { value: 'Query Hash' } })
     /** To check the order of the queries we can use regex to find
-     * all the row items in an array and then compare the items 
+     * all the row items in an array and then compare the items
      * one by one in the order we expect it
      * @reference https://github.com/testing-library/react-testing-library/issues/313#issuecomment-625294327
-    */
+     */
     queries = await screen.findAllByText(/\["query-[1-3]"\]/)
     expect(queries[0]?.textContent).toEqual(query1Hash)
     expect(queries[1]?.textContent).toEqual(query2Hash)
@@ -300,7 +300,7 @@ describe('ReactQueryDevtools', () => {
     expect(queries[1]?.textContent).toEqual(query3Hash)
     expect(queries[2]?.textContent).toEqual(query1Hash)
 
-    // When sorted by the status and then last updated date the queries 
+    // When sorted by the status and then last updated date the queries
     // query-3 takes precedence because its stale time being infinity, it
     // always remains fresh, the rest of the queries are sorted by their last
     // updated time, so the resulting order is -> query-3, query-2, query-1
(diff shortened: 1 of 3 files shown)


prateek3255:
@TkDodo @tannerlinsley  Just adding a comment here to make sure you didn't miss this PR 😬. Also, do let me know if I should squash the commits.

src/devtools/tests/devtools.test.tsx
```diff
@@ -0,0 +1,321 @@
...
+    defaultOptions: {
+      queries: {
+        staleTime: 0,
+      },
+    },
+  })
+
+  beforeEach(() => {
+    queryCache.clear()
+  })
```
TkDodo, line 24: I would prefer if we'd either:
- create a new QueryClient for each test so that they can run in isolation
- use the same cache and separate concerns by using unique query keys (there is a `queryKey()` util function that we use in other tests to ensure this.

That way, tests are more isolated and we can also run them potentially in parallel.

src/devtools/tests/devtools.test.tsx
```diff
@@ -0,0 +1,321 @@
...
+
+      return (
+        <div>
+          <h1>{data}</h1>
+        </div>
+      )
+    }
+
+    function PageParent() {
+      const [isPageVisible, setIsPageVisible] = React.useState(true)
```
TkDodo, line 85: little nit-picky tip:

```
const [isPageVisible, togglePageVisible] = React.useReducer(visible => !visible, true)
```

then you can pass it to onClick via:

```
onClick={togglePageVisible}
```

src/devtools/tests/devtools.test.tsx
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    renderWithClient(queryClient, <PageParent />)
+
+    fireEvent.click(
+      screen.getByRole('button', { name: /open react query devtools/i })
+    )
+
+    // Find the current query from the cache
+    const currentQuery = queryCache.find('check')
```
TkDodo, line 108: let's not over-extend on comments please. The comment here (`find the query from the cache`) doesn't add much which the code itself doesn't tell us via: `queryCache.find`. It does however make the flow of the code harder to read if every line of code has 3 lines of comments stating the WHAT is happening.

So let's try to keep comments describing maybe the WHY something is happening, if we need them at all. tests written with react-testing-library are pretty declarative and self-explanatory, so only a few comments should be needed please.

TkDodo (changes requested):
good work, thank you 💪. Just a couple of small comments from my side please:

src/devtools/tests/devtools.test.tsx
prateek3255: Got it using the former method here because we rely on only a certain number of items being visible in devtools for some tests, hence creating a fresh client and cache for every test via a helper

src/devtools/tests/devtools.test.tsx
prateek3255: ✅

src/devtools/tests/devtools.test.tsx
prateek3255: Thanks for pointing that out 🙌  . Toned down the unnecessary comments and kept only the ones that I think are necessary to understand why something is happening

src/devtools/tests/devtools.test.tsx
prateek3255: ✅

prateek3255:
@TkDodo Let me know if anything else is pending here.

TkDodo (approved):

PR #2438, opened 2021-07-06
Instruction for Kotlin/JS declarations

turansky:


TkDodo:
Looking at the deploy preview (https://react-query-git-fork-turansky-patch-1-tannerlinsley.vercel.app/overview), I don't think the page is visible anywhere? I think you'd need to wire it to the menu or so, but I'm not sure it deserves it's own page ... ?

turansky:
I have used [this page](https://github.com/tannerlinsley/react-table/blob/master/TYPESCRIPT.md) as example.
It also unavailable [on site](https://react-table.tanstack.com/)

TkDodo:
Then it’s probably not a good source of inspiration 😅

turansky:
@TkDodo You are right. New PR required

PR #2453, opened 2021-07-12
clear trackedProps of the pre render

liaoliao666:
clear the previous trackedProps to make the current trackedProps is correct


TkDodo:
We had auto resetting when we first implemented the feature, but decided to turn it off. Please see this discussion on the original PR: https://github.com/tannerlinsley/react-query/pull/1578#discussion_r558838697

liaoliao666:
@TkDodo okay, thankssss

PR #2483, opened 2021-07-21
fix(typo): Update isError to error

drewbaumann:
* Updates the copy to be consistent with the success state copy below it.


drewbaumann:
I'd suggest changing one of these two lines for consistency.
Screen Shot 2021-07-20 at 7 55 01 PM

PR #2513, opened 2021-07-29
feat (QueryClient): getQueriesData for matching multiple queries

babycourageous:
Similar to `setQueriesData`, a "get" version to fuzzily return multiple matching queries.

Based the code off of `setQueriesData` and `getQueryData` - couldn't get typing quite right without the assertion to `TData` in the map. I'm sure there's a better way to go with that.

Hope this at least points in the right direction!

### DIFF
--- a/docs/src/pages/reference/QueryClient.md
+++ b/docs/src/pages/reference/QueryClient.md
@@ -23,7 +23,6 @@ await queryClient.prefetchQuery('posts', fetchPosts)
 
 Its available methods are:
 
-- [`QueryClient`](#queryclient)
 - [`queryClient.fetchQuery`](#queryclientfetchquery)
 - [`queryClient.fetchInfiniteQuery`](#queryclientfetchinfinitequery)
 - [`queryClient.prefetchQuery`](#queryclientprefetchquery)
--- a/docs/src/pages/reference/QueryClient.md
+++ b/docs/src/pages/reference/QueryClient.md
@@ -23,30 +23,32 @@ await queryClient.prefetchQuery('posts', fetchPosts)
 
 Its available methods are:
 
-- [`fetchQuery`](#queryclientfetchquery)
-- [`fetchInfiniteQuery`](#queryclientfetchinfinitequery)
-- [`prefetchQuery`](#queryclientprefetchquery)
-- [`prefetchInfiniteQuery`](#queryclientprefetchinfinitequery)
-- [`getQueryData`](#queryclientgetquerydata)
-- [`setQueryData`](#queryclientsetquerydata)
-- [`setQueriesData`](#queryclientsetqueriesdata)
-- [`getQueryState`](#queryclientgetquerystate)
-- [`invalidateQueries`](#queryclientinvalidatequeries)
-- [`refetchQueries`](#queryclientrefetchqueries)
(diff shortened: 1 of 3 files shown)


src/core/queryClient.ts
```diff
@@ -111,6 +111,19 @@ export class QueryClient {
...
 
+  getQueriesData<TData = unknown>(queryKey: QueryKey): [QueryKey, TData][]
+  getQueriesData<TData = unknown>(filters: QueryFilters): [QueryKey, TData][]
+  getQueriesData<TData = unknown>(
+    queryKeyOrFilters: QueryKey | QueryFilters
+  ): [QueryKey, TData][] {
+    return this.getQueryCache()
+      .findAll(queryKeyOrFilters)
+      .map(({ queryKey, state }) => {
+        const data = state.data as TData
```
TkDodo, line 122: I think the type assertion is fine. Users don't need to provide the `TData` generic, so they will get back `unknown` per default. Which is correct if you don't know the structure of the queries you are getting. For example, if you try to find all active queries, the structure of the result might be totally different, so `unknown` is good. If you, as a developer, know that you are only getting back todo-lists from `getQueriesData`, then you can basically do the assertion by adding the generic:

```
queryClient.getQueriesData<Array<Todo>>(['todo', 'list'])
```

so I think it just makes life easier for consumers 👍

TkDodo (changes requested):
please also add documentation to the queryClient API docs: https://react-query.tanstack.com/reference/QueryClient

src/core/queryClient.ts
babycourageous: Thanks! Sounds good. I put a note in the docs about this caveat, but let me know if it isn't necessary.

docs/src/pages/reference/QueryClient.md
```diff
@@ -23,30 +23,32 @@ await queryClient.prefetchQuery('posts', fetchPosts)
...
-- [`getDefaultOptions`](#queryclientsetdefaultoptions)
-- [`setDefaultOptions`](#queryclientgetdefaultoptions)
-- [`getQueryDefaults`](#queryclientgetquerydefaults)
-- [`setQueryDefaults`](#queryclientsetquerydefaults)
-- [`getMutationDefaults`](#queryclientgetmutationdefaults)
-- [`setMutationDefaults`](#queryclientsetmutationdefaults)
-- [`getQueryCache`](#queryclientgetquerycache)
-- [`getMutationCache`](#queryclientgetmutationcache)
-- [`clear`](#queryclientclear)
+- [`QueryClient`](#queryclient)
```
TkDodo, line 26: I think we should remove this line. It points basically to itself, and it doesn't make much sense that the text reads:

`Its available methods are:` and then the first thing is `QueryClient` 🙃

TkDodo (changes requested):
looks great, let's remove this one line please and then we can 🚢 🇮🇹

docs/src/pages/reference/QueryClient.md
babycourageous: Oh - i didn't even notice it added that line when I saved. I'll delete that line and do a save w/o format to remove it!

TkDodo:
oh damn, there was a blank after `feat ` and now we didn't release this as a minor version :/

PR #2525, opened 2021-08-01
Add useIsError hook

michalenr:
Related to #2524

### DIFF
--- a/src/react/tests/useIsError.test.tsx
+++ b/src/react/tests/useIsError.test.tsx
@@ -1,0 +1,32 @@
+import React from 'react'
+
+import { queryKey, renderWithClient } from './utils'
+import { useQuery, useIsError, QueryClient } from '../..'
+
+describe('useIsError', () => {
+  xit('should return number of errored queries', async () => {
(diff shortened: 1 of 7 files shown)


src/react/tests/useIsError.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+describe('useIsError', () => {
+  xit('should return number of errored queries', async () => {
+    const queryClient = new QueryClient()
+
+    const key1 = queryKey()
+
+    function Page() {
+      const { status } = useQuery(key1, async () => {
+        throw new Error()
```
michalenr, line 14: I would appreciate help in getting this to work.

It retries three times and then the test exits. Status remains 'loading'

TkDodo: you probably need to turn off retries for the query.
---
but let's please continue [the discussion](https://github.com/tannerlinsley/react-query/discussions/2524) to see if there aren't better ways to handle your issue. I don't think that a hook that gives you a number of failed queries is particularly useful.

PR #2548, opened 2021-08-06
Support the ability to provide a context

blackarctic:
This adds support for passing in a context as other robust libraries (such as Redux) support: https://react-redux.js.org/api/provider#props.

This is important as right now React Query supports creating multiple `queryClient` instances but does not currently support the ability to specify which `queryClient` instance should be used with `useQueryClient`. In the case of nested `QueryClientProvider` instances in an application, the ability to specify which context (and thus which `QueryClientProvider` and `queryClient`)  should be used is critical. This adds such capability.

These changes should be 100% backwards compatible.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,8 +1,0 @@
-{
-  "workbench.colorCustomizations": {
-    "titleBar.activeBackground": "#00da63", // change this color!
-    "titleBar.inactiveBackground": "#00da63", // change this color!
-    "titleBar.activeForeground": "#ffffff", // change this color!
-    "titleBar.inactiveForeground": "#ffffff" // change this color!
-  }
-}
--- a/src/reactjs/QueryClientProvider.tsx
+++ b/src/reactjs/QueryClientProvider.tsx
@@ -9,7 +9,9 @@ declare global {
   }
 }
 
-const defaultContext = React.createContext<QueryClient | undefined>(undefined)
+export const defaultContext = React.createContext<QueryClient | undefined>(
+  undefined
+)
 const QueryClientSharingContext = React.createContext<boolean>(false)
 
 // If we are given a context, we will use it.
--- a/src/reactjs/QueryClientProvider.tsx
+++ b/src/reactjs/QueryClientProvider.tsx
@@ -1,6 +1,7 @@
 import React from 'react'
 
 import { QueryClient } from '../core'
+import { ContextOptions } from '../reactjs/types'
 
 declare global {
   interface Window {
@@ -11,13 +12,20 @@ declare global {
 const defaultContext = React.createContext<QueryClient | undefined>(undefined)
 const QueryClientSharingContext = React.createContext<boolean>(false)
 
-// if contextSharing is on, we share the first and at least one
+// If we are given a context, we will use it.
+// Otherwise, if contextSharing is on, we share the first and at least one
 // instance of the context across the window
 // to ensure that if React Query is used across
 // different bundles or microfrontends they will
 // all use the same **instance** of context, regardless
 // of module scoping.
-function getQueryClientContext(contextSharing: boolean) {
+function getQueryClientContext(
+  context: React.Context<QueryClient | undefined> | undefined,
+  contextSharing: boolean
+) {
+  if (context) {
+    return context
+  }
   if (contextSharing && typeof window !== 'undefined') {
     if (!window.ReactQueryClientContext) {
       window.ReactQueryClientContext = defaultContext
@@ -29,9 +37,9 @@ function getQueryClientContext(contextSharing: boolean) {
   return defaultContext
 }
 
-export const useQueryClient = () => {
+export const useQueryClient = ({ context }: ContextOptions = {}) => {
   const queryClient = React.useContext(
-    getQueryClientContext(React.useContext(QueryClientSharingContext))
+    getQueryClientContext(context, React.useContext(QueryClientSharingContext))
   )
 
   if (!queryClient) {
@@ -41,15 +49,26 @@ export const useQueryClient = () => {
   return queryClient
 }
 
-export interface QueryClientProviderProps {
+type QueryClientProviderPropsBase = {
   client: QueryClient
-  contextSharing?: boolean
 }
+type QueryClientProviderPropsWithContext = ContextOptions & {
+  contextSharing?: never
+} & QueryClientProviderPropsBase
+type QueryClientProviderPropsWithContextSharing = {
+  context?: never
+  contextSharing?: boolean
+} & QueryClientProviderPropsBase
+
+export type QueryClientProviderProps =
+  | QueryClientProviderPropsWithContext
+  | QueryClientProviderPropsWithContextSharing
 
 export const QueryClientProvider: React.FC<QueryClientProviderProps> = ({
   client,
-  contextSharing = false,
   children,
+  context,
+  contextSharing = false,
 }) => {
   React.useEffect(() => {
     client.mount()
@@ -58,10 +77,10 @@ export const QueryClientProvider: React.FC<QueryClientProviderProps> = ({
     }
   }, [client])
 
-  const Context = getQueryClientContext(contextSharing)
+  const Context = getQueryClientContext(context, contextSharing)
 
   return (
-    <QueryClientSharingContext.Provider value={contextSharing}>
+    <QueryClientSharingContext.Provider value={!context && contextSharing}>
       <Context.Provider value={client}>{children}</Context.Provider>
     </QueryClientSharingContext.Provider>
   )
--- a/src/reactjs/types.ts
+++ b/src/reactjs/types.ts
@@ -8,6 +8,14 @@ import {
   MutationObserverOptions,
   MutateFunction,
 } from '../core/types'
+import type { QueryClient } from '../core/queryClient'
+
+export interface ContextOptions {
+  /**
+   * Use this to pass your React Query context. Otherwise, the default will be used.
+   */
+  context?: React.Context<QueryClient | undefined>
+}
 
 export interface UseBaseQueryOptions<
   TQueryFnData = unknown,
@@ -15,13 +23,8 @@ export interface UseBaseQueryOptions<
   TData = TQueryFnData,
   TQueryData = TQueryFnData,
   TQueryKey extends QueryKey = QueryKey
-> extends QueryObserverOptions<
-    TQueryFnData,
-    TError,
-    TData,
-    TQueryData,
-    TQueryKey
-  > {}
+> extends ContextOptions,
+    QueryObserverOptions<TQueryFnData, TError, TData, TQueryData, TQueryKey> {}
 
 export interface UseQueryOptions<
   TQueryFnData = unknown,
@@ -42,13 +45,14 @@ export interface UseInfiniteQueryOptions<
   TData = TQueryFnData,
   TQueryData = TQueryFnData,
   TQueryKey extends QueryKey = QueryKey
-> extends InfiniteQueryObserverOptions<
-    TQueryFnData,
-    TError,
-    TData,
-    TQueryData,
-    TQueryKey
-  > {}
+> extends ContextOptions,
+    InfiniteQueryObserverOptions<
+      TQueryFnData,
+      TError,
+      TData,
+      TQueryData,
+      TQueryKey
+    > {}
 
 export type UseBaseQueryResult<
   TData = unknown,
@@ -70,10 +74,11 @@ export interface UseMutationOptions<
   TError = unknown,
   TVariables = void,
   TContext = unknown
-> extends Omit<
-    MutationObserverOptions<TData, TError, TVariables, TContext>,
-    '_defaulted' | 'variables'
-  > {}
+> extends ContextOptions,
+    Omit<
+      MutationObserverOptions<TData, TError, TVariables, TContext>,
+      '_defaulted' | 'variables'
+    > {}
 
 export type UseMutateFunction<
   TData = unknown,
--- a/src/reactjs/types.ts
+++ b/src/reactjs/types.ts
@@ -12,7 +12,7 @@ import type { QueryClient } from '../core/queryClient'
 
 export interface ContextOptions {
   /**
-   * Use this to pass your React Query context. Otherwise, the default will be used.
+   * Use this to pass your React Query context. Otherwise, `defaultContext` will be used.
    */
   context?: React.Context<QueryClient | undefined>
 }
--- a/src/reactjs/useIsMutating.ts
+++ b/src/reactjs/useIsMutating.ts
@@ -2,22 +2,30 @@ import React from 'react'
 
 import { notifyManager } from '../core/notifyManager'
 import { MutationKey } from '../core/types'
+import { ContextOptions } from '../reactjs/types'
 import { MutationFilters, parseMutationFilterArgs } from '../core/utils'
 import { useQueryClient } from './QueryClientProvider'
 
-export function useIsMutating(filters?: MutationFilters): number
+interface Options extends ContextOptions {}
+
+export function useIsMutating(
+  filters?: MutationFilters,
+  options?: Options
+): number
 export function useIsMutating(
   mutationKey?: MutationKey,
-  filters?: Omit<MutationFilters, 'mutationKey'>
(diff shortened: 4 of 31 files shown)


TkDodo:
What would be the use-case for needing multiple, nested QueryClientProviders where you do _not_ want to lookup the nearest one?

ngallows:
The scenario I have been seeing is an app container that wants to provide some core data using React Query but wants to have no coupling to the apps that it renders that may or may not be using their own React Query for their own data. In that case, trying to access the core data will try to access the app’s React Query instead of the app container’s.

Assuming a single context for all React Query instances leads to this issue and is the reason robust libraries like Redux support the option to specify which context to use at the provider and at the consumer level.

TkDodo:
I think the code changes look fine, I'd still want @tannerlinsley's opinion here. We'd also need some tests and updates to the docs please. Feel free to wait with that until Tanner gives his okay :)

blackarctic:
Any updates here?

TkDodo:
@tannerlinsley what do you think please? Is that a feature that should make it into the library?

blackarctic:
To illustrate the need for this a bit more clearly here:

```jsx
<ContainerDataProvider> // <-- Provides container data using its own React Query provider
  ...
  <AppDataProvider> // <-- Provides app data using its own React Query provider
    ...
      <MyComponentDataProvider> // <-- Provides component data using its own React Query provider
        <MyComponent />
      </MyComponentDataProvider>
    ...
  </AppDataProvider>
  ...
</ContainerDataProvider>
```

Each "DataProvider" uses its own React Query provider for isolation and abstraction. Those that consume the data of these providers should not have to know what library they are using for data fetching and storage.

In this case, `MyComponent` needs to access `useQuery` from both  `MyComponentDataProvider` and `ContainerDataProvider`. Unless we can specify which context we wish to use for each `useQuery`, it will default to `MyComponentDataProvider` and fail.

tannerlinsley:
I think this is a valid addition as long as:
- It is backwards compatible
- Doesn't require any code gymnastics to support it. The code looks simple enough on my first look-through, so let's hope it stays that way
- It gets documented extremely well as an optional and often unnecessary feature
- It gets tested well.

TkDodo:
@blackarctic would you like to continue with the PR now that we have the approval from Tanner?

blackarctic:
Yes. Sounds like we need:

- Update branch to latest
- Add accompanying tests
- Add accompanying documentation

.vscode/settings.json
```diff
@@ -1,8 +0,0 @@
-{
-  "workbench.colorCustomizations": {
-    "titleBar.activeBackground": "#00da63", // change this color!
-    "titleBar.inactiveBackground": "#00da63", // change this color!
-    "titleBar.activeForeground": "#ffffff", // change this color!
-    "titleBar.inactiveForeground": "#ffffff" // change this color!
-  }
-}
```
blackarctic, line 8: This file should never have been commited on `master`. It is in the `.gitignore`

blackarctic:
@TkDodo PR should be ready for re-review

TkDodo:
code looks good - there seems to be one test failing consistently.

blackarctic:
Not seeing any failing tests but it looks like coverage dipped by `0.03%` in `src/devtools/devtools.tsx`. However, none of the uncovered lines of code were touched as part of this PR. Trying to get this PR in, though, so I added a test to test some of the untested code in the file to get the coverage up.

TkDodo:
Thank you. I am thinking that maybe we should wait with the PR until v4 though. We are planning to change the api for `useQueries` to become an object, which would fit this use-case very well, because we could provide the `context` option at the very top, working for all queries:

```
useQueries({
  queries: [query1, query2, ..,. queryN],
  context: MyContext
})
```

blackarctic:
> Thank you. I am thinking that maybe we should wait with the PR until v4 though. We are planning to change the api for `useQueries` to become an object, which would fit this use-case very well, because we could provide the `context` option at the very top, working for all queries:
> 
> ```
> useQueries({
>   queries: [query1, query2, ..,. queryN],
>   context: MyContext
> })
> ```

That would mean each query config (`query1, query2, ..,. queryN`) could not contain their own context? I would prefer to do that as a followup as that changes a good amount of types and implementation.

TkDodo:
Judging from the current implementation, I was under the impression that it is a requirement that each query in useQueries uses the same context?

blackarctic:
> Judging from the current implementation, I was under the impression that it is a requirement that each query in useQueries uses the same context?

Correct, but this now changes the types of these query objects so that they cannot include the `context` option. They are now a different type of query object than what is passed to `useQuery`. This also changes the implementation of `useQueries` since we can no longer blindly pass this different query object (without the `context`) to handlers of single queries because those might be handling the `context`. This will require us to re-add the `context` to each query object before passing it to any single query handlers.

TkDodo:
> Correct, but this now changes the types of these query objects so that they cannot include the context option. They are now a different type of query object than what is passed to useQuery. 

I think that's fine - we can `Omit` the `context` on type level for `useQueries`.

> This also changes the implementation of useQueries since we can no longer blindly pass this different query object (without the context) to handlers of single queries because those might be handling the context. This will require us to re-add the context to each query object before passing it to any single query handlers.

yes, we would need to spread the `context` from the top level into each of the queries. Is that a problem?

blackarctic:
> > Correct, but this now changes the types of these query objects so that they cannot include the context option. They are now a different type of query object than what is passed to useQuery.
> 
> I think that's fine - we can `Omit` the `context` on type level for `useQueries`.
> 
> > This also changes the implementation of useQueries since we can no longer blindly pass this different query object (without the context) to handlers of single queries because those might be handling the context. This will require us to re-add the context to each query object before passing it to any single query handlers.
> 
> yes, we would need to spread the `context` from the top level into each of the queries. Is that a problem?

Nope, not a problem. Just outlining the additional changes that will need to happen and the increased risk of issues in long-running PRs.

TkDodo:
I've just merged the new useQueries api into the `alpha` branch. can you please base your PR on `alpha`, thank you :)

src/reactjs/useQueries.ts
```diff
@@ -118,7 +118,18 @@ export function useQueries<T extends any[]>({
...
+    const next = queries[i + 1]
+    if (next && curr.context !== next.context) {
+      throw new Error(
+        'useQueries requires all queries to have the same context'
+      )
+    }
+  })
+
+  const context = queries[0]?.context
+  const queryClient = useQueryClient({ context })
```
TkDodo, line 132: since `useQueries` has the object syntax in v4, let's please do:

```
useQueries({
  queries: [...],
  context: context
})
```

on type level, we could `Omit` the `context` property from the object we pass to each `query`

blackarctic: Sorry for the delay. This is completed and pushed up.

src/reactjs/useIsMutating.ts
```diff
@@ -2,22 +2,32 @@ import React from 'react'
...
-  arg2?: Omit<MutationFilters, 'mutationKey'>
+  arg2?: Omit<MutationFilters, 'mutationKey'> | Options,
+  arg3?: Options
 ): number {
   const mountedRef = React.useRef(false)
-  const filters = parseMutationFilterArgs(arg1, arg2)
+  const [filters, options = {}] = parseMutationFilterArgs(arg1, arg2, arg3)
 
-  const queryClient = useQueryClient()
+  const queryClient = useQueryClient({ context: (options as Options).context })
```
TkDodo, line 30: why is the type assertion here necessary, while it is not necessary for `useIsFetching` ?

src/reactjs/types.ts
```diff
@@ -21,7 +22,9 @@ export interface UseBaseQueryOptions<
     TData,
     TQueryData,
     TQueryKey
-  > {}
+  > {
+  context?: React.Context<QueryClient | undefined>
```
TkDodo, line 26: I think it would be better to have this in a shared `ContextOptions` interface that we can extend everywhere

src/reactjs/useIsMutating.ts
blackarctic: Great point. Pushed a new commit that aligns `parseFilterArgs` with `parseMutationFilterArgs`. For some reason, their types were not aligned even though their functionality is identical.

src/reactjs/types.ts
blackarctic: Done. Pushed a new commit to address this.

TkDodo:
This looks very good, thank you 🙏 . It seems that the `useQueryClient` api docs do not mention the new context yet, could we add that there as well?

Also, I'm a bit worried that the name `context` might be confused with the `queryFunctionContext` that we already have - the object that is injected into the `queryFn`. Do you think it would be better to choose a different name?

blackarctic:
> This looks very good, thank you 🙏 . It seems that the `useQueryClient` api docs do not mention the new context yet, could we add that there as well?
> 
> Also, I'm a bit worried that the name `context` might be confused with the `queryFunctionContext` that we already have - the object that is injected into the `queryFn`. Do you think it would be better to choose a different name?

Hmmm, I see your point. Well we could call it `reactContext`, `reactQueryContext`, `customReactQueryContext`, `customReactContext`, `customReactQueryContext`, `queryContext`, or `customQueryContext`? Any suggestions?

src/reactjs/QueryClientProvider.tsx
```diff
@@ -50,6 +58,7 @@ export const QueryClientProvider: React.FC<QueryClientProviderProps> = ({
   client,
   contextSharing = false,
   children,
+  context,
```
TkDodo, line 61: from an api perspective - does it make sense to allow passing in `context` _and_ `contextSharing` ? I think this would be an invalid state. One solution would be to either combine them to the same prop, or to make the props a union type where you cannot pass in one when you have already provided the other ....

blackarctic: I have attempted to type this with https://github.com/tannerlinsley/react-query/pull/2548/commits/358dfd0aa1441301bc1b018ba6a995ec94827e61. Let me know if this does or does not accomplish what you were thinking.

TkDodo: > It shouldn't be an invalid state. context takes priority over contextSharing

I'm aware. I'm just thinking from an api perspective, you could provide props that make no sense. Similar to how:

`<Button isPrimary isSecondary>`

makes no sense and it becomes an implementation detail over what takes precedence, something like:

`<Button variant="primary">` or `<Button variant="secondary">`

makes for a better api.

So here, you can do:

```
<QueryClientProvider context={MyContext} contextSharing>
```

but then yeah contextSharing will be `false` even though you provided `true` as a prop. That is the impossible state :)

---

we could do `context = React.Context | 'sharing'` to unify the props. I'm not sure if I like that or not, but we use such polymorphic props in other places too, like:

`notifyOnChangeProps?: Array<keyof InfiniteQueryObserverResult> | 'all'` 🤔

blackarctic: @TkDodo I believe I have already captured this as part of https://github.com/tannerlinsley/react-query/commit/358dfd0aa1441301bc1b018ba6a995ec94827e61. TS will not let you pass both `contextSharing` and `context`. You can see this with:

```typescript
type QueryClientProviderPropsBase = {
  client: QueryClient
}
type QueryClientProviderPropsWithContext = ContextOptions &
  QueryClientProviderPropsBase
type QueryClientProviderPropsWithContextSharing = {
  contextSharing?: boolean
} & QueryClientProviderPropsBase

export type QueryClientProviderProps =
  | QueryClientProviderPropsWithContext
  | QueryClientProviderPropsWithContextSharing
```

TkDodo: Right, I missed that, sorry. Yes I think that is one way of ensuring that, even though it's two separate props :)

also sorry there are conflicts again 🙈 . can you update with alpha once more please?

src/reactjs/QueryClientProvider.tsx
```diff
@@ -41,24 +49,41 @@ export const useQueryClient = () => {
...
   }, [client])
 
-  const Context = getQueryClientContext(contextSharing)
+  const contextSharing = propsHasContext(props)
+    ? false
+    : props.contextSharing || false
+
+  const Context = propsHasContext(props)
+    ? getQueryClientContext(props.context, false)
+    : getQueryClientContext(undefined, contextSharing)
```
TkDodo, line 86: is that really necessary? I think we could:
- define `context?: never` on `QueryClientProviderPropsWithContextSharing`
- just pass all props to `getQueryClientContext` an destruct there
- then, the early return for `if (context)` should work just the same because if we have a context, we can't have contextSharing.

blackarctic: Yep, absolutely agree. Great callout on `context?: never`. The existing types were not enforcing the mutual exclusivity.  

Changes have been made and pushed up.

src/reactjs/QueryClientProvider.tsx
blackarctic: PR has been rebased to alpha branch and pushed up. Should be no conflicts now.

Also, `context` and `contextSharing` have been consolidated into the `context` prop for `QueryClientProvider`. Docs should also have been updated to match. See this commit:  https://github.com/tannerlinsley/react-query/pull/2548/commits/0794e1f28ed6b10035929858ad1ffc9779c2ecd4.

TkDodo:
can you also add something to the [New Features](https://react-query-alpha.tanstack.com/guides/migrating-to-react-query-4#new-features-) section in the migrating to v4 doc please. I think it's good to let people know that way that we now have this feature 🚀

bertho-zero:
Is it possible to export the default context too? I have the same structure as https://github.com/tannerlinsley/react-query/pull/2548#issuecomment-915496288 but I don't necessarily need to pass the context option everywhere.

I would like to use the parent context if it exists, much like `contextSharing` but without using `window`, the idea is to do something like:

```js
function SubApp({}) {
  // const parentQueryClient = useQueryClient(); // Does not work because throws an error if tested in isolation
  const parentQueryClient = useContext(QueryClientContext);
  const queryClient = useMemo(() => (parentQueryClient || new QueryClient()), [parentQueryClient]);

  return (
      {/* ... */}
    </QueryClientProvider>
  );
}
```

Or just an option that skips the error from `useQueryClient`.

TkDodo:
@blackarctic we are nearing the release of v4. is there a chance we could get this in?

src/reactjs/QueryClientProvider.tsx
blackarctic: Done

blackarctic:
> Is it possible to export the default context or `getQueryClientContext` too? I have the same structure as [#2548 (comment)](https://github.com/tannerlinsley/react-query/pull/2548#issuecomment-915496288) but I don't necessarily need to pass the context option everywhere.
> 
> I would like to use the parent context if it exists, much like `contextSharing` but without using `window`, the idea is to do something like:
> 
> ```js
> function SubApp({}) {
>   // const parentQueryClient = useQueryClient(); // Does not work because throws an error if tested in isolation
>   const parentQueryClient = useContext(QueryClientContext);
>   const queryClient = useMemo(() => (parentQueryClient || new QueryClient()), [parentQueryClient]);
> 
>   return (
>     <QueryClientProvider client={queryClient}>
>       {/* ... */}
>     </QueryClientProvider>
>   );
> }
> ```
> 
> Or just an option that skips the error from `useQueryClient`.

Exported the `defaultContext` from `QueryClientProvider`. @TkDodo Is that ok with you?

blackarctic:
> can you also add something to the [New Features](https://react-query-alpha.tanstack.com/guides/migrating-to-react-query-4#new-features-) section in the migrating to v4 doc please. I think it's good to let people know that way that we now have this feature 🚀

Done

blackarctic:
> @blackarctic we are nearing the release of v4. is there a chance we could get this in?

Sorry for the delay. Should be good to re-review.

blackarctic:
> @blackarctic we are nearing the release of v4. is there a chance we could get this in?

@TkDodo Sorry for the delay. Should be good to re-review.

TkDodo (approved):

TkDodo:
thank you so much for bearing with me here @blackarctic. It's been a long 7 months since you opened the PR. It will be a great addition to v4 🚀 . Thank you for the contribution 🙇

PR #2556, opened 2021-08-10
fix(core): change `shouldFetchOptionally` condition

JaeYeopHan:
Related issue: https://github.com/tannerlinsley/react-query/issues/2187

### Description

Change `shouldFetchOptionally` condition to not change the `status` of result to `loading` when `state.status` is `error`.

1. `result.isLoading` is `true` in [this line](https://github.com/tannerlinsley/react-query/blob/master/src/react/useBaseQuery.ts#L110) when `result.error != null`
2. So the error did not propagate to the error boundary. (next condition)
3. Check the `createResult` function and `shouldFetchOptionally` function

Fixes tannerlinsley/react-query#2187

> Special thanks to @flybayer

### DIFF
--- a/src/core/queryObserver.ts
+++ b/src/core/queryObserver.ts
@@ -763,7 +763,7 @@ function shouldFetchOptionally(
   return (
     options.enabled !== false &&
     (query !== prevQuery || prevOptions.enabled === false) &&
-    query.state.status !== 'error' &&
+    prevQuery.state.status !== 'error' &&
     isStale(query, options)
   )
 }
--- a/src/core/queryObserver.ts
+++ b/src/core/queryObserver.ts
@@ -763,6 +763,7 @@ function shouldFetchOptionally(
   return (
     options.enabled !== false &&
     (query !== prevQuery || prevOptions.enabled === false) &&
+    query.state.status !== 'error' &&
     isStale(query, options)
   )
 }
--- a/src/core/queryObserver.ts
+++ b/src/core/queryObserver.ts
@@ -763,7 +763,7 @@ function shouldFetchOptionally(
   return (
     options.enabled !== false &&
     (query !== prevQuery || prevOptions.enabled === false) &&
-    prevQuery.state.status !== 'error' &&
+    (prevQuery.state.status !== 'error' || prevOptions.enabled === false) &&
     isStale(query, options)
   )
 }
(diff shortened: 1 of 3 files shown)


src/core/queryObserver.ts
```diff
@@ -763,6 +763,7 @@ function shouldFetchOptionally(
   return (
     options.enabled !== false &&
     (query !== prevQuery || prevOptions.enabled === false) &&
+    query.state.status !== 'error' &&
```
TkDodo, line 766: do you know which implications this can have on non-suspense / errorBoundary usages?

JaeYeopHan: @TkDodo i didn't catch (

so, how about using `options.suspense` condition?

TkDodo: > what's the problem?

I was just trying to understand `shouldFetchOptionally` and if there is a scenario where we are in error state, but still want to fetch, and I think I just found something. Consider the scenario:

- you have an `enabled` query that errors
- you disable that query
- you enable it again
--> we should now refetch, but we don't

[Here](https://codesandbox.io/s/error-fetch-on-mount-39wc8?file=/src/App.tsx) is a working codesandbox for that scenario. You can see in the console that the `Error` is logged every time we fetch (clicking the button twice for disable/enable), and [here](https://codesandbox.io/s/error-fetch-on-mount-forked-6q6ug?file=/src/App.tsx) is the same example forked with the codesanbox preview build of this PR. You can see that we never fetch if we enable the query again :/

---

your commit message suggests we want to avoid fetching  if the status changes from `loading` to `error`, maybe we need to check the status of the previousQuery as well?

JaeYeopHan: Thanks for detail! i understood.

> we need to check the status of the previousQuery as well?

sounds good!

```
prevQuery.state.status !== 'error'
```
also working!

JaeYeopHan: > You can see that we never fetch if we enable the query again

i see. i'll check

JaeYeopHan: how about this?

```
(prevQuery.state.status !== 'error' || prevOptions.enabled !== options.enabled) &&
```

JaeYeopHan: Is this approach correct?

```ts
(prevQuery.state.status !== 'error' ||
      prevOptions.enabled !== options.enabled) &&
```

🤔

TkDodo: cool, thanks for the additional test. I think that should work 👍

TkDodo (approved):

PR #2557, opened 2021-08-10
feat(infiniteQuery): add possibility to decide which pages should be refetched

TkDodo:
refs #1866 

missing:

- [x] finalize api
- [x] tests
- [x] docs

### DIFF
--- a/src/core/queryClient.ts
+++ b/src/core/queryClient.ts
@@ -282,7 +282,11 @@ export class QueryClient {
     const [filters, options] = parseFilterArgs(arg1, arg2, arg3)
 
     const promises = notifyManager.batch(() =>
-      this.queryCache.findAll(filters).map(query => query.fetch())
+      this.queryCache.findAll(filters).map(query =>
+        query.fetch(undefined, {
+          meta: { refetchPage: options?.refetchPage },
+        })
+      )
     )
 
     let promise = Promise.all(promises).then(noop)
--- a/src/core/queryClient.ts
+++ b/src/core/queryClient.ts
@@ -24,7 +24,9 @@ import type {
   QueryObserverOptions,
   QueryOptions,
   RefetchOptions,
+  RefetchQueryFilters,
   ResetOptions,
+  ResetQueryFilters,
 } from './types'
 import type { QueryState, SetDataOptions } from './query'
 import { QueryCache } from './queryCache'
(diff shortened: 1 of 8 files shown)


src/core/queryClient.ts
```diff
@@ -282,7 +282,11 @@ export class QueryClient {
     const [filters, options] = parseFilterArgs(arg1, arg2, arg3)
 
     const promises = notifyManager.batch(() =>
-      this.queryCache.findAll(filters).map(query => query.fetch())
+      this.queryCache.findAll(filters).map(query =>
+        query.fetch(undefined, {
+          meta: { refetchPage: options?.refetchPage },
```
TkDodo, line 287: I played around with this a bit, but I'm not in love with the public api. Two things:

1) if you pass `refetchPage` on a normal useQuery (non-infinite-query), it doesn't do anything
2) if you have a queryKey, but no filters, you'd need to do:

```
queryClient.invalidateQueries('myKey', undefined, {
    refetchPage: (page, allPages) => page === allPages[0]
})
```
yes, you could do:

```
queryClient.invalidateQueries({ queryKey: 'myKey' }, {
    refetchPage: (page, allPages) => page === allPages[0]
})
```

instead, but the same drawback applies when you want to invalidate _everything_:

```
queryClient.invalidateQueries(undefined, {
    refetchPage: (page, allPages) => page === allPages[0]
})
```

---

we could alternatively add that `refetchPage` option to `queryFilters`, similar to `refetchActive` and `refetchInactive` flags we can already provide to `invalidateQueries`. It's not really a query "filter" per se, because it doesn't narrow down anything, but neither is `refetchActive`, so maybe that distinction doesn't matter that much?

---

the only thing that comes to my mind to solve issue 1) would be a separate `invalidateInfiniteQueries`, but that would also mean we'd need `refetchInfiniteQueries` and `resetInfiniteQueries` with all the overloads and I don't think that's worth it.

---

right now, I would go with adding it to the query filters. What do you think @tannerlinsley ?

TkDodo: I found a reason to maybe keep the current api: the `refetch` method returned from `useInfiniteQuery` already takes `RefetchOptions`, so it's rather easy to make `refetchPage` work there as well

tannerlinsley: I think I would prefer just adding to the query filters and having it be a noop for non-infinite queries. Thoughts?

TkDodo: yep, I can do that, let me see how that looks. Would we still want to have it as an option for `refetch` returned from `useQuery` ? I would say yes, and we could just add it to the options there as well

TkDodo: what do you think?

56a3e12b

tannerlinsley: Love it.

tannerlinsley (approved):

PR #2568, opened 2021-08-13
refactor(useQueries):add types for result

MishaKazakov:
Adds types for useQueries beacus  previously it was not possible to type the result of useQueries

### DIFF
--- a/src/react/useQueries.ts
+++ b/src/react/useQueries.ts
@@ -5,7 +5,7 @@ import { QueriesObserver } from '../core/queriesObserver'
 import { useQueryClient } from './QueryClientProvider'
 import { UseQueryOptions, UseQueryResult } from './types'
 
-export function useQueries(queries: UseQueryOptions[]): UseQueryResult[] {
+export function useQueries<TData = unknown, TError = unknown>(queries: UseQueryOptions[]): UseQueryResult<TData,TError>[] {
   const mountedRef = React.useRef(false)
   const [, forceUpdate] = React.useState(0)
 


src/react/useQueries.ts
```diff
@@ -5,7 +5,7 @@ import { QueriesObserver } from '../core/queriesObserver'
 import { useQueryClient } from './QueryClientProvider'
 import { UseQueryOptions, UseQueryResult } from './types'
 
-export function useQueries(queries: UseQueryOptions[]): UseQueryResult[] {
+export function useQueries<TData = unknown, TError = unknown>(queries: UseQueryOptions[]): UseQueryResult<TData,TError>[] {
```
TkDodo, line 8: I'm afraid I don't think it's that easy. we've had an open PR here that didn't quite make it: https://github.com/tannerlinsley/react-query/pull/1527

and there is an open issue with lots of discussions here: https://github.com/tannerlinsley/react-query/issues/1675

one problem is that every query in `useQueries` can have it's own type. It's not only for fetching N times the same query, I can also do:

```
useQueries([
  { queryKey: 'foo', queryFn: fetchFoo },
  { queryKey: 'bar', queryFn: fetchBar },
] as const)
```

where the result type should be ideal a tuple: `[UseQueryResult<Foo>, UseQueryResult<Bar>]`

MishaKazakov: True. I was too hopeful that it would be easy.

Thanks for your time. closing the pr

PR #2589, opened 2021-08-19
docs(ssr): Change useState to useMemo

jkjustjoshing:
Wouldn't `useMemo` be the more appropriate hook to use in this situation? While `useState` would accomplish the same thing, it seems like not the right tool for the job.


TkDodo:
No, it’s not. State or a ref will do, but not useMemo. Please have a read: https://tkdodo.eu/blog/use-state-for-one-time-initializations

PR #2604, opened 2021-08-27
Broadcast invalidation requests

aldebout:
This PR allows invalidating queries in other tabs that don't exist in the current tab cache.

The use case that we have is to invalidate the preview of a document in a list when editing the document in another tab, as [discussed here](https://github.com/tannerlinsley/react-query/discussions/2591).

The PR in its current form deletes the current `broadcastQueryClient` function in favor of a class, it might be worth keeping the existing code not to completely nuke the API, even on an experimental feature. There might also be a way to achieve the same result while keeping the function but my attempts at overriding `invalidateQueries` were not successful.

`getQueryFilters` removes the extra keys from the filters because functions can't be serialized and sent on a BroadcastChannel.


TkDodo:
could it be that your change to `parseFilterArgs` broke some tests?

aldebout:
> could it be that your change to `parseFilterArgs` broke some tests?

It looks like it indeed, it seems I forgot some keys in the pick function.
What threw me off is this signature, I'll add more keys until the tests pass.

```ts
export function parseFilterArgs<
  TFilters extends QueryFilters,
  TOptions = unknown
>
```

aldebout:
@TkDodo Anything I can do to move this forward?

TkDodo:
I'm sorry, I'm very unfamiliar with this plugin and how it should work, so I'll have to leave this to @tannerlinsley

tannerlinsley:
It will need to be brought up to date with any changes recently to get rid of the conflicts.

aldebout:
@tannerlinsley took me some time to come back but everything should be good now :)
FWIW it's been running in prod for a few months on our side, although it's admittedly a pretty niche feature so it has not seen a lot of action.

src/core/utils.ts
```diff
@@ -162,8 +204,8 @@ export function parseFilterArgs<
   arg3?: TOptions
 ): [TFilters, TOptions | undefined] {
   return (isQueryKey(arg1)
-    ? [{ ...arg2, queryKey: arg1 }, arg3]
-    : [arg1 || {}, arg2]) as [TFilters, TOptions]
+    ? [{ ...getQueryFilters(arg2), queryKey: arg1 }, arg3]
+    : [getQueryFilters(arg1) || {}, arg2]) as [TFilters, TOptions]
```
TkDodo, line 208: why is this change necessary please?

aldebout: At the moment, any keys outside of the QueryFilters ones still get passed.
Some types (functions for example) are not compatible with being broadcast, resulting in errors.
`getQueryFilters` narrows the type to only include `QueryFilters` without extra keys.

TkDodo: Why would it have extra keys? predicate is also a function, so how does it work with that valid key? If this change is specifically for the broadcast plugin, I think it would be better if it lived there rather than in the core

aldebout: I don't remember exactly, maybe a type solution would be more appropriate?
Good point, I've never used predicate I think, this might block the whole thing though :/

TkDodo:

PR #2610, opened 2021-08-27
docs: Add link to 3rd party React Native devtools

jacklj:
I've been using this Flipper plugin and it does the job well! Not super polished, but to have React Query devtools on React Native is a massive step up.

(I'm not affiliated, just found it the other day).

Also grammar improvements.


jacklj:
all-contributors please add @jacklj for doc

PR #2615, opened 2021-08-29
Do not render DevTools during SSR

verekia:
This PR makes the DevTools client-side-only to avoid hydration issues (see https://github.com/tannerlinsley/react-query/issues/1774).

It renders `ReactQueryDevtools` as `null` during SSR, and then renders normally after mount.

I'm experiencing a failed test on my machine, even on the `master` branch without any change of mine:

```
Expected: "[\"query-2\"]"
    Received: "[\"query-1\"]"

      324 |     fireEvent.change(sortSelect, { target: { value: 'Last Updated' } })
      325 |     queries = await screen.findAllByText(/\["query-[1-3]"\]/)
    > 326 |     expect(queries[0]?.textContent).toEqual(query2Hash)
          |                                     ^
      327 |     expect(queries[1]?.textContent).toEqual(query3Hash)
      328 |     expect(queries[2]?.textContent).toEqual(query1Hash)
      329 | 

      at src/devtools/tests/devtools.test.tsx:326:37
```

With my changes, the failed test is different:

```
Expected: "[\"query-2\"]"
    Received: "[\"query-3\"]"

      324 |     fireEvent.change(sortSelect, { target: { value: 'Last Updated' } })
      325 |     queries = await screen.findAllByText(/\["query-[1-3]"\]/)
    > 326 |     expect(queries[0]?.textContent).toEqual(query2Hash)
          |                                     ^
      327 |     expect(queries[1]?.textContent).toEqual(query3Hash)
      328 |     expect(queries[2]?.textContent).toEqual(query1Hash)
      329 | 

      at src/devtools/tests/devtools.test.tsx:326:37
```

Maybe @prateek3255 would know what's going on?

### DIFF
--- a/src/devtools/devtools.tsx
+++ b/src/devtools/devtools.tsx
@@ -5,7 +5,7 @@ import React from 'react'
 import { useQueryClient } from 'react-query'
 import { matchSorter } from 'match-sorter'
 import useLocalStorage from './useLocalStorage'
-import { useSafeState } from './utils'
+import { useIsMounted, useSafeState } from './utils'
 
 import {
   Panel,
@@ -95,6 +95,7 @@ export function ReactQueryDevtools({
   )
   const [isResolvedOpen, setIsResolvedOpen] = useSafeState(false)
   const [isResizing, setIsResizing] = useSafeState(false)
+  const isMounted = useIsMounted()
 
   const handleDragStart = (panelElement, startEvent) => {
     if (startEvent.button !== 0) return // Only allow left click for drag
@@ -195,6 +196,9 @@ export function ReactQueryDevtools({
     onClick: onToggleClick,
     ...otherToggleButtonProps
   } = toggleButtonProps
+
+  // Do not render on the server
+  if (!isMounted()) return null
 
   return (
     <Container ref={rootRef} className="ReactQueryDevtools">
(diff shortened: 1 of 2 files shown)


src/devtools/devtools.tsx
```diff
@@ -196,6 +197,9 @@ export function ReactQueryDevtools({
     ...otherToggleButtonProps
   } = toggleButtonProps
 
+  // Do not render on the server
+  if (!isMounted()) return null
```
TkDodo, line 201: why would we check for `isMounted` here? Wouldn't a check for `isServer` be sufficient?

https://github.com/tannerlinsley/react-query/blob/79ad5a9ccbce1486c10e7e364d8fc28bef3fa19f/src/devtools/utils.ts#L8

It would also let us get rid of the comment, because it would be self-explanatory :)

TkDodo (changes requested):

TkDodo:
> I'm experiencing a failed test on my machine, even on the master branch without any change of mine:

hmm, it works fine on CI 🤷

src/devtools/devtools.tsx
verekia: That's because doing this can break rehydration. You can read more about it here: https://www.joshwcomeau.com/react/the-perils-of-rehydration/
I actually tried and it breaks on the Next example. A two-pass rendering is safer. (renders `null` on the server, then `null` when rehydrating, then finally the client-side only component).
This way we don't break rehydration.

verekia: Screenshot 2021-08-29 at 19 41 56

TkDodo: Thanks, today I learned 😅

TkDodo (approved):

src/devtools/devtools.tsx
verekia: Heh, no problem, always something new to learn! If you every find your website broken in prod but not in dev with super weird styles or DOM, this can be the cause. Like some `div` closing on an other `div` or some crazy stuff like that.

PR #2620, opened 2021-08-31
fix(UseMutationResult): result type should be more specific to allow type narrowing

raoufswe:
fixes https://github.com/tannerlinsley/react-query/issues/2601

This PR refactors useMutationResult type to follow the same approach/logic of useQueryResult type.

### DIFF
--- a/src/core/mutationObserver.ts
+++ b/src/core/mutationObserver.ts
@@ -4,6 +4,7 @@ import type { QueryClient } from './queryClient'
 import { Subscribable } from './subscribable'
 import type {
   MutateOptions,
+  MutationObserverBaseResult,
   MutationObserverResult,
   MutationObserverOptions,
 } from './types'
@@ -129,7 +130,12 @@ export class MutationObserver<
       ? this.currentMutation.state
       : getDefaultState<TData, TError, TVariables, TContext>()
 
-    this.currentResult = {
+    const result: MutationObserverBaseResult<
+      TData,
+      TError,
+      TVariables,
+      TContext
+    > = {
       ...state,
       isLoading: state.status === 'loading',
       isSuccess: state.status === 'success',
@@ -138,6 +144,13 @@ export class MutationObserver<
       mutate: this.mutate,
       reset: this.reset,
     }
+
+    this.currentResult = result as MutationObserverResult<
+      TData,
+      TError,
+      TVariables,
+      TContext
+    >
   }
 
   private notify(options: NotifyOptions) {
--- a/src/react/types.ts
+++ b/src/react/types.ts
@@ -2,13 +2,13 @@ import { RetryValue, RetryDelayValue } from '../core/retryer'
 import {
   InfiniteQueryObserverOptions,
   InfiniteQueryObserverResult,
-  MutateOptions,
-  MutationStatus,
+  MutationObserverResult,
   MutationKey,
   QueryObserverOptions,
   QueryObserverResult,
   QueryKey,
   MutationFunction,
+  MutateOptions,
 } from '../core/types'
 
 export interface UseBaseQueryOptions<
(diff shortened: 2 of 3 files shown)


src/core/mutationObserver.ts
```diff
@@ -137,7 +137,7 @@ export class MutationObserver<
       isIdle: state.status === 'idle',
       mutate: this.mutate,
       reset: this.reset,
-    }
+    } as MutationObserverResult<TData, TError, TVariables, TContext>
```
TkDodo, line 140: could we maybe do the same thing here as we do in the queryObserver, which is: declare the type as  `MutationObserverBaseResult` and then only cast it after that, see:

https://github.com/tannerlinsley/react-query/blob/f9b23fcae9c5d45e3985df4519dd8f78a9fa364e/src/core/queryObserver.ts#L544-L569

I think this would give a bit more saftey because things we set here would need to confirm to the BaseResult type, which I think they don't need with a pure type assertion

TkDodo (changes requested):

src/react/useMutation.ts
```diff
@@ -97,9 +93,9 @@ export function useMutation<
   }, [])
 
   const mutate = React.useCallback<
-    UseMutateFunction<TData, TError, TVariables, TContext>
+    MutateFunction<TData, TError, TVariables, TContext>
   >((variables, mutateOptions) => {
-    obsRef.current!.mutate(variables, mutateOptions).catch(noop)
+    return obsRef.current!.mutate(variables, mutateOptions).catch(noop)
   }, [])
```
TkDodo, line 99: why this change please? It's per design that `mutate` doesn't return anything...

TkDodo (changes requested):

src/react/useMutation.ts
raoufswe: MutationFunction (core/types) and UseMutateFunction (react/types) are quite identical. The only difference is that MutationFuncation returns Promise<TData>, and useMutationFunction returns void. So I assumed that MutationFunction is more accurate since it's from the core package. 

If this isn't true, I'm happy to revert the change

TkDodo: yes please. `.mutate` shouldn't return a promise. we've had that in v2 and it was confusing for many, which is why we added a dedicated `mutateAsync` for that case.

TkDodo: what's the status here please? There seem to be conflicts now with the new release we just made, please have a look at resolving them :)

raoufswe: sorry to keep this on hold, Dominik

I have reverted the .mutate change and solved the conflict, could you please have a look now?

thanks for your time.

TkDodo (approved):

TkDodo:
hmm, for some reason, the type narrowing doesn't seem to work in the codesandbox preview builds. Can you have a look?

https://codesandbox.io/s/tannerlinsley-react-query-basic-typescript-forked-6zf7r?file=/src/index.tsx

src/react/types.ts
```diff
@@ -119,24 +119,22 @@ export type UseMutateAsyncFunction<
...
-  failureCount: number
-  isError: boolean
-  isIdle: boolean
-  isLoading: boolean
-  isPaused: boolean
-  isSuccess: boolean
+> = Exclude<
+  MutationObserverResult<TData, TError, TVariables, TContext>,
+  'mutate'
+> & {
```
raoufswe, line 130: @TkDodo I'm trying to override the type of mutate here and define it with UseMutateFunction so I can fix this type error

Screenshot 2021-09-08 at 11 09 12 PM

I tried to used Omit, but it broke the union in MutationObserverResult and I can't change it to interface and then extend as MutationObserverResult is a type 

do you have any idea what the best approach would be here?

TkDodo (approved):

PR #2622, opened 2021-08-31
feat(useQuery): make refetchInterval accept a function

TkDodo:
so that we can have intervals dependent on current data / error


src/core/queryObserver.ts
amerllica, line 1: I know a few years have passed and it's too late to say this, but using `refetchInterval` means using `setInterval` and it means it is `short-polling` not a `long-polling`.

The long-polling uses the Promise answer to set the next call, and it seems RQ hasn't received it yet.

cc: @TkDodo

PR #2623, opened 2021-09-01
Update query-functions.md Handle Error Section

k0d3d:
I was looking for an example of how to handle errors thrown from a query functions when using useQuery . This page was top among search results. The code example under the 'Handling and Throwing Errors' section does not clearly show how to handle errors. I am adding this bit of code for any one else who lands on this page looking for an answer.

### DIFF
--- a/docs/src/pages/guides/query-functions.md
+++ b/docs/src/pages/guides/query-functions.md
@@ -28,7 +28,7 @@ const { error } = useQuery(['todos', todoId], async () => {
   }
 
   return data
-})
+}, { onError: (error) => handleError(error)})
 ```
 
 ## Usage with `fetch` and other clients that do not throw by default


TkDodo:
hmm, the question is _how_ you handle errors. I have a blog post in the pipeline on that topic, but in short: if you add the `onError` callback here, it will execute for every observer, so if you use the same query multiple times, it will be executed multiple times. Also, that page wanted to show that if you reject the Promise (or throw in an async function) in the queryFn, it will have the `error` in the error property returned by useQuery.

So depending on what error "handling" means for you, it could also be:

```
const { error } = useQuery(...)

if (error) <div>error: {error} </div>
```

that also handles the error...

now if you want to show a toast notification or logging the error or send the result to an error tracking tool like sentry, that's a different topic, but as said above, this callback is also likely not the right place for that, because it is executed multiple times. There is an `onError` callback on the `QueryCache` that is better suited for that.

k0d3d:
Ill go ahead and close this.

PR #2630, opened 2021-09-03
feat(ssr): support SSR in Deno

justjavac:
I use react-query in Deno:

```tsx
import { QueryClient, QueryClientProvider, useQuery } from 'https://esm.sh/react-query';

export default function App() {
  return (
      <Example />
    </QueryClientProvider>
  );
}

function Example() {
  const { isLoading, data } = useQuery('repoData', () => Promise.resolve({ name: 'foo' }));

  if (isLoading) return 'Loading...';

  return <div>{data.name}</div>;
}
```

It output a warning:

> This browser doesn't support requestAnimationFrame. Make sure that you load a polyfill in older browsers. https://reactjs.org/link/react-polyfills
> This browser doesn't support cancelAnimationFrame. Make sure that you load a polyfill in older browsers. https://reactjs.org/link/react-polyfills

### DIFF
--- a/src/core/utils.ts
+++ b/src/core/utils.ts
@@ -72,7 +72,7 @@ export type QueryStatusFilter = 'all' | 'active' | 'inactive' | 'none'
 
 // UTILS
 
-export const isServer = typeof window === 'undefined'
+export const isServer = typeof window === 'undefined' || 'Deno' in window
 
 export function noop(): undefined {
   return undefined
--- a/src/core/utils.ts
+++ b/src/core/utils.ts
@@ -72,7 +72,10 @@ export type QueryStatusFilter = 'all' | 'active' | 'inactive' | 'none'
 
 // UTILS
 
-export const isServer = typeof window === 'undefined' || 'Deno' in window
+export const isServer =
+  typeof window === 'undefined' ||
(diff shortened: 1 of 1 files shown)


src/core/utils.ts
```diff
@@ -72,7 +72,7 @@ export type QueryStatusFilter = 'all' | 'active' | 'inactive' | 'none'
 
 // UTILS
 
-export const isServer = typeof window === 'undefined'
+export const isServer = typeof window === 'undefined' || 'Deno' in window
```
TkDodo, line 75: Is this the best / only way to find out if you’re running in deno? What if a `'Deno'` object exists for some other reason on the window?

also, according to this stackoverflow, it seems to also be there for code run on the client if it was compiled by deno: 

https://stackoverflow.com/questions/65095009/how-to-detect-in-typescript-code-if-its-run-in-deno-or-in-browser-compiled-by-d

justjavac: What about `!!(window.Deno && Deno.version && Deno.version.deno)`?

TkDodo: yeah I guess that would be safe enough

Issue #1675, opened 2021-01-20
Unable to type useQueries options or results without casting


matthewdavidrodgers:
**Describe the bug**
While the arguments and return value of useQuery can clearly be typed and even inferred via generics, useQueries offers no way to provide types that extend the default type - meaning what would normally be generic in useQuery is `unknown` in useQueries, and thus cannot be typed correctly (without an unsafe cast).

To illustrate more clearly:
This is the signature of useQuery
https://github.com/tannerlinsley/react-query/blob/ead2e5dd5237f3d004b66316b5f36af718286d2d/src/react/useQuery.ts#L9-L15
Notice how the generic `TQueryFnData` is used in both the type of the options and the result. Thus, if your options contain a queryFn that returns a `Promise<string>`, TypeScript knows via inference that the return value will (when settled successfully) have a data property that is a `string`.

Contrast this to the signature of useQueries
https://github.com/tannerlinsley/react-query/blob/ead2e5dd5237f3d004b66316b5f36af718286d2d/src/react/useQueries.ts#L8
There is no api to provide info about the types that these individual queries will be dealing with. UseQueryOptions has generics, as we can see from useQuery, but without any way to provide them, they all default to `unknown`. This also goes for UseQueryResult; its generics also default to `unknown`, meaning the `data` property of each settled successful result is typed as `unknown`.

**To Reproduce**
As this is a compile-time issue, I've provided a minimal example of the issue:
```tsx
import { useQueries, UseQueryOptions, UseQueryResult } from 'react-query';

function stringQueryor(): Promise<string> {
  return Promise.resolve("hello there!");
}

// While this compiles correctly, the return type of this function is
// UseQueryResult<unknown, unknown, unknown>[]
// Even though the queryFns are explicitly typed to strings,
// the data property of each result (when settled successfully)
// is unknown
function inferenceOnlyUseQueries() {
  return useQueries([
    { queryKey: "stringOne", queryFn: stringQueryor },
    { queryKey: "stringTwo", queryFn: stringQueryor }
  ]);
}

// This does NOT compile.
// TypeScript reports the error below:
//
// Type 'QueryObserverResult<unknown, unknown>[]' is not assignable to type 'QueryObserverResult<string, unknown>[]'.
//  Type 'QueryObserverResult<unknown, unknown>' is not assignable to type 'QueryObserverResult<string, unknown>'.
[...]

PR #2634, opened 2021-09-03
feat(types): enable types in useQueries

artysidorenko:
Hi folks, thanks for the great library, have been using it a lot recently!

I've had a stab at issue #1675 , taking inspiration from PR #1527 but with a slightly different approach:

- `useQueries` gets a type parameter in the shape of _some_ array (representing the data types returned from the `queryFn`). This is inferred from the passed-in `queries` array.
- the queries array itself is a typed as a mapped generic tuple, so the individual elements' types are preserved in the correct order.

It also required tweaking some of the options property types slightly - namely `initialData` and `placeholderData`, to enforce the type that was set on the `queryFn`. Adding some comments to the PR diff to explain further.

I think it should be backwards-compatible.

Let me know what you think.

### DIFF
--- a/src/react/tests/useQueries.test.tsx
+++ b/src/react/tests/useQueries.test.tsx
@@ -1,7 +1,13 @@
 import { waitFor } from '@testing-library/react'
 import React from 'react'
 
-import { queryKey, renderWithClient, setActTimeout, sleep } from './utils'
+import {
+  expectType,
+  queryKey,
+  renderWithClient,
+  setActTimeout,
+  sleep,
+} from './utils'
 import { useQueries, QueryClient, UseQueryResult, QueryCache } from '../..'
 
 describe('useQueries', () => {
@@ -234,5 +240,148 @@ describe('useQueries', () => {
       { status: 'success', data: 12, isPreviousData: false, isFetching: false },
       { status: 'success', data: 18, isPreviousData: false, isFetching: false },
     ])
+  })
+
+  it('should return the correct types', async () => {
+    const key1 = queryKey()
+    const key2 = queryKey()
+    const key3 = queryKey()
+
+    // @ts-ignore
+    // eslint-disable-next-line
+    function Page() {
+      // unspecified query functions should default to unknown
+      const noQueryFn = useQueries([
+        {
+          queryKey: key1,
+        },
+        {
+          queryKey: key2,
+        },
+      ])
+      expectType<unknown>(noQueryFn[0].data)
+      expectType<unknown>(noQueryFn[0].error)
+      expectType<unknown>(noQueryFn[1].error)
+      expectType<unknown>(noQueryFn[1].error)
+
+      // it should infer the result type from the query function
+      const fromQueryFn = useQueries([
+        {
+          queryKey: key1,
+          queryFn: () => 1,
+        },
+        {
+          queryKey: key2,
+          queryFn: () => 'string',
+        },
+        {
+          queryKey: key3,
+          queryFn: () => ['string[]'],
+        },
+      ])
+      expectType<number | undefined>(fromQueryFn[0].data)
+      expectType<string | undefined>(fromQueryFn[1].data)
+      expectType<string[] | undefined>(fromQueryFn[2].data)
+
+      // it should enforce the initialData's type
+      useQueries([
+        {
+          queryKey: key1,
+          queryFn: () => 1,
+          initialData: 2,
+        },
+        {
+          queryKey: key2,
+          queryFn: () => 'string',
+          initialData: 'another string',
+        },
+        {
+          queryKey: key3,
+          queryFn: () => 'string',
+          // @ts-expect-error
+          initialData: 2,
+        },
+      ])
+
+      // it should enforce the placeholderData's type
+      useQueries([
+        {
+          queryKey: key1,
+          queryFn: () => 1,
+          placeholderData: 2,
+        },
+        {
+          queryKey: key2,
+          queryFn: () => 'string',
+          placeholderData: 'another string',
+        },
+        {
+          queryKey: key3,
+          queryFn: () => 'string',
+          // @ts-expect-error
+          placeholderData: 2,
+        },
+      ])
+
+      // it should enforce the select function's parameter type
+      useQueries([
+        {
+          queryKey: key1,
+          queryFn: () => 1,
+          select: (_number: number) => null,
+        },
+        {
+          queryKey: key2,
+          queryFn: () => 'string',
+          select: (_string: string) => null,
+        },
+        {
+          queryKey: key3,
+          queryFn: () => 1,
+          // @ts-expect-error
+          select: (_number: string) => null,
+        },
+      ])
+
+      // it should enforce the onSettled function's parameter type
+      useQueries([
+        {
+          queryKey: key1,
+          queryFn: () => 1,
+          onSuccess: (_number: number) => null,
+        },
+        {
+          queryKey: key2,
+          queryFn: () => 'string',
+          onSuccess: (_string: string) => null,
+        },
+        {
+          queryKey: key3,
+          queryFn: () => 1,
+          // @ts-expect-error
+          onSuccess: (_number: string) => null,
+        },
+      ])
+
+      // it should enforce the onSettled function's parameter type
+      useQueries([
+        {
+          queryKey: key1,
+          queryFn: () => 1,
+          onSettled: (_number: number | undefined) => null,
+        },
+        {
+          queryKey: key2,
+          queryFn: () => 'string',
+          onSettled: (_string: string | undefined) => null,
+        },
+        {
+          queryKey: key3,
+          queryFn: () => 1,
+          // @ts-expect-error
+          onSettled: (_number: string | undefined) => null,
+        },
+      ])
+    }
   })
 })
--- a/src/react/useQueries.ts
+++ b/src/react/useQueries.ts
@@ -2,10 +2,29 @@ import React from 'react'
 
 import { notifyManager } from '../core/notifyManager'
 import { QueriesObserver } from '../core/queriesObserver'
+import { InitialDataFunction, PlaceholderDataFunction } from '../core/types'
 import { useQueryClient } from './QueryClientProvider'
 import { UseQueryOptions, UseQueryResult } from './types'
 
-export function useQueries(queries: UseQueryOptions[]): UseQueryResult[] {
+type InferredQueryOptions<TQueryFnData> = Omit<
+  UseQueryOptions<TQueryFnData, unknown, TQueryFnData>,
+  'select' | 'initialData' | 'placeholderData'
+> & {
+  select?: (data: TQueryFnData) => unknown
+  initialData?: TQueryFnData extends infer TInitialData
+    ? TInitialData | InitialDataFunction<TInitialData>
+    : never
+  placeholderData?: TQueryFnData extends infer TPlaceholderData
+  ? TPlaceholderData | PlaceholderDataFunction<TPlaceholderData>
+  : never
+}
+
+type QueriesOptions<T> = { [E in keyof T]: InferredQueryOptions<T[E]> }
+type QueriesResults<T> = { [E in keyof T]: UseQueryResult<T[E]> }
+
+export function useQueries<TQueriesFnData extends unknown[]>(
+  queries: [...QueriesOptions<TQueriesFnData>]
+): QueriesResults<TQueriesFnData> {
   const mountedRef = React.useRef(false)
   const [, forceUpdate] = React.useState(0)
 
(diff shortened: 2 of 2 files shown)


src/react/useQueries.ts
```diff
@@ -2,10 +2,29 @@ import React from 'react'
...
+  UseQueryOptions<TQueryFnData, unknown, TQueryFnData>,
+  'select' | 'initialData' | 'placeholderData'
+> & {
+  select?: (data: TQueryFnData) => unknown
+  initialData?: TQueryFnData extends infer TInitialData
+    ? TInitialData | InitialDataFunction<TInitialData>
+    : never
+  placeholderData?: TQueryFnData extends infer TPlaceholderData
+  ? TPlaceholderData | PlaceholderDataFunction<TPlaceholderData>
+  : never
```
artysidorenko, line 19: Using conditional type with inference, to check that the type of `initialData` and `placeholderData` corresponds to the same type as returned from the `queryFn`.

Without this, if you added one of these properties with a different type from the `queryFn`, TS would merge the type definitions into a union 😢 

For example, the below would make TS think that `TQueryFnData` was `string | number` instead of just `string`

```typescript
useQueries([
  {
    queryKey: key3,
    queryFn: () => 'string',
    initialData: 2,
  },
])
```

TkDodo: Hmm but why would select return unknown? We don’t get type inference of the returned data then depending on select, right?

TkDodo:
Thank you for working on this 🙏 

prettier formatting seems to be off, which is why the pipeline fails, please have a look.

Also, I played around with this a bit, here are my findings:

- input passed to `useQueries` should be `readonly`, otherwise, I can't pass tuples created by const assertions to it, e.g. this fails:

```
useQueries([
    {
      queryKey: "key1",
      queryFn: () => "string",
    }
  ] as const);
```
- it's awesome that tuple inputs and positions are preserved 🚀 
- as said above, `select` doesn't work:

```
  const [shouldBeString] = useQueries([
    {
      queryKey: "key2",
      queryFn: () => 1,
      initialData: 1,
      select: data => String(data)
    },
  ]);
```
result type should be an Observer of `string`, but it still says `QueryObserverResult<number>` 
- if I input two different `select` options, typeScript complains that they are both implicit any:

```
useQueries([
    {
      queryKey: "key2",
      queryFn: () => 1,
      initialData: 1,
      select: data1 => String(data1)
    },
    {
      queryKey: "key2",
      queryFn: () => "1",
      initialData: "1",
      select: data2 => Number(data2)
    },
  ]);
```

here, `data1` and `data2` are both implicit any. It works with just one element though 🤔 

- 🚨 it seems like I can't use dynamic lengths, e.g. mapping over an array, which is the main use-case of `useQueries`:

```
 const queries = useQueries(
    myInput.map((i) => [
      {
        queryKey: ["key", i] as const,
        queryFn: () => i + 10,
      }
    ])
  );
```

this errors with: 

```
Type '{ queryKey: readonly ["key", number]; queryFn: () => number; }[]' has no properties in common with type 'InferredQueryOptions<unknown>'
```

src/react/useQueries.ts
artysidorenko: aha - that was a knowledge gap on my part; I misunderstood `select` completely.. I think I get it now though (its return type should take precedence over `queryFn` when it's defined?)

artysidorenko:
Thanks for the feedback; I'm gunna have to think about the solution a bit more as it sounds like I had only covered the set of `TQueryFnData` types, and need to work out the `TData` types as another dimension (that missing `TData` is related to those other issues you mentioned)

Experimenting with a second type parameter of equal length but I may need to do some work to pass the types through the `QueriesObserver`

```typescript
function useQueries<
  TQueriesFnData extends unknown[],
  TQueriesData extends {[E in keyof TQueriesFnData] : TQueriesData[E]},
>(
  queries: [...QueriesOptions<TQueriesFnData, TQueriesData>]
)
```

Will keep you posted; it's an interesting puzzle 😄

src/react/useQueries.ts
TkDodo: What you return from select overrides what the queryFn usually returns from useQueries, but keep in mind that placeholderData and initialData still need to be of type TQueryFnData, because they also run through select

artysidorenko:
Hi again,

Sorry took a while to explore. I think I've come up with something that's workable. Originally I had hoped I could do it with a single mapped type, however, because it needs to track multiple types that can be different for each element, I've ended up having to write a bunch of helpers to recursively cycle through the array 1 entry at a time (there's _a lot_ of these nested "X extends infer Y?" statements unfortunately)

**Anyways, the good news:**

- it infers all three types (fnData/error/data) for each element if no type param is provided. There is validation; the one piece that isn't working 100% smoothly is the callback parameter types - see below, my response to your `noImplicitAny` comment.

- you can also add explicit type params, either as a nested array, something like `useQueries<[[TQueryFnData1, TError1, TData1], [TQueryFnData2, TError2, TData2]]>`, or as an array of objects `useQueries<[{queryFnData: X, error: Y, data: Z}, {queryFnData: A, error: B, data: C}]>. I've kept both options here to show what they look like. My preference would be to keep the nested array one only, which will shave off a bit of the code/complexity.

- fyi there's a depth limit of 20, so if your array has 20+ entries, it will give up and return the usual unknown `UseQueryOptions[]`. Without this limit TS would give a `Type instantiation is excessively deep and possibly infinite.` error after ~25 entries.

Let me know what you think. There's a bit of complexity, but I've tried to add tests to cover all the possibilities; it should be pretty stable unless the underlying types change significantly? (in which case we'd need to update the utility types here as well).

**Covering last week's comments specifically:**

> input passed to useQueries should be readonly, otherwise, I can't pass tuples created by const assertions to it

Think this should be working now

> as said above, select doesn't work

Updated it to discriminate between `queryFn` / `select`

> if I input two different select options, typeScript complains that they are both implicit any

I'm afraid I've hit a wall on this `noImplicitAny` issue. I'm able to type and validate the callbacks, but it won't allow you to pass the argument without a type. (same applies to `onSuccess` / `onSettled`)

```ts
useQueries([
  {
    queryKey: "key1",
    queryFn: () => 1,
    initialData: 1,
    select: data1 => String(data1) // TS will complain about noImplicitAny
  },
  {
    queryKey: "key2",
    queryFn: () => 1,
    initialData: 1,
    select: (data1: number) => String(data1) // TS will allow this, and my IDE even suggests the number type on hover, but I need to add it explicitly to avoid the noImplicitAny
  },
  {
    queryKey: "key3",
    queryFn: () => "1",
    initialData: "1",
    select: (data2: number) => data2 // TS will complain as it should
  },
]);
```

> it seems like I can't use dynamic lengths, e.g. mapping over an array

I believe Array.map / similar array methods all return arrays rather than a tuples, so this process can't handle them like array literals. However I've updated it so that if it can't infer a tuple, it will revert to a regular unknown `UseQueryOptions[]` as before.

src/react/useQueries.ts
artysidorenko: (latest commit should handle this now 🤞 )

src/react/useQueries.ts
```diff
@@ -1,11 +1,98 @@
...
+    ? UseQueryOptions<TQueryFnData, TError>
+    : T extends { data: infer TData; error?: infer TError }
+    ? UseQueryOptions<unknown, TError, TData>
+    : // Map params from tuple [TQueryFnData, TError, TData]
+    T extends [infer TQueryFnData, infer TError, infer TData]
+    ? UseQueryOptions<TQueryFnData, TError, TData>
+    : T extends [infer TQueryFnData, infer TError]
+    ? UseQueryOptions<TQueryFnData, TError>
+    : T extends [infer TQueryFnData]
+    ? UseQueryOptions<TQueryFnData>
```
artysidorenko, line 27: this can be slashed in half if we keep one of the two type parameter structures

src/react/useQueries.ts
```diff
@@ -1,11 +1,98 @@
...
+  Depth extends ReadonlyArray<number> = []
+> = Depth['length'] extends MAXIMUM_DEPTH
+  ? UseQueryOptions[]
+  : T extends []
+  ? []
+  : T extends [infer Head]
+  ? [...Result, GetOptions<Head>]
+  : T extends [infer Head, ...infer Tail]
+  ? QueriesOptions<[...Tail], [...Result, GetOptions<Head>], [...Depth, 1]>
+  : T
```
artysidorenko, line 77: Just to add a bit more context, why I ended up going down this recursive-array-map route...

Originally I had hoped I could do something a bit more elegant with two array type parameters `TQueryFnData[]` and `TData[]` mapped by a single operation, but ran into a brick wall as I can't index them with a single key. (or if I try and force it, [it breaks the tuple magic](https://stackoverflow.com/questions/64138789/why-are-typescript-mapped-tuple-types-behaving-differently-when-supplying-generi))

Idea 1:
```ts
// ERROR: "E cannot be used to index TData"
type QueriesOptions<TQueryFnData, TData> = { [E in keyof TQueryFnData]: InferredQueryOptions<TQueryFnData[E], TData[E]> }

useQueries<TQueryFnData extends unknown[], TData extends unknown[]>(queries: [...QueriesOptions<TQueryFnData, TData>])
```

Idea 2
```ts
// this can be made to work, but QueriesOptions is no longer an array-type
type QueriesOptions<TQueryFnData, TData> = keyof TData extends keyof TQueryFnData ? { [E in keyof TQueryFnData]: InferredQueryOptions<TQueryFnData[E], TData[E]> } : never

// ERROR: "a rest parameter must be of an array type"
useQueries<TQueryFnData extends unknown[], TData extends unknown[]>(queries: [...QueriesOptions<TQueryFnData, TData>])
```

artysidorenko:
Clicking the re-request review button just in case there was no notification sent out with the last updates I pushed (should have done this sooner but was travelling last couple of weeks). Any feedback let me know.

TkDodo:
sorry @artysidorenko, it really slipped past me and I didn't get any notifications about the new commits :(

I'll try to have a look this week

src/react/tests/useQueries.test.tsx
```diff
@@ -235,4 +248,363 @@ describe('useQueries', () => {
...
     ])
   })
+
+  it('handles type parameter - tuple of tuples', async () => {
+    const key1 = queryKey()
+    const key2 = queryKey()
+    const key3 = queryKey()
+
+    // @ts-ignore
+    // eslint-disable-next-line
```
TkDodo, line 258: why is that necessary please?

src/react/tests/useQueries.test.tsx
```diff
@@ -235,4 +248,363 @@ describe('useQueries', () => {
...
+    function Page() {
+      // Array.map does not throw error
+      const result1 = useQueries(
+        Array(10).map(i => ({
+          queryKey: ['key', i] as const,
+          queryFn: () => i + 10,
+        }))
+      )
+      expectType<QueryObserverResult<unknown, unknown>[]>(result1)
```
TkDodo, line 508: should this not be able to infer to `<number, unknown>` ?

TkDodo (commented):
Thanks again for working on this, and sorry again that I missed your commits.

- ✅ const assertions now work
- ✅ select now works
- 🟡 it's a bit unfortunate that `data` passed to `select` is implicitly any, e.g. here:
```
[
    {
      queryKey: "key2",
      queryFn: () => 1,
      select: (data) => String(data)
    }
  ]
```
but I guess this isn't an easy fix?
- ✅ success and error look good, even though we have to type it explicitly. I guess same restriction as with select?
- 🔴 as commented on the test, `useQueries` with an `array.map` infers to `<unknown, unknown>`. I think getting inference here would be crucial

TkDodo:
I think that overload is a good compromise. But please also amend the TypeScript docs that currently say you need 3.8, to something that says that 4.1 is needed if you want proper typings for useQueries.

The important part is that users with 3.8 will not get any type errors in their code if they are using useQueries at the moment, and especially if they are only using useQuery :)

artysidorenko:
Aha, so I did a bit more research (I wasn't very happy with the compromise I suggested after trying it out - it was looking quite fiddly trying to get the overload right, and there is enough complexity in there as it is 😅)... I found that we can get full backwards-compatibility after all:

Added a directive in `package.json` to tell older TS versions to pick up a compatible `useQueries` definition. It means an extra file in your src directory, but I figured you could also leverage this if any similar b/c issues come up in the future.

Let me know if it works for you. Added a snippet to the docs as well (let me know if it's worth adding something extra in the community section as well, to cover a bit more how to add the explicit type parameters?)

bennettdams:
@artysidorenko I just wanted to take a moment to say thank you. The use of `as` gives me anxiety and with this PR I can remove the last occurrence of it in my codebase.

PR #2637, opened 2021-09-05
feat(useQuery): add reset method to returned useQuery object

TkDodo:


juanPorti:
Hello @TkDodo, is there any update on the state of this PR? Seems it was validated but not merged, is there anything left on which I can contribute to push this functionality forward? Thanks!

TkDodo:
I'm not sure we should merge this tbh. The implementation is not the same as `queryClient.resetQueries`. I just called `refetch` directly, but that would bypass queries that are `enabled: false` for example.

Apart from "we have it for everything else", what's the use-case of getting `reset` returned from useQuery?

juanPorti:
Hi again @TkDodo and thanks for the fast response :) 

Use-case would be when a controlled error happens in the application, for example when a request fails and we want to reset the query without an error boundary. Also to avoid importing `useQueryClient` all over the application for it. 

This is the implementation we have right now:

`const useResettableQuery = (...args) => {
  const query = useQuery(...args)
  
  query.reset = () => {
    query.remove()
    query.refetch()
  }

  return query
}`

Do we need to respect the `enabled: false` when calling explicitly a `reset()` method?

What would be the advantage of using `queryClient.resetQueries` against this `reset()` method? 

Thanks again for the time and work!

TkDodo:
> Use-case would be when a controlled error happens in the application, for example when a request fails and we want to reset the query without an error boundary.

I don't quite understand that. If an error happens because a request fails, the query goes in to error state. You can decide if you show that error or not. If you `reset` the query, it will just `refetch` immediately. If you want to show a refetch button, I think you can just call `refetch()` because a query that has no data (because it error'd) will just go to hard loading state again ...

> Do we need to respect the enabled: false when calling explicitly a reset() method?

yes, both ways (`queryClient.resetQueries` and `reset`) need to behave the same.

juanPorti:
> I think you can just call refetch() because a query that has no data (because it error'd) will just go to hard loading state again

This assertion has worked for me, didn't know that a query that has error'd forces the loading state of it. Looked around the documentation but couldn't find a [reference to it](https://react-query.tanstack.com/reference/useQuery#:~:text=the%20query%20succeeds.-,refetch,-%3A%20(options%3A%20%7B%20throwOnError%3A%20boolean)
Anyways, thanks again for the time! :)

TkDodo:
> Looked around the documentation but couldn't find a reference to it

A query that is `fetching` and has no data yet is in `loading` state - that's how loading state is defined. That it has an error or not is irrelevant for that definition :)

As it stands, I'm gonna close that PR because I don't see a good use-case for it, and it's not trivial to implement correctly for all cases.

PR #2643, opened 2021-09-07
fix(docs): use hydration utilities from core

Andrey-Bazhanov:


TkDodo:
@Ephem after moving the hydration to the core, what's the preferred way of importing the hydration utils:

```
import { Hydrate } from 'react-query/hydration'
import { Hydrate } from 'react-query'
```

I guess it's the latter, and we only kept the `/hydration` imports for backwards compatibility, right?

Andrey-Bazhanov:
@TkDodo according to this, I think yes https://github.com/tannerlinsley/react-query/blob/master/src/hydration/index.ts

Ephem:
Yes, exactly!

I was thinking about when would be a good time to update the docs, if we do it right now and people read them but are not on the latest version yet (pretty likely) I thought maybe that would be confusing?

On the other hand, maybe we should update them right away to not lead people down a later deprecated track. If we do, should we add a disclaimer about the old way and mention the version it got moved to core maybe?

TkDodo:
> If we do, should we add a disclaimer about the old way and mention the version it got moved to core maybe?

yes, that sounds like a good idea. Also, I think this split in the API Reference is now not needed anymore, and probably not a good idea either:

Screenshot 2021-09-10 at 11 19 43

I think we should have one page there called `Hydration` that shows the new exports with the disclaimer that old exports still exist. What do you think please?

Ephem:
I dont have strong opinions on that really. Either we group them on a single page or just remove the `hydration/`-prefix but keep separate pages.

Looks like the rest of the docs under API are atomic, in that every separate function/class has their own page. I guess that makes a function easy to look up, but that should be easy even if they are grouped? 🤷

TkDodo:
yes, the search will work as well if we group it and user headers. for example, not every method of the `queryClient` has its own page, so why should `hydrate` and `dehydrate` have one 🤷 . So I'm more for the grouping

Ephem:
Yeah, I agree. I guess technically the queryClient methods are on a class which the hydrate stuff is not, but I don't think that's an important distinction. Makes sense to me to group it.

Andrey-Bazhanov:
@TkDodo @Ephem I grouped hydration pages into single page and put notice about versions. Please take a look.

TkDodo (approved):

PR #2674, opened 2021-09-16
Don't reset status to loading when refetching query that has never succeeded

### LINKED ISSUE #2673: Query status resets to "loading" when refetching query that has never succeeded
**Describe the bug**
When a successful query (with status "success") is being refetched its status remains as "success".

If a query has never succeeded, however, its status resets from "error" to "loading" whenever data is being refetched. I would have expected a similar behavior to a successful query.

This is also a problem during initial rendering in SSR. If a prefetched query fails, the status becomes "loading" when the server renders the React app.

**To Reproduce**
I made a reduced testcase on https://codesandbox.io/s/react-query-error-loading-status-bug-dxmhi

The testcase itself (with steps to reproduce) can be viewed directly on https://dxmhi.sse.codesandbox.io/

**Expected behavior**
Status "error" should not reset to "loading" when React Query attempts to refetch the query.

**Screenshots**
N/A

N/A

**Smartphone (please complete the following information):**
N/A

**React Query version:**
v3.23.2

**Additional context**
The same issue was identified by @rvil in https://github.com/tannerlinsley/react-query/discussions/1208#discussioncomment-181742

The [documentation](https://react-query.tanstack.com/reference/useQuery) says status is
> `loading` if the query is in a "hard" loading state.
> This means there is no cached data and the query is currently fetching, eg `isFetching === true`"

The way the documentation is written could indicate the behavior I see is intentional since it only refers to cached data, but I would argue it is such non-obvious behavior, especially during SSR, that it should be considered a bug.

The same bug was reported and fixed [back in 2020](https://github.com/tannerlinsley/react-query/issues/529). There is no associated PR so I was unable to find the original fix, or verify that the fix ever made it into a release.

esseb:
Fixes #2673

### DIFF
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -547,7 +547,10 @@ export class Query<
           fetchMeta: action.meta ?? null,
           isFetching: true,
           isPaused: false,
-          status: !state.dataUpdatedAt ? 'loading' : state.status,
+          status:
+            !state.dataUpdatedAt && !state.errorUpdatedAt
+              ? 'loading'
+              : state.status,
         }
       case 'success':
         return {
--- a/src/core/queryObserver.ts
+++ b/src/core/queryObserver.ts
@@ -446,7 +446,7 @@ export class QueryObserver<
 
       if (fetchOnMount || fetchOptionally) {
         isFetching = true
-        if (!dataUpdatedAt) {
+        if (!dataUpdatedAt && !errorUpdatedAt) {
           status = 'loading'
         }
       }


esseb:
This PR is very likely incomplete, but the two changes I made in this PR fixes the issue I see https://github.com/tannerlinsley/react-query/issues/2673

I think there are more locations where similar logic should be updated, I'll need some help to make a more robust PR.

The `src/react/tests/QueryResetErrorBoundary.test.tsx` testcase fails with my changes, but I'm not sure why.

Issue #2187, opened 2021-04-23
suspense useQuery infinite loop when query goes from success to false and query key changes


flybayer:
**Describe the bug**
The workaround in https://github.com/tannerlinsley/react-query/issues/2157 fixed only part of the issue we are having in Blitz. We still have a case on user logout where we get infinite loops.

Failing test case included below. They key thing that makes it go into the infinite loop is the query key changing. If the key doesn't change, then it works fine.

**To Reproduce**

Add the following test to `src/react/tests/suspense.test.tsx`

```ts
  it('should not have infinite loop', async () => {
    const key = queryKey()

    const consoleMock = mockConsoleError()

    let succeed = true

    function Page() {
      const [nonce] = React.useState(0)
      const fullKey = `${key}-${succeed}`
      const result = useQuery(
        fullKey,
        async () => {
          await sleep(10)
          console.log('query', fullKey)
          if (!succeed) {
            throw new Error('Suspense Error Bingo')
          } else {
            return nonce
          }
        },
        {
          retry: false,
          suspense: true,
        }
      )
      return (
        <div>
          <span>rendered</span> <span>{result.data}</span>
          <button
            aria-label="fail"
            onClick={async () => {
              await queryClient.resetQueries()
            }}
          >
            fail
          </button>
        </div>
      )
    }

    function App() {
      const { reset } = useQueryErrorResetBoundary()
      return (
        <ErrorBoundary
          onReset={reset}
          fallbackRender={({ resetErrorBoundary }) => (
            <div>
              <div>error boundary</div>
              <button
                onClick={() => {
                  resetErrorBoundary()
                }}
              >
                retry
              </button>
            </div>
          )}
        >
          <React.Suspense fallback="Loading...">
            <Page />
          </React.Suspense>
        </ErrorBoundary>
      )
    }

    const rendered = renderWithClient(queryClient, <App />)

    await waitFor(() => rendered.getByText('Loading...'))
    await waitFor(() => rendered.getByText('rendered'))

    succeed = false
    fireEvent.click(rendered.getByLabelText('fail'))
    await waitFor(() => rendered.getByText('error boundary'))
    consoleMock.mockRestore()
  })
```

**Version**

React-query 3.13.9

flybayer:
cc @tannerlinsley @TkDodo

flybayer:
I could look into to fixing it, but not sure where to even start on this one

flybayer:
[...]

PR #2686, opened 2021-09-18
fix(core): make sure erroneous queries refetch when query keys change

### LINKED ISSUE #2682: error query doesn't fire again when keys change
### Discussed in https://github.com/tannerlinsley/react-query/discussions/2676

<sup>Originally posted by **jdupont22** September 16, 2021</sup>
Hi,

I've a strange situation, i don't know if it's normal!
I try to load some "offers" with `useQuery(["offers, deepJson])`. 

**I have a "staleTime=20s" and the default "cacheTime=5m".**

`useQuery(["offers",{"order":"desc","page":1,"q":null,"size":10,"sort":"name","states":["ACTIVE"]}])`

I've the following result, react-query retry 3 times and failed! It's ok.

Capture d’écran 2021-09-16 à 21 26 05

Now, I click on the tab "Inactive", that change the deepJson states to `"states":["INACTIVE"]`
I see in the react-query devtool the params change, but the request is not trigger!!!

**There is a "disabled" query with this keys, but i would imagine that a query in error will retry the next time? it's based on the cache expiration?**

Capture d’écran 2021-09-16 à 21 26 12

Can you explain to me this current behavior? When the query return 200 OK, switching tab work perfectly!

Thx

</div>

thanks for reporting. This seems to have worked correctly up until v3.19.4. I'll create an issue out of this.

Here is a slight variation of your codesandbox without window focus refetching, without retries, with a 1 second fetch delay and with devtools, so that we can better see when fetching transitions are happening: https://codesandbox.io/s/interesting-elbakyan-zz28f?file=/src/App.js

TkDodo:
the change in shouldFetchOptionally was introduced in #2556 to avoid an infinite loop wen using suspense (see #2187), but it broke the case where we're switching keys between two erroneous queries - they now didn't refetch even though they should

this fix scopes the error check to only usages where suspense is turned on, so that we get the same behaviour as in v3.19.4 for non-suspense users, while still having the fix for suspense users

fixes #2682


TkDodo:
@JaeYeopHan please have a look

JaeYeopHan (approved):
LGTM 👍

PR #2688, opened 2021-09-18
Add types for devtools

prateek3255:
Make devtools and all related helpers type safe

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,5 +130,9 @@
     "rollup-plugin-visualizer": "^4.0.4",
     "type-fest": "^0.21.0",
     "typescript": "^4.1.2"
+  },
+  "resolutions": {
+    "@types/react": "^16.9.41",
+    "@types/react-dom": "^16.9.8"
   }
 }
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
     "eslint-plugin-prettier": "^3.1.3",
     "eslint-plugin-promise": "^4.2.1",
     "eslint-plugin-react": "7.20.0",
-    "eslint-plugin-react-hooks": "4.0.4",
+    "eslint-plugin-react-hooks": "4.2.0",
     "eslint-plugin-standard": "^4.0.1",
     "is-ci-cli": "^2.1.1",
     "jest": "^26.0.1",
--- a/src/devtools/devtools.tsx
+++ b/src/devtools/devtools.tsx
@@ -336,7 +336,7 @@ export function ReactQueryDevtools({
                   bottom: '0',
                   left: '0',
                 }),
-            ...(toggleButtonStyle as any),
+            ...(toggleButtonStyle as unknown),
           }}
         >
           <Logo aria-hidden />
--- a/src/devtools/devtools.tsx
+++ b/src/devtools/devtools.tsx
@@ -1,8 +1,6 @@
-// @ts-nocheck
-
 import React from 'react'
 
-import { useQueryClient } from 'react-query'
+import { Query, useQueryClient } from 'react-query'
 import { matchSorter } from 'match-sorter'
 import useLocalStorage from './useLocalStorage'
 import { useIsMounted, useSafeState } from './utils'
@@ -71,6 +69,18 @@ interface DevtoolsPanelOptions {
    * The standard React className property used to style a component with classes
    */
   className?: string
+  /**
+   * A boolean variable indicating whether the panel is open or closed
+   */
+  isOpen: boolean
+  /**
+   * A function that toggles the open and close state of the panel
+   */
+  setIsOpen: (isOpen: boolean) => void
+  /**
+   * Handles the opening and closing the devtools panel
+   */
+  handleDragStart: (e: React.MouseEvent<HTMLDivElement, MouseEvent>) => void
 }
 
 const isServer = typeof window === 'undefined'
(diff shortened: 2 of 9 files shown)


TkDodo:
@prateek3255 wow, thank you so much. This was on my list for quite some time, but I never got to it 🚀 . It will make working with the devtools so much better in the future

src/devtools/devtools.tsx
```diff
@@ -204,13 +224,13 @@ export function ReactQueryDevtools({
     <Container ref={rootRef} className="ReactQueryDevtools">
       <ThemeProvider theme={theme}>
         <ReactQueryDevtoolsPanel
-          ref={panelRef}
+          ref={panelRef as any}
```
TkDodo, line 227: what error would you be getting here please?

src/devtools/useLocalStorage.ts
```diff
@@ -23,6 +28,7 @@ export default function useLocalStorage(key, defaultValue) {
     } else {
       setValue(initialValue)
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [defaultValue, key])
```
TkDodo, line 32: what would be expected here from the linter please, and why do we have to disable it?

src/devtools/devtools.tsx
prateek3255, line 227: Had to use `any` here because it was giving a weird error -
```
Type 'LegacyRef<HTMLDivElement>' is not assignable to type 'Ref<HTMLDivElement> | undefined'.
```

I tried looking for the solution but couldn't find anything

src/devtools/utils.ts
```diff
@@ -77,12 +105,12 @@ export function useIsMounted() {
...
  * or when the component is not mounted anymore.
  */
-export function useSafeState(initialState) {
+export function useSafeState<T>(initialState: T): [T, (value: T) => void] {
   const isMounted = useIsMounted()
   const [state, setState] = React.useState(initialState)
 
   const safeSetState = React.useCallback(
-    value => {
+    (value: T) => {
```
TkDodo, line 113: getting now:

> React Hook React.useCallback has a missing dependency: 'T'. Either include it or remove the dependency array

I think this can potentially be solved with updating the `eslint-plugin-react-hooks` to the latest version

src/devtools/devtools.tsx
```diff
@@ -245,14 +265,14 @@ export function ReactQueryDevtools({
           <Button
             type="button"
             aria-label="Close React Query Devtools"
-            {...otherCloseButtonProps}
-            onClick={() => {
+            {...(otherCloseButtonProps as any)}
```
prateek3255, line 268: Same reason as above here as well, it was giving the same error as above for the `ref` attribute

TkDodo (changes requested):
looks really good 👍 , just some small questions

prateek3255 (commented):
@TkDodo These are a couple of places I couldn't figure out what's wrong so had to end up using `any`

TkDodo:
>  One more thing I am not sure why prettier has added spaces at the beginning at some places

Hm, not sure either. But the `devtools` directory should be included in our prettier checks, as it's under `src`:

https://github.com/tannerlinsley/react-query/blob/afcba6066d238b9ef34386e6da4f372b3e6f7ca0/package.json#L42

so if the pipeline is not complaining, it should be fine :)

src/devtools/utils.ts
prateek3255: Fixed ✅

src/devtools/useLocalStorage.ts
prateek3255: It was the same dependency T issue, gone after upgrading `eslint-plugin-react-hooks`. Removed the comment ✅

src/devtools/devtools.tsx
prateek3255: I added the error I was facing in this comment -
https://github.com/tannerlinsley/react-query/pull/2688#discussion_r711621930

src/devtools/devtools.tsx
TkDodo: weird - I couldn't find out why that is. What you can do however is:

```
{...(otherCloseButtonProps as unknown)}
```

spreading `any` onto the Button will make all the other props _not_ typecheck anymore, so it's kind of a "leaking" any. with unknown, all other props will still type check, so it's at least non-invasive.

src/devtools/devtools.tsx
TkDodo: yeah, me neither. I guess it's fine for now 🤷

src/devtools/devtools.tsx
prateek3255: ✅

src/devtools/devtools.tsx
```diff
@@ -316,7 +336,7 @@ export function ReactQueryDevtools({
                   bottom: '0',
                   left: '0',
                 }),
-            ...toggleButtonStyle,
+            ...(toggleButtonStyle as any),
```
TkDodo, line 339: ah, there's another one:

```suggestion
            ...(toggleButtonStyle as unknown),
```

TkDodo (approved):
very good work 🚀

src/devtools/devtools.tsx
prateek3255: @TkDodo  it looks like there might be some package that is using an incompatible version of @types/react, Fixed it with [yarn resolutions](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/) for now as recommended by this comment - https://github.com/react-dropzone/react-dropzone/issues/1000#issuecomment-880533066

package.json
```diff
@@ -130,5 +130,9 @@
     "rollup-plugin-visualizer": "^4.0.4",
     "type-fest": "^0.21.0",
     "typescript": "^4.1.2"
+  },
+  "resolutions": {
+    "@types/react": "^16.9.41",
+    "@types/react-dom": "^16.9.8"
```
TkDodo, line 136: @prateek3255 I'm just stumbling upon this and I don't really know what this does or what this is for. I must have missed this when we merged this. Can you explain what this does or why we need it? And do we need to update it for v4 to react18 or something? Thanks!

prateek3255: @TkDodo From what I remember this was added because there was some bunch of type errors when adding DOM related types to devtools and its utils probably because of some conflicting `@types/react` of some other package. So adding resolutions helped resolve to the type version we wanted to use.

From what I see now is that we are not getting any type errors while compiling both on master and alpha if we remove this, so this can just be safely removed.

PR #2704, opened 2021-09-22
feat(quick-start): improved guide

### LINKED ISSUE #2669: [docs]: Add short description of queries and mutations in the Quick Start Guide
The docs have a quick start guide but the point of a quick guide is not just to provide code, it's also to give the users a simple overview with an example. 

Currently the quick start guide only has code and 3 links to docs about queries, mutations and query invalidation. I think we should add a short paragraph for each of the above which summarises their functionality in a way a new comer is able to understand.

Apart from that, we should also give a code walkthrough(not just through comments). In this, we should also explain what the `QueryClientProvider` does because there is no guidance about that.

I would be glad to take up this issue and fix it, I'm raising this issue to learn about your opinions and whether you also think that this is a good addition. Thanks for reading, looking forward to hearing your response!

akashshyamdev:
closes #2669

### DIFF
--- a/docs/src/pages/quick-start.md
+++ b/docs/src/pages/quick-start.md
@@ -73,8 +73,6 @@ In the `Todo` component, we start by accessing the queryClient we defined earlie
 A `query` listens to a source of data, in our case, the `getTodos` method.
 The unique key you provide is used internally for re-fetching, caching, and sharing your queries throughout your application.
 
-Queries are generally use with the `GET` HTTP method, mutations are used for HTTP methods like `POST`, `PUT` etc which modify data on the server.
-
 Finally, let's add a button that adds a todo:
 ```js
 import {
--- a/docs/src/pages/quick-start.md
+++ b/docs/src/pages/quick-start.md
@@ -24,7 +24,7 @@ function App() {
   return (
     // Provide the client to your App
     <QueryClientProvider client={queryClient}>
-      // we will define this later
+      {/* we will look at this later */}
       <Todos />
     </QueryClientProvider>
   )
--- a/docs/src/pages/quick-start.md
+++ b/docs/src/pages/quick-start.md
@@ -28,7 +28,7 @@ function App() {
 }
 ```
 
-Here we are creating a client by creating a new instance of the `QueryClient` class.
+We are creating a new instance of the `QueryClient` class.
 A QueryClient can be used to interact directly with the underlying caches.
 
 
--- a/docs/src/pages/quick-start.md
+++ b/docs/src/pages/quick-start.md
@@ -15,6 +15,7 @@ import {
   QueryClient,
   QueryClientProvider,
 } from 'react-query'
+import Todos from './Todos';
 
 // Create a client
 const queryClient = new QueryClient()
@@ -23,6 +24,8 @@ function App() {
   return (
     // Provide the client to your App
     <QueryClientProvider client={queryClient}>
+      // we will define this later
+      <Todos />
     </QueryClientProvider>
   )
 }
--- a/docs/src/pages/quick-start.md
+++ b/docs/src/pages/quick-start.md
@@ -50,7 +50,7 @@ import { getTodos, postTodo } from '../my-api'
 
 function Todos() {
   // Access the client
-  const queryClient = useQueryClient();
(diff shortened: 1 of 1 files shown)


akashshyamdev:
@TkDodo @tannerlinsley   Is everything all right... are there any other issues/contributions I can take?

TkDodo:
Thanks for the contribution. I think we have to work on the wording a little bit, but I haven't gotten around to it yet, sorry

akashshyamdev:
@TkDodo cool... could you let me know where it feels off and I'll rewrite that

docs/src/pages/quick-start.md
```diff
@@ -26,14 +23,74 @@ function App() {
   return (
     // Provide the client to your App
     <QueryClientProvider client={queryClient}>
-      <Todos />
```
TkDodo, line 29: i think it's a bit weird that we provide an example that is not really runnable. Rendering a `QueryClientProvider` without `children` isn't even valid jsx ....

docs/src/pages/quick-start.md
```diff
@@ -26,14 +23,74 @@ function App() {
...
     </QueryClientProvider>
   )
 }
+```
+
+Here we are creating a client by creating a new instance of the `QueryClient` class.
```
TkDodo, line 31: `creating a client by creating ...` is a bit repetitive.

docs/src/pages/quick-start.md
```diff
@@ -26,14 +23,74 @@ function App() {
...
+A query client can be used to deal with things like cache.
```
TkDodo, line 32:
```suggestion
A QueryClient can be used to interact directly with the underlying caches.
```

docs/src/pages/quick-start.md
```diff
@@ -26,14 +23,74 @@ function App() {
...
+      </ul>
+    </div>
+  )
+}
+
+render(<App />, document.getElementById('root'))
+```
+
+In the `Todo` component, we start by accessing the queryClient we defined earlier using the `useQueryClient` hook.
+A `query` listens to a source of data, in our case, the `getTodos` method.
```
TkDodo, line 73: this makes it sound like we're listening to the function `getTodos`, which is not true. We're listening to the query key, not the function. It gives the impression that the functional reference must be stable, which it is here, but it doesn't have to be.

docs/src/pages/quick-start.md
```diff
@@ -26,14 +23,74 @@ function App() {
...
+The unique key you provide is used internally for re-fetching, caching, and sharing your queries throughout your application.
+
+Queries are generally use with the `GET` HTTP method, mutations are used for HTTP methods like `POST`, `PUT` etc which modify data on the server.
```
TkDodo, line 76: this is not a requirement and I wouldn't put it on the first page. react-query is promise based, so everything that produces a promise is welcome. It doesn't even have to fetch data over a network at all, and e.g. graphql exclusively uses POST requests for everything ....

docs/src/pages/quick-start.md
```diff
@@ -26,14 +23,74 @@ function App() {
...
+  QueryClientProvider,
+} from 'react-query'
+import { getTodos, postTodo } from '../my-api'
+
+// ...app component
 
 function Todos() {
   // Access the client
-  const queryClient = useQueryClient()
+  const queryClient = useQueryClient();
```
TkDodo, line 93: this is just inconsistent now:

```suggestion
  const queryClient = useQueryClient()
```

TkDodo (changes requested):
I don't really like that the quick start now shows the same example 3 times, always getting bigger, always repeating itself. Without something like line highlighting, it will get difficult to understand what has been added.

How about we just extend the comments that we already have in the code a little bit instead of adding paragraphs underneath it? Most of the descriptions are also a bit self-explanatory, like:

> We have a button, which on being clicked triggers ...

akashshyamdev:
@TkDodo I've made the necessary changes, let me know if it's alright. If everything's good, could you mark this PR as `hacktoberfest-accepted`?

TkDodo:
sorry, I don't think this is making the guide any better than it currently is. I'll try to make the docs code blocks better, maybe with line highlighting and then we can revisit this.

PR #2713, opened 2021-09-23
Return Statement

noor-codes:
## Return Statement

I know everyone loves the one-liner code. It looks awesome!

But in this certain case, I thought newTodo is the second argument to useMutation hook here `useMutation(newTodo => axios.post('/todos', newTodo))` which was very confusing for me when I tried to code and it didn't work. It took me a very long time to notice that small detail.

I think the return statement will drastically remove that confusion.

### DIFF
--- a/docs/src/pages/guides/mutations.md
+++ b/docs/src/pages/guides/mutations.md
@@ -9,7 +9,9 @@ Here's an example of a mutation that adds a new todo to the server:
 
 ```js
 function App() {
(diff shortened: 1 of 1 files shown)


noor-codes:
`Before:` 
Screen Capture_select-area_20210924015932

`After:` 
Screen Capture_select-area_20210924015849

Sugaml:
Yes

PR #2715, opened 2021-09-24
fix types of `UseQueryOptions`

KATT:
In tRPC we extend `UseQueryOptions` like this

```tsx
interface UseTRPCQueryOptions<TInput, TError, TOutput>
  extends UseQueryOptions<TInput, TError, TOutput, QueryKey>,
    TRPCUseQueryBaseOptions {}
```

This leads to `initialData` being incorrectly typed as `TInput` instead of `TOutput` - see https://github.com/trpc/trpc/issues/990

I think this is the culprit.

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -91,7 +91,7 @@ export interface QueryObserverOptions<
   TData = TQueryFnData,
   TQueryData = TQueryFnData,
   TQueryKey extends QueryKey = QueryKey
-> extends QueryOptions<TQueryFnData, TError, TData, TQueryKey> {
+> extends QueryOptions<TQueryFnData, TError, TQueryData, TQueryKey> {
   /**
    * Set this to `false` to disable automatic refetching when the query mounts or changes query keys.
    * To refetch the query, use the `refetch` method returned from the `useQuery` instance.
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -91,7 +91,7 @@ export interface QueryObserverOptions<
   TData = TQueryFnData,
   TQueryData = TQueryFnData,
   TQueryKey extends QueryKey = QueryKey
-> extends QueryOptions<TQueryFnData, TError, TQueryData, TQueryKey> {
+> extends QueryOptions<TQueryFnData, TError, TData, TQueryKey> {
   /**
    * Set this to `false` to disable automatic refetching when the query mounts or changes query keys.
    * To refetch the query, use the `refetch` method returned from the `useQuery` instance.
--- a/src/react/tests/types.test.tsx
+++ b/src/react/tests/types.test.tsx
@@ -1,0 +1,13 @@
+import { InitialDataFunction, UseQueryOptions } from 'react-query'
+import { expectType } from './utils'
+
+test('initialData', () => {
+  type TOutput = 'output'
+  type MyQuery = UseQueryOptions<'input', Error, TOutput>
+
+  type TInitialData = MyQuery['initialData']
+
+  type TExpected = undefined | TOutput | InitialDataFunction<TOutput>
+
+  expectType<TExpected>((null as any) as TInitialData)
+})


KATT:
Added a test now that fails when I revert my fix

KATT:
Maybe @boschni can add an additional test on this branch - maybe my thing breaks his thing? 😅

TkDodo:
I looked at the test, but I don't understand what it is doing. I think there is some confusion about the generics, let my try to clear it up:

- `TQueryFnData` (1st generic): The data type returned from the `queryFn`
- `TData`(3rd generic): The data _returned to_ the `useQuery` observer as `data` property.

If you don't use `select`, those two will be the same, because the observer will get what the queryFn returns:

```
const { data } = useQuery('key', () => Promise.resolve(1))
```

this will have the generics `<number, unknown, number, string>` (unknown for error and string for queryKey, irrelevant here).

so far, so good. If you use `select`, this changes:

```
const { data } = useQuery('key', () => Promise.resolve(1), { select: data => String(data)) }
```

now you'll have: `<number, unknown, string, string>`. `data` will be of type string, but in the query cache, you'll still have a number, because you can have a second observer that could do for example:

```
const { data } = useQuery('key', () => Promise.resolve(1), { select: data => [data, data]) }
```

which would give you: `<number, unknown, Array<number>, string>`.

There would still only be one cache entry (which has a number), and two observers, one getting a string and the other getting the Array. Hope that makes sense.

What does all of this have to do with `initialData` ? Nothing, really. `initialData` in the above example _always_ has to be of type number. Because `initialData` is persisted to the cache, and the result of `initialData` is run _through_ select.

so, if you're declaring `UseQueryOptions<'input', Error, 'output'>`, you're just declaring that this observer will get input from the queryFn and produce output from select. A valid observer for this definition would be:

```
useQuery('key', () => Promise.resolve('input'), { select: () => 'output' })
```

your `initialData` must produce the same type that the queryFn produces. if you want to use initialData in this example, it has to be `'input'` or `() => 'input'`

If you've seen my follow up comment on the original issue (https://github.com/tannerlinsley/react-query/issues/2060#issuecomment-812099685), the confusion comes from `TData` on `QueryOptions`. This actually has to be the same as `TQueryFnData` for "normal" useQuery, but is going to be _something else_ for infinite queries, namely `InfiniteData<TQueryFnData>`, you can see that here:

https://github.com/tannerlinsley/react-query/blob/1896ca27abf46d14df7c6f463d98eb285b8d9492/src/core/types.ts#L209

the confusion is that this is also named `TData`, but it's not the same `TData` used for the `QueryObserverOptions` 🙃 

---

So, I think it would help if you'd come up with a test / an example that actually uses `useQuery` that isn't working for your case please

KATT:
I was under the impression that `TQueryFnData` was the _input_ or _arguments_ to that goes to the query function. 

Given the generic of `UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>`, I guess it must be part of the `TQueryKey` then?

I will explore this further as soon as I got the time -- might be a non-issue, so park this PR for now! 😄

TkDodo:
yes, the input or argument that goes into the queryFn is the `QueryFunctionContext`, which is derived from `TQueryKey`. So, for example:

```
useQuery(['/foo'] as const, (context) => fetch(baseUrl + context.queryKey[0]))
```

the `context` passed to the queryFn contains a property `queryKey`, whose type will be inferred from the actual `queryKey` passed to useQuery, so in the above example: `readonly ["/foo"]`

the `TQueryFnData` on the other hand is the (unwrapped) result type returned from the queryFn. So if you return a `Promise<number>` from your fetcher, `TQueryFnData` will be `number`

KATT:
Yeah, a long time since I did this stuff in trpc so I forgot the nuances. Sorry for wasting your time! 😅

I'm surprised my change above didn't break any other test cases though.

TkDodo:
> I'm surprised my change above didn't break any other test cases though.

yep, me too 😅. I guess many of the tests are focused on correct runtime behaviour. I know there are test for type correctness, but not many / not enough, especially for these combinations like select + initialData. If you want to contribute some more tests in this area - I'd happily merge :)

let me know if we can close this pr

KATT:
I'll use this PR to either prove I was right or prove that I was wrong (more likely) and add some type assertion tests!

Issue #2617, opened 2021-08-30
useInfiniteQuery fetchNextPage function is inconsistent with refetch


lgenzelis:
**Describe the bug**

If you call `fetchNextPage` repeatedly, each call will result in the invocation fo the underlying async function, regardless of whether the first invocation has already resolved. 
This behavior  is inconsistent with that of e.g. `refetch`. If you call `refetch` 100 times in a row, before the first invocation has resolved, the underlying async function doesn't get re-invoked.  

**To Reproduce**
1. Go to this [sandbox](https://codesandbox.io/s/react-query-infinitequery-bug-in3bl?file=/src/index.js)
2. Open the console
3. Click on the `Refetch` button repeatedly. You'll notice the line `Fetching...` printed on the console, just once (unless you wait more than 5 seconds, which is how long the async call takes to resolve)
4. Click on the `Fetch Next Page` button repeatedly. You'll notice the line `Fetching next page...` printed on the console several times (once per button click)

**Expected behavior**
`fetchNextPage`  should behave as `refetch`: you should see the line `Fetching next page...` printed only once. 

**Desktop (please complete the following information):**
 - OS: macos
 - Browser: chrome
 - Version: 92

TkDodo:
it just seems that they have different default values. Both functions call `this.fetch` under the hood, but `fetchNextPage` passes `cancelRefetch: true` as options:

https://github.com/tannerlinsley/react-query/blob/de282541591a7c350035a2fb5c27f5404c5daffa/src/core/infiniteQueryObserver.ts#L93-L103

you can achieve the same behaviour by passing `cancelRefetch` to `refetch`:

```
refetch({ cancelRefetch: true })}
```

it will make sure that currently outgoing requests are cancelled and the new request will be fired immediately.

I can't really tell why `cancelRefetch` can't be passed to `fetchNextPage` (it's hard-coded to `true`) or why the defaults are different...

lgenzelis:
Thanks a lot for your answer @TkDodo ! This is interesting, the fact that it is hardcoded this way gives the impression that this is not a bug at all, and that it behaves this way for some specific reason. But I can't imagine what the reason may be. To me `false` would be a far more sensible default. Moreover,  since `cancelRefetch: true` is enforced (rather than just a default) makes me think that using `cancelRefetch: false` would break something 😬😬 
Let's see if @tannerlinsley remembers the rationale behind this.

TkDodo:
[...]

PR #2731, opened 2021-09-27
feat: respect cancelRefetch flag for infinite queries (fetchNextPage, fetchPreviousPage)

lgenzelis:
See issue #2617

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ stats-hydration.json
 stats-react.json
 stats.html
 .vscode/settings.json
+.idea/
--- a/src/core/infiniteQueryObserver.ts
+++ b/src/core/infiniteQueryObserver.ts
@@ -94,7 +94,8 @@ export class InfiniteQueryObserver<
     options?: FetchNextPageOptions
   ): Promise<InfiniteQueryObserverResult<TData, TError>> {
     return this.fetch({
-      cancelRefetch: true,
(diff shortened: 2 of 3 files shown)


TkDodo (commented):
awesome. could you please
- update the api docs (https://react-query.tanstack.com/reference/useInfiniteQuery), as `options.pageParam` is also mentioned for `fetchNextPage` / `fetchPreviousPage`)
- add a test for passing `cancelRefetch: false`

lgenzelis:
@TkDodo I'm sorry, I feel a bit lost about the testing part. Can you point me to an example where `cancelRefetch` is being tested? (for regular non-infinite queries, I mean)

TkDodo:
I would probably just test if `cancel` on the `query` itself has been called:

https://github.com/tannerlinsley/react-query/blob/e6da990703482cff8ea99f5a254975cb57463ed5/src/core/query.ts#L362-L364

lgenzelis:
@TkDodo I addressed your points 😃 
- updated documentation
- added tests for the cases where `cancelRefetch` is omitted (so, it's `true`) and when `cancelRefetch: false` is passed to `fetchNextPage`

TkDodo:
Would you consider this a feature rather than a bugfix ?

lgenzelis:
> Would you consider this a feature rather than a bugfix ?

@TkDodo I guess so. Mainly because we didn't agree that the previous behavior was a bug on the first place 🤷‍♀️

Issue #2729, opened 2021-09-27
throttleTime cause createWebStoragePersistor not work when open page


disflyer:
**Describe the bug**
use persistQueryClient and createWebStoragePersistor like the image blow.
It not work when access page at  first time. 
Because the default throttleTime will always block the store action at first time, since it init.

![image](https://user-images.githubusercontent.com/22723892/134948845-835d4954-f8df-4472-bb3a-814bb50a4823.png)

**To Reproduce**
https://codesandbox.io/s/autumn-firefly-9s4b6?file=/src/App.js:690-697

**Expected behavior**
use debounce replace throttle, so that persistQueryClient could not lose query state.

**Screenshots**
<img width="1434" alt="desc" src="https://user-images.githubusercontent.com/22723892/134948660-f51cca87-0633-41c4-8c7a-2c62b3beb758.png">

**Desktop (please complete the following information):**
 - Version: 3.23.2

TkDodo:
the issue with throttling seems to be that if data comes in during the `throttleTime`, it is not saved because:
- we start the timer
- data comes in, no new timer is started
- timer finishes, we still closure over the data from the initial cache, which is empty

I'm not sure that debouncing is the better solution though. It means that with a throttleTime of 1 second, if data comes in after 900ms, we only save to the localStorage after 1.9 seconds rather than after 1 second.

Could it be that we just want a different version of `throttling`, like something that lodash throttle offers?

https://lodash.com/docs/4.17.15#throttle

from the docs:

> The func is invoked with the last arguments provided to the throttled function.

Our throttle implementation also seems to only call the function at the trailing end of the time, not at the beginning. So I think it would be equivalent to:

`_.throttle(fn, throttleTime, { leading: false})`

disflyer:
Alright, throttling is better in this case. I will close my PR and reopen a new one. Hope it is helpful.  thank you for your time~

PR #2739, opened 2021-09-29
fix(createAsyncStoragePersistor-experimental) fix throttle func

disflyer:
Based on issue #2729

### DIFF
--- a/src/createWebStoragePersistor-experimental/index.ts
+++ b/src/createWebStoragePersistor-experimental/index.ts
@@ -48,11 +48,13 @@ function throttle<TArgs extends any[]>(
   wait = 100
 ) {
   let timer: number | null = null
-
+  let params: TArgs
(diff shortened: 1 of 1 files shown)


src/createWebStoragePersistor-experimental/index.ts
```diff
@@ -48,11 +48,13 @@ function throttle<TArgs extends any[]>(
   wait = 100
 ) {
   let timer: number | null = null
-
+  let params: TArgs
   return function (...args: TArgs) {
-    if (timer === null) {
+    if (timer !== null) {
```
TkDodo, line 53: I think we can remove the if? If the throttled function is called only once, we will invoke it with undefined after setTimeout

disflyer: yes, it is a mistake.I will fix it.

PR #2760, opened 2021-10-03
fix(style): replace all instances of rem with em.

shih-js:
closes https://github.com/tannerlinsley/react-query/issues/2756

- Changes all instances of `rem` to `em` to prevent styling conflict for developers.

### DIFF
--- a/src/devtools/Explorer.tsx
+++ b/src/devtools/Explorer.tsx
@@ -6,7 +6,7 @@ import { styled } from './utils'
 
 export const Entry = styled('div', {
   fontFamily: 'Menlo, monospace',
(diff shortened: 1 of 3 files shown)


TkDodo (approved):

TkDodo:
prettier check is failing, can you please fix it

jesperp:
Wait, doesn't this fix render very small font sizes on deeply nested objects in the dev tools? E.g the last item here with a font-size of 0.9em has a computed font-size of 6px:

Screenshot 2021-11-05 at 12 12 22

Could probably use CSS custom properties for the font sizes and set them at the top-level of the explorer..

PR #2797, opened 2021-10-14
Add tutorial from Marius Espejo

aaabramov:
That's how I actually learned react-query. Found this video very useful.

### DIFF
--- a/docs/src/pages/videos.mdx
+++ b/docs/src/pages/videos.mdx
@@ -64,3 +64,15 @@ title: Videos & Talks
     width: '100%',
   }}
 ></iframe>
+
+<iframe
+  width="280"
+  height="400"
+  src="https://www.youtube.com/embed/aLQbVd-2tIo"
+  frameborder="0"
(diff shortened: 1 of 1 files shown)


aaabramov:
@TkDodo sorry, what do you mean my legit source?) Just curios :)

TkDodo:
> @TkDodo sorry, what do you mean my legit source?) Just curios :)

well I haven't seen the video, so I can't comment on if it's a good / credible resource to learn react-query. I'd have to take your word for it. There are tons of videos on youtube that are probably not worth watching.

However, the guy who made the video (Marius Espejo) seems to have lots of subscribers (if that metric is worth something) and the video also has a good amount of thumbs-ups (if that tells us anything), so I'm guessing the video is indeed a good / credible source to learn react-query.

Which is why I suggested to move it to community resources, because that's where we want to accumulate external resources like blog posts or videos.

TkDodo:
closed because stale. feel free to reopen with the suggested changes

PR #2798, opened 2021-10-15
feat(queryClient): respect cancelRefetch flag

TkDodo:
for refetchQueries, as well as for invalidateQueries

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -274,10 +274,7 @@ export interface ResetQueryFilters<TPageData = unknown>
     RefetchPageFilters<TPageData> {}
 
 export interface InvalidateOptions extends RefetchOptions {}
-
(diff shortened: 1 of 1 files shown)


dimagimburg:
@TkDodo Hey, until now we used whats written in [this](https://react-query.tanstack.com/guides/query-cancellation) page for cancellation, the requests were cancelled without providing `cancelRefetch` flags.

Will this behavior change in some way?

TkDodo:
Just to be on the same page: This PR didn't change anything regarding the default behaviour. `cancelRefetch` was previously accepted as an `option`, but it was never passed to the actual `fetch` function - so it was always `false` per default.

Now, you have the option of setting it to `true` if you want. If things worked for your use-cases before, they should certainly work now, too :)

PR #2851, opened 2021-10-28
try to clean the old persistented data when storage full

### LINKED ISSUE #1701: persistWithLocalStorage - Failed to execute 'setItem' on 'Storage'  REACT_QUERY_OFFLINE_CACHE exceeded the quota
**Describe the bug**
After loading our app and browsing our a few pages (which make useQuery calls) we get a `localStorage` error.

```
Uncaught DOMException: Failed to execute 'setItem' on 'Storage': Setting the value of 'REACT_QUERY_OFFLINE_CACHE' exceeded the quota.
```

These are the options I am using for the query config, essentially using [defaults options](https://react-query.tanstack.com/plugins/persist-localstorage#options) where throttleTime is `1000` 

```ts
  defaultOptions: {
    queries: {
      useErrorBoundary: true,
      refetchOnWindowFocus: false,
      cacheTime: 1000 * 60 * 60 * 24,
    },
  },
});
persistWithLocalStorage(queryClient); 
```

**To Reproduce**
1. Use the app with [persistWithLocalStorage](https://react-query.tanstack.com/plugins/persist-localstorage#usage) enabled

**Expected behavior**
Should't error, we may need a new default value for the throttle time in options

**Screenshots**
image

NewFuture:
fix #1701
when storage exceeded the quota.

1. try to clean the mutations (from old to new by creating time)
2. try to clean the queries (from old to new by updating time)

### DIFF
--- a/src/createWebStoragePersistor-experimental/index.ts
+++ b/src/createWebStoragePersistor-experimental/index.ts
@@ -30,28 +30,30 @@ export function createWebStoragePersistor({
     return {
       persistClient: throttle(persistedClient => {
         if (trySave(persistedClient) !== true) {
-          const { mutations, queries } = persistedClient.clientState
+          const mutations = [...persistedClient.clientState.mutations]
+          const queries = [...persistedClient.clientState.queries]
+          const client: PersistedClient = {
+            ...persistedClient,
+            clientState: { mutations, queries },
+          }
 
-          // try to remove mutations and save
-          while (mutations.length > 0) {
-            mutations.unshift()
-            persistedClient.clientState.mutations = mutations
-            if (trySave(persistedClient)) {
-              return
+          // clean mutations and try to save
+          while (mutations.shift()) {
+            if (trySave(client)) {
+              return // save success
             }
           }
 
-          // clean old queries and try save
-          const sortedQueries = queries.sort(
+          // sort queries by dataUpdatedAt (oldest first)
+          const sortedQueries = [...queries].sort(
             (a, b) => a.state.dataUpdatedAt - b.state.dataUpdatedAt
           )
+          // clean old queries and try to save
           while (sortedQueries.length > 0) {
             const oldestData = sortedQueries.shift()
-            persistedClient.clientState.queries = queries.filter(
-              q => q !== oldestData
-            )
-            if (trySave(persistedClient)) {
-              return
+            client.clientState.queries = queries.filter(q => q !== oldestData)
+            if (trySave(client)) {
+              return // save success
             }
           }
         }
--- a/src/createWebStoragePersistor-experimental/index.ts
+++ b/src/createWebStoragePersistor-experimental/index.ts
@@ -16,10 +16,45 @@ export function createWebStoragePersistor({
   key = `REACT_QUERY_OFFLINE_CACHE`,
   throttleTime = 1000,
 }: CreateWebStoragePersistorOptions): Persistor {
+  //try to save data to storage
+  function trySave(persistedClient: PersistedClient) {
+    try {
+      storage.setItem(key, JSON.stringify(persistedClient))
+    } catch {
+      return false
+    }
+    return true
+  }
+
   if (typeof storage !== 'undefined') {
     return {
       persistClient: throttle(persistedClient => {
-        storage.setItem(key, JSON.stringify(persistedClient))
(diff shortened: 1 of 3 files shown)


src/createWebStoragePersistor-experimental/index.ts
```diff
@@ -16,10 +16,47 @@ export function createWebStoragePersistor({
...
+            ...persistedClient,
+            clientState: { mutations, queries },
+          }
+
+          // clean mutations and try to save
+          while (mutations.shift()) {
+            if (trySave(client)) {
+              return // save success
+            }
+          }
```
TkDodo, line 45: I think we shouldn't remove `paused` mutations - because those are the ones that are persisted so that they can be resumed later (e.g. when the user is online again).

see: https://github.com/tannerlinsley/react-query/blob/e6da990703482cff8ea99f5a254975cb57463ed5/src/core/mutationCache.ts#L115-L124

src/createWebStoragePersistor-experimental/index.ts
```diff
@@ -16,10 +16,47 @@ export function createWebStoragePersistor({
...
+            (a, b) => a.state.dataUpdatedAt - b.state.dataUpdatedAt
+          )
+          // clean old queries and try to save
+          while (sortedQueries.length > 0) {
+            const oldestData = sortedQueries.shift()
+            client.clientState.queries = queries.filter(q => q !== oldestData)
+            if (trySave(client)) {
+              return // save success
+            }
+          }
```
TkDodo, line 58: that looks good 👍 I can see that we don't have any tests yet, but since this is getting more complex could you maybe add some? I'm also wondering what would happen if the persistFunction now takes longer than the throttleTime 🤔

src/createWebStoragePersistor-experimental/index.ts
NewFuture: All the mutations are `paused` as default.

In fact, I have no idea about this question: which is more important, the `queries` or the `mutations` 😄 .

Do you mean the persisted mutations are more important?
If we think so, we can try to clean the `queries ` at first.

TkDodo: > All the mutations are paused as default.

not sure what you mean with that. A mutation can have one of the following states:

- `success`: it has been executed successfully
- `error`: it has been executed, but is erroneous
- `idle`: it has not been executed yet, or it has been reset
- `loading`: the mutation is currently running

Additionally, a mutation can be in `isPaused: true` "state" when it is running, but has been paused because the mutation cannot be retried because the network is offline.

I think it's very important to _not_ removed paused mutations, because they would likely be resumed when the network goes online again.

So, I think the order of removal should be:

- non-paused mutations
- queries
- paused mutations

src/createWebStoragePersistor-experimental/index.ts
NewFuture: Thank you,

1. I will take a look at the tests in this project, and try to add one for this function.
2. It will be `serial execution`, because, it's a sync function (blocking the timer).

src/createWebStoragePersistor-experimental/index.ts
NewFuture: I found the `defaultShouldDehydrateMutation`
https://github.com/tannerlinsley/react-query/blob/69a5d798fed3beb69f015c131b05074050677223/src/core/hydration.ts#L68-L70

src/createWebStoragePersistor-experimental/index.ts
NewFuture: A test file was added in the 3rd commit

src/createWebStoragePersistor-experimental/index.ts
TkDodo: aaah right, so per default, only paused mutations and successful queries are included in the hydration, but you can overwrite that. I still think the order above would be the most correct one, but it would mean that per default, there are no non-paused mutations, so we could maybe bypass that check.

However, it would still mean that queries should be removed before mutations I think, because mutations are likely _paused_, so removing them before successful queries seems wrong. What do you think?

NewFuture: updated with tests in the 4th commit

NewFuture:
please help review it 😄

TkDodo (approved):

PR #2855, opened 2021-10-28
expose `CancelOptions` & `SetDataOptions` 

KATT:
Background: it would be nice if I didn't have to do [this](https://github.com/trpc/trpc/pull/1186#discussion_r738841851) :)

Technically, this is a breaking change - the other idea I had was to just expose these types in the `types.ts`-file but then the linter complained about circular dep.

It's hard structuring what is "internal" and what is "external" types - the way I go about it in tRPC is that I have folders called `internal/` in various places and I make sure to only cherry-pick from those, and they aren't cherry-picked I do `export *`

### DIFF
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -15,12 +15,13 @@ import type {
   QueryFunctionContext,
   EnsuredQueryKey,
   QueryMeta,
+  CancelOptions,
 } from './types'
 import type { QueryCache } from './queryCache'
 import type { QueryObserver } from './queryObserver'
 import { notifyManager } from './notifyManager'
 import { getLogger } from './logger'
-import { Retryer, CancelOptions, isCancelledError } from './retryer'
+import { Retryer, isCancelledError } from './retryer'
 
 // TYPES
 
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -16,6 +16,7 @@ import type {
   EnsuredQueryKey,
   QueryMeta,
   CancelOptions,
+  SetDataOptions,
 } from './types'
 import type { QueryCache } from './queryCache'
 import type { QueryObserver } from './queryObserver'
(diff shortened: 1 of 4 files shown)


TkDodo:
theoretically, you could infer those types by extracting them from the `Parameters` of `setQueryData`.

> Technically, this is a breaking change

why would this be a breaking change?

KATT:
> theoretically, you could infer those types by extracting them from the `Parameters` of `setQueryData`.

I did that at first but it doesn't work with `invalidateQueries`:

```tsx
invalidateQueries<TPageData = unknown>(filters?: InvalidateQueryFilters<TPageData>, options?: InvalidateOptions): Promise<void>;
invalidateQueries<TPageData = unknown>(queryKey?: QueryKey, filters?: InvalidateQueryFilters<TPageData>, options?: InvalidateOptions): Promise<void>;
```


> > Technically, this is a breaking change
> 
> why would this be a breaking change?

It's changing publicly exposed types. Will break tRPC for instance since I merged that PR, but I'll just fix it and bump the `react-query` peer dependency once this one is merged.

TkDodo:
> It's changing publicly exposed types.

I think they are only considered publicly exposed when being imported directly from `react-query`, like:

```
import { QueryContext } from 'react-query'
```

if you import it `from 'react-query/types/core/query'` I don't think that is considered public :)

also, per [the docs](https://react-query.tanstack.com/typescript):

> Changes to types in this repository are considered non-breaking and are usually released as patch semver changes (otherwise every type enhancement would be a major version!).

So I'm happy to merge this if you are

KATT:
I'm happy!

KATT:
@all-contributors add @KATT for code? :)

TkDodo:
@KATT would you like to fix the `.all-contributorsrc` file, too 😁 ?

KATT:
> @KATT would you like to fix the `.all-contributorsrc` file, too 😁 ?

I can't see what's wrong with it!

TkDodo:
@KATT I think it might be the trailing comma which I added here manually 🤦 

https://github.com/tannerlinsley/react-query/blob/f348673dce003e82a818cc3cc093dea79b0177ea/.all-contributorsrc#L150

TkDodo:
@all-contributors please add @KATT for code

Issue #2680, opened 2021-09-17
useQueries with keepPreviousData:true can get to infinite fetch loop when toggling a query off and on


jvuoti:
When using `useQueries` to issue optional queries, you can seemingly get to an infinite fetch loop if you toggle the same query off and on again. This seems to only happen when having the `keepExistingData: true` set, as well as only when you toggle some other query than the last one. 

We noticed this in one of our views which uses `useQueries` to fetch one or two data series, depending on what the user chooses. If the user toggles one data series off and then on again, the view will start to refetch all the data over and over again.

**To Reproduce** 
1. Go to https://codesandbox.io/s/nice-sid-9tkgi?file=/src/index.js -> the view should load two queries
2. Click on 'Fetch with ID 1' to toggle the query off -> only query with ID 2 is fetched and result is shown
3. Click on 'Fetch with ID 1' again to toggle the query -> both queries are fetched and result shown
4. However, checking the dev tools you can see both the queries stay in "fetching" state. Also the console shows the queries are being issued over and over again:

![image](https://user-images.githubusercontent.com/3702781/133781512-f50bf030-d69e-4707-9d36-6e2f48445a20.png)

**Expected behavior**
If you toggle the first query back on, the data should be fetched again only once.

**Additional info**

react-query version 3.23.2
 
It seems that if you toggle only the _last_ query off and on, this does not happen. e.g. in the codesandbox, toggle "Fetch with ID 2" off and on again, and no additional fetches are issued.

Also, this only seems to happen if you have the `keepExistingData:true` set for the queries. We have now turned that off for our queries, and the view no longer issues extra queries.

**Desktop:**
 - OS: MacOS Big Sur 11.6
 - Browser: Chrome 93.0.4577.63, Firefox 91.0.2

TkDodo:
Thanks for reporting. I contributed that `keepPreviousData` works for `useQueries` not too long ago (it wasn't doing anything at all before that). The PR is here if you're curious: https://github.com/tannerlinsley/react-query/pull/2340

It would be great if you could take a look into this, I think we need to respect the enabled flag somehow - I clearly wasn't anticipating one query getting disabled, my main concern was getting it to work when the length of the query array changes between renders :/

jvuoti:
Thanks for the quick response! Looking at that PR, using the index to get the observer there might have an effect why this happens only for the first query and not the last one... 

[...]

PR #2866, opened 2021-11-01
fix(useQueries): search for unused observers if keepPreviousData (#2680)

jvuoti:
#closes 2680

When `keepPreviousData` is on, search for a query observer no longer used instead of relying on observer order and indexing. Trying to match observers by index could cause infinite request loops when switching queries. See #2680 for more 

This does complicate the `getOptimisticResult` and `updateObservers` calls a bit. Now they both essentially do the same search for unused observers. The observers lists for the `queriesObserver` are only updated in the `updateObservers`, though. 

This does cause some extra updates, as can be seen in the tests. If I understand correctly, they probably are due to the `setOptions` calls being called both on `getOptimisticResult` and `updateObservers` calls. I tried for quite a while to get rid of the extra updates, but couldn't find a way that would still always return previous data, so eventually just left it like this and edited the tests. Any ideas to get rid of the extra updates are very welcome 😊

### DIFF
--- a/src/core/queriesObserver.ts
+++ b/src/core/queriesObserver.ts
@@ -77,17 +77,17 @@ export class QueriesObserver extends Subscribable<QueriesObserverListener> {
     const newObservers: QueryObserver[] = []
     const newObserversMap: Record<string, QueryObserver> = {}
 
-    const defaultedQueryOptions = queries.map(o =>
-      this.client.defaultQueryObserverOptions(o)
+    const defaultedQueryOptions = queries.map(options =>
+      this.client.defaultQueryObserverOptions(options)
     )
     const matchingObservers = defaultedQueryOptions
-      .map(o => {
-        if (o.queryHash == null) {
+      .map(options => {
+        if (options.queryHash == null) {
           return null
         }
-        const match = prevObserversMap[o.queryHash]
+        const match = prevObserversMap[options.queryHash]
         if (match != null) {
-          match.setOptions(o, notifyOptions)
+          match.setOptions(options, notifyOptions)
           return match
         }
         return null
@@ -95,27 +95,28 @@ export class QueriesObserver extends Subscribable<QueriesObserverListener> {
       .filter(notNullOrUndefined)
 
     const matchedQueryHashes = matchingObservers
-      .map(o => o?.options.queryHash)
+      .map(observer => observer?.options.queryHash)
       .filter(notNullOrUndefined)
     const unmatchedQueries = defaultedQueryOptions.filter(
-      o =>
-        o.queryHash !== undefined && !matchedQueryHashes.includes(o.queryHash)
+      options =>
+        options.queryHash !== undefined &&
+        !matchedQueryHashes.includes(options.queryHash)
     )
 
     const unmatchedObservers = prevObservers.filter(
-      p => !matchingObservers.includes(p)
+      prevObserver => !matchingObservers.includes(prevObserver)
     )
 
-    const newlyMatchedOrCreatedObservers = unmatchedQueries.map(q => {
-      if (q.keepPreviousData && unmatchedObservers.length > 0) {
-        // use the first existing but no longer matched query to keep query data for any new queries
+    const newlyMatchedOrCreatedObservers = unmatchedQueries.map(options => {
+      if (options.keepPreviousData && unmatchedObservers.length > 0) {
+        // use the first observer but no longer matched query to keep query data for any new queries
         const firstObserver = unmatchedObservers.splice(0, 1)[0]
         if (firstObserver !== undefined) {
-          firstObserver.setOptions(q, notifyOptions)
+          firstObserver.setOptions(options, notifyOptions)
           return firstObserver
         }
       }
-      return this.getObserver(q)
+      return this.getObserver(options)
     })
 
     matchingObservers
@@ -148,7 +149,7 @@ export class QueriesObserver extends Subscribable<QueriesObserverListener> {
       )
 
       const hasIndexChange = updatedObservers.newObservers.some(
-        (o, i) => o !== prevObservers[i]
+        (observer, index) => observer !== prevObservers[index]
       )
       if (
         prevObservers.length === updatedObservers.newObservers.length &&
--- a/src/core/queriesObserver.ts
+++ b/src/core/queriesObserver.ts
@@ -1,4 +1,4 @@
-import { difference, notNullOrUndefined, replaceAt } from './utils'
+import { difference, replaceAt } from './utils'
 import { notifyManager } from './notifyManager'
 import type { QueryObserverOptions, QueryObserverResult } from './types'
 import type { QueryClient } from './queryClient'
@@ -77,17 +77,17 @@ export class QueriesObserver extends Subscribable<QueriesObserverListener> {
       this.client.defaultQueryObserverOptions(options)
     )
 
-    const matchingObservers: QueryObserverMatch[] = defaultedQueryOptions
-      .map(defaultedOptions => {
+    const matchingObservers: QueryObserverMatch[] = defaultedQueryOptions.flatMap(
+      defaultedOptions => {
         const match = prevObservers.find(
-          observer => observer.options.queryHash === defaultedOptions.queryHash!
+          observer => observer.options.queryHash === defaultedOptions.queryHash
         )
         if (match != null) {
-          return { defaultedQueryOptions: defaultedOptions, observer: match }
+          return [{ defaultedQueryOptions: defaultedOptions, observer: match }]
         }
-        return null
-      })
-      .filter(notNullOrUndefined)
+        return []
+      }
+    )
 
     const matchedQueryHashes = matchingObservers.map(
       match => match.defaultedQueryOptions.queryHash
@@ -104,7 +104,7 @@ export class QueriesObserver extends Subscribable<QueriesObserverListener> {
 
     const newOrReusedObservers: QueryObserverMatch[] = unmatchedQueries.map(
       (options, index) => {
-        if (options.keepPreviousData && unmatchedObservers.length > 0) {
+        if (options.keepPreviousData) {
           // return previous data from one of the observers that no longer match
           const previouslyUsedObserver = unmatchedObservers[index]
           if (previouslyUsedObserver !== undefined) {
--- a/src/core/queriesObserver.ts
+++ b/src/core/queriesObserver.ts
@@ -64,69 +64,64 @@ export class QueriesObserver extends Subscribable<QueriesObserverListener> {
   }
 
   getOptimisticResult(queries: QueryObserverOptions[]): QueryObserverResult[] {
-    return this.getUpdatedObservers(queries).newResult
+    return this.findMatchingObservers(queries).map(match =>
+      match.observer.getCurrentResult()
+    )
   }
 
-  private getUpdatedObservers(
-    queries: QueryObserverOptions[],
-    notifyOptions?: NotifyOptions
-  ): QueryObserverUpdate {
+  private findMatchingObservers(
+    queries: QueryObserverOptions[]
+  ): QueryObserverMatch[] {
     const prevObservers = this.observers
-    const prevObserversMap = this.observersMap
-    const newResult: QueryObserverResult[] = []
-    const newObservers: QueryObserver[] = []
-    const newObserversMap: Record<string, QueryObserver> = {}
-
     const defaultedQueryOptions = queries.map(options =>
       this.client.defaultQueryObserverOptions(options)
     )
-    const matchingObservers = defaultedQueryOptions
-      .map(options => {
-        const match = prevObserversMap[options.queryHash!]
+
+    const matchingObservers: QueryObserverMatch[] = defaultedQueryOptions
+      .map(defaultedOptions => {
+        const match = prevObservers.find(
+          observer => observer.options.queryHash === defaultedOptions.queryHash!
+        )
         if (match != null) {
-          match.setOptions(options, notifyOptions)
-          return match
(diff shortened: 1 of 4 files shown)


src/core/utils.ts
```diff
@@ -453,3 +453,10 @@ export function getAbortController(): AbortController | undefined {
     return new AbortController()
   }
 }
+
+/**
+ * Type predicate to filter an array of null or undefined elements
+ */
+export function notNullOrUndefined<T>(value: T | null | undefined): value is T {
```
TkDodo, line 460: we can simplify the types here:

```suggestion
export function notNullOrUndefined<T>(value: T): value is NonNullable<T> {
```

src/core/queriesObserver.ts
```diff
@@ -64,78 +64,120 @@ export class QueriesObserver extends Subscribable<QueriesObserverListener> {
...
+    const newResult: QueryObserverResult[] = []
+    const newObservers: QueryObserver[] = []
+    const newObserversMap: Record<string, QueryObserver> = {}
+
+    const defaultedQueryOptions = queries.map(o =>
+      this.client.defaultQueryObserverOptions(o)
+    )
+    const matchingObservers = defaultedQueryOptions
+      .map(o => {
+        if (o.queryHash == null) {
```
TkDodo, line 85: I'd appreciate it if we wouldn't use one-letter abbreviations. Using `option` here instead of `o` would't hurt I believe, especially because further down, `o` is also used as abbreviation for `observer`, which makes this quite confusing ....

TkDodo (commented):
thank you. Regarding the extra re-renders: Are these _only_ happening if we also have `keepPreviousData: true` on the queries, or are all queries affected? Judging from the unchanged tests, things should stay the same for `keepPreviousData: false`, right?

TkDodo:
> When keepPreviousData is on, search for a query observer no longer used instead of relying on observer order and indexing. Trying to match observers by index could cause infinite request loops when switching queries. See #2680 for more

I'm not yet fully understanding _how_ you're actually matching the observers together if not by index 🤔 

also, prettier formatting is failing, please have a look 👀

jvuoti:
Thank you for the comments, good points, I now fixed the namings and formatting.

And oh yeah, you are right that this still fulfills the observer for previous data by index 😅 Just now the indexing is done from the list of observers that are no longer being used.

Previously the code fulfilled the observer by index from the main observer list, assuming the order of the queries would have stayed the same. However, that was not always the case when e.g. toggling queries off and on. This could then cause the same observer being returned first to an existing query (by hash) and then to a new query as previous data (by index), causing the infinite request loop. The new test should test for this issue.

The matching could be more advanced, to e.g. try to find the nearest match of the query key parts. However, this now at least should fix the biggest issue of infinite request loops. It also fulfills our use case of toggling between similar queries, with only an identifier being changed.

The extra re-renders do happen only when keepPreviousData is turned on. If I have understood correctly, this happens because now the `getOptimisticResults` also needs to call the same search and update code as `updateObservers`. Because `getOptimisticResults` does not update the internal observer map caches, [firstObserver.setOptions](https://github.com/tannerlinsley/react-query/pull/2866/files#diff-ba20afedab2419b82900a184fbde1adb5d4408a0b7eaead6896c7482ae00bf7bR115) gets called twice. This then seems to trigger the listeners also twice, even if the `getOptimisticResults` has left the `notifyOptions` undefined - or even when explicitly setting all the options false.

src/core/queriesObserver.ts
```diff
@@ -64,78 +64,121 @@ export class QueriesObserver extends Subscribable<QueriesObserverListener> {
...
+    const newObservers: QueryObserver[] = []
+    const newObserversMap: Record<string, QueryObserver> = {}
+
+    const defaultedQueryOptions = queries.map(options =>
+      this.client.defaultQueryObserverOptions(options)
+    )
+    const matchingObservers = defaultedQueryOptions
+      .map(options => {
+        if (options.queryHash == null) {
+          return null
```
TkDodo, line 86: codecov is reporting here that his line isn't covered, and I think maybe we can even remove the check and use `options.queryHash!` like before?

judging from the code, we should _always_ have a query hash when working with defaulted options, seems just quite hard to convince the compiler about it.

the hash is added to the options here if it does not yet exist: https://github.com/tannerlinsley/react-query/blob/f21d932142e8618995936c12c5f5696aa8993473/src/core/queryClient.ts#L619-L624

jvuoti: Thanks for the comment, I hadn't thought about those filters too much, just assumed empty hashes could be filtered out. 

However, when I tested this out a bit more, I noticed that you actually _can_ pass an undefined `queryKey` via options. This will make the `queryHash` undefined as well. And it does seem to work - somewhat. All the falsy `queryKeys` will just have an undefined `queryHash`, but because Map does allow undefined keys, the same single observer will be returned for all falsy `queryKey` values. 

So assuming someone might be using this, I changed the code to not filter out undefined `queryHash`values, but just use them as keys. I added a couple of tests for this case, so the coverage should be up to 100% again.

I think I'd prefer to use the nullish coalescing operator (`queryHash ?? ''`) to make it more explicit we are allowing for null or undefined hashes here. However, the previous non-null assertion operator (`queryHash!`) would probably also work in this case, allowing undefined to be used in the Map keys. So I can change to that if that is more commonly used here?

TkDodo: > I noticed that you actually can pass an undefined queryKey via options

TypeScript types don't allow for it. Sure, if you use JavaScript, you can always pass in anything :)

> However, the previous non-null assertion operator (queryHash!) would probably also work in this case, allowing undefined to be used in the Map keys. So I can change to that if that is more commonly used here?

yes please. I'm usually not a fan of the bang operator, but I also don't like to add code in a library for cases that will not happen if it increases the codebase

jvuoti: :+1: Changed the Map indexing to use the bang operator. 

> TypeScript types don't allow for it. Sure, if you use JavaScript, you can always pass in anything :)

I think you can pass an undefined `queryKey` also in TypeScript, if you only use the QueryOptions object for the parameters, apparently because there [queryKey is marked as optional](https://github.com/tannerlinsley/react-query/blob/master/src/core/types.ts#L69). E.g. `useQueries([{ queryKey: undefined, queryFn: () => 3 }])` seems to compile and run.

But true, also JS allows you to pass in anything, so I guess we need to support it anyways :)

TkDodo: It will likely become required in v4 :)

TkDodo:
can you please update with the latest `master`, as we've added some more tests. And please make sure that the test coverage does not drop (it currently does, see the codecov report)

src/core/queriesObserver.ts
```diff
@@ -64,78 +64,116 @@ export class QueriesObserver extends Subscribable<QueriesObserverListener> {
...
+    )
+
+    const unmatchedObservers = prevObservers.filter(
+      prevObserver => !matchingObservers.includes(prevObserver)
+    )
+
+    const newlyMatchedOrCreatedObservers = unmatchedQueries.map(options => {
+      if (options.keepPreviousData && unmatchedObservers.length > 0) {
+        // use the first observer but no longer matched query to keep query data for any new queries
+        const firstObserver = unmatchedObservers.splice(0, 1)[0]
```
TkDodo, line 108: splice mutates `unmatchedObservers`. It is a newly created array due to `prevObservers.filter`, but it would still be appreciated if we keep this immutable for readability. Or is this intended (I don't think it is since we abort the loop if we find the `firstObserver`..)

src/core/queriesObserver.ts
```diff
@@ -64,78 +64,116 @@ export class QueriesObserver extends Subscribable<QueriesObserverListener> {
...
+        if (firstObserver !== undefined) {
+          firstObserver.setOptions(options, notifyOptions)
```
TkDodo, line 110: you mentioned in the PR description:

> This does cause some extra updates, as can be seen in the tests. If I understand correctly, they probably are due to the setOptions calls being called both on getOptimisticResult and updateObservers calls.

I don't think that we _can_ call `setOptions` in `getOptimisticResult`, because `getOptimisticResult` is called _during_ render. It doesn't have any side-effects in the current version, where `updateObserver` is only called:
- in the constructor
- in an effect in `useQueries`
which are both legal by the rules of react.

`setOptions` potentially triggers a fetch, so I'm afraid we really can't do that 🚨

TkDodo (changes requested):

src/core/queriesObserver.ts
jvuoti: Ah, you are right! Thank you, I must have been too busy trying to see if I could call `setOptions` without triggering a re-render to stop to think if I should 😅 

I now refactored the code so the search part is all encapsulated to `findMatchingObservers`. Then `updateObservers` will separately call `setOptions` on the returned list. 

This gets rid of the extra renders, but there still seems to be some small change in the order of the updates compared to master. This means I couldn't completely revert the tests in `useQueries.test.tsx` back. However, the difference seems to rather small: two renders back to back, which previously both showed the query with keepPreviousData as fetching and `isPreviousData:true`, whereas now the first round still shows it not fetching yet (`isFetching: false`). Couldn't quite track why this happens, but it doesn't seem to have any visible effect, so I just did the small change to the tests.

src/core/queriesObserver.ts
jvuoti: You are right, it's better to keep this immutable. Simple indexing works here nicely as well, so changed the code to get rid of splice 👍

src/core/queriesObserver.ts
```diff
@@ -64,57 +64,94 @@ export class QueriesObserver extends Subscribable<QueriesObserverListener> {
...
+      .map(defaultedOptions => {
+        const match = prevObservers.find(
+          observer => observer.options.queryHash === defaultedOptions.queryHash!
+        )
+        if (match != null) {
+          return { defaultedQueryOptions: defaultedOptions, observer: match }
+        }
+        return null
+      })
+      .filter(notNullOrUndefined)
```
TkDodo, line 90: I think we wouldn't need the `notNullOrUndefined` util if we'd just `flatMap` over the array and return `[]` for cases that we want to exclude and an array with a single entry for those that match. What do you think?

TkDodo (changes requested):
can you please add a `useQueries` test for the initial issue, the infinite render loop? Thanks.

src/core/queriesObserver.ts
TkDodo: that's fine I think 👍

src/core/queriesObserver.ts
jvuoti: Good point, `flatMap` would work here as well, and then we would not need the new util for the single use.

Replaced now with `flatMap` 👍

jvuoti:
> can you please add a `useQueries` test for the initial issue, the infinite render loop? Thanks.

The ["should keep previous data when switching between queries" test](https://github.com/tannerlinsley/react-query/pull/2866/files#diff-27ab6cf11862f134526488fd6c22381bf484a5ad9a6d7dc36535e1d3e2d5aa42R255) added to `useQueries.test` does test the infinite render loop issue as a side effect: if you run that test against master, the [expect on line 296](https://github.com/tannerlinsley/react-query/pull/2866/files#diff-27ab6cf11862f134526488fd6c22381bf484a5ad9a6d7dc36535e1d3e2d5aa42R296) will fail, because there are going to be over 250 result states in the list 😬 

But good point, it would be nice to have a simpler test explicitly for that issue. If needed, I can probably look at that over the weekend.

TkDodo:
> But good point, it would be nice to have a simpler test explicitly for that issue. If needed, I can probably look at that over the weekend.

that would be great, thanks

jvuoti:
Now added a [test explicitly for the infinite render loop](https://github.com/tannerlinsley/react-query/pull/2866/files#diff-27ab6cf11862f134526488fd6c22381bf484a5ad9a6d7dc36535e1d3e2d5aa42R355) when toggling queries on and off, i.e. the initial issue from the bug report. It is still a bit cumbersome to set up and validate the issue, unfortunately couldn't find a way to get it much simpler than that.

TkDodo (approved):
Thank you, I'm going to trust the test and merge this. Also went a bit through the logic and I think it's pretty readable now 🚀

src/core/queriesObserver.ts
```diff
@@ -64,57 +64,94 @@ export class QueriesObserver extends Subscribable<QueriesObserverListener> {
...
   getOptimisticResult(queries: QueryObserverOptions[]): QueryObserverResult[] {
-    return queries.map((options, index) => {
-      const defaultedOptions = this.client.defaultQueryObserverOptions(options)
-      return this.getObserver(defaultedOptions, index).getOptimisticResult(
-        defaultedOptions
-      )
-    })
+    return this.findMatchingObservers(queries).map(match =>
+      match.observer.getCurrentResult()
+    )
```
TkDodo, line 69: quick follow-up question here @jvuoti : previously, we called `getOptimisticResult` of each observer, and now we just call `getCurrentResult()`. I'm surprised that this doesn't seem to make a difference 🤔 ...

jvuoti: Ah, you are right, it should call `getOptimisticResult` there, so that is a bug. Sorry, must have been some lazy copy-pasting on my part when refactoring the code around. 🤦 

Testing this out quickly, the change does not seem to cause any visible effects - but it _does_ now fix the issue I had with some of the existing tests still returning `isFetching: false`. So yeah, I probably should have noticed the issue from those tests. But at least now I can revert the tests back to the original, correct form 😅 

So good catch, thank you so much! I'll prep a PR to fix the call and tests shortly.

TkDodo: Thank you 🙏

PR #2876, opened 2021-11-04
test(core): improve code coverage

arnaudbzn:
Add tests to improve code coverage for `./src/core`

Before:
Screenshot 2021-11-04 at 10 30 07

After:
Screenshot 2021-11-04 at 10 30 15

For some tests I had to "force" some use cases  by accessing private functions and props (I use `// Force....` in the test code comments. It means that we can maybe remove some pieces of existing code that are not necessary.

## Not covered
No solution found  to tests the following cases to reach 100% :pensive:
- [ ] `utils.scheduleMicrotask`: Mock `process.on` to test uncaught exception but issue with jest version support.
- [ ]  `utils.isPlainObject`: Find the corresponding object case!
- [ ]  `logger`:  Mock console to undefined or assume that the console is always defined and change the existing code `logger`definition.

## Impacts on existing code
Code changes to improve code coverage:
- [x]  `notifyManager.ts:` added a `notifyManagerTest` singleton for testing only
- [x] `queryCache.ts`: fixed comparison in`find()` filters object is always defined but could be empty.
- [x]  `queryObserver.ts`: removed useless condition  `shouldNotifyListeners`

### DIFF
--- a/src/core/notifyManager.ts
+++ b/src/core/notifyManager.ts
@@ -10,7 +10,7 @@ type BatchNotifyFunction = (callback: () => void) => void
 
 // CLASS
 
-class NotifyManager {
+export class NotifyManager {
   private queue: NotifyCallback[]
   private transactions: number
   private notifyFn: NotifyFunction
@@ -94,6 +94,3 @@ class NotifyManager {
 // SINGLETON
 
 export const notifyManager = new NotifyManager()
-
-// For testing purposes (not change made by jest.setup.js)
-export const notifyManagerTest = new NotifyManager()
--- a/src/core/notifyManager.ts
+++ b/src/core/notifyManager.ts
@@ -94,3 +94,6 @@ class NotifyManager {
 // SINGLETON
 
 export const notifyManager = new NotifyManager()
+
+// For testing purposes (not change made by jest.setup.js)
+export const notifyManagerTest = new NotifyManager()
--- a/src/core/queryObserver.ts
+++ b/src/core/queryObserver.ts
@@ -590,10 +590,6 @@ export class QueryObserver<
       return true
     }
 
-    if (result === prevResult) {
-      return false
-    }
-
     const { notifyOnChangeProps, notifyOnChangePropsExclusions } = this.options
 
     if (!notifyOnChangeProps && !notifyOnChangePropsExclusions) {
--- a/src/core/tests/focusManager.test.tsx
+++ b/src/core/tests/focusManager.test.tsx
@@ -1,0 +1,106 @@
+import { sleep } from '../utils'
+import { focusManager } from '../focusManager'
+
+describe('focusManager', () => {
+  afterEach(() => {
+    // Reset removeEventListener private property to avoid side effects between tests
+    focusManager['removeEventListener'] = undefined
+  })
+
+  it('should call previous remove handler when replacing an event listener', () => {
+    const remove1Spy = jest.fn()
+    const remove2Spy = jest.fn()
+
+    focusManager.setEventListener(() => remove1Spy)
+    focusManager.setEventListener(() => remove2Spy)
+
+    expect(remove1Spy).toHaveBeenCalledTimes(1)
+    expect(remove2Spy).not.toHaveBeenCalled()
+  })
+
+  it('should use focused boolean arg', async () => {
+    let count = 0
+
+    const setup = (setFocused: (focused?: boolean) => void) => {
+      setTimeout(() => {
+        count++
+        setFocused(true)
+      }, 20)
+      return () => void 0
+    }
+
+    focusManager.setEventListener(setup)
+
+    await sleep(30)
+    expect(count).toEqual(1)
+    expect(focusManager.isFocused()).toBeTruthy()
+  })
+
+  it('should not notify listeners on focus if already focused', async () => {
+    const subscriptionSpy = jest.fn()
+    const unsubscribe = focusManager.subscribe(subscriptionSpy)
+
+    focusManager.setFocused(true)
+    expect(subscriptionSpy).toHaveBeenCalledTimes(1)
+    subscriptionSpy.mockReset()
+
+    focusManager.setFocused(false)
+    expect(subscriptionSpy).toHaveBeenCalledTimes(0)
+
+    unsubscribe()
+  })
+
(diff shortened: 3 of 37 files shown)


src/core/notifyManager.ts
```diff
@@ -94,3 +94,6 @@ class NotifyManager {
 // SINGLETON
 
 export const notifyManager = new NotifyManager()
+
+// For testing purposes (not change made by jest.setup.js)
+export const notifyManagerTest = new NotifyManager()
```
TkDodo, line 99: can we tie this to `process.env.NODE_ENV !== 'production'`, or maybe even better, create the `new NotifyManager()` inside the test?

src/core/queryObserver.ts
```diff
@@ -590,10 +590,6 @@ export class QueryObserver<
       return true
     }
 
-    if (result === prevResult) {
-      return false
-    }
```
TkDodo, line 595: why was this removed please?

src/core/tests/focusManager.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+    focusManager.setFocused(false)
+    expect(subscriptionSpy).toHaveBeenCalledTimes(0)
+
+    unsubscribe()
+  })
+
+  it('should return true for isFocused if document is undefined', async () => {
+    const { document } = globalThis
+
+    // @ts-ignore
```
TkDodo, line 56: let's please use `ts-expect-error` instead of `ts-ignore` wherever possible: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments

src/core/tests/queryObserver.test.tsx
```diff
@@ -490,4 +495,275 @@ describe('queryObserver', () => {
...
+
+    const spy = jest.fn()
+    const unsubscribe = observer.subscribe(spy)
+    await queryClient.refetchQueries(key)
+    await sleep(10)
+
+    // Force isStale to false
+    // because no use case has been found to reproduce this condition
+    // @ts-ignore
+    observer['currentResult'].isStale = true
```
TkDodo, line 762: hmm, the `staleTime` is `20`, but we only wait for `10`. I think if we reverse this (set staleTime to 10 and wait for 20), `isStale` should automatically go to `true`.

TkDodo (changes requested):
thank you for all the effort that goes into this 🚀

TkDodo:
also, we just added codecov reporting. We should see a nice increase if you update your PR, and maybe we could even increase the expected coverage in the settings?

src/core/queryObserver.ts
arnaudbzn: `shouldNotifyListeners` is only used by `updateResult` that already does: 
```ts
     // Only notify if something has changed
    if (shallowEqualObjects(this.currentResult, prevResult)) {
      return
    }
```
before calling `shouldNotifyListeners`.
That why this condition was not in the code coverage so we can safely remove it or we can keep it and add a test  to directly  call `shouldNotifyListeners` with the appropriate params to cover this condition.

TkDodo: right, I think `shouldNotifyListeners` should not make the extra equals check on result and previousResult, so this is fine 👍

src/core/tests/focusManager.test.tsx
arnaudbzn: Done when possible in https://github.com/tannerlinsley/react-query/pull/2876/commits/ca22c597d27e0e4138f480f642ecbea374368b7c
Still have to use ` @ts-ignore ` to avoid `Unused '@ts-expect-error' directive.` in some cases.

src/core/notifyManager.ts
arnaudbzn: Done in https://github.com/tannerlinsley/react-query/pull/2876/commits/d3e40aff30fdb8d967c5e5c793e5b4b7757c6630
I had to export `NotifyManager` class. 
I think it's the best solution for testing.

src/core/tests/queryObserver.test.tsx
arnaudbzn: If I do this change the else in the block of code below is not covered anymore:


As mentioned in the test comment I have not found a use case for this condtion in the `setTimeout`:
https://github.com/tannerlinsley/react-query/blob/b52b2779b32b5cf0c21733e7d86b5882b8a611a6/src/core/queryObserver.ts#L364-L368

TkDodo: I think what we'd need to do is manually set the query to stale by calling `queryClient.invalidateQueries(queryKey, { refetchActive: false })`. This would mark the query as `isInvalidated`, so `isStale` will already be true on the currentResult when the timer fires:

https://github.com/tannerlinsley/react-query/blob/e7a3207f74b27bc112d6ab583831cceac60f94c5/src/core/query.ts#L288-L294

arnaudbzn: Tried but the `else` is still not covered and even the timeout handler is not called at all in this case.
It's because:  
`query.invalidate()`  => `dispatch()` => `onQueryUpdate()` => `updateTimers()` => `updateStaleTimeout()` => `clearStaleTimeout()`
That's why I don't see a use case for this code coverage.
It looks like a useless condition in the `setTimeout()` handler (or a bug smell?).

TkDodo: True, it could be that the timer is reset once the query is marked as stale by invalidation. Maybe the condition is unnecessary but I’d rather not remove it now. Do you think it makes sense to keep the test?

arnaudbzn: I was wondering the same for this kind of tests that test something that should never happen..
We can keep this test to improve the code coverage and I use the comment prefix `// Force ... ` to identify this kind of situation. 
So it makes sense to keep this test, it's a kind a code documentation about the tested condition.

arnaudbzn:
@TkDodo It will be great if we can achieve 100% code coverage for the `./core` code base.
We can try do that in this PR or do you prefer to do that in another PR for the 3 remaining tests?

TkDodo:
Let’s merge this one and do a follow up 👍

TkDodo (approved):

PR #2884, opened 2021-11-05
fix: rename react directory to reactjs

EddyVinck:
Fixes https://github.com/tannerlinsley/react-query/issues/2339

The directory being named "react" causes an error with the moduleDirectories option from jest

Reproduction (using Codesandbox package version of this PR): https://github.com/EddyVinck/react-query-2339


TkDodo (commented):
Thanks, this looks good. One question:

We have this top level `react` directory with a `package.json`:

https://github.com/tannerlinsley/react-query/blob/f348673dce003e82a818cc3cc093dea79b0177ea/react/package.json#L1-L6

that points to `lib/react`, `es/react` and `types/react`. I haven't seen any changes in the build process (`rollup.config.js`). could you check if these directories still exist when making a local build on your machine, or if we need to adapt this package.json, too ? Thanks!

TkDodo (changes requested):
ah, I just downloaded the codesandbox build from https://pkg.csb.dev/tannerlinsley/react-query/commit/3a734e5f/react-query, and it indeed looks like the folder is now `reactjs` in `/lib` etc as well. So I think we need to change that package-json please, too.

TkDodo:
I  think the whole file would need to be renamed please:

```diff
- react-query/react/package.json
+ react-query/reactjs/package.json 
```

TkDodo:
to be on the safe side, I've added this to `v4` because it might be breaking if someone was importing from `react-query/react`. Please make the PR go to the `v4` branch.

TkDodo (approved):

TkDodo (approved):

TkDodo:
@all-contributors please add @EddyVinck for code

PR #2886, opened 2021-11-05
Fix progressively smaller font size for Explorer

emzoumpo:
Resolved bug mentioned in this [comment](https://github.com/tannerlinsley/react-query/pull/2760#issuecomment-961812722)
Fix progressively smaller font size for Explorer entries in devtools
Set Entry font-size to 1em, because it is indented multiple time within each other.
This causes the font to get smaller and smaller as `em` is relative to the parent font size.
font-size-progressively-smaller


jesperp:
Thanks for this!

PR #2892, opened 2021-11-06
fix(Mutation): Export the `Mutation` type and change default `TVariables` from `void` to `unknown`

### LINKED ISSUE #2838: No exported member "Mutation"
**Describe the bug**
It looks like exports from `mutation.d.ts` are not picked up correctly by TS. Instead a 'no exported member' error is thrown, making explicit typing with `Mutation` and other members impossible. Contrast that with importing `Query` which works fine.

**To Reproduce**
[Repro sandbox here](https://codesandbox.io/s/fervent-leakey-9dt6f?file=/src/index.tsx)

**Expected behavior**
Type for `Mutation` is imported

Brodzko:
closed #2838 

Exported the `Mutation` type as per #2838.

I also changed the `TVariables` to default to `unknown` - but since I'm unfamiliar with why it originally defaulted to `void` I can revert this if it's an undesired or breaking change.

The reason for that was that in `MutationCache` error handlers, `variables` naturally default to `unknown` which led to type errors in error handlers unless `Mutation` wasn't explicitly retyped to `Mutation<unknown, unknown, unknown, unknown>`.

Let me know if you need this reverted 🙂

### DIFF
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -17,6 +17,7 @@ export { dehydrate, hydrate } from './hydration'
 // Types
 export * from './types'
 export type { Query } from './query'
+export type { Mutation } from './mutation'
 export type { Logger } from './logger'
 export type {
   DehydrateOptions,
--- a/src/core/mutation.ts
+++ b/src/core/mutation.ts
@@ -19,7 +19,7 @@ interface MutationConfig<TData, TError, TVariables, TContext> {
 export interface MutationState<
   TData = unknown,
   TError = unknown,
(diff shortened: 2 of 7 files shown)


TkDodo:
I think the reason shows in the test failures: you can't call `mutate()` _without_ parameters if it defaults to `unknown`. Additionally, in the global callback, we don't know what the variables will be, so we can't use `void` there either.

I think the best thing would be to revert the additional changes please.

TkDodo (approved):

PR #2905, opened 2021-11-08
Bail out of query update if data is undefined or same as previous data

### LINKED ISSUE #2925: Make undefined an illegal cache value (for successful queries)
Please see [this discussion](https://github.com/tannerlinsley/react-query/discussions/2823)

This is an ever-recurring issue that you can't bail out of updates with `setQueryData`. But because the updater function gives you a potentially undefined value (if there is no cache entry for the key yet), TypeScript is very unhappy because you're supposed to create a new cache entry, and there is no "bailing out of it" because react-query actually creates the cache entry _before_ the functional updater of setQueryData is called.

### Proposed solution

As stated [here](https://github.com/tannerlinsley/react-query/discussions/2823#discussioncomment-1562184):

- let's allow the setQueryData functional updater to return undefined
  - signature: T | ((prev: T | undefined) => T | undefined)
  - if `undefined` is returned, we do not create or update the cache entry
  - that means if we have a previous value and `undefined` is returned - `undefined` is _not_ written to the cache, and observers are not informed about that.
  - this is technically breaking if `undefined` is a legit cache entry value for someone

This would be in-ine with what the `initialData` function already does: if `undefined` is returned from that, it is _not_ persisted to the cache.

Also, if a fetch function returns `undefined`, we should:
- transform it to a failed promise (error)
- warn during development mode about it

phatmann:
See https://github.com/tannerlinsley/react-query/discussions/2823

closes #2925


src/core/query.ts
```diff
@@ -588,6 +606,16 @@ export class Query<
...
+      case 'ignore':
+        return {
+          ...state,
+          error: null,
+          fetchFailureCount: 0,
+          isFetching: false,
+          isInvalidated: false,
+          isPaused: false,
+          status: typeof state.data === 'undefined' ? 'idle' : 'success',
+        }
```
TkDodo, line 618: why do we need this please? Can we just _not_ dispatch anything in case of bail-out and let the query be in that state as if `setQueryData` was never called?

src/core/queryObserver.ts
```diff
@@ -68,7 +68,7 @@ export class QueryObserver<
     TQueryKey
   >
   private previousQueryResult?: QueryObserverResult<TData, TError>
-  private previousSelectError: Error | null
+  private previousSelectError: TError | null
```
TkDodo, line 71: this change seems unrelated. Why is it necessary please?

src/core/tests/queryClient.test.tsx
```diff
@@ -137,13 +137,13 @@ describe('queryClient', () => {
     test('should create a new query if query was not found', () => {
       const key = queryKey()
       queryClient.setQueryData(key, 'bar')
-      expect(queryClient.getQueryData(key)).toBe('bar')
+      expect(queryCache.find(key)).not.toBe(undefined)
```
TkDodo, line 140: why do we need to change that test - what was wrong with the previous assertion?

src/react/tests/useQuery.test.tsx
```diff
@@ -817,10 +890,11 @@ describe('useQuery', () => {
 
   it('should re-render when dataUpdatedAt changes but data remains the same', async () => {
     const key = queryKey()
-    const states: UseQueryResult<string>[] = []
+    const states: UseQueryResult<number>[] = []
+    let queryReturnValue = 0
 
     function Page() {
-      const state = useQuery(key, () => 'test', {
+      const state = useQuery(key, () => queryReturnValue++, {
```
TkDodo, line 897: this test is named:

`'should re-render when dataUpdatedAt changes but data remains the same'`

now, data no longer stays the same. I was wondering if we're not breaking this functionality: Apparently, we would still expect `dataUpdatedAt` to be updated if a new fetch comes in, which now likely doesn't happen because of the bail out...

---

sorry for the troubles, but I think the easiest solution would be to remove the `Object.is` related bailouts and stick with the `undefined` ones?

TkDodo (changes requested):

src/core/query.ts
```diff
@@ -450,10 +466,12 @@ export class Query<
       fn: context.fetchFn as () => TData,
       abort: abortController?.abort?.bind(abortController),
       onSuccess: data => {
-        this.setData(data as TData)
+        const updatedData = this.setData(data as TData)
 
-        // Notify cache callback
-        this.cache.config.onSuccess?.(data, this as Query<any, any, any, any>)
+        if (typeof updatedData !== 'undefined' && !Object.is(prevData, updatedData)) {
```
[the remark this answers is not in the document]
phatmann: I will add polyfill.

src/core/query.ts
phatmann: If a query function returns undefined or the same data, we need to reset state after bailing out. Otherwise the query will stay in fetching state.

src/core/queryObserver.ts
phatmann: These are type errors in the latest version of Typescript. I can fix in separate PR if you prefer.

src/core/tests/queryClient.test.tsx
phatmann: getQueryData returns undefined for existing idle queries or non-existing queries. I wanted to ensure query was actually created.

src/react/tests/useQuery.test.tsx
phatmann: Your instinct to combine undefined and same-data bailout was good, since the code changes happen in the same places. Let's stick with this. It is not hard to add code to update dataUpdatedAt. I just need to enhance the 'ignore' action handler to do this.

src/core/tests/queryObserver.test.tsx
```diff
@@ -143,7 +144,7 @@ describe('queryObserver', () => {
     await sleep(1)
     await observer.refetch()
     unsubscribe()
-    expect(count).toBe(2)
+    expect(count).toBe(3)
```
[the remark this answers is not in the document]
phatmann: If the query function returned a constant value, then only the first call would have any effect because of the bail out change. So now it returns a different value on each call. This results in one extra call to `select`.

src/react/tests/useQuery.test.tsx
TkDodo: Hm, I don't really think so. When I wrote [this suggestion](https://github.com/tannerlinsley/react-query/discussions/2823#discussioncomment-1562184), I only had `setQueryData` in mind, because that's where the issue stems from: That you cannot bail out of it even if you get `undefined` passed in and can't create valid data.

I really think we should keep all changes confined to `setQueryData`. If we fetch data, we can just put it in the cache without bailing out. structural sharing will make sure that data stays referentially stable, and RQ won't even re-render observers if the data doesn't change. So what's the point in bailing out here? I think it would also remove lots of code.
Additionally, we _should_ in fact call the `onSuccess` callback of observers that the fetch was successful, even if the data stays the same. I think bailing out here does more harm than good.

phatmann: The structural sharing takes place in `Query.setData`, which is used by `setQueryData` and fetching. So if we are going to give up on the Object.is bailout for fetching, we should do so for setQueryData as well.

Similarly, the call to onSuccess takes place in `Query.setData` as well. So if we want to return onSuccess for fetching but not setQueryData, it would require some kind of flag, which is smelly.

Since these changes are for React Query 4, I am going to venture my opinion. A large part of the confusion here is that we are using `onSuccess` in two very different ways. We use it to indicate a successful fetch, and we use it to indicate a data change. Instead, we need two different callbacks, `onSuccess` when a fetch succeeds and `onDataChanged` whenever cache is updated with new data via `setQueryData` or the fetch. I can move this discussion to Discussions if that seems more appropriate.

TkDodo: > The structural sharing takes place in Query.setData, which is used by setQueryData and fetching.

this is what I meant when I said we'd need some internal restructurings. calling the functional updater + performing structural sharing / isDataEqual check can be a pure function - it only depends on `data` and `options` - both of which we have available. This only leaves us with

```
    // Set data and mark it as cached
    this.dispatch({
      data,
      type: 'success',
      dataUpdatedAt: options?.updatedAt,
    })
```

that we would need to somehow put into a function on the query, like `setSuccessData` and call it in two places: from `query.setData` and from `queryClient.setQueryData`, where we could only do it after we've computed the new data by calling the aforementioned function, so potentially, not calling `query.setData` from `queryClient.setQueryData`, thus keeping the separation between "setting data after fetching" and "setting data from setQueryData"

---

i like the idea of an extra callback - it was confusing to ppl that `onSuccess` is also called after `setQueryData`, but it's a different topic I think

phatmann: Ok, this all makes sense. I will get to work on it.

If we ever split out onDataChanged and onSuccess, we will need to do the kind of code split you are proposing above anyway. I will propose this change that in a discussion.

Only one issue to resolve: what do we do if the query function returns undefined? If we are not bailing out, I suggest we treat this as an error.

phatmann: Proposed onDataChanged callback: https://github.com/tannerlinsley/react-query/discussions/2911

TkDodo: > Only one issue to resolve: what do we do if the query function returns undefined? If we are not bailing out, I suggest we treat this as an error.

this is also what Tanner proposed yesterday for v4: we throw an error and put the query in error state. In development mode, we could additionally warn, and if possible, we should exclude undefined on type level :)

phatmann: Sounds good, I will make this change.

TkDodo:
I've created an issue for this here: https://github.com/tannerlinsley/react-query/issues/2925 and assigned you to it.

please also change the base of your PR to the `v4` branch

src/core/queryObserver.ts
TkDodo: yes lets please do this in a different branch, where we also update our TS version to the latest (4.5 as of a couple of minutes ago)

TkDodo:
@phatmann any updates here please?

phatmann:
Sorry for the activity gap. I'm preparing an app release for my employer and that is sucking up my time. I'll be back on this in about a week.

phatmann:
I should finally have time soon to get back to this. I have not forgotten about it!

TkDodo:
@phatmann hope you're having a great new year. Let me know if I can help you with the PR :)

phatmann:
Still waiting for that magical moment when my new app release has landed with no outstanding issues. Really hoping that is any day now. And then I can get back to this PR.

TkDodo:
@phatmann since we are aiming for a release soon, I would really like to get this over the finish line. If I'm not hearing back from you, I will close the PR and start a new one.

phatmann:
@TkDodo I will work on it today!

phatmann:
@TkDodo after reviewing the comments, I see that we will be better off making a new (narrower) PR and closing this one. So I will do that. Expect a new PR today.

phatmann:
https://github.com/tannerlinsley/react-query/pull/3271

PR #2906, opened 2021-11-09
feat: meta to Mutation

danielbartsch:
the same meta that already exists in Query and is accessible in
queryCache is now also usable in mutationCache and can be passed in
useMutation, just as in useQuery

Context: https://github.com/TkDodo/blog-comments/issues/29#issuecomment-963095663

### DIFF
--- a/src/core/mutation.ts
+++ b/src/core/mutation.ts
@@ -85,7 +85,7 @@ export class Mutation<
   state: MutationState<TData, TError, TVariables, TContext>
   options: MutationOptions<TData, TError, TVariables, TContext>
   mutationId: number
-  meta?: MutationMeta
+  meta: MutationMeta | undefined
 
   private observers: MutationObserver<TData, TError, TVariables, TContext>[]
   private mutationCache: MutationCache
--- a/src/core/mutation.ts
+++ b/src/core/mutation.ts
@@ -1,4 +1,4 @@
-import type { MutationOptions, MutationStatus } from './types'
+import type { MutationOptions, MutationStatus, MutationMeta } from './types'
 import type { MutationCache } from './mutationCache'
 import type { MutationObserver } from './mutationObserver'
 import { getLogger } from './logger'
@@ -14,6 +14,7 @@ interface MutationConfig<TData, TError, TVariables, TContext> {
   options: MutationOptions<TData, TError, TVariables, TContext>
   defaultOptions?: MutationOptions<TData, TError, TVariables, TContext>
   state?: MutationState<TData, TError, TVariables, TContext>
+  meta?: MutationMeta
 }
 
 export interface MutationState<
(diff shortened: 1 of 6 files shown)


src/react/tests/useMutation.test.tsx
```diff
@@ -519,4 +519,56 @@ describe('useMutation', () => {
...
+        </div>
+      )
+    }
+
+    const { getByText } = renderWithClient(queryClientAy, <Page />)
+
+    fireEvent.click(getByText('succeed'))
+    fireEvent.click(getByText('error now'))
+
+    await sleep(500)
```
TkDodo, line 567: instead of waiting an arbitrary time, we should probably just render out the error returned from the mutation and wait for it to appear on the screen (the test right above does something similar)

src/core/mutation.ts
```diff
@@ -84,6 +85,7 @@ export class Mutation<
   state: MutationState<TData, TError, TVariables, TContext>
   options: MutationOptions<TData, TError, TVariables, TContext>
   mutationId: number
+  meta?: MutationMeta
```
TkDodo, line 88:
```suggestion
  meta: MutationMeta | undefined
```

we want to always pass it to the constructor, but it might be `undefined`. With the current notation, we could theoretically not pass it to the constructor and TypeScript would be fine with it

TkDodo (changes requested):
very good work, thank you 🚀

src/react/tests/useMutation.test.tsx
danielbartsch: is that how it should be? The test checks if the 'global' `onSuccess` and `onError` are called. They shouldn't change something of the render of a component.

danielbartsch: 5ea0c1c

TkDodo: that solution also works, but is not really what I meant. I think he asserting that the callbacks have been called were totally fine.

to get around the sleep, I would've just rendered any error message, like:

```
{ isError ? <h1>Oh noes</div> : null }
```

and then instead of waiting 500ms, wait until that error is visible:

```
await waitFor(() => {
  expect(getByRole('heading').textContent).toBe('Oh noes')
})
```

then you have basically waited until the error mutation is finished, in which case both callbacks should've been called.

danielbartsch: but then where is the actual test that `meta` has been passed?

TkDodo: like you had it before - with `expect(errorMock).toHaveBeenCalledWith(myErrorMessage)` ?

danielbartsch: 4699132

TkDodo (approved):

TkDodo (changes requested):
Sorry, I totally missed that we also need to update the documentation please

danielbartsch:
b464b49
I hope this is fine

TkDodo (approved):

PR #2908, opened 2021-11-10
docs(optimistic-updates): ease wording in introduction

BerniWittmann:
the introductionary text for optimistic update was in my opinion not easily understandable. Thus, i tried to improve the wording a little bit, to make it easier to grasp the objective.

### DIFF
--- a/docs/src/pages/guides/optimistic-updates.md
+++ b/docs/src/pages/guides/optimistic-updates.md
@@ -3,7 +3,7 @@ id: optimistic-updates
 title: Optimistic Updates
 ---
 
-When you optimistically update your state before performing a mutation, there is a non-zero chance that the mutation will fail. In most cases, you can just trigger a refetch for your optimistic queries to revert them to their true server state. In some circumstances though, refetching may not work correctly and the mutation error could represent some type of server issue that won't make it possible to refetch. In this event, you can instead choose to rollback your update.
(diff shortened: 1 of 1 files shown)


BerniWittmann:
cc @alejo-monkee

docs/src/pages/guides/optimistic-updates.md
```diff
@@ -3,7 +3,7 @@ id: optimistic-updates
 title: Optimistic Updates
 ---
 
-When you optimistically update your state before performing a mutation, there is a non-zero chance that the mutation will fail. In most cases, you can just trigger a refetch for your optimistic queries to revert them to their true server state. In some circumstances though, refetching may not work correctly and the mutation error could represent some type of server issue that won't make it possible to refetch. In this event, you can instead choose to rollback your update.
+When you optimistically update your state before performing a mutation, the chance of mutation failure is not exempt. In most of these failure cases, you can just trigger a refetch for your optimistic queries to revert them to their true server state. In some circumstances though, refetching may not work correctly and the mutation error could represent some type of server issue that won't make it possible to refetch. In this event, you can instead choose to rollback your update.
```
TkDodo, line 6: I'm not a native english speaker, but maybe that's a good thing because I don't find `the chance of mutation failure is not exempt` to be easily understandable. I guess your issue was with the `non-zero chance` ? If so, how about just removing that:

```suggestion
When you optimistically update your state before performing a mutation, there is a chance that the mutation will fail. In most of these failure cases, you can just trigger a refetch for your optimistic queries to revert them to their true server state. In some circumstances though, refetching may not work correctly and the mutation error could represent some type of server issue that won't make it possible to refetch. In this event, you can instead choose to rollback your update.
```

BerniWittmann: even better so :+1: just committed that. 
yes it was mostly about the `non-zero chance`, thanks for the quick reply :)

PR #2934, opened 2021-11-12
Add pitfall case for hydration with Next.js' Incremental Static Regeneration to the docs

bennettdams:
This change will add a pitfall for hydration with Next.js' Incremental Static Regeneration to the docs.

One thing I'm not sure about: How would you like to link to another paragraph on the same page? In this commit, I do it like this:

`Please read the [general tips for this topic](#staleness-is-measured-from-when-the-query-was-fetched-on-the-server) first.`

Feel free to change the text based on your preferences, I just think having this pitfall in the docs is valuable, because I stumbled upon this issue and it took me a long time to figure out what was going on.

---

_Here is the full text for reference:_

Please read the [general tips for this topic](#staleness-is-measured-from-when-the-query-was-fetched-on-the-server) first.

You will need to synchronize your `getStaticProps` revalidation time (via `revalidate`) and the `staleTime` of your query.
Without the synchronization, it would prevent us from using the static page & the data to hydrate of Next.js' revalidation result.

Via `getStaticProps`, we hydrate the query state with fresh data every revalidation cycle.
Let's say revalidation is configured to be every 5 minutes. After 5 minutes, `getStaticProps` is executed again when a new user request comes in.
But everything inbetween these 5 minutes, the "old" props result will be sent to the client.

Now, let's say a query has a `staleTime` of 3 minutes. This means, that React Query wants to refetch a query after 3 minutes (under most circumstances).
This stale time is bound to the query and the query is bound to the query client. We create the client in `getStaticProps`, which only
gets created once per revalidation cycle. This means that the internal timer of `staleTime` will continue to run.
Let's imagine in our example (5 minutes revalidation cycle and 3 minutes stale time), a request is sent at minute 4 of the revalidation cycle.
Next.js does not need to revalidate again (because it's less than 5 minutes), so it sends the already existent client with the running `staleTime` timer.
The query is considered stale after 3 minutes, our request comes in after 4 minutes, so React Query wants to refetch.
The pitfall: This refetching will not happen in `getStaticProps` automatically, because the query is not "used" on the server.
Rather, Next.js sends the existing client to the browser with it's data like it always does. The client therefore has the existing data
of the last revalidation cycle and can use it for hydration.
But: This query will immediately refetch in the browser, because the `staleTime` was reached, even though the data is there.

### DIFF
--- a/docs/src/pages/guides/ssr.md
+++ b/docs/src/pages/guides/ssr.md
@@ -124,9 +124,9 @@ gets created once per revalidation cycle. This means that the internal timer of 
 Let's imagine in our example (5 minutes revalidation cycle and 3 minutes stale time), a request is sent at minute 4 of the revalidation cycle.
 Next.js does not need to revalidate again (because it's less than 5 minutes), so it sends the already existent client with the running `staleTime` timer.
 The query is considered stale after 3 minutes, our request comes in after 4 minutes, so React Query wants to refetch.
+
 The pitfall: This refetching will not happen in `getStaticProps` automatically, because the query is not "used" on the server.
-Rather, Next.js sends the existing client to the browser with it's data like it always does. The client therefore has the existing data
-of the last revalidation cycle and can use it for hydration.
+Rather, Next.js sends the existing client to the browser with its data like it always does. The client therefore has the existing data of the last revalidation cycle and can use it for hydration.
 But: This query will immediately refetch in the browser, because the `staleTime` was reached, even though the data is there.
 
 ## Using Other Frameworks or Custom SSR Frameworks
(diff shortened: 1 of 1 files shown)


TkDodo:
correct me if I'm wrong, but isn't that exactly what the docs are saying at the bottom of that page?

> This refetching of stale queries is a perfect match when caching markup in a CDN! You can set the cache time of the page itself decently high to avoid having to re-render pages on the server, but configure the staleTime of the queries lower to make sure data is refetched in the background as soon as a user visits the page. Maybe you want to cache the pages for a week, but refetch the data automatically on page load if it's older than a day?

I think this is exactly your example of 5 minutes markup caching (with ISR, but still), but the query itself is only valid for 3 minutes. It depends on the use-case of course, but the static generated site might contain a lot more things than just this one query. Having that cached for a longer time than a single query can make sense.

This is also not quite correct:

> Now, let's say a query has a staleTime of 3 minutes. This means, that React Query wants to refetch a query after 3 minutes (under most circumstances).

React-query will not refetch automatically after 3 minutes. It will just mark the query as stale. If you work on your screen for 8 more hours, nothing will happen. Only if you mount a new observer, refetch the window etc, RQ will do a background refetch.

bennettdams:
> correct me if I'm wrong, but isn't that exactly what the docs are saying at the bottom of that page?

I agree that it's very similar. It's just that for me, this explanation would have been helpful, so I thought it might be for others.

PR #2935, opened 2021-11-12
fix(useQuery): don't throw error if errorBoundary has just been reset

TkDodo:
the fix for disabled queries was wrong, because disabled queries still need to throw if they are fetching due to some other means, like `refetch`. However, if a query has just been reset, we want to skip throwing for one render cycle. The useEffect that clears the reset will then make sure that further errors will be thrown

### DIFF
--- a/src/react/tests/QueryResetErrorBoundary.test.tsx
+++ b/src/react/tests/QueryResetErrorBoundary.test.tsx
@@ -204,6 +204,65 @@ describe('QueryErrorResetBoundary', () => {
     consoleMock.mockRestore()
   })
 
+  it('should throw error if query is disabled and manually refetched', async () => {
+    const key = queryKey()
+
+    const consoleMock = mockConsoleError()
+
+    function Page() {
+      const { data, refetch, status } = useQuery(
+        key,
+        async () => {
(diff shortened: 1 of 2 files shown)


TkDodo:
@WeiShengv99 I tried your lates example with this preview, and it seems to work fine:

https://codesandbox.io/s/tannerlinsley-react-query-basic-forked-vspe0?file=/src/index.js

WeiShengv99:
> @WeiShengv99 I tried your lates example with this preview, and it seems to work fine:
> 
> [codesandbox.io/s/tannerlinsley-react-query-basic-forked-vspe0?file=/src/index.js](https://codesandbox.io/s/tannerlinsley-react-query-basic-forked-vspe0?file=/src/index.js)

Yes, this pr fixed my problem. 
By the way , i have some question about this  behavior .

`enable:false` instance  get  cache data  in second component, sometimes i dont know i have cache , before  i set `enable:true`, i thought i get data is undefined,  but i got  a  queryInstance's  status is successed ,   this is a litter confusing ,  
what do you think about that , its  right now   ?

if  `enable:false` cant get cache data , the throw error problem also fixed.

PR #2947, opened 2021-11-15
Fix/2901 improve devtools a11y

### LINKED ISSUE #2901: Improve aria-tagging of Devtools
**Describe the bug**

1. Devtools is now defaulting to being rendered as `footer`, which doesn't make a lot of sense to me.
2. Menu-items for sorting seem to be rendered and visible in the Accessibility Tree regardless of open/closed state.
3. The Devtools toggle-button have no aria-attributes for toggled state or link to the panel it toggles.

**To Reproduce**
1. Render React-Query Devtools in an app
2. Explore the DOM tree and the Accessibility tree

**Expected behavior**
1. A better default should probably be `aside`, and it should also have an `aria-label` (like "React Query Devtools" for the landmark it generates.
2. Not sure if this is an actual issue, or just how the accessibility tree works, but I'm thinking the panel should maybe either not be rendered at all, or be rendered with `aria-hidden` or something like that when not opened?
3. Toggle-button should have appropriate aria attributes (`aria-controls`, `aria-expanded` and be linked to the panel as it should.

**Screenshots**
In the screenshot below, the React Query Devtools "root" is the generic `contentinfo` node
image
Changing the element-type to `aside` and adding an `aria-label` (just edited the HTML in Chrome for this) gives this instead, which to me at least looks more appropriate:
image

Andrewnt219:
Closes #2901 

## Test plan

Devtools is rendered as `aside` by default
image

Button aria: expanded
image

Button aria: collapsed
image

Hide `menuitem` when Devtools is closed 
image

### DIFF
--- a/src/devtools/devtools.tsx
+++ b/src/devtools/devtools.tsx
@@ -55,7 +55,7 @@ interface DevtoolsOptions {
   /**
    * Use this to render the devtools inside a different type of container element for a11y purposes.
    * Any string which corresponds to a valid intrinsic JSX element is allowed.
-   * Defaults to 'footer'.
+   * Defaults to 'aside'.
    */
   containerElement?: string | any
 }
@@ -91,7 +91,7 @@ export function ReactQueryDevtools({
   closeButtonProps = {},
   toggleButtonProps = {},
   position = 'bottom-left',
-  containerElement: Container = 'footer',
+  containerElement: Container = 'aside',
 }: DevtoolsOptions): React.ReactElement | null {
   const rootRef = React.useRef<HTMLDivElement>(null)
   const panelRef = React.useRef<HTMLDivElement>(null)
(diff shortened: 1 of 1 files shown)


src/devtools/devtools.tsx
```diff
@@ -494,7 +510,7 @@ export const ReactQueryDevtoolsPanel = React.forwardRef<
             maxHeight: '100%',
             overflow: 'auto',
             borderRight: `1px solid ${theme.grayAlt}`,
-            display: 'flex',
+            display: isOpen ? 'flex' : 'none',
```
Andrewnt219, line 513: This change causes one test to fail. I'm investigating.
```bash
useInfiniteQuery › should render correct states even in case of concurrent renders with different properties
```

TkDodo (approved):

PR #2948, opened 2021-11-15
docs(cancel): add graphql-request example

arnaudbzn:
It is possible to set an `AbortSignal` in the `GraphQLClient` constructor.

```js
const query = useQuery('todos', ({ signal }) => {
  const client = new GraphQLClient(endpoint, {
    signal,
  });
  return client.request(query, variables)
})
```

But unfortunately, it is not possible to define a `signal` per request:
https://github.com/prisma-labs/graphql-request/issues/182

It means that the GraphQL Code Gen React Query plugin cannot support a `signal` for the moment with graphql-request :pensive:

The solution is to:
- PR graphql-request to support a signal per request
- PR GraphQL Code Gen React Query Plugin to send the signal to the graphql-request request.

### DIFF
--- a/docs/src/pages/guides/query-cancellation.md
+++ b/docs/src/pages/guides/query-cancellation.md
@@ -97,6 +97,19 @@ const query = useQuery('todos', ({ signal }) => {
 })
 ```
 
+## Using `graphql-request`
+
+An `AbortSignal` can be set in the `GraphQLClient` constructor.
+
+```js
+const query = useQuery('todos', ({ signal }) => {
+  const client = new GraphQLClient(endpoint, {
+    signal,
+  });
+  return client.request(query, variables)
+})
+```
+
 ## Manual Cancellation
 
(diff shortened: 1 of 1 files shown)


TkDodo:
thank you. do you know if there are other graphql fetching libs that would support it right now? Like maybe [@urql/core](https://formidable.com/open-source/urql/docs/api/core/)?

arnaudbzn:
New graphq-request PR to support a `signal` argument in the request function:
https://github.com/prisma-labs/graphql-request/pull/303

arnaudbzn:
> thank you. do you know if there are other graphql fetching libs that would support it right now? Like maybe [@urql/core](https://formidable.com/open-source/urql/docs/api/core/)?

I have not found in the `urql` docs and code where to set a `signal`.
I may not use `urql` as a graphql fetcher for React Query, `graphql-request` is super lean and does the job.
I'm not aware of other basic graphql fetchers.

TkDodo:
oh wow, you already PR'ed to graphql-request 🚀 

should we wait with this docs until it lands, or should we merge this doc PR right away?

arnaudbzn:
> oh wow, you already PR'ed to graphql-request 🚀
> 
> should we wait with this docs until it lands, or should we merge this doc PR right away?

:grin:
We can merge now because it is useful to know the current possibilities with graphql-request and I've no visibility on when the PR will be merged... :snail:

I'll update the doc when the graphql-request PR will be merged in another RQ PR.

PR #2950, opened 2021-11-15
fix(useQuery): cleanup queries even if they have been fetching

### LINKED ISSUE #2946: leak in query garbage collection
**Describe the bug**

Queries that should be garbage collected because the `cacheTime` has elapsed are kept in memory forever if the query was fetching when the last observer was removed.

**To Reproduce**
https://codesandbox.io/s/youthful-neumann-rh8t6?file=/src/index.js

note:
- `cacheTime` is 2 seconds
- the query takes 5 seconds to fetch
- it's best to wait until the query has loaded once, then trigger a refetch via window focus and click the button which will unmount the query while the fetch is happening.

**Expected behavior**
queries to be garbage collected correctly

**Additional context**

I think the problem is that `optionalRemove`, once it determines that it has no observers but doesn't remove the query because it's still fetching doesn't restart the timer:

https://github.com/tannerlinsley/react-query/blob/c1ae82ba188fd5abda5e256cac070145e5941447/src/core/query.ts#L216-L220

TkDodo:
fixes #2946

### DIFF
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -214,8 +214,12 @@ export class Query<
   }
 
   private optionalRemove() {
-    if (!this.observers.length && !this.state.isFetching) {
-      this.cache.remove(this)
+    if (!this.observers.length) {
+      if (this.state.isFetching) {
+        this.scheduleGc()
+      } else {
+        this.cache.remove(this)
+      }
     }
   }
 
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -163,9 +163,11 @@ export class Query<
   private observers: QueryObserver<any, any, any, any, any>[]
   private defaultOptions?: QueryOptions<TQueryFnData, TError, TData, TQueryKey>
   private abortSignalConsumed: boolean
+  private hadObservers: boolean
 
   constructor(config: QueryConfig<TQueryFnData, TError, TData, TQueryKey>) {
     this.abortSignalConsumed = false
+    this.hadObservers = false
     this.defaultOptions = config.defaultOptions
     this.setOptions(config.options)
     this.observers = []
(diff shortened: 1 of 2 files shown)


TkDodo:
@arnaudbzn This is a fix for:
- #2946 

however, it re-introduces the issue worked around in
- #2821

The test you added there fails again :(

I'd appreciate any help here, because I really think the fix here is better, as it re-schedules garbage collection if it wanted to remove a query, but couldn't because it's still fetching and thus ensures that eventually, all queries that are eligible for garbage collection will also be collected.

TkDodo:
I was able to narrow the issue down and work around it, but I'm not super happy with it. The root cause seems to be that we throw the Promise _before_ we even subscribe, because subscription happens in an effect, and the optimistic result puts us right in loading state, so we throw during render when we have suspense on.

This is all intended so that suspense queries never render in loading state, however, it also means that we have no subscribers - we'll get the subscription once the component resumes and the effect runs. This is also why we have a staleTime of 1 second - so that the component does not immediately re-fetch :)

Now the problem comes with a cacheTime that is smaller than the time the fetch actually takes. The query tries to GC, but previously didn't because the query is still fetching. Now, I'm rescheduling the GC to avoid the leak, which led to suspense queries being prematurely garbage collected again. I wanted to check for suspense, but we have no information in the query, and we also have no observer yet.

The workaround in a840e7cdea030d45215d5e7b8d299da5b8a500be is to keep track if the query ever had any observers, and if we are still fetching while trying to GC but we never had any observers, it probably means we are suspending (how else could we be fetching without ever having a subscriber?) so I just fall back to the previous behavior - not garbage collecting at all ...

I hope this is acceptable, please have a look @tannerlinsley @arnaudbzn

TkDodo:
a different fix that I could offer is to generally _not_ kickoff garbage collection in the constructor, but only when the last observer unmounts. This means that if you create an observer, but never subscribe to it, the query would not gc. This is something that you cannot do with useQuery and friends, because it always also subscribes. But theoretically, you could create an observer and never subscribe to it.

this works around the suspense issue because we only add observers after we've loaded as we throw before that. I think it's the more elegant solution, code wise, but it might have a larger impact.

Issue #2943, opened 2021-11-14
Docs CSS Issue. Text get hidden


kothariji:
**Describe the bug**
There is `width` issue in Right Sidebar Layout of Docs. The text gets hidden, and the horizontal scroll bar is at the bottom of the page (so its hardly visible, until a user scrolls to bottom of page.) Also I find out that its just a matter of `25 px`.

## Screenshots of API Reference Section

 https://react-query.tanstack.com/reference/QueryClient
![image](https://user-images.githubusercontent.com/60866749/141676021-8c4cefef-af6b-483e-96cb-5265dcd6e72f.png)

https://react-query.tanstack.com/reference/onlineManager
![image](https://user-images.githubusercontent.com/60866749/141676255-e1a629b7-7a79-41a3-a8ac-dbb0a7f8767b.png)

**To Reproduce**
1. Go to ['React-Query Reference'](https://react-query.tanstack.com/reference/QueryClient)
2. Check the right side bar
3. See `li` like queryClient.prefetchInfiniteQuery

**Expected behavior**
We can prevent this behavior by increasing the width by `25 px` 

#### Before
![image](https://user-images.githubusercontent.com/60866749/141676496-a2e29309-8218-4ccf-8074-b5afb5dd3df4.png)

#### After modifying width to `225 px`
![image](https://user-images.githubusercontent.com/60866749/141676482-8766c1b6-a832-488f-a12b-e77671d6c8d7.png)

# Solution - Changing width to `225 px` helps prevent this horizontal scroll from Entire API reference Section.

Also please let me know if I can work on this issue 🙏 
Thank You

PR #2952, opened 2021-11-15
docs(reference): fixed right sidebar width issue

nikasepiskveradze:
This is the possible solution of this issue #2943


TkDodo:
this will work for now, but will break again if we have any larger texts in the sidebar :)

PR #2963, opened 2021-11-17
2920 mutation cache duration

### LINKED ISSUE #2920: Mutation Cache Duration
As discussed in [this discussion](https://github.com/tannerlinsley/react-query/discussions/2768), the MutationCache is currently ever growing. This can be a problem because mutations are not shared like queries are, and are also never garbage collected. If mutations return big responses, those will be held in memory indefinitely.

### Proposed solution

Add a `cacheTime` to `MutationCache`, and default to 5 minutes (same time as for queries). We can't remove mutations from the cache instantly after the last observer unmounts, because callacks on `useMutation` might fire after that.

For that, we also need to notify the cache when an observer is added or removed for a mutation. The code for scheduling garbageCollection could be shared between queries and mutations via a base class that both `Query` and `Mutation` could extend.

TkDodo:
fixes #2920


Noitidart:
Oh wow thank you Sir!!

docs/src/pages/reference/useMutation.md
```diff
@@ -39,6 +40,9 @@ mutate(variables, {
   - **Required**
   - A function that performs an asynchronous task and returns a promise.
   - `variables` is an object that `mutate` will pass to your `mutationFn`
+- `cacheTime: number | Infinity`
```
Noitidart, line 43: Should we mention a default here is Infinity?

TkDodo: The default is 5 minutes, like it is for queries.

Noitidart: oh got it, thanks!

Noitidart:
I think `cacheTime` isn't in the typescript of the `useMutation`. I keep getting errors. I'll take a look at PR it.

ShiChenCong:
Does useMutation cacheTime work with you？ It seems not working in my project

ShiChenCong:
> > Does useMutation cacheTime work with you？ It seems not working in my project
> 
> can you elaborate?

https://github.com/TanStack/query/issues/5058

PR #2969, opened 2021-11-18
2964 changes to on success callback

### LINKED ISSUE #2964: changes to onSuccess callback
### Discussed in https://github.com/tannerlinsley/react-query/discussions/2911

<sup>Originally posted by **phatmann** November 11, 2021</sup>
Calling `QueryClient.setQueryData` method triggers the `onSuccess` callback. For many users, including me, this is unexpected behavior, and can be the source of recursion issues, since the `onSuccess` handler might make calls to `setQueryData`.

I propose we create a new `onDataChanged` callback. This would be trigged when cache data is changed via `setQueryData` or via a fetch.  The `onSuccess` callback would **only** be called after a successful fetch.

For completeness, the same `onDataChanged` callback would be added to QueryCache.</div>

TkDodo:
closes #2964


TkDodo:
@phatmann FYI in case you want to check the implementation. This will likely conflict with your PR

PR #2987, opened 2021-11-19
feat(QueryObserver): track queries as default

babycourageous:
The tests were fine after removing `notifyOnChangePropsExclusion` but there are a bunch of tests failing after the updates to `notifyOnChangeProps`. 

I gather this probably makes some sense since the underlying default behavior of the query is changing? I'm just not sure how to go about updating them - change how `useQuery` gets called (ie with { notifyOnChangeProps: 'all' }) or change the `expects` outcome - perhaps it depends on the test. 

Also, do we need to add any new tests now that some of these default assumptions change?

I am also realizing I still have the corresponding docs for "tracked" to update.


TkDodo:
yeah, the test failures make sense because components re-render less often. I think we really need to see this on a case-by-case basis I'm afraid. for example, the first failure:

> should return the correct states for an unsuccessful query

only observes the `state.status` field, but then checks for different `failureCount`s. Here, the best solution would to actually observe `state.failureCount` by printing it on the screen inside a div or so. That would make the test pass.

For tests that just `return null` and just track an assert renders, I think it's better to just set `notifyOnChangeProps: 'all'`. I am re-writing most of those tests to a better / more stable approach (less hard-coded timeouts) in the react-18 branch, and I can take a look at this there :)

Let's just make sure we have a sufficient coverage for the tracking behaviour :)

src/core/queryObserver.ts
```diff
@@ -596,27 +596,25 @@ export class QueryObserver<
...
       return true
     }
 
-    const includedProps =
-      notifyOnChangeProps === 'tracked'
-        ? this.trackedProps
-        : notifyOnChangeProps
+    const includedProps = !notifyOnChangeProps
+      ? this.trackedProps
+      : notifyOnChangeProps
```
TkDodo, line 611: negated ifs are pretty hard to read imo, so is this the same as:

```
    const includedProps = notifyOnChangeProps ?? this.trackedProps
```

?

src/core/types.ts
```diff
@@ -160,11 +160,7 @@ export interface QueryObserverOptions<
    * When set to `['data', 'error']`, the component will only re-render when the `data` or `error` properties change.
    * When set to `tracked`, access to properties will be tracked, and the component will only re-render when one of the tracked properties change.
```
TkDodo, line 161: the comment is no longer accurate here :)

TkDodo (commented):
looks good so far 👍

src/core/types.ts
babycourageous: Yah - this is for sure part of the docs i need to update

src/core/queryObserver.ts
babycourageous: Totally agree on using `??`

babycourageous:
@TkDodo gonna follow your advice for the tests and see if I can get the fails back to 0.

babycourageous:
@TkDodo I think I got the tests passing 

Used your guidance for `useQuery` and `useInfiniteQuery` but diverted a bit for the `QueryResetErrorBoundary` tests. I set the queries in question to `{notifyOnChangeProps:'all'}` for that file. 
Not sure if the diversion is ok or not.

Let me know if we need anything else in test land. In the meantime, I'm gonna move on to the docs/type def comments updates.

src/core/queryObserver.ts
```diff
@@ -596,27 +596,23 @@ export class QueryObserver<
...
-    const { notifyOnChangeProps, notifyOnChangePropsExclusions } = this.options
+    const { notifyOnChangeProps } = this.options
 
-    if (!notifyOnChangeProps && !notifyOnChangePropsExclusions) {
+    if (notifyOnChangeProps === 'all') {
       return true
     }
 
-    if (notifyOnChangeProps === 'tracked' && !this.trackedProps.length) {
+    if (!notifyOnChangeProps && !this.trackedProps.length) {
```
TkDodo, line 605: let’s combine these two if statements that return true with an ||

TkDodo:
Hmm I think we should have at least one test for QueryResetErrorBoundary that uses the tracked properties. If we can’t get this to work then maybe we have a general issue there?

TkDodo:
I've found the issue - it's actually a bug with tracked properties that I will fix on master: if you `useErrorBoundary`, you're likely not observing the `error`, because you're not using the returned `error` field - the ErrorBoundary handles the error for you. So when the actual error then happens, we don't inform the observer about it, because we're only informing them if a prop has changed that they are interested in.

The solution would be to always track the `error` property if you've set `useErrorBoundary`. I'll add a bugfix.

TkDodo:
see:
- https://github.com/tannerlinsley/react-query/pull/2993

TkDodo:
Fix is merged, I also updated the v4 branch. Please solve the conflicts, and I think you should be able to revert most of the changes to the `QueryResetErrorBoundary` tests :)

babycourageous:
I've reverted those QueryResetErrorBoundary tests, combined those two `if`s, and updated docs. Happy to change any of those doc changes if ya need!

TkDodo:
@all-contributors please add @babycourageous for code, docs

PR #2996, opened 2021-11-20
refactor: Remove deprecated promise cancel

### LINKED ISSUE #2915: remove deprecated .cancel on promises
With the addition of out-of-the-box support for Query Cancellation via AbortController and AbortSignal (v3.30.0) we will remove the old way to cancel Promises by providing a .cancel method.

- [ ] remove old code
- [ ] update docs

prateek3255:
- Removes cancel method from wherever possible
- Updates query cancellation tests to use abort controller and remove tests using the older method
- Update the migration docs

Fixes #2915

### DIFF
--- a/docs/src/pages/guides/migrating-to-react-query-4.md
+++ b/docs/src/pages/guides/migrating-to-react-query-4.md
@@ -148,6 +148,10 @@ Since these plugins are no longer experimental, their import paths have also bee
 + import { createAsyncStoragePersister } from 'react-query/createAsyncStoragePersister'
 ```
 
+### The `cancel` method on promises is no longer supported
+
+The [old `cancel` method](/query-cancellation#old-cancel-function) that allowed you to define a `cancel` function on promises which were then used by the library to support query cancellation has been removed. We recommend you to use the [newer API](/query-cancellation) (introduced with v3.30.0) for query cancellation that uses the [`AbortController` API](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) internally and provides you with an [`AbortSignal` instance](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) for your query function to support query cancellation.
+
 ## New Features 🚀
 
 ### Mutation Cache Garbage Collection
--- a/docs/src/pages/guides/migrating-to-react-query-4.md
+++ b/docs/src/pages/guides/migrating-to-react-query-4.md
@@ -150,7 +150,7 @@ Since these plugins are no longer experimental, their import paths have also bee
 
 ### The `cancel` method on promises is no longer supported
 
-The [old `cancel` method](../guides/query-cancellation#old-cancel-function) that allowed you to define a `cancel` function on promises which were then used by the library to support query cancellation has been removed. We recommend you to use the [newer API](../guides/query-cancellation) (introduced with v3.30.0) for query cancellation that uses the [`AbortController` API](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) internally and provides you with an [`AbortSignal` instance](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) for your query function to support query cancellation.
+The [old `cancel` method](../guides/query-cancellation#old-cancel-function) that allowed you to define a `cancel` function on promises, which was then used by the library to support query cancellation, has been removed. We recommend to use the [newer API](../guides/query-cancellation) (introduced with v3.30.0) for query cancellation that uses the [`AbortController` API](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) internally and provides you with an [`AbortSignal` instance](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) for your query function to support query cancellation.
 
 ## New Features 🚀
 
--- a/docs/src/pages/guides/migrating-to-react-query-4.md
+++ b/docs/src/pages/guides/migrating-to-react-query-4.md
@@ -150,7 +150,7 @@ Since these plugins are no longer experimental, their import paths have also bee
 
 ### The `cancel` method on promises is no longer supported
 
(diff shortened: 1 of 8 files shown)


prateek3255:
@TkDodo I wasn't able to remove this method completely from the retryer here -

https://github.com/tannerlinsley/react-query/blob/1445e8a8909615220b96632fe58dd38e34f50c6e/src/core/retryer.ts#L149-L153

because it is internally by infiniteQuery to cancel promises since it has its own `AbortController` -

https://github.com/tannerlinsley/react-query/blob/1445e8a8909615220b96632fe58dd38e34f50c6e/src/core/infiniteQueryBehavior.ts#L148-L151

docs/src/pages/guides/migrating-to-react-query-4.md
```diff
@@ -148,6 +148,10 @@ Since these plugins are no longer experimental, their import paths have also bee
 + import { createAsyncStoragePersister } from 'react-query/createAsyncStoragePersister'
 ```
 
+### The `cancel` method on promises is no longer supported
+
+The [old `cancel` method](../guides/query-cancellation#old-cancel-function) that allowed you to define a `cancel` function on promises which were then used by the library to support query cancellation has been removed. We recommend you to use the [newer API](../guides/query-cancellation) (introduced with v3.30.0) for query cancellation that uses the [`AbortController` API](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) internally and provides you with an [`AbortSignal` instance](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) for your query function to support query cancellation.
```
TkDodo, line 153:
```suggestion
The [old `cancel` method](../guides/query-cancellation#old-cancel-function) that allowed you to define a `cancel` function on promises, which was then used by the library to support query cancellation, has been removed. We recommend to use the [newer API](../guides/query-cancellation) (introduced with v3.30.0) for query cancellation that uses the [`AbortController` API](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) internally and provides you with an [`AbortSignal` instance](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) for your query function to support query cancellation.
```

TkDodo:
looks good. one thing I think we can remove for sure is `this.isTransportCancelable` on the `retryer`. It is only used here:

https://github.com/tannerlinsley/react-query/blob/542cc99fafe972e34a6ef59d691b4bddc07868e8/src/core/query.ts#L315-L324

but every transport is now cancelable - and we check if the user consumed the `AbortSignal` with `abortSignalConsumed`. If the user doesn't use the signal, we'll still let the promise resolve so it can be cached - and if they have used the signal, we can abort.

TkDodo:
> I wasn't able to remove this method completely from the retryer here -

I also don't see at the moment how to do that. Tagging @jcready as he implemented the AbortController, maybe he knows. If not, I think its also good as long as all tests pass :)

prateek3255:
> one thing I think we can remove for sure is this.isTransportCancelable on the retryer. It is only used here:

I tried removing it but a couple of tests like `should stop fetching additional pages when the component is unmounted` start failing in `useInfiniteQuery`.

TkDodo:
> I tried removing it but a couple of tests like should stop fetching additional pages when the component is unmounted start failing in useInfiniteQuery.

right yeah, I think it's related to the other thing you mentioned - using the `cancel` function on the `finalPromise`. I'm fine with keeping it if we don't know a better solution 👍

src/core/infiniteQueryBehavior.ts
```diff
@@ -153,9 +148,6 @@ export function infiniteQueryBehavior<
         finalPromiseAsAny.cancel = () => {
           cancelled = true
           abortController?.abort()
-          if (isCancelable(promise)) {
-            promise.cancel()
-          }
         }
```
jcready, line 151: I think you _might_ be able to get rid of this internal `.cancel` function by replacing this with

```ts
context.signal?.addEventListener('abort', () => {
  cancelled = true;
  abortController?.abort()
})
```

prateek3255: Thanks, this seems to work now, although the signal wasn't being passed down to the `QueryBehavior`'s `FetchContext`, so I added the signal to it similar to how we added it to `QueryFunctionContext`

prateek3255:
@TkDodo I was able to remove the internal cancel function, and was also able to remove `isCancellable` method and its related interface and the `this.isTransportCancellable` method as you mentioned. But the test for Node 16 is failing, although I am not sure why its only happening in CI because I tried running the test with Node 16 on my local system and it seemed to be running fine.

TkDodo:
@prateek3255 AbortController only exists in node 15 and above, so it makes sense that it passes in 14 but not in 16.

But I re-ran the checks, so maybe it's just a flaky test? Can you maybe try running it a bunch of times locally to see if it fails, and try to make it a bit more stable?

src/core/query.ts
```diff
@@ -65,6 +65,7 @@ export interface FetchContext<
 > {
   fetchFn: () => unknown | Promise<unknown>
   fetchOptions?: FetchOptions
+  signal?: AbortSignal
```
TkDodo, line 68: this looks unused - is this maybe a leftover?

src/core/query.ts
```diff
@@ -412,6 +413,17 @@ export class Query<
...
 
+    Object.defineProperty(context, 'signal', {
+      enumerable: true,
+      get: () => {
+        if (abortController) {
+          this.abortSignalConsumed = true
+          return abortController.signal
+        }
+        return undefined
+      },
```
TkDodo, line 424: is there a way to extract this function into a const so that we can use the same getter for both signals?

TkDodo (changes requested):

src/core/query.ts
prateek3255: No, I had to add it because we weren't passing the signal in the `FetchContext` for the `QueryBehaviour` as I mentioned in this comment https://github.com/tannerlinsley/react-query/pull/2996#discussion_r753832455

src/core/query.ts
prateek3255: The whole `Object.defineProperty` was exactly the same, so I moved the whole thing to a separate function used in both of the contexts.

prateek3255:
> Can you maybe try running it a bunch of times locally to see if it fails and try to make it a bit more stable?

@TkDodo I tried running it a bunch of times on both Node 15/16 versions, but it didn't fail at all locally

src/core/query.ts
TkDodo: ah, it just looks unused, but its added via `Object.defineProperty` 👍

TkDodo:
@prateek3255 alright then, since this goes to v4, I'd just merge it, and lets see if we have any errors showing up in other PRs 🚀

TkDodo (approved):

PR #3052, opened 2021-12-04
fix(queriesObserver): fix getOptimisticResult, order of returned observers

jvuoti:
Fix for `queriesObserver.getOptimisticResult` to correctly call `getOptimisticResult` on the observers. 

This was an issue I introduced in PR #2866. There I had needed to tweak [two](https://github.com/tannerlinsley/react-query/pull/2866/files#diff-27ab6cf11862f134526488fd6c22381bf484a5ad9a6d7dc36535e1d3e2d5aa42R209) [original](https://github.com/tannerlinsley/react-query/pull/2866/files#diff-27ab6cf11862f134526488fd6c22381bf484a5ad9a6d7dc36535e1d3e2d5aa42R131) tests, which now in hindsight were off due to this issue. I now reverted those tests back to the original form.

Now testing the issue, I also noticed a small issue with the order of the observers returned by `findMatchingObservers`: with `keepPreviousData: true`, any repurposed observers would be returned last, potentially in different order than the given queries. 

The fix is just to sort the observer entries by the order of the queries before returning them. This also gets rid of an extra render at least in some tests. As this was a rather small fix, I included the commit for that fix here, but I can also put it in a separate PR if that would be cleaner.

### DIFF
--- a/src/core/queriesObserver.ts
+++ b/src/core/queriesObserver.ts
@@ -65,7 +65,7 @@ export class QueriesObserver extends Subscribable<QueriesObserverListener> {
 
   getOptimisticResult(queries: QueryObserverOptions[]): QueryObserverResult[] {
     return this.findMatchingObservers(queries).map(match =>
-      match.observer.getCurrentResult()
+      match.observer.getOptimisticResult(match.defaultedQueryOptions)
     )
   }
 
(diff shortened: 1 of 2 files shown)


TkDodo:
> Now testing the issue, I also noticed a small issue with the order of the observers returned by findMatchingObservers: with keepPreviousData: true, any repurposed observers would be returned last, potentially in different order than the given queries.

I actually _did_ notice this in the other PR, but then somehow thought it didn't matter 🙈

TkDodo (approved):

TkDodo:
@all-contributors add @jvuoti for code

PR #3064, opened 2021-12-07
Feature/react 18

TkDodo:


src/reactjs/useBaseQuery.ts
```diff
@@ -76,30 +72,21 @@ export function useBaseQuery<
       )
   )
 
-  let result = observer.getOptimisticResult(defaultedOptions)
+  const result = observer.getOptimisticResult(defaultedOptions)
 
-  React.useEffect(() => {
-    mountedRef.current = true
+  useSyncExternalStore(
```
Ephem, line 77: Is it intentional to not do anything with the result here? How does that work?

src/reactjs/tests/ssr-hydration.test.tsx
```diff
@@ -79,14 +115,17 @@ describe('Server side rendering with de/rehydration', () => {
 
     // Check that we have no React hydration mismatches
     expect(mockLogger.error).not.toHaveBeenCalled()
-    expect(fetchDataSuccess).toHaveBeenCalledTimes(1)
+    expect(fetchDataSuccess).toHaveBeenCalledTimes(2)
```
Ephem, line 118: The point of this expect is to check that we did not have to actually go and fetch the data on the client, because hydration worked and it was already there. Was this a temporary fix to get tests working for now, or was there some other reason?

Ephem: I dug into this, and I think something was fishy with the tests to begin with. As we've documented, since `staleTime` defaults to 0, unless you set it to something else we automatically refetch on mount. So this should probably have been 2 all along and these new tests and wrapping the new hydration in an `act` uncovered that.

I think we should update these tests to have a `staleTime` instead and change this back to 1 (requires updating expectedMarkup as well). Then we can also run them in both the 17 and 18 variations. I can take care of that while working on hydration.

TkDodo: @Ephem I think this is the same conclusion I cam to when changing it:

https://github.com/tannerlinsley/react-query/commit/a4b456431519dae67f6005602b81f9eedbb6b302

Changing the test to make sure we don't fetch on the client by setting a staleTime is a good idea 👍

src/reactjs/useBaseQuery.ts
TkDodo: it's a bit hacky. We need a stable result to subscribe to, which is `observer.getCurrentResult()`. However, the actual result displayed to the user is two lines above:

```
```

the optimistic result makes sure that we go into loading state immediately when mounting a component, even though it is not yet fetching. However, the result itself is not stored.

When we re-render because `getCurrentResult()` changes, we compute the optimistic result again, which will yield the same result, unless we are doing optimistic stuff.

src/reactjs/tests/ssr-hydration.test.tsx
Ephem: I read the linked discussion, sorry for causing that confusion in the first place.. 😄  I think it would have made sense to write the tests with a set staleTime to begin with (the common case for this) and have a single test to test the refetch with staleTime 0.

src/reactjs/useBaseQuery.ts
Ephem: I see! So if I understand this correctly, this is safe because `getOptimisticResults()` only rely on things that we know will trigger re-renders elsewhere (e.g. via the `uSES` and `observer.getCurrentResult()`)?

Say `getOptimisticResults` would rely on some other piece of external state, that was not linked to `observer.getCurrentResult()`, in that case we would have a bug because that state could change without triggering a re-render of the component?

In a sense, we _could_ conceptually do something like this?

```js
const currentResult = useSyncExternalStore(
  React.useCallback(
    onStoreChange =>
      observer.subscribe(notifyManager.batchCalls(onStoreChange)),
    [observer]
  ),
  () => observer.getCurrentResult(),
  () => observer.getCurrentResult()
)

const optimisticResults = this.deriveOptimisticResults(currentResult)
```

Not a suggestion, just trying to formulate my understanding of why this hack is safe. 😄

Ephem: To be clear I'm trying to understand and explore 3 things:

* If and why this is safe (seems to be!)
* Potential avenues for this to _become_ unsafe via changes in other places of the code
* How optimistic results and hydration play together (because it seems conceptually like the dehydrated data that we plan to put on a context is a form of optimistic result)

TkDodo: > In a sense, we could conceptually do something like this?

@Ephem yes, I think so. This is a good idea! From what I can see, the only thing that we need for optimisticResults is this:

https://github.com/tannerlinsley/react-query/blob/9d260e78b45e93ec6b5043de3bb61cac2ddb4345/src/core/queryObserver.ts#L417-L434

One problem might be that on the first render, `currentResult` is still `undefined` because we don't even have a cache entry yet. That's why `getOptimisticResult` primes the cache via `queryCache.build()`(should be safe, even in concurrent mode):

https://github.com/tannerlinsley/react-query/blob/9d260e78b45e93ec6b5043de3bb61cac2ddb4345/src/core/queryObserver.ts#L207-L219

so yeah, we might need to add `createResult()` into the mix if the uSES subscription returns `undefined`, but it still sounds like a better approach.

TkDodo: ah, one thing that comes to mind is renders _not_ being triggered by react-query, but just top-down. For example, some local state might change, which you use inside the `select` function. Since `getOptimisticResult` calls `createResult`, which is also the function that executes select, we would also need to do that in the hypothetical `deriveOptimisticResults` 🤔

najamelan:
React 18 has been released today and is not compatible with react-query according to npm:

```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: some_app@0.1.0
npm ERR! Found: react@18.0.0
npm ERR! node_modules/react
npm ERR!   react@"^18" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^16.8.0 || ^17.0.0" from react-query@3.34.16
npm ERR! node_modules/react-query
npm ERR!   react-query@"=3.34.16" from the root project
```

src/reactjs/useBaseQuery.ts
TkDodo: for now, I would say to keep it as-is, and explore changes further down the road.

PR #3071, opened 2021-12-08
feat(useQueries): v4 api to take an object as input

babycourageous:
Update the `useQueries` API to take an object as input. This way we can define top level options that are true for all queries and a list of queries.

```js
useQueries({ 
  queries: [{ queryKey1, queryFn1, options1 }, { queryKey2, queryFn2, options2 }, ... ] 
})
```

### DIFF
--- a/docs/src/pages/guides/migrating-to-react-query-4.md
+++ b/docs/src/pages/guides/migrating-to-react-query-4.md
@@ -207,6 +207,16 @@ new QueryClient({
 })
 ```
 
+### new API for `useQueries`
+
+The `useQueries` hook now accepts an object with a `queries` prop as its input. The value of the `queries` prop is an array of queries (this array is identical to what was passed into `useQueries` in v3).
+
+```diff
+- useQueries([{ queryKey1, queryFn1, options1 }, { queryKey2, queryFn2, options2 }])
++ useQueries({ queries: [{ queryKey1, queryFn1, options1 }, { queryKey2, queryFn2, options2 }] })
+```
+
+
 ### Removed undocumented methods from the `queryClient`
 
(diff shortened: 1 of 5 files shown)


babycourageous:
Let me know if anything seems amiss and I'll make any necessary adjustments for ya!

etabackman:
I am curious why this is a better params signature than the first param being an array of queries and an optional second param for options which apply to all queries?

That way the simple case where you don't want to provide any options does not require wrapping your param in an object with a `queries` key.

Like this:

```typescript
useQueries(
  [{ queryKey1, queryFn1, options1 }, { queryKey2, queryFn2, options2 }, ... ],
  { ...optionalOptions }
)
```

TkDodo:
@etabackman our idea was to go to a unified structure with just a single object _also_ for `useQuery`:

```
useQuery({ queryKey, queryFn, staleTime, cacheTime, onSuccess })
```

this is currently possible as well because we have overloads, but a unified way of how to do things would be preferred. However, an unexpected [typescript limitation](https://github.com/microsoft/TypeScript/issues/43371#issuecomment-984112995) stopped us from making this the only available syntax.

However, since TypeScript is getting smarter, I'm sure we'll get to this syntax eventually. It might also be that we will make _some_ options unavailable on each query that is usually available (because it has to be the same for all queries), and hoist this option to the top (see e.g. [this PR](https://github.com/tannerlinsley/react-query/pull/2548)). There have also been some [good proposals here](https://github.com/tannerlinsley/react-query/issues/2923#issuecomment-968040760).

The change itself is straight-forward, and guarded by TypeScript as well as the codemod we will likely provide for it, I think upgrading will be smooth.

PR #3081, opened 2021-12-10
fix(useInfiniteQuery): fix infinite re-renders when select produced a new result

### LINKED ISSUE #3068: Infinite re-renders when using `select` in `useInfiniteQuery`
**Describe the bug**
When I transform data from infinite query with `select` option, I get infinite re-renders in my component

**To Reproduce**
Add following component and render it (implied that it'll be wrapped in `QueryClientProvider` of course):
```js
import { useInfiniteQuery } from 'react-query';
import { nanoid } from 'nanoid';

const App = () => {
  const query = useInfiniteQuery(
    'todos',
    ({ pageParam = 0 }) => fetchTodo(pageParam),
    {
      getNextPageParam: ({ nextCursor }) => nextCursor,
      select: ({ pages, pageParams }) => ({
        pages: pages.map(({ todos }) => (
          todos.map(todo => ({
            ...todo,
            localId: nanoid(),
          }))
        )),
        pageParams: [...pageParams],
      })
    },
  );

  console.log('re-render')

  return null;
};

const todos = [];

for (let i = 0; i < 100; i++) {
  todos.push({ id: nanoid(), text: `Todo ${i + 1}` });
}

const TODOS_FETCH_PORTION = 10;

const fetchTodo = (cursor) => {
  const nextCursor = cursor + TODOS_FETCH_PORTION;
  const result = {
    todos: todos.slice(cursor, nextCursor),
  };
  if (todos.length - 1 >= nextCursor) {
    result.nextCursor = nextCursor;
  }
  return result;
};

export default App;
```

**Expected behavior**
Expected `App` to not re-render infinitely.

 - OS: macOS 10.13.6
 - Browser: Chrome 96.04

**Additional info**
With `useQuery` this bug does not occur

TkDodo:
by correctly forwarding notifyOptions

closes #3068

### DIFF
--- a/src/core/infiniteQueryObserver.ts
+++ b/src/core/infiniteQueryObserver.ts
@@ -6,7 +6,11 @@ import type {
   InfiniteQueryObserverResult,
 } from './types'
 import type { QueryClient } from './queryClient'
-import { ObserverFetchOptions, QueryObserver } from './queryObserver'
+import {
+  NotifyOptions,
+  ObserverFetchOptions,
+  QueryObserver,
+} from './queryObserver'
 import {
   hasNextPage,
   hasPreviousPage,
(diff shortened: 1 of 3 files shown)


TkDodo:
@carljohnesan this is the proposed fix and it does fix the infinite loop for me in the test (useCallback is not necessary, just a way to ensure that the selector is only called once)

I've forked your sandbox with the csb preview (with a safeguard because it kept crashing my browser :/) and it now only renders twice as expected: https://codesandbox.io/s/stupefied-shockley-3hk2c?file=/src/App.js

PR #3098, opened 2021-12-15
fix(queryObserver): make sure that memoized select function only runs once per input

### LINKED ISSUE #3095: select runs twice
### Describe the bug

a memoized select function runs twice under some circumstances, even though it should only run once.

circumstances:

- query key stays the same
- but select becomes a new value

### Your Example Website or App

https://codesandbox.io/s/select-runs-twice-bf3pp

### Steps to reproduce

1. click the button
2. observe the console.log - it should only execute select once.

### Expected behavior

select should only be executed once per input

Every time

### Screenshots or Videos

_No response_

### Platform

all

### Additional context

if the query key changes with the input, select is correctly only executed once.

in this sandbox, I've only added `input` to the query key and things work:

https://codesandbox.io/s/select-runs-twice-forked-qrtqm?file=/src/App.tsx

TkDodo:
closes #3095 

we compute the result optimistically, and then one more time when calling setOptions. Both times run the select function because only setOptions will update the previous result and previous options. storing the "previous" selectFn separately on the observer whenever it was invoked (independent of the fact from where it was called) side-steps the issue

### DIFF
--- a/src/core/queryObserver.ts
+++ b/src/core/queryObserver.ts
@@ -69,6 +69,7 @@ export class QueryObserver<
   >
   private previousQueryResult?: QueryObserverResult<TData, TError>
   private previousSelectError: Error | null
+  private previousSelectFn?: (data: TQueryData) => TData
   private staleTimeoutId?: number
   private refetchIntervalId?: number
   private currentRefetchInterval?: number | false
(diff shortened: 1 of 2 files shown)


TkDodo:
@chrislacey89 FYI

chrislacey89:
@TkDodo works like a charm. Good work!

PR #3100, opened 2021-12-16
SWR supports Offline Mutation

zzelcyy:
### Ref
- https://swr.vercel.app/docs/revalidation#revalidate-on-reconnect


TkDodo:
revalidateOnReconnect is not the same as offline mutations. With react-query, if you are offline and fire off a mutation, it will be paused and then continued in the same order once you go back online. 

I think react-query and swr have different concepts of „mutate“? In react-query, a mutation is usually a function that fires a post/put/patch/delete request, thus mutating a remote resource. I don’t think swr has that?

zzelcyy:
> revalidateOnReconnect is not the same as offline mutations. With react-query, if you are offline and fire off a mutation, it will be paused and then continued in the same order once you go back online.
> 
> I think react-query and swr have different concepts of „mutate“? In react-query, a mutation is usually a function that fires a post/put/patch/delete request, thus mutating a remote resource. I don’t think swr has that?

Thanks for the feedback. In SWR, a method for transforming data received through useSWR() is 'mutation'. I think I was confused with the concept in this part. I think this PR can be closed. Have a nice day.

PR #3103, opened 2021-12-16
feat: remove logging in prod mode

TkDodo:


alexandernanberg:
Oh this has indeed confused me before, tbh I'd vote for removing logging altogether. Most cases where I've been confused is in dev

TkDodo:
I think it’s still good practice to log in dev mode to hint users when they are doing something wrong (like using a string key instead if an array)

PR #3122, opened 2021-12-22
fix(types): use core types in useMutation instead of redefining them

DamianOsipiuk:
Using core types, instead of custom ones for `useMutation`.

One question though. Why `variables` and `_defaulted` exist for `MutationObserver` but not for `useMutation`?

@TkDodo

### DIFF
--- a/src/react/types.ts
+++ b/src/react/types.ts
@@ -1,15 +1,12 @@
-import { RetryValue, RetryDelayValue } from '../core/retryer'
 import {
   InfiniteQueryObserverOptions,
   InfiniteQueryObserverResult,
   MutationObserverResult,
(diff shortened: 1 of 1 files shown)


TkDodo:
I know that `_defaulted` is definitely something we set internally and shouldn't be exposed. `variables` is set in a deferred way. When you call `useMutation`, we create the observer, but don't subscribe to it. That only happens once you call `.mutate`, which gets the variables, which we then set on the options of the observer:

https://github.com/tannerlinsley/react-query/blob/c956f1af937f5f5186c7426dee1b07541264646b/src/core/mutationObserver.ts#L117-L121

PR #3128, opened 2021-12-23
make meta optional in QueryFunctionContext

jbguild:
fix(QueryFunctionContext): make meta optional

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -20,7 +20,7 @@ export interface QueryFunctionContext<
   queryKey: EnsuredQueryKey<TQueryKey>
   signal?: AbortSignal
   pageParam?: TPageParam
-  meta: QueryMeta | undefined
+  meta?: QueryMeta | undefined
 }
 
 export type InitialDataFunction<T> = () => T | undefined


TkDodo:
why would meta be optional please? The key on the object will always be there, but the value might be `undefined`. You can see that here where we construct the object:

https://github.com/tannerlinsley/react-query/blob/c989bde9e13a59a58428ea6dde9d958251cc09d5/src/core/query.ts#L397-L401

So I think the types are in line with the implementation.

jbguild:
Perhaps I'm not doing things properly.  If I create a query function

  async function query({
    queryKey,
  }: QueryFunctionContext<MyQueryKeys>): Promise<MyQueryResult> {...}

I get a typescript error if I call this function without a value for meta.   If the query function is intended to be called only from a useQuery or equivalent, that's fine.  I just thought it would be cleaner code for it to be optional.
 One thing to consider is that meta  is optional on the Mutation equivalent (MutationOptions).

Thanks for looking at this!

TkDodo:
> One thing to consider is that meta is optional on the Mutation equivalent (MutationOptions).

`meta` is also optional on the `QueryOptions`:

https://github.com/tannerlinsley/react-query/blob/b84681fdc7d37645cd149f40e98cb18fd8bc393c/src/core/types.ts#L93

There is no equivalent of the `QueryFunctionContext` for mutations, so there is nothing to compare it to. I see the QueryFunctionContext as something that we use internally to invoke the queryFn. I think it would be a mistake to call the function _without_ `meta`, which is what actually happened when we added `meta`. If it's optional, it means that we can just _not_ pass it. But in this case, we _always_ want to pass it - it might just be `undefined` if it hasn't been set. So this type helps us avoids bugs internally.

If you want to manually invoke a function that takes a QueryFunctionContext, you can still pass `meta: undefined` to it.

PR #3141, opened 2021-12-28
Break down `persistQueryClient`

jonathanstanley:
Per discussion #3129 

1. Break down `persistQueryClient` into its various tasks
2. Return an unsubscribe callback when subscribing a persister
3. Update docs

holding on the other updates to keep it more straightforward for now.

### DIFF
--- a/docs/src/pages/plugins/persistQueryClient.md
+++ b/docs/src/pages/plugins/persistQueryClient.md
@@ -89,7 +89,7 @@ persistQueryClientRestore({
   persister,
   maxAge = 1000 * 60 * 60 * 24, // 24 hours
   buster = '',
-  hydrateOptions,
+  hydrateOptions = undefined,
 })
 ```
 
@@ -102,7 +102,7 @@ persistQueryClientSave({
   queryClient,
   persister,
   buster = '',
-  dehydrateOptions,
+  dehydrateOptions = undefined,
 })
 ```
 
@@ -118,7 +118,7 @@ persistQueryClientSubscribe({
   queryClient,
   persister,
   buster = '',
-  dehydrateOptions,
+  dehydrateOptions = undefined,
 })
 ```
 
@@ -132,8 +132,8 @@ persistQueryClient({
   persister,
   maxAge = 1000 * 60 * 60 * 24, // 24 hours
   buster = '',
-  hydrateOptions,
-  dehydrateOptions,
+  hydrateOptions = undefined,
+  dehydrateOptions = undefined,
 })
 ```
 
--- a/docs/src/pages/plugins/persistQueryClient.md
+++ b/docs/src/pages/plugins/persistQueryClient.md
@@ -30,7 +30,9 @@ const queryClient = new QueryClient({
   },
 })
 
-const localStoragePersister = createWebStoragePersister({storage: window.localStorage})
+const localStoragePersister = createWebStoragePersister({
+  storage: window.localStorage,
+})
 
 persistQueryClient({
   queryClient,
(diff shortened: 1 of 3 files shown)


src/persistQueryClient/index.ts
```diff
@@ -84,8 +90,53 @@ export async function persistQueryClient({
...
+ * (Retained for backwards compatibility)
+ */
+export async function persistQueryClient(props: PersistQueryClientOptions) {
+  if (typeof window !== 'undefined') {
+    // Attempt restore
+    persistQueryClientRestore(props)
 
     // Subscribe to changes in the query cache to trigger the save
-    queryClient.getQueryCache().subscribe(saveClient)
+    persistQueryClientSubscribe(props)
```
TkDodo, line 140: should we not `return persistQueryClientSubscribe(props)` here in order to fix the issue of being unable to unsubscribe?

docs/src/pages/plugins/persistQueryClient.md
```diff
@@ -48,31 +50,91 @@ You can also pass it `Infinity` to disable garbage collection behavior entirely.
...
+
+### Removal
+
+- If data is found to be expired (see `maxAge`), busted (see `buster`), error (ex: `throws ...`), or empty (ex: `undefined`), the persister `removeClient()` is called and the cache is immediately discarded.
+
 ## API
 
+### `persistQueryClientRestore`
+
+This will attempt to restore a persister's stored cached to the active query cache.
```
TkDodo, line 84: hmm, why the "active query cache" ? I think something like: "to the query cache of the passed queryClient" would be better fitting?

src/persistQueryClient/index.ts
jonathanstanley: I'm happy either way.

if a user wants to control the subscribe/unsubscribe, my assumption was that they would call the new `persistQueryClientSubscribe` directly.  The way I submitted, `persistQueryClient` works **exactly** as before.  But I perhaps some would find it nice to also add the feature here.  Let me know if you want me to add this to the PR.

docs/src/pages/plugins/persistQueryClient.md
jonathanstanley: Yes, I like your wording 👍

src/persistQueryClient/index.ts
TkDodo: yes lets please add / fix that. We're towards a new major version anyways :)

docs/src/pages/plugins/persistQueryClient.md
```diff
@@ -48,31 +50,91 @@ You can also pass it `Infinity` to disable garbage collection behavior entirely.
...
+
+```ts
+persistQueryClientRestore({
+  queryClient,
+  persister,
+  maxAge = 1000 * 60 * 60 * 24, // 24 hours
+  buster = '',
+  hydrateOptions = undefined,
```
TkDodo, line 92: just a question: what's the reason of defaulting to `undefined` ? If `hydrateOptions` isn't passed, it will be `undefined` anyhow 🤔

jonathanstanley: When I was reading the docs before, the defaults and required parameters were a little difficult for me to follow.  My hope here was to indicate that the `maxAge`, `buster`, `hydrateOptions` etc. have defaults and aren't required.

src/persistQueryClient/index.ts
jonathanstanley: done: https://github.com/tannerlinsley/react-query/pull/3141/commits/c4651b1851745f9a03f92deb392d1c2778de33f6

docs/src/pages/plugins/persistQueryClient.md
jonathanstanley: done: https://github.com/tannerlinsley/react-query/pull/3141/commits/a1584d3a4dce9924acdf825aa79871fa0e6cddfb

src/persistQueryClient/index.ts
```diff
@@ -84,8 +90,53 @@ export async function persistQueryClient({
...
+}
+
+/**
+ * Restores persisted data to QueryCache and persists further changes.
+ * (Retained for backwards compatibility)
+ */
+export async function persistQueryClient(props: PersistQueryClientOptions) {
+  if (typeof window !== 'undefined') {
+    // Attempt restore
+    persistQueryClientRestore(props)
```
TkDodo, line 137: I think we need to `await persistQueryClientRestore(props)` before we trigger the subscription, do we not?

jonathanstanley: My reading of the existing version is that it does not await:  

https://github.com/tannerlinsley/react-query/blob/d7faf86005052654451dd7d5b3665bf6c971b66e/src/persistQueryClient-experimental/index.ts#L63-L89

However, I can add it if you like?

TkDodo: > the try is synchronous?

I don't think it is. The whole function is an async function, and in line 65 we `await persister.restoreClient() `. If we move that to a separate function, we have to await it to behave the same as before.

jonathanstanley: https://github.com/tannerlinsley/react-query/pull/3141/commits/df7686c11dfc58bc8989b03bb3f9565040bdabd2

TkDodo (approved):

PR #3143, opened 2021-12-29
Pr/make useIsMounted reused

DidrikLind:
I noticed that the "useIsMounted" hook is basically used on quite some places in the code implicitly (mountedRef defined in several of hooks) and saw the opportunity to refactor out useIsMounted from src/devtools -> src/react in order to be in a more general util place and therefore it would make sense to reuse the hook everywhere in the project.

I think it makes the code more readable by using this hook rather than introducing a new useRef in every component in order to know if the component is mounted or not:
```
// OLD
...
const mountedRef = React.useRef(false)
...

useEffect(() => {
  ...
  mountedRef.current = true
  ...
  return () => {
    mountedRef.current = false
  }
}, [])
```

```
// "NEW"
...
const isMounted = useIsMounted() // use as isMounted()
...
```

**Some thinking out loud stuff:**
1. The actual hook
I have not introduced the hook myself as can be seen in the code diffing tool, but I think I moved it to a better place.
Though there are some naming conflicts with the "utils.tsx" file under the same directory. Maybe one should rename "utils.tsx" to "testUtils.tsx"? Maybe make an own file for this hook, e.g "useIsMounted.ts".
I still need to suffix the testing file with "tsx" in order to use jsx in there it seems, but does it make sense to have the following combination of files if so: useIsMounted.ts / useIsMounted.test.tsx (or maybe just useIsMounted.tsx / useIsMounted.test.tsx even though there is no jsx in the hook, but for the sake of consistent suffixes of the files)? 

2. About the tests
I haven't done much opensource before and am a bit new with the "@testing-library/react" testing library, but I think it makes sense to have some basic testing for this hook in the longer run. I peaked at the "@testing-library/react-hooks" library a bit but I dont think I need it for this case, and also did not want to add another depedency for the project.

I have a bunch more questions, but I thought maybe it could help me to throw this pr up and get some feedback on it.

Cheers!

### DIFF
--- a/src/devtools/devtools.tsx
+++ b/src/devtools/devtools.tsx
@@ -3,7 +3,7 @@ import React from 'react'
 import { Query, useQueryClient } from 'react-query'
 import { matchSorter } from 'match-sorter'
 import useLocalStorage from './useLocalStorage'
-import { useIsMounted, useSafeState } from './utils'
+import { useSafeState } from './utils'
 
 import {
   Panel,
@@ -20,6 +20,7 @@ import { getQueryStatusLabel, getQueryStatusColor } from './utils'
 import Explorer from './Explorer'
 import Logo from './Logo'
 import { noop } from '../core/utils'
+import { useIsMounted } from '../react/utils'
 
 interface DevtoolsOptions {
   /**
--- a/src/devtools/utils.ts
+++ b/src/devtools/utils.ts
@@ -3,8 +3,7 @@ import { Query } from '../core'
 
 import { Theme, useTheme } from './theme'
 import useMediaQuery from './useMediaQuery'
-
-export const isServer = typeof window === 'undefined'
+import { useIsMounted } from '../react/utils'
 
 type StyledComponent<T> = T extends 'button'
   ? React.DetailedHTMLProps<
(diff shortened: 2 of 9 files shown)


TkDodo:
Please have a look at the `alpha` branch. For v4, we've mostly removed all occurrences of `useIsMounted`, because the actual warning caused by calling setState on an unmounted component will be removed in react18. I will likely not merge this PR because of the conflicts it will create with v4 / the alpha branch.

I think we still have it in the devtools, but only there.

DidrikLind:
@TkDodo alrighty. 

I guess you are referring to this change in react 18: [removage of warning react 18](https://github.com/reactwg/react-18/discussions/82).
Sorry missed that totally.

Though I cannot see in the `alpha` branch that anything is changed regarding the usage of the mountedRef scenario, see diff [master diff alpha](https://github.com/tannerlinsley/react-query/compare/alpha)
Am I missing something obvious?



I have closed the PR for now. Thnx for the quick answer!

TkDodo:
@DidrikLind I'e removed the mounted things in this draft PR:
- https://github.com/tannerlinsley/react-query/pull/3064

where we upgrade to react18. This one will very likely make it for v4.

DidrikLind:
@TkDodo roger that. Thnx for the guiding answers! :love_you_gesture:

PR #3175, opened 2022-01-10
fix(examples/load-more-infinite-scroll): fix wrong condtion

maricabertarini:
Fixes [issue](https://github.com/tannerlinsley/react-query/issues/3171).

Scope: **[Load More / Infinite Scroll example]**
The top button contained wrong text due to wrong variables used in the corresponding condition that determines its text. 
This PR fixes the use of the wrong variables.


maricabertarini:
I do not have rights to edit the _Linked issue_ field, so I pasted the corresponding issue in the PR description.

PR #3196, opened 2022-01-14
fix(examples): react-native: update .gitignore to exclude lib folder

arnaudbzn:
This fix the missing `src/lib` folder.

### DIFF
--- a/examples/react-native/src/lib/api.ts
+++ b/examples/react-native/src/lib/api.ts
@@ -18,7 +18,7 @@ function delay(t: number, v: () => void) {
   });
 }
 
-export function fetchMovies(): Promise<Movie[]> {
+export async function fetchMovies() {
   console.log('fetchMovies');
   return delay(200 + Math.floor(Math.random() * 2000), () =>
     movies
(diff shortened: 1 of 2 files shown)


examples/react-native/src/lib/api.ts
```diff
@@ -0,0 +1,39 @@
...
+export function fetchMovies(): Promise<Movie[]> {
+  console.log('fetchMovies');
+  return delay(200 + Math.floor(Math.random() * 2000), () =>
+    movies
+      .slice(0, 100)
+      .map((movie) => ({ title: movie.title, year: movie.year }))
+  ) as Promise<Movie[]>;
+}
+
+export function fetchMovie(title: string): Promise<MovieDetails> {
```
TkDodo, line 30: lets make these `async` functions, which should get rid of the type assertion to a Promise because async functions always return a Promise.

TkDodo: I meant removing `as Promise<Movie[]>;` - or why do we need that?

arnaudbzn: We need that because movies.json is not the same type as Movie[]. TS seems not capable to infer this.

TkDodo: oh okay, I didn't see that

TkDodo (approved):

PR #3209, opened 2022-01-18
docs(ssr): Hydration caveat for Next.js rewrites

bennettdams:
This adds a small section for a caveat about hydration when using some specific feature/configuration of Next.js.

See initial RQ discussion: https://github.com/tannerlinsley/react-query/discussions/3192

See Next.js issue: https://github.com/vercel/next.js/issues/33028#issuecomment-1012126770

---

I'm not sure if we should additionally reference the Next.js issue for this topic in the RQ docs, but my hopes are not high that the issue will be fixed "this time", as it is not really a bug per se, so I didn't here.


bennettdams:
Maybe @Ephem has some more insights on this topic, so we could give guidance on how to tackle this in the docs?

PR #3223, opened 2022-01-23
fix(core): do not refetch disabled queries

### LINKED ISSUE #3202: Disabled queries (enabled: false) get affected by invalidateQueries (refetchInactive: true)
### Describe the bug

`
queryClient.invalidateQueries(queryKey, {
        refetchInactive: true,
      })
`
also refetches disabled queries (`enabled`: false)

but  - https://github.com/tannerlinsley/react-query/issues/947#issuecomment-1005020047

looks like this issue was introduce in this commit  - https://github.com/tannerlinsley/react-query/pull/949/commits/f5ae636e2ba697d56e1789d731e5d382f7b444bd

https://codesandbox.io/s/quiet-dust-hegny?file=/src/App.js

### Steps to reproduce

1. Create a query with `useQuery` and set `enabled` option to false.
2. Invalidate this query with `invalidateQueries` and set `refetchInactive` to true

### Expected behavior

I expected that disabled queries (`enabled: false`) would not refetch when I use `invalidateQueries`

### Platform

- OS: macOs
- Browser: Not related

### react-query version

v3.34.8

### TypeScript version

v4.2.2

TkDodo:
with refetchQueries or invalidateQueries + refetchType "inactive"

disabled queries (=queries that have observers which are all enabled:false) are matched as "inactive"; this is okay when searching for them via findAll or for removeQueries, but the docs clearly state that refetchQueries / invalidateQueries do not refetch disabled queries, and that the only way to refetch them is via refetch returned from useQuery; this is important when using enabled to signal that some dependencies are not yet ready

some tests needed to be adapted because we used disabled observer + refetchQueries a lot. The easiest way to emulate the observers we wanted here was mostly with initialData + staleTime, and to get a real inactive query, we just need to subscribe + unsubscribe immediately

closes #3202

### DIFF
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -250,6 +250,10 @@ export class Query<
     return this.observers.some(observer => observer.options.enabled !== false)
   }
 
+  isDisabled(): boolean {
+    return this.getObserversCount() > 0 && !this.isActive()
+  }
+
   isStale(): boolean {
     return (
       this.state.isInvalidated ||
--- a/src/core/queryClient.ts
+++ b/src/core/queryClient.ts
@@ -283,13 +283,16 @@ export class QueryClient {
     const [filters, options] = parseFilterArgs(arg1, arg2, arg3)
 
     const promises = notifyManager.batch(() =>
-      this.queryCache.findAll(filters).map(query =>
-        query.fetch(undefined, {
-          ...options,
-          cancelRefetch: options?.cancelRefetch ?? true,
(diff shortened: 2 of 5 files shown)


tannerlinsley (approved):
Wonderful!

TkDodo:
`enabled` is an observer level property that only works on `useQuery`. For imperative methods, you are expected to _not_ invoke them if you don’t want to invoke them.

PR #3244, opened 2022-01-29
docs(setQueriesData): Make setQueriesData more searchable

villesau:
- Makes setQueriesData more searchable by adding keyword `partial` there
- Change mention of fuzzily matching to partial matching since fuzzy matching is more arbitrary algorithm
- Mention setQueriesData in setQueryData to avoid confusion of behaviours

The purpose of this change is to make setQueriesData a bit more searchable from the docs to avoid confusion like this: https://github.com/tannerlinsley/react-query/discussions/3239

### DIFF
--- a/docs/src/pages/reference/QueryClient.md
+++ b/docs/src/pages/reference/QueryClient.md
@@ -203,7 +203,7 @@ This distinction is more a "convenience" for ts devs that know which structure w
 
 ## `queryClient.setQueryData`
 
-`setQueryData` is a synchronous function that can be used to immediately update a query's cached data. If the query does not exist, it will be created. **If the query is not utilized by a query hook in the default `cacheTime` of 5 minutes, the query will be garbage collected**.
+`setQueryData` is a synchronous function that can be used to immediately update a query's cached data. If the query does not exist, it will be created. **If the query is not utilized by a query hook in the default `cacheTime` of 5 minutes, the query will be garbage collected**. To do partial mathing of query keys, you need to use [`queryClient.setQueriesData`](#queryclientsetqueriesdata) instead. 
 
 After successful changing query's cached data via `setQueryData`, it will also trigger `onSuccess` callback from that query.
 
@@ -250,7 +250,7 @@ console.log(state.dataUpdatedAt)
 
 ## `queryClient.setQueriesData`
 
-`setQueriesData` is a synchronous function that can be used to immediately update cached data of multiple queries. Only queries that match the passed queryKey or queryFilter will be updated - no new cache entries will be created. Under the hood, [`setQueryData`](#queryclientsetquerydata) is called for each query.
+`setQueriesData` is a synchronous function that can be used to immediately update cached data of multiple queries by partially matching the query key. Only queries that match the passed queryKey or queryFilter will be updated - no new cache entries will be created. Under the hood, [`setQueryData`](#queryclientsetquerydata) is called for each query.
 
 ```js
 queryClient.setQueriesData(queryKey | filters, updater)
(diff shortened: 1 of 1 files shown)


docs/src/pages/reference/QueryClient.md
```diff
@@ -250,7 +250,7 @@ console.log(state.dataUpdatedAt)
 
 ## `queryClient.setQueriesData`
 
-`setQueriesData` is a synchronous function that can be used to immediately update cached data of multiple queries. Only queries that match the passed queryKey or queryFilter will be updated - no new cache entries will be created. Under the hood, [`setQueryData`](#queryclientsetquerydata) is called for each query.
+`setQueriesData` is a synchronous function that can be used to immediately update cached data of multiple queries by partially matching the query key. Only queries that match the passed queryKey or queryFilter will be updated - no new cache entries will be created. Under the hood, [`setQueryData`](#queryclientsetquerydata) is called for each query.
```
TkDodo, line 253: `setQueriesData` per se doesn't always partially match the query key. it takes a `filters` argument, and if the filter matches, the data will be set. You can use a filter that doesn't contain a query key. This is the general filters concept of react-query that applies to all functions that have a plural in their name, like `refetchQueries` or `removeQueries`. see: https://react-query.tanstack.com/guides/filters#_top

docs/src/pages/reference/QueryClient.md
```diff
@@ -203,7 +203,7 @@ This distinction is more a "convenience" for ts devs that know which structure w
 
 ## `queryClient.setQueryData`
 
-`setQueryData` is a synchronous function that can be used to immediately update a query's cached data. If the query does not exist, it will be created. **If the query is not utilized by a query hook in the default `cacheTime` of 5 minutes, the query will be garbage collected**.
+`setQueryData` is a synchronous function that can be used to immediately update a query's cached data. If the query does not exist, it will be created. **If the query is not utilized by a query hook in the default `cacheTime` of 5 minutes, the query will be garbage collected**. To do partial mathing of query keys, you need to use [`queryClient.setQueriesData`](#queryclientsetqueriesdata) instead.
```
TkDodo, line 206: I find this more confusing than helpful. Most of the time, you will only update _one_ query. The reference to `partial matching` is also weird. What we could do is:

```
To update multiple queries at once, you can use setQueriesData
```

villesau: @TkDodo I would not have understood the usage of `setQueriesData` from that sentence. I spent a lot of time figure out how to partially match keys and was giving up on evaluating react-query since such feature did not seem possible.

villesau: the reason for using "partial" is for searchability. I could not find a way to do what I needed from the documentation since the word was not mentioned anywhere. I only found outdated usages from previous versions which made me think that the feature is completely deprecated.

docs/src/pages/reference/QueryClient.md
villesau: the reason for highlighting partial matching is searchability. This part of documentation does not appear by any word that indicates there is such possibility. All the examples uses `setQueryData` and outdated examples indicates that it would support partial matching. what about `by using filter function or partially matching the query key`?

```suggestion
`setQueriesData` is a synchronous function that can be used to immediately update cached data of multiple queries by using filter function or partially matching the query key. Only queries that match the passed queryKey or queryFilter will be updated - no new cache entries will be created. Under the hood, [`setQueryData`](#queryclientsetquerydata) is called for each query.
```

docs/src/pages/reference/QueryClient.md
villesau: What about: `To update multiple queries at once and partially matching of query keys, you need to use [`queryClient.setQueriesData`](#queryclientsetqueriesdata) instead.` ?

```suggestion
`setQueryData` is a synchronous function that can be used to immediately update a query's cached data. If the query does not exist, it will be created. **If the query is not utilized by a query hook in the default `cacheTime` of 5 minutes, the query will be garbage collected**. To update multiple queries at once and partially matching of query keys, you need to use [`queryClient.setQueriesData`](#queryclientsetqueriesdata) instead. 
```

PR #3246, opened 2022-01-29
feat(*): add optional logger prop to QueryClient config

anilanar:
See the commit message for full(ish) list of changes. 

Addresses https://github.com/tannerlinsley/react-query/discussions/3240.

I'll update docs if we are happy with these changes.


src/reactjs/logger.native.ts
```diff
@@ -1,7 +0,0 @@
-import type { Logger } from '../core/logger'
```
anilanar, line 1: See `getDefaultLogger` in `../core/logger` for what would happen in react-native.

Basically, I went for this version of the logger instead of plain` console` as the default logger just to support both cases. 

If you think it's really necessary to have different loggers for normal/react-native users to save ~50 bytes; sure, I'll do it.

src/reactjs/logger.native.ts
```diff
@@ -1,7 +0,0 @@
...
-
-export const logger: Logger = {
-  log: console.log,
-  warn: console.warn,
-  error: console.warn,
```
anilanar, line 6: Was this a mistake or was it on purpose to sure `console.warn` for `logger.error` for react-native?

anilanar: Ok, I digged out https://github.com/tannerlinsley/react-query/issues/795 so it's undesirable to use `console.error` for react-native users and that was the primary reason for `logger.native.ts` I assume. 

I'll change it.

anilanar: Or should we ask react-native users to pass a custom logger to `QueryClient` instead?

src/core/logger.ts
```diff
@@ -8,14 +8,25 @@ export interface Logger {
...
-  return logger
-}
-
-export function setLogger(newLogger: Logger) {
-  logger = newLogger
+  return {
+    log: console.log,
+    warn: console.warn,
+    error: console.error,
+  }
```
anilanar, line 31:
```suggestion
  return console
```

I return this object instead of just `console` because I mimicked `logger.native.ts` assuming there was a problem with react-native builds if not done so, but now I understand why there was a `logger.native.ts` so I can probably change this.

jest.config.js
```diff
@@ -7,4 +7,5 @@ module.exports = {
   moduleNameMapper: {
     'react-query': '<rootDir>/src/index.ts',
   },
+  clearMocks: true,
```
anilanar, line 10: This option resets all calls for jest mock functions after every test, which is false by default.

It's important to set this to true, because otherwise `expect(x).toHaveBeenCalledTimes(1)` can return false positives because `x` was called in a previous test.

src/core/query.ts
```diff
@@ -365,11 +368,8 @@ export class Query<
       }
     }
 
-    if (
-      process.env.NODE_ENV !== 'production' &&
-      !Array.isArray(this.options.queryKey)
-    ) {
-      getLogger().error(
+    if (!Array.isArray(this.options.queryKey)) {
+      this.logger.error(
```
TkDodo, line 372: I understand why you made that change, since the default logger is now stubbed for production. However, if you pass in your own logger, you will now get these logs also in production unless you make the env check yourselves.

src/reactjs/logger.native.ts
TkDodo: good question. @arnaudbzn ?

src/reactjs/tests/utils.tsx
```diff
@@ -24,10 +33,10 @@ export function mockNavigatorOnLine(value: boolean) {
...
 }
 
-export function mockConsoleError() {
-  const consoleMock = jest.spyOn(console, 'error')
-  consoleMock.mockImplementation(() => undefined)
-  return consoleMock
+export const mockLogger = {
+  log: jest.fn(),
+  warn: jest.fn(),
+  error: jest.fn(),
```
TkDodo, line 39: why do we need he mockLogger if the defaultLogger defaults to `jest.fn()` in test env?

TkDodo (commented):
docs would need to be updated please, as well as the migration guide

TkDodo:
I think we can remove `setLogger` from the side-effects, right?

https://github.com/tannerlinsley/react-query/blob/5e2f2c2b644c03433c15e3ba61f884e142098fc7/package.json#L17-L26

src/core/query.ts
anilanar: Which is the point of custom loggers, requiring finer control on logging and logging levels in different environments. For example, @userlike, we set log levels to "error" and catch console errors with sentry that happens in production.

TkDodo:
The size of `react-query.production.min.js` seems to go up from `11.74KB` on the latest `alpha` release to `11.88KB` on this PR.
This seems significant for such a tiny change 🤔

src/reactjs/tests/utils.tsx
anilanar: Default logger (i.e. the return value of `getDefaultLogger`) doesn't default to mock logger/jest.fn because it's tricky to dead code eliminite `jest.fn` calls in non-test environments. So the reason default logger doesn't use jest.fn() in test environments is because we don't want to refer to the global `jest` variable in files that go into production development/builds.

TkDodo: but, it does, for `process.env.NODE_ENV === 'test'`, or am I missing something?

https://github.com/tannerlinsley/react-query/pull/3246/files#diff-b26e2d2c63ca623f55f84e869194f983f9dbb81fd0703cf38e5f11edc091f5c3R11-R18

anilanar:
> The size of react-query.production.min.js seems to go up from 11.74KB on the latest alpha release to 11.88KB on this PR.

I think `console` vs `{ log: console.log, warn: console.warn. error: console.error }` by itself adds ~ 50bytes by itself, so once I revert that, it should be a less significant number, but I can probably reduce it further, I'll check.

src/reactjs/tests/utils.tsx
anilanar: I tried that approach and gave up for the reasons I give above. It seems I forgot to remove that code, my bad 😸

TkDodo:
With the old approach, log statements were removing completely in production due to the env check. Now they stay in the bundle and do nothing per default, and it will increase with every log statement we make. The string itself also can't be minified. 
The key warning for example is similar to the react key warning that only shows up in dev mode. Not sure why we'd ever want that in production.

src/reactjs/logger.native.ts
arnaudbzn: The current default logger behavior for React Native is fine. 
We should not require React Native developers to define a custom logger.

TkDodo:
any updates here please, especially in regards to my last statement? thank you

anilanar:
@TkDodo I'll check today and get back to you.

anilanar:
@TkDodo I wrapped all log usages with production check. I'll add docs and this should be good to go.

The increase in bundle size comes from `logger` class fields and `getLogger` getter, which cannot be shortened to a single character by the uglifier unlike `getLogger` function. 

If you prefer bundle size over code readability, you can rename all private class fields to single character names and that can save some bytes.

TkDodo:
> @TkDodo I wrapped all log usages with production check. I'll add docs and this should be good to go.

Thanks. Have you removed the `sideEffects` form `package.json` ?

> The increase in bundle size comes from logger class fields and getLogger getter, which cannot be shortened to a single character by the uglifier unlike getLogger function.

> If you prefer bundle size over code readability, you can rename all private class fields to single character names and that can save some bytes.

Thanks, I do not prefer that, but I think there are ways to let terser know to mangle private fields, but it's out of scope for here.

The gain is also minimal now since you updated with alpha:

```
alpha: 11.64KB
here:  11.68KB
```

so it's all good 👍

anilanar:
@TkDodo 

 I updated the docs :tada: 
 
 I wanted to document the `Logger` interface but noticed that there's no documentation convention for pure types/interfaces so I created a section in `Guides` instead.

TkDodo:
@anilanar the side-effects still need to be removed from package.json, right?

https://github.com/tannerlinsley/react-query/pull/3246#issuecomment-1025151854

TkDodo (approved):

TkDodo:
thank you for your effort in contributing this 🙇

TkDodo:
Now that I've merged it, I just had an idea for a follow up regarding the tests 😅 

I think it would be better understandable if we don't rely on the global `mockLogger`. We do have `createQueryClient()` anyways, so how about:

- `createQueryClient()` in the tests always creates a _new_ `mockLogger` for encapsulation
- in the tests where we need it, we just call `queryClient.getLogger()` because we have a queryClient anyways.

what do you think?

anilanar:
@TkDodo That's what I did at first, and then noticed that every test needs to create a mock logger in the exact same way. 

Because jest mocks are isolated between tests, then I thought, there would be 0 harm for having a global mock logger to reduce some boilerplate from unit tests.

TkDodo:
okay yeah that's good enough. I just noticed something else: we get error logs in the tests now, see here:

https://github.com/tannerlinsley/react-query/runs/5344881152?check_suite_focus=true

```
PASS src/reactjs/tests/useQuery.test.tsx (13.6[42](https://github.com/tannerlinsley/react-query/runs/5344881152?check_suite_focus=true#step:5:42) s)
  ● Console

    console.error
      Error: Uncaught 'Error test jaylen'
          at reportException (/home/runner/work/react-query/react-query/node_modules/jsdom/lib/jsdom/living/helpers/runtime-script-errors.js:62:24)
          at innerInvokeEventListeners (/home/runner/work/react-query/react-query/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:346:9)
          at invokeEventListeners (/home/runner/work/react-query/react-query/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:281:3)
          at HTMLUnknownElementImpl._dispatch (/home/runner/work/react-query/react-query/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:228:9)
          at HTMLUnknownElementImpl.dispatchEvent (/home/runner/work/react-query/react-query/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:101:17)
          at HTMLUnknownElement.dispatchEvent (/home/runner/work/react-query/react-query/node_modules/jsdom/lib/jsdom/living/generated/EventTarget.js:207:34)
          at Object.invokeGuardedCallbackDev (/home/runner/work/react-query/react-query/node_modules/react-dom/cjs/react-dom.development.js:237:16)
          at invokeGuardedCallback (/home/runner/work/react-query/react-query/node_modules/react-dom/cjs/react-dom.development.js:292:31)
          at beginWork$1 (/home/runner/work/react-query/react-query/node_modules/react-dom/cjs/react-dom.development.js:23203:7)
          at performUnitOfWork (/home/runner/work/react-query/react-query/node_modules/react-dom/cjs/react-dom.development.js:22157:12) Error test jaylen

      at VirtualConsole.<anonymous> (node_modules/jsdom/lib/jsdom/virtual-console.js:29:45)
      at VirtualConsole.emit (node:events:520:28)
      at reportException (node_modules/jsdom/lib/jsdom/living/helpers/runtime-script-errors.js:66:28)
      at innerInvokeEventListeners (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:346:9)
      at invokeEventListeners (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:281:3)
      at HTMLUnknownElementImpl._dispatch (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:228:9)
      at HTMLUnknownElementImpl.dispatchEvent (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:101:17)
      at HTMLUnknownElement.dispatchEvent (node_modules/jsdom/lib/jsdom/living/generated/EventTarget.js:207:34)

    console.error
      The above error occurred in the <Page> component:
          in Page
          in ErrorBoundary
          in QueryClientProvider
      
      React will try to recreate this component tree from scratch using the error boundary you provided, ErrorBoundary.

      at logCapturedError (node_modules/react-dom/cjs/react-dom.development.js:19527:21)
      at logError (node_modules/react-dom/cjs/react-dom.development.js:19564:5)
      at ErrorBoundary.callback (node_modules/react-dom/cjs/react-dom.development.js:207[44](https://github.com/tannerlinsley/react-query/runs/5344881152?check_suite_focus=true#step:5:44):9)
      at callCallback (node_modules/react-dom/cjs/react-dom.development.js:12[49](https://github.com/tannerlinsley/react-query/runs/5344881152?check_suite_focus=true#step:5:49)0:12)
      at commitUpdateQueue (node_modules/react-dom/cjs/react-dom.development.js:12[51](https://github.com/tannerlinsley/react-query/runs/5344881152?check_suite_focus=true#step:5:51)1:9)
      at commitLifeCycles (node_modules/react-dom/cjs/react-dom.development.js:198[58](https://github.com/tannerlinsley/react-query/runs/5344881152?check_suite_focus=true#step:5:58):11)
      at commitLayoutEffects (node_modules/react-dom/cjs/react-dom.development.js:22[80](https://github.com/tannerlinsley/react-query/runs/5344881152?check_suite_focus=true#step:5:80)3:7)
      at HTMLUnknownElement.callCallback (node_modules/react-dom/cjs/react-dom.development.js:188:14)
```

tests are passing but this wasn't here before and is quite distracting when running tests. do you know how to fix that?

TkDodo:
this seems to be for all tests with `ErrorBoundaries`. I've found an issue and apparently the only way is to spy on the console?

https://github.com/facebook/react/issues/11098#issuecomment-523977830

How did that work before?

TkDodo:
I've pushed a quickfix to alpha, but if you know a proper fix, I would appreciate another PR. thanks.

https://github.com/tannerlinsley/react-query/commit/4253ff3f715bea277eccfebee5daf62320fbe8db

anilanar:
@TkDodo Sadly that's still an open issue and React team didn't fix it nor they promised it yet: https://github.com/facebook/react/issues/15069 (you can see my face couple of times in that thread :smile:).

anilanar:
@TkDodo At Userlike, we use a custom logger that fails the test when an error is logged and use a different custom logger when `log.error`s are expected.

We were forced to not use `console` as a logger at some point because it was impossible to prevent some libraries log errors in some cases.

PR #3248, opened 2022-01-30
feat(persistQueryClient): PersistQueryClientProvider

TkDodo:


src/persistQueryClient/index.ts
```diff
@@ -1,142 +1,2 @@
...
-  if (typeof window !== 'undefined') {
-    // Attempt restore
-    await persistQueryClientRestore(props)
-
-    // Subscribe to changes in the query cache to trigger the save
-    return persistQueryClientSubscribe(props)
-  }
-}
+export * from './persist'
+export * from './PersistQueryClientProvider'
```
TkDodo, line 2: I am not sure where to put the `PersistQueryClientProvider`. It is both react specific as well as specific for the plugin. If I put it to `react`, it will likely wind up in the main bundle (even though treeshakeable). If we keep it here, the `persistQueryClient` plugin is dependent on React...

src/persistQueryClient/PersistQueryClientProvider.tsx
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    void run()
+
+    return unsubscribe
+  }, [client])
+
+  if (!initialized) {
+    return <>{loading}</>
+  }
```
TkDodo, line 39: I guess this wouldn't work with SSR? I mean it would work, but you would only get `loading` rendered on the server. Is this even a use-case: hydrating from the server and restoring from local storage? I fear that if we make an exception for SSR here, we will get hydration mismatches. Also rendering on the server and then overriding that with state from localStorage seems weird... what do you think?

@Ephem fyi

Ephem: Hmm, I haven't thought about the persisting use case much, but yeah, does seem a little weird. I guess one use case could be if you have a very long cache time in a CDN or something (or static generation) where the data in the server rendered html can be pretty old. If the user has visited the site before and persisted fresher data, I _guess_ it could be nice if you first hydrated the server rendered html and as soon as the app has hydrated, you update the data to the fresher version that was previously persisted.

Feel like a lot of hoops to jump through though, if data freshness matters you probably wanna revalidate by refetching anyway.

TkDodo: The problem is that you can't really use this approach together with SSR, as its an async rehydrate. What happens if you use the two together, but don't have something in localstorage?

- SSR will render `<>{loading}</>`
- on the client, we'll render `<>{loading}</>`, followed by a read from localstorage
- we don't read anything, so we just client render the whole page

That basically means if you use this component, you can't SSR anything on the page, which seems bad :/

redux-persist has uses the same approach with the [PersistGate](https://github.com/rt2zz/redux-persist/blob/d8b01a085e3679db43503a3858e8d4759d6f22fa/src/integration/react.ts#L16) component, but they also have a ton of open issues regarding SSR, e.g.:

- https://github.com/rt2zz/redux-persist/issues/1008
- https://github.com/vercel/next.js/issues/8240#issuecomment-647699316

all the provided solutions are basically: don't use `PersistGate`, or use the children render function, which basically also doesn't render the Gate, which means that the restore actions might be dispatched _after_ the app has started rendering, which is the exact same issue I was trying to solve.

if we disable the Gate on the server only, as [suggested on stackoverflow](https://stackoverflow.com/questions/58238525/gatsby-not-generating-correct-static-html-files/62501513#62501513), we will get hydration mismatches, because the server renderes the whole markup while the client only renders the loading gate.

I don't see a solution right now. Maybe we should just document this properly, that this doesn't work with SSR and that people can still subscribe to / restore from their storage whenever they want manually with the provided functions?

Ephem: Ahh, yes, now that I actually read the code (😬 ) I see that this components job is specifically to delay rendering until this is done, which is as you say incompatible with SSR at a fundamental level. 😄 

If there exists a low level API to restore a persist manually, that should be enough for any wonky SSR usecases people might have.

TkDodo: @Ephem please have a look here: 5c4de915

As briefly discussed on Discord, I added another `IsHydrating` context and I've deferred observer subscriptions until we have finished hydrating.

Further, if we are hydrating, I've put the query into the new `fetchState: paused`. We could also provide a `pausedReason` or something similar, but I think at least it gives the users the opportunity to _not_ show a loading spinner, because the `status` of the query will be `loading` in the meantime. However, no actual request is being made. We could also put it to `fetchStatus: 'idle'` if we want.

Now currently, this only applies to the new `PersistQueryClientProvider`, but maybe the new hydrate that should run in a useEffect could use the same mechanism? Or would it need to be a separate context then? If we do SSR + persistence, we would somehow need to make sure that restoring from localStorage would run _after_ hydration from SSR is finished 🤔

src/persistQueryClient/PersistQueryClientProvider.tsx
```diff
@@ -0,0 +1,44 @@
...
+  persistOptions,
+  ...props
+}: PersistQueryClientProviderProps): JSX.Element => {
+  const [isHydrating, setIsHydrating] = React.useState(true)
+  const options = React.useRef(persistOptions)
+  React.useEffect(() => {
+    let unsubscribe: (() => void) | undefined
+
+    async function run() {
+      unsubscribe = await persistQueryClient({
```
Ephem, line 27: I think this might have a potential (theoretical?) issue. You don't get back the unsubscribe callback until after having awaited the `persistQueryClient`-call. If the effect cleans up while that is going on, it will never call unsubscribe and you will have a hanging subscription (not sure what consequences that will have).

TkDodo: yes, that's true. `persistQueryClient` is implemented as:

```
await persistQueryClientRestore(props)
return persistQueryClientSubscribe(props)
```

subscribing is sync, but restoring is async. In order to `await` it, we must return a promise. We could subscribe _before_ restoring, but I have bigger doubts about that because the restore itself might trigger a write then. Do you see a good way to solve this?

Ephem: I'm not sure it would be an issue in reality, but, maybe a "solution" could be to return an unsubscribe and a Promise synchronously. If the unsubscribe is called before the subscription happened, we just keep track of that fact and don't subscribe when the time comes. After subscription happens, unsubscribe just does its thing?

src/persistQueryClient/PersistQueryClientProvider.tsx
Ephem: So it's still in a loading state (`isLoading` is true), but you could look at the `fetchStatus` specifically to avoid showing a spinner if you want to take it a step further? I like it!

For SSR hydration/incremental page hydration this will get trickier. On the initial hydrate from the server markup, `isHydrating` should be false because we already have the data in the cache (hydration happened on cache/queryClient creation so to speak). Only _after_ that should we set `isHydrating`.

I think it would be nice to reuse the same context for all hydration if we can, but I'm also not sure about how the SSR+persistance use case should work, it's tricky.

Just to think out loud, I guess this provider as well as the planned hydration provider are both really "gates" so to speak. While the hooks versions of these would work fine on their own, the real problem with those is that the queries would still fire and trigger a fetch, even if we are currently in the process of hydrating "something". (We aren't really sure what we are hydrating is everything that is actually needed to render btw, so by blocking all queries some we actually do need to fetch might get slightly delayed, but I think we are talking about a neglectable amount of time compared to network latency.)

If we are rendering the `PersistQueryClientProvider` on the server, I guess we could want one of three things to happen:

1. We are telling the server we'll just go ahead and restore on the client, so don't worry about fetching any data, just render loading state placeholders
2. We are telling the server "on the client, I'm going to try to restore fresh data from some external cache, but since you don't have that available now you'll still need to go ahead and fetch all data needed"
3. Throw an error

Nr 1 just seems pointless to me. Nr 2 seems pretty complex since now the initial state on the client _can't_ be `isLoading && fetchStatus: 'paused'` since that's not the status it had on the server, there it was successful. We would have to delay restoring until after SSR hydration is complete. But if we have data from the server, we still don't want the queries to be in a `isLoading` state, since we already have data to show.

Or maybe this is even more complex? If we first hydrate from the server and then there are some components that were client rendered that are still waiting to fetch data, then in the persistance case _those_ queries should now be paused, and the queries that now have data should be successful, but in a refetching state while restoring..?

🥴

TkDodo: > So it's still in a loading state (isLoading is true), but you could look at the fetchStatus specifically to avoid showing a spinner if you want to take it a step further? I like it!

Yes, this is a new feature in v4. Currently, we use `paused` only when a query is paused due to no internet connection, but we could widen that case. We currently have `fetchStatus: 'idle' | 'paused' | 'fetching'`. we could even have a fourth state but I don't see the need for that. I just [blogged](https://tkdodo.eu/blog/offline-react-query) about it yesterday :)

> For SSR hydration/incremental page hydration this will get trickier. On the initial hydrate from the server markup, isHydrating should be false because we already have the data in the cache (hydration happened on cache/queryClient creation so to speak). Only after that should we set isHydrating (on page navigation and the like).

The plan is to have initial hydration happen via the `initialState` we pass to the `QueryClientProvider`, right? This would by synchronous and we wouldn't need any isHydrating state 👍 

> Just to think out loud, I guess this provider as well as the planned hydration provider are both really "gates" so to speak. While the hooks versions of these would work fine on their own, the real problem with those is that the queries would still fire and trigger a fetch, even if we are currently in the process of hydrating "something". (We aren't really sure what we are hydrating is everything that is actually needed to render btw, so by blocking all queries some we actually do need to fetch might get slightly delayed, but I think we are talking about a neglectable amount of time compared to network latency.)

Yes, I think that's correct. The `gates` basically aim to stop triggering a fetch. The observer is created, but not subscribed, so no fetch is initiated. That also means that `initialData` is taken into account, so without `initialDataUpdatedAt`, that data would be newer than the restored one.
Anyhow, the delay is acceptable I guess. You'd explicitly _want_ to delay fetching while restoring, because if you have _staleTime_ set and you hydrate sufficiently fresh data, there shouldn't be any data fetching at all!

> Or maybe this is even more complex? If we first hydrate from the server and then there are some components that were client rendered that are still waiting to fetch data, then in the persistance case those queries should now be paused, and the queries that now have data should be successful, but in a refetching state while restoring..?

For the combo of the two (SSR + persist reloading), I was think more of the SSG use-case:

- you pre-render a site on the server, it produces markup, we send that to the client.
- the client hydrates that via `initialState`, synchronously.
- then, the `persist` provider would try to restore additional state from the localStorage. This might be queries that we haven't SSGed on the server, or it might be newer variants of already cached queries. On the first visit, nothing would happen because you have no localStorage data. But if you do, you would get your data restored because it very likely has a newer `dataUpdatedAt` timestamp.

With SSR, this would likely have the opposite ordering: Since the SSRed data has a brand new timestamp, it is likely newer than your localStorage data, so the persist plugin would read out data but _not_ hydrate it. I guess that's okay and needs to be documented.

And for additional page navigations where we aim to hydrate in useEffect, the persist "gate" wouldn't need to do anything because restoring should only do something on page load - so we shouldn't have any conflicts here...

src/persistQueryClient/PersistQueryClientProvider.tsx
TkDodo: need to add some tests to see if it works, but is this what you had in mind?

261777cc

Ephem: Left a small comment on a detail, but yes, exactly the approach I was thinking about! 😃

TkDodo: good catch. next attempt: 0f6de7a7

src/reactjs/index.ts
```diff
@@ -13,7 +13,12 @@ export { useMutation } from './useMutation'
 export { useQuery } from './useQuery'
 export { useQueries } from './useQueries'
 export { useInfiniteQuery } from './useInfiniteQuery'
-export { useHydrate, Hydrate } from './Hydrate'
+export {
+  useHydrate,
+  Hydrate,
+  IsHydratingProvider,
+  useIsHydrating,
```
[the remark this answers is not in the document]
TkDodo: 459ac568

src/core/queryObserver.ts
```diff
@@ -456,6 +456,9 @@ export class QueryObserver<
           status = 'loading'
         }
       }
+      if (options._optimisticResults === 'isHydrating') {
+        fetchStatus = 'paused'
```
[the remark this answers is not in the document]
TkDodo: 363e3f78

TkDodo:
This would now be ready for review. @Ephem @tannerlinsley please have a look

docs/src/pages/plugins/persistQueryClient.md
```diff
@@ -48,6 +48,58 @@ It should be set as the same value or higher than persistQueryClient's `maxAge`
...
+  persister: localStoragePersister,
+})
+
+// 🚨 happens at the same time as restoring
+ReactDOM.render(<App />, rootElement)
+```
+
+### PeristQueryClientProvider
+
+For this use-case, you can use the `PersistQueryClientProvider`. It will make sure to subscribe / unsubscribe correctly according to the React lifecycle, and it will also make sure that queries will not start fetching while we are still restoring. Queries will still render though, they will just be put into `fetchingState: 'idle'` until data has been restored. Then, they will refetch unless the restored data is _fresh_ enough, and _initialData_ will also be respected. It can be used _instead of_ the normal `QueryClientProvider`:
```
arnaudbzn, line 72: Does it makes sense to have a new `fetchStatus = 'hydrating'` or `'restoring'`?

TkDodo: I thought about it, but with the removal of the idle state, all combinations of status + fetchStatus are possible. If we add this new one it will only work with loading state so that's not worth it imo

arnaudbzn: Ok!  
`useIsHydrating()` hook can be used  (if this will be officially supported in the docs) to know that queries are hydrating.

src/persistQueryClient/PersistQueryClientProvider.tsx
```diff
@@ -0,0 +1,46 @@
...
+      ...refs.current.persistOptions,
+      queryClient: client,
+    })
+
+    promise.then(() => {
+      refs.current.onSuccess?.()
+      setIsHydrating(false)
+    })
+
+    return unsubscribe
```
Ephem, line 38: Bit of an edge case and therefor a nit, but what would currently happen if `client` changes before the restore has happened? (I _think_ it would currently still restore the data from the old one, even if we have unsubscribed?) What should happen? Might make a good test.

I'm also wondering about the `promise.then` above, will that (and therefor `onSuccess`) get called even if we have done the effect cleanup? Should it?

src/persistQueryClient/index.ts
Ephem: Don't have a solution, just wanted to say I had/have the same problem with the hydrate stuff. :)

src/reactjs/Hydrate.tsx
```diff
@@ -3,6 +3,11 @@ import React from 'react'
 import { hydrate, HydrateOptions } from '../core'
 import { useQueryClient } from './QueryClientProvider'
 
+const IsHydratingContext = React.createContext(false)
+
+export const useIsHydrating = () => React.useContext(IsHydratingContext)
```
Ephem, line 8: Since we are at least _aiming_ at putting hydration data on a context now to avoid async hydration, we might end up not needing this for the regular hydration case. If that is the case, we might want to extract this out into persist instead and call it something not related to hydration to avoid confusion. A bit too early to say for sure though, since we might still need it if having hydration data on a context is too hard or complex for some reason. 🤷😄

src/persistQueryClient/PersistQueryClientProvider.tsx
Ephem: I just read this again and I think a bunch of these issues are more straightforward if normal hydration is sync via putting the data on a context like we've talked about.

Just wanted to note on the SSG/SSR usecase, since I never replied to that, that yes, often what you are describing is the case, but if you cache SSR you end up with the same case as the SSG one you are describing. 😄

src/persistQueryClient/PersistQueryClientProvider.tsx
TkDodo: yes, it would still restore from the old one and also try to restore from the new one. So there might be race conditions. I don't think the client is meant to change at this level, but it can. Some possible solutions:

- move `client` to the `refs` and change the dependency array to an empty array. This communicates that we only want to restore once, but it would not work if the client really changes.
- keep as-is, because race conditions wouldn't really matter. `hydration` can be done multiple times, even from different sources. The query with the highest `dataUpdatedAt` time "wins". So it doesn't matter which one gets hydrated first.
- somehow defer the second call until the first promise has finished - maybe by storing the promise on a ref?

I would think that keep as-is should be good enough. One related problem is that `promise.then` would set `isHydrating` to `false`. I think we need to set it to `true` when the effect starts, or maybe even keep a count of currently running hydrations?
I'm also missing error handling, so if hydration promise rejects, we'll stay in `isHydrating` forever. That I definitely need to fix!

TkDodo: just realised: the client is the target, not the source of hydration. For the source to change, it would be a different persister. I think that's very unlikely to be the case. For a different client, I can see the case:

- you create one client, start restoring
- then you change the client e.g. because you change something in the `defaultOptions`, maybe in a `useEffect` - you still want to restore and subscribe I guess?

In that case, order really doesn't matter. However, we'd to somehow "wait" until both hydrations are done before updating the Provider 🤔

TkDodo: okay so the count is tricky because if we start at `0` and increment in the effect, we get one render cycle with `isHydrating: false`, which is not good.

I tried to do nr. 3 from above with chained promises, storing the previous one on a ref and added a test. Honestly not sure if that code does what I think it does :/

507edca1

src/persistQueryClient/index.ts
TkDodo: hydration was moved to the core, but I really don't want to move this plugin to the core. I guess I'll leave it here for now, until someone complains that they want to use the plugin without react or so 😅

src/persistQueryClient/PersistQueryClientProvider.tsx
Ephem: Hmm, I see what you mean, it's a bit hard to reason about.

I guess if the client changes, we don't really care if the old client was hydrated or not (it's fine either way), we only care that _when_ the new client has been hydrated, then and only then do we call onSuccess and set isHydrating to false?

Just thinking out loud, but maybe you could leverage a `isStale` variable? Something like:

```jsx
  React.useEffect(() => {
    let isStale = false
    setIsHydrating(true)
    const [unsubscribe, promise] = persistQueryClient({
      ...refs.current.persistOptions,
      queryClient: client,
    })

    promise.then(() => {
      if (!isStale) {
        refs.current.onSuccess?.()
        setIsHydrating(false)
      }
    })

    return () => {
      isStale = true
      unsubscribe()
    }
}, [client])
```

TkDodo: i committed the change, but now the test I made fails that should check that onSuccess is only called once. I somehow can't run tests locally because of some weird errors. Can we keep the previous implementation, or do you know why onSuccess would be called twice with this implementation if the client changes?

TkDodo: I fixed my tests issue. I misread the test results. With the new implementation, `onSuccess` will only be called once, and it was called (and asserted in the test) twice before.

I like your proposal, so I updated the test and it should be green now: 433fd8a2

Ephem: Nice! Looks good to me. :)

Ephem (approved):
I haven't really tested it, but from a code review perspective I think this looks good to go, great work!

PR #3249, opened 2022-01-31
doc(queryClient): add dev warning with queryDefaults

GLabat:
Whenever several queryDefaults match a queryKey, a dev warning message is raised since this is error-prone.
Test added to highlight the behaviour.

Related to https://github.com/tannerlinsley/react-query/discussions/3199.


src/core/queryClient.ts
```diff
@@ -532,13 +533,35 @@ export class QueryClient {
...
+      partialMatchKey(queryKey, x.queryKey)
+    )
+    // It is ok not having defaults, but it is error prone to have more than 1 default for a given key
+    if (matchingDefaults?.length > 1) {
+      getLogger().warn(
+        `[QueryClient] Several defaults match with key '${JSON.stringify(
+          queryKey
+        )}'. The first matching query defaults are used. Please check how query defaults are registered. Order does matter here. cf. https://react-query.tanstack.com/reference/QueryClient#queryclientsetquerydefaults.`
+      )
+    }
```
TkDodo, line 554: let's please hide this behind a `if (process.env.NODE_ENV !== 'production')` check

src/core/queryClient.ts
```diff
@@ -532,13 +533,35 @@ export class QueryClient {
...
+    // Explicitly returns the first one
+    const firstMatchingDefaults = matchingDefaults?.[0]
+    return firstMatchingDefaults?.defaultOptions || undefined
+  }
+
   getQueryDefaults(
```
TkDodo, line 560: what about `getMutationDefaults` ?

src/core/queryClient.ts
```diff
@@ -532,13 +533,35 @@ export class QueryClient {
...
+    if (!queryKey) {
+      return undefined
+    }
+
+    // First retrieve all matching defaults for the given key
+    const matchingDefaults = this.queryDefaults.filter(x =>
+      partialMatchKey(queryKey, x.queryKey)
+    )
+    // It is ok not having defaults, but it is error prone to have more than 1 default for a given key
+    if (matchingDefaults?.length > 1) {
```
TkDodo, line 548: can `matchingDefaults` be undefined here? I don't think so as `.filter` will always return an aray:

```suggestion
    if (matchingDefaults.length > 1) {
```

src/core/queryClient.ts
```diff
@@ -532,13 +533,35 @@ export class QueryClient {
...
+      getLogger().warn(
+        `[QueryClient] Several defaults match with key '${JSON.stringify(
+          queryKey
+        )}'. The first matching query defaults are used. Please check how query defaults are registered. Order does matter here. cf. https://react-query.tanstack.com/reference/QueryClient#queryclientsetquerydefaults.`
+      )
+    }
+    // Explicitly returns the first one
+    const firstMatchingDefaults = matchingDefaults?.[0]
+    return firstMatchingDefaults?.defaultOptions || undefined
```
TkDodo, line 557: I think this could just be:

```
return matchingDefaults[0]?.defaultOptions
```

src/core/queryClient.ts
GLabat: Yep, I removed it but can't remember why…

src/core/queryClient.ts
GLabat: I haven't played much with mutation yet. Let me check and apply the same pattern if needed.

src/core/queryClient.ts
GLabat: ok.

src/core/queryClient.ts
GLabat: New `assert` helper in utils to hide all this.

src/core/queryClient.ts
GLabat: Indeed, both query and mutation defaults are working the same way.
I'm not familiar enough with TS to find a way to factorise the code, so I duplicated the code and changed what was needed.

I still think factorisation is required here. WDYT?
Could you give me some hints on how both `findFirstMatchingQueryDefaults` and `findFirstMatchingMutationDefaults` could be merged? `findFirstMatchingDefaults(defaultsStore, queryKey)` could do it. But I'm struggling with overload…

TkDodo: the main goal should be:
- not affect production bundle size
- not affect production performance

right now, we `filter` even though a `find` would be sufficient, which can terminate early. Also, we add a bit of code for production that is not necessary.

How about:
- doing the `.find` like before
- adding an env check, behind which we can:
  - do an additional `filter`
  - log if that produces more than one result

that code could also stay duplicated because it is stripped way in production anyhow. And during dev mode, it doesn't matter if we do find + filter, as in prod, it will stay only one `find`

TkDodo (changes requested):

src/core/utils.ts
```diff
@@ -74,6 +75,15 @@ export function noop(): undefined {
...
 }
 
+/**
+ * Raise a dev warning message when the condition is not fulfilled
+ */
+export function assert(condition: boolean, message: string): void {
+  if (process.env.NODE_ENV !== 'production' && !condition) {
+    getLogger().warn(message)
+  }
+}
```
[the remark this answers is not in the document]
GLabat: If needed, we could still have a version of assert for release build (noop-like), compatible with dead code elimination.
You know the codebase better than me 😅 so I remove the `assert`.

src/core/queryClient.ts
GLabat: Yep will do as suggested. I don't think the number of defaults would be too high but I agree, let's keep production code as performant as it could be.

src/core/queryClient.ts
```diff
@@ -535,10 +599,8 @@ export class QueryClient {
   getQueryDefaults(
     queryKey?: QueryKey
   ): QueryObserverOptions<any, any, any, any, any> | undefined {
-    return queryKey
-      ? this.queryDefaults.find(x => partialMatchKey(queryKey, x.queryKey))
-          ?.defaultOptions
-      : undefined
+    const queryDefaults = this.findFirstMatchingQueryDefaults(queryKey)
+    return queryDefaults
```
TkDodo, line 603: I don't see the need for an extra function here, as this would basically be:

```
return this.findFirstMatchingQueryDefaults(queryKey)
```

and we don't intend on using this function again. So I'd suggest we inline it please.

TkDodo (commented):
please update the branch with the latest `alpha` so that we can do a size comparison between the latest build and this branch.

src/core/queryClient.ts
GLabat: There is no "need". It is just the way I'm used to coding with the "1 function - 1 purpose" mantra.
I'll inline both functions, no pb.

GLabat:
> please update the branch with the latest `alpha` so that we can do a size comparison between the latest build and this branch.

Yup thanks.

TkDodo (approved):

TkDodo:
awesome, size is now the same as on `alpha`. Thank you 🙇

GLabat:
> awesome, size is now the same as on `alpha`. Thank you 🙇

Thanks for the review.

PR #3267, opened 2022-02-05
fix(types): useQueries to also handle array-params with typed QueryKey

artysidorenko:
Relates to discussion https://github.com/tannerlinsley/react-query/discussions/3242, and very similar to PR https://github.com/tannerlinsley/react-query/pull/3214 - this update ensures we also pass down the QueryKey type when working with regular array parameters (i.e. not a variadic tuple), making it easier to create wrappers around `useQueries`.


TkDodo:
thank you 🙏

PR #3271, opened 2022-02-07
Bail out if query data undefined

### LINKED ISSUE #2925: Make undefined an illegal cache value (for successful queries)

phatmann:
closes #2925 

@TkDodo here is the wildly simpler PR we have all been hoping for :-)

### DIFF
--- a/docs/src/pages/guides/migrating-to-react-query-4.md
+++ b/docs/src/pages/guides/migrating-to-react-query-4.md
@@ -254,7 +254,7 @@ Further, it is an easy bug to produce `Promise<void>` by adding logging in the q
 ```js
 useQuery(
   ['key'],
-  () => axios.get(url).then(result => console.log(result.data)
+  () => axios.get(url).then(result => console.log(result.data))
 )
 ```
 
--- a/docs/src/pages/guides/migrating-to-react-query-4.md
+++ b/docs/src/pages/guides/migrating-to-react-query-4.md
@@ -245,6 +245,21 @@ Types now require using TypeScript v4.1 or greater
 Starting with v4, react-query will no longer log errors (e.g. failed fetches) to the console in production mode, as this was confusing to many.
 Errors will still show up in development mode.
 
+### Undefined is an illegale cache value for successful queries
+
+In order to make bailing out of updates possible by returning `undefined`, we had to make `undefined` an illegal cache value. This is in-line with other concepts of react-query, for example, returning `undefined` from the [initialData function](guides/initial-query-data#initial-data-function) will also _not_ set data.
+
+Further, it is an easy bug to produce `Promise<void>` by adding logging in the queryFn:
+
+```js
+useQuery(
+  ['key'],
+  () => axios.get(url).then(result => console.log(result.data)
+)
+```
+
+This is now disallowed on type level; at runtime, `undefined` will be transformed to a _failed Promise_, which means you will get an `error`, which will also be logged to the console in development mode.
+
 ## New Features 🚀
 
 ### Proper offline support
@@ -265,3 +280,14 @@ Mutations can now also be garbage collected automatically, just like queries. Th
 ### Tracked Queries per default
 
 React Query defaults to "tracking" query properties, which should give you a nice boost in render optimization. The feature has existed since [v3.6.0](https://github.com/tannerlinsley/react-query/releases/tag/v3.6.0) and has now become the default behavior with v4.
+
+### Bailing out of updates with setQueryData
+
+When using the [functional updater form of setQueryData](../reference/QueryClient#queryclientsetquerydata), you can now bail out of the update by returning `undefined`. This is helpful if `undefined` is given to you as `previousValue`, which means that currently, no cached entry exists and you don't want to / cannot create one, like in the example of toggling a todo:
+
+```js
+ queryClient.setQueryData(
+   ['todo', id],
+   (previousTodo) => previousTodo ? { ...previousTodo, done: true } : undefined
+)
+ ```
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -1,7 +1,5 @@
 import {
   getAbortController,
-  Updater,
-  functionalUpdate,
   noop,
   replaceEqualDeep,
   timeUntilStale,
@@ -195,13 +193,10 @@ export class Query<
   }
 
   setData(
-    updater: Updater<TData | undefined, TData>,
+    data: TData,
     options?: SetDataOptions & { notifySuccess: boolean }
   ): TData {
     const prevData = this.state.data
-
-    // Get the new data
-    let data = functionalUpdate(updater, prevData)
 
     // Use prev data if an isDataEqual function is defined and returns `true`
     if (this.options.isDataEqual?.(prevData, data)) {
@@ -438,11 +433,41 @@ export class Query<
       this.dispatch({ type: 'fetch', meta: context.fetchOptions?.meta })
     }
 
+    const onError = (error: TError | { silent?: boolean }) => {
+      // Optimistically update state if needed
+      if (!(isCancelledError(error) && error.silent)) {
+        this.dispatch({
+          type: 'error',
+          error: error as TError,
+        })
+      }
+
+      if (!isCancelledError(error)) {
+        // Notify cache callback
+        this.cache.config.onError?.(error, this as Query<any, any, any, any>)
+
+        if (process.env.NODE_ENV !== 'production') {
+          getLogger().error(error)
+        }
+      }
+
+      if (!this.isFetchingOptimistic) {
+        // Schedule query gc after fetching
+        this.scheduleGc()
+      }
+      this.isFetchingOptimistic = false
+    }
+
     // Try to fetch the data
     this.retryer = createRetryer({
       fn: context.fetchFn as () => TData,
       abort: abortController?.abort?.bind(abortController),
       onSuccess: data => {
+        if (typeof data === 'undefined') {
+          onError(new Error('Query data cannot be undefined') as any)
+          return
+        }
+
         this.setData(data as TData)
 
         // Notify cache callback
@@ -454,30 +479,7 @@ export class Query<
         }
         this.isFetchingOptimistic = false
       },
-      onError: (error: TError | { silent?: boolean }) => {
-        // Optimistically update state if needed
-        if (!(isCancelledError(error) && error.silent)) {
-          this.dispatch({
-            type: 'error',
-            error: error as TError,
-          })
-        }
-
-        if (!isCancelledError(error)) {
-          // Notify cache callback
-          this.cache.config.onError?.(error, this as Query<any, any, any, any>)
-
-          if (process.env.NODE_ENV !== 'production') {
-            getLogger().error(error)
-          }
-        }
-
-        if (!this.isFetchingOptimistic) {
-          // Schedule query gc after fetching
-          this.scheduleGc()
-        }
-        this.isFetchingOptimistic = false
-      },
+      onError,
       onFail: () => {
         this.dispatch({ type: 'failed' })
       },
--- a/src/core/queryClient.ts
+++ b/src/core/queryClient.ts
@@ -8,6 +8,7 @@ import {
   partialMatchKey,
   hashQueryKeyByOptions,
   MutationFilters,
+  functionalUpdate,
 } from './utils'
 import type {
   QueryClientConfig,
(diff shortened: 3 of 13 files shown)


src/core/queryClient.ts
```diff
@@ -129,14 +130,28 @@ export class QueryClient {
...
+      return
+    }
+
+    if (!query) {
+      const parsedOptions = parseQueryArgs(queryKey)
+      const defaultedOptions = this.defaultQueryOptions(parsedOptions)
+      query = this.queryCache.build(this, defaultedOptions)
+    }
+
+    return query.setData(data, options)
```
TkDodo, line 154: if we go this way, I think there would be no need for `query.setData` to accept a functional updater, as we now only call it with `data`. That would allow us to get rid of some code in `query.ts` that we have re-introduced here.

However, from our discussion on the other PR, I thought we were going a slightly different route by _disallowing_ `undefined` as a cache value altogether.

It is a common source of error that people do not return anything from their `queryFn` because they do things like:

```
axios.get('/api').then(result => console.log(result))
```

which then returns `Promise<void>`. Of course, TypeScript helps :)

But I like the current approach much more because it's simpler. So how about we do:

- the proposed change here to `query.setData`
- add a dev mode error logging here if we get `undefined` just to make people aware that this is disallowed
https://github.com/tannerlinsley/react-query/blob/c0fc916cb3e7d32a05e9356b2f1359c3cd5de573/src/core/query.ts#L446
- try to tweak the typescript types to disallow `undefined` for `TData` on type level - if that's possible. `null` should still be allowed though.
- we'd need to update the docs (migration guide), but I can do that too if you want

phatmann: @TkDodo If I understand you correctly, the changes that I made so far is good. That is, if `setQueryData` is passed undefined data or a function that returns undefined, bail out. We just need additional changes to prevent undefined being written to the cache. Per #2925:

> Also, if a fetch function returns undefined, we should:
> * transform it to a failed promise (error)
> * warn during development mode about it

That is, I need to modify `setData` as you propose above (including removing extra functional update handling.)

I will also look into the type change.

I will look at the docs and see if I can update them satisfactorily.

phatmann: Looking at your comments again, I believe you are proposing to _not_ fail a query when the queryFn returns undefined, but instead to bail out. This is what I tried before, and I discovered it breaks clients that are expecting the query status to change as a result of a fetch. I recommend we stick with your original proposal and fail the query if the queryFn returns undefined.

TkDodo: Transforming to error is fine for me 👍

phatmann:
@TkDodo The checks failed due to a TS error in useQueries.test.tsx. However, I am not seeing the error on my machine, and the line number in the error is for the generated JS code. Any idea what I need to do to reproduce it?

src/core/queryClient.ts
TkDodo, line 154: looking at this once more, I don't think we need the `if` guard:

```suggestion
const parsedOptions = parseQueryArgs(queryKey)
const defaultedOptions = this.defaultQueryOptions(parsedOptions)
return this.queryCache
  .build(this, defaultedOptions)
  .setData(data, options)
```

src/core/query.ts
```diff
@@ -219,14 +217,19 @@ export class Query<
...
-  ): TData {
-    const prevData = this.state.data
+  setData(data: TData, options?: SetDataOptions): TData {
+    if (typeof data === 'undefined') {
+      const message = 'Cache data cannot be undefined'
+      getLogger().error(message)
+      const error = new Error(message) as any
+      this.dispatch({
+        type: 'error',
+        error,
```
TkDodo, line 227: my idea was to _not_ do this in `setData`, but keep `setData` to only accept `TData` and do the transformation on the only call-side:

https://github.com/tannerlinsley/react-query/blob/c0fc916cb3e7d32a05e9356b2f1359c3cd5de573/src/core/query.ts#L445-L446

there we could do the check and then:
- extract the `onError` function of the retrier to it's own function
- just call `onError` with the newly created error. That would:
  - dispatch to error state
  - already does the logging in dev mode for us:
https://github.com/tannerlinsley/react-query/blob/c0fc916cb3e7d32a05e9356b2f1359c3cd5de573/src/core/query.ts#L470-L472

TkDodo:
> @TkDodo The checks failed due to a TS error in useQueries.test.tsx. However, I am not seeing the error on my machine, and the line number in the error is for the generated JS code. Any idea what I need to do to reproduce it?

that error is weird indeed. I will have to check it out locally, too.

phatmann:
@TkDodo I see that I again misread your earlier comments:

> The proposed change here to query.setData

You meant removing the functional handling. **Got it.**

> add a dev mode error logging here if we get undefined just to make people aware that this is disallowed

By "here" you meant [query.ts at line 446](https://github.com/tannerlinsley/react-query/blob/c0fc916cb3e7d32a05e9356b2f1359c3cd5de573/src/core/query.ts#L446). **Got it.**

src/core/queryClient.ts
phatmann: Right, since we did not have the guard before.

src/core/queryClient.ts
```diff
@@ -129,14 +130,28 @@ export class QueryClient {
...
-  ): TData {
-    const parsedOptions = parseQueryArgs(queryKey)
-    const defaultedOptions = this.defaultQueryOptions(parsedOptions)
-    return this.queryCache
-      .build(this, defaultedOptions)
-      .setData(updater, options)
+  ): TData | undefined {
+    if (typeof updater === 'undefined') {
+      return undefined
+    }
```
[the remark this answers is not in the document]
phatmann: Makes sense. Will update code.

src/core/query.ts
phatmann: Sounds way better On it.

phatmann:
@TkDodo I made all the changes you requested.

src/core/query.ts
```diff
@@ -450,6 +447,12 @@ export class Query<
       fn: context.fetchFn as () => TData,
       abort: abortController?.abort?.bind(abortController),
       onSuccess: data => {
+        if (typeof data === 'undefined') {
+          const error = new Error('Query data cannot be undefined') as any
+          this.onError(error)
```
TkDodo, line 452: does this work?

```suggestion
          this.onError(new Error('Query data cannot be undefined') as TError)
```

src/core/query.ts
```diff
@@ -460,28 +463,7 @@ export class Query<
...
-          // Log error
-          getLogger().error(error)
-        }
-
-        // Remove query after fetching if cache time is 0
-        if (this.cacheTime === 0) {
-          this.optionalRemove()
-        }
-      },
+      onError: this.onError,
```
TkDodo, line 466: it doesn't have to live on the `this` scope, I think that is too wide. Also binding comes with an overhead in bundle size I believe. It was an inline function before, so I think it can just be:

`const onError: (error: TError | { silent?: boolean }) => ...`

defined right above `this.retryer = new Retryer({`

src/core/queryClient.ts
```diff
@@ -129,14 +130,21 @@ export class QueryClient {
 
   setQueryData<TData>(
     queryKey: QueryKey,
-    updater: Updater<TData | undefined, TData>,
+    updater: Updater<TData | undefined, TData> | undefined,
     options?: SetDataOptions
-  ): TData {
+  ): TData | undefined {
+    let query = this.queryCache.find<TData>(queryKey)
```
TkDodo, line 136: can be `const` now:

```suggestion
    const query = this.queryCache.find<TData>(queryKey)
```

src/core/queryClient.ts
```diff
@@ -129,14 +130,21 @@ export class QueryClient {
...
+    const prevData = query?.state.data
+    const data = functionalUpdate(updater, prevData)
+
+    if (typeof data === 'undefined') {
+      return
```
TkDodo, line 141: just to be explicit and consistent with the rest of the codebase:

```suggestion
      return undefined
```

src/core/queryClient.ts
```diff
@@ -129,14 +130,21 @@ export class QueryClient {
...
+    }
+
     const parsedOptions = parseQueryArgs(queryKey)
     const defaultedOptions = this.defaultQueryOptions(parsedOptions)
-    return this.queryCache
-      .build(this, defaultedOptions)
-      .setData(updater, options)
+    query = this.queryCache.build(this, defaultedOptions)
+    return query.setData(data, options)
```
TkDodo, line 147: there is no need to reassign to `query` here. We can just do:

```
return this.queryCache.build(this, defaultedOptions).setData(data, options)
```

TkDodo (commented):
looks really good! Just when I commented about `new Retryer`, I realized that this PR goes to `master` and not `alpha`, which is our v4 branch (I noticed because I refactored retryer to a function 😅). Since this is a breaking change we want this for v4 please. Can you rebase that PR to v4 ? Thanks 🙏

src/core/query.ts
phatmann: I tried that before. TS says:
> Conversion of type 'Error' to type 'TError' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  'TError' could be instantiated with an arbitrary type which could be unrelated to 'Error'.

So I need to cast to `any`. This makes sense, since errors coming in via try/catch are of type `any` (or `unknown` in newer TS versions). I wish JS had a real base Error type. Oh well.

TkDodo: yep, TS is right here. The user can specify anything as `TError`, for example:

```
const { error } = useQuery<number, string>('key', () => Promise.resolve(5))

error?.toUppercase()
```

in that example, they would expect `error` to be a `string` - which it may or may not be. They would be in for a surprise if they then get an `Error`. We do the same thing already if `select` throws an error - we transform to `Error` state.

There are bunch of things we could do:
- ensure that `TError extends Error`. But you can reject a promise with non-Errors, too.
- remove the possibility to type `TError` completely and just always use `unknown`, which is the only "safe" thing to do. That would be pretty breaking for TS users though.

Jus thinking out loud. Nothing we need to do on this PR, the `any` cast is fine :)

phatmann:
@TkDodo when I try to force push the rebased branch, I get this error:

> ! [remote rejected] pr/set-query-data-undefined-bail-out -> pr/set-query-data-undefined-bail-out (refusing to allow an OAuth App to create or update workflow `.github/workflows/test-and-publish.yml` without `workflow` scope)

Should I open a new PR and close this one?

phatmann:
@TkDodo actually I cannot push my branch at all. Nothing to do with the PR. Some workflow changes must have crept into branch, will see if I can find them.

phatmann:
@TkDodo this is apparently a [VS Code Github auth issue](https://stackoverflow.com/questions/64059610/how-to-resolve-refusing-to-allow-an-oauth-app-to-create-or-update-workflow-on). Will fix and push ASAP.

phatmann:
@TkDodo changes are all in and rebased to alpha.

(I just had to switch to using SSH URL ☺️ )

phatmann:
By using this command in the failing test above

> npm run test:types && npm run test:format && npm run test:eslint && jest

I can see the error. Looking into it.

TkDodo (approved):
❤️ it, thank you 🙇

phatmann:
@TkDodo Tests are all passing now

TkDodo:
> @TkDodo Tests are all passing now

oh nice! That was a sneak `return undefined` in that test from the queryFn 😅 . I guess that shows that types are now catching this :)

I'll add some documentation to the upgrade guide and then we can ship this - thank you so much for this 🙏❤️

docs/src/pages/guides/migrating-to-react-query-4.md
```diff
@@ -245,6 +245,21 @@ Types now require using TypeScript v4.1 or greater
...
+### Undefined is an illegale cache value for successful queries
+
+In order to make bailing out of updates possible by returning `undefined`, we had to make `undefined` an illegal cache value. This is in-line with other concepts of react-query, for example, returning `undefined` from the [initialData function](guides/initial-query-data#initial-data-function) will also _not_ set data.
+
+Further, it is an easy bug to produce `Promise<void>` by adding logging in the queryFn:
+
+```js
+useQuery(
+  ['key'],
+  () => axios.get(url).then(result => console.log(result.data)
```
TkDodo, line 257:
```suggestion
  () => axios.get(url).then(result => console.log(result.data))
```

TkDodo:
@phatmann I missed that we are now still having lots of log statements during the tests:

Screenshot 2022-02-10 at 15 32 44

could you check from which tests these are coming from?

TkDodo:
I fixed it here: f6a6ae3f

PR #3274, opened 2022-02-08
docs(mutations): describe difference between using callbacks in useMu…

zrwsk:
Hi Folks!

I just noticed today that there is no explicit mention in Mutations guide about the fact that callbacks passed to `useMutation` will behave slightly different than the ones used in `.mutate`.

There is only small remark about that [here](https://react-query.tanstack.com/reference/useMutation):
> If you make multiple requests, onSuccess will fire only after the latest call you've made.

I hope it may be useful to have this information more visible and that I haven't broken any contribution rules :)

### DIFF
--- a/docs/src/pages/guides/mutations.md
+++ b/docs/src/pages/guides/mutations.md
@@ -180,7 +180,9 @@ mutate(todo, {
 ```
 
 ### Consecutive mutations
-For consecutive mutations, you may consider passing `onSuccess`, `onError` and `onSettled` to the `mutate` function instead of `useMutation`. Then, these callbacks will fire only _once_ after last mutation is fulfilled, and only if the component is still mounted. It might be useful for performance optimisation (ie. when invalidating queries after mutation). This is due to the fact that the mutation observer is removed and resubscribed every time when the `mutate` function is called.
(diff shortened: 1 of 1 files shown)


docs/src/pages/guides/mutations.md
```diff
@@ -179,6 +179,24 @@ mutate(todo, {
 })
 ```
```
TkDodo, line 181: I would appreciate it if we could add another sub-heading here (level 3). That would make it easier to send that information around.

docs/src/pages/guides/mutations.md
```diff
@@ -179,6 +179,24 @@ mutate(todo, {
...
 
+For consequent mutations you may consider passing `onSuccess`, `onError` and `onSettled` to `mutate` function instead of `useMutation`. Then, these callbacks will fire only _once_ after last mutation is fulfilled. It might be useful performance optimisation (ie. when invalidating queries after mutation). This is due to the fact that mutation observer is removed and resubscribed every time when `mutate` function is called.
```
TkDodo, line 182:
```suggestion
For consecutive mutations, you may consider passing `onSuccess`, `onError` and `onSettled` to the `mutate` function instead of `useMutation`. Then, these callbacks will fire only _once_ after last mutation is fulfilled, and only if the component is still mounted. It might be useful for performance optimisation (ie. when invalidating queries after mutation). This is due to the fact that the mutation observer is removed and resubscribed every time when the `mutate` function is called.
```

TkDodo (commented):
love the example ❤️

TkDodo (approved):

TkDodo:
@all-contributors add @zrwsk for docs

PR #3286, opened 2022-02-10
Update react-native.md

engWafula:


TkDodo:
I think the title is fine as it is - everything in the docs is about `React Query` after all :)

PR #3288, opened 2022-02-10
docs(useQuery): note effect of initialData on loading status

jaykaron:
The fact that `initialData` affects `isLoading` during the inital fetch was not clear to me on first read (it will always be `false`).

### DIFF
--- a/docs/src/pages/reference/useQuery.md
+++ b/docs/src/pages/reference/useQuery.md
@@ -188,7 +188,7 @@ const result = useQuery({
 - `status: String`
   - Will be:
     - `idle` if the query is idle. This only happens if a query is initialized with `enabled: false` and no initial data is available.
-    - `loading` if the query is in a "hard" loading state. This means there is no cached data and the query is currently fetching, eg `isFetching === true`
+    - `loading` if the query is in a "hard" loading state. This means there is no cached data or `initialData` and the query is currently fetching, eg `isFetching === true`
     - `error` if the query attempt resulted in an error. The corresponding `error` property has the error received from the attempted fetch
     - `success` if the query has received a response with no errors and is ready to display its data. The corresponding `data` property on the query is the data received from the successful fetch or if the query's `enabled` property is set to `false` and has not been fetched yet `data` is the first `initialData` supplied to the query on initialization.
 - `isIdle: boolean`


docs/src/pages/reference/useQuery.md
```diff
@@ -188,7 +188,7 @@ const result = useQuery({
 - `status: String`
   - Will be:
     - `idle` if the query is idle. This only happens if a query is initialized with `enabled: false` and no initial data is available.
-    - `loading` if the query is in a "hard" loading state. This means there is no cached data and the query is currently fetching, eg `isFetching === true`
+    - `loading` if the query is in a "hard" loading state. This means there is no cached data or `initialData` and the query is currently fetching, eg `isFetching === true`
```
TkDodo, line 191: if we change that, it shouldn't be here please, but rather somewhere here:
- https://react-query.tanstack.com/guides/initial-query-data

From a query perspective, it doesn't matter where the data is coming from. It could get into the cache via a variety of ways:
- prefetchQuery
- initialData
- setQueryData
- hydration
- restore from a persist plugin

So here, the sentence is alright: A hard loading state just means there is no data in the cache.

On the page about initial query data, we explicitly say:

> IMPORTANT: initialData is persisted to the cache, so it is not recommended to provide placeholder, partial or incomplete data to this option and instead use placeholderData

I also go into great details in the community resources:

https://react-query.tanstack.com/community/tkdodos-blog#9-placeholder-and-initial-data-in-react-query

jaykaron: Right. After going through more of the docs and resources I realized what was going on and that with context just stating "no cached data" is sufficient. However, before reading all the resources, this was the first place I looked when I was confused why `isLoading` was `false` and the current text wasn't fully clear to me.

My phrasing of "cached data or `initialData`" should probably be reworded because it makes it sound like they are separate cases, which they aren't, but it would've been helpful to me to have a note here that "cached data" can come from the variety of sources that you mentioned and not necessarily from a fetch.

Maybe linking the phrase "cached data" to one of the other resources would make it clearer without mixing in `initialData`. I couldn't find an exact page that would be appropiate.

TkDodo: I'll close this for now. if you want, you can make a PR to the initial data docs that improve it and states that initialData will put a query in success state.

PR #3289, opened 2022-02-10
refactor: inline / remove some functions for size improvements

TkDodo:


jcready:
Size of what exactly? The file size or memory size? I see that this reduced the total number of lines by 120, but it's not clear if it actually reduced the *significant* lines of code that much.

When we inline these functions we're trading idle memory usage for more garbage collection. I'd almost always choose to more memory usage over GC so I'm curious why this choice was made.

TkDodo:
yes, file size. The  improvements were not significant, that's right - about 1.2%. I only stumbled upon one function that could be completely removed, and the `default` cases in the reducers could also safely be removed.

> When we inline these functions we're trading idle memory usage for more garbage collection. I'd almost always choose to more memory usage over GC so I'm curious why this choice was made.

I'm not an expert here, so I'm happy to revert the inlining changes. As far as I'm aware, the inline functions can be created and then garbage collected right after the usage, so it shouldn't use more memory? Maybe even less so because the functions don't have to be held for as long as we hold the class? So are you talking about the cost of garbage collection itself? I've never really worried about function creation + GC before 😅  but maybe I should?

PR #3290, opened 2022-02-11
docs: add caveats about memory to ssr.md

meskill:


docs/src/pages/guides/ssr.md
```diff
@@ -196,3 +199,9 @@ A query is considered stale depending on when it was `dataUpdatedAt`. A caveat h
 Because `staleTime` defaults to `0`, queries will be refetched in the background on page load by default. You might want to use a higher `staleTime` to avoid this double fetching, especially if you don't cache your markup.
 
 This refetching of stale queries is a perfect match when caching markup in a CDN! You can set the cache time of the page itself decently high to avoid having to re-render pages on the server, but configure the `staleTime` of the queries lower to make sure data is refetched in the background as soon as a user visits the page. Maybe you want to cache the pages for a week, but refetch the data automatically on page load if it's older than a day?
+
+### High memory consumption on server
+
+In case you are creating `QueryClient` for every request React Query creates the isolated cache for this client and this cache is preserved in memory for the `cacheTime` period. That may lead to high memory consumption on server in case of high number of requests during `cacheTime` period.
```
TkDodo, line 205:
```suggestion
In case you are creating the `QueryClient` for every request, React Query creates the isolated cache for this client, which is preserved in memory for the `cacheTime` period (which defaults to 5 minutes). That may lead to high memory consumption on server in case of high number of requests during that period.
```

docs/src/pages/guides/ssr.md
```diff
@@ -196,3 +199,9 @@ A query is considered stale depending on when it was `dataUpdatedAt`. A caveat h
...
+
+To clear cache after it is not needed and by that lower memory consumption you can add call to the [`queryClient.clear()`](../reference/QueryClient#queryclientclear) after request is handled and Query dehydrated state has been sent to the client.
```
TkDodo, line 207:
```suggestion
To clear the cache after it is not needed and to lower memory consumption, you can add a call to [`queryClient.clear()`](../reference/QueryClient#queryclientclear) after the request is handled and dehydrated state has been sent to the client.

Alternatively, you can set a smaller `cacheTime`.
```

TkDodo (approved):

PR #3302, opened 2022-02-16
Feature/remove idle state

TkDodo:


arnaudbzn:
@TkDodo Great change that simplifies Query `status` while providing more details about what is going on with the `fetchStatus`⚡️

TiagoPortfolio:
Hi @TkDodo !

I want to display a loading spinner specifically on the first data fetch, I guess now I have to change my code from
`myQuery.status === 'loading'`
to
`myQuery.status === 'loading' && myQuery.fetchStatus === 'fetching'`
?

Cheers!

TkDodo:
@TiagoPortfolio to be on the safe side, yes, that would be the equivalent. However, the other cases are quite edgy:

- `status: 'loading', fetchStatus: 'idle'` - this can only happen if you have a query that is `enabled: false` and no data exists, or if you restore a query from one of the persist plugins while we are restoring
- `status: 'loading', fetchStatus: 'idle'` - this can only happen if you mount a query for the first time, but have no internet connection

The second case is not handled in v3 at all, so you would also just show a loading spinner there until you go online. So if you don't use the `enabled` option, checking for `status: loading` should still be enough. If you do, you have to use the extra check and "do something else" if you are idle.

sharpSteff:
This should be mentioned on the breaking changes for v4.

TkDodo:
It is: https://react-query-beta.tanstack.com/guides/migrating-to-react-query-4#the-idle-state-has-been-removed

sharpSteff:
Great! Sorry. My bad, somehow missed it ...

PR #3320, opened 2022-02-20
feat(tkdodos-blog): add Offline React Query post

MostafaNawara:


TkDodo:
Thank you. I wanted to wait with this until v4 was released - but maybe you can make this go to the `alpha` branch instead? Then we would have it on https://react-query-alpha.tanstack.com :)

MostafaNawara:
@TkDodo Thank you!. I have changed the base branch to `alpha`.

docs/src/pages/community/tkdodos-blog.md
```diff
@@ -53,3 +53,9 @@ React Query maintainer [TkDodo](https://twitter.com/tkdodo) has a series of blog
 ## [#12: Mastering Mutations in React Query](https://tkdodo.eu/blog/mastering-mutations-in-react-query)
 
 > Mutations are the important, second part necessary to work with server data - for situations where you need to update it. This blog post covers what mutations are and how they are different from queries. You'll learn the difference between `mutate` and `mutateAsync` as well as how you can tie queries and mutations together. [Read more...](https://tkdodo.eu/blog/mastering-mutations-in-react-query)
+
+
+## [#13: Offline React Query](https://tkdodo.eu/blog/offline-react-query)
+
+> You'll learn about the different offline strategies with React Query.
+[Read more...](https://tkdodo.eu/blog/offline-react-query)
```
TkDodo, line 61: I'd like to have a bit of an excerpt here usually, which is why I've copied the relevant paragraph here:

```suggestion
> There are many ways to produce promises - which is everything React Query needs - but by far the biggest use-case is data fetching. Very often, that requires an active network connection. But sometimes, especially on mobile devices where, the network connection can be unreliable, you need your app to also work without it. In this article, you'll learn about the different offline strategies React Query offers. [Read more...](https://tkdodo.eu/blog/offline-react-query)
```

PR #3322, opened 2022-02-20
docs: Extend simple example to handle failed request

ASchwad:
Hey there,
I recently stumpled across a [misunderstanding](https://github.com/tannerlinsley/react-query/issues/2258#issuecomment-1034707667) of the library and the [referenced simple example](https://codesandbox.io/s/github/tannerlinsley/react-query/tree/master/examples/simple).
Since the example contains an error handling with respect to the UI, it should also correctly recognize occurring errors. As mentioned in the docs, [fetch and other clients do not throw by default](https://react-query.tanstack.com/guides/query-functions#usage-with-fetch-and-other-clients-that-do-not-throw-by-default) an error - so I added the error handling to the example within the result handler.

### DIFF
--- a/examples/simple/src/index.js
+++ b/examples/simple/src/index.js
@@ -18,7 +18,14 @@ function Example() {
   const { isLoading, error, data, isFetching } = useQuery("repoData", () =>
     fetch(
       "https://api.github.com/repos/tannerlinsley/react-query"
-    ).then((res) => res.json())
+    ).then((res) => {
+      // Manually throwing error if request was not successful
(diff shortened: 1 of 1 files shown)


TkDodo:
I would want to keep the simple example as simple as possible (minimal lines of code). `fetch` has this unfortunate spec that 4xx and 5xx are not treated as failed promises. Maybe it would be better to just use `axios` or `ky` for this example?

ASchwad:
Fair point. I changed the PR to use Axios instead

TkDodo:
interesting that we had axios in the dependencies already 😅

TkDodo (approved):

PR #3336, opened 2022-02-24
fix(createAsyncStoragePersister): persistClient not respect throttleTime and not persisted final state queryClient (#3331)

### LINKED ISSUE #3331: createAsyncStoragePersistor's persistClient not respect throttleTime and not persisted final state queryClient
### Describe the bug

The persistClient of createAsyncStoragePersistor does not execute according to the throttleTime and not persisted final state queryClient when make many queries to refetch. To be precise, there is a problem with the function [asyncThrottle](https://github.com/tannerlinsley/react-query/blob/aa853d826d6776e073b9f5dbdd38e5ffa27a52bb/src/createAsyncStoragePersister/index.ts#L54).

https://www.typescriptlang.org/play?#code/FAYw9gdgzgLgBDAprAjHAvHAhlAnhEAFQAsAnMGGAG0QAod8Q5aBLAEwC44IBXAWwBGiUgEoMAPjgBvYHDjhoYGgDoqYAOa0ABiByI4aVpzgASKewC+I5QhZ9kMLHwAOXMwBEsSZRDAB3WhELLREAbllsPywWeDZEKixcWgBmAAYw4AsAGmkIlggkUgA3LCouFFTKrMyM0EhYBAcAJgxsPAISckoaenamIy5eQWExdEkZOQUoJURVDW1dKH0WgdNzNisbGDsHJ1c1z29fAKCQ8LksKJi4OISkpsr00LgAehf5MAbt+zgBFnV1MIEMQsBA4PlCiUqJkchNwQVhFDyo9qlZwsA3nAAMJUFggADWcAA5AAlHgQIm-HiUSAIMDyYiIAlwMDU5zUuBqdRQDjADHvRb6FC8zEAUQAHs4mfAuQg-PTcRBkFT4OoKHAAGas0jYCBsTk4eDscFQeHgUikaXKYC9RjMUbjCJIVC0FAZOTOmAoWhNd2RaKxeKJFIAVj9nu9yT9lwDNyDSWSYfOjRdABYMlZAujMYK4E0Re8JVKQDKNHKFfllQIOWr4JB9KD9QkGsaWKb8ubLSXrbaCPaJLkPc1XeHh77wpnagw+xrySWWLTbsG+DzuPwhKJB3J-dclX44AAFch8Nt0WiW6ZUIqIB1b7fbpYwQg7Vkwc-IJTXnIrv1yNGZPl+TgYhKGcHk3nUGJiB4ARlHAPgXkcCAlVIRUoBoXAXktLASwAWgARx4YRMIENQBBeLAsAADhDZI2CopoADY2EYgB2VjGMQVJWOSAQAE4NRDNgBDYNhkioxAQw1DUsCaVisBDJoBHIqBSBAF4QGwpAAEE+gAZRgMBSCwQED2EKA20KF58jicVlBgKAAGIABkQ1TYBZwIbZaWnIgyAoahEAAHm00huTgRBxSQPVTWwthICoXA4HJfFjggABtABdHISWQHgqBgcRaAiTyQC4WhlEqrAwtXULuVvI8wBPJYgtyqB8sK6o5CkeFIVKVoKiqTk7GuTA0AsLgeohRFSgAfkGddhGeXETxgea12GHULFaKQLGAMQ4RYDVmBgXApTAY7SrgABCdBMCJUrvIpMQYH8-c9zgUULSM2giWq9R+EQAoTW4dVHoXCBlCJEQIimeBSHJCB8nUHb5B4C0gZgLgZKoJY4D2uQaBlQ1nx+TBUgiInbHsV8uFymB0YgQgzuC07zuOx9ScQV9xFh+p4EIxAiK4ULjNwEKaskTAsoiS0GdIMEKqqmqRZqh0IjkXt+lvOF7yO5gEeQ5G4PRy0CgOjX7wNWAudaQ5ZmOQJLb147aEFojVCB9RXrgSQVpiC2rat93ZigYgjrfGGg7-Plo7gEPlHZMP6DV53txAGhqq519aG+bnqSjq2Cat-XaHtnx-ECOBcOtp8dl93qZqoQPo8NpGIHUE2MeBzAYARxA04uK54FKpXlH+qBC6D5s67JuBy8dqerbb42QFNzHWhxpY0+2+I8d1kvXYTmgO59yR0jvaO4drurTUwEP0uPr2fZrlBMuuwftzz19Wk5l9qS1v2MYl846l2uivDuXczYwBbnHe8EDO5r27vAXu-dP5BxjNcUelVVCGlvkvOBcAEFQI3pgLeA9CEx0IdkRuxRSgEO3MXe8TD8YiCdhYIAA

### Steps to reproduce

Just run the code on the TypeScript playground.
1. Click 'Run' button on the left panel.
2. Click  'Logs'  button on the right panel.
3. Check output logs.

### Expected behavior

Case1: Expect log two lines and print maximum id but got four lines and not print maximum id.
Case2: Expect log two lines and print maximum id but got one line and not print maximum id.

_No response_

### Screenshots or Videos

_No response_

### Platform

Only react native.

### react-query version

v4.0.0-alpha.14 and v3.34.16

### TypeScript version

_No response_

### Additional context

_No response_

huanguolin:
closes #3331

fix(createAsyncStoragePersister): persistClient not respect throttleTime and not persisted final state queryClient (#3331).
And add unit test for asyncThrottle.

### DIFF
--- a/src/createAsyncStoragePersister/asyncThrottle.ts
+++ b/src/createAsyncStoragePersister/asyncThrottle.ts
@@ -1,0 +1,58 @@
+export interface AsyncThrottleOptions {
+  interval?: number
+  onError?: (error: unknown) => void
+}
+
+const noop = () => {
+  /* do nothing */
+}
+
+export function asyncThrottle<Args extends readonly unknown[]>(
+  func: (...args: Args) => Promise<void>,
+  { interval = 1000, onError = noop }: AsyncThrottleOptions = {}
+) {
+  if (typeof func !== 'function') throw new Error('argument is not function.')
+
+  let running = false
+  let lastTime = 0
+  let timeout: ReturnType<typeof setTimeout>
+  let currentArgs: Args | null = null
+
+  const execFunc = async () => {
+    if (currentArgs) {
+      const args = currentArgs
+      currentArgs = null
+      try {
+        running = true
+        await func(...args)
+      } catch (error) {
+        onError(error)
+      } finally {
+        lastTime = Date.now() // this line must after 'func' executed to avoid two 'func' running in concurrent.
+        running = false
+      }
+    }
+  }
+
+  const delayFunc = async () => {
+    clearTimeout(timeout)
+    timeout = setTimeout(() => {
+      if (running) {
+        delayFunc() // Will come here when 'func' execution time is greater than the interval.
+      } else {
+        execFunc()
+      }
+    }, interval)
+  }
+
+  return (...args: Args) => {
+    currentArgs = args
+
+    const tooSoon = Date.now() - lastTime < interval
+    if (running || tooSoon) {
+      delayFunc()
+    } else {
+      execFunc()
+    }
+  }
+}
--- a/src/createAsyncStoragePersister/index.ts
+++ b/src/createAsyncStoragePersister/index.ts
@@ -1,4 +1,5 @@
 import { PersistedClient, Persister } from '../persistQueryClient'
+import { asyncThrottle } from './asyncThrottle'
 
 interface AsyncStorage {
   getItem: (key: string) => Promise<string | null>
@@ -49,44 +50,4 @@ export const createAsyncStoragePersister = ({
     },
     removeClient: () => storage.removeItem(key),
   }
-}
-
-function asyncThrottle<Args extends readonly unknown[], Result>(
-  func: (...args: Args) => Promise<Result>,
-  { interval = 1000, limit = 1 }: { interval?: number; limit?: number } = {}
-) {
-  if (typeof func !== 'function') throw new Error('argument is not function.')
-  const running = { current: false }
-  let lastTime = 0
-  let timeout: ReturnType<typeof setTimeout>
-  const queue: Array<Args> = []
-  return (...args: Args) =>
-    (async () => {
-      if (running.current) {
-        lastTime = Date.now()
-        if (queue.length > limit) {
-          queue.shift()
-        }
-
-        queue.push(args)
-        clearTimeout(timeout)
-      }
-      if (Date.now() - lastTime > interval) {
-        running.current = true
-        await func(...args)
-        lastTime = Date.now()
-        running.current = false
-      } else {
-        if (queue.length > 0) {
-          const lastArgs = queue[queue.length - 1]!
-          timeout = setTimeout(async () => {
-            if (!running.current) {
-              running.current = true
-              await func(...lastArgs)
-              running.current = false
-            }
-          }, interval)
-        }
-      }
-    })()
 }
--- a/src/createAsyncStoragePersister/tests/asyncThrottle.test.ts
+++ b/src/createAsyncStoragePersister/tests/asyncThrottle.test.ts
@@ -1,0 +1,127 @@
+import { asyncThrottle } from '../asyncThrottle'
+import { sleep as delay } from '../../reactjs/tests/utils'
+
+describe('asyncThrottle', () => {
+  test('basic', async () => {
+    const interval = 10
+    const execTimeStamps: number[] = []
+    const mockFunc = jest.fn(
+      async (id: number, complete?: (value?: unknown) => void) => {
+        await delay(1)
+        execTimeStamps.push(Date.now())
+        if (complete) {
+          complete(id)
+        }
+      }
+    )
+    const testFunc = asyncThrottle(mockFunc, { interval })
+
+    testFunc(1)
+    await delay(1)
+    testFunc(2)
+    await delay(1)
+    await new Promise(resolve => testFunc(3, resolve))
+
+    expect(mockFunc).toBeCalledTimes(2)
+    expect(mockFunc.mock.calls[1]?.[0]).toBe(3)
+    expect(execTimeStamps.length).toBe(2)
+    expect(execTimeStamps[1]! - execTimeStamps[0]!).toBeGreaterThanOrEqual(
+      interval
+    )
+  })
+
+  test('Bug #3331 case 1: Special timing', async () => {
+    const interval = 1000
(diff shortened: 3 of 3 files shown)


TkDodo (approved):
thank you 🙏

src/createAsyncStoragePersister/asyncThrottle.ts
```diff
@@ -0,0 +1,51 @@
...
+  const execFunc = async () => {
+    if (currentArgs) {
+      const args = currentArgs
+      currentArgs = null
+      try {
+        running = true
+        await func(...args)
+      } catch (error) {
+        getLogger().error(error)
+      } finally {
```
TkDodo, line 23: one small thing: in v4, we put all loggings behind a production env check, like here:

https://github.com/tannerlinsley/react-query/blob/fe98d533cc62fab8ab88166a31707d6cfc02d058/src/core/query.ts#L449-L451

can we do that here as well please

TkDodo: oh but actually, I don't think we should be logging here on this level. We have [another PR](https://github.com/tannerlinsley/react-query/pull/3246) where we removed the global logger and made it an option on the queryClient instead, so this would be problematic.

If we want to catch errors and log them, how about doing it in the `persistQueryClient` itself, when the writing happens? I think this would be somewhere around here:

https://github.com/tannerlinsley/react-query/blob/1abb3714637c3f9044a4294cc737cc41966a2f49/src/persistQueryClient/index.ts#L114

Not sure if it makes sense to just swallow and log the error though. Shouldn't we at least re-throw it to the consumers?

huanguolin:
@TkDodo 
I totally agree with you.
The asyncThrottle has to catch errors to avoid state corruption (like running). I currently have asyncThrottle provide an onError callback.
It's right to handle errors in `persistQueryClient`, but I'm not sure how it's appropriate to do that at the moment: 

- Is it to use a user-uniformly configured error handling callback from `queryClient`?
- Or do I modify the createAsyncStoragePersister interface?
- Or split this to another pr, to be handled in the future?

TkDodo:
We can do another PR and think about it. I don't think errors were handled in the previous version either

huanguolin:
@TkDodo 
OK!
The current pull request has been updated, we can now move on...

src/createAsyncStoragePersister/asyncThrottle.ts
```diff
@@ -0,0 +1,56 @@
...
+  const execFunc = async () => {
+    if (currentArgs) {
+      const args = currentArgs
+      currentArgs = null
+      try {
+        running = true
+        await func(...args)
+      } catch (error) {
+        onError(error)
+      } finally {
```
TkDodo, line 28: hm, I don't fully understand why we would need the `onError` callback. If remove the `catch` statement here and just let the error bubble up to the call side, what would happen?

huanguolin: Remove `catch` directly, keep `try...finally` is fine for `asyncThrottle`. But then error handling must be handled internally by the `func` itself. `asyncThrottle` returns a non-waitable function.

If we decide to handle errors inside the `func`, I'll change it right away.

TkDodo: i think it's expected that if `asyncThrottle` returns a `Promise` that it can fail

huanguolin: Sorry, I wasn't clear.

If the `onError` callback is retained, the error handling will look like this：
```typescript
return {
    persistClient: asyncThrottle(
      (persistedClient: PersistedClient) => storage.setItem(key, serialize(persistedClient)),
      { 
        interval: throttleTime,
        onError: (err: unknown) => {
          // handing error here
        } 
      }
    ),
// ....
```

If the `onError` callback is removed directly：
```typescript
return {
    persistClient: asyncThrottle(
      async (persistedClient: PersistedClient) => {
        try {
          storage.setItem(key, serialize(persistedClient))
        } catch (error) {
          // handing error here
        }
      }, 
      throttleTime
    ),
// ....
```



> i think it's expected that if `asyncThrottle` returns a `Promise` that it can fail

`asyncThrottle` returns a non-waitable function. Do you mean to return a function that returns a `promise`?
If so, this is strange. `asyncThrottle` returns a function that may be called 10 times， but the 'func' may be executed only twice. `promise` either resolves or rejects, and cannot be left dangling.
And the code will be more complex than now, I still think the previous two better a little , do you think?
☺

TkDodo:
getting a lint error now:

```
/home/runner/work/react-query/react-query/src/createAsyncStoragePersister/asyncThrottle.ts
Error:   6:20  error  Unexpected empty arrow function  @typescript-eslint/no-empty-function
```

src/createAsyncStoragePersister/asyncThrottle.ts
TkDodo: I see, my mistake. let's keep the callback then 👍

huanguolin:
Fixed. And improve unit test coverage.

TkDodo (approved):

TkDodo:
prettier fails now:

```
Checking formatting...
[warn] src/createAsyncStoragePersister/asyncThrottle.ts
[warn] src/createAsyncStoragePersister/tests/asyncThrottle.test.ts
```

huanguolin:
@TkDodo 
I'm sorry for that.
I had problems with my local environment, I changed a lot of version numbers in package.json, and the configuration of `.eslintrc`, and I was barely able to run eslint. But I was still prompted with some errors, but not caused by my commit this time (I didn't commit the code just to make eslint work). 
image

I've fixed the formatting problem. Hope everything goes well this time. No more wasting your time.

PR #3367, opened 2022-03-04
fix(mem): Stop leak from cacheTime outliving Query given WeakRef support (v3)

klarstrup:
This change fixes the issue described in [SSR: High memory consumption on server](https://react-query.tanstack.com/guides/ssr#high-memory-consumption-on-server) in environments with `WeakRef` support (most importantly Node 14.6+). I believe this will allow entire `QueryClient` instances to be garbage collected promptly now, even with `cacheTime`s much longer than the lifetime of the `QueryClient` instance.

I've included a couple of skipped tests that require being run with `--expose-gc` to reproduce the problem and confirm the fix, I didn't want to mess with y'all test setup but there they are.

Cheers
_Io_

### DIFF
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -7,6 +7,7 @@ import {
   replaceEqualDeep,
   timeUntilStale,
   ensureQueryKeyArray,
+  CompatWeakRef,
 } from './utils'
 import type {
   InitialDataFunction,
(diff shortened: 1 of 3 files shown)


TkDodo:
superseded by:
- https://github.com/tannerlinsley/react-query/pull/3377

PR #3370, opened 2022-03-04
fix(mem): Stop leak from cacheTime outliving Query given WeakRef support (v4)

klarstrup:
Like #3367 but for v4


TkDodo:
superseded by:
- https://github.com/tannerlinsley/react-query/pull/3377

PR #3377, opened 2022-03-09
feat(cacheTime): default cacheTime to Infinity for SSR

JacobLey:
Cache persists for the lifecycle of request and can be immediately GCed afterward

Helps close Node process immediately after use

https://github.com/tannerlinsley/react-query/blob/d8c46ce30a646f55651accbbfd551236ac483768/docs/src/pages/guides/testing.md#set-cachetime-to-infinity-with-jest
https://react-query.tanstack.com/guides/ssr#high-memory-consumption-on-server

These caveats may no longer be necessary if done by default? Although might be worth keeping in places where `cacheTime` is explicitly passed.

Related discussion of whether this is a 100% QoL improvement https://github.com/tannerlinsley/react-query/discussions/3373


klarstrup:
It's funny, I imagined this would make https://github.com/tannerlinsley/react-query/blob/3981f5219a1e3d8f48d13108b55673dcaea46319/src/core/query.ts#L203-L211 memory leak forever, but  `setTimeout(fn, Infinity)` actually makes it run immediately.

That has a different consequence though, that secondary renders server-side will be in a loading state as everything is invalidated immediately. (For example in a Next.js setup using react-ssr-prepass to prefill the RQ cache, dehydrating it and letting Next.js render the full page with a populated cache with passing a dehydrated one to clients).

My proposal for a fix to https://react-query.tanstack.com/guides/ssr#high-memory-consumption-on-server and https://github.com/tannerlinsley/react-query/blob/master/docs/src/pages/guides/testing.md#set-cachetime-to-infinity-with-jest is to instead set up the timeout with a weak reference such that an orphaned QueryClient/QueryCache instance simply gets garbage collected(see #3367)

JacobLey:
`setTimeout(fn, Infinity)` does actually work the way you described (counter intuitively) but in React Query that is not being called at all.

[isValidTimeout](https://github.com/tannerlinsley/react-query/blob/master/src/core/utils.ts#L91) returns false for `Infinity`. The result is there is no timeout and no manual garbage collection, and instead handled purely via Javascript's default garbage collection at the end of a request.

Even if the query itself were garbage collected early (via something like `WeakRef` or `WeakMap`) the `setTimeout` would still keep the process open until completion

> as everything is invalidated immediately
Nothing is ever invalidated on SSR. So _if_ you were to re-use a query client, it would be pre-filled. But..

https://react-query.tanstack.com/guides/ssr#using-hydration
> Create a new QueryClient instance for each page request. This ensures that data is not shared between users and requests.

Re-using a Query Client is actively discouraged

klarstrup:
Reusing a query client within a request is necessary though, if you need to render more than once for whatever reason.

Good shout about the `isValidTimeout` stuff though I hadn't thought that through.

JacobLey:
> Reusing a query client within a request is necessary though

Agreed, and that should be handled natively because the cache never preemptively expires. Re-usage _across_ multiple requests is discouraged.

TkDodo:
I'm unsure which of the two PRs (infinite staleTime or weakRef) is the better fix for the problem. @Ephem ?

Ephem:
Good question, possibly both?

I think in the distant past, we simply did not do `setTimeout`s at all on the server, with an explicit server-check. `Infinite` `cacheTime` accomplishes that and avoids scheduling a `setTimeout´ at all, which is good because it lets processes/tests close immediately instead of having to rely on `.clear()`. As @JacobLey points out, this is a good default _because_ we are actively discouraging sharing clients between requests. I do think it would be a breaking change though.

There _are_ valid usecases where you share clients between requests though (if you know what you are doing), and in those cases it's important to have a `cacheTime` so the query cache doesn't grow forever. You can still configure this manually though, and since this is the advanced case I think it makes sense to flip the defaults for better DX in the standard case. (It's pretty dangerous and hard to debug if you forget to set a `cacheTime` in those cases though, so we need to document that properly).

That said, in environments where `WeakRef` is available, it's still nice that queries can get GCd immediately if nothing outside is keeping a reference to them. So if a QueryClient no longer has a reference to it (maybe a subapp was unloaded), we don't have to wait until the timeout before the queries can be GCd (though the timeout is still scheduled, so the optimal thing would be to call `client.clear()`, but meh). That change also isn't a breaking one.

Ultimately though, maybe neither of these are the perfect solution? Maybe we should have a setTimeout on a much shorter timeout (maybe run and re-schedule itself every second, do cache cleanup every 10 or something) at the top that keeps track of queries to GC instead? With a WeakRef, this wouldn't re-schedule itself if the client had been GCd. This would let us keep a 5 minute `cacheTime` as default even on the server, but only if there were still references to the client and if not, the entire cache would be GCd in short order and the process could close?

JacobLey:
> Maybe we should have a setTimeout on a much shorter timeout (maybe run and re-schedule itself every second, do cache cleanup every 10 or something) at the top that keeps track of queries to GC instead?

Apologies, I'm not a master of the internals of React Query, so I want to make sure I understand something. (I guess this applies to existing functionality as well as proposed)

Lets say I have a server-side usage of React Query, and create a client to prefetch a couple queries.

```ts
export const getServerSideProps = async () => {

    await queryClient.prefetch('my-key', () => 'my-cached-data');
    await queryClient.prefetch('my-slow-key', async () => {
        await sleep(11000); // longer than default 10 second cache;
        return 'my-slow-data';
    });

    return {
        props: {
            dehydratedState: dehydrate(queryClient),
        },
    };
};
```
Would 
```ts
    useQuery('my-key', () => 'THIS SHOULD NOT RUN ON SERVER');
```
be cached properly? Or cleared before render? As there is a period where queryClient is not "in use", and also past the set cacheTime? 

Perhaps naive of me, but I would _assume_ that the `my-key` query remains cached on the server for lifetime of request.

Sorry this question is a bit off topic for this specific PR, happy to move to a separate issue if desired.

At first glance, my opinion is that any custom garbage collection goes against _recommended_ server side usage, and continues to open up a can of worms of potential memory bloat and hanging processes. So if a user _explicitly_ opts into manual server-side caching, the usage of WeakRef can help alleviate those pains.

I like the idea of an "actual cache clear timeout" built up of many smaller `setTimeout` with `WeakRef` checks in-between as a safety net, I just don't feel it is the sensible default

TkDodo:
we never garbage collect a query that is currently fetching - we schedule for gc after it is done fetching. So that would mean if cacheTime is 10 seconds, and your fetch takes 11 seconds, we try to remove it after 10 seconds, which we cannot. Then, after 11 seconds, we schedule another gc, which will then remove it after an additional 10 seconds - so in total after 21 seconds. Hope that makes sense.

JacobLey:
> we try to remove it after 10 seconds, which we cannot

What about the first `'my-key'` query? Which stopped fetching 10 seconds ago and is just waiting to be used?

Agreed about `'my-slow-key'` as it is still "in use". Although what if it took 9.9 seconds to resolve? And the 10 second timeout just happened to occur right between end of request and actual page render.

I'm aware this is an edge case, and a request that takes ~10 seconds sucks for other reason. Mostly trying to make a case for server-side manual GC being non-ideal as a default.

TkDodo:
> What about the first 'my-key' query? Which stopped fetching 10 seconds ago and is just waiting to be used?

oh, you are absolutely right. if we `await` another query that takes a long time before dehydrating the whole store, it might be garbage collected.

So with that in mind, I would think a good course of action would be to default to `cacheTime: Infinity` on the server, but since it is technically breaking, we should do that for v4 (which is just around the corner - on the `alpha` branch).

If users still want to share data on the server, which is rare, they could still set a different cacheTime.

JacobLey:
Took a swing at docs, let me know if any phrasing needs to be adjusted

TkDodo (approved):

PR #3389, opened 2022-03-13
docs(example): algolia

arnaudbzn:
New example with Algolia as a fetcher.


TkDodo:
Hey @arnaudbzn 

I like this example, but I think a lot has changed. E.g. we're now using `vite` for all examples (except nextjs) exclusively.

I think it would be easier to open a new PR

PR #3391, opened 2022-03-13
fix executeMutation: on error cache gets updated before promise is ev…

Toyotomi-clan:
**Desired result: when mutation fails and error boundary is set to true, I want to see the error boundary being triggered.**

Example
```
export function useMutationSignUp(....){
  return useMutation(args) =>{
    return postSignUpForm(args);
  },{
    retry: false,
    useErrorBoundary: true,
....
```

**Issue**

Currently, when an error is triggered it gets [dispatched](https://github.com/tannerlinsley/react-query/blob/master/src/core/mutation.ts#L225) before it is thrown and before the promise is rejected by the retryer. I do lack the domain knowledge to explain why this is a problem, however it is definitely not creating the desired behaviour.

Take for example an application, that has these defaults shown above (`  useErrorBoundary: true,  retry: false,`). Once a post fails i.e. returns a failed request, the expect behaviour would been to throw to the error boundary immediately since retry is also set to false. However, this is not the case. Here is an example application to demonstrate this current behaviour.

before-fix

Here is the current code base for this application specific [sign up page](https://github.com/Toyotomi-clan/borngladiator/blob/master/apps/born/src/app/signup.tsx) (sorry for the mess)

Initially, I thought no dispatch was being called. This turns out to be false. In line [225](https://github.com/tannerlinsley/react-query/blob/master/src/core/mutation.ts#L225) a dispatch gets called. However, as stated above this happens before the error is thrown or promise is rejected.  Having executeMutation implement error allows the [retryer to raise the error](https://github.com/tannerlinsley/react-query/blob/master/src/core/retryer.ts#L111). 

Doing this results in the desired behaviour.

Ultimately, I am not sure why rising a dispatch at the retryer level results in the desired behaviour while doing so before a throw does not. I welcome any feedback into this.

Here is the working example of the error raising and propagating to the error boundary with the changes implemented. 

after-fix

### DIFF
--- a/src/core/mutation.ts
+++ b/src/core/mutation.ts
@@ -190,7 +190,6 @@ export class Mutation<
         )
       )
       .then(() => {
-        this.dispatch({ type: 'success', data })
         return data
       })
       .catch(error => {
@@ -222,7 +221,6 @@ export class Mutation<
             )
           )
           .then(() => {
-            this.dispatch({ type: 'error', error })
             throw error
           })
       })
@@ -235,6 +233,13 @@ export class Mutation<
           return Promise.reject('No mutationFn found')
         }
         return this.options.mutationFn(this.state.variables!)
+      },
+
+      onSuccess: (data) => {
+        this.dispatch({ type: 'success', data })
+      },
+      onError: (error) =>{
+        this.dispatch({ type: 'error', error: error })
       },
       onFail: () => {
         this.dispatch({ type: 'failed' })
--- a/src/core/mutation.ts
+++ b/src/core/mutation.ts
@@ -190,6 +190,7 @@ export class Mutation<
         )
       )
       .then(() => {
+        this.dispatch({ type: 'success', data })
         return data
       })
       .catch(error => {
@@ -235,9 +236,6 @@ export class Mutation<
         return this.options.mutationFn(this.state.variables!)
       },
 
-      onSuccess: (data) => {
-        this.dispatch({ type: 'success', data })
-      },
       onError: (error) =>{
         this.dispatch({ type: 'error', error: error })
       },
--- a/src/core/tests/mutations.test.tsx
+++ b/src/core/tests/mutations.test.tsx
@@ -78,25 +78,6 @@ describe('mutations', () => {
     expect(cache?.state.error).toEqual(error)
   })
 
-  test('executeMutation should propagate success to the mutation cache after retrier resolves', async () => {
-    const key = queryKey()
-
-    const data = "hello world";
-    queryClient.setMutationDefaults(key, {
-      mutationFn: async () => { return  data},
-      retry: false
-    })
-
-    await queryClient.executeMutation({
-        variables: 'todo',
-        mutationKey: key,
-      })
-
-    const cache = queryClient.getMutationCache().getAll().find(x => x.state.data);
-
-    expect(cache?.state.data).toEqual(data)
-  })
-
   test('mutation should set correct success states', async () => {
     const mutation = new MutationObserver(queryClient, {
       mutationFn: async (text: string) => {
--- a/src/core/tests/mutations.test.tsx
+++ b/src/core/tests/mutations.test.tsx
@@ -56,6 +56,47 @@ describe('mutations', () => {
     expect(result).toBe(result)
   })
 
+  test('executeMutation should propagate error to the mutation cache after retrier rejects', async () => {
+    const key = queryKey()
+
+    const error = new Error("bad error / network issue");
+    queryClient.setMutationDefaults(key, {
+      mutationFn: async () => { throw error},
+      retry: false
+    })
+
+    try{
+      await queryClient.executeMutation({
+        variables: 'todo',
+        mutationKey: key,
+      })
+    }
+    catch (e){ expect(e).toEqual(error);}
+
+    const cache = queryClient.getMutationCache().getAll().find(x => x.state.error);
+
+    expect(cache?.state.error).toEqual(error)
+  })
+
+  test('executeMutation should propagate success to the mutation cache after retrier resolves', async () => {
+    const key = queryKey()
+
+    const data = "hello world";
+    queryClient.setMutationDefaults(key, {
+      mutationFn: async () => { return  data},
+      retry: false
+    })
+
+    await queryClient.executeMutation({
+        variables: 'todo',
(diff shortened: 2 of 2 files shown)


TkDodo:
does this supersede:
- https://github.com/tannerlinsley/react-query/pull/3385

?

---

FYI, you have prettier errors:

```
Checking formatting...
[warn] src/core/mutation.ts
[warn] src/core/tests/mutations.test.tsx
error Command failed with exit code 1.
```

src/core/mutation.ts
```diff
@@ -190,7 +190,6 @@ export class Mutation<
         )
       )
       .then(() => {
-        this.dispatch({ type: 'success', data })
```
TkDodo, line 193: one problem I'm seeing with this change is that with mutations, we are intentionally `awaiting` the Promises returned from `onSuccess`, `onError` and `onSettled`, and only after that, we go dispatch success / error state. With your change, I think we now go into that state too early. It could also be that the promise returned from `onSuccess` fails, which should put the query into `error` state, but that's not the case now anymore.

Please also have a look at the `alpha` branch, where I've refactored this code to async/await:

https://github.com/tannerlinsley/react-query/blob/f42689951de3ef5c56150bcc90fffb0cdd54bbd5/src/core/mutation.ts#L190-L263

Toyotomi-clan:
@TkDodo it does, I will abandon that pr. I need to flash out this prs description a bit more (accidently triggered a pr before preview). I will tag you once it is ready 👯.

src/core/mutation.ts
Toyotomi-clan: The all the then chaining gave me a hard time not going to lie 😢 

I can revert that and leave the dispatch for success on line [193](https://github.com/tannerlinsley/react-query/blob/master/src/core/mutation.ts#L193).

The root issue of this is the error part. The error gets dispatched on line [225](https://github.com/tannerlinsley/react-query/blob/master/src/core/mutation.ts#L225) but because it is not thrown and the promise is not rejected yet it puts as in a unknown state (user defaults such as useErrorBoundary is not trigger)

Give me just a moment I will clip all this in the description.

Toyotomi-clan: @TkDodo undone the change :D but still removed the dispatch at the [throw](https://github.com/tannerlinsley/react-query/blob/master/src/core/mutation.ts#L225) as I don't think that holds any value since when it is rejected dispatch gets called again. 

Let me know if that is incorrect / needs to be changed.

TkDodo:
formatting is still off! Tests won't run until you fix this please.

src/core/tests/mutations.test.tsx
```diff
@@ -56,6 +56,28 @@ describe('mutations', () => {
...
+        variables: 'todo',
+        mutationKey: key,
+      })
+    }
+    catch (e){ expect(e).toEqual(error);}
+
+    const cache = queryClient.getMutationCache().getAll().find(x => x.state.error);
+
+    expect(cache?.state.error).toEqual(error)
+  })
```
TkDodo, line 79: I tried this test on `master` and it's also green there. I'm still no sure what exactly the problem is we're trying to solve here, but it's not documented with this test. We'd need a test that _fails_ on `master` and _works_ on this PR please.

src/core/tests/mutations.test.tsx
```diff
@@ -364,7 +386,7 @@ describe('mutations', () => {
     consoleMock.mockRestore()
   })
 
-  test('cancel mutation should not call mutationFn if the current retrier is undefined', async () => {
+  test('cancel mutation should not call mutationFn if the current retier is undefined', async () => {
```
TkDodo, line 389: This looks like an accidental change?

```suggestion
  test('cancel mutation should not call mutationFn if the current retrier is undefined', async () => {
```

TkDodo (changes requested):

src/core/tests/mutations.test.tsx
Toyotomi-clan: Tried to spell, turns out I can't, I will revert it.

src/core/tests/mutations.test.tsx
Toyotomi-clan: Sure, I can update it to fail on master.

Here is my reasoning behind why I did this:

The reason nothing fails on master is because nothing should have been broken. This test captures it.

You can see in line [225](https://github.com/tannerlinsley/react-query/pull/3391/files#diff-9c4fdb8c2806ce1f8b1e9a48d82639ac533c39eecc9ef1f1c6cd18562642634bL225) has been removed. This would cause this test to fail.

**Current branch pass**
It passes because onError is implemented on executeMutation which triggers a dispatch when the promise is rejected by the retyer.

**Master branch pass**
It passes in master because master implements line  [225](https://github.com/tannerlinsley/react-query/pull/3391/files#diff-9c4fdb8c2806ce1f8b1e9a48d82639ac533c39eecc9ef1f1c6cd18562642634bL225) which by defaults triggers a dispatch.

**Why do this then?**

This I am not 100% sure of, but triggering a dispatch at the retryer level allows us to hit the error boundary compared to the earlier dispatch that happens before the error is thrown  [225](https://github.com/tannerlinsley/react-query/pull/3391/files#diff-9c4fdb8c2806ce1f8b1e9a48d82639ac533c39eecc9ef1f1c6cd18562642634bL225) .


**Improvement I can make**

If I test the retryer's resolve error gets called, would ensure a fail on master.  I can do that, just don't know how to for now xD.

I will try to re-write the test to fail on master by implementing the above test (see if config.onError gets called)

I hope that makes sense, let me know if there is anything else.

TkDodo: if this is about error boundaries, please create a test case that tries to show an error boundary, as that should fail then. We already have tests for error boundaries.

Toyotomi-clan:
I spent hours trying to implement a test that fails on master. It turns out I screwed up. 
For anyone having similar issues. 

Please **AWAIT** your promise.

the entire issue was caused by the onError being called, which then utilised an external tool which returns a promise, whose promise is not awaited.

Thus nothing was getting propagated.

I am very happy to resolved this. Thank you @TkDodo  for you guidance.

TkDodo:
@Toyotomi-clan thanks for taking the time to look into this, and I am happy that the issue is not with react-query itself :)

PR #3410, opened 2022-03-16
Fix useIntersectionObserver bug by working with react-intersection-observer

lironezra:
Hi :)

I Found a bug on the example of infinite-scrolling in documentation, there was two fetch request on the first scroll because there is an issue with the `useIntersectionObserver.js` hook (the function `onIntersect` called twice and that is what cause for double fetching)
I was in touch with @TkDodo, he asked me to help him with the issue by using `react-intersection-observer` package instead of using the custom hook.

hope you will like it and approve my solution :)


TkDodo (commented):
looks good. can you please:
- remove `hooks/useIntersectionObserver` as we don't need it anymore
- remove the `package-lock` file, as we'd like the examples to be up-to-date :)

TkDodo (approved):

PR #3411, opened 2022-03-17
add test for `status`=`error`

KATT:
Related to https://github.com/trpc/trpc/pull/1645

### DIFF
--- a/src/react/tests/useQuery.test.tsx
+++ b/src/react/tests/useQuery.test.tsx
@@ -4677,4 +4677,38 @@ describe('useQuery', () => {
 
     consoleMock.mockRestore()
   })
+
+  it('it should have status=error on mount when a query has failed', async () => {
+    const consoleMock = mockConsoleError()
+    const key = queryKey()
+    const states: UseQueryResult<number>[] = []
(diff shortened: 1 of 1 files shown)


src/react/tests/useQuery.test.tsx
```diff
@@ -4677,4 +4677,38 @@ describe('useQuery', () => {
...
+    const error = new Error('oops')
+
+    const queryFn = async () => {
+      throw error
+    }
+
+    function Page() {
+      const state = useQuery(key, queryFn, {
+        retry: false,
+        retryOnMount: false,
```
KATT, line 4694: Cc @TkDodo what puzzles me here is that the `retryOnMount: false` is needed. Shouldn't `retry: false` override it?

TkDodo: it's a different setting. `retryOnMount` might be not ideally named:
-  `retry` retries internally when a query fails, and then it goes to `error` state after all retries are done.
- `retryOnMount` is more like `refetchOnMountIfWeHaveAnErrorButNoDataYet`

PR #3414, opened 2022-03-18
docs: updates for new react-query course

alexanderson1993:
Change removes convert kit for email capture and migrates it to Bytes. Copy updates to promote the new React Query Course are included. Also includes a PPP discount banner that provides coupon codes to folks.


TkDodo:
is this the same as:
- https://github.com/tannerlinsley/react-query/pull/3400

?

I'm confused 😅 . Also, should this go to the `alpha` branch so that we sync it with the v4 release?

alexanderson1993:
> is this the same as:
> 
> - https://github.com/tannerlinsley/react-query/pull/3400
> 
> 
> 
> ?
> 
> 
> 
> I'm confused 😅 . Also, should this go to the `alpha` branch so that we sync it with the v4 release?

There were issues syncing the remote branch, so this is the one that needs to be merged. That other one should be closed.

Also, this should go to the main branch as soon as the course launches. The React Query docs will be a primary traffic driver to the course. We're planning to launch the course as soon as possible, regardless of  whether v4 is out of alpha. The first lesson in the course will explain that it's covering the v4 alpha version - at least until the alpha launches.

Sorry, we've been discussing this with Tanner, but we should have looped you in sooner.

PR #3440, opened 2022-03-24
fix(devtools): serialize bigint types

### LINKED ISSUE #3082: React Query Devtools cannot serialize BigInt type
**Describe the bug**
One of the models that I am caching with React Query is a BigInt type. Unfortunately, when I try and inspect the data in the devtools explorer, I get an uncaught error: `TypeError: Do not know how to serialize a BigInt`.

**To Reproduce**
1. Add an object using a BigInt type to the query cache
2. Open the Devtools
3. Select the query in the browser to inspect the data
4. See the error

**Expected behavior**
I should be able to view properly serialised data in the devtools browser.

**Screenshots**
image

 - OS: Pop_OS 21.04
 - Browser: Chrome
 - Version: 94

**Additional context**
The error can be traced to this line: https://github.com/tannerlinsley/react-query/blob/master/src/devtools/Explorer.tsx#L112. If there's a reason that the data is being stringified, then hopefully there's a way to keep doing it, but there are data types which cannot be parsed by JSON.stringify. There may be other options, such as https://github.com/blitz-js/superjson but I understand that would add further complexity.

HenriqueDerosa:
closes #3082 

### Detail
I detected the issue `TypeError: Do not know how to serialize a BigInt` was caused by `JSON.stringify` in the devtools explorer. I created a separate function so I could also write a test. 

This is my first contribution here, so please, let me know how I can improve this implementation. 😄

### DIFF
--- a/src/devtools/tests/Explorer.test.tsx
+++ b/src/devtools/tests/Explorer.test.tsx
@@ -2,6 +2,7 @@ import React from 'react'
 import { fireEvent, render, screen } from '@testing-library/react'
 
 import { chunkArray, DefaultRenderer } from '../Explorer'
+import { displayValue } from '../utils'
 
 describe('Explorer', () => {
   describe('chunkArray', () => {
@@ -53,5 +54,9 @@ describe('Explorer', () => {
 
       expect(toggleExpanded).toHaveBeenCalledTimes(1)
     })
+  })
+
+  it('when the value is a BigInt', () => {
+    expect(displayValue(BigInt(1))).toBe('"1n"')
   })
 })
--- a/src/devtools/tests/Explorer.test.tsx
+++ b/src/devtools/tests/Explorer.test.tsx
@@ -56,7 +56,13 @@ describe('Explorer', () => {
     })
   })
 
-  it('when the value is a BigInt', () => {
-    expect(displayValue(BigInt(1))).toBe('"1n"')
+  describe('displayValue', () => {
+    it('when the value is a boolean', () => {
+      expect(displayValue(true)).toBe('true')
+    })
+
+    it('when the value is a BigInt', () => {
+      expect(displayValue(BigInt(1))).toBe('"1n"')
+    })
   })
 })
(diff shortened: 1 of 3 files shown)


src/devtools/utils.ts
```diff
@@ -123,6 +123,19 @@ export function useSafeState<T>(initialState: T): [T, (value: T) => void] {
   return [state, safeSetState]
 }
 
+/**
+ * Displays a string regardless the type of the data
+ * @param {any} value Value to be stringified
+ */
+ export const displayValue = (value: any) => {
```
TkDodo, line 130: let's please use `unknown` instead of `any`:

```suggestion
/**
 * Displays a string regardless the type of the data
 * @param {any} value Value to be stringified
 */
 export const displayValue = (value: unknown) => {
```

src/devtools/tests/Explorer.test.tsx
```diff
@@ -54,4 +55,8 @@ describe('Explorer', () => {
       expect(toggleExpanded).toHaveBeenCalledTimes(1)
     })
   })
+
+  it('when the value is a BigInt', () => {
+    expect(displayValue(BigInt(1))).toBe('"1n"')
+  })
```
TkDodo, line 61: please put this in it's own `describe('displayValue')` block, and add at least one test that also tests for non-big-int please.

src/devtools/utils.ts
```diff
@@ -123,6 +123,19 @@ export function useSafeState<T>(initialState: T): [T, (value: T) => void] {
...
+  let newValue = value
+  const name = Object.getOwnPropertyNames(Object(value))
+  if (typeof value === 'bigint') {
+    newValue = `${value.toString()}n`
+  }
+  return JSON.stringify(newValue, name)
```
TkDodo, line 136: how about:

```
const name = Object.getOwnPropertyNames(Object(value))
const newValue = typeof value === 'bigint' ? `${value.toString()}n` : value


return JSON.stringify(newValue, name)
```

TkDodo (commented):
thank you for this contribution - it looks good already 👍. Just some minor comments from my side

TkDodo:
also, prettier formatting is off:

```
Checking formatting...
[warn] src/devtools/utils.ts
error Command failed with exit code 1.
```

TkDodo (approved):

PR #3447, opened 2022-03-27
chore(eslint): Add react/jsx-key lint rule

Liam-Tait:
Rule is set to error.
option checkFragmentShorthand option set to true,  meaning the rule is used for shorthand fragment syntax too.
option will default to true next major eslint-plugin-react version

See rule docs [here](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-key.md)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -15,6 +15,7 @@
     "sourceType": "module"
   },
   "rules": {
+    "react/jsx-key": ["error", { "checkFragmentShorthand": true }],
     "@typescript-eslint/ban-types": "off",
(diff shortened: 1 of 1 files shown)


Liam-Tait:
This pr is expected to fail based on the new rule until https://github.com/tannerlinsley/react-query/pull/3447 is merged where rule violations are fixed

TkDodo:
please update with master to make the new rule pass

Liam-Tait:
@TkDodo rebased

PR #3451, opened 2022-03-29
Rename query data generics from TData to TQueryFnData for consistency

mattbrandlysonos:
Hey! Thank you for `react-query`. It makes my job a lot easier.

Recently we ran into some confusion about the expected return value from `getQueryData`. The existing return type is called `TData`:

https://github.com/tannerlinsley/react-query/blob/fc4671f0731d1dd757f9fd037fd8e6755b3393d1/src/core/queryClient.ts#L105-L110

However, looking at the type for `UseQueryOptions`, `TData` refers to the value after the `select` function has been applied. The type that's returned from `getQueryData` is called `TQueryFnData` in this context.

https://github.com/tannerlinsley/react-query/blob/fc4671f0731d1dd757f9fd037fd8e6755b3393d1/src/react/types.ts#L26-L31

This PR presents no changes in functionality, but it aligns the generics names between `UseQueryOptions` and `getQueryData`! Now when working with `UseQueryOptions`, the names will help indicate the value returned from `getQueryData`.

To illustrate further, you might write code like this if you're expecting `getQueryData` to return `TData`, but the types here are incorrect!

```tsx
type QueryOptions = UseQueryOptions<A, B, C>
type ExtractTData<P> = P extends UseQueryOptions<any, any, infer TData> ? TData : never;
type TData = ExtractTData<QueryOptions>

// ...

// `value` is not actually of type `TData`!
const value = queryClient.getQueryData<TData>(queryKey)
```

### DIFF
--- a/src/core/queryClient.ts
+++ b/src/core/queryClient.ts
@@ -102,31 +102,35 @@ export class QueryClient {
     return this.mutationCache.findAll({ ...filters, fetching: true }).length
   }
 
-  getQueryData<TData = unknown>(
+  getQueryData<TQueryFnData = unknown>(
     queryKey: QueryKey,
     filters?: QueryFilters
-  ): TData | undefined {
-    return this.queryCache.find<TData>(queryKey, filters)?.state.data
+  ): TQueryFnData | undefined {
+    return this.queryCache.find<TQueryFnData>(queryKey, filters)?.state.data
   }
 
-  getQueriesData<TData = unknown>(queryKey: QueryKey): [QueryKey, TData][]
-  getQueriesData<TData = unknown>(filters: QueryFilters): [QueryKey, TData][]
-  getQueriesData<TData = unknown>(
+  getQueriesData<TQueryFnData = unknown>(
+    queryKey: QueryKey
+  ): [QueryKey, TQueryFnData][]
+  getQueriesData<TQueryFnData = unknown>(
+    filters: QueryFilters
+  ): [QueryKey, TQueryFnData][]
+  getQueriesData<TQueryFnData = unknown>(
     queryKeyOrFilters: QueryKey | QueryFilters
-  ): [QueryKey, TData][] {
+  ): [QueryKey, TQueryFnData][] {
     return this.getQueryCache()
       .findAll(queryKeyOrFilters)
       .map(({ queryKey, state }) => {
-        const data = state.data as TData
+        const data = state.data as TQueryFnData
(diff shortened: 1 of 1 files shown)


TkDodo:
yeah, I think I'm fine with the change :)

src/core/queryClient.ts
```diff
@@ -102,70 +102,74 @@ export class QueryClient {
...
+    return this.queryCache.find<TQueryFnData>(queryKey, filters)?.state.data
   }
 
-  getQueriesData<TData = unknown>(queryKey: QueryKey): [QueryKey, TData][]
-  getQueriesData<TData = unknown>(filters: QueryFilters): [QueryKey, TData][]
-  getQueriesData<TData = unknown>(
+  getQueriesData<TQueryFnData = unknown>(
+    queryKey: QueryKey
+  ): [QueryKey, TQueryFnData][]
+  getQueriesData<TQueryFnData = unknown>(
```
TkDodo, line 115: we also have some references to `TData` on the docs page here:

https://react-query.tanstack.com/reference/QueryClient#queryclientgetqueriesdata

would we also need to change that?

src/core/queryClient.ts
```diff
@@ -102,70 +102,74 @@ export class QueryClient {
...
     queryKeyOrFilters: QueryKey | QueryFilters
-  ): [QueryKey, TData][] {
+  ): [QueryKey, TQueryFnData][] {
     return this.getQueryCache()
       .findAll(queryKeyOrFilters)
       .map(({ queryKey, state }) => {
-        const data = state.data as TData
+        const data = state.data as TQueryFnData
```
TkDodo, line 124: looking at these types and the type assertion, I actually think the types are wrong here. If you have a query that is in `error` state, or that hasn't run yet (e.g. because it is `disabled`), and you find it with `getQueriesData`, you will get `undefined` back. I think we need to at least union the return type with undefined:

```diff
- [QueryKey, TQueryFnData][]
+ [QueryKey, TQueryFnData | undefined][]
```

This would be in line to what `getQueryData`. What do you think? Should we add this here as a fix?

TkDodo:
Any updates regarding my comments here? Also, there are conflicts now.

TkDodo:

SammRosenthal:
If you end up closing this I can take the current work and resolve your comments.

TkDodo:
@SammRosenthal yes please go ahead and create a new PR, thank you 🙏

PR #3464, opened 2022-04-02
fix(persistQueryClient): add a failing test for persisting mutations

### LINKED ISSUE #3460: Paused Queries/Mutations are not persisted
### Describe the bug

I'm fairly new to React query and first of all I want to appreciate sharing such a cool library with the community.
I'm trying to make use of new offline capabilities introduced in v4. I'm facing some problems regarding queries/mutations' dehydration in localStorage which is reproduced [here](https://codesandbox.io/s/usemutation-demo-forked-fluu13). If I set my connection to offline (using React query's built-in offline toggle), I expect the queries/mutations that are now in `paused` state to be persisted in local storage so I'll not lose them the user closes the window and then goes online. But in my code, the localStorage item is not created at all.  

I tried updating an official demo to use the latest version of react query. The result is available [here](https://codesandbox.io/s/tannerlinsley-react-query-forked-5x7482). This time the localStorage is created but is not populated with paused queries when I go offline. 

- Regarding my first case, the localStorage item is not even created. I suspect the storage is initialised only when `useQuery` is involved and in my sample, I only used mutations.
- in my second sample, I believe the correct behaviour should be to have the paused queries in localStorage with a maximum of 1 second delay (based on what I read about `createWebStoragePersister`).

I have tried to use `<PersistQueryClientProvider>` as well as `persistQueryClient` method with `<QueryClientProvider>`. Both led to the same result.

Of course, this might not be a bug and I might have misconfigured the library and in that case, I'd greatly appreciate if you let me know about the correct way.

https://codesandbox.io/s/usemutation-demo-forked-fluu13

### Steps to reproduce

- Open the sample
- Go offline using react query's dev tools toggle
- Make a query/mutation and watch `REACT_QUERY_OFFLINE_CACHE` localStorage changes

### Expected behavior

When using `createWebStoragePersister` as the persister, React Query should dehydrate mutations/queries in paused state every 1 second.

_No response_

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: 99.0.4844.83

### react-query version

4.0.0-alpha.23

### TypeScript version

4.1.3

### Additional context

_No response_

TkDodo:
fixes #3460

### DIFF
--- a/docs/src/pages/guides/mutations.md
+++ b/docs/src/pages/guides/mutations.md
@@ -195,7 +195,7 @@ useMutation(addTodo, {
   mutate(todo, {
     onSuccess: (data, error, variables, context) => {
       // Will execute only once, for the last mutation (Todo 3),
-      // regardless which mutation resolves first 
+      // regardless which mutation resolves first
     },
   })
 })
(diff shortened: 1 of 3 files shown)


TkDodo:
@KamyarP your example works "better" now with this PR. the local storage entry is created, but upon reloading the page, it errors with `no mutationFn`, because you need to provide a default mutation function for this to work. I've also updated the documentation accordingly.

PR #3468, opened 2022-04-04
fix: change devtools env condition

dngwoodo:
We basically use process.env.NODE_ENV as 'localhost', so devtools does not work.

So I changed the devtools conditional statement. Please check it out.


TkDodo:
this will also add devtools to testing environments where people might not expect it. or, if someone is building their production app with a special NODE_ENV, the devtools will not be split.

if you don't use DEVELOPMENT as NODE_ENV, you can also import the `Devtools` directly:

```
import { ReactQueryDevtools } from 'react-query/devtools/development'
```

PR #3480, opened 2022-04-08
fix: 🐛 make types @types/react@18 compatible

Andrey-Bazhanov:
In react 18 types children was removed from FC type.

Changes of this PR are compatible with 17 and 18 typings versions.

### DIFF
--- a/src/react/Hydrate.tsx
+++ b/src/react/Hydrate.tsx
@@ -25,7 +25,7 @@ export interface HydrateProps {
   options?: HydrateOptions
 }
 
-export const Hydrate: React.FC<HydrateProps> = ({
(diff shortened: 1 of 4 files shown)


TkDodo:
thank you for the PR, but please see: https://github.com/tannerlinsley/react-query/issues/3476#issuecomment-1092501739

I don't think we should update the v3 typings to work with react18, as v3 explicitly needs v16 or v17:

https://github.com/tannerlinsley/react-query/blob/b5675e9718d4b5ba7ee899bdf8bcea10003d0176/package.json#L72-L74

for v4-beta, the `QueryClientProvider` already does not use `React.FC`, and I also plan on making the changes to the other components, unless you want to?

PR #3481, opened 2022-04-08
fix: 🐛 make types @types/react@18 compatible

Andrey-Bazhanov:


Andrey-Bazhanov:
@TkDodo here we go again)

Any eta of v4 release?

TkDodo:
I wanted to achieve two things:

1) get rid of `React.FC` and similar things altogether and just type the props accordingly.
2) use react 18 typings locally. this has shown some other issues around typings that we might have with useQueries

I've created a PR to supersede this one:
- https://github.com/tannerlinsley/react-query/pull/3482

PR #3482, opened 2022-04-08
fix(types): use react 18 types

TkDodo:


src/persistQueryClient/tests/PersistQueryClientProvider.test.tsx
```diff
@@ -147,7 +147,7 @@ describe('PersistQueryClientProvider', () => {
 
       return (
         <div>
-          <h1>{state.data}</h1>
+          <h1>{String(state.data)}</h1>
```
TkDodo, line 150: @artysidorenko I think the react 18 type upgrades have shown an issue with the types of `useQueries`: Here, the type of `state.data` is inferred to `string | Promise<string> | undefined`, which cannot be rendered.

I will create a separate issue for this with a failing test case, just to let you know :)

TkDodo: follow-up issue: https://github.com/tannerlinsley/react-query/issues/3483

TkDodo:
> Any chance we could get a fix for 3.x as well?

Why do you want to use react18 with react-query 3.x? Imo this is just asking for troubles as soon as you start to use concurrent features. They will not work in all cases with v3. If you want to use react18, please use v4.

symbiont-matthew-pirocchi:
> Why do you want to use react18 with react-query 3.x? Imo this is just asking for troubles as soon as you start to use concurrent features. They will not work in all cases with v3. If you want to use react18, please use v4.

Thanks, I missed that 3.x isn't intended to work with react18.

PR #3490, opened 2022-04-09
fix: React implicit children type

maddhruv:
With this recent change in `@types/react` - https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210 the type check fails for `QueryClientProvider` and other components which have `children` as a prop.

This is a particular failure when using `^@types/react@18`. Adding the `children` type explicitly to these components.

### DIFF
--- a/docs/src/components/LayoutDocs.js
+++ b/docs/src/components/LayoutDocs.js
@@ -128,14 +128,19 @@ export const LayoutDocs = props => {
                       </div>
                       <div className="mt-12 relative">
                         <h4 className="font-semibold uppercase text-sm mb-2 mt-2 text-gray-500">
-                          Subscribe to Bytes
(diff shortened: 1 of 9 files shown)


maddhruv:
Closed since this is already being fixed here - https://github.com/tannerlinsley/react-query/pull/3482

clothoo:
> Closed since this is already being fixed here - #3482

Is this a mistake? As in #3482, there's no change to `QueryClientProviderProps`

clothoo:
My mistake. Reinstall the package the types got updated!

TkDodo:
> Is this a mistake? As in https://github.com/tannerlinsley/react-query/pull/3482, there's no change to QueryClientProviderProps

Usage of React.FC was removed in an earlier PR for v4, so there was no change necessary

Issue #3483, opened 2022-04-08
Types issue with useQueries


TkDodo:
useQueries type includes `Promise` in the inferred type of `.data`, which is not renderable as `React.ReactNode` with the new react 18 typings.

failing test on `beta`:

```
  it('infers returned Promises', async () => {
    const key1 = queryKey()

    // @ts-expect-error (Page component is not rendered)
    // eslint-disable-next-line
    function Page() {
      const result = useQueries({
        queries: [
          { queryKey: key1, queryFn: () => Promise.resolve(1)}
        ]
      })

      expectType<number | undefined>(result[0].data)
    }
  })
```

fails with:

```
TS2345: Argument of type 'number | Promise<number> | undefined' is not assignable to parameter of type 'number | undefined'.   Type 'Promise<number>' is not assignable to type 'number'.
```

@artysidorenko could you take a look at this please?

artysidorenko:
Heya, had a look, added a PR to fix. Wasn't so much react-18 related; previously it had been correctly getting the `Promise<string>` => `string` inference, but got tripped up by [this union added in a recent PR](https://github.com/tannerlinsley/react-query/pull/3271/files#diff-80ebcd3dc48e85c12b2c8484c5355662ab98ceea358660eb3af536c584f6fa44R16)

artysidorenko:
Btw, I noticed something else along the way. My understanding from reading [your other issue](https://github.com/tannerlinsley/react-query/issues/2925)  is that you want to warn with a type error anytime `queryFn` resolves to undefined?

I think at the moment (`beta` branch) it only works when it's synchronous `undefined` but not `Promise<undefined>`

Errors as expected
```typescript
// @ts-expect-error
useQuery(key, () => undefined)
```

This doesn't complain
```typescript
useQuery(key, async () => undefined)
```

Can look into it later unless that's intentional?

TkDodo:
> Happy to look into it later unless that's intentional

oh yes good catch.  Promise<void> should also error on v4. Thank you 🙏

TkDodo:
just to be sure, @artysidorenko, the `async () => undefined` issue wasn't fixed with #3493 right?

from what I can see:

- `useQuery(key, () => undefined)` errors, as expected ✅ 
- `useQuery(key, () => undefined)` no error ❌ 
- `useQueries({ queries: [{ queryKey: key, queryFn: () => undefined }]})` no error ❌ 
[...]

PR #3493, opened 2022-04-10
fix(types): fix promise inference in useQueries

artysidorenko:
fixes issue #3483 

This new helper `QueryFunctionData` was introduced in the latest branch:

```typescript
type QueryFunctionData<T> = T extends undefined ? never : T

type QueryFunction<
  T = unknown,
  TQueryKey extends QueryKey = QueryKey
> = (
  context: QueryFunctionContext<TQueryKey>
) => QueryFunctionData<T> | Promise<QueryFunctionData<T>>
```

Looks like when a conditional expression is duplicated like this in a union, it's not able to pass along any `infer` logic correctly.

[(TS Playground link with a simplified example - not restricted to promises though... you can replace `Promise<X>` with `Array<X>` for the same effect)](https://www.typescriptlang.org/play?#code/C4TwDgpgBAigrhATiAYnAdgY2ASwPboAiAhsMQDwAqAfFALxSVQQAewE6AJgM5QacQAZjnQROUAPxRRANyRQAXIwBQy0JCgB5OO0QBVdPnRVaDeElQZsRkmRNQAPlAAKiPAFsc3COXPI0WLgEthQ01Grg0ACS6IJIetwiAObauvYMTKzsXLyp8YYE5CJxiIx+ICG0UpTlIYrSEHKIERooxDgANvRQMSUJyXmI5K4eXj7cwIjJ1OGYBBNQgu0dShNT6EndAOTcHhDAABbJW1DEvG2dqurR6KL6BcY03eUB1sGkoY4ubp7eJuHXHqxeKJDYxO7pRjMNgcHhAu4GIxFYGlGoIZCVSRldEVD71WRIFrQZxnXgMXog5LgpDDH5jchraazebAKBgUmrSbJba7dz7I4bE5nFykoA)

This means the `useQueries` helper types are getting confused, thinking `TQueryFnData` could itself be a promise.

Interestingly, if we make it so there's only 1 instance of `QueryFunctionData`, with a union as its type argument, as I'm doing in this PR, it seems to work just fine. [Rules of TS claim that the two expressions should be equivalent](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types), so it's a bit counter-intuitive that 1 way works and the other doesn't.

So my hunch is that the issue was caused by how TS goes about evaluating... when it's doing inference on a union of two nested conditional expressions, it may be evaluating twice, and merges the results together, meaning we end up with both `string` and `Promise<string>` that fit the bill.

 (I'm asking around to get a better understanding of the mechanics - will let you know if it's caused by something else).

### DIFF
--- a/src/reactjs/tests/useQueries.test.tsx
+++ b/src/reactjs/tests/useQueries.test.tsx
@@ -776,8 +776,25 @@ describe('useQueries', () => {
       expectType<QueryObserverResult<string, unknown>>(result4[1])
       expectType<QueryObserverResult<number, unknown>>(result4[2])
 
-      // Array as const does not throw error
+      // handles when queryFn returns a Promise
       const result5 = useQueries({
+        queries: [
+          {
+            queryKey: key1,
+            queryFn: () => Promise.resolve('string'),
+            onSuccess: (a: string) => {
+              expectType<string>(a)
+              expectTypeNotAny(a)
+            },
+            // @ts-expect-error (refuses to accept a Promise)
+            onSettled: (a: Promise<string>) => null,
+          },
+        ],
+      })
+      expectType<QueryObserverResult<string, unknown>>(result5[0])
+
+      // Array as const does not throw error
+      const result6 = useQueries({
         queries: [
           {
             queryKey: ['key1'],
@@ -789,8 +806,8 @@ describe('useQueries', () => {
           },
         ],
       } as const)
-      expectType<QueryObserverResult<string, unknown>>(result5[0])
-      expectType<QueryObserverResult<number, unknown>>(result5[1])
+      expectType<QueryObserverResult<string, unknown>>(result6[0])
+      expectType<QueryObserverResult<number, unknown>>(result6[1])
 
       // field names should be enforced - array literal
       useQueries({
--- a/src/core/tests/query.test.tsx
+++ b/src/core/tests/query.test.tsx
@@ -803,7 +803,8 @@ describe('query', () => {
 
     const observer = new QueryObserver(queryClient, {
       queryKey: key,
-      queryFn: (() => undefined) as any,
+      // @ts-expect-error (queryFn must not return undefined)
+      queryFn: (() => undefined),
       retry: false,
     })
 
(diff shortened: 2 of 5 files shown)


TkDodo (approved):
great work, as always. thank you 🙏

src/reactjs/tests/useQueries.test.tsx
```diff
@@ -776,8 +776,25 @@ describe('useQueries', () => {
...
 
-      // Array as const does not throw error
+      // handles when queryFn returns a Promise
       const result5 = useQueries({
+        queries: [
+          {
+            queryKey: key1,
+            queryFn: () => Promise.resolve('string'),
+            onSuccess: (a: string) => {
+              expectType<string>(a)
```
symbiont-matthew-pirocchi, line 786: @artysidorenko Out of curiosity, what is the motivation for `expectType` and `expectTypeNotAny` here?

Looking at their [definitions](https://github.com/tannerlinsley/react-query/blob/5562cfaaa04bbcea141c553dbdb211fec8224357/src/reactjs/tests/utils.tsx#L76), they are compile-time checks rather than runtime checks. In this case, the compile-time check is satisfied by `(a: string)` from the previous line in this test. In other words, it's testing the test code rather than the product code.

Am I missing something? I'm not involved in this codebase, just noticed this and thought I'd ask.

artysidorenko: Hi @symbiont-matthew-pirocchi,

Yeah, the `expectType` and `expectTypeNotAny` are not strictly necessary here, the way the types work at the moment. I think while I was writing the other tests I got into the pattern of test-asserting _every_ type to avoid any surprises.

Am still holding out hope that at some point we'll figure out a way to do the context-sensitive typing on the function parameter, or TS will release a new version that does this more effectively. Then we would replace that snippet with something like this, where the helpers are more important:

```typescript
onSuccess: (a) => {
  expectType<string>(a)
  expectTypeNotAny(a)
},
```

I suppose having these helpers in there already could be a good primer to keep the checks when that happens 🤞

TkDodo: @artysidorenko I just stumbled upon eslint-plugin-expect-type, and the PR that adds support for two-slash annotations. Do you think that would be a good replace for type-level tests?
https://twitter.com/joshuakgoldberg/status/1514382066795876352

artysidorenko: I'm sort of on the fence; the annotation-style looks cool, but reckon you'd still need the same volume of tests to check all the types (just replacing all those `expectType` calls with the two-slashes).

PR #3498, opened 2022-04-12
docs: provide a fully-working example for the quick start docs

mozz100:
Updates the documentation to solve the issue encountered in https://github.com/tannerlinsley/react-query/issues/3461

### DIFF
--- a/docs/src/pages/quick-start.md
+++ b/docs/src/pages/quick-start.md
@@ -9,7 +9,10 @@ This example very briefly illustrates the 3 core concepts of React Query:
 - [Mutations](./guides/mutations)
 - [Query Invalidation](./guides/query-invalidation)
 
+You can copy and paste it into `index.js` in a freshly-created app made with `create-react-app`.
+
 ```js
+import { render } from 'react-dom'
 import {
   useQuery,
   useMutation,
@@ -17,7 +20,16 @@ import {
   QueryClient,
   QueryClientProvider,
 } from 'react-query'
-import { getTodos, postTodo } from '../my-api'
+
+// Provide some stub/dummy query functions.
+// In a real implementation, getTodos and postTodo would
+// make requests to a backend.
+let todos = []
+const getTodos = () => todos
+const postTodo = async (todo) => {
+  todos.push(todo);
+  return todos;
+};
 
 // Create a client
 const queryClient = new QueryClient()
@@ -49,7 +61,7 @@ function Todos() {
   return (
     <div>
       <ul>
-        {query.data.map(todo => (
+        {query.data?.map(todo => (
           <li key={todo.id}>{todo.title}</li>
         ))}
       </ul>


docs/src/pages/quick-start.md
```diff
@@ -9,15 +9,27 @@ This example very briefly illustrates the 3 core concepts of React Query:
...
+
+// Provide some stub/dummy query functions.
+// In a real implementation, getTodos and postTodo would
+// make requests to a backend.
+let todos = []
+const getTodos = () => todos
+const postTodo = async (todo) => {
+  todos.push(todo);
+  return todos;
+};
```
HenriqueDerosa, line 32: Do you think it's better to explicitly have these functions rather than importing from a fake path?
Maybe, then, it would be easier to understand if you actually make an axios call in the getTodos function?
I'm just wondering what's the best option 🤔

TkDodo: i would keep the import from `my-api` please. this page should just show you the concepts, I don't think its meant to be copy-paste able. The idea is that you read on with the docs, and understand the concepts, not that you go to the quickstart page and copy-paste the example :) If you want something to copy-paste, the `Examples` section is the best bet, and the `simple` example only has a couple of lines of code.

mozz100: I'm a newcomer here, and I can only tell you my own point of view.  I'm fine with being overruled, but I'd like you to hear my feedback.

Given the title "Quick Start", I navigated to this code _expecting it to work_.  I clicked through to the page because I wanted to try out the project quickly, to actually get started.  Further, the words at the top "This example..." give the impression that it's a working example.

If the code doesn't/can't possibly work, this should be prominently highlighted!  Don't make newcomers discover that the hard way.  As a newcomer to the project, I had a really tough time getting past this: honestly, I nearly gave up.

Could you reconsider?  Could a quick start example both illustrate the concepts and work with copy-paste?  (I've tested that the code in this merge request works as-is, and I only needed to add seven lines and some comments.

mozz100: P.S.  I just tried out the "simple" example.  It's 40 lines (a bit more than "a couple").  But it is awesome.  I'd suggest making it more prominently linked from the Quick Start.

TkDodo: The simple example is also visible on the first page of the docs (`overview`) after the intro text: https://react-query.tanstack.com/overview#enough-talk-show-me-some-code-already

it is fully runnable because it has an inlined query function that fetches from the github api.

I understand that you had a bad time, and I appreciate you taking the time to contribute to the docs. It seems like this particular flow (skip the overview page, go to quick start, expect something fully running) doesn't work so well with how the docs are currently structured. But other people have different impressions. Extending this Quick Start guide with a "mock" backend, people can go like:
- why do I need `let todos` here?
- why do you want me to use create-react-app? I've heard Dan Abramov say that you should use nextjs nowadays
- why is the QuickStart over 60 lines of code? I've heard react-query is simple ...

and so on ...

really, I'd rather just leave it and extend the text to:

```diff
- This example very briefly illustrates the 3 core concepts of React Query:
+ This code snippet very briefly illustrates the 3 core concepts of React Query:
[Queries](https://react-query.tanstack.com/guides/queries)
[Mutations](https://react-query.tanstack.com/guides/mutations)
[Query Invalidation](https://react-query.tanstack.com/guides/query-invalidation)
+ If you're looking for a fully functioning example, please have a look at our [codesandbox examples](https://react-query.tanstack.com/examples/simple)
```

even if people won't read the texts much :)

TkDodo:
I'm gonna close this due to staleness.

PR #3506, opened 2022-04-14
nonce for devtools style

### LINKED ISSUE #3495: CSP enabled DevTool show errors (insert inline style)
### Describe the bug

When CSP enabled, (Style-SRC='xxx')

the console error
```
 Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self' 'nonce-0mihlgp7' 'report-sample'
```

https://codesandbox.io/s/heuristic-cannon-it3v0r?file=/public/index.html

### Steps to reproduce

1. start the app
2. open the app in single window
3.  open the console and refresh

### Expected behavior

No errors or passing the nonce as params

Every time

### Screenshots or Videos

_No response_

### Platform

Browsers

### react-query version

3.34.19

### TypeScript version

_No response_

### Additional context

_No response_

NewFuture:
* pass nonce to devtools
* __webpack__nonce__
* set style scoped 

fix #3495

### DIFF
--- a/src/devtools/devtools.tsx
+++ b/src/devtools/devtools.tsx
@@ -238,7 +238,7 @@ export function ReactQueryDevtools({
       <ThemeProvider theme={theme}>
         <ReactQueryDevtoolsPanel
           ref={panelRef as any}
-          styleNonce={styleNonce || (window as any).__webpack_nonce__}
+          styleNonce={styleNonce}
           {...otherPanelProps}
           style={{
             position: 'fixed',
@@ -485,7 +485,6 @@ export const ReactQueryDevtoolsPanel = React.forwardRef<
         <style
           scoped
           nonce={styleNonce}
-          data-testid="react-query-devtools-panel-style"
           dangerouslySetInnerHTML={{
             __html: `
             .ReactQueryDevtoolsPanel * {
--- a/src/devtools/devtools.tsx
+++ b/src/devtools/devtools.tsx
@@ -58,6 +58,10 @@ interface DevtoolsOptions {
    * Defaults to 'aside'.
    */
   containerElement?: string | any
+  /**
+   * nonce for style element for CSP
+   */
+  styleNonce?: string
 }
 
 interface DevtoolsPanelOptions {
@@ -73,6 +77,10 @@ interface DevtoolsPanelOptions {
    * A boolean variable indicating whether the panel is open or closed
    */
   isOpen?: boolean
+  /**
+   * nonce for style element for CSP
+   */
+  styleNonce?: string
   /**
    * A function that toggles the open and close state of the panel
    */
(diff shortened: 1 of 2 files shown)


src/devtools/devtools.tsx
```diff
@@ -229,6 +238,7 @@ export function ReactQueryDevtools({
       <ThemeProvider theme={theme}>
         <ReactQueryDevtoolsPanel
           ref={panelRef as any}
+          styleNonce={styleNonce || (window as any).__webpack_nonce__}
```
TkDodo, line 241: I'm fine with the style nonce, but I'd rather not couple this to window and `__webpack_nonce__`. It should be easily doable in user land to forward this:

```
<ReactQueryDevtools styleNonce={window.__webpack_nonce__}
```

src/devtools/devtools.tsx
```diff
@@ -467,6 +483,9 @@ export const ReactQueryDevtoolsPanel = React.forwardRef<
         {...panelProps}
       >
         <style
+          scoped
+          nonce={styleNonce}
+          data-testid="react-query-devtools-panel-style"
```
TkDodo, line 488: we can just find the `style` tag without needing a test-id I guess, just by doing:

```
const { container } = render(<Page />)
const button = container.querySelector('style')
```

TkDodo (changes requested):

src/devtools/devtools.tsx
NewFuture: removed

NewFuture:
@TkDodo help review 😸

src/devtools/devtools.tsx
```diff
@@ -467,6 +483,8 @@ export const ReactQueryDevtoolsPanel = React.forwardRef<
         {...panelProps}
       >
         <style
+          scoped
```
TkDodo, line 486: what does this attribute do please? according to the docs, it's a non-standard, deprecated thing?

https://developer.mozilla.org/en-US/docs/Web/API/HTMLStyleElement/scoped

NewFuture: I did not realize that it's deprecated now.
The purpose is to make this style with no side effects on the whole document.

TkDodo: I still think we should remove it please.

NewFuture: yeah, removed

src/devtools/tests/devtools.test.tsx
```diff
@@ -391,4 +391,21 @@ describe('ReactQueryDevtools', () => {
     expect(queries[1]?.textContent).toEqual(query2Hash)
     expect(queries[2]?.textContent).toEqual(query3Hash)
   })
+
+  it('style should have scoped and nonce', async () => {
```
TkDodo, line 395:
```suggestion
  it('style should have a nonce', async () => {
```

TkDodo (approved):

PR #3509, opened 2022-04-14
fix: type QueryClientProvider children explicitly

filiptammergard:
[React 18 introduced some breaking type changes](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210). One is that `children` is no longer implicitly typed with `React.FC`. This breaks `QueryClientProvider`, since it relies on implicit children through `React.FC`. It can be fixed by typing it explicitly, while it will be no change for users of React 17 or lower.

### DIFF
--- a/src/react/QueryClientProvider.tsx
+++ b/src/react/QueryClientProvider.tsx
@@ -42,6 +42,7 @@ export const useQueryClient = () => {
 }
 
 export interface QueryClientProviderProps {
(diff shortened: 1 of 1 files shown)


TkDodo:
This is the 3rd time a PR for this was opened. While I appreciate the initiative, please look at the other closed PRs for the reasons why we haven't merged them.

PR #3512, opened 2022-04-15
fix(persistQueryClient): rename isHydrating to isRestoring

TkDodo:


docs/src/pages/plugins/persistQueryClient.md
```diff
@@ -220,6 +220,10 @@ ReactDOM.createRoot(rootElement).render(
   - will be called when the initial restore is finished
   - can be used to [resumePausedMutations](../reference/QueryClient#queryclientresumepausedmutations)
 
+### useIsRestoring
+
+If you use the `PersistQueryClientProvider`, you can also use the `useIsRestoring` hook alongside it to check if you a restore is currently in progress. `useQuery` and friends also check this internally to avoid race conditions between the restore and mounting queries.
```
TkDodo, line 225:
```suggestion
If you are using the `PersistQueryClientProvider`, you can also use the `useIsRestoring` hook alongside it to check if a restore is currently in progress. `useQuery` and friends also check this internally to avoid race conditions between the restore and mounting queries.
```

PR #3518, opened 2022-04-16
feat(query): functions for refetchOn options

TkDodo:
refs #3110, #3517

### DIFF
--- a/docs/src/pages/reference/useQuery.md
+++ b/docs/src/pages/reference/useQuery.md
@@ -110,27 +110,27 @@ const result = useQuery({
 - `refetchIntervalInBackground: boolean`
   - Optional
   - If set to `true`, queries that are set to continuously refetch with a `refetchInterval` will continue to refetch while their tab/window is in the background
-- `refetchOnMount: boolean | "always" | (query: Query) => boolean | "always")`
+- `refetchOnMount: boolean | "always" | ((query: Query) => boolean | "always")`
   - Optional
   - Defaults to `true`
   - If set to `true`, the query will refetch on mount if the data is stale.
   - If set to `false`, the query will not refetch on mount.
   - If set to `"always"`, the query will always refetch on mount.
-  - If set to a function, the function will be executed with the latest data and query to compute the value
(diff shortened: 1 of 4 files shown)


Noitidart:
Wow this feature is so great, I literally just needed this right now. As I have cacheTime at Infinity, and retry is also infinite. On unmount it goes into error state as retries are canceled. My refetchOnMount was false. So when came back, it would not refetch as its in error state. I need to make it refetch on mount only if its in error state. Thank you!!

Noitidart:
> Wow this feature is so great, I literally just needed this right now. As I have cacheTime at Infinity, and retry is also infinite. On unmount it goes into error state as retries are canceled. My refetchOnMount was false. So when came back, it would not refetch as its in error state. I need to make it refetch on mount only if its in error state. Thank you!!


Oh man I just tested this, and it seems refetchOnMount is not being called while the query is in the `error` status.

TkDodo:
@Noitidart i don't fully understand the problem. Can you open a discussion with a codesandbox reproduction please

Noitidart:
Oh yes totally, thanks!

Noitidart:
Oh man in codesandbox, when I unmount it, its not going into error state. I can't reproduce. Maybe because I just updated react-query thigns are cached.

Here's the codesandbox which is working fine - https://codesandbox.io/s/try-to-get-infinite-retry-to-error-55gf0l

1. Click button to mount it
2. We see in console "reading from disk" every 2 seconds
3. Click button to destroy it.
4. Wait 3 seconds, till the query throws.
5. Click button to mount it, and its still in loading state, weird. In my app when I click button to destroy it, it goes into error state, then on mount it stays in error state. But in the codesandbox on mount it always starts in loading state.

Noitidart:
Ohhh it was the `retryOnMount` I had set that to false, man that's so silly I mixed it up with `refetchOnMount`. Man thanks man you're always helping out the newbies even though they're wasting your time. Man sorry about that. Awwwesome awwwwesome lib!

PR #3521, opened 2022-04-17
ESM compliant bundling

### LINKED ISSUE #3513: `react-query/devtools`'s `module` entry points to a CJS file
### Describe the bug

Essentially reposting https://github.com/tannerlinsley/react-query/issues/3505 here.
Though Vite reverted that change, I believe it is a bug in `react-query`.

The [`module` field of `react-query/devtools`](https://unpkg.com/browse/react-query@3.34.19/devtools/package.json) points to a [CJS file](https://unpkg.com/browse/react-query@3.34.19/devtools/index.js) that reexports from a subfolder containing nothing but a `package.json`, and it expects the reexport resolves to the `module` field of that subfolder.

It violates the convention that [`module` field should contain ESM syntax](https://github.com/stereobooster/package.json#module).

---

To provide different exports for different environments in the ES module system, maybe the `exports` field can help here: https://nodejs.org/api/packages.html#community-conditions-definitions

https://github.com/akharkhonov/vite-react-query

### Steps to reproduce

1. Install dependencies `yarn --frozen-lockfile`
2. Run dev mode `yarn dev`
3. Open devttools and see errors

### Expected behavior

No errors.

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: Version 100.0.4896.88 (Official Build) (arm64)

### react-query version

3.34.19

### TypeScript version

4.6.3

### Additional context

_No response_

sachinraja:
closes #3513

- add package.json `"exports"`
- Top level files (exports) now just point to dist files instead of having their own `package.json`. Individual package.json files are not esm-compliant and this is already handled by `"exports"`.
- Types are now emitted into `lib` instead of a separate `types` directory. This will help support TypeScript, which will add support for `"exports"` in the next release.
- upgrade and remove various packages related to bundling and building

Unfortunately this makes the process of adding a new export more difficult:
1. add UMD global to `scripts/rollup/umd-config.js`
2. add entry to `"exports"`
3. add top level directory with files that re-export built files to support bundlers that do not support `"exports"`
4. (not sure) add input file to `tsconfig.types.json`


sachinraja:
Looks like it failed on node v12 because a dependency required at least node v14. Considering v12 is about to go EOL in 13 days, maybe it should be removed from CI?

TkDodo:
> Looks like it failed on node v12 because a dependency required at least node v14. Considering v12 is about to go EOL in 13 days, maybe it should be removed from CI?

yes, we can totally do that for v4

package.json
```diff
@@ -11,17 +11,61 @@
...
+    "./core": {
+      "import": "./lib/core/index.mjs",
+      "default": "./lib/core/index.js"
+    },
+    "./reactjs": {
+      "import": "./lib/reactjs/index.mjs",
+      "default": "./lib/reactjs/index.mjs"
+    },
+    "./devtools": {
+      "development": {
```
TkDodo, line 41: will we still be able to import from react-query/devtools/development to do e.g. this:

https://twitter.com/tannerlinsley/status/1392857647037091849

?

sachinraja: No, although I could add support for that.

sachinraja: added support for it

sachinraja:
Well that's unfortunate and I'm really not sure why it's happening. Seems like it's not an issue in `devtools` either, but rather an issue with the actual query client now.

TkDodo:
> Seems like it's not an issue in devtools either, but rather an issue with the actual query client now.

no, it does seem to be an issue with the devtools. The devtools are a separate entry point, but they import directly from `react-query`.

But, whatever you just changed, it seems to be working now 🎉

sachinraja:
Awesome! It was probably that I set it up to resolve `react-query` to the index file instead of leaving it. Also looks like `basic` works, but `basic-typescript` doesn't (and has the error I was referring to above).

sachinraja:
Looks like both examples are working now :tada:!

TkDodo:
> Looks like both examples are working now 🎉!

Awesome, thank you for this contribution. I have to admit I don't really understand the changes made, so let's wait until @tannerlinsley has taken a look at them.

In the meantime:
- please add the breaking changes to the migration guide (see: https://github.com/tannerlinsley/react-query/pull/3521#issuecomment-1102791682)
- For current `beta`, I can see a bundle size of:
`dist/react-query.production.min.js: 11.3KB`
but on this PR, it increases to:
`dist/react-query.production.min.js: 11.36KB`
this could be explained if this branch is not up-to-date with the latest `beta`. could you check that please?

tannerlinsley (approved):
I approve! But only if you promise to give the same treatment to my publish scripts for React Table 😉

sachinraja:
@TkDodo I've switched back to Babel to building the dist so output should be the same now as before now.

> But only if you promise to give the same treatment to my publish scripts for React Table

@tannerlinsley Of course! I'm always willing to fix build configs :smile:.

TkDodo:
thanks. I'm still seeing size differences I cannot quite explain. Compared to `beta`:

```diff
ESM PR: dist/broadcastQueryClient-experimental.production.min.js: 3.55KB
BETA  : dist/broadcastQueryClient-experimental.production.min.js: 3.55KB
ESM PR: dist/createAsyncStoragePersister.production.min.js: 625B
BETA  : dist/createAsyncStoragePersister.production.min.js: 625B
ESM PR: dist/createWebStoragePersister.production.min.js: 674B
BETA  : dist/createWebStoragePersister.production.min.js: 674B
- ESM PR: dist/persistQueryClient.production.min.js: 2.19KB
+ BETA  : dist/persistQueryClient.production.min.js: 1.4KB
ESM PR: dist/react-query-core.production.min.js: 9.85KB
BETA  : dist/react-query-core.production.min.js: 9.85KB
ESM PR: dist/react-query-devtools.production.min.js: 13.71KB
BETA  : dist/react-query-devtools.production.min.js: 13.71KB
- ESM PR: dist/react-query.production.min.js: 11.36KB
+ BETA  : dist/react-query.production.min.js: 11.3KB
```

so it looks to me like some code of the that was previously in the `persistQueryClient` bundle has somehow moved towards the main `react-query` bundle 🤔 
maybe I've added a wrong import there that is now resolved differently?

sachinraja:
Looks like that was caused by upgrading to the very latest version of `@rollup/plugin-commonjs` (published 2 days ago). I downgraded to the version before and I'm seeing the same numbers as [the latest beta version published](https://unpkg.com/browse/react-query@4.0.0-beta.5/dist/).

TkDodo:
splendid 🤩 . We can take a look at the latest rollup plugin separately. would be curious to know what is changing here :)

latin-1:
```
Module not found: Error: Can't resolve 'use-sync-external-store/shim' in '/home/runner/work/mono/mono/node_modules/react-query/lib/reactjs'
Did you mean 'index.js'?
BREAKING CHANGE: The request 'use-sync-external-store/shim' failed to resolve only because it was resolved as fully specified
```

Since `use-sync-external-store` doesn't include the `exports` field in its `package.json`, we should use `use-sync-external-store/shim/index.js` instead.

TkDodo:
@latin-1 when are you getting this error?

@sachinraja can you have a look please?

latin-1:
@TkDodo I got that error from Webpack. Since Webpack 5, `fullySpecified` is enabled by default.
https://webpack.js.org/configuration/module/#resolvefullyspecified

sachinraja:
I don't have time to fix it right now, but @latin-1 is correct. We should change all instances of importing from `use-sync-external-store/shim` to `use-sync-external-store/shim/index.js`.

TkDodo:
@latin-1 can you contribute that maybe ?

sachinraja:
@TkDodo I'm not sure exactly what happened but it seems like this was (unintentionally?) reverted when the project was converted to a monorepo.

TkDodo:
@sachinraja yes, seems so. Can you fix it please?

TkDodo:
@sachinraja I'm pretty sure this is also stopping me from fixing this issue:
- https://github.com/TanStack/query/pull/3861

I tried to make the exports conditional, but it doesn't work for the esm builds. I remember you changed something there as well?

PR #3529, opened 2022-04-19
feat(types): make QueryClientConfig generic

SevenOutman:
Add type argument for `QueryClientConfig` generic type.

Please see discussion #3528 for detailed explanation.

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -676,10 +676,10 @@ export type MutationObserverResult<
   | MutationObserverErrorResult<TData, TError, TVariables, TContext>
   | MutationObserverSuccessResult<TData, TError, TVariables, TContext>
 
-export interface QueryClientConfig {
+export interface QueryClientConfig<TError = unknown> {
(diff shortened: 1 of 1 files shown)


src/core/types.ts
```diff
@@ -676,10 +676,10 @@ export type MutationObserverResult<
   | MutationObserverErrorResult<TData, TError, TVariables, TContext>
   | MutationObserverSuccessResult<TData, TError, TVariables, TContext>
 
-export interface QueryClientConfig {
+export interface QueryClientConfig<TError = unknown> {
   queryCache?: QueryCache
   mutationCache?: MutationCache
-  defaultOptions?: DefaultOptions
+  defaultOptions?: DefaultOptions<TError>
```
TkDodo, line 682: I don't think this works. The `config` passed to the `QueryClient` constructor is still not generic. Additionally, the caches also have global onError callbacks that wouldn't get this generic.

all in all, I think its simpler to go with the runtime check.

PR #3532, opened 2022-04-19
Exposing errorUpdateCount on useQuery

OmerCohenAviv:
exposing errorUpdateCount state,


docs/src/pages/reference/useQuery.md
```diff
@@ -235,6 +235,8 @@ const result = useQuery({
   - The failure count for the query.
   - Incremented every time the query fails.
   - Reset to `0` when the query succeeds.
+- `errorUpdateCount: number`
+  - The sum of all refetch failures.
```
TkDodo, line 239:
```suggestion
  - The sum of all errors.
```

src/core/queryObserver.ts
```diff
@@ -586,6 +586,7 @@ export class QueryObserver<
       error,
       errorUpdatedAt,
       failureCount: state.fetchFailureCount,
+      errorUpdateCount: state.errorUpdateCount,
```
TkDodo, line 589: let's maybe name it `errorCount` when exposing, to be in line with `failureCount` ?

src/react/tests/useQuery.test.tsx
```diff
@@ -4677,4 +4677,38 @@ describe('useQuery', () => {
...
+        </div>
+    }
+    const rendered = renderWithClient(queryClient, <Page />)
+    const fetchBtn = rendered.getByRole('button', { name: 'refetch' })
+    fireEvent.click(fetchBtn)
+    await waitFor(() => rendered.getByText('data: 1'))
+    fireEvent.click(fetchBtn)
+    await waitFor(() => rendered.getByText('data: 2'))
+    fireEvent.click(fetchBtn)
+    await waitFor(() => rendered.getByText('data: 3'))
```
TkDodo, line 4710: since the query is enabled per default, I think we should wait for the count to become `1` before we click refetch the first time. the error count is not about refetches - its about all errors :)

TkDodo (commented):
looks really good already 🙌

src/react/tests/useQuery.test.tsx
OmerCohenAviv: Thanks, will update the test :)

src/core/queryObserver.ts
Omercoav: > don't you think errorCount / failureCount will cause confusion?
> 
> could maybe provide better DOC description for what is errorCount and how is different from failureCount , The sum of all errors could be interpreted to many cases I suppose.
> 
> if not I can proceed with your suggestion :)

TkDodo: yeah, we can also keep it at `errorUpdateCount`. If we expose a `query`, e.g. via the predicate filter function, it will also be accessible as:

```
query => query.state.errorUpdateCount
```

the rename might cause confusion, too.

TkDodo (approved):

TkDodo:
please check prettier:

```
> react-query@1.0.0-semantic-release test:format
> yarn prettier --check

Checking formatting...
[warn] src/react/tests/useQuery.test.tsx
```

TkDodo:
some test are failing - please add `errorUpdateCount` to the failing tests

OmerCohenAviv:
> now there are conflicts, sorry 🙈 . can you please resolve them?

hopefully it's ok now sorry for trouble

PR #3545, opened 2022-04-22
Error update count test

OmerCohenAviv:
it seems the test was removed ,
 could we use this branch instead?
Thanks


TkDodo:
yeah, not sure what happened. can you make a PR that only re-adds the tests? Everything else should already be in master...

PR #3549, opened 2022-04-23
Deduplicate packages in yarn.lock

Naoya9922:
Deduplicated packages in yarn.lock ~~to reduce bundle size~~.

```bash
$ npx yarn-deduplicate --strategy highest
```

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -18,44 +18,14 @@
   optionalDependencies:
     chokidar "^2.1.8"
 
-"@babel/code-frame@^7.0.0":
-  version "7.5.5"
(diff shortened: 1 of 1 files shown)


TkDodo:
How would this affect bundle size?

Naoya9922:
> How would this affect bundle size?

As it will depend fewer libraries after deduplicating.

TkDodo:
> As it will depend fewer libraries after deduplicating.

but, we don't have runtime dependencies - only devDependencies 🤔

Naoya9922:
Oh, I got it.
It won't affect bundle size as you say.
It will just eliminating duplicated devDependencies.

Naoya9922:
Should PR be closed?

PR #3554, opened 2022-04-24
fix: cache select errors

TkDodo:
refs #3530

there is no need to re-run the select function if it is referentially stable, even when it throws an error. if no dependency has changed, we can re-use the cached error, as the select function should be idempotent and thus return the same error anyway.

further, if we have cached data from a previously successful select, that should be returned as `data` alongside the error.


TkDodo:
working forked sandbox from the discussion with the preview build: https://codesandbox.io/s/error-logging-forked-jp52dv?file=/src/App.jsx

PR #3557, opened 2022-04-24
Keep data in useQuery result defined when initialData is defined

### LINKED ISSUE #3310: better type narrowing for initialData
when providing `initialData` directly on a `query`, the type of `data` is guaranteed to be `TData`, _not_ `TData | undefined`:

```
const { data } = useQuery(['test'], () => Promise.resolve('test'), { initialData: 'something' })

// 🚨 Object is possibly 'undefined'.(2532)
data.toUpperCase()
```

[TypeScript playground](https://www.typescriptlang.org/play?ssl=6&ssc=19&pln=1&pc=1#code/JYWwDg9gTgLgBAbzgVwM4FMCKz1QJ5wC+cAZlBCHAORToCGAxjALQCOO+VAUFwxAHap4SACZ0YdInAC8KDNlx4AFAG0qMdEKoBdADRwlAShkA+OAAVyIYBgB0tVBAA2AN3RL1mmFUP6kwfmAYYDonABFxOgAuakcQdBgACwCAcyoiQx4Aeiy4QF4NwAq9uAB5ACMAK3QmOBs4SFRUYFKnAipkfhF0EgD0ESpbJQAmAFYAZkHMsQlbGAgAVTAwXABhOgwjIA)

ideally, we could narrow the type when `initialData` is provided. This _only_ works for `initialData` because `initialData` is executed in the constructor of the query, and because it is synchronous. Also, it can only work on `v4` because in `v3`, `undefined` is technically still valid for the `TData` generic, so we cannot remove it from the union. In `v4`, `undefined` is illegal for successful queries, so that would now work.

Keep in mind that `initialData` can also be a function that returns `TData | undefined`, so this should also narrow if the function returns `TData`.

It does _NOT_ work for:
- `placeholderData`, because that will revert back to `undefined` when a query errors
- `initialData` being provided globally, because TS cannot know that on an instance level. It needs to be explicitly passed to `useQuery`.
- `suspense` or any other case because a query can be `cancelled` while it is initially `fetching`, in which case it will be `reset` to its initial state. `queryClient.resetQueries` also does that. Also, without network connection in the default network mode, the query will be in `fetchStatus: 'paused'` and will also not have data.

I guess we could potentially do this with another overload that has `initialData` provided and removes `undefined` from the result union, but we would also have do this for `useQueries`.
I would like to avoid having to add _another_ generic to `useQuery`, as we already have four of them.

Because this is for v4, PRs should go to the `beta` branch please.

prateek3255:
Fixes #3310


examples/basic-typescript/src/index.tsx
```diff
@@ -26,13 +26,10 @@ type Post = {
 };
 
 function usePosts() {
-  return useQuery(
+  return useQuery<Array<Post>>(
```
TkDodo, line 29: not really a fan of annotating call-side generics. This should be inferred, should it not?

prateek3255: Yeah it is inferred automatically, I was just testing something, that's why added it, forgot to remove it.

TkDodo:
superseded by:
- https://github.com/TanStack/query/pull/3834

PR #3561, opened 2022-04-26
Use fully-specified path to import use-sync-external-store/shim

latin-1:
https://github.com/tannerlinsley/react-query/pull/3521#issuecomment-1109810824


latin-1:
I've tested it locally, and it appears to work properly.

sachinraja (approved):

TkDodo:
thank you for the swift fix 🥳

PR #3571, opened 2022-04-29
feat: make InfiniteQueryObserver's type more robust

liaoliao666:
Currently i face with this problem that i defined a `InfiniteQueryObserverOptions` that contains `QueryKey`. and then i pass it into `InfiniteQueryObserver`. but unfortunately its not compatible with options of  `InfiniteQueryObserver`'s constructor


TkDodo:
I think it looks good, but could you maybe show an example that didn't work before? Best would be a type level test that fails on master, but works on your PR. Thanks

PR #3577, opened 2022-04-30
feat(devtools): update react query logo to close devtools

### LINKED ISSUE #3067: [ReactQueryDevtools] Handle in a better way the "close" button
Hello,

I've got an issue with the "close" button on the interface of the dev tools.

image

The "close" button doesn't allow you to see what's the name of the key behind.

Two suggestions to deal with it:
- Use the react-query logo to close the panel
- Add at least a bit of padding in the list of queries to be able to scroll enough to have some space to see the name of the key

Thank you 😊

himankpathak:
### Description
Added button wrapper on react query logo which onClick closes devtools window

### Verification Process
Tested locally

### Applicable Issue
Closes #3067

### DIFF
--- a/src/devtools/devtools.tsx
+++ b/src/devtools/devtools.tsx
@@ -278,7 +278,6 @@ export function ReactQueryDevtools({
         {isResolvedOpen ? (
           <Button
             type="button"
-            aria-label="Close React Query Devtools"
             aria-controls="ReactQueryDevtoolsPanel"
             aria-haspopup="true"
             aria-expanded="true"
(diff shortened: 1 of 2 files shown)


himankpathak:
@TkDodo We can also add padding to the query list to allow scrolling past overlapping queries if needed.

image

src/devtools/devtools.tsx
```diff
@@ -540,12 +540,24 @@ export const ReactQueryDevtoolsPanel = React.forwardRef<
               alignItems: 'center',
             }}
           >
-            <Logo
-              aria-hidden
+            <button
```
TkDodo, line 543: can you please add a test that makes sure clicking the button actually closes the devtools?

src/devtools/devtools.tsx
```diff
@@ -540,12 +540,24 @@ export const ReactQueryDevtoolsPanel = React.forwardRef<
...
+              type="button"
+              aria-label="Close React Query Devtools"
```
TkDodo, line 545: I don't think its good to have two buttons with the same `aria-label`. This will also make `findByRole('button', { name: 'Close React Query Devtools' })` error because it will find both buttons.

It's probably best to keep the aria-label here, but remove it from the button that has the `Close` text, as that button is already labelled by the `Close` text printed on it.

TkDodo:
> We can also add padding to the query list to allow scrolling past overlapping queries if needed.

I think its fine as-is. you can always hide the close button via props if you want to.

src/devtools/devtools.tsx
himankpathak: Added test case for it. 👍

src/devtools/devtools.tsx
himankpathak: Removed `aria-label` from Close and kept it in Logo button.

TkDodo (approved):
nice work, thank you

IanVS:
This is my favorite change in recent versions.  Thank you!

PR #3578, opened 2022-05-02
fix: move usage of React Context out of core package

### LINKED ISSUE #3542: v4 introduces dependency on react in core package
### Describe the bug

In the beta branch i have noticed that [hydration.ts](https://github.com/tannerlinsley/react-query/blob/beta/src/core/hydration.ts#L1) contains reference to `react` types.
This will introduce transitive dependency on `react` for `vue` package.

N/A

### Steps to reproduce

N/A

### Expected behavior

I think we need to remove `react` reference from `core` package.
Maybe introducing some kind of overload in `react` sub-package would be a solution?

_No response_

### Screenshots or Videos

_No response_

### Platform

N/A

### react-query version

v4

### TypeScript version

_No response_

### Additional context

https://github.com/tannerlinsley/react-query/blob/beta/src/core/hydration.ts#L1
https://github.com/tannerlinsley/react-query/blob/beta/src/reactjs/types.ts#L12
https://github.com/tannerlinsley/react-query/blob/beta/src/reactjs/types.ts#L18

TkDodo:
fixes #3542


TkDodo:
@DamianOsipiuk fyi

DamianOsipiuk (approved):

Issue #3584, opened 2022-05-03
[Beta] Disabled query has `isLoading: true` property


maxsbelt:
### Describe the bug

In `react-query@beta` version has strange behaviour. By default query that has `enabled: false` option has `isLoading: true` property (see code sandbox for details). In stable release `react-query@^3` is has `isLoading: false`. I didn't find any notes about this update in change logs so it seems like a bug.

### Your minimal, reproducible example

https://codesandbox.io/s/eager-bell-epuvfi?file=/src/App.js

### Steps to reproduce

Define simple query with `enabled: false` option

### Expected behavior

Query should have `isLoading: false`

### How often does this bug happen?

Every time

### Platform

Any

### react-query version

4.0.0-beta.7

### TypeScript version

Not relevant

TkDodo:
https://react-query-beta.tanstack.com/guides/migrating-to-react-query-4#the-idle-state-has-been-removed

maxsbelt:
@TkDodo in such case I think it makes sense to mention in the migration guild that `query.isLoading` state was updated from `false` to `true` for disabled queries, WDYT? I didn't find any note about this change.

TkDodo:
The migration guide mentions that the idle state was replaced with isLoading: true and fetchStatus: idle. If your query was in idle state on v3, it will be in idle state on v4. My guess is that's what's happening to you, but not every disabled query will be in loading state, just like not every disabled query was in idle state. If you have data and then disable the query, it will still be in success state.

olan-go2:
This is messed up. If the enabled = false, isLoading should be in false as well

kimFrost:
This caught me off guard as well.

This page gives a bit of info on the subject
[https://tanstack.com/query/v4/docs/guides/queries#fetchstatus](https://tanstack.com/query/v4/docs/guides/queries#fetchstatus)

houssemAmiri-bulldozer:
How to deal with disabled false if we have a loading page that work if isloading is true and the loader is showing

atreya2011:
This totally caught me off guard too! I am leaving a link to this in my repo for my future self 😂

epicmau5time:
So what to do in this case? If I have a query that is disabled, am I to just use isFetching? There is not really as use for isLoading now that it will always show a spinner on components that do not meet a url param criteria (for my use case).

Is param in url? fetch, if it is not always be enabled = false. Due to this change the spinner is always going even if it does not need to be.

[...]

PR #3585, opened 2022-05-04
fix: set isLoading false if query is disabled

maxsbelt:
Updated `isLoading` query property default value from `true` to `false` if query `enabled` option is `false`.

See #3584 for details.

### DIFF
--- a/src/reactjs/tests/useQuery.test.tsx
+++ b/src/reactjs/tests/useQuery.test.tsx
@@ -3743,7 +3743,7 @@ describe('useQuery', () => {
   })
 
   // See https://github.com/tannerlinsley/react-query/issues/360
-  test('should init to status:loading, fetchStatus:idle when enabled is false', async () => {
+  test('should init to status:loading, fetchStatus:idle, isLoading:false when enabled is false', async () => {
     const key = queryKey()
 
     function Page() {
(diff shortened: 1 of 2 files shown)


src/reactjs/tests/useQuery.test.tsx
```diff
@@ -4616,7 +4616,7 @@ describe('useQuery', () => {
 
     expect(states[0]).toMatchObject({
       data: undefined,
-      isLoading: true,
+      isLoading: false,
```
maxsbelt, line 4619: This test in current master (version 3.x.x) has `isLoading: false`:
https://github.com/tannerlinsley/react-query/blob/09a375b/src/react/tests/useQuery.test.tsx#L4494

But seems like this behaviour was missed during `isIdle` refactoring in this PR:
https://github.com/tannerlinsley/react-query/pull/3302/files#diff-eee051dcfac05e804da529ab297862d9341cf6d3963d5fe0c2b3ed1918ca5decR4452

TkDodo:
No, it's on purpose: https://react-query-beta.tanstack.com/guides/migrating-to-react-query-4#the-idle-state-has-been-removed

gregg-cbs:
> No, it's on purpose: https://react-query-beta.tanstack.com/guides/migrating-to-react-query-4#the-idle-state-has-been-removed

Dead link...

PR #3599, opened 2022-05-10
Update useQueries.md

Barna-Molnar:
useQueries does not accept an array or an object with options, you can add them as shown above, I think it should be in the doc

### DIFF
--- a/docs/src/pages/reference/useQueries.md
+++ b/docs/src/pages/reference/useQueries.md
@@ -9,12 +9,13 @@ The `useQueries` hook can be used to fetch a variable number of queries:
 const results = useQueries([
   { queryKey: ['post', 1], queryFn: fetchPost },
   { queryKey: ['post', 2], queryFn: fetchPost },
+  { queryKey: ['post', 3], queryFn: fetchPost, staleTime: 0, enable: true, refetchOnWindowFocus; true, ... },
 ])
 ```
 
 **Options**
 
-The `useQueries` hook accepts an array with query option objects identical to the [`useQuery` hook](/reference/useQuery).
+The `useQueries` hook accepts query options which are identical to the [`useQuery` hook](/reference/useQuery).
 
 **Returns**
 


TkDodo:
useQueries accepts an array _with_ query option objects - so the original sentence is perfectly fine. Also, we've changed the api of useQueries for v4, so the whole sentence is obsolete soon.

Barna-Molnar:
If I try to add an array with options then I always get an error .... but if I try to add the options without an array then It works just fine, and I even get autocomplete for them. Do you have any idea why it behaves differently for me? I am using it with TS. By the way... What have you changed for v4? and when is the date of release?

TkDodo:
I can't help you if you don't show the code. But please no screenshots. A codesandbox reproduction or a typescript playground!

changes for v4 are here: https://react-query-beta.tanstack.com/guides/migrating-to-react-query-4
note that the useQuery api for v4 has changed!

Barna-Molnar:
Yes, of course, I would not have done a screenshot, here is a ts playground [click here](https://playcode.io/898069) as you will see,  I can add any options without {} or [] as a ''list'' but with {} or [] I always get an error. Did I implement something wrong? Let me know pls if you need any info about it. Thank you

TkDodo:
but, your code is alright the way it is?

The docs say:

> The `useQueries` hook accepts an array with query option objects identical to the `useQuery` hook.

and the useQuery hook can be called with:

```
useQuery({ queryKey: myKey, queryFn: myFn, staleTime: 0, refetchOnWindowFocus: true })
```

This is known as the object syntax for useQuery, and it's documented here: https://react-query.tanstack.com/guides/query-functions#using-a-query-object-instead-of-parameters

Barna-Molnar:
Hi, yes it is working fine, 
So that it means, that I have not dug enough deep into the doc, but to be honest I could be a bit clearer or a simple example on that page in the doc, but you are right, and with your explanation I totally understand. The only thing was that we have not used the object syntax for useQuery and therefore It was strange for me. Again thx for your help now is everything clear.
Have a nice week

PR #3600, opened 2022-05-10
Add QueryFunctionContext object properties

aliazizi:


aliazizi:
@TkDodo I see typescript interface and the value of `meta` can't be undefined. but in document you mentioned `meta?` so I think it needs to be changed to `meta`

TkDodo:
`meta` is defined as:

```
meta: QueryMeta | undefined
```

so yes, it can be undefined if you don't supply it.

PR #3613, opened 2022-05-13
fix(query): updating with setQueryData should not affect fetchStatus

### LINKED ISSUE #3594: setQueryData incorrectly sets isFetching to false
### Describe the bug

If `setQueryData` is invoked while a query is fetching, `isFetching` automatically goes to `false`.

Quoting @TkDodo (see [his comment](https://github.com/tannerlinsley/react-query/issues/3579#issuecomment-1120430029) in #3579 ), when `setQueryData` is called
> only the status, data and dataUpdatedAt fields should be set

https://codesandbox.io/s/setquerydata-and-isfetching-bug-gck1zp?file=/src/App.js

### Steps to reproduce

1. Go to [the MWE](https://codesandbox.io/s/setquerydata-and-isfetching-bug-gck1zp?file=/src/App.js)
2. Click "Refetch", and note that "isFetching: true" is displayed
3. Before the query ends (you have 5 seconds), click "Manually update data" 
4.  "isFetching: false" is displayed, even though the query is still being fetched 🐛🐛🐛

### Expected behavior

Calling `setQueryData` shouldn't change the value of `isFetching` 

Every time

### Screenshots or Videos

https://user-images.githubusercontent.com/4582440/167303377-e254b757-d84a-480f-9cdf-0abab3227f9a.mov

### Platform

- OS: MacOs
- Browser: Chrome
- Version: 100

### react-query version

3.39.0 and also 4.0.0-beta.9

### TypeScript version

_No response_

### Additional context

_No response_

TkDodo:
queries can be fetching _while_ we are making a manual update are still fetching, so we have to set fields that affect the fetch conditionally (fetchStatus, fetchFailureCount)

fixes #3594

### DIFF
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -98,7 +98,7 @@ interface SuccessAction<TData> {
   data: TData | undefined
   type: 'success'
   dataUpdatedAt?: number
-  notifySuccess?: boolean
+  manual?: boolean
 }
 
 interface ErrorAction<TError> {
@@ -195,10 +195,7 @@ export class Query<
     }
   }
 
-  setData(
-    data: TData,
-    options?: SetDataOptions & { notifySuccess: boolean }
-  ): TData {
+  setData(data: TData, options?: SetDataOptions & { manual: boolean }): TData {
     const prevData = this.state.data
 
     // Use prev data if an isDataEqual function is defined and returns `true`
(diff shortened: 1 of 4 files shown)


TkDodo:
@lgenzelis fyi

PR #3625, opened 2022-05-19
Properly type useInfiniteQuery to respect the select result

### LINKED ISSUE #3065: TypeScript: Selectors for inifinite query less expressive
**Describe the bug**
Selectors in combination with `useQuery` look very good. Creating multiple hooks on top of the original `useQuery` hook, allowing you to get different parts of the data out the query result as described in https://tkdodo.eu/blog/react-query-data-transformations in particular (e.g. `useTodosCount`, `useTodos`, `useTodo(id)`).

Now, when using `useInfiniteQuery`, I'm able to do the same thing. But TypeScript won't let me. It forces me into the `InfiniteData` type, hence the `{ pages, pageParams }` format.

**To Reproduce**

What I'd like to do is twofold:

* Get an array of all items concatenated over all pages
* Get the total count of items available in the backend

Type definitions of a simplified example:

```ts
interface Entity {
  id: string;
  // ...
}

interface APIResponse {
  total: number;      // Total number of entities available in the backend
  totalPages: number; // Total number of pages with entities
  items: Entity[];    // The entities for this requested page
}

interface DesiredDataResult {
  total: number;   // Same value as total in APIResponse
  items: Entity[]; // Concatenation of all items on all fetched pages (flatMap of `page.items`)
}
```

Now, I try to build the `DesiredDataResult` in the selector:

```ts
const { data: result } = useInfiniteQuery(
  ['entities', 'list'],
  async ({ pageParam = 0 }) => getPageFromAPI(pageParam),
  {
    getNextPageParam: (lastPage, allPages) =>
      lastPage.totalPages > allPages.length ? allPages.length : undefined,
    select: React.useCallback(
      (data: InfiniteData<APIResponse>) => ({
        total: data.pages[0].total,
        entities: data.pages.flatMap((page) => page.items),
      }),
      [],
    ),
  },
);
```

And I get this TypeScript error. It wants me to use the `{ pages, pageParams }` type.

```
Type '(data: InfiniteData<APIResponse>) => { total: number; entities: Entity[]; }' is not assignable to type '(data: InfiniteData<APIResponse>) => InfiniteData<APIResponse>'.
  Type '{ total: number; entities: Entity[]; }' is missing the following properties from type 'InfiniteData<APIResponse>': pages, pageParams ts(2322)
```

Therefore, selectors in combination with `useInfiniteQuery` are less expressive than with `useQuery`. Maybe I'm missing something?

**Expected behavior**
I would have expected to be able to apply selectors with the same expressiveness as `useQuery` on `useInfiniteQuery`.

**Screenshots**
Screenshot 2021-12-07 at 09 03 29

 - OS: macOS
 - Browser: Safari
 - Version: 14.1.2

bradennapier:
Test: https://codesandbox.io/s/tannerlinsley-react-query-forked-gw8jm2?file=/src/index.tsx

Currently the type returned by `useInfiniteQuery` is not honoring the result of the `select` function.  This is due to the fact that it is wrapped at one point in `InfiniteData<TData>` which strips the context once it is passed to the `Query` which doesn't have any wrapping. 

The solution was to allow `InfiniteData` to have two parameters, one would be the normal `TData` and the other is a type parameter `SData` which represents the result of the select fn.  This way it doesn't strip the context and can store both.  The SData ends up taking precedence, but rules do still exist:

- Select must return a compatible type to `InfiniteData<TData>` (it must have a pages and pagesParams property and both must be arrays, but they can be made a diff type if needed). 
- Any extra keys can be returned on top of those keys and the type is automatically inferred.

Such that:

```ts
type FetchResult = {
  totalPages: number;
  page: number;
  items: {
    n: number;
  }[];
};

async function exampleFetch(page: number): Promise<FetchResult> {
  return {
    totalPages: 50,
    page,
    items: [{ n: page + 1 }, { n: page + 2 }, { n: page + 3 }],
  };
}

/*

function transformResult(data: InfiniteData<FetchResult>) {
  return {
    ...data,
    items: data.pages.flatMap(x => x.items),
  };
}

export function useExampleInfiniteQuery() {
  const result = useInfiniteQuery({
    queryKey: ['example'],
    queryFn: ({ pageParam = 1 }) => exampleFetch(pageParam),
    getNextPageParam: (lastPage: FetchResult) =>
      lastPage.page === lastPage.totalPages ? undefined : lastPage.page + 1,
    getPreviousPageParam: () => undefined,
    select: useCallback(data => transformResult(data), []),
  });
  
  // correct
  const data: InfiniteData<FetchResult, {
    items: {
        n: number;
    }[];
  }> | undefined = result.data

  return result.data?.items;
}
```

image

It also allows overwriting the `pages` and `pagesParams` keys, but it will error if the return transformation does not include them and they must be arrays.

So:

```ts
// @ts-expect-error - type must have pages and pageParams arrays
function transformResult(data: InfiniteData<FetchResult>) {
  return {
    items: data.pages.flatMap(x => x.items),
  };
}


/*
InfiniteData<never, {
    pages: never[];
    pageParams: never[];
    items: {
        n: number;
    }[];
}>
*/
function transformResult(data: InfiniteData<FetchResult>) {
  return {
    pages: [],
    pageParams: [],
    items: data.pages.flatMap(x => x.items),
  };
}

/*
InfiniteData<number, {
    pages: number[];
    pageParams: number[];
    items: {
        n: number;
    }[];
}>
*/
function transformResult(data: InfiniteData<FetchResult>) {
  return {
    pages: [1, 2, 3],
    pageParams: [1, 2, 3],
    items: data.pages.flatMap(x => x.items),
  };
}
```

closes #3065

### DIFF
--- a/src/react/useInfiniteQuery.ts
+++ b/src/react/useInfiniteQuery.ts
@@ -11,6 +11,7 @@ export function useInfiniteQuery<
   TQueryFnData = unknown,
   TError = unknown,
   TData = TQueryFnData,
+  SData = unknown,
   TQueryKey extends QueryKey = QueryKey
 >(
   options: UseInfiniteQueryOptions<
@@ -18,13 +19,15 @@ export function useInfiniteQuery<
     TError,
     TData,
     TQueryFnData,
+    SData,
     TQueryKey
   >
-): UseInfiniteQueryResult<TData, TError>
+): UseInfiniteQueryResult<TData, TError, SData>
 export function useInfiniteQuery<
   TQueryFnData = unknown,
   TError = unknown,
   TData = TQueryFnData,
+  SData = unknown,
   TQueryKey extends QueryKey = QueryKey
 >(
   queryKey: TQueryKey,
@@ -34,15 +37,17 @@ export function useInfiniteQuery<
       TError,
       TData,
       TQueryFnData,
+      SData,
       TQueryKey
     >,
     'queryKey'
   >
-): UseInfiniteQueryResult<TData, TError>
+): UseInfiniteQueryResult<TData, TError, SData>
 export function useInfiniteQuery<
   TQueryFnData = unknown,
   TError = unknown,
   TData = TQueryFnData,
+  SData = unknown,
   TQueryKey extends QueryKey = QueryKey
 >(
   queryKey: TQueryKey,
@@ -53,15 +58,17 @@ export function useInfiniteQuery<
       TError,
       TData,
       TQueryFnData,
+      SData,
       TQueryKey
     >,
     'queryKey' | 'queryFn'
   >
-): UseInfiniteQueryResult<TData, TError>
+): UseInfiniteQueryResult<TData, TError, SData>
 export function useInfiniteQuery<
   TQueryFnData,
   TError,
   TData = TQueryFnData,
+  SData = unknown,
   TQueryKey extends QueryKey = QueryKey
 >(
   arg1:
@@ -71,6 +78,7 @@ export function useInfiniteQuery<
         TError,
         TData,
         TQueryFnData,
+        SData,
         TQueryKey
       >,
   arg2?:
@@ -80,6 +88,7 @@ export function useInfiniteQuery<
         TError,
         TData,
         TQueryFnData,
+        SData,
         TQueryKey
       >,
   arg3?: UseInfiniteQueryOptions<
@@ -87,12 +96,14 @@ export function useInfiniteQuery<
     TError,
     TData,
     TQueryFnData,
+    SData,
     TQueryKey
   >
-): UseInfiniteQueryResult<TData, TError> {
+): UseInfiniteQueryResult<TData, TError, SData> {
   const options = parseQueryArgs(arg1, arg2, arg3)
   return useBaseQuery(
     options,
     InfiniteQueryObserver as typeof QueryObserver
-  ) as UseInfiniteQueryResult<TData, TError>
+  ) as UseInfiniteQueryResult<TData, TError, SData>
 }
+
--- a/src/react/useInfiniteQuery.ts
+++ b/src/react/useInfiniteQuery.ts
@@ -106,4 +106,3 @@ export function useInfiniteQuery<
     InfiniteQueryObserver as typeof QueryObserver
   ) as UseInfiniteQueryResult<TData, TError, SData>
 }
-
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,5 +4,6 @@
     "titleBar.inactiveBackground": "#00da63", // change this color!
     "titleBar.activeForeground": "#ffffff", // change this color!
     "titleBar.inactiveForeground": "#ffffff" // change this color!
-  }
+  },
+  "typescript.tsdk": "node_modules/typescript/lib"
 }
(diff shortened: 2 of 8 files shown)


bradennapier:
That being said, the api here does seem broken.  While this works, it doesn't make sense that we are running against the entire set of pages on every request.  

Instead infinite query should receive (page, pages) where pages would be our previously selected pages and the page is the raw page from the fetch.  

Then we'd simply provide our own logic for inserting the new page into the pages and pageParams arrays (as well as any other properties added to InfiniteData as this pr provides capability for

src/react/useInfiniteQuery.ts
```diff
@@ -34,15 +37,17 @@ export function useInfiniteQuery<
...
     >,
     'queryKey'
   >
-): UseInfiniteQueryResult<TData, TError>
+): UseInfiniteQueryResult<TData, TError, SData>
 export function useInfiniteQuery<
   TQueryFnData = unknown,
   TError = unknown,
   TData = TQueryFnData,
+  SData = unknown,
```
TkDodo, line 50: shouldn't `SData` default to `TData` to be backwards compatible?

bradennapier: SData defaults to an empty object and intersects Tdata.  The issue stems around infinite query types passing TData then wrapping it in InfiniteData when it calls the standard query observer type. 

That is how it is passed to the select and how it identifies TData with the select fn. 

So by adding the Sdata type and intersecting it as a second type param, typescript is able to infer that the select fn gives SData its value which is the return type of the select fn whereas TData remains the value that it assigns from either the nextParam fn or the async fn (unless you change that via the select which is interesting to me, but ideal result) 

It's actually a fundamental error in the types causing the issue in the first place, and this PR is technically incorrect (select could return a primitive but we require you return the pages/pagesParams values (although they can be changed) or a type error occurs. 

This is because of the underlying issue around how infinite query ties to query. 

In a deeper rework of types, it may be possible to resolve this, but it would have been a huge changeset from what I can tell. 

This at least provides a framework for Typescript users to be able to transform the infinite data...

bradennapier:
Not sure why this failed for node 16?  Seems odd it works for 14 and 12?

TkDodo:
> Not sure why this failed for node 16? Seems odd it works for 14 and 12? It says that the objects are not equal - I did not make a single change to the codebase, it was all type level changes. Not to mention 12 and 14 were fine... and i developed with 16 and the tests passed.

seems like a flaky test :( I've re-run the pipeline

TkDodo:
can you please add a type level test that shows that these things are working?

we have some examples, e.g. here:

https://github.com/tannerlinsley/react-query/blob/9f6c992feecb4694758f8767063d8ad89d842c3c/src/react/tests/useQueries.test.tsx#L473-L498

TkDodo:
@bradennapier how are we with this PR? We have released v4 in the meantime, so there are a couple of conflicts that would need to be addressed please

TkDodo:
> I am investigating a better fix which will actually work 

@bradennapier what's the status here please?

TkDodo:

TkDodo:
if someone wants to pick up this PR where it left off, please feel free. But no feedback for 3 months means I'm gonna close this stale PR.

joseito-terence:
Any update on this? Or type of select is still not inferred for infinite query?

TkDodo:
It's fixed in v5

PR #3635, opened 2022-05-20
Dates should compare properly

kuberlog:
Hi. I noticed that Date comparisons for `replaceEqualDeep` weren't working and weren't congruent with how `partialDeepEqual` was working.

I added a test to show what `partialDeepEqual` does with Dates (properly returns true if Dates are equivalent)

I then added a test and implementation for a Date comparison. It isn't the same comparison for partialDeepEqual, but I think its equivalent for dates.

### DIFF
--- a/src/core/tests/utils.test.tsx
+++ b/src/core/tests/utils.test.tsx
@@ -93,7 +93,12 @@ describe('core/utils', () => {
       const b = { a: { b: 'b' }, c: 'c', d: [] }
       expect(partialDeepEqual(a, b)).toEqual(true)
     })
+    it('should return true if a and b are dates', () => {
+      const a = { a: new Date('December 17, 1995 03:24:00') }
+      const b = { a: new Date('December 17, 1995 03:24:00') }
 
+      expect(partialDeepEqual(a, b)).toEqual(true)
+    })
     it('should return `false` if a does not include b', () => {
       const a = { a: { b: 'b' }, c: 'c', d: [] }
       const b = { a: { b: 'b' }, c: 'c', d: [{ d: 'd ' }] }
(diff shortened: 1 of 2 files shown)


src/core/tests/utils.test.tsx
```diff
@@ -140,9 +145,16 @@ describe('core/utils', () => {
       expect(replaceEqualDeep(null, null)).toBe(null)
       expect(replaceEqualDeep(undefined, undefined)).toBe(undefined)
     })
+
+    it('should return the previous value when the next value is an equal Date', () => {
+      const a = new Date('December 17, 1995 03:24:00')
```
kuberlog, line 150: nothing special about this date BTW. I stole it from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date

TkDodo:
thanks for the PR. It is a known (and purposeful) limitation that structural sharing only works for json serializable values, and `Date` is not json serializable. This is also noted in the docs on the [important defaults page](https://react-query.tanstack.com/guides/important-defaults):

> Structural sharing only works with JSON-compatible values, any other value types will always be considered as changed. If you are seeing performance issues because of large responses for example, you can disable this feature with the config.structuralSharing flag. If you are dealing with non-JSON compatible values in your query responses and still want to detect if data has changed or not, you can define a data compare function with config.isDataEqual

The problem of diverging from this principle would be that we would likely have to support other non-json-serializable datataypes, like Maps or Sets as well. This is where [superjson](https://github.com/blitz-js/superjson) comes in that solves this problem really well, but adds a bit to bundle size.

As the docs suggest, you can still implement the custom comparison via `config.isDataEqual`

PR #3666, opened 2022-06-02
fix(types): remove non-void constraint from queryFn result

artysidorenko:
discussed in #3541 

trying to enforce not-void + not-undefined on the queryFn return type is creating pockets of complexity / unexpected typing behaviour, as TS is trying to both enforce and infer too many things at the same time.

removing for now, and adding some additional tests to check for the pieces that were clashing with the non-void constraint
- able to use fetch() in queryFn with `Promise<any>` return type
- able to use wrapped queries with custom fetcher functions typed to return `TQueryFnData`


TkDodo (approved):
thank you 🙏

PR #3669, opened 2022-06-04
fix: `replaceDeepEqual` special case for non-plain arrays

jxom:
It turns out that the `replaceDeepEqual` function wasn't returning the intended value for objects that are non-plain arrays.

E.g.

```tsx
const prev = Object.assign([1, 2], { a: { b: 'b' }, c: 'c' })
const next = Object.assign([1, 2], { a: { b: 'b' }, c: 'c' })

replaceDeepEqual(prev, next) // result: [1, 2]
```

This data structure is used in libraries such as [ethers](https://docs.ethers.io/v5/) and their [contract `Result` value](https://docs.ethers.io/v5/api/utils/abi/interface/#Result).

I'm not sure if we should perform structural sharing on these values as they are not really "JSON-compatible" values as mentioned in the [docs](https://react-query.tanstack.com/guides/important-defaults), so I added a change which just checks if the given array is a "plain array", and if it is not, it has changed.

### DIFF
--- a/src/core/utils.ts
+++ b/src/core/utils.ts
@@ -314,7 +314,7 @@ export function replaceEqualDeep(a: any, b: any): any {
     return a
   }
 
-  const array = Array.isArray(a) && Array.isArray(b)
+  const array = isPlainArray(a) && isPlainArray(b)
 
   if (array || (isPlainObject(a) && isPlainObject(b))) {
     const aSize = array ? a.length : Object.keys(a).length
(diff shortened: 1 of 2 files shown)


src/core/utils.ts
```diff
@@ -355,6 +355,10 @@ export function shallowEqualObjects<T>(a: T, b: T): boolean {
   return true
 }
 
+export function isPlainArray(a: any) {
+  return Array.isArray(a) && a.length === Object.keys(a).length
+}
```
TkDodo, line 360: can you also add a separate test for this function please?

src/core/utils.ts
```diff
@@ -355,6 +355,10 @@ export function shallowEqualObjects<T>(a: T, b: T): boolean {
   return true
 }
 
+export function isPlainArray(a: any) {
```
TkDodo, line 358: a can probably be of type `unknown`. Also, I would appreciate it if we rename it to something like `value`:

```
export function isPlainArray(value: unknown) {

jxom:
@TkDodo your wishes have been granted

TkDodo (approved):

PR #3672, opened 2022-06-06
fix(persist): introduce in-house synchronous Storage interface

todorone:
### Rationale
Currently RQ has `createWebStoragePersistor` and `createAsyncStoragePersistor` which assumes that non-web platforms can have async storages only, which is not true. React Native has performant synchronous MMKV adapters support, which can't be used with RQ 3 right now directly because of interface incompatibility and `window` check(already fixed in v4).
Related discussion: https://github.com/tannerlinsley/react-query/discussions/3667

### Changes
Introduce own `Storage` interface to make possible using `createWebStoragePersistor` in React Native without TS errors.

### DIFF
--- a/src/core/mutation.ts
+++ b/src/core/mutation.ts
@@ -178,7 +178,7 @@ export class Mutation<
         this.options.onSuccess?.(
           data,
           this.state.variables!,
-          this.state.context!
+          this.state.context
         )
       )
       .then(() =>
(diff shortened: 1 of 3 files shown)


TkDodo:
thanks. can you make a PR towards the `beta` branch ?

todorone:
> thanks. can you make a PR towards the `beta` branch ?

Oops, missed this point, sorry.
Is it ok now?

TkDodo:
there are conflicts now because we've renamed the directory - in v4, the plugin is no longer experimental

todorone:
> there are conflicts now because we've renamed the directory - in v4, the plugin is no longer experimental

sorry for a mess, created new "clean" PR

PR #3698, opened 2022-06-11
feat: slightly increase array difference performance

oas89:
Actual implementation of array difference is not optimal, we could improve performance using Set to check if the second array not contains element from the first one

Here some performance benchmarking for both implementations https://jsbench.me/w6l49kdxls/1


TkDodo:
thanks. could you solve the conflicts please?

TkDodo:
sorry this was auto closed. can you re-open the PR pointing to `main` please?

PR #3700, opened 2022-06-12
docs: stabilise query client in quick start guide

Carlwirkus:
This caused me some pain when initially learning react query. HMR would break when I added functionality like error toasts to the global onError handlers. I figure wrapping the query client in a useState hook would be sensible default.


TkDodo:
In this example, the queryClient is defined outside of the app component, making it referentially stable. There is no need to copy it to useState.

PR #3757, opened 2022-06-28
docs: update invalidation behavior

41y08h:


docs/guides/query-invalidation.md
```diff
@@ -17,7 +17,7 @@ queryClient.invalidateQueries(['todos'])
 When a query is invalidated with `invalidateQueries`, two things happen:
 
 - It is marked as stale. This stale state overrides any `staleTime` configurations being used in `useQuery` or related hooks
-- If the query is currently being rendered via `useQuery` or related hooks, it will also be refetched in the background
+- If the query is currently being rendered via `useQuery` or related hooks, it will be refetched, even if in the background
```
TkDodo, line 20: the wording is correct as it is. it means: the query will be "refetched in the background", meaning it will happen while the user still sees the old data.

PR #3762, opened 2022-06-29
fix(core): added in timer and interval id checks before clearing

dwilt:
[`clearInterval` and `clearTimeout` are expensive to run](https://jsbench.me/mjl4ybxzwe/1) and should only be ran if there is for sure an id to be cleared. This PR adds the checks in as discussed in #3756

### DIFF
--- a/src/core/query.ts
+++ b/src/core/query.ts
@@ -211,8 +211,10 @@ export class Query<
   }
 
   private clearGcTimeout() {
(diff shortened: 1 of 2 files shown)


TkDodo:
prettier is failing

dwilt:
@TkDodo should be fixed now

TkDodo (approved):

PR #3781, opened 2022-07-06
Upgrading use-sync-external-store version

sanjaiyan-dev:


TkDodo:
Why is there no change in yarn.lock 🤔

TkDodo:
according to our lock files, we already resolve to `1.2.0`, so this change is not needed.

PR #3786, opened 2022-07-08
fix: add react-dom as peer dependency

### LINKED ISSUE #3534: Missing dependency on react-dom leads to runtime errors when using yarn berry
### Describe the bug

When using react-query in a project using yarn berry (3+), in combination with workspace, then dependencies are no longer shared across the project and instead only direct dependencies of each library are visible to it.

As a result, using react-query from a next.js project and client side rendering fails as it is unable to import react-dom, which does not appear as a dependency, or a peer dependency, in react query.

It is possible to work around this by making react and react-dom dependencies of the top level project in a yarn workspace.

N/A

### Steps to reproduce

1. Create a new yarn 3 workspace, with a module containing a next.js project with dependency on react-query
2. Attempt to use react-query in a client side component of this project
3. Observe that react-query fails to find react-dom, as it is not a dependency of react-query and thus invisible to it under yarn 3 resolution rules

### Expected behavior

React-dom is resolved as dependency of react-query, or alternatively, a react-dom specific version of / addon to react-query is required in a react-dom environment.

Every time

### Screenshots or Videos

_No response_

### Platform

- Linux Ubuntu 18, any browser

### react-query version

3.35.0

### TypeScript version

4.6.3

### Additional context

_No response_

duarten:
If missing, then strict package managers like pnpm will not link this
react-dom in, causing react-query to fail at runtime cause it can't load
react-dom.

Fixes #3534


duarten:
> thanks. I guess we need `react-native` as well then, because we have both `react-dom` and `react-native` marked as `optional` in `peerDependenciesMeta`, right ?

Yeah, that's right. I can add it to the PR, but I'm not sure what the version selector should be?

TkDodo:
> Yeah, that's right. I can add it to the PR, but I'm not sure what the version selector should be?

yeah me neither. @arnaudbzn might know, or @hirbod ?

hirbod:
Depends. Expo users need 0.68.2, while RN is available at 0.69.1.
Expo SDK will release SDK 46 soon, which will also be at 0.69.1 (or .2)

hirbod:
It's hard to say. I think folks might end up with the need to add a yarn resolution when different versions being installed.

TkDodo:
are optional peer dependencies also installed automatically ? I didn't think so, but not sure 🤔

TkDodo:
see: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta

> Marking a peer dependency as optional ensures npm will not emit a warning if the soy-milk package is not installed on the host. This allows you to integrate and interact with a variety of host packages without requiring all of them to be installed.

hirbod:
Yeah, that was something similar he was tweeting. Most people did the mistake to add it as devDependencY instead of peer. Adding * should be the correct and safe way as peer

TkDodo:
thanks. @duarten then let's do `react-native: '*'` please as a peer dependency

duarten:
Done!

satya164:
Keep in mind that adding `react-native` as a peer dependency means people who don't use react-native and use `npm` will get `react-native` installed since npm auto-installs missing peer deps. If they use `yarn` then they'll get a missing peer dep warning. Same for `react-dom` and `react-native` users.

I'd suggest not adding this peer dep. I haven't used `pnpm` but afaik `pnpm` users could add some extra config to make this work: https://github.com/pnpm/pnpm/issues/4183#issuecomment-1008045670 or maybe there's another way to specify this.

TkDodo (approved):

TkDodo:
sorry @satya164 I just saw your comment right now. I was under the impression that since the dependencies are marked as `optional` in `peerDependenciesMeta`, they would _not_ be auto installed with `npm`. Is that not right?

satya164:
@TkDodo ah I right, I hadn't checked the full file, yes that should be fine then.

ingvaldlorentzen:
Seems like `npm` sadly does auto-install optional `peerDepedencies`. There is a discussion about it on in the `npm/feedback` repo. Though there does seem to be some confusion as to what `npm` is supposed to do, and wether this is a bug or not. 

I would suggest `react-query` remove `react-native` and `react-dom` as `peerDependencies`, i.e. reverse this PR, until this is solved in `npm`. As installing `react-query` in a web project now, results in `react-native` being installed, which adds a lot of potential issues to otherwise simple projects.

PR #3796, opened 2022-07-11
Fix the npm installation snippet

domino2:
The `@tanstack/react-query` is not found and npm ended with to 404.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -10,7 +10,7 @@ or a good ol' `<script>` via
 ### NPM
 
 ```bash
-$ npm i @tanstack/react-query
+$ npm i react-query
 ```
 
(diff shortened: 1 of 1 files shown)


TkDodo:
we plan to release it soon

domino2:
I noticed that there are other miss leads in the documentation. Will the release fix the others as well?

TkDodo:
i hope so :)

which ones are you referring to, specifically ?

domino2:
After that I committed I found other i the "Quick start"
```
import {
  useQuery,
  useMutation,
  useQueryClient,
  QueryClient,
  QueryClientProvider,
} from '@tanstack/react-query'
```
Where it should by `import ... from 'react-query'`.

TkDodo:
yes. after the release, the imports will work as described in the docs.

PR #3798, opened 2022-07-12
fix intallation script and added script for Yarn

Akash187:
The previous script is broken. Got package not found Error.


TkDodo:
we plan to release it soon. there is an open issue until then

PR #3801, opened 2022-07-13
feat(codemod): add codemod that replaces the react-query import specifiers

TkDodo:


TkDodo:
the build errors are fixed in another PR :/

PR #3829, opened 2022-07-15
feat: determine whehter suspend by user

liaoliao666:
sometimes I wanna suspend the component when the query is fetching, It make sense when the data as initialValues in the form scene. but I can not found any way to do that.
I think It should give the control to developer.


liaoliao666:
Maybe u would say that u can set the cacheTime to 0. But I have I have another complex scene. like below

when I push to the new page it should be suspend and when go back the old page it should not be suspend. So I wanna control by myself.

TkDodo:
I'm not against the idea, but I don't like the interface of another option. What we have been doing instead is to make boolean properties also be a function instead, like:

```
refetchOnWindowFocus: boolean | "always" | ((query: Query) => boolean | "always")
 ```

we could do the same with suspense I think?

```
suspense: boolean | ((query: Query) => boolean)
```

`useErrorBoundary` also supports this syntax already.

liaoliao666:
@TkDodo I completely understand what u mean. but what i want is control the suspend behavior of the suspense and not just set the suspense true or false that make no sense to me.

TkDodo:
not sure I understand the difference. If suspense is a function, we can invoke it in every render to determine if it should throw the promise or not...

liaoliao666:
@TkDodo oh Its my fault. If so,  I agree that ur suggestion is better😂

TkDodo:
yes, looks good. please remove the `yarn.lock` file that you added, thx.

liaoliao666:
done

TkDodo:
sorry this was auto closed. can you re-open the PR pointing to `main` please?

PR #3833, opened 2022-07-16
Update link to the `network-mode` page

GabrielKastenbaum:
The present link is a 404.
https://tanstack.com/query/v4/docs/guides/network-mode


TkDodo:
I'm glad for everyone who tries to fix the broken links. What I would really like to have is _relative_ links. linking to `https://tanstack.com/query/v4/` will break if we go to v5!

I'm not entirely sure why the relative links are broken. It seems that going to `../guides/network-mode` works. I'm not sure why though.

If you want, please replace all links that go to `./` with `../` + the respective slug. Also, there are some absolute links (those that start with `/`) that do not work and would need to be replaced.

If you want to contribute that, go ahead. If not, I'll close this PR. Thank you

GabrielKastenbaum:
I'll make a more complete PR. Just like you I could not completely understand why the relative link with `./Network-mode` does not work.

PR #3834, opened 2022-07-16
Added a proof of concept for initialData making TData never undefined

### LINKED ISSUE #3310: better type narrowing for initialData

mattpocock:
closes #3310

### DIFF
--- a/packages/react-query/src/useQuery.ts
+++ b/packages/react-query/src/useQuery.ts
@@ -4,7 +4,7 @@ import {
   QueryKey,
   QueryObserver,
 } from '@tanstack/query-core'
-import { UseQueryOptions, UseQueryResult } from './types'
+import { DefinedUseQueryResult, UseQueryOptions, UseQueryResult } from './types'
 import { useBaseQuery } from './useBaseQuery'
 
 // HOOK
@@ -31,7 +31,7 @@ export function useQuery<
     UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
     'initialData'
   > & { initialData: TData | (() => TData) },
-): Omit<UseQueryResult<TData, TError>, 'data'> & { data: TData }
+): DefinedUseQueryResult<TData, TError>
 
 export function useQuery<
   TQueryFnData = unknown,
@@ -57,7 +57,7 @@ export function useQuery<
     UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
     'queryKey' | 'initialData'
   > & { initialData: TData | (() => TData) },
-): Omit<UseQueryResult<TData, TError>, 'data'> & { data: TData }
+): DefinedUseQueryResult<TData, TError>
 
 export function useQuery<
   TQueryFnData = unknown,
@@ -85,7 +85,7 @@ export function useQuery<
     UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
     'queryKey' | 'queryFn' | 'initialData'
   > & { initialData: TData | (() => TData) },
-): Omit<UseQueryResult<TData, TError>, 'data'> & { data: TData }
+): DefinedUseQueryResult<TData, TError>
 
 export function useQuery<
   TQueryFnData,
--- a/packages/react-query/src/useQuery.ts
+++ b/packages/react-query/src/useQuery.ts
@@ -30,7 +30,7 @@ export function useQuery<
   options: Omit<
     UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
     'initialData'
-  > & { initialData: TData | (() => TData) },
+  > & { initialData: TQueryFnData | (() => TQueryFnData) },
 ): DefinedUseQueryResult<TData, TError>
 
 export function useQuery<
@@ -56,7 +56,7 @@ export function useQuery<
   options?: Omit<
     UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
     'queryKey' | 'initialData'
-  > & { initialData: TData | (() => TData) },
+  > & { initialData: TQueryFnData | (() => TQueryFnData) },
 ): DefinedUseQueryResult<TData, TError>
 
 export function useQuery<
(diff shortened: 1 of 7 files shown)


packages/react-query/src/useQuery.ts
```diff
@@ -15,8 +15,24 @@ export function useQuery<
...
+export function useQuery<
+  TQueryFnData = unknown,
+  TError = unknown,
+  TData = TQueryFnData,
+  TQueryKey extends QueryKey = QueryKey,
+>(
+  options: Omit<
+    UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+    'initialData'
+  > & { initialData: TData | (() => TData) },
```
TkDodo, line 33: this needs to be `TQueryFnData` as `TData` is the type that runs through select if it exists:

```suggestion
  > & { initialData: TQueryFnData | (() => TQueryFnData) },
```

packages/react-query/src/useQuery.ts
```diff
@@ -26,9 +42,23 @@ export function useQuery<
...
+  TQueryFnData = unknown,
+  TError = unknown,
+  TData = TQueryFnData,
+  TQueryKey extends QueryKey = QueryKey,
+>(
+  queryKey: TQueryKey,
+  options?: Omit<
+    UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+    'queryKey' | 'initialData'
+  > & { initialData: TData | (() => TData) },
```
TkDodo, line 59:
```suggestion
  > & { initialData: TQueryFnData | (() => TQueryFnData) },
```

packages/react-query/src/useQuery.ts
```diff
@@ -39,9 +69,24 @@ export function useQuery<
...
+  TError = unknown,
+  TData = TQueryFnData,
+  TQueryKey extends QueryKey = QueryKey,
+>(
+  queryKey: TQueryKey,
+  queryFn: QueryFunction<TQueryFnData, TQueryKey>,
+  options?: Omit<
+    UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+    'queryKey' | 'queryFn' | 'initialData'
+  > & { initialData: TData | (() => TData) },
```
TkDodo, line 87:
```suggestion
  > & { initialData: TQueryFnData | (() => TQueryFnData) },
```

TkDodo:
@all-contributors add @mattpocock for code

Alcas1:
Is this going to be in V3 as well or only for V4?

TkDodo:
@Alcas1 only on v4. please see the issue description for the reasoning behind this:
- https://github.com/TanStack/query/issues/3310

Console-buche (commented):
nothing (mistake)

moris-moze:
@TkDodo Hi, in what specific v4 version is this available? I've updated to [4.0.0-beta.23], but the `data` type still remains `T | undefined`. Thanks

TkDodo:
just use latest v4 ?

moris-moze:
@TkDodo Looking at https://www.npmjs.com/package/react-query/v/4.0.0-beta.23?activeTab=versions, 4.0.0-beta.23 seems to be the latest?

TkDodo:
the package is `@tanstack/react-query`:

https://www.npmjs.com/package/@tanstack/react-query

moris-moze:
@TkDodo Is it possible to make a query run only once, while having `initialData` set and `staleTime` set to Infinity?

E.g.:

```
export const useAllUsers = () =>
  useQuery<UserResponseDto[]>({
    queryKey: [QueryClientKeys.GET_ALL_USERS],
    queryFn: async () => {
      const response = await http.get<
        UserResponseDto[],
        AxiosResponse<UserResponseDto[]>
      >(`${API_BASE}/api/user/all`);
      return response.data;
    },
    staleTime: Infinity,
    initialData: [],
  });
```

The goal here is not to have possibly undefined data, hence the `initialData`. Besides that, I want to run the query only once so there's the `staleTime: Infinity`. This obviously can't work since query has initial data and stale time is infinite, so there is no triggering of the queryFn. Is there a way around this?

TkDodo:
> @TkDodo Is it possible to make a query run only once, while having initialData set and staleTime set to Infinity?

maybe with `refetchOnMount: 'always'` ? But it will fetch whenever a new observer mounts

PR #3861, opened 2022-07-19
fix(react-query-devtools): only export devtools in development mode

### LINKED ISSUE #3860: @tanstack/react-query-devtools v4 does not include `NODE_ENV` check
### Describe the bug

According to the documention, devtools should only be included in `development` mode:
https://tanstack.com/query/v4/docs/devtools#install-and-import-the-devtools

React Query v3 devtools include check for that:
https://github.com/TanStack/query/blob/v3/devtools/index.js

React Query v4 are missing this check which results in devtools being bundled in production and increasing overall bundle size:
https://github.com/TanStack/query/blob/main/packages/react-query-devtools/src/index.ts

https://stackblitz.com/edit/nextjs-odbs3c?file=pages/_app.js

React Query v3 does not have this issue:
https://stackblitz.com/edit/nextjs-ssq6g3?file=pages/_app.js

### Steps to reproduce

1. Open `_app.js` file
2. Open terminal and run `npm run build`
3. Comment out `<ReactQueryDevtools initialIsOpen={false} />` and `import { ReactQueryDevtools } from '@tanstack/react-query-devtools';`
4. Run `npm run build`
5. Notice that shared bundle size decreased from `103kB` to `90kB`

### Expected behavior

React Query Devtools should not be bundled in production build by default.

Every time

### Screenshots or Videos

Next 12.2 project with devtools enabled:
image

Next 12.2 project with devtools disabled:
image

### Platform

- OS: MacOS
- Browser: Chromium
- Version: 4.0.3

### react-query version

4.0.3

### TypeScript version

4.7.4

### Additional context

_No response_

TkDodo:
fixes #3860


packages/react-query-devtools/src/index.ts
```diff
@@ -1 +1,12 @@
-export * from './devtools'
+if (process.env.NODE_ENV !== 'development') {
```
satya164, line 1: I find it's better to do `process.env.NODE_ENV === 'production'` since it matches with the behaviour of most tools (e.g. npm recognizes `production`) etc. and also what tools usually recommend to set for production (e.g. webpack). While most do set `development` in case it's not production, in case someone sets it to a different one (say for building with staging API etc.) then it'll cause different behaviour between other tools and React Query.

TkDodo: we can do that, I just got this from the v3 implementation. Problem is that this doesn't really work with ESM. If we take a look at the codesandbox build:
- https://codesandbox.io/s/tanstack-query-example-react-basic-hofver

we can see the error:

```
Could not find module in path: '@tanstack/react-query-devtools/build/esm/devtools' relative to '/node_modules/@tanstack/react-query-devtools/build/esm/index.js'
```

I think the cjs builds are fine, and in the umd builds, I can see that it is properly there for dev builds and removed for prod builds. But I don't think this is how it can work with esm builds. Any ideas?

satya164: @TkDodo afaik with webpack, we should be able to do something like this, in the package.json of devtools:

```js
  "exports": {
    "production": "./index-without-devtools.js",
    "default": "./index-with-devtools.js"
  }
```

https://webpack.js.org/guides/package-exports/

I haven't tried it, but looks like it would work.

TkDodo: right, we used to have this in the v4 beta, but it got lost during the monorepo move. let me see if I can restore it.

tannerlinsley: As per usual, esm 💩 -ing on the party. Let's try our best on the esm, but if we need to ship a version with the others fixed while we figure it out, we can.

TkDodo: @sachinraja fyi

packages/react-query-devtools/src/index.ts
```diff
@@ -1 +1,16 @@
...
+  function () {
+    return null
+  }
+
+export let ReactQueryDevtoolsPanel: typeof devtools['ReactQueryDevtoolsPanel'] =
+  function () {
+    return null
+  } as any
+
+if (process.env.NODE_ENV !== 'production') {
```
MattyBalaam, line 13: I believe this will cause issues when the environment is 'test'. 

Perhaps something like this?

`!['production', 'test'].includes(process.env.NODE_ENV)`

packages/react-query-devtools/src/index.ts
MattyBalaam: @satya164 @TkDodo in case my comment gets lost, `process.env.NODE_ENV === 'production' will cause the dev tools to be run when running tests in test which uses 'test' for NODE_ENV

packages/react-query-devtools/src/index.ts
satya164: `!['production', 'test'].includes(process.env.NODE_ENV)` may not work since minifier needs simple logic to do DCE. `process.env.NODE_ENV !== 'production' || process.env.NODE_ENV !== 'test'` may work but it's usually better to bundle in `production` mode when running e2e tests to avoid development related stuff in test.

Many other tools use `process.env.NODE_ENV !== 'production'` (e.g. redux: https://github.com/reduxjs/redux-devtools/blob/a8883f287de2e36a12a4058635f8474cda34a0d4/packages/redux-devtools-extension/src/logOnlyInProduction.ts#L22) so a custom env will leave mix of dev and production code in the bundle.

MattyBalaam: So, using 4.0.5 I am seeing this issue because react-testing-library uses `jsdom` which is currently failing because `window.matchMedia` is not supported:
Screenshot 2022-07-20 at 11 54 39

TkDodo: the latest attempt doesn't use an env check at all

TkDodo: okay I think I got it now - added the !== development env check again for commonjs builds, and esm builds now use the separate `exports` field in package.json

PR #3868, opened 2022-07-19
feat: allow custom merge function for structuralSharing

richsilv:
As discussed [here](https://github.com/TanStack/query/discussions/3715) and [here](https://github.com/TanStack/query/discussions/3690), this allows the provision of a function as the `structuralSharing` value, which effectively acts as a replacement for `replaceEqualDeep`. This way, consumers can still achieve the performance benefits of retained references even when cache data contains non-serializable values.

### DIFF
--- a/packages/query-core/src/tests/queryClient.test.tsx
+++ b/packages/query-core/src/tests/queryClient.test.tsx
@@ -13,6 +13,7 @@ import {
   QueryClient,
   QueryFunction,
   QueryObserver,
+  QueryObserverOptions,
 } from '..'
 import { focusManager, onlineManager } from '..'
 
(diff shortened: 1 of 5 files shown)


packages/query-core/src/tests/queryClient.test.tsx
```diff
@@ -357,6 +358,38 @@ describe('queryClient', () => {
...
+          newData: { value: Date },
+        ) => {
+          if (!prevData) {
+            return newData
+          }
+          return newData.value.getTime() === prevData.value.getTime()
+            ? prevData
+            : newData
+        },
+      } as QueryObserverOptions
```
richsilv, line 376: It's not clear to me how I could avoid type casting here, since `setDefaultOptions` takes `DefaultOptions`, which has `QueryObserverOptions<unknown, TError>` as the type of `query`, which isn't going to match any function which accepts a concrete type for `TData`. I think you have exactly the same problem if you try to update `onSuccess` this way, so the problem isn't limited to the change introduced in this PR.

TkDodo: can you just inline the object? like:

```
queryClient.setDefaultOptions({ queries: { structuralSharing: () ... } })
```

or use a const assertion: `as const` instead of `as QueryObserverOptions`

TkDodo: will you still address this @richsilv ?

richsilv: Sorry for the delay, fixing this up now.

richsilv: No, I'm afraid that neither of those will work. The problem is that `setDefaultOptions` is typed to take a concrete (non-generic) type, `DefaultOptions`, which has a concrete sub-type of `QueryObserverOptions` for it's `query` field with `TData` set to `unknown`. So I don't think there will be any way to make it type check when passed a callback for a query which has a non-`unknown` type for `TData` without asserting like this, as TS will always complain that the function you're providing expects something other than `unknown` for its data argument. The only other alternative is to change the typing for `DefaultOptions`, which I'm happy to try but I suspect will be a much larger diff.

As I say, this isn't restricted to the new `structuralSharing` field, it also applies to `onSuccess`.

TkDodo: ah yeah, if you use defaults, there is no way to know which value will be provided for data, because it can and will be used for multiple queries with different structures.

I'm not really sure why the `as QueryObserverOptions` solves it, because that one is also bound to `unknown` but 🤷

TkDodo (approved):

packages/query-core/src/tests/queryClient.test.tsx
richsilv: I think it's because by asserting as a type that's bound to `unknown` you're pretending that `TData` is `unknown` and therefore consistent with `DefaultOptions`, even though the provided function actually takes a different type. But as you say, :shrug:.

PR #3873, opened 2022-07-19
fix(react-query): allow passing options to useQuery without initialData

juanhenriquez:
On `react-query` v3, users were allowed to pass custom options to `useQuery` and `useInfiniteQuery` without needing of passing `initialData`.

```ts
function useQueryPosts(isAdmin, options?: UseQueryOptions<number, Error, number>) {
  return useQuery<number, Error, number>(["posts"], () => 10, {
    enabled: isAdmin,
    ...options
  });
}
```

With the launch of `react-query` v4, this is not possible anymore as all the overload functions of `useQuery` forces the user to pass `initialData`:

```ts
function useQueryPosts(isAdmin, options?: UseQueryOptions<number, Error, number>) {
  return useQuery<number, Error, number>(["posts"], () => 10, {
    enabled: isAdmin,
    initialData: () => null,
    ...options
  });
}
```

So, this Pull Request adds a new overload function which allow users to pass options without forcing them to pass `initialData` (as it was on v3).

### DIFF
--- a/packages/react-query/src/useQuery.ts
+++ b/packages/react-query/src/useQuery.ts
@@ -39,6 +39,15 @@ export function useQuery<
   TData = TQueryFnData,
   TQueryKey extends QueryKey = QueryKey,
 >(
+  options: UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+): UseQueryResult<TData, TError>
+
+export function useQuery<
+  TQueryFnData = unknown,
+  TError = unknown,
+  TData = TQueryFnData,
+  TQueryKey extends QueryKey = QueryKey,
+>(
   queryKey: TQueryKey,
   options?: Omit<
     UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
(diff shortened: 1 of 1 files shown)


packages/react-query/src/useQuery.ts
```diff
@@ -59,6 +59,34 @@ export function useQuery<
...
+  TData = TQueryFnData,
+  TQueryKey extends QueryKey = QueryKey,
+>(
+  queryKey: TQueryKey,
+  queryFn: QueryFunction<TQueryFnData, TQueryKey>,
+  options?: Omit<
+    UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+    'queryKey' | 'queryFn'
+  >,
+): DefinedUseQueryResult<TData, TError>
```
TkDodo, line 88: this would need to be `UseQueryResult`:

```suggestion
): UseQueryResult<TData, TError>
```

packages/react-query/src/useQuery.ts
```diff
@@ -59,6 +59,34 @@ export function useQuery<
   > & { initialData: TQueryFnData | (() => TQueryFnData) },
 ): DefinedUseQueryResult<TData, TError>
 
+export function useQuery<
```
TkDodo, line 62: overload ordering matters, so please put these two as the very last overloads.

TkDodo (changes requested):

packages/react-query/src/useQuery.ts
juanhenriquez: Missed this! Just moved them to be the last overloads.

packages/react-query/src/useQuery.ts
juanhenriquez: I added 2 overloads to follow your pattern, one that uses `UseQueryResult` and the other one uses `DefinedUseQueryResult`. Should I just leave the one with `UseQueryResult`?

TkDodo: the result is `Defined` if we have an `initialData` other than undefined, or a function that returns something that is not undefined.

so I think we would need one last overload for each "version" (the one with 3, 2 and one parameter) that returns a normal `UseQueryResult` - meaning 3 new overloads in total.

juanhenriquez: Cool, I just added them 😄. Could you please check?

TkDodo: looks right 🚀

PR #3876, opened 2022-07-20
Fix a typo and add install command

spiftire:


docs/plugins/createAsyncStoragePersister.md
```diff
@@ -6,6 +6,13 @@ title: createAsyncStoragePersister
 ## Installation
 
 This utility comes as a separate package and is available under the `'@tanstack/query-async-storage-persister'` import.
+```bash
+npm install @tanstack/query-async-storage-persister @tanstack/react-query-persist-client
+```
+or
+```bash
+yarn add @tanstack/query-async-storage-persister @tanstack/react-query-persist-client
+```
```
TkDodo, line 15: could you add the same for the sync storage persister?

spiftire: Will do

PR #3878, opened 2022-07-20
Change Query Keys Url

SamTLT:
Current url doesn't work. Replaced it with working one.

### DIFF
--- a/docs/guides/query-functions.md
+++ b/docs/guides/query-functions.md
@@ -65,7 +65,7 @@ function fetchTodoList({ queryKey }) {
 
 The `QueryFunctionContext` is the object passed to each query function. It consists of:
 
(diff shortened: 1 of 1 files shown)


docs/guides/query-functions.md
```diff
@@ -65,7 +65,7 @@ function fetchTodoList({ queryKey }) {
 
 The `QueryFunctionContext` is the object passed to each query function. It consists of:
 
-- `queryKey: QueryKey`: [Query Keys](./query-keys)
+- `queryKey: QueryKey`: [Query Keys](./guides/query-keys)
```
TkDodo, line 68: I think it should actually be:

```suggestion
- `queryKey: QueryKey`: [Query Keys](query-keys)
```

PR #3883, opened 2022-07-20
fix: add "require" to "exports" in package.json of devtools

jaworek:
Fresh NextJS project does not build after #3861 was merged. It looks like this is due to missing "require" path in "exports".

I'm not 100% sure why it is needed, but this suggestion from @satya164 seems to resolve the issue.

Repro: https://stackblitz.com/edit/nextjs-odbs3c

image

### DIFF
--- a/packages/react-query-devtools/package.json
+++ b/packages/react-query-devtools/package.json
@@ -22,6 +22,7 @@
     "development": {
       "default": "./build/esm/index.js"
     },
+    "require": "./build/cjs/packages/react-query-devtools/src/index.js",
     "default": "./build/esm/noop.js"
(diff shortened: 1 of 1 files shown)


TkDodo:
Thanks 🙏 
I'm a noob in all this bundling things and it clearly shows 🙈

TkDodo:
Looks like this PR is a more complete fix?
- https://github.com/TanStack/query/pull/3884

TkDodo:
I'll merge the other PR. Thank you 🙏

jaworek:
Awesome! I agree that the other PR is better.

PR #3884, opened 2022-07-20
fix(react-query-devtools): support cjs in noop build

Songkeys:
Just noticed that someone also made a PR #3883 . But we should export the noop version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,6 @@
     "test:jest:dev": "jest --config ./jest.config.ts --watch",
     "test:size": "npm run build && bundlewatch",
     "build": "rollup --config rollup.config.js && npm run typecheck",
-    "postbuild": "cp ./packages/react-query-devtools/src/noop.ts ./packages/react-query-devtools/build/esm/noop.js",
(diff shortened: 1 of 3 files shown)


TkDodo:
Do we need your PR, the other PR, or both?

Songkeys:
Yes. The other PR pointed out the issue but doesn't export for noop actually. This PR should fix them all.

TkDodo:
Thank you very much 🙇‍♂️

TkDodo:
@Songkeys could it be that we point to the wrong direction for the default require exports. It points towards:

```
"require": "./build/cjs/packages/react-query-devtools/src/noop.js",
```

but looking at the build directory when I create a build locally, `noop.js` is in :

`build/cjs/packages/react-query-devtools/noop.js`

so it's _not_ inside the src directory...

Songkeys:
@TkDodo Oops. I'm sorry about this. You are right. Is it a fix yet? Or do you want me to submit another PR?

TkDodo:
another PR would be great, thanks

Issue #3855, opened 2022-07-19
Module '"@tanstack/react-query"' has no exported member 'QueryClient'


Paultje52:
### Describe the bug

When trying to import `QueryClient`, TypeScript says that `@tanstack/react-query` doesn't export `QueryClient`. The QuickStart guide clearly mentiones that `QueryClient` needs to be imported (https://tanstack.com/query/v4/docs/quick-start).

### Your minimal, reproducible example

It's just one line: `import { QueryClient } from "@tanstack/react-query";`

### Steps to reproduce

1. Install `@tanstack/react-query`
2. Try to import `QueryClient`

### Expected behavior

`QueryClient` should be available

### How often does this bug happen?

Every time

### Screenshots or Videos

![image](https://user-images.githubusercontent.com/31313717/179708472-cf7e624c-439d-4256-8dc6-b6868f62ea2e.png)

```ts
Type error: Module '"@tanstack/react-query"' has no exported member 'QueryClient'.

  3 | import { Fragment, useState } from "react";
  4 | import Head from "next/head";
> 5 | import { QueryClient } from "@tanstack/react-query";
    |          ^
  6 | 
  7 | function MyApp({ Component, pageProps }: AppProps) {
  8 |   // const [ queryClient ] = useState(() => new QueryClient());
```

### Platform

Every platform

### react-query version

4.0.3

### TypeScript version

4.7.4

### Additional context

Why are there so many problems regarding TypeScript? Yesterday I couldn't use react-query either, because the typings weren't exported. How can I be confident that no more issues occur, when two massive oversights regarding TypeScript already happend?

Karol-Perec:
Try to reinstall your node_modules, this issue should not appear since 4.0.3

TkDodo:
We can also see in the official examples that this works now:

https://codesandbox.io/s/github/tanstack/query/tree/main/examples/react/simple

Paultje52:
> We can also see in the official examples that this works now:
> 
> https://codesandbox.io/s/github/tanstack/query/tree/main/examples/react/simple

The official example isn't working either when converted to TypeScript.
![image](https://user-images.githubusercontent.com/31313717/179728892-d483c0cf-4497-49f4-9cfc-7db57ed743a7.png)



> Try to reinstall your node_modules, this issue should not appear since 4.0.3

[...]

PR #3889, opened 2022-07-21
Update package.json

HXavS:
-beta.1 not published to NPM and v4 no longer in beta

### DIFF
--- a/packages/react-query/package.json
+++ b/packages/react-query/package.json
@@ -43,7 +43,7 @@
     "react-error-boundary": "^3.1.4"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


TkDodo:
I don't think it matters much, as we are depending on a package from within the same monorepo. As you can see in `package-lock.json`, it is resolved to a file reference:

```
    "@tanstack/query-core": {
      "version": "file:packages/query-core"
    },
```

I've verified this by changing something in the query-core, then running a build, then looking at the output in the react-query bundle, and the change was reflected.

HXavS:
Okay, wasn't sure was having this issue #3855, but fixed now

Issue #3882, opened 2022-07-20
ESM support


TkDodo:
we had it in v4 beta, but it went away with the rebranding.

prior art:

- https://github.com/TanStack/query/pull/3521

there were some useSyncExternalStore issues reported after that, which prompted another fix:

- https://github.com/TanStack/query/pull/3601

TkDodo:
@sachinraja FYI

TkDodo:
let's try to keep discussion about ESM support here :)

I tried out `4.0.11-beta.0` and it works fine in our simple example, but that one uses `vite`:
- https://codesandbox.io/s/amazing-carson-7fqsfe

There have been reports on discord that this version "breaks the app", but we'd need more info and try out more bundlers I guess.

cc @DamianOsipiuk

sachinraja:
@TkDodo Yeah if they share reproductions or at least error messages, I can look into it.

TkDodo:
@sachinraja I can confirm that `4.0.11-beta.0` doesn't work with react-scripts v4. Here is an (older) repo of mine that uses v4 of react-scripts, and I've merely switched from `4.1.0` where it works (but doesn't render devtools at all as per the issue) to the beta:

- https://github.com/TkDodo/testing-react-query/blob/feature/breaking-react-query/package.json#L16-L17

what I'm seeing when I run it is this:

![Screenshot 2022-08-12 at 20 06 57](https://user-images.githubusercontent.com/1021430/184418172-af476704-1cf3-4d46-8e40-179a6e4f8d41.png)

I can also confirm that removing the rendering of the `<ReactQueryDevtools initialIsOpen />` makes the error go away.

TkDodo:
Another thing I found out with @DamianOsipiuk is that the size of the beta is 5x times too large:

https://bundlephobia.com/package/@tanstack/react-query@4.0.11-beta.0

I think @DamianOsipiuk mentioned that react-dom is included because it wasn't specified as external. Just wanted to note all known issues here :)

sachinraja:
I've seen that Provider error, it has to do with one package importing the cjs format and the other package importing the esm format.

TkDodo:
@DamianOsipiuk I tried the latest beta (4.0.11-beta.4) with `react-scripts` v4 and it now fails with:

![Screenshot 2022-08-18 at 14 06 25](https://user-images.githubusercontent.com/1021430/185390801-70f56d56-27af-48d3-bda1-f127cb40eb9c.png)


```
[...]

PR #3892, opened 2022-07-21
fix: restore missing `exports` declarations

DamianOsipiuk:
Closes: https://github.com/TanStack/query/issues/3882


TkDodo:
Thanks. Are you sure that this is all we need? According to @sachinraja, there is more work to be done to be fully ESM compliant - e.g. to use the `.mjs` extension. That's why I've created this issue:

- https://github.com/TanStack/query/issues/3882

We've also had some problems with `useSyncExternalStore` and react-native, so I'd rather not rush this issue. Also not sure how we can properly test this with various bundlers before we release 🤔 

I would also really like to get the feature back that allowed us to import the devtools in production [via a separate entry point](https://github.com/TanStack/query/pull/3521/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R47-R50)

DamianOsipiuk:
@TkDodo Yup, marked as WIP for now as there are still more things to be moved/refactored.

DamianOsipiuk:
@TkDodo @tannerlinsley So most of the things are covered by this PR to restore EMS compliant exports.

The only thing left is this (https://github.com/TanStack/query/pull/3521):
```
Types are now emitted into lib instead of a separate types directory. This will help support TypeScript, which will add support for "exports" in the next release.
```

I'm not entirely sure why some packages are build to `cjs` with subdirectories, while others are not.
I assume that packages depending on other packages from monorepo have that behavior.
However it's kind of weird that `react-query/build/cjs` has also `query-core` build copy.
Is that something to be expected?

packages/react-query/package.json
```diff
@@ -10,10 +10,17 @@
...
   "browser": "build/umd/index.production.js",
   "types": "build/types/packages/react-query/src/index.d.ts",
+  "exports": {
+    ".": {
+      "import": "./build/lib/index.mjs",
+      "default": "./build/lib/index.js"
+    },
+    "./package.json": "./package.json"
+  },
   "sideEffects": [
```
TkDodo, line 24: the files listed here don't exist - can we try to point them towards existing files?
right now, I think `setBatchedUpdatesFn` is treeshaked away because it's unused and the sideEffects listings are wrong

packages/query-async-storage-persister/package.json
```diff
@@ -10,10 +10,17 @@
...
   "types": "build/types/packages/query-async-storage-persister/src/index.d.ts",
+  "exports": {
+    ".": {
+      "import": "./build/lib/index.mjs",
+      "default": "./build/lib/index.js"
+    },
+    "./package.json": "./package.json"
+  },
   "files": [
     "build/*",
```
TkDodo, line 25: the build directory contains things that we probably shouldn't ship, like `stats.html` etc

TkDodo (commented):
somehow, codesandbox shows that types can't be picked up anymore:
Screenshot 2022-07-26 at 07 19 03

see this preview sandbox: https://codesandbox.io/s/tanstack-query-example-react-basic-typescript-ssd0yy?file=/src/index.tsx

TkDodo:
@sachinraja please have a look at this PR if you find the time :)

packages/react-query/package.json
DamianOsipiuk: Fixed

packages/query-async-storage-persister/package.json
DamianOsipiuk: Picked more granular paths. But this could still change if i find a way to bundle types into lib properly

PR #3893, opened 2022-07-21
fix(react-query-devtool): correct the noop entry for cjs

Songkeys:
https://github.com/TanStack/query/pull/3884#issuecomment-1191264046

Sorry about the previous PR.

The current build system is actually somehow messy. We might need to tidy it up once v4 is stable.

@TkDodo


TkDodo (approved):

TkDodo:
fully agree, thank you 🙏 

we can track improvements here:
- #3882

PR #3909, opened 2022-07-24
improve useQuery type overload

mh-alahdadian:


TkDodo:
please fix the pipeline (prettier)

DamianOsipiuk:
@mh-alahdadian I believe this types improvement requires Typescript 4.2 to work right?
Currently we have [Types currently require using TypeScript v4.1 or greater](https://tanstack.com/query/v4/docs/typescript)
Therefore this would be a breaking change and we would need to push it out to next major version.

mh-alahdadian:
> @mh-alahdadian I believe this types improvement requires Typescript 4.2 to work right? Currently we have [Types currently require using TypeScript v4.1 or greater](https://tanstack.com/query/v4/docs/typescript) Therefore this would be a breaking change and we would need to push it out to next major version.

Did you test it on TS4.1 and it doesn't work ? I don't know if it requires a special new version of TS , if it didn't worked with 4.1 then I think we can wait for next major version release

TkDodo:
Thanks for the contribution. Our plan is to remove all overloads eventually, and for now, there is no drawback to stick with the current implementation.

PR #3912, opened 2022-07-24
Migrate examples/react/simple from CRA to Vite

zorzysty:


examples/react/simple/package.json
```diff
@@ -3,26 +3,21 @@
   "private": true,
   "version": "0.0.1",
   "scripts": {
-    "start": "rescripts start",
-    "build": "rescripts build",
-    "test": "rescripts test",
-    "eject": "rescripts eject"
+    "dev": "vite",
+    "start": "vite",
```
zorzysty, line 7: `dev` script is the default way of running Vite apps. `npm run dev` is also the command used in the root of tanstack query to run dev env for packages. 

So I wonder whether to remove the `start` script or just leave it here for "historical" reasons.

TkDodo: it depends what codesandbox picks up per default 🤔

zorzysty: I believe it works fine with `dev`: https://codesandbox.io/s/qrf93

TkDodo: Then let's remove it

zorzysty: [done](https://github.com/TanStack/query/pull/3912/commits/a8ae567022fbee339150784895737feb9b3f07b9 )

PR #3914, opened 2022-07-25
Optimizing by memoizing the `<QueryClientProvider>` component using `React.memo` 🚀 

sanjaiyan-dev:
Memoizing `<QueryClientProvider>` and some other `react-query` components using `React.memo` which might improve the performance of that components :)

For more info please refer [the documentation](https://reactjs.org/docs/react-api.html#reactmemo).
_Extremely sorry, if I made any mistakes :(_


TkDodo:
memoizing components that take children doesn't do anything because children are a new referential identity in each render cycle.

PR #3915, opened 2022-07-25
Already checked in line 566.

ro0t:
(╯°□°)╯︵ ┻━┻

Typescript doesn't like it when we are checking if we're on production twice. First we check at Line 566 `if (process.env.NODE_ENV !== 'production') {` and then again @ line 573. 

```Failed to compile.

../../node_modules/@tanstack/query-core/src/queryClient.ts:573:13
Type error: This condition will always return 'true' since the types '"development" | "test"' and '"production"' have no overlap.
```

Simple fix 🥇


TkDodo:
there is another, similar check in this file that you could remove

ro0t:
> there is another, similar check in this file that you could remove

Correct! I also removed the other unnecessary check 🤝

Issue #3916, opened 2022-07-25
umd file index.development.js returns null for devtools


geraldodev:
### Describe the bug

The UMD file https://unpkg.com/browse/@tanstack/react-query-devtools@4.0.10/build/umd/index.development.js returns null for devtools.

I'm my case I'm consuming via shadow-cljs , the clojurescript compiler,
According to @tkdodo it also can affect cdn consumers.

### Your minimal, reproducible example

https://unpkg.com/browse/@tanstack/react-query-devtools@4.0.10/build/umd/index.development.js

### Steps to reproduce

1- consume the umd package via cdn
2 - render query tools component
3 - it'll render null

### Expected behavior

Query tools component render

### How often does this bug happen?

Every time

### Platform

Linux

### react-query version

4.0.10

TkDodo:
it's fixed in the beta we just released:

https://unpkg.com/browse/@tanstack/react-query-devtools@4.0.11-beta.0/build/umd/index.development.js

there are some other issues in there though, so it will be available once we generally fix this :)

PR #3924, opened 2022-07-26
fix/umd-build

JohnDaly:
UMD builds are configured to always generate output files with the format: `${packageDir}/build/umd/index.production.js` and `${packageDir}/build/umd/index.development.js`. 

This is a problem, because there are two sets of build configs for the dev tools package: `react-query-devtools` and `react-query-devtools-noop`. Since they are both generating output files with the same name, this causes `react-query-devtools-noop` to always override the output of `react-query-devtools`.

To fix this, I updated the `umdDev` and `umdProd` functions to use the `outputFile` variable to determine the name of the output file. Then I provided different `outputFile` values for `react-query-devtools` and `react-query-devtools-noop`


packages/react-query-devtools/package.json
```diff
@@ -12,7 +12,7 @@
   },
   "module": "build/esm/index.js",
   "main": "build/cjs/packages/react-query-devtools/src/index.js",
-  "browser": "build/umd/index.production.js",
+  "browser": "build/umd/noop.production.js",
```
JohnDaly, line 15: Since the CJS and ESM production build entrypoints are noop's, I updated this to also follow that pattern

rollup.config.ts
```diff
@@ -194,7 +194,7 @@ function umdDev({
     output: {
       format: 'umd',
       sourcemap: true,
-      file: `${packageDir}/build/umd/index.development.js`,
+      file: `${packageDir}/build/umd/${outputFile}.development.js`,
```
JohnDaly, line 197: Since all UMD builds were generating an output file with the `index` name, this change will only affect the output of the `react-query-devtools` package

rollup.config.ts
```diff
@@ -90,7 +90,7 @@ export default function rollup(options: RollupOptions): RollupOptions[] {
       name: 'react-query-devtools-noop',
       packageDir: 'packages/react-query-devtools',
       jsName: 'ReactQueryDevtools',
-      outputFile: 'react-query-devtools',
+      outputFile: 'noop',
```
JohnDaly, line 93: Setting this to `noop` will allow `react-query-devtools` and `react-query-devtools-noop` to generate their own output files

JohnDaly:
This change should allow for a new `react-query-devtools/build/umd/index.development.js` to be generated, which can be used to fix https://github.com/TanStack/query/issues/3916

packages/react-query-devtools/package.json
TkDodo: we now have `index.development.js`, `index.production.js`, `noop.developtment.js` and `noop.production.js` produced. I don't think that's on purpose?

rollup.config.ts
```diff
@@ -37,39 +37,39 @@ export default function rollup(options: RollupOptions): RollupOptions[] {
...
+      outputFile: 'index',
       entryFile: 'src/index.ts',
       globals: {},
     }),
     ...buildConfigs({
       name: 'query-async-storage-persister',
       packageDir: 'packages/query-async-storage-persister',
       jsName: 'QueryAsyncStoragePersister',
-      outputFile: 'query-async-storage-persister',
+      outputFile: 'index',
```
TkDodo, line 48: was `outputFile` not used at all previously?

JohnDaly: It was not used previously, from what I could tell

packages/react-query-devtools/package.json
JohnDaly: My thought was that `build/umd/index.development.js` would be distributed via CDN (to support use-cases like https://github.com/TanStack/query/issues/3916). 

I'm not entirely sure what we want to do about the `"browser"` field here. It seems that the convention for CJS and ESM production builds is to resolve to a noop for the dev tools package? The counter-point to this would be supporting bundlers like Webpack 4, which might not know about the `"exports"` field and instead use the `"browser"` field. When I tried to use the dev tools package on a project that used an older version of Webpack, I ran into that problem and couldn't use the dev tools.

TkDodo: > When I tried to use the dev tools package on a project that used an older version of Webpack, I ran into that problem and couldn't use the dev tools.

yeah me to (react-scripts v3). would be good if we could support this

TkDodo:
@JohnDaly we want to release all build related things in a pre-release version. There is also this PR that re-adds proper ESM support:
- https://github.com/TanStack/query/pull/3892

both of them should go to the `beta` branch please, and there are likely conflicts. thanks 🙏

phil-saam:
This bug is preventing me from upgrading to v4 as I currently have to use an older version of react-scripts.  When do you think this will be merged?

packages/react-query-devtools/package.json
```diff
@@ -12,7 +12,7 @@
   },
   "module": "build/lib/index.mjs",
   "main": "build/lib/index.js",
-  "browser": "build/umd/index.production.js",
+  "browser": "build/lib/index.js",
```
JohnDaly, line 15: This should allow for older bundlers (e.g. Webpack 4) to properly resolve `@tanstack/react-query-devtools`

packages/react-query-devtools/package.json
JohnDaly: I tested the change from https://github.com/TanStack/query/pull/3924/commits/d85684bb3b6e9d6b78388c46a7936130ad019690 with `react-scripts` v3 and v4. Both versions are able to load `@tanstack/react-query-devtools` properly with that change

packages/react-query-devtools/package.json
DamianOsipiuk: I would argue that we should remove `browser` field from all packages.
UMD build should be linked directly from CDN.
We do not ship separate code for `node` and `browser` (ie. xxx.server.js and xxx.client.js), so we could stick to `main` field.
Or remap all packages this way.

TkDodo:
@JohnDaly it seems that this PR increased the bundle-size of react-query by 5x 😅 

https://bundlephobia.com/package/@tanstack/react-query@4.0.11-beta.0

We can also see this in the bundlewatch logs:

`packages/react-query/build/umd/index.production.js: 43.89KB (gzip)`

size on main is:

`packages/react-query/build/umd/index.production.js: 11.37KB (gzip)`

any idea what that could be about?

JohnDaly:
> @JohnDaly it seems that this PR increased the bundle-size of react-query by 5x 😅
> 
> https://bundlephobia.com/package/@tanstack/react-query@4.0.11-beta.0
> 
> We can also see this in the bundlewatch logs:
> 
> `packages/react-query/build/umd/index.production.js: 43.89KB (gzip)`
> 
> size on main is:
> 
> `packages/react-query/build/umd/index.production.js: 11.37KB (gzip)`
> 
> any idea what that could be about?

I applied this changes from this PR onto the code in the `main` branch, and this is what I got:

`packages/react-query/build/umd/index.production.js... index.production.js ⏤  11.6 kB`

There are a few things that I noticed that are different between the `beta` branch and the `main` branch, that might be contributing to this:

### 1. The Rollup configs are different

Specifically the `main` branch doesn't include some of the `globals` that are specified in the `beta` branch.

Example `'@tanstack/query-core': 'QueryCore'` is [not present in `main`](https://github.com/TanStack/query/blob/b3201e991d8737d2a4ddbb99324f3282156a19ef/rollup.config.ts#L74), but is [in `beta`](https://github.com/TanStack/query/blob/4a4543c39ae500b537000c0cab807327d3ce5ff3/rollup.config.ts#L82):
```diff
    ...buildConfigs({
      name: 'react-query',
      packageDir: 'packages/react-query',
      jsName: 'ReactQuery',
      outputFile: 'index',
      entryFile: 'src/index.ts',
      globals: {
        react: 'React',
-       '@tanstack/query-core': 'QueryCore',
      },
    }),
```

### 2. The `package.json` for packages are different

There are changes to the fields that bundlers use to determine how to resolve packages. We can see that the entrypoints (e.g. `module`, `main`, `exports`) are quite different between branches and that the `sideEffects` field is as well:

- Beta: https://github.com/TanStack/query/blob/4a4543c39ae500b537000c0cab807327d3ce5ff3/packages/react-query/package.json
- Main: https://github.com/TanStack/query/blob/b3201e991d8737d2a4ddbb99324f3282156a19ef/packages/react-query/package.json

Example: Partial diff for the `package.json` file from `"@tanstack/react-query"` 
```diff
"name": "@tanstack/react-query",
- "module": "build/lib/index.mjs",
+ "module": "build/esm/index.js",
- "main": "build/lib/index.js",
+ "main": "build/cjs/react-query/src/index.js",
  "browser": "build/umd/index.production.js",
- "types": "build/lib/index.d.ts",
+ "types": "build/types/packages/react-query/src/index.d.ts",
- "exports": {
-   ".": {
-     "types": "./build/lib/index.d.ts",
-     "import": "./build/lib/index.mjs",
-     "default": "./build/lib/index.js"
-   },
-   "./package.json": "./package.json"
- },
  "sideEffects": [
-   "./src/setBatchUpdatesFn.ts"
+   "lib/index.js",
+   "lib/index.mjs",
+   "lib/reactjs/index.js",
+   "lib/reactjs/index.mjs",
+   "lib/reactjs/setBatchUpdatesFn.js",
+   "lib/reactjs/setBatchUpdatesFn.mjs"
  ],
```

DamianOsipiuk:
@JohnDaly Yeah basically it's due to missing global declaration of react-dom. I will make the fix for it. Tested already that it fixes the issue.

PR #3935, opened 2022-07-28
feat: determine whehter suspend by user

liaoliao666:


packages/react-query/src/useBaseQuery.ts
```diff
@@ -102,10 +102,10 @@ export function useBaseQuery<
...
   // Handle suspense
   if (
-    defaultedOptions.suspense &&
-    result.isLoading &&
-    result.isFetching &&
-    !isRestoring
+    !isRestoring &&
+    (typeof defaultedOptions.suspense === 'function'
+      ? defaultedOptions.suspense(observer.getCurrentQuery())
+      : defaultedOptions.suspense && result.isLoading && result.isFetching)
```
TkDodo, line 108: can you please add a test for when `suspense` is a function?

there is also the thing that `useErrorBoundary` is set to `true` if we have suspense set:

https://github.com/TanStack/query/blob/454c58481f6f5d5c3e182e6e418d65fdb4a9ec1f/packages/query-core/src/queryClient.ts#L681-L683

so with that change, `useErrorBoundary` will be `true` when a function is passed to `suspense`. I'm not sure that this is right?

liaoliao666: It's fine, cuz the throw promise behavior of `useBaseQuery`  depend not only on suspense, and still depend on `isLoading` and `isFetching`. If u wanna put the suspend function on suspense, it's unavoidable.

and I think,  If people set `suspense` to function, It shoud be a metaphor that people will wanna this behavior like set it to `true`.

kamilkazmierczakMtab:
Unfortunately with `() => true` as `suspense` it goes into infinite fetch loop
Reproduction link -> https://codesandbox.io/s/heuristic-neumann-ohjy5e

liaoliao666:
@kamilkazmierczakMtab I think it is no problem. cuz it is different between `() => true` with `true` for `suspense`. if set `suspense` as `function` that it means that  this `function` should have a boundary conditions to return `true`

kamilkazmierczakMtab:
I see, I thought along with @TkDodo that it can allow to force Suspense to be triggered on non-initial fetches e.g. by calling `invalidateQueries` like in attempt to do it in my previous example in codesandbox

liaoliao666:
@kamilkazmierczakMtab But this is what i wanted. I wanna force `Suspense` whenever i want. I'm still do not understand what u worry about, cuz it's completely control by developer.

TkDodo:
> I wanna force Suspense whenever i want.

I think you can get the same behaviour with `queryClient.resetQueries` actually. `reset` will clear the data from the cache, and then re-render your component. At that point, it will see that there is no data, go fetch it and suspend. I think this would be the better solution to the problem you're trying to solve, wouldn't it ?

TkDodo:
given that we have `resetQueries` that can likely do exactly what we want, I don't think I'm going to merge this. Also, there is some talk on twitter about a new suspense api coming, and it might be a hook, so calling suspense conditionally might make it more difficult to adapt.

liaoliao666:
okay😂

PR #3940, opened 2022-07-28
Fixed incorrect syntax in QueryCache docs

phoenixbeats01:
The docs were showing that to fetch data from the query cache, you pass a string into the `find` and `findAll` method. That is incorrect as in react query v4, keys are arrays only.

### DIFF
--- a/docs/reference/QueryCache.md
+++ b/docs/reference/QueryCache.md
@@ -19,7 +19,7 @@ const queryCache = new QueryCache({
   }
 })
 
-const query = queryCache.find('posts')
+const query = queryCache.find(['posts'])
 ```
 
 Its available methods are:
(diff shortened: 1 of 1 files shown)


docs/reference/QueryCache.md
```diff
@@ -44,14 +44,15 @@ The `onError` and `onSuccess` callbacks on the QueryCache can be used to handle
 - `defaultOptions` can be overridden by each Query - the global callbacks will **always** be called.
 - `defaultOptions` callbacks will be called once for each Observer, while the global callbacks will only be called once per Query.
 
-## `queryCache.find`
+## `
+`
```
TkDodo, line 49: thank you, but this change looks wrong?

phoenixbeats01: Good catch. I accidentally deleted some text while I was writing the commit message. fixed

docs/reference/QueryCache.md
```diff
@@ -71,7 +71,7 @@ const query = queryCache.find(queryKey)
 > Note: This is not typically needed for most applications, but can come in handy when needing more information about a query in rare scenarios
 
 ```tsx
-const queries = queryCache.findAll(queryKey)
+const queries = queryCache.findAll([queryKey])
```
TkDodo, line 74: i would leave these two as-is. the options then state that queryKey needs to be of type QueryKey:

```
`queryKey?: QueryKey`: [Query Keys](../guides/query-keys)
```

and `QuerKey` is already an array!

phoenixbeats01: That's a good point, missed that, sorry!

PR #3951, opened 2022-07-30
docs(examples): remove redundant examples

zorzysty:
Removing:
- "custom hooks", because "basic" example uses custom hooks already
- "window focus refetching", because this feature is already turned on by default
- "optimistic updates" because "optimistic updates typescript" already covers this functionality


TkDodo:
I think we also need to update the docs that the links in the sidebar are not created ?

PR #3977, opened 2022-08-03
Add type:module property to package.json

ArtyCodingart:
If the "--experimental-vm-modules" property is set, modules are processed with an error.

image

Adding the property type: "module" solves this problem.


DamianOsipiuk:
I do not think that adding `type: module` is a good idea.
I think we should rather switch `exports` statements so `default` points to `cjs` and `import` points to `esm`

PR #3988, opened 2022-08-05
feat(react-query): isLoading of type DefinedUseQueryResult is always false

ronny1020:
Since there is always cache data, the `isLoading` of type `DefinedUseQueryResult`  should be false. Besides, the `Omit` is not required.

### DIFF
--- a/packages/query-core/src/types.ts
+++ b/packages/query-core/src/types.ts
@@ -425,11 +425,14 @@ export interface QueryObserverSuccessResult<TData = unknown, TError = unknown>
   status: 'success'
 }
 
-export type QueryObserverResult<TData = unknown, TError = unknown> =
-  | QueryObserverLoadingErrorResult<TData, TError>
-  | QueryObserverLoadingResult<TData, TError>
+export type DefinedQueryObserverResult<TData = unknown, TError = unknown> =
   | QueryObserverRefetchErrorResult<TData, TError>
   | QueryObserverSuccessResult<TData, TError>
+
+export type QueryObserverResult<TData = unknown, TError = unknown> =
+  | DefinedQueryObserverResult<TData, TError>
+  | QueryObserverLoadingErrorResult<TData, TError>
+  | QueryObserverLoadingResult<TData, TError>
 
 export interface InfiniteQueryObserverBaseResult<
   TData = unknown,
--- a/packages/react-query/src/types.ts
+++ b/packages/react-query/src/types.ts
@@ -65,10 +65,10 @@ export type UseQueryResult<
   TError = unknown,
 > = UseBaseQueryResult<TData, TError>
 
-export type DefinedUseQueryResult<TData = unknown, TError = unknown> = Omit<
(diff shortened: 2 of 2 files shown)


TkDodo:
Why is the omit not required? It should be so that the type of data will be properly narrowed...?

ronny1020:
I'm not sure what case will it not be appropriately narrowed.

```ts
type Prototype = { a: number | null };

type NewType = Prototype & { a: number };

type NewNeverType = Prototype & { a: string };

// OK
const a: NewType = { a: 0 };

// Error
const b: NewType = { a: null };

// Error, the type of property a is never
const c: NewNeverType = { a: '3' };
```

I think it will be fine. Besides, it can avoid using the type which is not in the prototype.

Moreover, if I used `Omit`, there would be `overload` error in the `useQuery`, or  I have to write something like `UseQueryResult<TData, TError> | DefinedUseQueryResult<TData, TError>`.

TkDodo:
> I'm not sure what case will it not be appropriately narrowed.

Interesting! I'm pretty sure there was a time when you needed to remove a property from an object type before you could refine it, or the intersection wouldn't work properly. But I've tested this as far back as TS v3.3 and it all works fine 🤷

TkDodo:
I'm still not sold on setting `isLoading` to `false`. If we do that, wouldn't we also need to remove `'loading'` from the `'status'` field? Why is it important?

ronny1020:
> > I'm not sure what case will it not be appropriately narrowed.
> 
> Interesting! I'm pretty sure there was a time when you needed to remove a property from an object type before you could refine it, or the intersection wouldn't work properly. But I've tested this as far back as TS v3.3 and it all works fine 🤷

I think that could happen if given a type that does not extend from the original one. In that case `Omit` is required.

ronny1020:
> I'm still not sold on setting `isLoading` to `false`. If we do that, wouldn't we also need to remove `'loading'` from the `'status'` field? Why is it important?

In my opinion, I agree to remove the `loading` from the `status`. Since the type wouldn't happen, the type should not include it.

To be honest, the reason why I create this PR is that in my project I had used the previous version of `react-query`, and the data from the `useQuery` could be undefined so I write something like `const value = data ?? defaultValue` instead of using `initialData`. After migrating to the current version, I find out there is a fantastic type,` DefinedUseQueryResult` so I start to use this and the `initialData`, However, I forget to change the `isLoading` to `isFetching` which cause some bugs.

It's not a big deal, I just think that still could happen to some other guys. With a strict type, maybe that could be avoided. Moreover, I can't see a reason to allow a type that will not exist.

TkDodo:
alright, please remove the `loading` status from the union, too

ronny1020:
Hi, I think I found a better way to solve this issue. The `Omit` and `loading` things wouldn't be a problem anymore. Other properties should also work properly such as `isLoadingError`. Moreover, if there is some change related to the status, the type will still work properly.

TkDodo:
oh yeah I like that solution 👍

Issue #2741, opened 2021-09-29
disabled label not updating in the devtools


TkDodo:
**Describe the bug**

As of recently, the `devtools` show a `disabled` label when all observers of a query are disabled. However, this label doesn't update dynamically.

**To Reproduce**
1. sandbox: https://codesandbox.io/s/serverless-currying-js20q?file=/src/index.js
2. Make sure devtools are open
3. Toggle enabled via the button

devtools still look like this (no disabled label):

![Screenshot 2021-09-29 at 17 23 47](https://user-images.githubusercontent.com/1021430/135299706-efab4595-948a-4ec0-b431-673f91ad7cc9.png)

4. close the devtools, and open them again

--> they now look good

![Screenshot 2021-09-29 at 17 24 31](https://user-images.githubusercontent.com/1021430/135299781-c0c41402-2926-45f3-bf35-62e9ef06186b.png)

**Expected behavior**
the correct enabled state should always be reflected in the devtools

TkDodo:
alright, this is not the easy fix that I thought it would be. devtools subscribe to the queryCache, but there is no `QueryCacheNotifyEvent` for when the options of an observer change. disabling a query changes nothing in the cache itself 🤔

TkDodo:
one quick workaround would be to add a `reload devtools` button somewhere to ensure that the devtools are up-to-date

armandabric:
I just get caught with this one. I spend some time to ensure my query was disabled event if it's not shown by the devtools. Did you have an idea to get this fixed? Maybe only in v4 if this need to much change

TkDodo:
@armandabric the devtools subscribe to the `queryCache`, and the cache emits `events` that we can listen to. Events are:

- `queryAdded`
- `queryRemoved`
- `queryUpdated`
- `observerAdded`
- `observerRemoved`

there is no event for `observerUpdated`, but that is the only thing that changes when the `enabled` option changes. So I don't see a good way to make this work at the moment.

DamianOsipiuk:
@TkDodo Adding `observerOptionsUpdated` event to queryCache and sending it in `setOptions` of `queryObserver` after `updateQuery` would do?
https://github.com/TanStack/query/blob/main/packages/query-core/src/queryObserver.ts#L171

TkDodo:
Yes this sounds about right 👍

PR #3989, opened 2022-08-05
feat: notify about observer options change

DamianOsipiuk:
https://github.com/TanStack/query/issues/2741

### DIFF
--- a/packages/react-query/src/__tests__/useQuery.test.tsx
+++ b/packages/react-query/src/__tests__/useQuery.test.tsx
@@ -2149,14 +2149,16 @@ describe('useQuery', () => {
     await sleep(20)
     unsubscribe()
 
-    // 1. Subscribe observer
-    // 2. Query loading
-    // 3. Observer loading
-    // 4. Query success
-    // 5. Observer success
-    // 6. Query stale
-    // 7. Unsubscribe observer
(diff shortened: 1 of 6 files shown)


packages/react-query/src/__tests__/useQuery.test.tsx
```diff
@@ -2149,14 +2149,16 @@ describe('useQuery', () => {
...
-    expect(fn).toHaveBeenCalledTimes(7)
+    // 1. Query added -> loading
+    // 2. Observer result updated -> loading
+    // 3. Observer added
+    // 4. Query updated -> success
+    // 5. Observer result updated -> success
+    // 6. Query updated -> stale
+    // 7. Observer options updated
+    // 8. Observer result updated -> stale
+    // 9. Observer options updated
```
DamianOsipiuk, line 2160: @TkDodo Not sure why observer options is called twice...

TkDodo: not sure, but it shouldn't matter much I guess..

I can confirm that the devtools now update when an observer disables, which is awesome. thank you!

https://codesandbox.io/s/tanstack-query-example-react-basic-forked-gfwimm?file=/src/index.js

TkDodo (approved):

PR #4011, opened 2022-08-09
fix(react-query-devtools): add required `react` versions

### LINKED ISSUE #4004: ReactQueryDevtools v4: Unable to find required version for "react" in description file
### Describe the bug

After upgrading to v4 (and going through the migration guide) we get this warning (from webpack-dev-server):

"Compiled with problems:

WARNING in shared module react

No required version specified and unable to automatically determine one. Unable to find required version for "react" in description file (/path/to/project/node_modules/@tanstack/react-query-devtools/package.json). It need to be in dependencies, devDependencies or peerDependencies."

ReactQuery sometimes stops working too when this happens. If we remove the devtools, everything works as usual.

**Manually adding react as a peer dependency in `node_modules/@tanstack/react-query-devtools/package.json` fixes it!**

I tried reproducing this in codesandbox with just a clean create-react-app, but that setup doesn't seem to have the same issues and I haven't had the time to setup a full single-spa/module-federation project there to test with.

Is it reasonable to ask you to add react as a peer dependency within this package.json file? Or maybe there is some other way to tell webpack about what react version this package need? Or do you think this is an issue we need to figure out within our project?

[WIP]

### Steps to reproduce

Still working on this, but I suspect it can be replicated by setting up a project that uses webpack module federation.

### Expected behavior

Should run without error.

_No response_

### Screenshots or Videos

_No response_

### Platform

- OS: MacOS (ARM64)
- Browser: Chrome
- Version: 103.0.5060.134 (Official Build) (arm64)

### react-query version

4.0.10

### TypeScript version

^4.7.2

### Additional context

We are running a monorepo with single-spa (which uses webpack module federation I think). I suspect this may have something to do with it, but I'am not sure.

Small tangent: Can not overstate how great the transition from Redux/Saga to ReactQuery has been! Amazing library you people have made, we really appreciate it! :)

aboqasem:
Closes #4004


packages/react-query-devtools/package.json
```diff
@@ -36,5 +36,14 @@
...
     "use-sync-external-store": "^1.2.0"
+  },
+  "peerDependencies": {
+    "react": "^16.8.0 || ^17.0.0 || ^18.0.0",
+    "react-dom": "^16.8.0 || ^17.0.0 || ^18.0.0"
+  },
+  "peerDependenciesMeta": {
+    "react-dom": {
+      "optional": true
+    }
```
TkDodo, line 47: for the devtools, I don't think `react-dom` is optional, as it's a browser ui :)

TkDodo:
a similar PR was just merged, please address the conflicts and the above comment, thanks.

packages/react-query-devtools/package.json
aboqasem: Noted!

TkDodo (approved):

PR #4013, opened 2022-08-09
fix: subpackage dependencies

DamianOsipiuk:
This should allow for automatic deps install on sub packages. Currently when you try to install one of those, you might get an error about missing imports.

@TkDodo This should go to beta for pre-release

### DIFF
--- a/packages/query-sync-storage-persister/package.json
+++ b/packages/query-sync-storage-persister/package.json
@@ -30,5 +30,8 @@
   "scripts": {
     "clean": "rm -rf ./build",
     "test:eslint": "../../node_modules/.bin/eslint --ext .ts,.tsx ./src"
+  },
+  "dependencies": {
+    "@tanstack/react-query-persist-client": "^4.0.10"
   }
 }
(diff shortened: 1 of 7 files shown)


packages/query-sync-storage-persister/package.json
```diff
@@ -30,5 +30,8 @@
   "scripts": {
     "clean": "rm -rf ./build",
     "test:eslint": "../../node_modules/.bin/eslint --ext .ts,.tsx ./src"
+  },
+  "dependencies": {
+    "@tanstack/react-query-persist-client": "^4.0.10"
```
TkDodo, line 35: i think it's wrong that `query-*` packages depend on `react-query-*` packages! It seems that we only import some `types`, so maybe it should just be the other way around and we should in fact move the types ?

DamianOsipiuk: That is a good point. However the types that are imported are common for both persisters. I would say that we should extract common logic from `react-query-persist-client` to `query-persist-client` so framework-specific adapters would be possible.
But that could be next step.

PR #4030, opened 2022-08-14
feat: enabled option that should support to function

liaoliao666:


knyzorg:
@TkDodo @liaoliao666  Does this PR need any more work? I'm really struggling with the lack of this feature.

TkDodo:
It has conflicts in package lock and it doesn't have any tests for the new functionality. Also, I'd need to check if all cases have been taken into account.

What is your use case that requires this?

liaoliao666:
@knyzorg Currently I develop a toolkit for ReactQuery that contains this feature, maybe u would like to try to use it. here is the link https://github.com/liaoliao666/react-query-kit
@TkDodo In some case, I do not wanna refetch data when already fetched

TkDodo:
@liaoliao666 that's what `staleTime` is for ?

liaoliao666:
@TkDodo It seems so. Never thought this option can be used like this😂. I have no more question for this issue

TkDodo:
> I have some queries which monitor jobs that where the status goes from "In Queue", to "In Progress", and then to either "Complete" or "Failed". Once a query becomes "Complete" or "Failed", it is in a terminal state (It will never change). I don't want to be refreshing queries in a terminal state.

I would personally use a `refetchInterval` that's a function so that it stops polling after the terminal state is reached, and then set an infinite staleTime. But you are right, query invalidation would still "target" it.

What you could do is implement your queryFn in a way that it checks the current state of the query before deciding to make a call:

```
const queryClient = useQueryClient
const query = useQuery(key, () => {
  const existingData = queryClient.getQueryData(key)
  return existingData?.state === 'Complete'
    ? existingData
    : fetchTheData()
})

**Comment by knyzorg** [comment]
@TkDodo I don't like that solution because it requires coupling the query client to the data fetching function.

Is there a reason for not allowing functions to handle whether a query should be enabled or not? 

It's not as if it's introducing a breaking change.

**Comment by TkDodo** [comment]
> Is there a reason for not allowing functions to handle whether a query should be enabled or not?

yes and no :) `enabled` is a per-observer option, so technically, a query can't be "disabled" - only observers can. If we're only passing the `query` to it, we are missing the result of the `select` option, which can also be different per observer. If you want to calculate being enabled depending on "your data" (as an observer), this might be a crucial, missing information 

For comparisons (to see if enabled has changed between renders), the function must also run with the previousQuery, but this might yield different results if the function itself closures over data from the component. It might be expected that the result from the previous invocation is cached.

So yeah my main concern is "it's not that easy" and it doesn't add a ton of value and there are workarounds ...

**Comment by knyzorg** [comment]
Almost every other property can be a function. Is that because none of them relate to observers, and are solely related to queries?
I don't really understand the difference between a query and an observable, so these are just probing question, that might not make sense technically:

 - what would be the issue with managing if a query is enabled in the query instead of the observer? This could allow us to run the query through the select, right?

 - could the function receive the observable as the argument to the function instead of the query? Could that be a way of approaching it?

 - I cant imagine select being strictly necessary to finding information. Tbh, select feels like an antipattern to me as a whole.

As for comparing the previous and old results, I agree with you. In my PR I approached it with the logic of that the function should have the final say on it, and theres no reason to call the function with the previous query. Caching would significantly increase the complexity of a currently simple implementation.

I did end up working around it by generating a wrapper around the data fetching function, similar to your previous comment. While I do appreciate that there are complexities here, I don't think "there are workarounds" is a fair response. 

React Query has a lot of features that couldve been dismissed with a "there are workarounds", such as select. And yet, it is still there, because it is convenient.

The only difference here, is that my need for it is a lot more niche.

**Comment by TkDodo** [comment]
> Is that because none of them relate to observers, and are solely related to queries?

Those that are related to observers either receive the transformed data (TData) additionally, or they don't care about data that much because they are for errors.

> I cant imagine select being strictly necessary to finding information. Tbh, select feels like an antipattern to me as a whole.

select is the recommended way for fine grained subscriptions and data transformations. You can read about it here: https://tkdodo.eu/blog/react-query-data-transformations#3-using-the-select-option

> React Query has a lot of features that couldve been dismissed with a "there are workarounds", such as select. And yet, it is still there, because it is convenient.

I am hoping that there is a reason for every feature apart from convenience. For example, because you mentioned select, there is actually no other way to subscribe to parts of the query cache. It's the only place where you an do data transformation and react-query can still bail out of re-rendering your component as a whole, which is huge.

> As for comparing the previous and old results, I agree with you. In my PR I approached it with the logic of that the function should have the final say on it, and theres no reason to call the function with the previous query.

In order to know if a fetch should be triggered when a query goes from disabled to enabled, we need to know if the query was previously disabled. This currently happens here:

https://github.com/TanStack/query/blob/7e51776bff00b2f6f8496a7a64b3a048238a4ee7/packages/query-core/src/queryObserver.ts#L745

So yeah, we definitely need to know the previous value of `enabled`. Calculating it every time seems like the wrong thing to do because of closures, and caching is too much complexity for too little gain. Sorry.

**Comment by liaoliao666** [comment]
@TkDodo I think we can pre-execute the `enabled` function in he `useBaseQuery.ts`. This way we do not have to worry about it affecting the logic of queryObserver. like below

```ts
// useBaseQuery.ts
if (typeof defaultedOptions.enabled === 'function') {
    defaultedOptions.enabled = defaultedOptions.enabled(queryClient.getQueryData(defaultedOptions.queryKey))
 }
```

And I think this feat is very convenience in some case by add a few of code

TkDodo:
This would make enabled work differently on `useQuery` than when creating a new QueryObserver manually, and we'd likely have to replicate this functionality for other framework adapters, too.

If it's just one line of code, you can also do this in userland ?

```
const enabled = computeEnabled(queryClient.getQueryData(key))
const query = useQuery(key, fn, { enabled })
```

PR #4044, opened 2022-08-16
fix: remove browser entry, fix umd size

DamianOsipiuk:
@TkDodo
This should fix:
- some bundlers picking up `browser` entry in package.json (`umd` build instead of `cjs` or `esm`)
- unusual size of `react-query`, due to missing rollup global definition for react-dom

This won't fix:
- bundlers that support `exports`, but does not support `development/production` conditionals (https://webpack.js.org/guides/package-exports/#optimizations)
  This might be the case as those conditionals were mostly proposed by `webpack`.
  We could potentially overcome this by linking `exports/./default/default` not to `noop.js`, but to intermediate `js` file that conditionally require devtools based on `NODE_ENV`, like in the v3.
  This however i would like to test separately.

Previous beta release also have a side effect that `react-query` UMD build does not contain `query-core` UMD build and therefore is smaller. Is that something we should fix to be bundled together?

--

I would like to get this as a beta pre-release to test it out.

### DIFF
--- a/packages/query-async-storage-persister/package.json
+++ b/packages/query-async-storage-persister/package.json
@@ -10,10 +10,9 @@
     "type": "github",
     "url": "https://github.com/sponsors/tannerlinsley"
   },
+  "types": "build/lib/index.d.ts",
   "module": "build/lib/index.mjs",
   "main": "build/lib/index.js",
-  "browser": "build/umd/index.production.js",
-  "types": "build/lib/index.d.ts",
   "exports": {
     ".": {
       "types": "./build/lib/index.d.ts",
--- a/packages/query-broadcast-client-experimental/package.json
+++ b/packages/query-broadcast-client-experimental/package.json
@@ -10,10 +10,9 @@
     "type": "github",
     "url": "https://github.com/sponsors/tannerlinsley"
   },
+  "types": "build/lib/index.d.ts",
   "module": "build/lib/index.mjs",
   "main": "build/lib/index.js",
-  "browser": "build/umd/index.production.js",
-  "types": "build/lib/index.d.ts",
   "exports": {
     ".": {
       "types": "./build/lib/index.d.ts",
--- a/packages/query-core/package.json
+++ b/packages/query-core/package.json
@@ -10,10 +10,9 @@
     "type": "github",
     "url": "https://github.com/sponsors/tannerlinsley"
   },
+  "types": "build/lib/index.d.ts",
   "module": "build/lib/index.mjs",
   "main": "build/lib/index.js",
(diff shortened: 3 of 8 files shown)


TkDodo (approved):

TkDodo:
I merged `main` back into `beta` - there were lots of conflicts. Can you update your PR with `beta` too please?

TkDodo:
then I'll make the next beta release. Current release is `4.2.0` - does that make the next beta be `4.3.0-beta.0` ?

DamianOsipiuk:
> then I'll make the next beta release. Current release is `4.2.0` - does that make the next beta be `4.3.0-beta.0` ?

I would say 4.2.2-beta.0, but you can also do 4.3.0-beta.0. It should not matter too much.

TkDodo:
@DamianOsipiuk I wanted to check the bundle size, but it seems the config is no longer correct:

https://github.com/TanStack/query/blob/8025dbf46ce0dff3b9ab98c4dc2ac275ab60550e/package.json#L120-L126

Not sure where the umd builds are now ? I think we just need to point towards `packages/**/build/lib/index.js` ?

DamianOsipiuk:
@TkDodo UMD build files should be in the same place and the config should still work.

image

TkDodo:
@DamianOsipiuk yeah I've seen the results, but they look weird, like, why is `packages/react-query/build/umd/index.production.js` only 2.22 kb?

DamianOsipiuk:
> Previous beta release also have a side effect that react-query UMD build does not contain query-core UMD build and therefore is smaller. Is that something we should fix to be bundled together?

Basically in current rollup setup, if we mark `query-core` as external to strip it out from `cjs` and `esm` builds, it also strips it from `umd` build. Previously it was not the case.

Therefore if we want to maintain previous way of bundling `react-query` and `query-core` together but only for `UMD` build, rollup config needs to be restructured.

TkDodo:
ah yes. so [in the docs](https://tanstack.com/query/v4/docs/installation#cdn) we say that react-query can be installed via:

```
<script src="https://unpkg.com/@tanstack/react-query@4/build/umd/index.production.js"></script>
```

this should still work, so we can't really break that

DamianOsipiuk:
Hmmm, i think that right now you would have to 
```
<script src="https://unpkg.com/@tanstack/query-core@4/build/umd/index.production.js"></script>
<script src="https://unpkg.com/@tanstack/react-query@4/build/umd/index.production.js"></script>
```

So it is kind of a breaking change unfortunately 😞

DamianOsipiuk:
@TkDodo Should be good now

PR #4053, opened 2022-08-18
fix release script

TkDodo:


.github/workflows/ci.yml
```diff
@@ -25,7 +25,6 @@ jobs:
           cache: 'npm'
       - run: |
           npm i
-          npm run test:ci
```
TkDodo, line 28: not needed because the publishci script also runs test

scripts/publish.ts
```diff
@@ -167,7 +167,7 @@ async function run() {
     ? packages
     : changedFiles.reduce((changedPackages, file) => {
         const pkg = packages.find((p) =>
-          file.startsWith(path.join('packages', p.packageDir, p.srcDir)),
+          file.startsWith(path.join('packages', p.packageDir)),
```
TkDodo, line 170: makes sure that we publish new versions whenever there are changes anywhere in the package directory, not just the `src` directory!

scripts/publish.ts
```diff
@@ -177,16 +177,28 @@ async function run() {
...
+    const allDependencies = Object.keys(
+      Object.assign(
+        {},
+        packageJson.dependencies ?? {},
+        packageJson.peerDependencies ?? {},
+      ),
+    )
+
     if (
-      pkg.dependencies?.find((dep) =>
```
TkDodo, line 182: `pkg.dependencies` does not exist - the type were wrong
so we need to find them in packgage.json

scripts/publish.ts
```diff
@@ -419,59 +427,55 @@ async function run() {
       path.resolve(rootDir, 'packages', pkg.packageDir, 'package.json'),
       async (config) => {
         await Promise.all(
-          (pkg.dependencies ?? []).map(async (dep) => {
+          Object.keys(config.dependencies ?? {}).map(async (dep) => {
```
TkDodo, line 430: same here - pkg.dependencies does not exist

scripts/publish.ts
```diff
@@ -419,59 +427,55 @@ async function run() {
...
             const depPackage = packages.find((d) => d.name === dep)
 
-            if (!depPackage) {
-              throw new Error(`Could not find package ${dep}`)
```
TkDodo, line 426: throwing here is wrong as it fails the promise and `Promise.all` stops after the first rejection. We can just ignore dependencies that are _not_ dependencies we care about

DamianOsipiuk (approved):
Nice 🥇

Issue #2994, opened 2021-11-20
Move dev tools option


claytron5000:
Sometimes the ergonomics of a button- stuck tool is hard to use. Could there be a button, like in native dev tools to pin the query tool to the sides, or top/ bottom

TkDodo:
Can you be a bit more specific? What is a button-stuck tool please?

i think there are already options to position the devtools differently if that’s what you mean..

claytron5000:
No problem, I probably shouldn't write features requests while walking the dog. 

Currently we can set the position of the flower tool by passing the position parameter:
```
<ReactQueryDevtools position={"top-right"} />
```
That puts the flower trigger in the top right corner. However, the drawer always opens at the bottom.

A user could pass styles and classes in through `panelProps`. And position the r-query devtools where-ever they please.


What I'm thinking is a toggler within the devtools that allows the user to switch the drawer position on the fly. The example I'm thinking of is the layout switcher in Chrome:
![image](https://user-images.githubusercontent.com/5767672/142743719-afaea8bd-6a3d-43d4-9b4e-aa0099bc8a7c.png)

This would allow  a user to reposition the devtool drawer on the fly as they might switch page layouts, or other window sizes.

TkDodo:
❤️ the idea, would you like to add it?

claytron5000:
Yes, I'll work on it.
═════════
Clayton Salem
413-768-8522
═════════

salamaashoush:
@TkDodo  @claytron5000  I could work on this if there is one currently working on it

TkDodo:
@salamaashoush go ahead, I've assigned you to the issue.

nirtamir2:
Hi,
I created https://github.com/nirtamir2/react-query-devtools-extension to solve this problem. It works well in development but I have a little problem with its event when building it for production. You can use it and integrate it back to react query

salamaashoush:
@nirtamir2  this is a pretty cool browser extension, but I am just adding the ability to reposition the current devtools so you don't need to install a browser extension

salamaashoush:
@TkDodo  if you have time could you please have a look at this PR https://github.com/TanStack/query/pull/3678

yairEO:
I would love to be able to hold with the mouse and drag it somewhere else, in case it covers something I want to interact with on the page (or see). It doesn't need to remember the new dragged position after refresh, but simply solve an anecdotal situation

PR #4060, opened 2022-08-20
feat: Add move devtools select menu

salamaashoush:
closes https://github.com/TanStack/query/issues/2994

Added the option to move dev tools to one of these sides `left`,  `right`, `top`, and `bottom` and refactored the code to accommodate the new capability.

image

image

image


salamaashoush:
@TkDodo here you go a new PR, please try to review and give feedback

TkDodo:
please check the prettier formatting:

```
> prettier
> prettier "packages/*/{src/**,examples/**/src/**}.{md,js,jsx,ts,tsx,json}" "--check"

Checking formatting...
[warn] packages/react-query-devtools/src/__tests__/devtools.test.tsx
Error: Process completed with exit code 1.
```

docs/devtools.md
```diff
@@ -87,3 +90,7 @@ Use these options to style the dev tools.
   - The standard React style object used to style a component with inline styles
 - `className: string`
   - The standard React className property used to style a component with classes
+- `showCloseButton?: boolean`
```
TkDodo, line 93: I tried this prop in codesandbox, but it seems that it doesn't do anything?

https://codesandbox.io/s/tanstack-query-example-react-basic-forked-x3s37v?file=/src/index.jsx

In that sandbox, it is set to `false`, but the closeButton still shows up.

packages/react-query-devtools/src/devtools.tsx
```diff
@@ -681,6 +686,48 @@ export const ReactQueryDevtoolsPanel = React.forwardRef<
...
+                onClick={(e) => {
+                  setIsOpen(false)
+                  onCloseClick?.(e)
+                }}
+              >
+                Close
+              </Button>
+            ) : null}
+            {position && onPositionChange ? (
+              <Select
```
TkDodo, line 717: I think we'd also need a way to hide this select. One reason to hide the closeButton was because it overlaps with queries if you have many of them, and the Select in this position probably has the same drawback.

TkDodo (commented):
love this feature, thank you. One thing about responsiveness:
if the devtools are positioned on the bottom, they are responsive, e.g. the sorting select is hidden if there is not enough space:

large screen:

Screenshot 2022-08-21 at 10 16 37

smaller screen:
Screenshot 2022-08-21 at 10 16 50

This avoids horizontal scrollbars, however, it doesn't seem to work if devtools are positioned on the right / left? Also, the position select itself disappears on smaller resolutions. I think this is because `Select` itself has a media query 😅: 

https://github.com/TanStack/query/blob/357ec041a6fcc4a550f3df02c12ecc7bcdefbc05/packages/react-query-devtools/src/styledComponents.ts#L103-L107

packages/react-query-devtools/src/devtools.tsx
salamaashoush: should I move the position select on the top or just add a flag to hide it

TkDodo: is there enough place on the top? But yeah, I would like to avoid more flags ...

salamaashoush: I will try to squeeze it in there

docs/devtools.md
salamaashoush: I will fix that

salamaashoush:
@TkDodo  about the responsiveness in the left/right positions, we use media queries and it works on the screen width, not container width so that's why it is not working when you resize the panel, it would require changing the approach to use `resize` event or `ResizeObserver` since `Container queries` are not widely implemented yet, WDYT, should I go in that direction and fix it or it is a minor thing we can fix later?

TkDodo:
@salamaashoush you are right. I think listening to a resize observer might be the best thing until container queries become a thing, but we can do this in a separate issue / follow up 👍

PR #3854, opened 2022-07-19
fix: Adding screenreader text to the DevTools open icon

nicholasgriffintn:
While testing our local site with our [accessibility testing tool](https://github.com/bbc/bbc-a11y), we recieved the following error:

```
✗ Main Dashboard
  * Forms: Labelling form controls: Fields must have labels or titles
    - Button has no text: //div[@id='ReactQueryDevtoolsPanel']/div[2]/div[1]/button
    - Button has no text: //main[@id='main-content']/aside/button
```

The issue is that not all screenreaders support aria values, and so this test is checking that buttons at least have a text value within them, which can be visibility hidden if so required.

To resolve this, this PR adds visually hidden text to the button.

TODO:

- [ ] I'm still working out if this is the best way of doing it (I haven't been through all of the code yet)
- [ ] I need to test this locally to ensure it meets the requirements.

PR #4063, opened 2022-08-21
fix: Adding screenreader text to the DevTools open icon

nicholasgriffintn:
> Recreated from https://github.com/TanStack/query/pull/3854

While testing our local site with our [accessibility testing tool](https://github.com/bbc/bbc-a11y), we recieved the following error:

```
✗ Main Dashboard
  * Forms: Labelling form controls: Fields must have labels or titles
    - Button has no text: //div[@id='ReactQueryDevtoolsPanel']/div[2]/div[1]/button
    - Button has no text: //main[@id='main-content']/aside/button
```

The issue is that not all screenreaders support aria values, and so this test is checking that buttons at least have a text value within them, which can be visibility hidden if so required.

To resolve this, this PR adds visually hidden text to the button.

### DIFF
--- a/packages/react-query-devtools/src/screenreader.tsx
+++ b/packages/react-query-devtools/src/screenreader.tsx
@@ -9,7 +9,6 @@ export default function ScreenReader({ text }: { text: string }) {
         height: '0.1px',
         overflow: 'hidden',
       }}
-      className="screenreader"
     >
       {text}
     </span>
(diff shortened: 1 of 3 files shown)


packages/react-query-devtools/src/screenreader.tsx
```diff
@@ -0,0 +1,17 @@
...
+export default function ScreenReader({ text }: { text: string }) {
+  return (
+    <span
+      style={{
+        position: 'absolute',
+        width: '0.1px',
+        height: '0.1px',
+        overflow: 'hidden',
+      }}
+      className="screenreader"
```
TkDodo, line 12: I don't think this className exists?

```suggestion

```

nicholasgriffintn: Yes it makes sense to remove that.

PR #4064, opened 2022-08-21
Add new guides for React Native createAsyncStoragePersister

amerllica:
Add new documentation guides about using `PersistQueryClientProvider` instead of using `persistQueryClient`.


amerllica:
Dear @TkDodo, I sent this PR, but I didn't have an issue with this PR, if you think using `fixes/resolves` works are mandatory please give me a shout here, then I will create an issue and will connect it to this PR. thanks.

PR #4074, opened 2022-08-23
fix: umd build size, force prod devtools

DamianOsipiuk:


packages/react-query-devtools/package.json
```diff
@@ -14,37 +14,32 @@
   "main": "build/lib/index.js",
   "exports": {
     ".": {
-      "development": {
```
DamianOsipiuk, line 17: generated files `index.js`/`index.mjs` still contain `NODE_ENV` conditions which will be used by the bundler to treeshake it from production

rollup.config.ts
```diff
@@ -94,19 +99,31 @@ export default function rollup(options: RollupOptions): RollupOptions[] {
...
+        'use-sync-external-store/shim/index.js': 'UseSyncExternalStore',
       },
+      bundleUMDGlobals: [
+        '@tanstack/match-sorter-utils',
+        'use-sync-external-store/shim/index.js',
+      ],
     }),
     ...buildConfigs({
-      name: 'react-query-devtools-noop',
+      name: 'react-query-devtools-prod',
```
DamianOsipiuk, line 113: This config was changed to contain non-treeshakable devtools build accessible via `@tanstack/react-query-devtools/build/lib/index.prod.js`

rollup.config.ts
```diff
@@ -94,19 +99,31 @@ export default function rollup(options: RollupOptions): RollupOptions[] {
...
+      outputFile: 'index.prod',
+      entryFile: 'src/index.ts',
       globals: {
         react: 'React',
+        'react-dom': 'ReactDOM',
         '@tanstack/react-query': 'ReactQuery',
+        '@tanstack/match-sorter-utils': 'MatchSorterUtils',
+        'use-sync-external-store/shim/index.js': 'UseSyncExternalStore',
       },
+      forceDevEnv: true,
```
DamianOsipiuk, line 125: This forces the build to thinks it's bundling for dev env

rollup.config.ts
```diff
@@ -94,19 +99,31 @@ export default function rollup(options: RollupOptions): RollupOptions[] {
...
+      skipUmdBuild: true,
```
DamianOsipiuk, line 126: we do not want to generate extra set for `umd`, since those are already generated

DamianOsipiuk:
@TkDodo So this seems to be working for `react-scripts 4/5` and `vite`.
Also devtools do not appear on the `prod` builds.

The only issue i can see is that `react-scripts` do not properly treeshake the devtools (some code is left despite exports returning `null`). Possibly due to the fact that they fall back to `cjs` build.
This behavior is also present in previous beta releases.

TkDodo:
@DamianOsipiuk okay, let's make a new beta and I'll try it out!

PR #4085, opened 2022-08-25
refactor(react-query-devtools): use `superjson` to display values

### LINKED ISSUE #1552: Error displays as an empty object
When my query function throws an error, the devtools displays it as an empty object.
I would expect it to display the `message` and maybe `stack` string properties that the object contains.

If I throw a custom error with additional properties, such as `status`, it shows up just fine in the devtools. So it seems to only be the default error properties that are not being shown.

image

aboqasem:
Fixes #1552


package.json
```diff
@@ -112,6 +112,7 @@
     "rollup-plugin-visualizer": "^5.6.0",
     "solid-js": "^1.3.15",
     "stream-to-array": "^2.3.0",
+    "superjson": "^1.9.1",
```
TkDodo, line 115: This needs to be a `dependency` of the `react-query-devtools` package please.

TkDodo (changes requested):

TkDodo:
not sure that there should be so many changes in package-lock.json, also, builds are failing. I would suggest reverting the changes to packge-lock (restoring what we have in main), then running `npm install` again.

aboqasem:
I tried and it produces similar result. Which version of `npm` should I use? Currently I have to `npm i --legacy-peer-deps` on `8.18.0`.

TkDodo:
our `.nvmrc` uses node `v16.14.2`, so please use that

TkDodo (approved):

TkDodo:
sorry, I wanted to wait until we merged our beta branch with bundling changes. If you fix the merge conflicts, I'll merge this PR

TkDodo (approved):

TkDodo:
yeah, we fixed it immediately. sorry 🙈

PR #4092, opened 2022-08-28
ESM support

### LINKED ISSUE #3882: ESM support

TkDodo:
fixes #3882


TkDodo:
@sachinraja could you have a look a this PR please to see if we're missing something? All the reported issues and things we've identified seems to be resolved with this. It is also already available as beta release:

```
"@tanstack/react-query": "4.3.0-beta.4",
"@tanstack/react-query-devtools": "4.3.0-beta.7",
```

Thank you

packages/react-query-devtools/src/useSyncExternalStore.native.ts
```diff
@@ -0,0 +1,3 @@
+// Temporary workaround due to an issue with react-native uSES - https://github.com/TanStack/query/pull/3601
+// @ts-ignore
+export { useSyncExternalStore } from 'use-sync-external-store/shim/index.native.js'
```
sachinraja, line 3: This doesn't work since the code is bundled. Since rollup only sees `useSyncExternalStore.ts`, it just bundle it into `build/lib/index.mjs` and `build/lib/index.js`.

packages/react-query-devtools/package.json
```diff
@@ -10,36 +10,46 @@
...
-    "development": {
-      "require": "./build/cjs/packages/react-query-devtools/src/index.js",
-      "default": "./build/esm/index.js"
+    ".": {
+      "types": "./build/lib/index.d.ts",
+      "import": "./build/lib/index.mjs",
+      "default": "./build/lib/index.js"
+    },
+    "./production": {
+      "types": "./build/lib/index.d.ts",
```
sachinraja, line 22: These exports won't work with TypeScript versions < 4.7 or bundlers that don't support `exports` (older bundlers and React Native). Not sure what platforms/versions we're targeting but something to keep in mind.

If we do want to support them, we'll need to match the top level file structure of the package with the entry points in `exports`.

packages/react-query/src/useSyncExternalStore.native.ts
```diff
@@ -0,0 +1,3 @@
+// Temporary workaround due to an issue with react-native uSES - https://github.com/TanStack/query/pull/3601
+// @ts-ignore
+export { useSyncExternalStore } from 'use-sync-external-store/shim/index.native.js'
```
sachinraja, line 3: Same issue as what I mentioned before for `react-query-devtools`

sachinraja (commented):
Looks mostly good but had a few concerns

packages/react-query-devtools/package.json
TkDodo: yep, we are aware of that. I've documented that here: Do you think it's good enough?

https://github.com/TanStack/query/blob/beta/docs/devtools.md#modern-bundlers

sachinraja: Ah yeah that's good

packages/react-query-devtools/src/useSyncExternalStore.native.ts
TkDodo: How can we fix it?

DamianOsipiuk: I guess we could create a proxy package with proper exports so we would not need to think about bundling.

But there are two more `.native.` files in source. One for logger and one for batched updates.

The logger is not needed anymore as someone mentioned in some ticket. And even if so it can be configured with QueryClient.

Not sure about batched updates though.

PR #4133, opened 2022-09-07
Docs: update mutate options description

SirMoustache:
Hey, in [`useMutation` options docs](https://tanstack.com/query/v4/docs/reference/useMutation) description for `onSuccess`, `onError`  and  `onSettled` says 

> If a promise is returned, it will be awaited and resolved before proceeding

So in this example, the mutation will wait for the promise in `onSuccess` to be resolved before changing mutation status and data

```tsx
{
  const { data, isLoading } = useMutation(
    () => {
      /**Some async stuff */
    },
    {
      onSuccess: () => {
        /** Will wait to be resolved before changing the mutation status */
        return delay(5000);
      },
    }
  );
}
```

It also says in `mutate` options 

> Remaining options extend the same options described above in the useMutation hook.

But if used like this the promise will not be awaited

```tsx
  const { data, isLoading, mutate } = useMutation(() => {
    /**Some async stuff */
  });

  mutate(undefined, {
    onSuccess: () => {
      /** Will NOT wait to be resolved before changing the mutation status */
      return delay(5000);
    },
  });
```

As I understand, this is intended behavior, so I want to update the docs to clarify this.

### DIFF
--- a/packages/query-core/src/types.ts
+++ b/packages/query-core/src/types.ts
@@ -579,19 +579,19 @@ export interface MutateOptions<
   onSuccess?: (
     data: TData,
     variables: TVariables,
-    context: TContext,
-  ) => Promise<unknown> | unknown
+    context: TContext | undefined,
+  ) => void
   onError?: (
     error: TError,
     variables: TVariables,
     context: TContext | undefined,
-  ) => Promise<unknown> | unknown
+  ) => void
   onSettled?: (
     data: TData | undefined,
     error: TError | null,
     variables: TVariables,
     context: TContext | undefined,
-  ) => Promise<unknown> | unknown
+  ) => void
 }
 
 export type MutateFunction<
--- a/packages/query-core/src/types.ts
+++ b/packages/query-core/src/types.ts
@@ -579,7 +579,7 @@ export interface MutateOptions<
   onSuccess?: (
     data: TData,
     variables: TVariables,
-    context: TContext | undefined,
+    context: TContext,
   ) => void
   onError?: (
     error: TError,
--- a/packages/query-core/src/types.ts
+++ b/packages/query-core/src/types.ts
@@ -576,11 +576,7 @@ export interface MutateOptions<
   TVariables = void,
   TContext = unknown,
 > {
-  onSuccess?: (
-    data: TData,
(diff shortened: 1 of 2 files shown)


TkDodo:
yeah it's intended, but I wouldn't phrase it as "if a promise is returned, it's not awaited". The signature is different, as those callbacks can only return `void` according to the typescript definitions.

SirMoustache:
> yeah it's intended, but I wouldn't phrase it as "if a promise is returned, it's not awaited". The signature is different, as those callbacks can only return `void` according to the typescript definitions.

Thank you for your reply, but are you sure about the return types?

Both [MutationOptions interface](https://github.com/TanStack/query/blob/main/packages/query-core/src/types.ts#L528)  and [MutateOptions](https://github.com/TanStack/query/blob/main/packages/query-core/src/types.ts#L573) have same return types for `onSuccess`, `onError` and `onSettled` functions, so maybe replacing it to `void` in MutateOptions could be a good solution?

TkDodo:
@SirMoustache you are absolutely right. I don't know why we should allow to return a Promise from the mutate callbacks, as we are not awaiting them.

> so maybe replacing it to void in MutateOptions could be a good solution?

Yes, I'd say so. Would you like to add that to this PR ?

SirMoustache:
@TkDodo I have done that, also I changed the wording in docs and explicitly provided a description for mutation callbacks

packages/query-core/src/types.ts
```diff
@@ -579,19 +579,19 @@ export interface MutateOptions<
   onSuccess?: (
     data: TData,
     variables: TVariables,
-    context: TContext,
-  ) => Promise<unknown> | unknown
+    context: TContext | undefined,
```
TkDodo, line 582: I think it was on purpose that `onSuccess` always gets the context passed. The type is inferred from whatever `onMutate` returns. It's just potentially undefined in `onError` and `onSettled` because the `onMutate` function is async and can throw / return a rejected Promise, in which case we have no context.

SirMoustache: Got it, I have created undo commit for this.

TkDodo (approved):

Issue #4097, opened 2022-08-29
Devtools are re-rerendering constantly on every cache update even when closed


pristas-peter:
### Describe the bug

We are encountering performance issues with the app when using devtools with many queries (around 1000). We are using normalised cache keys, when for example fetch of list items updates data in each single fetch query via setQueryData. 

However in v4, by using `useSyncExternalStore` things got really slow (app freezes for a short time when doing updates).

Is it possible to update devtools by batches, or is it even necessary to "observe" every query when the devtools are closed? 

### Your minimal, reproducible example

-

### Steps to reproduce

-

### Expected behavior

-

### Platform

-

### react-query version

v4 3.2.1

### TypeScript version

4.7.4

TkDodo:
Please provide a codesandbox reproduction!
By using useSyncExternalStore, we actually fixed a perf issue in the devtools:
- https://github.com/TanStack/query/issues/2742#issuecomment-993923506

pristas-peter:
Hi, check it out here:
https://codesandbox.io/s/suspicious-jasper-xh7k88?file=/src/App.js

Try to scroll beyond 500 items or so, the app will become un-responsible. 
If you comment out react query devtools component, all is smooth. 

This was not happening in react-query 3. 

Thanks for checking it out.

TkDodo:
i can confirm that this is not how it should be :)

Given that it gets worse with the number of queries, my best guess is that all uSES subscriptions are being triggered to re-render even though they should not. We've switched to fine-grained subscriptions because they are supposed to be more performant. Before that, every update would re-render the whole devtools, which is why we had the workaround of _not_ subscribing at all when the devtools are closed.

I see a couple of ways forward:

1) re-add the workaround that we removed in #2742. Basically, render `null` when the devtools are closed. That would be a quick win, and probably not wrong because why would we subscribe to anything if we are closed. However, app would still lag with open devtools.
2) investigate what re-renders so often and why, and stop those re-renders.
3) if the renders are not coming from uSES subscriptions themselves but from top-down renders, we could try memoizing each query row. But I don't think that's the case.

I'm afraid I won't have much time to look into this right now. Would you like to contribute any of those steps @pristas-peter ?

pristas-peter:
Yes, I will try to take a look at it next week

DhrjArun:
> i can confirm that this is not how it should be :)
> 
[...]

PR #4146, opened 2022-09-11
fix(devtools): not subscribing to queryCache if devtool is closed

DhrjArun:
This will stop unnecessary rerender of devtools when they are closed.

fix of issue #4097

### DIFF
--- a/packages/react-query-devtools/src/devtools.tsx
+++ b/packages/react-query-devtools/src/devtools.tsx
@@ -377,12 +377,18 @@ export function ReactQueryDevtools({
 const useSubscribeToQueryCache = <T,>(
   queryCache: QueryCache,
   getSnapshot: () => T,
+  skip: boolean = false,
 ): T => {
   return useSyncExternalStore(
     React.useCallback(
-      (onStoreChange) =>
-        queryCache.subscribe(notifyManager.batchCalls(onStoreChange)),
-      [queryCache],
+      (onStoreChange) => {
+        if (!skip)
+          return queryCache.subscribe(notifyManager.batchCalls(onStoreChange))
+        return () => {
+          return
+        }
+      },
+      [queryCache, skip],
     ),
     getSnapshot,
     getSnapshot,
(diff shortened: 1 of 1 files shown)


TkDodo:
wouldn't a simpler way be to conditionally render the whole `ReactQueryDevtoolsPanel` if we want to have it closed, possibly here:

https://github.com/TanStack/query/blob/ea80700c2c8d48c6d0cb4ae0165a424ebea5d35c/packages/react-query-devtools/src/devtools.tsx#L246

The variable to check is `!isResolvedOpen`

DhrjArun:
> 

Yeah, I tried that, but Transition(animation) gets removed.

packages/react-query-devtools/src/devtools.tsx
```diff
@@ -377,12 +377,18 @@ export function ReactQueryDevtools({
...
     React.useCallback(
-      (onStoreChange) =>
-        queryCache.subscribe(notifyManager.batchCalls(onStoreChange)),
-      [queryCache],
+      (onStoreChange) => {
+        if (!skip)
+          return queryCache.subscribe(notifyManager.batchCalls(onStoreChange))
+        return () => {
+          return
+        }
```
TkDodo, line 389: I think this could just be `return undefined`

so I would do, to avoid the negated condition:

```
return skip ? undefined : queryCache.subscribe(notifyManager.batchCalls(onStoreChange))
```

DhrjArun: Tried it

```ts
useSyncExternalStore<T>(subscribe: (onStoreChange: () => void) => () => void, getSnapshot: () => T, getServerSnapshot?: (() => T) | undefined): T
```

Screenshot 2022-09-11 at 23 20 18
Screenshot 2022-09-11 at 23 20 34

TkDodo: oh i didn't know that, I thought the API is the same as the one from useEffect. nevermind then 🙈

TkDodo (approved):

PR #4153, opened 2022-09-12
feat: add support for bigint keys

paul-sachs:
Support bigint as keys within the default hashQueryKey function.

### DIFF
--- a/packages/query-core/src/tests/utils.test.tsx
+++ b/packages/query-core/src/tests/utils.test.tsx
@@ -7,6 +7,7 @@ import {
   scheduleMicrotask,
   sleep,
   isPlainArray,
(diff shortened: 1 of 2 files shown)


TkDodo:
There are two problems with this:

1) BigInt(12) will be equivalent to "12" (the string), which might lead to key clashes and overwriting caches.
2) if we start to support things like `bigint`, questions for support about `Map`, `Set` etc. will come up, in which case we could just as well use `superjson`.

I think we'd rather stay JSON compatible per default. You can always `.toString()` the value when passing it to `useQuery`, or you can pass your own `queryKeyHashFn` in user land. You can set this up globally once and it will be used for all keys.

PR #4158, opened 2022-09-13
fix(queryObserver): allow refetch interval in service worker (#4018)

ElGrecode:
Resolves discussion in #4018 

Note: The use of `// eslint-disable-next-line` is an intentional choice. The lint rule `@typescript-eslint/no-unnecessary-condition` works when typing is known. In this case `globalThis` is scoped to `NodeJS`'s global namespace and therefore has the typings for only one environment of `globalThis`. It's a weird case which is why I put the ignore for this line only consistent with other uses of ignoring `no-unnecessary-condition`

### DIFF
--- a/packages/query-core/src/queryObserver.ts
+++ b/packages/query-core/src/queryObserver.ts
@@ -364,7 +364,8 @@ export class QueryObserver<
     this.currentRefetchInterval = nextInterval
 
     if (
-      isServer ||
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+      !globalThis.setInterval ||
       this.options.enabled === false ||
       !isValidTimeout(this.currentRefetchInterval) ||
       this.currentRefetchInterval === 0
@@ -372,7 +373,7 @@ export class QueryObserver<
       return
     }
 
-    this.refetchIntervalId = setInterval(() => {
+    this.refetchIntervalId = globalThis.setInterval(() => {
       if (
         this.options.refetchIntervalInBackground ||
         focusManager.isFocused()


TkDodo:
Thanks for the PR. I think the thing I have missed in the discussion is that `setInterval` also exists in nodeJs. So with that change, we would likely add `setInterval` calls during server side rendering, wouldn't we?

ElGrecode:
Thanks for the quick comment @TkDodo !

I was under the assumption from your comment that this subscription code would only be run in effects

> No, I think we can decouple setInterval calls from isServer checks. None if this code runs on the server anyways, because subscriptions are only created on the client (they run in effects / useSyncExternalStore, so not on the server)

TkDodo:
Huh I guess you're right. Let me think about it and test it some more. Thanks for the contribution 🙌

TkDodo:
what about if we would extend the `isServer` check to exclude the service worker environment? It seems like checking for a service worker can be done with checking for the `WorkerGlobalScope` ?

https://stackoverflow.com/questions/7931182/reliably-detect-if-the-script-is-executing-in-a-web-worker

I think other timeouts like `staleTime` and `cacheTime` would also profit from this?

ElGrecode:
> what about if we would extend the isServer check to exclude the service worker environment? It seems like checking for a service worker can be done with checking for the WorkerGlobalScope

Makes sense to me. I'll update

TkDodo:
@ElGrecode how's it going? Did you have time to update the PR?

ElGrecode:
Closing as I'm unable to dedicate time to it this month unfortunately. I'm trying to maintain good house keeping for these PRs. I will reopen with this context next month.

PR #4160, opened 2022-09-13
feat(useQuery): allow callback format for `enabled` option in useQuery

knyzorg:
Add support for using a callback in the `enabled` option, to allow queries to decide whether they should be refetched depending on the query. This saves the need to use predicates in invalidate calls.

There are no breaking changes introduced by this feature.

**I have not yet tested this, nor written unit tests**

### DIFF
--- a/packages/query-core/src/queryObserver.ts
+++ b/packages/query-core/src/queryObserver.ts
@@ -165,9 +165,10 @@ export class QueryObserver<
 
     if (
       typeof this.options.enabled !== 'undefined' &&
-      typeof this.options.enabled !== 'boolean'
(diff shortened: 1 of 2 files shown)


TkDodo:
isn't this the same as:
- #4030

?

knyzorg:
It appears so. I looked in the issues but didn't think to look into the PRs.

Is the linked PR to your satisfaction? 

I'm not here for the glory, I just really want to see the feature land.

knyzorg:
Is there anyone taking a look at the PR? This would be a major QOL improvement for me.
Would it possible to release that changeset as a beta/RC release?

With how the package is built, I don't think I could simply apply a patch to the package in my node_modules.

PR #4165, opened 2022-09-14
fix(react-query-persist-client): Make dependencies to peerDependencies

### LINKED ISSUE #4161: @tanstack/react-query-persist-client not working with yarn 3 and pnp
### Describe the bug

We are using yarn 3 and want to add offline functionality to our application. However, it seems that the module resolution somehow fails when we added the package `@tanstack/react-query-persist-client`. After replacing the default `QueryClientProvider` with `PersistQueryClientProvider`, we get the error message `Error: No QueryClient set, use QueryClientProvider to set one`.

After some debugging we found out that `@tanstack/react-query` is bundled two times and the `PersistQueryClientProvider` uses a different instance than `useQuery`, therefore not finding the query context.

The output of `yarn why @tanstack/react-query` shows this:

```
$ yarn why @tanstack/react-query
├─ @tanstack/react-query-persist-client@npm:4.3.4
│  └─ @tanstack/react-query@npm:4.3.4 [e51bb] (via npm:4.3.4 [e51bb])
│
└─ react-query-test@workspace:.
   └─ @tanstack/react-query@npm:4.3.4 [7208f] (via npm:4.3.4 [7208f])
```

https://github.com/amuttsch/react-query-persister-yarn-berry

### Steps to reproduce

```
yarn
yarn dev
```

Open localhost:3000 and you should see the error message Error: No QueryClient set, use QueryClientProvider to set one.

### Expected behavior

The QueryClient should be in the correct context and the error should not occur

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: MacOS Montary
- Browser: Chromium 106

### react-query version

4.3.4 / 4.3.5

### TypeScript version

_No response_

### Additional context

_No response_

amuttsch:
For yarn 3 pnp, having react-query as a normal dependencies causes it to be bundled multiple times. The query client set in react-query-persist-client is not visible for the normal useQuery imports from react-query. Setting react-query as a peerDependency instead fixes this issue.

Fixes #4161

### DIFF
--- a/packages/react-query-persist-client/package.json
+++ b/packages/react-query-persist-client/package.json
@@ -31,7 +31,7 @@
     "clean": "rm -rf ./build",
     "test:eslint": "../../node_modules/.bin/eslint --ext .ts,.tsx ./src"
   },
-  "dependencies": {
+  "peerDependencies": {
     "@tanstack/query-core": "4.3.7",
     "@tanstack/react-query": "4.3.7"
   }


TkDodo:
> For yarn 3 pnp, having react-query as a normal dependencies causes it to be bundled multiple times.

interesting - why would that be the case? This doesn't sound right.

without auto installation of `peerDependency`, it would mean that users would need to manually add a dependency to `@tanstack/query-core`, which is not really intended ...

amuttsch:
@TkDodo I'm not quite sure how the module resolution works, but it differs a bit from node_modules to pnp. Especially regarding peer dependencies, packages should put transitive dependencies into `peerDependencies` instead of regular `dependencies`. At least this is what I understood when reading about it.

But we can move `@tanstack/query-core` back to regular dependencies, but `@tanstack/react-query` must be a `peerDependency` in order to work. I tested this by installing `react-query-persist-client` from disk instead of npm and changing the `package.json`.

I changed the MR.

amuttsch:
We see this when we use `yarn why @tanstack/react-query`. With a peer dependency:

```
yarn why @tanstack/react-query
├─ @tanstack/react-query-persist-client@npm:4.3.4
│  └─ @tanstack/react-query@npm:4.3.4 [e51bb] (via npm:4.3.4 [e51bb])
│
└─ react-query-test@workspace:.
   └─ @tanstack/react-query@npm:4.3.4 [7208f] (via npm:4.3.4 [7208f])
```

and without the peer dependency:

```
yarn why @tanstack/react-query
├─ @tanstack/react-query-persist-client@file:../query/packages/react-query-persist-client#../query/packages/react-query-persist-client::hash=142afd&locator=react-query-test%40workspace%3A.
│  └─ @tanstack/react-query@npm:4.3.4 [e51bb] (via npm:4.3.4 [e51bb])
│
├─ @tanstack/react-query-persist-client@npm:4.3.4
│  └─ @tanstack/react-query@npm:4.3.4 [e51bb] (via npm:4.3.4 [e51bb])
│
└─ react-query-test@workspace:.
   └─ @tanstack/react-query@npm:4.3.4 [7208f] (via npm:4.3.4 [7208f])
```

We see that it is included again from my local installation too. I'm not quite sure why we see the npm version here too (@tanstack/react-query-persist-client@npm:4.3.4), but it works this way.

TkDodo:
If a package manager includes the same dependency twice, isn't this a bug in the package manager? For example, we also have `useSyncExternalStore` as a dependency in react-query, because we don't want to force users to install it themselves.

If you use `redux` additionally, they also depend on `useSyncExternalStore` (with the same version, as `dependency`). Does that mean you would get two versions bundled with pnp?

TkDodo:
I mean yeah, very likely, no one would use the `react-query-persist-client` _without_ also having a dependency on `react-query`, so I think the change might be alright. I'd just like to understand it better :)

amuttsch:
Unfortunately, I'm no expert in the whole module resolution /pnp topic, but my understanding is that if multiple packages depend on some dependency A with different versions, they have their own version accessible. If your library is an extension of some other library, you add it as a peer dependency, so the project that implements your library has to install the proper dependency version.

Yarn PNP even shows a warning message if you haven't installed your peer dependencies correctly, see https://yarnpkg.com/advanced/error-codes#yn0002---missing_peer_dependency and https://dev.to/arcanis/implicit-transitive-peer-dependencies-ed0

TkDodo:
> but my understanding is that if multiple packages depend on some dependency A with different versions, they have their own version accessible

that is for sure, but in our example, multiple packages depend on the same dependency A with _the same_ version, not different versions! And that's where I'm having trouble to understand why that wouldn't be deduped?

amuttsch:
I created a discussion on the yarn repo: https://github.com/yarnpkg/berry/discussions/4848

Hopefully we get some answers on why we have this behavior.

Would it be possible to get this merged even if we don't have a definite answer yet? I need the offline persister for my project and cannot fall back to `node_modules` resolution easily. Unfortunately, patching `@tanstack/react-query-persist-client` in my project doesn't work either. Thanks! :-)

TkDodo:
idk, we'd have to make sure that npm and normal yarn users wouldn't get peer dependency install errors because again, no one should need to specify a dependency on query-core directly. It's an implementation detail that the core is a separate package, which is why the react-query package bundles the core as a direct dependency.

but yeah if that's ensured it's fine

amuttsch:
`react-query` has a direct dependency on `query-core`, so defining it as a peer dependency for `react-query-persist-client` will be fine as users will usually install both `react-query` and `react-query-persist-client` as dependencies.

I updated the PR.

TkDodo (approved):

TkDodo:
can you please solve the merge conflicts, then we can merge this.

amuttsch:
Done.

PR #4177, opened 2022-09-15
Error boundary handling for useQueries

### LINKED ISSUE #2395: useErrorBoundary option works in useQueries?
### Discussed in https://github.com/tannerlinsley/react-query/discussions/2355

<sup>Originally posted by **luizeboli** June  8, 2021</sup>
Hello guys, I have this hook:

```
export const useOptions = () => {
  return useQueries([
    {
      queryKey: 'groups',
      queryFn: () => axios({ method: 'GET', url: '/groups' }),
      useErrorBoundary: true,
    },
    {
      queryKey: 'types',
      queryFn: () => axios({ method: 'GET', url: '/types' }),
      useErrorBoundary: true,
    },
    {
      queryKey: 'sourcedTypes',
      queryFn: () => axios({ method: 'GET', url: '/sourcedTypes' }),
      useErrorBoundary: true,
    },
  ])
}
```

The component that consumes this hook is wrapped by a `ErrorBoundary`, but it doesn't throw while using `useQueries`. It just works with `useQuery`. 

The only change needed to work correctly is replace `useQueries` by `useQuery`.

Is this an expected behavior? I didn't find it in docs</div>

zorzysty:
closes #2395


packages/react-query/src/useQueries.ts
```diff
@@ -188,5 +190,35 @@ export function useQueries<T extends any[]>({
     observer.setQueries(defaultedQueries, { listeners: false })
   }, [defaultedQueries, observer])
 
+  const errorResetBoundary = useQueryErrorResetBoundary()
+
+  defaultedQueries.forEach((query) => {
```
TkDodo, line 195: can we rename this from `queries` to `options` ? I think this is highly confusing right now 😅

packages/react-query/src/useQueries.ts
```diff
@@ -188,5 +190,35 @@ export function useQueries<T extends any[]>({
...
+    if (query.suspense || query.useErrorBoundary) {
+      // Prevent retrying failed query if the error boundary has not been reset yet
+      if (!errorResetBoundary.isReset()) {
+        query.retryOnMount = false
+      }
```
TkDodo, line 200: could this be a single, abstracted function used in useBaseQuery as well? Something like:

(I'm using options instead of query already, see my previous comment)

```
ensurePreventErrorBoundaryRetry(options, errorResetBoundary)
```

packages/react-query/src/useQueries.ts
```diff
@@ -188,5 +190,35 @@ export function useQueries<T extends any[]>({
...
+
+  const firstSingleResultWhichShouldThrow = result.find(
+    (singleResult, index) =>
+      singleResult.isError &&
+      !errorResetBoundary.isReset() &&
+      !singleResult.isFetching &&
+      shouldThrowError(defaultedQueries[index]?.useErrorBoundary ?? false, [
+        singleResult.error,
+        observer.getQueries()[index]!,
+      ]),
```
TkDodo, line 216: I also think that just this part could be a single extracted util function

TkDodo (commented):
very good work ❤️

packages/react-query/src/errorBoundaryUtils.ts
```diff
@@ -0,0 +1,67 @@
...
+  result: QueryObserverResult<TData, TError>
+  errorResetBoundary: QueryErrorResetBoundaryValue
+  useErrorBoundary: boolean | TUseErrorBoundaryFn | undefined
+  query: Query<TQueryFnData, TError, TQueryData, TQueryKey>
+}) => {
+  return (
+    result.isError &&
+    !errorResetBoundary.isReset() &&
+    !result.isFetching &&
+    shouldThrowError(useErrorBoundary, [result.error, query] as Parameters<TUseErrorBoundaryFn>)
```
zorzysty, line 65: Wonder if this `as Parameters<TUseErrorBoundaryFn>` could be avoided somehow

packages/react-query/src/useQueries.ts
zorzysty: [237912b](https://github.com/TanStack/query/pull/4177/commits/237912b2459af5df2397416a57cb185ac3dc534f)

packages/react-query/src/useQueries.ts
zorzysty: [[7f81e88](https://github.com/TanStack/query/pull/4177/commits/7f81e8893bd674f103d6b72a93b77151abea86a3)](https://github.com/TanStack/query/pull/4177/commits/7f81e8893bd674f103d6b72a93b77151abea86a3#diff-b44eb6740558a471060188bab4175b44802e0b0d56d4be245cc87acc748e44fdR11)

packages/react-query/src/useQueries.ts
zorzysty: [7f81e88](https://github.com/TanStack/query/pull/4177/commits/7f81e8893bd674f103d6b72a93b77151abea86a3)

packages/react-query/src/__tests__/useQueries.test.tsx
```diff
@@ -1142,6 +1142,7 @@ describe('useQueries', () => {
         fallbackRender={({ error }) => (
           <div>
             <div>error boundary</div>
+            {/* @ts-expect-error `error` here is actually of type `string` and not `Error` */}
```
TkDodo, line 1145: Can we just reject the queryFn with a new Error instead?

zorzysty: [c92a897](https://github.com/TanStack/query/pull/4177/commits/c92a8976fc60c70beb6fc43bf3a1798a19d1e935)

packages/react-query/src/__tests__/useQueries.test.tsx
```diff
@@ -1104,4 +1104,55 @@ describe('useQueries', () => {
...
+              Promise.reject(
+                new Error(
+                  'this should not throw because useErrorBoundary is not set',
+                ),
+              ),
+          },
+          {
+            queryKey: key2,
+            queryFn: () => Promise.reject(new Error('single query error')),
+            useErrorBoundary: true,
```
TkDodo, line 1128: can we also have a test where `useErrorBoundary` is a function?

packages/react-query/src/errorBoundaryUtils.ts
TkDodo: it works if we hardcode the type of `useErrorBoundary` without the extra generic:

```
useErrorBoundary: boolean | undefined | ((
    e: TError,
    q: Query<TQueryFnData, TError, TQueryData, TQueryKey>,
  ) => boolean)
```

maybe it would just be good to extract this type:

https://github.com/TanStack/query/blob/f828528c935089a3adbf7b0474c6c91603ff47e4/packages/query-core/src/types.ts#L208-L213

and then re-use it here?

zorzysty: Thanks, I'll try it out.

packages/react-query/src/__tests__/useQueries.test.tsx
zorzysty: [3fd6426](https://github.com/TanStack/query/pull/4177/commits/3fd6426d0ccf7c200a8e715d22aa69d98a57b886)

packages/react-query/src/errorBoundaryUtils.ts
zorzysty: [3518f3c](https://github.com/TanStack/query/pull/4177/commits/3518f3ca403d52c8be23073c9d66af8a250e34b8)

TkDodo (approved):

packages/react-query/src/useQueries.ts
TkDodo: sorry, I meant more globally that `defaultedQueries` should also probably be `defaultedOptions`. It's not a query 😅

TkDodo: okay nevermind, this is actually a bigger refactoring. we have `queries` everywhere and also `setQueries`...

zorzysty: I was struggling to grasp the naming conventions here, and when I thought I finally got it, this comment got me confused again 😁

Let me explain how I understand it. First, here are 3 facts about the code in this PR its current state:
1. `useQueries` takes `queries` array as an argument
```ts
export function useQueries<T extends any[]>({
  queries: queriesOptions,
  context,
}: {
  queries: readonly [...QueriesOptions<T>]
  context?: UseQueryOptions['context']
})
```
2. Each item of `queries` array is a _query_, but _query_ is basically an `options` object
3. These `queries` are then _defaulted_ and stored as `defaultedQueries`:
```ts
const defaultedQueries = React.useMemo(
  () =>
    queriesOptions.map((options) => {
      const defaultedOptions = queryClient.defaultQueryOptions(options)

      // Make sure the results are already in fetching state before subscribing or updating options
      defaultedOptions._optimisticResults = isRestoring
        ? 'isRestoring'
        : 'optimistic'

      return defaultedOptions
    }),
  [queriesOptions, queryClient, isRestoring],
)
```

It would be strange if mapping with default values would suddenly convert `queries` to `options`. Furthermore in the code below `QueriesObserver` constructor has second parameter named `queries` and not `options` (although `queries` is typed as `QueryObserverOptions`, because why not xD).
```ts
const [observer] = React.useState(
  () => new QueriesObserver(queryClient, defaultedQueries),
)
```


So here are my thoughts to sum it up:
If we rename `defaultedQueries` to `defaultedOptions` it would be confusing because `options` are of a single query and not of `queries` array (`useQueries` does not have `options` on its own). To make it more consistent and less confusing, I would instead suggest to take the example from how `queries` are typed in `useQueries` arguments (type is called `QueriesOptions`) and do the name change this way:

`defaultedQueries` -> `defaultedQueriesOptions`

WDYT?

If that makes sense, maybe we could also do this to make it even more consistent:
```ts
export function useQueries<T extends any[]>({
  queries: queriesOptions,
  context,
}
```

TkDodo: yeah the naming is weird, I agree. But we have a public method called `setQueries` that actually takes `options` in, but they are named `queries` :/

so for now, let's:

- revert the lats commit (sorry) and just keep it `queries` for now. Being consistent is better I think.

there are also conflicts now, sorry. If those are solved we can go ahead and merge 🚀

PR #4180, opened 2022-09-16
Get rid of TS errors in tests files

zorzysty:
Test files in most packages were importing `tests/utils` from the root of the repo causing hundreds of typescript errors. This commit collocates these utils in each of the packages and removes 'exclude' statements from all tsconfig files as they are no longer necessary for the typechecking to pass.


TkDodo (approved):
❤️

Issue #4155, opened 2022-09-12
Webpack build issue v4.3.4


parkm:
### Describe the bug

When running webpack `webpack --config webpack.config.js` I'm getting this error on v4.3.4:
```
ERROR in ./node_modules/@tanstack/react-query/build/lib/reactBatchedUpdates.mjs 3:32-64
Can't import the named export 'unstable_batchedUpdates' from non EcmaScript module (only default export is available)
 @ ./node_modules/@tanstack/react-query/build/lib/setBatchUpdatesFn.js
 @ ./node_modules/@tanstack/react-query/build/lib/index.js
 @ ./src/root.component.js
 @ ./src/communications-ui.js
```

This does not happen on v4.3.3, any ideas what happened here?

### Your minimal, reproducible example

n/a

### Steps to reproduce

Build app with webpack on v4.3.4

### Expected behavior

I should not get import errors when building with webpack

### How often does this bug happen?

Every time

### Platform

MacOS v11.6.5
Webpack v4.41.6

### react-query version

4.3.4

parkm:
It looks like this commit might have introduced the problem https://github.com/TanStack/query/commit/c5d8582b25a8242bf9e6e632d9551e6b29b21713

TkDodo:
yes, we're aware of this by now, but it's good we are tracking it as an issue now. According to [this stackoverflow answer](https://stackoverflow.com/questions/71054449/cant-import-the-named-export-directive-from-non-ecmascript-module-only-defau/71059718#71059718), the bundler picks up `.mjs` files but doesn't understand ESM.

If you are using raw webpack, this config seems to do the trick:

```
rules: [
  {
    test: /\.mjs$/,
    include: /node_modules/,
    type: "javascript/auto"
  }
]
```

if you are using CRA v4, the best course of action is to upgrade to V5, or downgrade RQ until we have a fix. We had it working for CRA v4 in the `beta` but I think, as you've correctly pointed out, in an attempt to fix react-native support, this issue was somehow introduced.

Blackglade:
What version would you recommend downgrading to in the meantime while we wait for a fix?

nosovsh:
Thanks @TkDodo 

To summarize the quick fix:
if you cannot upgrade CRA to v5 do this:
`npm i @tanstack/react-query@4.3.3 --save`

and fix react-query version in your package.json (remove ^)

TkDodo:
[...]

Issue #4154, opened 2022-09-12
@tanstack/react-query@4.3.4 react-native example does not work - queryFn not being called


tomskopek:
### Describe the bug

I'm trying to migrate from react-query v3. I couldn't get react-query v4.3.4 to work in my react-native app, so I eventually tried the [demo app](https://github.com/TanStack/query/tree/main/examples/react/react-native).

The react-native example provided in the repo also doesn't work.
Specifically, the queryFn doesn't seem to be getting called. You can verify this by noting that `console.log` in `fetchMovies()` never gets called.

I downgraded the example app to use react-query@4.0.10, a version I have working in my react web project. I also tried 4.2.3, the last version before 4.3 minor version was released.

Now the `console.log` in `fetchMovies()` does get called, as expected.
However, note that now `data` seems to be returning a function, instead of an array, as I'd expect, so the example repo still is not working out of the box.

Thanks!

### Your minimal, reproducible example

https://github.com/TanStack/query/tree/main/examples/react/react-native

### Steps to reproduce

Run the expo code example provided in the package, with react-query set to v4.3.4

### Expected behavior

As a user, I expect to see a movie list, but I do not.

### Platform

- OS - iOS 12.5.1
- iPhone 13 Pro Max simulator 15.2

### react-query version

4.3.4

TkDodo:
yeah two different things here:
1. the example needs to be updated, but that's a minor thing.
1. queryFn not being called. I've seen that one before, it has to do with `useSyncExternalStore`. I think it was this issue:
- https://github.com/TanStack/query/issues/3582

TkDodo:
@DamianOsipiuk FYI

TkDodo:
upstream issue:
- https://github.com/facebook/react/pull/25231

Gedoz:
Me too, downgrade to v4.2.3 and works fine!

Mari-yun:
我在写小程序时也遇到了这样的问题，降级到3，正常

gagamil:
I have updated my expo based project to the latest expo v 46 (react 18) (from v 45) and queryFn is being called.

Willham12:
can confirm the issue!

TkDodo:
@Willham12 @gagamil @Gedoz @tomskopek can you please try with this version?

```
"@tanstack/react-query": "https://pkg.csb.dev/TanStack/query/commit/fd83a352/@tanstack/react-query",
```

it's from the preview build of this PR:
[...]

PR #4183, opened 2022-09-16
fix: serve esm.js for webpack 4

DamianOsipiuk:
@TkDodo Quick check:
- scripts v4
scipts4
- scripts 5
scripts5

Both seems to work

bundle

Could you please check with react-native 🙏 

Fixes: https://github.com/TanStack/query/issues/4155

### DIFF
--- a/packages/query-async-storage-persister/package.json
+++ b/packages/query-async-storage-persister/package.json
@@ -12,6 +12,7 @@
   },
   "types": "build/lib/index.d.ts",
   "main": "build/lib/index.js",
+  "module": "build/lib/index.esm.js",
   "exports": {
     ".": {
       "types": "./build/lib/index.d.ts",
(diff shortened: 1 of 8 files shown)


TkDodo:
> Could you please check with react-native 🙏

not sure how I would check react-native 😅 

@hirbod @arnaudbzn could you quickly try this version with react-native?

```
"@tanstack/react-query": "https://pkg.csb.dev/TanStack/query/commit/fd83a352/@tanstack/react-query",
```

thanks 🙏

TkDodo:
@DamianOsipiuk it seems like react-native is broken anyways because of the issue with uSES ?

- https://github.com/TanStack/query/issues/4154

So I think we can merge this because it has no negative effect on react-native

DamianOsipiuk:
Yup. Will try to fix native in another one

PR #4185, opened 2022-09-18
fix: react native uSES

### LINKED ISSUE #4154: @tanstack/react-query@4.3.4 react-native example does not work - queryFn not being called

DamianOsipiuk:
@TkDodo
This is a potential fix following the same pattern as other `.native` files.
Unfortunately i do not have a good way to test it out. So people that have a reliable native setup should say if it works as intended.

fixes #4154

### DIFF
--- a/packages/react-query/src/useSyncExternalStore.native.ts
+++ b/packages/react-query/src/useSyncExternalStore.native.ts
@@ -1,3 +1,5 @@
 // Temporary workaround due to an issue with react-native uSES - https://github.com/TanStack/query/pull/3601
 // @ts-ignore
-export { useSyncExternalStore } from 'use-sync-external-store/shim/index.native.js'
+import { useSyncExternalStore } from 'use-sync-external-store/shim/index.native.js'
+
+export { useSyncExternalStore }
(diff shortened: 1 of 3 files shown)


TkDodo:
this feedback is good enough for me:

- https://github.com/TanStack/query/issues/4154#issuecomment-1250611072

shipping it

PR #4190, opened 2022-09-19
Feature/extended filters get query data

ecyrbe:
Hello @TkDodo @tannerlinsley ,

This PR is following this discussion :  
https://twitter.com/ecyrbedev/status/1571561333992079363

Consider the following pattern :  
image

The idea is now that people are using the factory pattern to create keys and `useQueryOptions` in general, it might be helpful to be consistent for every function and allow to pass directly the result of the factory to all functions.

People might want to have everywhere the same ability to pass the factory result like they do for :  
- useQuery
- useInfiniteQuery
- fetchQuery

This PR add this possibility to : 
- getQueryData

Might also need to do it on (i can improve this PR upon agreement, after fixing details) :  
- getQueriesData

The solution i took is to be backward compatible with current `getQueryData`, but also allow to pass only one parameter that would be the filter + ignore the additionnal parameters.

For this i added an `ExtendedQueryFilters` :  
```ts
export interface ExtendedQueryFilters<
  TQueryFnData = unknown,
  TError = unknown,
  TData = TQueryFnData,
> extends Omit<QueryFilters, 'queryKey'>,
    Omit<QueryOptions<TQueryFnData, TError, TData, TQueryKey>, 'queryKey'> {
  queryKey: TQueryKey
}
```
```ts
getQueryData<TData = unknown>(
    filters:  ExtendedQueryFilters,
  ): TData | undefined
```

### DIFF
--- a/packages/query-core/src/queryClient.ts
+++ b/packages/query-core/src/queryClient.ts
@@ -10,7 +10,6 @@ import {
   MutationFilters,
   functionalUpdate,
   ExtendedQueryFilters,
-  isQueryKey,
 } from './utils'
 import type {
   QueryClientConfig,
@@ -119,11 +118,7 @@ export class QueryClient {
     arg1: QueryKey | ExtendedQueryFilters<TData, any, TData, QueryKey>,
     arg2?: QueryFilters,
   ): TData | undefined {
-    if (isQueryKey(arg1))
-      return this.queryCache.find<TData>(arg1, arg2)?.state.data
-    if (arg1.queryKey)
-      return this.queryCache.find<TData>(arg1.queryKey, arg1)?.state.data
-    return undefined
+    return this.queryCache.find<TData>(arg1, arg2)?.state.data
   }
 
   getQueriesData<TData = unknown>(queryKey: QueryKey): [QueryKey, TData][]
--- a/packages/query-core/src/queryClient.ts
+++ b/packages/query-core/src/queryClient.ts
@@ -9,6 +9,8 @@ import {
   hashQueryKeyByOptions,
   MutationFilters,
   functionalUpdate,
+  ExtendedQueryFilters,
+  isQueryKey,
 } from './utils'
 import type {
   QueryClientConfig,
@@ -107,10 +109,21 @@ export class QueryClient {
   }
 
   getQueryData<TData = unknown>(
+    filters: ExtendedQueryFilters<TData, any, TData, QueryKey>,
+  ): TData | undefined
+  getQueryData<TData = unknown>(
     queryKey: QueryKey,
     filters?: QueryFilters,
+  ): TData | undefined
+  getQueryData<TData = unknown>(
+    arg1: QueryKey | ExtendedQueryFilters<TData, any, TData, QueryKey>,
+    arg2?: QueryFilters,
   ): TData | undefined {
-    return this.queryCache.find<TData>(queryKey, filters)?.state.data
+    if (isQueryKey(arg1))
+      return this.queryCache.find<TData>(arg1, arg2)?.state.data
+    if (arg1.queryKey)
+      return this.queryCache.find<TData>(arg1.queryKey, arg1)?.state.data
+    return undefined
   }
 
   getQueriesData<TData = unknown>(queryKey: QueryKey): [QueryKey, TData][]
--- a/docs/reference/QueryClient.md
+++ b/docs/reference/QueryClient.md
@@ -168,12 +168,12 @@ The options for `prefetchInfiniteQuery` are exactly the same as those of [`fetch
 `getQueryData` is a synchronous function that can be used to get an existing query's cached data. If the query does not exist, `undefined` will be returned.
 
 ```tsx
-const data = queryClient.getQueryData(queryKey)
+const data = queryClient.getQueryData(queryKey| filters | queryOptions)
 ```
 
 **Options**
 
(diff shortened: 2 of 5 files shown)


packages/query-core/src/queryClient.ts
```diff
@@ -106,11 +107,22 @@ export class QueryClient {
...
+  ): TData | undefined
+  getQueryData<TData = unknown, Filters extends QueryFilters = QueryFilters>(
+    arg1: QueryKey | Filters,
+    arg2?: Filters,
   ): TData | undefined {
-    return this.queryCache.find<TData>(queryKey, filters)?.state.data
+    if (isQueryKey(arg1))
+      return this.queryCache.find<TData>(arg1, arg2)?.state.data
+    if (arg1.queryKey)
+      return this.queryCache.find<TData>(arg1.queryKey, arg1)?.state.data
```
TkDodo, line 124: I think we have a util function that normalizes query key / filters (`parseQueryArgs`)

TkDodo: also, this seems like we could pass a filter only without a query key, and that should not be allowed on type level imo.

packages/query-core/src/queryClient.ts
```diff
@@ -106,11 +107,22 @@ export class QueryClient {
...
   }
 
+  getQueryData<TData = unknown, Filters extends QueryFilters = QueryFilters>(
+    filters: Filters,
+  ): TData | undefined
   getQueryData<TData = unknown>(
     queryKey: QueryKey,
     filters?: QueryFilters,
+  ): TData | undefined
+  getQueryData<TData = unknown, Filters extends QueryFilters = QueryFilters>(
```
TkDodo, line 117: the generic trick is nice, but it sadly won't work if we provide a generic to TData:

```
const query = {
  queryKey,
  queryFn
}
queryClient.getQueryData<string>(query)
```

this will still fail because TypeScript has no partial inference for generics, so because we provide the first generic, the second one will default to `QueryFilters`, and then TS complains that we cannot pass the `queryFn`...
and in order to work with `getQueryData` in a meaningful way in TypeScript, it is necessary to supply the `TData` generic.

ecyrbe: Indeed.
So another option would be the solution of my first commit ?

packages/query-core/src/queryClient.ts
ecyrbe: fixed.

packages/query-core/src/queryClient.ts
ecyrbe: One cool thing about using the extended filters, is that if your factory returns has a `queryFn` the return type is now inferred : 
```ts
const query = {
  queryKey,
  queryFn // return a string
}
const data = queryClient.getQueryData(query);
//        ^? typeof data : string
const data = queryClient.getQueryData<string>(query); // stil works
const data = queryClient.getQueryData<number>(query); // typescript error number and queryFn missmatch
```

TkDodo:
what about `getQueriesData` ? Would we need that there as well?

packages/query-core/src/queryClient.ts
```diff
@@ -106,11 +107,18 @@ export class QueryClient {
     return this.mutationCache.findAll({ ...filters, fetching: true }).length
   }
 
+  getQueryData<TData = unknown>(
+    filters: ExtendedQueryFilters<TData, any, TData, QueryKey>,
```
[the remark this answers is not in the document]
ecyrbe: I think so. I'll try it

ecyrbe:
> what about `getQueriesData` ? Would we need that there as well?

Yes, if you agree, i'll also add it

TkDodo:
> Yes, if you agree, i'll also add it

yep, awesome. I am trying to think if there is a situation where this will bite us, like when we introduce new query filters, they cannot overlap in namings with a query option that we already have, or vice-versa?

like, we have a query filter called `type`, and if we add an option called `type` later, this will clash.

For typesafety, we would only need `{ queryKey, queryFn }` I think. If we pick only those two fields, the problem would be that we couldn't pass "more" to it, right?

So:

```
const query = {
  queryKey,
  staleTime
  queryFn // return a string
}

queryClient.getQueryData(query)
```

would then become invalid because it contains `staleTime` ? Still, I think it would be the "less risky" approach ... What do you think?

ecyrbe:
You are correct on all points. So maybe, it would be safer to even go further. 
Because what about a user that wants to have filters alongside of is refuse of a Factory function?
So maybe we should not mix them at all. And have two simple overloads:
```ts
getQueryData(queryKey: QueryKey, filters: Filters);
// and
getQueryData(queryOptions: QueryOptions, filters: Filters);
```
What do you think?

TkDodo:
Yes I think that would work 👍

ecyrbe:
For information, i'm strugling with where to handle this new overload.
Internally no current parseArgs helpers handles this use case, and modifying `find` and `findAll` to accept queryOptions would break everything.
So i'm currently trying to handle this use case by adding a dedicated parse helper that will not break everything else.
I understand why @tannerlinsley said he would not design with overload if he had to do this now.
But i'll try to finish this week.

ecyrbe:
i'm sorry dom, but i can't make this work smoothly.
internally everything is using unpredictable overloads. And since code coverage is not there, i can't guaranty i'll not break any user code out there.
So i'll back down on this PR.
Again sorry.

TkDodo:
no worries, thank you for trying ❤️

PR #4193, opened 2022-09-20
docs: add react query kit community contribution

liaoliao666:
this PR is for adding `react-query-kit` package to the docs.

I recently wrote this package to get rid of the tedious management of `queryKey` and make `queryKey` strongly related to `queryFn`.

Maybe some people would like to use it.

### DIFF
--- a/docs/community/liaoliao666-react-query-kit.md
+++ b/docs/community/liaoliao666-react-query-kit.md
@@ -9,8 +9,8 @@ title: React Query Kit
 
 - Make `queryKey` strongly related with `queryFn`
 - Manage `queryKey` in a type-safe way
-- Generate quickly a custom ReactQuery hook
-- Make `queryClient`'s operations more clearly associated with which custom ReactQuery hook
+- Generate a custom ReactQuery hook quickly
+- Make `queryClient`'s operations clearly associated with custom ReactQuery hooks
 - Set defaultOptions for custom ReactQuery hooks easier and clearer
 
 ## Installation
--- a/docs/community/liaoliao666-react-query-kit.md
+++ b/docs/community/liaoliao666-react-query-kit.md
@@ -3,7 +3,7 @@ id: liaoliao666-react-query-kit
 title: React Query Kit
 ---
 
-🕊️ A toolkit for ReactQuery that make ReactQuery hooks reusable and typesafe
+🕊️ A toolkit for ReactQuery that makes ReactQuery hooks reusable and typesafe
 
 ## what do you benefit from it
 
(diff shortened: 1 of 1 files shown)


docs/community/liaoliao666-react-query-kit.md
```diff
@@ -0,0 +1,77 @@
...
+  )
+}
+
+console.log(usePost.getKey()) //  ['/posts']
+console.log(usePost.getKey(variables)) //  ['/posts', { id: 1 }]
+
+export async function getStaticProps() {
+  const queryClient = new QueryClient()
+
+  await queryClient.prefetchQuery(usePost.getKey(variables), usePost.queryFn)
```
TkDodo, line 67: wouldn't `usePost.prefetch()` be nice?

liaoliao666: I wanna the custom hook is agnostic with which `queryClient`.

liaoliao666: If the api design as this. it would also get a problem form such api.
```
// Here is type definition
type prefetch = (variables?: Variables, options?:FetchQueryOptions) => Promise<void>

// If the `prefetch` do not need pass `variables`
// but I need pass `options`
// it will be like below. it's ugly
usePost.prefetch(undefined, options)
```
Even If I can do it. `queryClient` also has many apis, do I need to provide all of these api😂.

I don't think it's a wise choice

docs/community/liaoliao666-react-query-kit.md
```diff
@@ -0,0 +1,77 @@
...
+title: React Query Kit
+---
+
+🕊️ A toolkit for ReactQuery that make ReactQuery hooks reusable and typesafe
+
+## what do you benefit from it
+
+- Make `queryKey` strongly related with `queryFn`
+- Manage `queryKey` in a type-safe way
+- Generate quickly a custom ReactQuery hook
```
TkDodo, line 12:
```suggestion
- Generate a custom ReactQuery hook quickly
```

docs/community/liaoliao666-react-query-kit.md
```diff
@@ -0,0 +1,77 @@
...
+- Make `queryClient`'s operations more clearly associated with which custom ReactQuery hook
```
TkDodo, line 13: not sure about this sentence, it doesn't make much sense grammatically. Maybe:

```suggestion
- Make `queryClient`'s operations clearly associated with custom ReactQuery hooks
```

?

liaoliao666: It's done. thz u so much for check this.

PR #4200, opened 2022-09-21
Added Turborepo

mattpocock:


package.json
```diff
@@ -81,7 +79,6 @@
     "babel-eslint": "^10.1.0",
     "babel-plugin-transform-async-to-promises": "^0.8.18",
     "bundlewatch": "^0.3.2",
-    "concurrently": "^7.1.0",
```
TkDodo, line 84: I think we could remove the lerna dependency as well, right?

turbo.json
```diff
@@ -0,0 +1,42 @@
...
+    "test:eslint": {
+      "outputs": []
+    },
+    "//#typecheck": {
+      "outputs": []
+    },
+    "//#format:check": {
+      "outputs": []
+    },
+    "//#build": {
```
TkDodo, line 33: can you explain or point to the documentation about what this `//#` syntax means? It looks pretty foreign to me 😅

mattpocock: Yes, here you go!

https://turborepo.org/docs/core-concepts/running-tasks#running-tasks-from-the-root

TkDodo:
Now that we have a solid-query package, a runner like turborepo becomes more important. For example, we wouldn't want to run tests for the react-query package if we only have changes in solid-query. However, if we have changes in query-core, we have to run all tests because every package depends on it.

@mattpocock is this something turborepo would help us with? We're using NX at work and it knows about dependencies. Would you be willing to help us get this PR in a working state? Thank you 🙏

TkDodo:
I will close this PR for now as it doesn't seem like we're getting anywhere with it, and it will likely be easier to start from scratch.

If someone wants to pick this up, please let me know and go ahead.

Otherwise, I will look into adding `NX` somewhen in the near future, simply because I know how it works 😅

PR #4213, opened 2022-09-24
feat(query-core): log queryKey in createRetryer

fobos531:
@TkDodo As per our chat on Discord, I made a small change in the error that throws in `createRetryer`'s `onSuccess` callback if the fetched data turns out to be `undefined`. The error now includes the affected `queryKey` - I have confirmed this in my personal project and successfully resolved the issue I had that way.

Open to any suggestions.

Thanks!

### DIFF
--- a/packages/query-core/src/query.ts
+++ b/packages/query-core/src/query.ts
@@ -452,7 +452,13 @@ export class Query<
       abort: abortController?.abort.bind(abortController),
       onSuccess: (data) => {
         if (typeof data === 'undefined') {
-          onError(new Error('Query data cannot be undefined') as any)
+          onError(
+            new Error(
+              `Query data cannot be undefined - affected query key: ${JSON.stringify(
+                this.queryKey,
+              )}`,
+            ) as any,
+          )
           return
         }
 
--- a/packages/query-core/src/query.ts
+++ b/packages/query-core/src/query.ts
@@ -454,9 +454,7 @@ export class Query<
         if (typeof data === 'undefined') {
           onError(
             new Error(
-              `Query data cannot be undefined - affected query key: ${JSON.stringify(
-                this.queryKey,
-              )}`,
+              `Query data cannot be undefined - query key in question: ${this.queryHash}`,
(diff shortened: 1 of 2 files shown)


packages/query-core/src/query.ts
```diff
@@ -452,7 +452,13 @@ export class Query<
       abort: abortController?.abort.bind(abortController),
       onSuccess: (data) => {
         if (typeof data === 'undefined') {
-          onError(new Error('Query data cannot be undefined') as any)
+          onError(
+            new Error(
+              `Query data cannot be undefined - affected query key: ${JSON.stringify(
+                this.queryKey,
+              )}`,
+            ) as any,
```
TkDodo, line 460: awesome, thank you. What we should likely use instead of stringifying the querykey is just `this.queryHash`, as this contains the stringified version already, and also respects custom hash functions passed in by the user.

TkDodo:
also you'd need to amend a test please:

```
FAIL packages/query-core/src/tests/query.test.tsx
  ● query › fetch should dispatch an error if the queryFn returns undefined

    expect(received).toMatchObject(expected)

    - Expected  - 1
    + Received  + 1

      Object {
    -   "error": [Error: Query data cannot be undefined],
    +   "error": [Error: Query data cannot be undefined - affected query key: ["query_30"]],
        "isError": true,
      }

      819 |     const error = new Error('Query data cannot be undefined')
      820 |
    > 821 |     expect(observerResult).toMatchObject({
          |                            ^
      822 |       isError: true,
      823 |       error,
      824 |     })

      at Object.<anonymous> (packages/query-core/src/tests/query.test.tsx:821:28)
```

packages/query-core/src/query.ts
fobos531: Awesome, I wasn't aware of that one!

fobos531:
I have applied your suggestions - all tests pass now locally.

Let me know if you need anything else.

TkDodo (approved):

PR #4218, opened 2022-09-25
fix(query): log error about undefined query data being returned

TkDodo:


TkDodo:
@fobos531 this is a follow-up to your fix from
- #4213 

I think that using our logger to log the error properly during development mode is better, as we can be more explanatory in the message (as verbose as we want actually). The error thrown at runtime will remain in the production bundle, so I'd prefer if it were not as verbose. Thoughts?

fobos531:
> @fobos531 this is a follow-up to your fix from
> 
> * [feat(query-core): log queryKey in createRetryer #4213](https://github.com/TanStack/query/pull/4213)
> 
> I think that using our logger to log the error properly during development mode is better, as we can be more explanatory in the message (as verbose as we want actually). The error thrown at runtime will remain in the production bundle, so I'd prefer if it were not as verbose. Thoughts?

Hey @TkDodo 

I see where you're coming from. Is there a particular reason why you'd want the error to not be as verbose in production? I would assume logging framework such as Sentry would pick up on this and developers could use this information to help pinpoint the issue faster even if this happened in production. I'm not entirely sure how TanStack Query's logger works so I don't know how this change would affect existing developer workflows. I think you're the better person for that :D

If I were developing an app in, say, React Native, how would I be made aware of this error (using the logger) given this change?

I'm okay with this change, I just think most of my concerns come from the fact I'm not entirely familiar with how tanstack query's logging system works. I'd appreciate any insight!

TkDodo:
> Is there a particular reason why you'd want the error to not be as verbose in production?

mainly bundle size. those texts stay in the bundle as-is, which is why they compound really quickly. libraries like `react` will only give you an error code, and you can translate that on their webpage to a real error with description (in https://reactjs.org/docs/error-decoder.html)

I think for this case, it is fine to just log the error in development mode. It only happens if you forget to return something from the queryFn, or if you explicitly return undefined. You _should_ be able to see that during development.

> If I were developing an app in, say, React Native, how would I be made aware of this error (using the logger) given this change?

it would log to the console in development mode and the console does work in react-native as well.

fobos531:
Makes sense. Thank you for the explanation! LGTM

PR #4224, opened 2022-09-26
 test(solid-query) Add test for transition

jeski-bright:
Add test for running a query during transition.

~~Change `createResource` usage to only indicate the suspended loading state. Always serve the underlying data from the query state. Do not trigger mutations and refetches as often.~~

~~This is a workaround for the following issue: https://github.com/solidjs/solid/issues/1249.~~


TkDodo:
Hi, let's wait for @ardeora to review your change, but in the meantime:

- please create a test case that fails with the current implementation on `main`, but doesn't fail with your fix.

thank you 🙏

ardeora:
Thanks! Yeah this a very interesting/smart approach. I had looked at something similar before but the current implementation worked fine. Though, I'm curious as to which exact part of the current implementation is breaking transitions. I'm at work right now. But will take a deeper look at it in a few hours.

packages/solid-query/src/createBaseQuery.ts
```diff
@@ -76,7 +68,21 @@ export function createBaseQuery<
...
-  onCleanup(() => unsubscribe())
+  createComputed(() => {
+    batch(() => {
+      if (state.fetchStatus === 'idle') {
+        res()
+        res = noop
+      } else {
+        refetch()
+      }
+    })
```
ardeora, line 79:
```suggestion
    if (state.fetchStatus === 'idle') {
      res()
      res = noop
    } else {
      refetch()
    }
```

We don't need a batch here

packages/solid-query/src/createBaseQuery.ts
```diff
@@ -41,30 +41,22 @@ export function createBaseQuery<
...
-          resolve(unwrap(state.data))
-        }
-      })
-    },
-  )
-
-  batch(() => {
-    mutate(() => unwrap(state.data))
-    refetch()
+  let res = noop
```
ardeora, line 44: Can we give this variable a more descriptive name? Maybe `resolveQueryData` would be nice? 
```suggestion
  let resolveQueryData = noop
```

packages/solid-query/src/createBaseQuery.ts
```diff
@@ -41,30 +41,22 @@ export function createBaseQuery<
...
+  const [dataResource, { refetch }] = createResource<void>(() => {
+    return new Promise((resolve) => {
+      if (!(state.isFetching && state.isLoading)) {
+        resolve()
+      } else {
+        res = resolve
```
ardeora, line 50:
```suggestion
        resolveQueryData = resolve
```

packages/solid-query/src/createBaseQuery.ts
```diff
@@ -76,7 +68,21 @@ export function createBaseQuery<
...
+        res = noop
+      } else {
+        refetch()
+      }
+    })
+  })
+
+  onCleanup(() => {
+    unsubscribe()
+    res()
```
ardeora, line 84:
```suggestion
    resolveQueryData()
```

packages/solid-query/src/__tests__/transition.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+  it('should render the content when the transition is done', async () => {
+    const key = queryKey()
+
+    function Suspended() {
+      const state = createQuery(key, async () => {
+        await sleep(10)
+        return true
+      })
+
+      return <div>{state.data ? 'Message' : 'never'}</div>
```
ardeora, line 20: One thing that still breaks (even with the current implementation) is that if you use the `Show` component instead so 

```tsx
<Show when={state.data}>
    <div>Message</div>
</Show>
```

still doesn't work 😭  There's maybe something goofy over here with solid js not resolving the reactivity correctly. I'm trying to create a minimal reproduction of where things are going wrong, but it'll take me more time

ardeora (changes requested):

packages/solid-query/src/__tests__/transition.test.tsx
jeski-bright: That's true :( But I believe this must be fixed in `solid-js` core.

This however works:
```js
<div>
  <Show when={state.data}>
    <div>Message</div>
  </Show>
</div>
```

It seems that the conditional component always needs to be wrapped in another one and `Fragment` doesn't help here either.
For this one I'd expect something wrong in the transformations made by `X-plugin-solid` packages.

jeski-bright: This (along with the `Show`) issue seem to be fixed in `solid-js` already: https://github.com/solidjs/solid/commit/ea2c976add11b20798267c1fa218672189955ef6, though not released yet. Currently my fix breaks the `<Show when={...}` case 😂 , so how should we proceed with this?

I believe that the test case could still be useful.

ardeora: Ah yes! Thanks for creating an issue with Ryan about this. I built the latest version and it seems like I can get it working fine (current implementation) with the new fixes Ryan implemented. I will let solid-js push a new release out to solve this upstream issue. However, the transitions tests would be really useful. I would suggest only adding tests in this pull request for now :D Let me know if you have any questions or concerns!

ardeora: Our CI wont let us publish until all tests pass. So for now can we make it `it.skip`. I'll go back ad remove the skip when we have a new solid-js version out

jeski-bright: Left only the test & updated the commit message.

ardeora:
@TkDodo  We are looking good with this PR. @jeski-bright added a new test for solid-js transitions. The upstream issue should be fixed in a few days and we can unskip the test then. Feel free to merge 😄  

Or do you suggest we hold on to this PR until the upstream issue is fixed?

TkDodo:
@ardeora since this PR only adds a skipped test, I would wait until the upstream issue is fixed and then merge the test when we actually run it.

jeski-bright:
Since Ryan released those changes, updated solid to 1.5.7 and enabled the test. Please double check this, as during the first run of the test suite I got 1 failing test: `useQuery › should not fetch when switching to a disabled query`, subsequent re-runs were all green...

ardeora:
@TkDodo The changes are looking good here and `test:ci` works fine on my local machine. But CI on github actions fails. Any idea what's going on here? 😄 I can see it complaining about not finding `pnpm` on disk

TkDodo:
Yes I fixed this in my other PR. Sorry, I should port the fix directly to main. Will do later today.

ardeora:
No worries! Thanks for the quick reply. Thanks @jeski-bright! We'll get this merged when CI is fixed

TkDodo:
it's merged. please update with main to make the pipeline work :)

TkDodo (commented):
what is the empty file named `change` please? can we remove that?

jeski-bright:
> what is the empty file named `change` please? can we remove that?

Sorry for that, must've been checking something. Removed.

TkDodo (approved):

PR #4227, opened 2022-09-26
chore: add LICENSE to each package

KiwiKilian:
Why should the LICENSE be included in the package? Well [it's included in a distributed package per default](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#files).

But we actually have a use case for this. The MIT License requires the following:

> The above copyright notice and this permission notice shall be included in all
copies or substantial portions of the Software.

Therefore a package should include the licenses when packaged.

Furthermore we want to accomplish this in a native app by listing all packages and their licenses. To do so it's common to scan the `node_modules` for the `LICENSE` files. #4205 points to the same problem.


TkDodo:
looks good, we have one more package now: `query-persist-client-core`. Can you add it there as well?

KiwiKilian:
Done!

TkDodo:
we now have `pnpm` and it copies the license file to each released package, as described here:

https://pnpm.io/cli/publish

> When publishing a package inside a [workspace](https://pnpm.io/workspaces), the LICENSE file from the root of the workspace is packed with the package (unless the package has a license of its own).

So I don't think we need this PR anymore :)

KiwiKilian:
Yes that works perfectly fine, thanks!

PR #4228, opened 2022-09-26
refactor(types): rename query data generics

SammRosenthal:
This PR suggests aligning the generic names between `UseQueryOptions` and `getQueryData`! Now when working with `UseQueryOptions`, the names will help indicate the value returned from `getQueryData`.

This confusion comes from the fact that [`UseQueryOptions`](https://github.com/TanStack/query/blob/fc4671f0731d1dd757f9fd037fd8e6755b3393d1/src/react/types.ts#L26-L31) uses `TQueryFnData` and [`getQueryData`](https://github.com/TanStack/query/blob/fc4671f0731d1dd757f9fd037fd8e6755b3393d1/src/core/queryClient.ts#L105-L110) uses `TData`. (This is also the case for `getQueriesData`, `setQueryData`, and `setQueriesData`.

To help clear up any confusion we now have the matching generic name `TQueryFnData`.

This also fixes the `getQueriesData` return type by now returning `[queryKey: QueryKey, data: TQueryFnData | undefined]`. This aligns its return type with `getQueryData` where there can possibly be an `undefined` value in some cases explained by @TkDodo [here](https://github.com/TanStack/query/pull/3451/commits/258efb462d958421f876a279cfd481a4b32ced92#r846630179).

This PR also includes the changes to the documentation to reflect this update.


TkDodo (approved):

packages/query-core/src/queryClient.ts
```diff
@@ -105,32 +105,36 @@ export class QueryClient {
...
+  ): [QueryKey, TQueryFnData | undefined][]
+  getQueriesData<TQueryFnData = unknown>(
     queryKeyOrFilters: QueryKey | QueryFilters,
-  ): [QueryKey, TData][] {
+  ): [QueryKey, TQueryFnData | undefined][] {
     return this.getQueryCache()
       .findAll(queryKeyOrFilters)
       .map(({ queryKey, state }) => {
-        const data = state.data as TData
+        const data = state.data as TQueryFnData
```
TkDodo, line 127: the type assertion should be a union with undefined:

```suggestion
        const data = state.data as TQueryFnData | undefined
```

PR #4235, opened 2022-09-28
feat: make `data` defined when `suspense` is `true`

lukemorales:
This PR adds the overloads to `useQuery` that makes `data` be defined if `{ suspense: true }` is defined in the options

### DIFF
--- a/packages/react-query/src/useQuery.ts
+++ b/packages/react-query/src/useQuery.ts
@@ -17,6 +17,46 @@ export function useQuery<
 >(
   options: Omit<
     UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+    'suspense' | 'enabled' | 'initialData'
+  > & { suspense: true; enabled: boolean; initialData?: () => undefined },
+): UseQueryResult<TData, TError>
+
+export function useQuery<
+  TQueryFnData = unknown,
+  TError = unknown,
+  TData = TQueryFnData,
+  TQueryKey extends QueryKey = QueryKey,
+>(
+  options: Omit<
+    UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+    'suspense' | 'enabled' | 'initialData'
+  > & {
+    suspense: true
+    enabled: boolean
+    initialData?: TQueryFnData | (() => TQueryFnData)
+  },
+): DefinedUseQueryResult<TData, TError>
+
+export function useQuery<
+  TQueryFnData = unknown,
+  TError = unknown,
+  TData = TQueryFnData,
+  TQueryKey extends QueryKey = QueryKey,
+>(
+  options: Omit<
+    UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+    'suspense'
+  > & { suspense: true },
+): DefinedUseQueryResult<TData, TError>
+
+export function useQuery<
+  TQueryFnData = unknown,
+  TError = unknown,
+  TData = TQueryFnData,
+  TQueryKey extends QueryKey = QueryKey,
+>(
+  options: Omit<
+    UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
     'initialData'
   > & { initialData?: () => undefined },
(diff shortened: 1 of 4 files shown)


packages/react-query/src/useQuery.ts
```diff
@@ -81,6 +106,20 @@ export function useQuery<
...
 
+export function useQuery<
+  TQueryFnData = unknown,
+  TError = unknown,
+  TData = TQueryFnData,
+  TQueryKey extends QueryKey = QueryKey,
+>(
+  queryKey: TQueryKey,
+  queryFn: QueryFunction<TQueryFnData, TQueryKey>,
+  options: Omit<
```
lukemorales, line 117: `options` needs to be defined for this overload, otherwise, typescript will require `options` to be at least an empty object to fallback to the following overloads

packages/solid-query/src/createQuery.ts
```diff
@@ -86,6 +109,19 @@ export function createQuery<
...
 ): CreateQueryResult<TData, TError>
+export function createQuery<
+  TQueryFnData = unknown,
+  TError = unknown,
+  TData = TQueryFnData,
+  TQueryKey extends SolidQueryKey = SolidQueryKey,
+>(
+  queryKey: TQueryKey,
+  queryFn: QueryFunction<TQueryFnData, ReturnType<TQueryKey>>,
+  options: Omit<
```
lukemorales, line 120: `options` needs to be defined for this overload, otherwise, typescript will require `options` to be at least an empty object to fallback to the following overloads

TkDodo:
Thanks for the PR, but there are a couple of problems with relying on suspense to have data defined:

1) if you set `enabled: false`, you won't have data. I can see that you did some things with `enabled`, but it's always a `boolean`, so I'm not sure that this would work correctly?
2) if a query is cancelled via `queryClient.cancelQueries`, e.g. because of an optimistic update, while it is initially loading, it will be set back to its "previous state", which in this case would be the initialState, which doesn't have any data either. It's an edge case for sure but still a valid usage.

`suspense` (in comparison to initialData) is also something that is often set globally, and the type narrowing only works if we pass `suspense` directly to `useQuery`, so I don't really see the big gain.

I'm rather thinking about adding a `useSuspenseQuery` hook that will solve this holistically, but let's keep in mind that suspense is still very much experimental...

lukemorales:
> if you set enabled: false, you won't have data. I can see that you did some things with enabled, but it's always a boolean, so I'm not sure that this would work correctly?

I'm treating `enabled` as a dynamic prop in the overloads. Assuming they should switch from enabled to disabled as the conditions apply, data should be a union with undefined if that `enabled` is present. While technically someone just passing a literal `{ enabled: true }` is possible, I'd question if that's an edge case worth addressing in the overloads? If yes, a couple more variations should also account for those use cases.

> suspense (in comparison to initialData) is also something that is often set globally, and the type narrowing only works if we pass suspense directly to useQuery, so I don't really see the big gain.

while that makes perfect sense, I'd imagine how big is the adoption of Suspense on a global level vs per query? I might be completely wrong, but my impression is that Suspense is something that started being adopted not too long ago and I'd imagine most apps have been adopting it per query basis as the refactor allows it to happen.
About the second part of your sentence, I would personally pass `suspense: true` to a `useQuery` even if it was set globally just to have `data` be defined. But, as you very well pointed out, there are side effects that can happen that might not make sense for this to be true... perhaps on an app that implements API calls without the integration and query control that react query provides this might be true, but with the possibility of it being canceled in its loading state, the definition of `data` might not be accurate 🤔

TkDodo:
Instead of more overloads that add complexity to useQuery, why not have a better typed `useSuspenseQuery` that maybe doesn't support the enabled option at all?
It could still call useQuery under the hood and it would basically be just an opt-in convenience for TS users?

The cancellation problem would remain but I guess it's really an edge case...

On the other hand, you could really also implement that in user land with a simple non-null assertion in a custom hook. I'd actually prefer to wait with suspense changes until react has figured out everything around suspense for data fetching

PR #4236, opened 2022-09-28
refactor(solid-query): `vite-plugin-solid` backwards compatibility

ardeora:
Add backwards compatibility to the most popular `vite-plugin-solid` version (2.3.0). 
It also removes the version lock from all `solid-query` examples


ardeora:
Investigating issues with example. Closing this for now

PR #4237, opened 2022-09-28
Feature/pnpm

TkDodo:


sachinraja:
I tried `pnpm --filter "./packages/**" install` and it worked fine with this message as output:
```
➜  react-query git:(feature/pnpm) pnpm --filter "./packages/**" install
Scope: 9 of 32 workspace projects
Lockfile is up to date, resolution step is skipped
```

TkDodo:
@sachinraja you're right, it worked. Not sure sure what I did wrong before:
- 47a05dc14a0c7c663d17182ff574703f8fe10f8f

TkDodo:
hm but now I'm getting errors like `rollup: not found` because we have that defined as dependency in the top-level `package.json`. Is there a way to include the top-level dependencies, too?

p6l-richard:
> Is there a way to include the top-level dependencies, too? 

I got curious as well and for anyone reading this here, @juliusmarminge 
 added this [here](https://github.com/TanStack/query/pull/4239#discussion_r982750644)

PR #4248, opened 2022-09-30
docs: add more detail to the loading status in useQuery

tomskopek:
Builds on top of https://github.com/TanStack/query/pull/4244

I think some more details in the useQuery docs for status would be helpful to those who miss the migration guide.

### DIFF
--- a/docs/reference/useQuery.md
+++ b/docs/reference/useQuery.md
@@ -198,8 +198,8 @@ const result = useQuery({
 
 - `status: String`
   - Will be:
-    - `loading` if there's no cached data and no query attempt was finished yet.
-    - `error` if the query attempt resulted in an error. The corresponding `error` property has the error received from the attempted fetch
+    - `loading` if there's no cached data and no query attempt was finished yet. Queries (including disabled ones) will start in `loading` state.
+    - `error` if the query attempt resulted in an error. The corresponding `error` property has the error received from the attempted fetch.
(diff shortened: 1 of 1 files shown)


docs/reference/useQuery.md
```diff
@@ -198,8 +198,8 @@ const result = useQuery({
 
 - `status: String`
   - Will be:
-    - `loading` if there's no cached data and no query attempt was finished yet.
-    - `error` if the query attempt resulted in an error. The corresponding `error` property has the error received from the attempted fetch
+    - `loading` if there's no cached data and no query attempt was finished yet. Queries (including disabled ones) will start in `loading` state.
```
TkDodo, line 201: I am all for adding more context here, however, let's please keep in mind that a query can be disabled an in success state (if data was put into the cache via some other means), or even in error state. So saying that disabled queries (or any query really) will "start in loading state" is not correct, as it depends on many factors.

tomskopek: Oh wow - I didn't consider that.

While I think this would be a good place for some helpful explanation regarding the slightly unexpected loading status while the query is disabled, I can't think of text that is helpful and still correct, so I'll close this 😅 

Thanks for https://github.com/TanStack/query/pull/4244 and all your work, btw!

tomskopek: Yeah, that is probably sufficient 👍 😄

PR #4253, opened 2022-10-01
refactor(query-core): change QueryMeta and MutationMeta to an interface

DaleWebb:
Changes the `QueryMeta` and `MutationMeta` types to interfaces, so that they can have their shapes overridden in the implementing codebase for safer typing and autocompletion.

```ts
interface Meta {
  persist?: boolean;
}

  interface QueryMeta extends Meta {}
}

useQuery(queryKey, {
  meta: {
    persist: false,
  },
});
```


TkDodo (commented):
TypeScript refers to these as index types in the docs:

https://www.typescriptlang.org/docs/handbook/2/objects.html#index-signatures

this is just a stylistic change though

oleksandr-danylchenko:
Is the suggested syntax still applicable? 
Unfortunately, after declaring my custom `Meta` type the following way:
```ts
interface Meta {
	foo: boolean;
}

declare module '@tanstack/query-core' {
	// eslint-disable-next-line @typescript-eslint/no-empty-interface
	interface QueryMeta extends Meta {}
}
```
The typescript still says that the `foo` is not declared on the `QueryMeta` type.

PR #4254, opened 2022-10-01
feat: vue-query adapter

DamianOsipiuk:
Todo
- [x] test if everything still works
- [x] test UMD build
- [x] migrate basic example
- [x] migrate vue-specific docs


packages/vue-query/README.md
```diff
@@ -0,0 +1,81 @@
+[![Vue Query logo](./media/vue-query.png)](https://damianosipiuk.github.io/vue-query/)
+
+[![npm version](https://img.shields.io/npm/v/vue-query)](https://www.npmjs.com/package/vue-query)
+[![npm license](https://img.shields.io/npm/l/vue-query)](https://github.com/DamianOsipiuk/vue-query/blob/main/LICENSE)
+[![bundle size](https://img.shields.io/bundlephobia/minzip/vue-query)](https://bundlephobia.com/result?p=vue-query)
+[![npm](https://img.shields.io/npm/dm/vue-query)](https://www.npmjs.com/package/vue-query)
```
TkDodo, line 6: does this need to be adapted? it still references `https://damianosipiuk.github.io/vue-query/` and `vue-query` on bundlephobia and `npm`

packages/vue-query/README.md
```diff
@@ -0,0 +1,81 @@
...
+- [![npm bundle size](https://img.shields.io/bundlephobia/minzip/vue-query)](https://bundlephobia.com/result?p=vue-query) (depending on features imported)
+
+# Quick Start
+
+1. Install `vue-query`
+
+   ```bash
+   npm install vue-query
+   # or
+   yarn add vue-query
```
TkDodo, line 40: same here (and below), this likely needs to be adapted

packages/vue-query/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "@tanstack/vue-query",
+  "version": "4.7.2",
```
TkDodo, line 3:
```suggestion
  "version": "4.8.0",
```

packages/vue-query/package.json
```diff
@@ -0,0 +1,55 @@
...
+  ],
+  "devDependencies": {
+    "@vue/composition-api": "1.7.1",
+    "vue": "^3.2.40",
+    "vue2": "npm:vue@2"
+  },
+  "dependencies": {
+    "@tanstack/query-core": "workspace:*",
+    "@vue/devtools-api": "^6.4.2",
+    "match-sorter": "^6.3.1",
```
TkDodo, line 43: we've had problems with `match-sorter`, so we switched to the one Tanner has exposed from the react-table:

https://github.com/TanStack/query/blob/9f48dffaf506a4730162a7490f00394c0c39f7c6/packages/react-query-devtools/package.json#L50

Maybe it's worth also using this here?

packages/vue-query/src/index.ts
```diff
@@ -0,0 +1,26 @@
+/* istanbul ignore file */
+
+export * from '@tanstack/query-core'
+
+export { useQueryClient } from './useQueryClient'
+export { VueQueryPlugin } from './vueQueryPlugin'
+
+export { QueryClient } from './queryClient'
+export { QueryCache } from './queryCache'
+export { MutationCache } from './mutationCache'
```
TkDodo, line 10: do we need to re-export the core if you have your own wrappers? I mean the `query-core` also has a `QueryClient`, so what happens if we export this twice?

packages/vue-query/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "composite": true,
+    "rootDir": "./src",
+    "outDir": "./build/lib",
+    "tsBuildInfoFile": "./build/.tsbuildinfo"
+  },
+  "include": ["src"],
+  "exclude": ["src/__tests__"],
```
TkDodo, line 10: we have removed this lately because it causes tons of errors in local IDEs. Can we remove it here as well?

```suggestion
  },
  "include": ["src"],
```

pnpm-workspace.yaml
```diff
@@ -2,3 +2,4 @@ packages:
   - 'packages/**'
   - 'examples/react/**'
   - 'examples/solid/**'
+  # - 'examples/vue/**'
```
TkDodo, line 5: why is this commented out? I think we'd need it so that pnpm installs dependencies for all examples?

TkDodo (commented):
looks really good 🚀

packages/vue-query/package.json
DamianOsipiuk: I will take a look at that afterwards

packages/vue-query/src/index.ts
DamianOsipiuk: The latter will overwrite the former. It works well.

packages/vue-query/tsconfig.json
DamianOsipiuk: Yeah, since i was using `tsc` to build vue-query before it was necessary. But now it's not 👍

pnpm-workspace.yaml
DamianOsipiuk: Right now it does not work, cause `@tanstack/vue-query` does not exist and pnpm fails on install

TkDodo: oh shoot yeah that makes sense. ok then let's please create a follow-up PR after we release vue-query that enables this :)

DamianOsipiuk: I wonder though if we could do the same trick as with referencing core - `"@tanstack/query-core": "workspace:*",` in examples? 🤔

TkDodo: I thought about this, but I don't think so because examples need to work in isolation in codesandbox...

TkDodo:
@allcontributors please add @DamianOsipiuk for code, maintenance

PR #4258, opened 2022-10-02
fix: use match-sorter-utils, simplify sorting

DamianOsipiuk:
While looking at the `match-sorter-utils` i noticed that `react-query-devtools` makes unnecessary passes trough the query array. 
Also for a big amount of entries in the cache, filtering first would be beneficial in regards of performance.

What i have noticed is that we only need `filtering` from the `match-sorter-utils`, since we have predefined list of sort functions.
If we implement `hash sort function` as `a.queryHash.localeCompare(b.queryHash)`, we would have natural sorting done (`[1, 10]` instead of `[10, 1]`). So the second sort with give us virtually no benefit.


packages/react-query-devtools/src/devtools.tsx
```diff
@@ -418,9 +418,9 @@ export const ReactQueryDevtoolsPanel = React.forwardRef<
 
   const [filter, setFilter] = useLocalStorage('reactQueryDevtoolsFilter', '')
 
-  const [sortDesc, setSortDesc] = useLocalStorage(
+  const [baseSort, setBaseSort] = useLocalStorage(
     'reactQueryDevtoolsSortDesc',
```
TkDodo, line 422: should we chose a new name here? Otherwise, stored values might be picked up:

```suggestion
    'reactQueryDevtoolsBaseSort',
```

packages/react-query-devtools/src/devtools.tsx
```diff
@@ -438,26 +438,23 @@ export const ReactQueryDevtoolsPanel = React.forwardRef<
...
-    ranked = ranked.filter((d) => d[1].passed)
+    const filtered = filter
+      ? unsortedQueries.filter(
+          (item) => rankItem(item.queryHash, filter).passed,
+        )
+      : [...unsortedQueries]
 
-    ranked = ranked.sort((a, b) => compareItems(a[1], b[1]))
+    const sorted = sortFn
+      ? filtered.sort((a, b) => sortFn(a, b) * (baseSort as number))
```
TkDodo, line 453: all of our usages of `useLocalStorage` provide the optional default value, so if we make it mandatory, we don't have to do type assertions here. Also, with the current implementation, there is a render cycle where the value is indeed `undefined`. We can fix that by passing the `defaultValue` directly to `useState` inside `useLocalStorage` and can then also get rid of this whole block:

https://github.com/TanStack/query/blob/357ec041a6fcc4a550f3df02c12ecc7bcdefbc05/packages/react-query-devtools/src/useLocalStorage.ts#L24-L27

TkDodo (commented):
nice work 🙌

packages/react-query-devtools/src/devtools.tsx
DamianOsipiuk: @TkDodo Not sure why but when i touch anything about `useLocalStorage` to get rid of initial `undefined` render for values, tests start to break.
Basically this code block stops devtools from being rendered:
```ts
// Do not render on the server
if (!isMounted()) return null
```
🤷

TkDodo: ok, 🤷

TkDodo (approved):

PR #4260, opened 2022-10-03
Add questions

nhducit:
1. create a single ton queryClient and reuse it
    
    we can create an angular service
    
2. consumer notify query-core
    1. `queryArray`, `options` changes
        - query-core update query options
        
        `queryArray` and `options` are a behavior subject
        
        should we use the behavior subject to store options? is it ok to store both primitive data and callback in the behavior subject?
        
    2. onDestroy
        - query-core unsubscribe store’s cache
        
        notify the query-core observer to call observer.unsubscribe()
        
3. query-core notify the  consumer
    1. result is updated 
        1. the consumer uses the new data
            
            should we wrap the result inside a behavior subject?
            
        2. consumer/query-core trigger angular changeDetections
            
            if we use behavior subject should cover this already

instead of returning raw data, we return an observable and pipe that observable with rxjs's finalized operator to call observer.unsubscribe()

### DIFF
--- a/packages/react-query/src/useBaseQuery.ts
+++ b/packages/react-query/src/useBaseQuery.ts
@@ -29,7 +29,7 @@ export function useBaseQuery<
   >,
   Observer: typeof QueryObserver,
 ) {
-  // Duc - we need to call create a queryClient and reuse it
+  // Duc - we need to create a queryClient and reuse it per query
 
   // { context: options.context }
   // Duc - why do we need context here?
@@ -115,6 +115,10 @@ export function useBaseQuery<
   )
 
   // Duc - update observer option if consumer options are changed
+  // Duc - need to find a good API to for angular life cycle
+  // Duc - a good canidate is provide options as BehaviorSubject
+  // Duc - consumer will subject this subject whenever options are update
+  // Duc - angular bindings should listen for changes and update observer options
   React.useEffect(() => {
     // Do not notify on updates because of changes in the options because
     // these changes should already be reflected in the optimistic result.
--- a/packages/react-query/src/useBaseQuery.ts
+++ b/packages/react-query/src/useBaseQuery.ts
@@ -29,6 +29,9 @@ export function useBaseQuery<
   >,
   Observer: typeof QueryObserver,
 ) {
+  // Duc - we need to call create a queryClient and reuse it
+
+  // { context: options.context }
   // Duc - why do we need context here?
   // Duc - how to implement this behavior in angular
   // Duc - FROM the document: Required, but only if no default query function has been defined See Default Query Function for more information.
@@ -38,7 +41,7 @@ export function useBaseQuery<
   // Duc - skip this in MVP
   const isRestoring = useIsRestoring()
   // Duc - SKIP: since angular does not have ErrorBoundary
-  const errorResetBoundary = useQueryErrorResetBoundary()
+  // const errorResetBoundary = useQueryErrorResetBoundary()
   const defaultedOptions = queryClient.defaultQueryOptions(options)
 
   // Make sure results are optimistically set in fetching state before subscribing or updating options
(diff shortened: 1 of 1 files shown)


packages/react-query/src/useBaseQuery.ts
```diff
@@ -29,29 +29,43 @@ export function useBaseQuery<
...
   Observer: typeof QueryObserver,
 ) {
+  // Duc - we need to create a queryClient and reuse it per query
+
+  // { context: options.context }
+  // Duc - why do we need context here?
+  // Duc - how to implement this behavior in angular
+  // Duc - FROM the document: Required, but only if no default query function has been defined See Default Query Function for more information.
+  // https://tanstack.com/query/v4/docs/reference/useQuery?from=reactQueryV3&original=https://react-query-v3.tanstack.com/reference/useQuery
+  // maybe we can skip this for angular e.g. the default query function is required
```
TkDodo, line 39: context is needed to support multiple contexts :) see:

https://tanstack.com/query/v4/docs/guides/migrating-to-react-query-4#custom-contexts-for-multiple-providers

I don't know if angular has a similar concept. But it's not part of the query-core, so you can totally omit it.

packages/react-query/src/useBaseQuery.ts
```diff
@@ -29,29 +29,43 @@ export function useBaseQuery<
...
   const queryClient = useQueryClient({ context: options.context })
+  // Duc - skip this in MVP
   const isRestoring = useIsRestoring()
```
TkDodo, line 42: the check for restoring is needed to make sure that we are not triggering queries while an async persister is potentially still restoring data from an external location into the query cache.

You'll need something like this if the persist-query-client plugins are supported.

packages/react-query/src/useBaseQuery.ts
```diff
@@ -78,57 +96,70 @@ export function useBaseQuery<
       ),
   )
 
+  // Duc - skip this in MVP
   const result = observer.getOptimisticResult(defaultedOptions)
```
TkDodo, line 100: this is what returns the result of the query so it cannot be skipped :)

packages/react-query/src/useBaseQuery.ts
```diff
@@ -78,57 +96,70 @@ export function useBaseQuery<
...
 
+  // Duc - why do we need useSyncExternalStore
+  // need to check React document to find out
   useSyncExternalStore(
```
TkDodo, line 104: this makes sure our observers are subscribed at the right time. it's react specific.

packages/react-query/src/useBaseQuery.ts
```diff
@@ -78,57 +96,70 @@ export function useBaseQuery<
...
+  // Duc - angular bindings should listen for changes and update observer options
   React.useEffect(() => {
     // Do not notify on updates because of changes in the options because
     // these changes should already be reflected in the optimistic result.
     observer.setOptions(defaultedOptions, { listeners: false })
   }, [defaultedOptions, observer])
 
+  // Duc - is this react specific suspense logic?
+  // Duc - I guess we don't need this for Angular
   // Handle suspense
```
TkDodo, line 130: yes, it's suspense specific.

TkDodo (commented):
well most of the code in this file is react specific. If I were you I would rather look at the query-core. As you've seen, all we basically do is create a `QueryObserver` at the "correct" time (depends on the lifecycle of your framework) and make sure the options are forwarded. The result returned from useQuery is the optimistic result from the observer - so that we also have a result on the first render, before the actual fetch has happened. The observer receives an update in the callback of the subscription and then you need to re-render your component (or somehow else update the dom with the new information).

The rest is suspense / errorBoundary related.

I think the tests for the QueryObserver in the query-core would be a good starting point:

https://github.com/TanStack/query/blob/ca767771f3720ba9abb541871c22ea106172e9ab/packages/query-core/src/tests/queryObserver.test.tsx

nhducit:
thanks @TkDodo for helpful comments 🙏

PR #4263, opened 2022-10-04
fix: add proper rollup config for superjson

DamianOsipiuk:


TkDodo (approved):
hm, I didn't know that we have to list all of these manually. maybe in the future, we can extract those dependencies from package.json ....

DamianOsipiuk:
I think there are rollup plugins to auto-external deps from node_modules. But it was kind of finicky when i last tested it out.
Might give it another try to simplify this though.

PR #4288, opened 2022-10-09
chore: improve CI

TkDodo:
this PR:

- tries to fix most peer dependency warnings
- splits GitHub Actions into multiple steps so that they run in parallel
- moves test execution from one big jest run into the separate packages


packages/solid-query/src/__tests__/createQuery.test.tsx
```diff
@@ -5191,7 +5149,7 @@ describe('createQuery', () => {
           isFetching,
           isSuccess,
           isStale,
-        })
+        } as any)
```
TkDodo, line 5152: @lukesmurray @ardeora any idea why I had to make these changes here? What is the type `Optional` - I got errors that this is not found.

Also, the types we're trying to push into this Array here don't seem to be matching :/

packages/solid-query/jest-preset.js
```diff
@@ -0,0 +1,7 @@
+const solidPreset = require('solid-jest/preset/browser/jest-preset')
+const tanStackPreset = require('../../jest-preset')
+
+module.exports = {
+  ...tanStackPreset,
+  ...solidPreset,
+}
```
TkDodo, line 7: @ardeora @lukesmurray I'm having problems getting the solid tests to work in this PR. They are the only ones that use their own preset. I've merged the solid preset with our own preset, and I think that works.

I've had general troubles with `babel-jest` - it just gave me lots of "unexpected token" errors, as if it didn't transpile jsx / typescript at all. So I switched to ts-jest and things just work. But the solid preset also uses `babel-jest` under the hood as far as I can see, so the errors are still there for the solid package :/

Would appreciate if you could take a look here 🙏

.github/workflows/pr.yml
```diff
@@ -2,33 +2,98 @@ name: pr
...
-        if: matrix.react == '17'
-        env:
-          REACTJS_VERSION: ${{ matrix.react }}
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v2
+  lint:
+    name: 'Lint'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
karlhorky, line 25: This can use v3 now, right?

packages/solid-query/jest-preset.js
TkDodo: I don't really know what I'm doing, but adding a custom transformer (copied from the preset one + added more babel presets) made it work:

21fa9392

.github/workflows/pr.yml
TkDodo: probably, yes. what's the difference? are there breaking changes?

karlhorky: Node 16 by default and a minimum actions runner version: https://github.com/actions/checkout/releases/tag/v3.0.0

packages/solid-query/jest-preset.js
ardeora: Hello! Apologies for the late reply. I have been travelling internationally this weekend. So I will be AFK till Monday 😅 From a quick look. It seems like it breaks whenever it encounters type definitions/ types in the test file. Maybe @lukesmurray can take a look? If not I'll take a look first thing Monday

.github/workflows/pr.yml
TkDodo: done: [42f3c3d](https://github.com/TanStack/query/pull/4288/commits/42f3c3db718340f86f58a45a0f83523b6b7e031a)

packages/solid-query/src/__tests__/createQuery.test.tsx
ardeora: Hello! This is a bug that is caused by an issue upstream. I am writing a small PR to fix all of these any/Partial typings.

lukesmurray: thank you @ardeora

packages/solid-query/jest-preset.js
lukesmurray: let me know if you still need help with this. I'm not super familiar with solid's babel + jest setup but I can do my best to help if you're still running into issues.

TkDodo: it works fine now, so this is resolved for me 🤷

packages/solid-query/src/__tests__/createQuery.test.tsx
ardeora: No worries! #4308 aims to solve this and removes all the vague typings too

.github/workflows/pr.yml
```diff
@@ -2,33 +2,98 @@ name: pr
...
+    name: 'Test React 17'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: pnpm/action-setup@v2.2.2
+        with:
+          version: 7
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node }}
```
DamianOsipiuk, line 76: I guess this matrix could be swapped to specific version, like in other jobs?

DamianOsipiuk (approved):

.github/workflows/pr.yml
TkDodo: well spotted - this was a left-over: dc8819cc

PR #4294, opened 2022-10-09
Specify end of line type

hichemfantar:
Specify end of line type for consistency across operating systems

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,6 @@
 {
   "semi": false,
   "singleQuote": true,
-  "trailingComma": "all"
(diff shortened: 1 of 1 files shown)


.prettierrc
```diff
@@ -1,5 +1,6 @@
 {
   "semi": false,
   "singleQuote": true,
-  "trailingComma": "all"
+  "trailingComma": "all",
+  "endOfLine": "lf"
```
TkDodo, line 5: endOfLine lf is the prettier default since v2

https://prettier.io/docs/en/options.html#end-of-line

PR #4295, opened 2022-10-09
feat(query-core): add 'fetchedPages' option

aliakbarazizi:
Related discussions, https://github.com/TanStack/query/discussions/4286.

Also i'm not sure the way I approach nist, I added a new state called meta, since fetchMeta will be overwrite in every fetch and there is no way to access to old value in behavior.


packages/query-core/src/infiniteQueryBehavior.ts
```diff
@@ -134,6 +134,11 @@ export function infiniteQueryBehavior<
           // Fetch remaining pages
           for (let i = 1; i < oldPages.length; i++) {
             promise = promise.then((pages) => {
+              context.dispatch({
```
aliakbarazizi, line 137: `i` is the value for previous page also in every promise I changed fetchPages for previous one

packages/query-core/src/query.ts
```diff
@@ -130,6 +137,7 @@ export type Action<TData, TError> =
   | PauseAction
   | SetStateAction<TData, TError>
   | SuccessAction<TData>
+  | MetaAction
```
aliakbarazizi, line 140: `SetStateAction` has a weird property called `setStateOptions` I don't see anywhere it used, maybe its better to merge this action with `SetState` and `setStateOptions`

packages/query-core/src/query.ts
```diff
@@ -49,6 +49,7 @@ export interface QueryState<TData = unknown, TError = unknown> {
   isInvalidated: boolean
   status: QueryStatus
   fetchStatus: FetchStatus
+  meta?: any
```
aliakbarazizi, line 52: I marked meta as undefined cause its usually is undefined.

TkDodo: `meta` is not a good name because we already have `meta` inside the `Query` itself:

https://github.com/TanStack/query/blob/04945f38b3f2ff82a597083bf1a36724f973bc7e/packages/query-core/src/query.ts#L152

could we just put this value inside `fetchMeta` ?

packages/query-core/src/query.ts
TkDodo: yeah `setStateOptions` seems to be unsed! We could remove that I think :)

packages/query-core/src/query.ts
```diff
@@ -560,6 +569,11 @@ export class Query<
             ...state,
             isInvalidated: true,
           }
+        case 'meta':
+          return {
+            ...state,
+            meta: action.meta,
+          }
```
TkDodo, line 576: I think this is too generic. How about an action that says `'pageFetched'` or something that increases the current state? and we'd re-set the counter to zero when a `'fetch'` action is initiated

TkDodo (commented):
I think it looks pretty good already. I left some comments, and we'd also need some tests for the new behaviour please

TkDodo:
hmm it seems that this breaks lots of existing tests 🤔

packages/query-core/src/query.ts
aliakbarazizi: It's not that easy to put it on fetchMeta, because when fetchingNextPage called there is no good way to access previous  fetchMeta and this method will overwrite fetchMeta. So I need to change before this line
https://github.com/TanStack/query/blob/89da08fe0723249a86b681e855937787b3d14eb9/packages/query-core/src/infiniteQueryBehavior.ts#L16

and need lot of check to check `fetchOptions` and `fetchOptions.meta` is not undefined and add new meta on it.

packages/query-core/src/query.ts
aliakbarazizi: How about putting meta in setStateOptions action and dispatch `setState`? I think its better to remove it.

aliakbarazizi:
I didn't test I just want share my idea, because I'm new to this library and I wasn't sure my way is a good way, I will check test and update test

packages/query-core/src/query.ts
aliakbarazizi: That actually sound very good I will do that

aliakbarazizi:
The tests are broken because we check length of state and I need to change it all
for example this code
`expect(states.length).toBe(2)`

aliakbarazizi:
I did check some of the test
there is two main issue
one of them checking how many re-render call and it expect `2` but because of fetchedPages is `3` now I think I just need change 2 to 3 since as you say we tracked property for re-render and it should be `2` without getting fetchedPages, I will add proper test for it but i just want to be sure am i correct?

the another one is about size of states and I need to change it in all test becasue the size of state is increased by 1

TkDodo:
> one of them checking how many re-render call and it expect 2 but because of fetchedPages is 3 now I think I just need change 2 to 3 since as you say we tracked property for re-render and it should be 2 without getting fetchedPages, I will add proper test for it but i just want to be sure am i correct?

the way I see it, with tracked properties being on per default, there shouldn't be additional re-renders of the component even if we dispatch more events during fetching. But it seems we get one additional re-render per each update...

aliakbarazizi:
In this test It's expected, in this test we fetch 3 pages and after all of that we refetch all pages again, so we fetch 6 pages, and we have one more re-render pages.

aliakbarazizi:
Actually you are right, in the test we don't call `fetchedPages` I will look in to it whats going on

TkDodo:
> In this test It's expected, in this test we fetch 3 pages and after all of that we refetch all pages again, so we fetch 6 pages, and we have one more re-render pages.

not sure I understand that. Unless the component is using the newly exposed counter, there shouldn't be any additional re-render ...

aliakbarazizi:
> > one of them checking how many re-render call and it expect 2 but because of fetchedPages is 3 now I think I just need change 2 to 3 since as you say we tracked property for re-render and it should be 2 without getting fetchedPages, I will add proper test for it but i just want to be sure am i correct?
> 
> the way I see it, with tracked properties being on per default, there shouldn't be additional re-renders of the component even if we dispatch more events during fetching. But it seems we get one additional re-render per each update...
> 
> there is even one tests that has:
> 
> ```
> Expected: 8
> Received: 14
> ```

I checked the example you provided

```
Expected: 8
Received: 14
```

I think thats because when using this like it broke tacked property feature
https://github.com/TanStack/query/blob/main/packages/solid-query/src/__tests__/createInfiniteQuery.test.tsx#L445

And I think I have to change `await waitFor(() => expect(states.length).toBe(8))` to `await waitFor(() => expect(states.length).toBe(14))` in these kind of situations.

Can you check and see am I right?

TkDodo:
yeah the one that has `notifyOnChangeProps: 'all'` will need to be updated, that's for sure.

I have to say I don't know about the solid tests. I hope that @ardeora can have a look?

packages/query-core/src/infiniteQueryBehavior.ts
```diff
@@ -134,6 +134,10 @@ export function infiniteQueryBehavior<
...
+                type: 'pageFetched',
+              })
```
TkDodo, line 139: shouldn't we dispatch this event at the end of the `fetchPage` function? That way, it would be in one place, right after really fetching a page ...

packages/query-core/src/query.ts
TkDodo: yeah it would need shallow merging of the meta. My biggest concern is with the word `meta`, as we already have three different "meta" properties:

- the `fetchMeta`
- meta on queryOptions
- meta on mutationOptions

if we can't unify this with fetchMeta, we have to please find a better name 😅

TkDodo:
there's one more react-query tests that needs changes:

```
FAIL packages/react-query/src/__tests__/useInfiniteQuery.test.tsx
  ● useInfiniteQuery › should return the correct states for a successful query

    expect(received).toEqual(expected) // deep equality

    - Expected  - 0
    + Received  + 1

    @@ -13,10 +13,11 @@
        "errorUpdatedAt": 0,
        "failureCount": 0,
        "fetchNextPage": Any<Function>,
        "fetchPreviousPage": Any<Function>,
        "fetchStatus": "idle",
    +   "fetchedPages": 1,
        "hasNextPage": true,
        "hasPreviousPage": undefined,
        "isError": false,
        "isFetched": true,
        "isFetchedAfterMount": true,
```

and there are some more failing tests that have `notifyOnChangeProps: 'all'`, which would be fine to adapt

aliakbarazizi:
> there's one more react-query tests that needs changes:
> 
> ```
> FAIL packages/react-query/src/__tests__/useInfiniteQuery.test.tsx
>   ● useInfiniteQuery › should return the correct states for a successful query
> 
>     expect(received).toEqual(expected) // deep equality
> 
>     - Expected  - 0
>     + Received  + 1
> 
>     @@ -13,10 +13,11 @@
>         "errorUpdatedAt": 0,
>         "failureCount": 0,
>         "fetchNextPage": Any<Function>,
>         "fetchPreviousPage": Any<Function>,
>         "fetchStatus": "idle",
>     +   "fetchedPages": 1,
>         "hasNextPage": true,
>         "hasPreviousPage": undefined,
>         "isError": false,
>         "isFetched": true,
>         "isFetchedAfterMount": true,
> ```
> 
> and there are some more failing tests that have `notifyOnChangeProps: 'all'`, which would be fine to adapt

I did change all of them I just need confirmation about notifyOnChangeProps, I'll push it as soon as possible.

aliakbarazizi:
I did merge `meta` with `fetchMeta`, I also fixed tests and one test specifically for `fetchedPages`.

I have weird problem in local with `userQuery` tests.

I pushed my code to see the github testings result too, I think my problem is local but still I need to see.

packages/query-core/src/query.ts
```diff
@@ -516,7 +515,12 @@ export class Query<
           return {
             ...state,
             fetchFailureCount: 0,
-            fetchMeta: action.meta ?? null,
+            fetchMeta: {
+              ...action.meta,
+              fetchedPages: action.meta?.fetchMore
+                ? state.fetchMeta?.fetchedPages
+                : 0,
```
TkDodo, line 522: I think when a `fetch` action is triggered, we should always set `fetchPages` back to `0`. I see no case where we actually pass `fetchedPages` to the `fetch` action?

packages/query-core/src/query.ts
```diff
@@ -560,6 +564,14 @@ export class Query<
...
             isInvalidated: true,
           }
+        case 'pageFetched':
+          return {
+            ...state,
+            fetchMeta: {
+              ...state.fetchMeta,
+              fetchedPages: state.fetchMeta.fetchedPages + 1,
+            },
+          }
```
TkDodo, line 574: I'm a little concerned about the triple equals comparison of `state.fetchMeta` here:

https://github.com/TanStack/query/blob/04945f38b3f2ff82a597083bf1a36724f973bc7e/packages/query-core/src/query.ts#L419

because every `pageFetched` action will create a "new" `fetchMeta`. Maybe it would be better, for the internal state, to keep it on a separate state variable ?

ardeora:
I took a look at this PR. It seems like most (all?) solid tests fail because `solid-query` resolves `queryObserver` updates a little differently than `react-query`. Take a look at this createInfiniteQuery function:

```js
const query = createInfiniteQuery(
  () => ['pages'],
  ({ pageParam = 20 }): number => pageParam,
  {
    initialData: { pages: [10], pageParams: [undefined] },
    getNextPageParam: () => undefined,
  },
)
```

When the query is mounted. These are the results queryObserver initialises and updates with consecutively -

```js
// Initial Result
{ data:  {pages: [10], pageParams: [undefined] },  fetchStatus: 'fetching', fetchedPages: 0 }
// Update
{ data:  {pages: [10], pageParams: [undefined] },  fetchStatus: 'fetching', fetchedPages: 1 }
// Update
{ data:  {pages: [20], pageParams: [undefined] },  fetchStatus: 'idle', fetchedPages: 1 }
```

Now `react-query` batches the 'Updates'
```js
// Update
{ data:  {pages: [10], pageParams: [undefined] },  fetchStatus: 'fetching', fetchedPages: 1 }
// Update
{ data:  {pages: [20], pageParams: [undefined] },  fetchStatus: 'idle', fetchedPages: 1 }
```

 by using a using a setTimeout to defer the commit after all macrotasks have been executed. We can see the definition for it here https://github.com/TanStack/query/blob/main/packages/query-core/src/utils.ts#L413-L415. 

`solid-query` however batches these updates queuing a microtask 
https://github.com/TanStack/query/blob/main/packages/solid-query/src/createBaseQuery.ts#L81-L87
which performs updates between macrotasks. This is the reason why we see 3 query state updates instead of 2 in solid-query. @TkDodo any idea why the `fetchedPages` property is updated differently than other state updates?

aliakbarazizi:
> I took a look at this PR. It seems like most (all?) solid tests fail because `solid-query` resolves `queryObserver` updates a little differently than `react-query`. Take a look at this createInfiniteQuery function:
> 
> ```js
> const query = createInfiniteQuery(
>   () => ['pages'],
>   ({ pageParam = 20 }): number => pageParam,
>   {
>     initialData: { pages: [10], pageParams: [undefined] },
>     getNextPageParam: () => undefined,
>   },
> )
> ```
> 
> When the query is mounted. These are the results queryObserver initialises and updates with consecutively -
> 
> ```js
> // Initial Result
> { data:  {pages: [10], pageParams: [undefined] },  fetchStatus: 'fetching', fetchedPages: 0 }
> // Update
> { data:  {pages: [10], pageParams: [undefined] },  fetchStatus: 'fetching', fetchedPages: 1 }
> // Update
> { data:  {pages: [20], pageParams: [undefined] },  fetchStatus: 'idle', fetchedPages: 1 }
> ```
> 
> Now `react-query` batches the 'Updates'
> 
> ```js
> // Update
> { data:  {pages: [10], pageParams: [undefined] },  fetchStatus: 'fetching', fetchedPages: 1 }
> // Update
> { data:  {pages: [20], pageParams: [undefined] },  fetchStatus: 'idle', fetchedPages: 1 }
> ```
> 
> by using a using a setTimeout to defer the commit after all macrotasks have been executed. We can see the definition for it here https://github.com/TanStack/query/blob/main/packages/query-core/src/utils.ts#L413-L415.
> 
> `solid-query` however batches these updates queuing a microtask https://github.com/TanStack/query/blob/main/packages/solid-query/src/createBaseQuery.ts#L81-L87 which performs updates between macrotasks. This is the reason why we see 3 query state updates instead of 2 in solid-query. @TkDodo any idea why the `fetchedPages` property is updated differently than other state updates?

`fetchedPages` change on every page fetched. this action will dispatch in `Promise` that suppose to run `fetching` and all of other state is changing outside of this `Promise` so that's why, but I don't see anyway to do that without this behavior

aliakbarazizi:
@ardeora should `tracked` work fine in solid? if it's working then we shouldn't have any state changes because there is no subscribe to `fetchedPages` in those tests.

TkDodo:
@ardeora I think you might be looking in the wrong direction. The `scheduleMicrotask` is done in the `query-core`, which `solid-query` also uses. The `setTimeout` is admittedly a bit hacky - ideally, we would just use the browser-native [queueMicrotask](https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask), but a lot of tests failed (more in-between re-renders) so I didn't do that.

what react-query additionally does is that we are using our `notifiyManager.setBatchNotifyFunction` and set it to `unstable_batchedUpdates` from `react-dom` (or `react-native` respectively):

https://github.com/TanStack/query/blob/357ec041a6fcc4a550f3df02c12ecc7bcdefbc05/packages/react-query/src/setBatchUpdatesFn.ts#L1-L4

We use this function in the query-core when invoking callbacks, but also in `useBaseQuery` to batch updates coming from react:

https://github.com/TanStack/query/blob/34ca5581217dae4133046f6cfbc746a22be14dd9/packages/react-query/src/useBaseQuery.ts#L88

If the solid way to do this is via the `queueMicrotask` you've pointed towards, I would still suggest to call `notifyManager.setBatchNotifyFunction` somewhere to this custom function, and then use `notifyManager.batchCalls` when you create the observer subscription.

> @ardeora shouldn't tracked works fine in solid? if it's working then we shouldn't have any state changes because there is no subscribe to fetchedPages in those tests.

This is a good question. property tracking is part of the query-core 🤔

aliakbarazizi:
@ardeora @TkDodo any updates?

TkDodo:
@aliakbarazizi there are still some open questions from my side:
https://github.com/TanStack/query/pull/4295#pullrequestreview-1138764366

aliakbarazizi:
@TkDodo I did answered all of them before. I don't know my answer is enough or need an action to fix them

ardeora:
> @ardeora @TkDodo any updates?

Yeah I have spent a few hours today trying to get `notifyManager` working with `solid-query` but it seems to break even more tests than without it. I think I might try something different tomorrow to switch to `notifyManager.setBatchNotifyFunction` which might help with this PR too

ardeora:
Great question! Solid components are invoked once throughout their lifecycle. So the `store` that you see is referentially stable between state updates. If I just `states.push(store)` it will just push the same store repeatedly on state updates. 

Also notifyOnChangeProps is irrelevant in solid query since the `store` itself updates states in a fine grained fashion

TkDodo:
I'm not sure how to continue with this PR. In the original discussion, we talked about how displaying a loading indicator for refetches of all pages (e.g. all 20 in the cache) would be too long. In v5, we've introduced the `maxPages` option, which can limit how many pages are kept in the cache. Maybe this is the better fix for the underlying problem? If you limit the number of pages in the cache, you can continue to display the loading spinner until all pages are refetched, because you can control what "all pages" means. Also, you can limit the number of refetches happening this way.

aliakbarazizi:
> I'm not sure how to continue with this PR. In the original discussion, we talked about how displaying a loading indicator for refetches of all pages (e.g. all 20 in the cache) would be too long. In v5, we've introduced the `maxPages` option, which can limit how many pages are kept in the cache. Maybe this is the better fix for the underlying problem? If you limit the number of pages in the cache, you can continue to display the loading spinner until all pages are refetched, because you can control what "all pages" means. Also, you can limit the number of refetches happening this way.

Sorry for late respond, the issue was when you want to show 2 different loading ( loading for first page refetching, and loading for other page ) 
The purpose of this pull request is not limit number of refetch pages, its provide a way to know which page is refetching currently.

I don't think `maxPages` is good alternative, because its only limit number of page to refetch, but what if you want to refetch all pages and still show different loading for first page refetching?


Also it may be better to tell why sometime you need 2 different loading, if you create pull down to refetch ability (something like twitter in webapp) you need a loading when fetching new data (not syncing old data) at the top of screen, but when user scrolling to see more pages you need a loading at bottom of screen to show we are fetching new pages

aliakbarazizi:
Also its worse to mention this pull request was completely ready, the only issue remain is solid test case and as I mention above it seems the issue is how solid test are, and the issue is not because of my codes

PR #4299, opened 2022-10-10
Memoizing the return values 🛥️ 

sanjaiyan-dev:
Memoizing the return values using [`React.useMemo`](https://beta.reactjs.org/apis/react/useMemo) to prevent from unnecessary computations when, any unnecessary state changes occurs or when parent component re reder.

_Extremely sorry if I made any mistakes :(_

### DIFF
--- a/packages/react-query/src/useBaseQuery.ts
+++ b/packages/react-query/src/useBaseQuery.ts
@@ -131,7 +131,7 @@ export function useBaseQuery<
   }
 
   // Handle result property usage tracking
-  return !defaultedOptions.notifyOnChangeProps
+  return React.useMemo(() => !defaultedOptions.notifyOnChangeProps
     ? observer.trackResult(result)
-    : result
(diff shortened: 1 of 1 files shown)


packages/react-query/src/useBaseQuery.ts
```diff
@@ -131,7 +131,7 @@ export function useBaseQuery<
   }
 
   // Handle result property usage tracking
-  return !defaultedOptions.notifyOnChangeProps
+  return React.useMemo(() => !defaultedOptions.notifyOnChangeProps
     ? observer.trackResult(result)
-    : result
+    : result, [defaultedOptions.notifyOnChangeProps, observer.trackResult, result])
```
TkDodo, line 136: `result` comes from `observer.getOptimisticResult`:

https://github.com/TanStack/query/blob/34ca5581217dae4133046f6cfbc746a22be14dd9/packages/react-query/src/useBaseQuery.ts#L81

which is re-computed on every render and will yield a new object every time. So this memoization does nothing in practice I'm afraid.

Issue #4187, opened 2022-09-19
useQueries with duplicate query keys returns wrong value from select when placeholderData is not an object


allansson:
### Describe the bug

When passing two queries with the same query key to `useQueries`, with a `select` function and a non-object value as placeholder data, the data of both entries in the query results array will be the value of the first call to `select`.

```ts
const results = useQueries({
  queries: [
    { queryKey: ["abc"], queryFn: () => fetchSomething(), placeholderData: null, select: data => ({ id: 1, data }) },
    { queryKey: ["abc"], queryFn: () => fetchSomething(), placeholderData: null, select: data => ({ id: 2, data }) },
  ]
})

// prints { id: 1, data: null }
console.log(results[0].data)

// prints { id: 1, data: null }, but id should be 2
console.log(results[1].data)
```

This causes a "duplicate key" warning in React, because the `id` is the same for two items.

Once the fetch is completed the data is correct. It also works if you provide an object value (e.g. `{}`).

### Your minimal, reproducible example

https://codesandbox.io/s/gracious-solomon-pvbv9m

### Steps to reproduce

1. Run the following code in a component.

```ts
const results = useQueries({
  queries: [
    {
      queryKey: ["abc"],
      queryFn: () => new Promise(() => {}),
      placeholderData: null,
      select: (data) => ({ id: 1, data }),
      enabled: false
    },
    {
      queryKey: ["abc"],
      queryFn: () => new Promise(() => {}),
      placeholderData: null,
      select: (data) => ({ id: 2, data }),
      enabled: false
    }
  ]
});

console.log(results[0].data)
console.log(results[1].data)
```
2. Check the output in the console.

### Expected behavior

As a user, I expected the `data` returned for each query to be mapped using the `select`-function.

### How often does this bug happen?

Every time

### Platform

- OS: macOS
- Browser: Firefox Developer Edition
- Version: 105.0b9

### react-query version

v4.3.9

### TypeScript version

v4.1.3

### Additional context

In our app, we are fetching time series data for some charts based on user configuration. The configuration consists of some UI-stuff (e.g. name and color of the time series) and a query used to fetch the time series. It looks roughly like this:

```ts
interface ChartConfig {
  series: TimeSeriesConfig[]
}

interface TimeSeriesConfig {
  label: string
  color: string
  query: { metric: string, method: string }
}
```

[...]

PR #4303, opened 2022-10-11
docs(useQueries): duplicate query keys disclaimer

allansson:
This PR documents the issue described in #4187.

I considered adding it to `Parallell queries` as well, but since it's such a specific issue I decided not to.

### DIFF
--- a/docs/reference/useQueries.md
+++ b/docs/reference/useQueries.md
@@ -21,6 +21,8 @@ The `useQueries` hook accepts an options object with a **queries** key whose val
 - `context?: React.Context<QueryClient | undefined>`
   - Use this to use a custom React Query context. Otherwise, `defaultContext` will be used.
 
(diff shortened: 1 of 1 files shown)


docs/reference/useQueries.md
```diff
@@ -21,6 +21,8 @@ The `useQueries` hook accepts an options object with a **queries** key whose val
 - `context?: React.Context<QueryClient | undefined>`
   - Use this to use a custom React Query context. Otherwise, `defaultContext` will be used.
 
+> Having the same query key more than once in the array of query objects may cause some data to be shared data between queries, e.g. when using `placeholderData` and `select`. To avoid this, consider de-duplicating the queries and map the results back to the desired structure.
```
TkDodo, line 24:
```suggestion
> Having the same query key more than once in the array of query objects may cause some data to be shared between queries, e.g. when using `placeholderData` and `select`. To avoid this, consider de-duplicating the queries and map the results back to the desired structure.
```

PR #4308, opened 2022-10-11
fix(solid-query): fix disappearing data key during fetching state

ardeora:
#4288 uncovered a bug in createBaseQuery.tsx, this PR aims to solve this. 

### The issue
`createBaseQuery` uses a SolidJS store to persist the queryObserver result and exposes it when `createQuery` is called. Scenarios when a query is reset or when a new query key is mounted on the same observer, The queryObserver result initially updates into this state -> `{ data: undefined, isLoading: true; }`.  Updating the store with `setStore` when `data` is `undefined` removes the `data` property from the store. ref: [Updating Stores in SolidJS](https://www.solidjs.com/docs/latest/api#updating-stores)

This is why this fix uses an empty `Symbol` in place of undefined data to prevent SolidJS from deleting the data key and keeps the state consistent between updates.

### DIFF
--- a/packages/solid-query/src/createBaseQuery.ts
+++ b/packages/solid-query/src/createBaseQuery.ts
@@ -64,7 +64,7 @@ export function createBaseQuery<
   const unsubscribe = observer.subscribe((result) => {
     taskQueue.push(() => {
       batch(() => {
-        let unwrappedResult = { ...unwrap(result) }
+        const unwrappedResult = { ...unwrap(result) }
         if (unwrappedResult.data === undefined) {
           // This is a hack to prevent Solid
           // from deleting the data property when it is `undefined`
--- a/packages/solid-query/src/createBaseQuery.ts
+++ b/packages/solid-query/src/createBaseQuery.ts
@@ -31,7 +31,7 @@ export function createBaseQuery<
   Observer: typeof QueryObserver,
 ): QueryObserverResult<TData, TError> {
   const queryClient = useQueryClient({ context: options.context })
-
+  const emptyData = Symbol('empty')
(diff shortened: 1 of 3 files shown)


packages/solid-query/src/createBaseQuery.ts
```diff
@@ -31,7 +31,7 @@ export function createBaseQuery<
   Observer: typeof QueryObserver,
 ): QueryObserverResult<TData, TError> {
   const queryClient = useQueryClient({ context: options.context })
-
+  const emptyData = Symbol('empty')
```
TkDodo, line 34: would it be safer if we crated the symbol outside of `createBaseQuery` ? creating the symbol twice will not give referential identity. Just wondering if there is a case where `createBaseQuery` would run again and the comparison would fail.

I think because it's a const, when the resource functions runs it will closure over the emptyData from when it was created, but still: shouldn't there be _exactly one_ `emptyData` symbol ?

ardeora: That's a great question 😁 Solid JS hooks and components are created only once. Unlike react, They never 're-render'. The options and functions are made reactive using reactive utility functions that SolidJS provides like 'createComputed', 'createEffect', etc. The emptyData const will always have the same referential identity since it will only be created once, throughout the hook's lifespan.

TkDodo: of course 🤦‍♂️. I'm still thinking in react re-renderings lol.

TkDodo (approved):

TkDodo:
@allcontributors please add @ardeora for code, maintenance

PR #4315, opened 2022-10-13
feat(query-core): Expose retry error as `failureReason`

Kuirak:
Added `failureReason: TError | null` to query as suggested here:
https://github.com/TanStack/query/discussions/4305

Open Question: Would it make sense to add this to mutations as well?


packages/query-core/src/query.ts
```diff
@@ -473,8 +475,8 @@ export class Query<
         this.isFetchingOptimistic = false
       },
       onError,
-      onFail: () => {
-        this.dispatch({ type: 'failed' })
+      onFail: (_, error) => {
+        this.dispatch({ type: 'failed', error })
```
TkDodo, line 479: there seems to be no consumer that actually needs the `failureCount` that is passed here as the first argument to `onFail`, so maybe we can just remove that? The `failureCount` is just incremented internally ...

Or, we use the `failureCount` here and pass it to our internal reducer instead of incrementing it there. I just think the middle ground we have now is kinda weird because we increment in two places:

https://github.com/TanStack/query/blob/ca767771f3720ba9abb541871c22ea106172e9ab/packages/query-core/src/retryer.ts#L176-L179

TkDodo:
> Open Question: Would it make sense to add this to mutations as well?

I would say so, yes.

packages/query-core/src/query.ts
Kuirak: Since `failureCount` is used within `retryer.ts` at multiple points it might make more sense to use this value in the `reducer` as well. I will change it accordingly

Kuirak:
> > Open Question: Would it make sense to add this to mutations as well?
> 
> I would say so, yes.

While implementing it I stumbled over:
1. `failureCount` was not documented for `useMutation`
2. `failureCount` implementation differs as it doesn't reset to `0` after a successful mutation, this led to not using the `failureCount` from the action but incrementing the one in the state

docs/reference/useMutation.md
```diff
@@ -129,3 +131,9 @@ mutate(variables, {
   - The error object for the query, if an error was encountered.
 - `reset: () => void`
   - A function to clean the mutation internal state (i.e., it resets the mutation to its initial state).
+- `failureCount: number`
+  - The failure count for the query.
+  - Incremented every time the query fails.
+- `failureReason: null | TError`
+  - The failure reason for the query retry.
```
TkDodo, line 138:
```suggestion
- `failureCount: number`
  - The failure count for the mutation.
  - Incremented every time the mutation fails.
  - Reset to `0` when the mutation succeeds.
- `failureReason: null | TError`
  - The failure reason for the mutation retry.
  - Reset to `null` when the mutation succeeds.
```

packages/query-core/src/mutation.ts
```diff
@@ -171,8 +173,8 @@ export class Mutation<
           }
           return this.options.mutationFn(this.state.variables!)
         },
-        onFail: () => {
-          this.dispatch({ type: 'failed' })
+        onFail: (_failureCount, error) => {
+          this.dispatch({ type: 'failed', error })
```
TkDodo, line 177: I thought we'd also wanted to dispatch the failureCount here as well and _not_ increment it in the reducer?

```
this.dispatch({ type: 'failed', failureCount, error })
```

TkDodo:
> failureCount implementation differs as it doesn't reset to 0 after a successful mutation

I'd say this is a bug, isn't it? Why would the failureCount behave differently for mutations

TkDodo:
@ardeora @DamianOsipiuk if you have some time, please take a look at the PR and let me know if you agree with the api decision

packages/query-core/src/mutation.ts
Kuirak: This is what I meant with:
> failureCount implementation differs as it doesn't reset to 0 after a successful mutation, this led to not using the failureCount from the action but incrementing the one in the state

I didn't want to alter the behavior here. But if it is considered a bug I can fix it.

TkDodo: I'm pretty sure this is not the intended behavior

DamianOsipiuk:
Looks good to me 👍

packages/query-core/src/mutation.ts
```diff
@@ -298,6 +302,8 @@ export class Mutation<
           return {
             ...state,
             data: action.data,
+            failureCount: 0,
+            failureReason: null,
```
TkDodo, line 306: I think when `loading` is dispatched, we should also reset the `failureCount` and the `failureReason`. At least this is what's happening in `useQuery`:

https://github.com/TanStack/query/blob/04945f38b3f2ff82a597083bf1a36724f973bc7e/packages/query-core/src/query.ts#L518

TkDodo: seems like we don't have a test for this behaviour? Could you maybe add one?

Kuirak: Ok will do :)

TkDodo:
seems like the new test fails and there are also lint errors.

Kuirak:
> seems like the new test fails and there are also lint errors.

Oh I am sorry, I forgot to run prettier. Haven't setup format on save in this repo yet.

=> Fixed Tests and linting

TkDodo (approved):

PR #4330, opened 2022-10-16
docs(examples/vue): add example for dependent queries

cngJo:
This basically adds a copy of the `vue/basic` example thats adds code to show of how dependent queries work with vuejs.

This is vaguely documented in `packages/vue-query/README.md` by showing the `enabled` option and passing a `ref(false)` to it. 
This examples shows how to use a value from the response of a previous query and feed it into a new query.

 I haven't found any documentation about this with vue's rectivity / composition api and it took quite some tome for me to figure this out. That my be related to be being a backend person and not that familiar yet with the new vue composition API.
But i wanted to build a example for this to make it easier for others to find this.

The basic example gets updated by fetching the author for the loaded post by the `userId` contained in the post response.

I thought about integrating it into the `vue/basic` example but wanted to leave it as a clean basic example. I can integrate my changes into the basic example if wanted.

---

NOTE: I haven't created a issue for this new example, since I didn't want to create a bug report for this change which would be the only option to create a new issue.


examples/vue/dependent-queries/src/Post.vue
```diff
@@ -0,0 +1,66 @@
...
+  setup(props) {
+    const { isLoading, isError, isFetching, data: post, error } = useQuery(
+      ["post", props.postId],
+      () => fetchPost(props.postId)
+    );
+
+    const authorId = computed(() => post.value?.userId);
+
+    const { data: author } = useQuery(
+      ["author", authorId.value],
```
DamianOsipiuk, line 35:
```suggestion
      ["author", authorId],
```
You should not unwrap reactive value here, otherwise it will stay under the same key, as in vue `setup` function is run only once.

DamianOsipiuk (changes requested):

DamianOsipiuk:
If you are looking for some docs, until they will be migrated: https://vue-query.vercel.app/#/guides/dependent-queries

examples/vue/dependent-queries/src/Post.vue
cngJo: Thank you, updated the code.

cngJo:
Thanks for the link to the docs, thats exactly what I was looking for when I first tried the dependent queries :D

I Updated the code to don't unwrap in the queryKey. Thanks for that suggestion!

DamianOsipiuk:
I think you need to use the same versions of libs as in the other example, or run `pnpm i` and commit also lockfile changes

reslear:
> If you are looking for some docs, until they will be migrated: https://vue-query.vercel.app/#/guides/dependent-queries

By the way, is there any news about the updated documentation?)

TkDodo:
> By the way, is there any news about the updated documentation?

it's coming:
- https://github.com/TanStack/query/pull/4608

cngJo:
> I think you need to use the same versions of libs as in the other example, or run pnpm i and commit also lockfile changes

I've executed `pnpm i` in the `examples/vue/dependent-queries` directory and committed the lock file changes. 

I'm not sure if the changes are correct since I'm not familiar with pnpm, could you please take a look?

Thank you!

DamianOsipiuk:
@cngJo I think you still need to prettify the files 😄

cngJo:
@DamianOsipiuk Done ✅ 

<details>
<summary><code>npm run test:format</code></summary>
<pre>
> test:format
> pnpm run prettier --check
> query@ prettier /path/to/TanStack/query
> prettier "{packages,examples}/**/src/**/*.{md,js,jsx,ts,tsx,json,vue}" "--check"

Checking formatting...
All matched files use Prettier code style!
</pre>
</details>

PR #4339, opened 2022-10-18
fix(vue-query): Make use mutate reactive to argument changes

matthewhausman:
Definitely need someone to help verify this implementation as I am not intimately familiar with how tanstack query is used professionally, only generally familiar.


DamianOsipiuk:
So, from my point of view, there are a lot of unnecessary changes.
How i would see this implemented is to add something like this in `types.ts`:
```ts
// A Vue version of MutationObserverOptions from "@tanstack/query-core"
// Accept refs as options
export type VueMutationObserverOptions<
  TData = unknown,
  TError = unknown,
  TVariables = void,
  TContext = unknown,
> = {
  [Property in keyof MutationObserverOptions<
    TData,
    TError,
    TVariables,
    TContext
  >]: MaybeRefDeep<
    MutationObserverOptions<TData, TError, TVariables, TContext>[Property]
  >
}
```

and then use it in https://github.com/TanStack/query/blob/main/packages/vue-query/src/useMutation.ts#L23

This will also require change in
https://github.com/TanStack/query/blob/main/packages/vue-query/src/useMutation.ts#L168
and
https://github.com/TanStack/query/blob/main/packages/vue-query/src/useMutation.ts#L183

to use non reffed version of options which is currently at https://github.com/TanStack/query/blob/main/packages/vue-query/src/useMutation.ts#L23

matthewhausman:
@DamianOsipiuk let me know what ya think! I'd like if functions could also be accepted as refs by tanstack query if possible but what do you think? (hence the chance to the deep ref function)

packages/vue-query/src/__tests__/useMutation.test.ts
```diff
@@ -35,11 +35,8 @@ describe('useMutation', () => {
 
   test('should return error when request fails', async () => {
     const mutation = useMutation(errorMutator)
-
-    mutation.mutate()
```
DamianOsipiuk, line 39: I do not think changes in this file are relevant

packages/vue-query/src/useMutation.ts
```diff
@@ -125,11 +137,10 @@ export function useMutation<
     })
   }
 
-  watch(
-    [() => arg1, () => arg2, () => arg3],
+  watch(options,
```
DamianOsipiuk, line 140: Could you please explain the reason for this change?
Also could you add a test for it to verify if deep property watching still works?

matthewhausman: Essentially just to reuse the logic for providing an argument to queryClient.defaultMutationOptions and the initial declaration for the observer var. Realized 143 is definitely incorrect however, the argument should just be options.value!

Also reduces the redundancy in defining when the mutation options are recalced, but not super important. lmk if I should just do something more similar to what was already implemented :)

Will absolutely write a test. thanks for being patient with this really appreciate it

packages/vue-query/src/__tests__/useMutation.test.ts
matthewhausman: Ok will take another look - for some reason or another passing no args was failing the tests.

matthewhausman: Screen Shot 2022-10-22 at 4 33 44 PM

Any idea why these changes would cause something as such?

packages/vue-query/src/useMutation.ts
matthewhausman: In terms of testing deep reactivity - a library I'm trying to replicate (the reason I'm making this PR) is passing an object as the mutationKey so I think this should be sufficient basis for testing deep reactivity

packages/vue-query/src/__tests__/useMutation.test.ts
DamianOsipiuk: Actually no idea.
Does it work on `main` branch?
What TS version do you have installed?

packages/vue-query/src/useMutation.ts
DamianOsipiuk: Since you have removed `defaultMutationOptions` from watch, i think you need to move it to `computed` and remove it as well from line 124.

As for deep reactivity, i'm not sure if `computed` can detect such changes. And since now watcher is based on `computed` result it will not run until `computed` value is updated. You are removing reacivity from all options when calling `parseMutationArgs` there, so this is why i'm afraid that it will only work when you update the immediate child prop of an object for example.
I might be wrong though, that is why it would be good to write a test case that is checking this, and will make sure that it would not break in the future.

packages/vue-query/src/__tests__/useMutation.test.ts
matthewhausman: that test is running on your test suite - not my local so not sure if my ts version is relevant

matthewhausman: export type MutateFunction<
  TData = unknown,
  TError = unknown,
  TVariables = void,
  TContext = unknown,
> = (
  variables: TVariables,
  options?: MutateOptions<TData, TError, TVariables, TContext>,
) => Promise<TData> 

it seems completely reasonable to get an error for not passing args - variables is not optional. Maybe I somehow removed a Partial<MutateFunction> somewhere... pretty strange

packages/vue-query/src/useMutation.ts
matthewhausman: test case for deep reactivity is working - let me know if it is bad style thanks

matthewhausman: "Since you have removed defaultMutationOptions from watch, i think you need to move it to computed and remove it as well from line 124." 

are you suggested to make queryClient.defaultMutationOptions(options.value) its own computed property? Definitely think that can be done, just want to clarify

DamianOsipiuk: Was thinking of
```ts
const options = computed(() => {
  return queryClient.defaultMutationOptions(parseMutationArgs(arg1, arg2, arg3))
})
```
and reuse it in both places, since you are already moving stuff

matthewhausman: in the instantiation of the queryClient we need options.value.queryClientKey, so above snippet is not achievable as far as I can see...

matthewhausman: As far as I can see that wouldn't be possible because in order to create the const queryClient we need the options.value.queryClient and useQueryClient(options.value.queryClientKey) as a fallback

matthewhausman:
I have a separate concern with the current typing of the arguments - I see the most popular way to create reactive arguments to be to make a single options object, so const options = ref({ mutationKey: ..., mutationFn: ... })

currently, we do not allow for such an argument to be passed. I generally think that we should adjust the type signature to something like what I have just committed

1) 
export type UseMutationOptions<TData, TError, TVariables, TContext> =
  WithQueryClientKey<
    MutationObserverOptions<TData, TError, TVariables, TContext>
  >

The mutation options above are all non-reactive

2) 
export type VueMutationObserverOptions<
  TData = unknown,
  TError = unknown,
  TVariables = void,
  TContext = unknown,
> = {
  [Property in keyof UseMutationOptions<
    TData,
    TError,
    TVariables,
    TContext
  >]: MaybeRefDeep<
    UseMutationOptions<TData, TError, TVariables, TContext>[Property]
  >
}

the Vue options use maybe deep ref on each key....

3)
options: MaybeRef<VueMutationObserverOptions<TData, TError, TVariables, TContext>>,

the object itself is allowed to be reactive

matthewhausman:
Lastly I have tried all of the prettier formatting commands defined in package.json but still getting errors in the workflow tests somehow, any idea how to fix that?

DamianOsipiuk:
> Lastly I have tried all of the prettier formatting commands defined in package.json but still getting errors in the workflow tests somehow, any idea how to fix that?

yeah, i think the `write` script was not updated after the switch to `pnpm`.

Change `pnpm run prettier -- --write` to `pnpm run prettier --write` and then run it and it should work

matthewhausman:
worked like a charm thanks!!!!!

matthewhausman:
So the test that's failing is a coverage test for the util that I added right?? pretty hard to tell lol

matthewhausman:
Woooo 🎉🎉🎉🎉🎉🎉🎉 test passed! lmk if you want to zoom or something and go over before we move forward. Thank you so much for the help it means so much to help with my first open source contribution :)

DamianOsipiuk:
Thanks for the contribution @matthewhausman 🚀

TkDodo:
@allcontributors add @matthewhausman for code

PR #4363, opened 2022-10-23
fix(vue-query): fix bug where ref is destructured

matthewhausman:
non-plain object is no longer destructured. tests added to prevent similar bugs from occuring going forward. improve parseMutationArgs function


matthewhausman:
getting a react query error on the top set of tests

TkDodo:
sorry about that, it's a flaky test. I really need to get around to remove all those weird sleep timings in most test assertions :/

matthewhausman:
all good !

PR #4364, opened 2022-10-23
feat(eslint-plugin-query): initial commit

Newbie012:
This PR contains two projects:

- an ESLint plugin
- an example project that uses the plugin

A few notes:

- This PR support (currently) only array expression of `queryKey` (i.e, query key factory pattern is not supported).
- I wasn't quite sure what's the preferences for building a production/dev build, so I stuck with my original setup ([tsup](https://github.com/egoist/tsup))


packages/eslint-plugin-query/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "url": "https://github.com/sponsors/tannerlinsley"
+  },
+  "main": "build/lib/index.js",
+  "scripts": {
+    "typecheck": "tsc",
+    "build": "tsup --minify",
+    "dev": "tsup --watch --sourcemap",
+    "test:jest": "../../node_modules/.bin/jest --config ./jest.config.ts",
+    "test:dev": "pnpm run test:jest --watch",
+    "lint:eslint": "eslint --ext .js,.jsx,.ts,.tsx src"
```
TkDodo, line 20: it has to be named `test:eslint` like in all the other packages so that `pnpm run test:estlint` picks it up for this package as well:

```suggestion
    "test:eslint": "eslint --ext .js,.jsx,.ts,.tsx src"
```

examples/react/basic-eslint-plugin/package.json
```diff
@@ -0,0 +1,35 @@
...
+  },
+  "dependencies": {
+    "axios": "^0.21.1",
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0",
+    "@tanstack/react-query": "^4.7.1",
+    "@tanstack/react-query-devtools": "^4.7.1"
+  },
+  "devDependencies": {
+    "@tanstack/eslint-plugin-query": "^1.0.0",
```
TkDodo, line 18: I don't think that we need a dedicated example for the lint plugin. would we actually see somewhere in the code that the lint rule shows a warning?

If you want, we can add the eslint package just to the already existing basic example?

packages/eslint-plugin-query/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@tanstack/eslint-plugin-query",
+  "version": "1.0.0",
```
TkDodo, line 3: versions have to be in line with the rest of the monorepo I guess, so let's put this to v4.10.3 please

packages/eslint-plugin-query/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "repository": "tanstack/query",
+  "homepage": "https://tanstack.com/query",
+  "funding": {
+    "type": "github",
+    "url": "https://github.com/sponsors/tannerlinsley"
+  },
+  "main": "build/lib/index.js",
+  "scripts": {
+    "typecheck": "tsc",
+    "build": "tsup --minify",
```
TkDodo, line 16: for the build step to work and this to be published, it has to be included in `pnpm run build`, which currently executes this:

https://github.com/TanStack/query/blob/17a362b141b042f0fd9ff3ce849fd8858446fb30/package.json#L15

It's fine if we want to build this with tsup rather than rollup, though it being a dev dependency for consumers it doesn't really matter. I don't know the rollup config very well but if we could include it there, maybe that would be easier?

@DamianOsipiuk knows better how this works, and also knows if we have to add `module` and `exports` to `package.json` to make esm work?

packages/eslint-plugin-query/jest.config.ts
```diff
@@ -0,0 +1,4 @@
+export default {
+  displayName: 'query-core',
```
TkDodo, line 2:
```suggestion
  displayName: 'eslint-plugin-query',
```

packages/eslint-plugin-query/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "outDir": "build/lib"
+  },
+  "devDependencies": {
+    "@typescript-eslint/eslint-plugin": "^5.40.1",
+    "@typescript-eslint/parser": "^5.40.1",
+    "@typescript-eslint/types": "^5.40.1",
+    "@typescript-eslint/utils": "^5.40.1",
+    "eslint": "^8.26.0",
+    "tsup": "^6.3.0",
+    "typescript": "4.8.4"
```
TkDodo, line 43: typescript is defined as devDependency in the root of the monorepo, and I think having only one version of TS is important:

```suggestion
    "tsup": "^6.3.0"
```

TkDodo (commented):
This is very good work, thank you ❤️ 

What we would need to add is some docs for the eslint plugin, how to download it, how to use it, what the rule does etc.

TkDodo:
also, @KubaJastrz I really like your eslint-plugin-react-query, especially this rule:

https://github.com/KubaJastrz/eslint-plugin-react-query/blob/master/docs/rules/prefer-query-object-syntax.md

So if you don't object, I would eventually like to add that one here, too :)

packages/eslint-plugin-query/src/rules/exhaustive-deps.test.ts
```diff
@@ -0,0 +1,83 @@
+import { ESLintUtils } from '@typescript-eslint/utils'
+import { exhaustiveDepsRule } from './exhaustive-deps.rule'
+
+const ruleTester = new ESLintUtils.RuleTester({
+  parser: '@typescript-eslint/parser',
+  settings: {},
+})
+
+ruleTester.run('exhaustive-deps', exhaustiveDepsRule, {
+  valid: [
```
TkDodo, line 10: can we add a test that just shows that a basic useQuery is okay?

```
useQuery(['todos'], fetchTodos)
```

packages/eslint-plugin-query/src/rules/exhaustive-deps.test.ts
```diff
@@ -0,0 +1,83 @@
...
+            queryKey: ["entity"],
+            queryFn: () => {
+                const { data } = axios.get(\`.../\${id}\`);
+                return data;
+            }
+        });
+      `,
+      output: `
+        const [id] = React.useState(1);
+        useQuery({
```
TkDodo, line 72: i can see that all tests use the object syntax for useQuery. is this a limitation that it will only work with that syntax, or does it also work for `useQuery(key, fn)` ?

(note that we want to move towards `useQuery(options)` in v5 anyways, so it wouldn't be a big deal)

packages/eslint-plugin-query/src/rules/exhaustive-deps.rule.ts
```diff
@@ -0,0 +1,173 @@
...
+    queryFn === undefined ||
+    queryFn.value.type !== AST_NODE_TYPES.ArrowFunctionExpression
+  ) {
+    return
+  }
+
+  if (queryKey.value.type !== AST_NODE_TYPES.ArrayExpression) {
+    // TODO support query key factory
+    return
+  }
```
TkDodo, line 99: wouldn't a rather simple approach at tackling this be to look if the queryKey is a function, and if it is, look at the params of that function?

So:

`useQuery({ queryKey: ["todos", id] })`

would be the same as:

`useQuery({ queryKey: todoKeys.detail(id) })`

they both reference a variable `id`.

packages/eslint-plugin-query/src/rules/exhaustive-deps.rule.ts
```diff
@@ -0,0 +1,173 @@
...
+    fixable: 'code',
+    schema: ruleSchema,
+  },
+  defaultOptions: [defaultRuleOptions],
+  create(context) {
+    return {
+      CallExpression(node) {
+        if (
+          ExtraUtils.isIdentifier(node.callee) &&
+          node.callee.name === 'useQuery'
```
TkDodo, line 56: instead of checking for `useQuery`, would it be possible to check for `queryKey` and `queryFn`? In the future (v5), when we only allow objects, I think this pattern will be popular:

```
const todoQueries = {
  list: () => ({ queryKey: ['todos'], queryFn: fetchTodos })
  detail: (id) => ({ queryKey: ['todos', id], queryFn: () => fetchTodo(id) })
}

useQuery(todoQueries.list())
useQuery(todoQueries.detail(5))

queryClient.prefetchQuery(todoQueries.detail(23))
```

this kind of takes the "query key factory" towards a full blown "query factory" where you can share the resulting object between `useQuery` and methods like `prefetchQuery` or even `invalidateQueries`...

packages/eslint-plugin-query/src/rules/exhaustive-deps.test.ts
Newbie012: I haven't implemented the "traditional" syntax, but I guess I could write an edge case for this. But what if I could write another lint rule that will auto-fix "traditional" syntax into "object" syntax?

packages/eslint-plugin-query/src/rules/exhaustive-deps.rule.ts
Newbie012: Correct. But it can be a bit more complex:

```
queryKey: todoKeys.a({ x })
queryKey: todoKeys.a([b, { c }])
```

I need to recursively iterate over the nodes and then determine:

1. If all of the dependencies are there (trivial)
2. If there's a missing dependency, how should I auto-fix it? the function can accept anything. I can idiomatically put it inside the last function expression.

packages/eslint-plugin-query/src/rules/exhaustive-deps.rule.ts
Newbie012: That is a very interesting approach! I'll write some tweaks.

KubaJastrz:
@TkDodo I'm okay with that, let me know if you want to merge projects, as I'm no longer actively maintaining it. 

Let me know if you need npm package name too.

packages/eslint-plugin-query/package.json
Newbie012: eslint doesn't work with esm. It has to be bundled as cjs.

packages/eslint-plugin-query/src/rules/exhaustive-deps.test.ts
Newbie012: added as an "object" syntax

Newbie012:
Additional improvements:

- Works with `@babel/eslint-parser` (the default parser).
- Automatically whitelists call expressions (e.g `fetch(...)`, `axios.get(...)`, `api.entity.get(...)`).

packages/eslint-plugin-query/src/rules/exhaustive-deps.test.ts
DamianOsipiuk: This would be useful for transition from `v4` to `v5`, as instead of creating codemod, we could say `please install eslint plugin` for auto-fix.
This could also increase adoption rate.

packages/eslint-plugin-query/src/rules/exhaustive-deps.rule.ts
```diff
@@ -0,0 +1,176 @@
...
+    return
+  }
+
+  if (queryKey.value.type !== AST_NODE_TYPES.ArrayExpression) {
+    // TODO support query key factory
+    return
+  }
+
+  const queryKeyValue = queryKey.value
+  const queryKeyDeps =
```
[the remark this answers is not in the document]
Newbie012: You're correct. I think I have something in mind that will address it. I'll write a test case for it as well.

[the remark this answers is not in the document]
Newbie012: [recast](https://github.com/benjamn/recast) can simplify this process by node manipulation.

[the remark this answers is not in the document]
Newbie012: It should be ok now. Added test cases as well.

packages/eslint-plugin-query/package.json
DamianOsipiuk: Yeah, it does not bring to much benefit to build `esm` for eslint right now, maybe in the future.
Nevertheless it would be neat to use the same build tools for all of the packages.
But i guess we could do that later, when we do some refactoring of current rollup config.

Newbie012:
I'm not exactly sure why the CI is failing

TkDodo:
>  I'm not exactly sure why the CI is failing

eslint is failing:

```
packages/eslint-plugin-query test:eslint:   0:0  error  Parsing error: Cannot read file '/home/runner/work/query/query/packages/eslint-plugin-query/tsconfig.base.json'
```

tests are flaky so I re-ran them. I think this got worse since we parallelized the pipeline, not sure why. I need to get around to stabilize those, sorry.

DamianOsipiuk (approved):

Newbie012:
Wait, I found an issue when specifying a member expression. I'll fix it later:

```diff
function Component() {
-    useQuery({ queryKey: ["todos"], queryFn: () => fetch(props.id) });
+    useQuery({ queryKey: ["todos", props], queryFn: () => fetch(props.id) }); // ❌ should be `props.id`
}
```

Newbie012:
Fixed. I believe everything is (hopefully) working as expected.

packages/eslint-plugin-query/package.json
TkDodo: > Nevertheless it would be neat to use the same build tools for all of the packages.

okay so can we switch to `rollup` just without ESM ?

packages/eslint-plugin-query/src/rules/exhaustive-deps.test.ts
TkDodo: yeah this is actually a rule that @KubaJastrz has in his eslint-plugin:

- [prefer-object-syntax](https://github.com/KubaJastrz/eslint-plugin-react-query/blob/master/docs/rules/prefer-query-object-syntax.md)

and it's also fixable ❤️ 

since we got his permission to use it, maybe we should just copy it over and then it's totally fine that the exhaustive-deps rule only works on the object syntax. It will give people another incentive to switch towards that syntax, and it will also make the migration path to v5 better.

packages/eslint-plugin-query/src/rules/exhaustive-deps.rule.ts
TkDodo: I don't think auto-fixing is required if it gets too complex. The order of arrays is important (for example, for invalidation), so you'd likely need to change that anyways (unlike the react hooks rule, where the order doesn't matter).

packages/eslint-plugin-query/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@tanstack/eslint-plugin-query",
+  "version": "4.13.0",
+  "description": "ESLint plugin for React Query",
```
TkDodo, line 4:
```suggestion
  "description": "ESLint plugin for TanStack Query",
```

packages/eslint-plugin-query/src/rules/exhaustive-deps.test.ts
```diff
@@ -0,0 +1,179 @@
...
+          }
+        `,
+      output: `
+          function Component({ map, key }) {
+            useQuery({ queryKey: ["key", map[key]], queryFn: () => api.get(map[key]) });
+          }
+        `,
+      errors: [{ messageId: 'missingDeps', data: { deps: 'map[key]' } }],
+    },
+  ],
```
TkDodo, line 178: another edge-case that just came to my mind that is actually legit and might yield false positives is:

```
useQuery({
  queryKey: ['foo', dep1],
  queryFn: ({ queryKey: [,dep] }) => api.get(dep)
```

because the `queryKey` is passed to the `queryFn` inside the `queryFunctionContext` - the object that the `queryFn` receives as input.

packages/eslint-plugin-query/src/rules/exhaustive-deps.rule.ts
Newbie012: Seems right. I'll write an implementation without an auto fix

packages/eslint-plugin-query/src/rules/exhaustive-deps.test.ts
Newbie012: I wish I knew this approach earlier. It would've make typescript happier... I believe I ignore all of the "local dependencies", but I'll have to check that once I get back to it.

Newbie012: added a test case. It seems to be working as expected (not adding dep1 to the deps)

TkDodo: nice 🚀

packages/eslint-plugin-query/docs/rules/prefer-query-object-syntax.md
```diff
@@ -0,0 +1,53 @@
+# Prefer object syntax for useQuery (`@tanstack/query/prefer-query-object-syntax`)
+
+You can use [`useQuery`](https://tanstack.com/query/v4/docs/reference/useQuery) in two different ways.
```
Newbie012, line 3: When v5 is out, we should update the link here

packages/eslint-plugin-query/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "homepage": "https://tanstack.com/query",
+  "funding": {
+    "type": "github",
+    "url": "https://github.com/sponsors/tannerlinsley"
+  },
+  "main": "build/lib/index.js",
+  "scripts": {
+    "typecheck": "tsc",
+    "build": "tsup --minify",
+    "dev": "tsup --watch --sourcemap",
```
Newbie012, line 17: if we decide to use [rollup to bundle as cjs](https://rollupjs.org/guide/en/#--bundleconfigascjs), then we should ensure it'll generate source maps when working in development mode for easier debugging (breakpoints, etc).

Newbie012 (commented):
I have merged [eslint-plugin-react-query](https://github.com/KubaJastrz/eslint-plugin-react-query) with this PR. Hopefully, it won't be too big for a merge.

packages/eslint-plugin-query/docs/rules/prefer-query-object-syntax.md
TkDodo: when v5 is out, we can probably delete this rule because there will only be the query syntax 🤔. Or, we'll keep it so that people can use the auto fix to migrate and then delete it later...

packages/eslint-plugin-query/package.json
TkDodo: I'm fine with keeping tsup for now if it works.

packages/eslint-plugin-query/docs/rules/exhaustive-deps.md
```diff
@@ -0,0 +1,41 @@
+# Prefer object syntax for useQuery (`@tanstack/query/exhaustive-deps`)
+
+TODO @TkDodo a description of this rule
```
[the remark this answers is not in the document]
TkDodo: done: bc4c70a216a8c7dbc8ca1750ed0f7d21ff2f4527

examples/react/basic/.eslintrc
```diff
@@ -1,7 +1,8 @@
 {
   "extends": ["react-app", "prettier"],
+  "plugins": ["@tanstack/eslint-plugin-query"],
   "rules": {
-    // "eqeqeq": 0,
-    // "jsx-a11y/anchor-is-valid": 0
+    "@tanstack/query/exhaustive-deps": "error",
+    "@tanstack/query/prefer-query-object-syntax": "error"
   }
```
TkDodo, line 7: is there a "recommended" ruleset? how would I use it?

TkDodo: would this be:

```
"extends": ["@tanstack/eslint-plugin-query/recommended"]
```

?

Just trying to get this right for the docs 😅

packages/eslint-plugin-query/src/rules/prefer-query-object-syntax/prefer-query-object-syntax.ts
```diff
@@ -0,0 +1,118 @@
+import type { TSESLint } from '@typescript-eslint/utils'
+import { createRule } from '../../utils/create-rule'
+import { ASTUtils } from '../../utils/ast-utils'
+
+const USE_QUERY = 'useQuery'
```
TkDodo, line 5: if we want this plugin to work for `solid-query` as well, we'd need to check for `createQuery` as well, right @ardeora ?

packages/eslint-plugin-query/src/utils/detect-react-query-imports.ts
```diff
@@ -0,0 +1,72 @@
...
+        })
+      },
+    }
+
+    const detectionInstructions: TSESLint.RuleListener = {
+      ImportDeclaration(node) {
+        if (
+          node.specifiers.length > 0 &&
+          node.importKind === 'value' &&
+          node.source.value === '@tanstack/react-query'
```
TkDodo, line 38: also here, this wouldn't work for other adapters 🤔

examples/react/basic/.eslintrc
Newbie012: Yes:

```
{
  "extends": ["plugin:@tanstack/eslint-plugin-query/recommended"]
}
```

packages/eslint-plugin-query/src/utils/detect-react-query-imports.ts
Newbie012: I agree. Although, there are other libraries who have `useQuery` as well, such as [Chakra UI's useQuery](https://github.com/chakra-ui/chakra-ui/blob/f0e3fb61e3064f03e64222650cf0f1570843a146/packages/components/media-query/src/media-query.tsx#L13) (I always accidentally import that hook 😅).

I copied the same logic that was in `react-query-eslint-plugin`. I can remove that part.

TkDodo: I think checking the import is fine, but it would need to be widened to include all tanstack adapters

packages/eslint-plugin-query/src/rules/prefer-query-object-syntax/prefer-query-object-syntax.ts
Newbie012: done

packages/eslint-plugin-query/src/utils/detect-react-query-imports.ts
Newbie012: done

examples/react/basic/.eslintrc
TkDodo: are you sure it's `@tanstack/eslint-plugin-query` and not `@tanstack/query` ? I'm always confused by when to include `eslint-plugin` and when not 😅

TkDodo: the rules in the example say:

```
"@tanstack/query/exhaustive-deps": "error",
```

TkDodo: also: does it need the `typescript-eslint` parser to be present, or will it add it manually? Do the rules only work on TypeScript files?

Sorry for all the questions, just trying to write something for the docs and want to understand it properly 😅

Newbie012: I understand why it doesn't look very clear, It depends on the plugin's package name. Whether it's a scoped package or a standard name (in our case it's scoped).

the pattern of `extends` is "plugin:{package_name}/{config_name}"
the pattern of `rules` is: "{package_name minus prefix 'eslint-plugin-'}/{rule_name}

`eslint-plugin-` is omitted since it's redundant.

You don't have to pass `typescript-eslint` as a parser, and it won't be added manually. The plugin now works with babel's parser (which is the default parser).

docs/config.json
```diff
@@ -298,6 +298,23 @@
         }
       ]
     },
+    {
+      "label": "Eslint",
+      "children": [
+        {
+          "label": "Eslint Plugin Query",
```
Newbie012, line 305:
```suggestion
          "label": "ESLint Plugin Query",
```

docs/eslint/eslint-plugin-query.md
```diff
@@ -0,0 +1,49 @@
+---
+id: eslint-plugin-query
+title: Eslint Plugin Query
+---
+
+TanStack Query comes with its own EsLint plugin. This plugin is used to enforce best practices and to help you avoid common mistakes.
```
Newbie012, line 6:
```suggestion
TanStack Query comes with its own ESLint plugin. This plugin is used to enforce best practices and to help you avoid common mistakes.
```

Newbie012 (commented):
I think it should be written as `eslint` or `ESLint` rather than `EsLint`

examples/react/basic/.eslintrc
TkDodo: Thanks for explaining ❤️ 

I've added the docs and moved the rule specific docs to our docs structure as well. I think it's good to go ?

docs/eslint/eslint-plugin-query.md
TkDodo: fixed all `EsLint` -> `ESLint`: 348f4614a3e7e58ce976f5726134fc47faf19b62

docs/config.json
TkDodo: 348f4614a3e7e58ce976f5726134fc47faf19b62

examples/react/basic/.eslintrc
Newbie012: Let's 🚀 it

KubaJastrz (approved):
If you could include my name somewhere in this, that would be great. Thank you!

TkDodo:
@KubaJastrz absolutely:  6ec2643

TkDodo:
@all-contributors add @KubaJastrz for code

TkDodo:
@all-contributors add @Newbie012 for code

TkDodo:
release is here 🎉 

https://github.com/TanStack/query/releases/tag/v4.14.0

Thank you for contributing this. @KubaJastrz if you have a twitter handle, let me know to give you credit in the announcement :)

TkDodo:
@Newbie012 I'm still not so sure about the auto fixing. Consider:

```
const foo = 5
const bar = 10

useQuery({ queryKey: [{ hello: 'world' }], queryFn: () => foo + bar, enabled: false })
```

the key will be fixed to:

```
[{ hello: 'world' }, foo, bar]
```

but it could also be:

```
[{ hello: 'world' }, bar, foo]
```

or, more likely:

```
[{ hello: 'world', foo, bar }]
```

The thing is: we don't know. Maybe we should downgrade this to a suggestion rather than a fix?

Newbie012:
I think you're correct. I'll open a new PR that will downgrade it to a suggestion rather than an auto-fix.

Newbie012:
fixes in https://github.com/TanStack/query/pull/4425

PR #4211, opened 2022-09-23
feat(solid-query): Solid Query Adapter for TanStack Query

lukesmurray:
This pull request aims to add the TanStack Query adapter for Solid JS. This pull request will add the following key primitives to the package.

-   `createQuery`
-   `createQueries`
-   `createInfiniteQueries`
-   `createMutation`
-   `useIsFetching`
-   `useIsMutating`
-   `useQueryClient`
-   `QueryClient`
-   `QueryClientProvider`

The adapter docs have been updated as well as examples have been added.

## Acknowledgments

The work on solid-query would not have been possible without

- @ardeora
- @lukesmurray
- @jennyckaplan

This is a follow-up to #4195 but rebased on `main` and with commits from `beta` removed per discussion with @TkDodo and @ardeora.

PR #4380, opened 2022-10-25
solid-query-persist-client: init at 4.13.0

milahu:
add persistent caching for solid-query

**based on**

- packages/react-query-persist-client
- packages/solid-query
  -  #3730 
  - #4211
- examples/react/offline

**todo**

fix tests?

example: remove `workspace:*` deps

**done**

make it build: `packages/solid-query-persist-client/build/` is not generated
&rarr; add `buildConfigs` block to rollup.config.ts

fix example: transitive dependencies are not installed by pnpm
vite: `✘ [ERROR] Could not resolve "@mswjs/cookies"`
dependency graph: @tanstack/query-example-solid-offline &rarr; msw &rarr; @mswjs/cookies
solved by removing `{ resolve: { preserveSymlinks: true } }` from vite.config.ts

fixme: solid build is not produced
packages/solid-query-persist-client/build/solid/
but is declared in tsconfig.json
fixme: solid jsx is compiled to React.createElement
packages/solid-query-persist-client/src/PersistQueryClientProvider.tsx
solved by adding
`{ "path": "packages/solid-query-persist-client" },`
to the root tsconfig.json

fixme: service worker in typescript is not compiled to javascript?
`A bad HTTP response code (404) was received when fetching the script.`
fixed by running `npx msw init .` to generate `mockServiceWorker.js` ([docs](https://mswjs.io/docs/getting-started/integrate/browser))

fixme: double fetching

```
rest-deps.js:82 [MSW] 21:58:24 GET /movies (200 OK)
rest-deps.js:82 [MSW] 21:58:24 GET /movies (200 OK)
```

fixed by query options `{ refetchOnMount: false }`

i use `unwrap` to remove solidjs Proxy objects from queryKey
before storing the query cache in indexeddb.
is this a lossy transformation?
or is it okay to restore the query cache without solidjs Proxy objects
&rarr; okay because we use `handleResult(observer.getCurrentResult())`
which calls `setState(unwrap(unwrappedResult))`
which restores the solidjs Proxy objects

fix examples/solid/offline

createQueries: add persistence

~~createMutation: add persistence~~
no. `useIsRestoring` is used only in
packages/react-query/src/useBaseQuery.ts
packages/react-query/src/useQueries.ts

**ping**

@lukesmurray


packages/solid-query-persist-client/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@tanstack/solid-query-persist-client",
+  "version": "4.13.0",
+  "description": "Solid bindings to work with persisters in TanStack/react-query",
```
TkDodo, line 4:
```suggestion
  "description": "Solid bindings to work with persisters in TanStack Query",
```

TkDodo (commented):
👍 
I think one thing that's missing is that `createQueries` would also need to listen to `isRestoring` - at least that's what we also have in react!

ardeora:
Hello! Thanks for working on the persister! This is looking very promising. One thing I noticed that is preventing the `solid` folder to be generated is because you need to add another reference in the root tsconfig.

```json
{ "path": "packages/solid-query-persist-client" },
```

milahu:
currently blocked by solidjs store functions, which cannot be stored in indexeddb

```
DOMException: Failed to execute 'put' on 'IDBObjectStore': function () { [native code] } could not be cloned.
```

caused by `client.clientState.queries[0].state.data.movie.comment` in `examples/solid/offline/src/persister.ts`

what would others do?
[solid-dexie](https://github.com/faassen/solid-dexie)
[createFetch().withCacheStorage()](https://github.com/solidjs-community/solid-primitives/tree/main/packages/fetch#modifiers)

ardeora:
Closing this in favor of #5858

PR #4397, opened 2022-10-28
Yet another test stability improvements PR

zorzysty:


packages/react-query/src/__tests__/useQuery.test.tsx
```diff
@@ -873,21 +873,33 @@ describe('useQuery', () => {
...
-    renderWithClient(queryClient, <Page />)
+    const rendered = renderWithClient(queryClient, <Page />)
 
-    await sleep(100)
+    await waitFor(() => {
+      rendered.getByText('data')
+    })
+
+    fireEvent.click(rendered.getByRole('button', { name: 'remove' }))
+
```
TkDodo, line 899: I think here we need a longer sleep (100) after the button click. As the title suggests:

> 'should not get into an infinite loop when removing a query with cacheTime 0 and rerendering'

How do we detect an infinite loop? We wait a bit and wait what the states length is. If it's still 5 after 100ms, it's not an infinite loop. If it were, it would be anywhere between 30 and 50 or so :)

zorzysty: [1e05f1e](https://github.com/TanStack/query/pull/4397/commits/1e05f1ef29020e5d0f77269377a4d9a253665538)

PR #4409, opened 2022-10-30
docs: prefer object syntax in docs

TkDodo:


dimitropoulos:
I like this change a lot, personally.  That said, from a "what react-query looks like to most people" perspective this seems like a big change (i.e. to point people in the object syntax direction).  I'm a bit new to this community so apologies if I missed something obvious before this PR, but I'm curious if you could say just a bit about why you've decided to move in this direction?

TkDodo:
Have a look at the v5 roadmap (under removing overloads)
https://github.com/TanStack/query/discussions/4252

Additionally, it will be helpful for solid-query

PR #4412, opened 2022-10-31
fix(env): Run eslint+jest with pnpm exec

IvanLisz:
Makes eslint + jest run on windows

Before:
image

After:
image


TkDodo:
why was this closed? I think the solution works and is better than what we currently have @IvanLisz

PR #4419, opened 2022-10-31
docs(guides): minor fixes and typos

dimitropoulos:
I'm doing a full read-through of the docs, and I noticed a few minor things along the way.  I'm very low-ego about these changes, but thought I'd submit them.  I'd be happy to pull each commit into separate PRs if you like, but each commit has a description and should be atomic.

Definitely more to come, but I thought I'd submit what I have for today.

### DIFF
--- a/docs/guides/background-fetching-indicators.md
+++ b/docs/guides/background-fetching-indicators.md
@@ -30,7 +30,7 @@ function Todos() {
 }
 ```
 
-# Displaying Global Background Fetching Loading State
+## Displaying Global Background Fetching Loading State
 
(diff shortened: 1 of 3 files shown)


docs/guides/background-fetching-indicators.md
```diff
@@ -30,7 +30,7 @@ function Todos() {
 }
 ```
 
-# Displaying Global Background Fetching Loading State
+## Displaying Global Background Fetching Loading State
```
dimitropoulos, line 33: | BEFORE | AFTER |
| - | - |
| Screenshot_20221031_155131 | Screenshot_20221031_155229

PR #4428, opened 2022-11-02
Stabilize another batch of flaky tests

zorzysty:


packages/react-query/src/__tests__/useInfiniteQuery.test.tsx
```diff
@@ -1725,13 +1724,13 @@ describe('useInfiniteQuery', () => {
       const promise = new Promise<string>((resolve, reject) => {
         cancelFn = jest.fn(() => reject('Cancelled'))
         signal?.addEventListener('abort', cancelFn)
-        sleep(20).then(() => resolve('OK'))
+        sleep(1000).then(() => resolve('OK'))
```
zorzysty, line 1727: This could be an arbitrarily long value, we're cancelling the query before it's resolved.

TkDodo (approved):

Issue #4459, opened 2022-11-04
eslint-plugin-query incorrectly thinks TS generic is dependency


mckelveygreg:
### Describe the bug

Love that you are putting out an eslint plugin!! It already found several issues in our codebase.

However, we have a custom hook where a generic can be passed along in order to type the response of an api call. The linter is incorrectly thinking that the TS generic should be a dependency. 

```
error  The following dependencies are missing in your queryKey: Result  @tanstack/query/exhaustive-deps
```

### Your minimal, reproducible example

https://stackblitz.com/edit/vitejs-vite-wgscv2?file=src/useGenericHook.ts

### Steps to reproduce

1. Open up terminal in stackblitz
2. Run `yarn lint`
3. Check out error in console from file `useGenericHook.ts`

### Expected behavior

I expect that types shouldn't be a dependency for a query key

### How often does this bug happen?

Every time

### Platform

- OS Mac ventura M1
- node 18.12.0

### react-query version

v4.14.3

### TypeScript version

4.8.4

TkDodo:
@Newbie012 could you have a look?

Newbie012:
Sure, I'll push a fix later this day.

PR #4462, opened 2022-11-05
fix(eslint-plugin): omit generics/types from dependencies

Newbie012:
fixes https://github.com/TanStack/query/issues/4459


TkDodo:
@all-contributors please add @Newbie012 for code.

PR #4465, opened 2022-11-06
Make isRefetching false when fetching next/previous page in an infinite query

zorzysty:
Rationale discussed under https://github.com/TanStack/query/discussions/3744

### DIFF
--- a/packages/query-core/src/infiniteQueryObserver.ts
+++ b/packages/query-core/src/infiniteQueryObserver.ts
@@ -140,7 +140,7 @@ export class InfiniteQueryObserver<
     const { state } = query
     const result = super.createResult(query, options)
 
-    const { isFetching, isLoading } = result
+    const { isFetching, isRefetching } = result
 
     const isFetchingNextPage =
       isFetching && state.fetchMeta?.fetchMore?.direction === 'forward'
@@ -157,7 +157,7 @@ export class InfiniteQueryObserver<
       isFetchingNextPage,
       isFetchingPreviousPage,
       isRefetching:
-       result.isRefetching &&
+        isRefetching &&
         !isFetchingNextPage &&
         !isFetchingPreviousPage,
     }
--- a/packages/query-core/src/infiniteQueryObserver.ts
+++ b/packages/query-core/src/infiniteQueryObserver.ts
@@ -157,9 +157,7 @@ export class InfiniteQueryObserver<
       isFetchingNextPage,
       isFetchingPreviousPage,
       isRefetching:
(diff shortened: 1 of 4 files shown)


TkDodo (commented):
thx. do the existing tests sufficiently cover the different behaviour?

zorzysty:
> thx. do the existing tests sufficiently cover the different behaviour?

I had to update update few tests to cover `isRefetching`.

packages/query-core/src/infiniteQueryObserver.ts
```diff
@@ -139,18 +139,28 @@ export class InfiniteQueryObserver<
...
-        state.fetchStatus === 'fetching' &&
-        state.fetchMeta?.fetchMore?.direction === 'forward',
-      isFetchingPreviousPage:
-        state.fetchStatus === 'fetching' &&
-        state.fetchMeta?.fetchMore?.direction === 'backward',
+      isFetchingNextPage,
+      isFetchingPreviousPage,
+      isRefetching:
+        isFetching &&
+        !isLoading &&
```
TkDodo, line 161: slight correction: let's use `result.isRefetching &&` here instead of duplicating `isFetching &6 !isLoading` - just in case the "base condition" from `super.createResult` changes

 ```suggestion
        result.isRefetching &&
```

File: packages/query-core/src/infiniteQueryObserver.ts [hunk 2]
```diff
@@ -139,18 +139,27 @@ export class InfiniteQueryObserver<
   ): InfiniteQueryObserverResult<TData, TError> {
     const { state } = query
     const result = super.createResult(query, options)
+
+    const { isFetching, isLoading } = result
```
TkDodo, line 143:
```suggestion
    const { isFetching, isRefetching } = result
```

packages/query-core/src/infiniteQueryObserver.ts
```diff
@@ -139,18 +139,27 @@ export class InfiniteQueryObserver<
...
-      isFetchingNextPage:
-        state.fetchStatus === 'fetching' &&
-        state.fetchMeta?.fetchMore?.direction === 'forward',
-      isFetchingPreviousPage:
-        state.fetchStatus === 'fetching' &&
-        state.fetchMeta?.fetchMore?.direction === 'backward',
+      isFetchingNextPage,
+      isFetchingPreviousPage,
+      isRefetching:
+       result.isRefetching &&
```
TkDodo, line 160:
```suggestion
        isRefetching &&
```

PR #4466, opened 2022-11-06
stabilize the test for notifying query cache when a query becomes stale

zorzysty:


packages/react-query/src/__tests__/useQuery.test.tsx
```diff
@@ -2231,7 +2228,28 @@ describe('useQuery', () => {
     // 7. Observer options updated
     // 8. Observer result updated -> stale
     // 9. Observer options updated
-    expect(fn).toHaveBeenCalledTimes(9)
```
TkDodo, line 2234: the comments about the "9 updates" is still here, but the assertion is gone 🤔 

I also don't know what the `states.push` are actually doing here. Nothing it seems.

I think this is one of the tests that should actually live in the query-core, not react-query. There is no nothing react specific about `should notify query cache when a query becomes stale`

we could do:

- subscribe to queryCache
- create an observer with a staleTime
- do the same checks there ...

this would also get rid of the test duplication in solidJs

TkDodo (changes requested):

packages/react-query/src/__tests__/useQuery.test.tsx
```diff
@@ -2231,7 +2228,28 @@ describe('useQuery', () => {
...
+          type: 'updated',
+          action: expect.objectContaining({ type: 'success' }),
+        }),
+      )
+    })
+
+    unsubscribe()
   })
 
   it('should not re-render when it should only re-render on data changes and the data did not change', async () => {
```
TkDodo, line 2255: this test is also failing now. It wasn't change so I think it's just flaky too:

```
packages/react-query test:jest: FAIL react-query src/__tests__/useQuery.test.tsx (31.7 s)
packages/react-query test:jest:   ● useQuery › should not re-render when it should only re-render only data change and the selected data did not change
packages/react-query test:jest:     expect(received).toBe(expected) // Object.is equality
packages/react-query test:jest:     Expected: 2
packages/react-query test:jest:     Received: 1
packages/react-query test:jest:       1038 |     await sleep(10)
packages/react-query test:jest:       1039 |
packages/react-query test:jest:     > 1040 |     expect(states.length).toBe(2)
packages/react-query test:jest:            |                           ^
packages/react-query test:jest:       1041 |     expect(states[0]).toMatchObject({ data: undefined })
packages/react-query test:jest:       1042 |     expect(states[1]).toMatchObject({ data: 'test' })
packages/react-query test:jest:       1043 |   })
packages/react-query test:jest:       at Object.<anonymous> (src/__tests__/useQuery.test.tsx:1040:27)
```

zorzysty: [2630a1f](https://github.com/TanStack/query/pull/4466/commits/2630a1f4d45e3b33df87cd266e7a0de7a6bc5e1c)

packages/react-query/src/__tests__/useQuery.test.tsx
zorzysty: All right, I'll give that a try

TkDodo (approved):

PR #4468, opened 2022-11-06
feat(devtools): add 'copy' button in devtools 

StefanDjokovic:
This PR adds a copy button in the devtools. This will appear next to objects in Data Explorer. 

Let me know if the design of this makes sense or how this can be improved.

Screenshot 2022-11-06 at 19 56 46

image


packages/react-query-devtools/src/Explorer.tsx
```diff
@@ -49,6 +59,11 @@ type ExpanderProps = {
   style?: React.CSSProperties
 }
 
+type CopierProps = {
+  onClick: () => {}
```
TkDodo, line 63:
```suggestion
  onClick: () => void
```

packages/react-query-devtools/src/Explorer.tsx
```diff
@@ -62,6 +77,20 @@ export const Expander = ({ expanded, style = {} }: ExpanderProps) => (
   </span>
 )
 
+export const Copier = ({ onClick, style = {} }: CopierProps) => (
+  <span onClick={onClick} style={{
+    display: 'span',
```
TkDodo, line 82: I don't think `display: 'span'` exists?

https://developer.mozilla.org/en-US/docs/Web/CSS/display

packages/react-query-devtools/src/Explorer.tsx
```diff
@@ -74,6 +103,7 @@ type RendererProps = {
   subEntryPages: Entry[][]
   type: string
   expanded: boolean
+  copiable: boolean
```
TkDodo, line 106: typo: should probably by `copyable`

packages/react-query-devtools/src/Explorer.tsx
```diff
@@ -124,6 +155,9 @@ export const DefaultRenderer: Renderer = ({
               {subEntries.length} {subEntries.length > 1 ? `items` : `item`}
             </Info>
           </ExpandButton>
+          {copiable ? (<CopyButton>
+            <Copier onClick={() => navigator.clipboard.writeText(displayValue(value))} />
+          </CopyButton>) : null}
```
TkDodo, line 160: why isn't the click handler on the `button` but on the `span` inside it?

please also give the button:

- a proper `aria-label`
- a `tooltip`

for a11y. Like we also do for the mock offline behaviour button:

https://github.com/TanStack/query/blob/1a1012bf901b99f703b4b219675767ec2bcb8049/packages/react-query-devtools/src/devtools.tsx#L647-L656

packages/react-query-devtools/src/Explorer.tsx
```diff
@@ -124,6 +155,9 @@ export const DefaultRenderer: Renderer = ({
               {subEntries.length} {subEntries.length > 1 ? `items` : `item`}
             </Info>
           </ExpandButton>
+          {copiable ? (<CopyButton>
+            <Copier onClick={() => navigator.clipboard.writeText(displayValue(value))} />
```
TkDodo, line 159: putting `displayValue` into the clipboard seems weird, because we use `superjson` to show non-serializable values. Why don't we just put the raw values that are really in the cache into the clipboard?

TkDodo (commented):
Nice work, thank you 🙏 

I think some sort of user confirmation that the copying worked would be nice. Maybe the button could turn into a green checkmark or so for a short period of time before it turns back into the button via an animation?

FYI, in the codesandbox iframe, it doesn't seem to work, which got me confused. Opening it in full view does work, but the ux is the same: you click and nothing happens...

also, please add some tests for the new feature, thank you 🙏

StefanDjokovic:
Thank you TkDodo for the feedback and thorough review 🙏 

Will be adding the confirmation, tests, and investigate why it's not working in the iframe. Thanks for pointing them out! :)

StefanDjokovic:
Looks like it might not be possible to add to clipboard from the iframe :/ 

`DOMException: The Clipboard API has been blocked because of a permissions policy applied to the current document. See https://goo.gl/EuHzyv for more details.`

TODOs left to address:
- [ ] Add better error visual for the user on copy error (probably a red X with the error on the console)
- [ ] Add tests
- [ ] Check if it's possible to add the raw value to the clipboard (right now I have changed to a JSON.stringify)
- [ ] Double-check the code before submitting it for review

StefanDjokovic:
Confirmation state added (same SVG as GitHub):

image

I'm still unsure about how to best handle what is being pasted to the clipboard. I used superjson to stringify the value because I couldn't find any way to copy the javascript object :/ Let me know if you have some pointer on how to do that!

To handle situations where the user can't copy to clipboard (e.g. permissions in iframe) I've added an error state that should hopefully make things clearer. 

image

The other solution that I have been thinking of is, instead of a copy-to-clipboard action, having it become a "print on console". That should help with the error state. Let me know your thoughts and any feedback/comments on the code :)

packages/react-query-devtools/src/Explorer.tsx
```diff
@@ -29,6 +30,55 @@ export const ExpandButton = styled('button', {
   padding: 0,
 })
 
+enum CopyState {
```
TkDodo, line 33: please don't use typescript enums. We prefer string literal types:

```
type CopyState = 'noCopy' | 'successCopy' | 'errorCopy'
```

packages/react-query-devtools/src/Explorer.tsx
```diff
@@ -29,6 +30,55 @@ export const ExpandButton = styled('button', {
...
+          },
+          (err) => {
+            console.error('Failed to copy: ', err)
+            setCopyState(CopyState.ErrorCopy)
+            setTimeout(() => {
+              setCopyState(CopyState.NoCopy)
+            }, 1500)
+          },
+        )}
+      }
```
TkDodo, line 60: looks good, but we're looking at a mini-state-machine here already. The user shouldn't be able to click the button while it is in the `SuccessCopy` or `ErrorCopy` state - only when it is in the `NoCopy` state. I would probably just attach the click handler when we are in that state.

packages/react-query-devtools/src/Explorer.tsx
```diff
@@ -29,6 +30,55 @@ export const ExpandButton = styled('button', {
...
+      onClick={() => {
+        navigator.clipboard.writeText(superjson.stringify(value)).then(
+          () => {
+            setCopyState(CopyState.SuccessCopy)
+            setTimeout(() => {
+              setCopyState(CopyState.NoCopy)
+            }, 1500)
+          },
+          (err) => {
+            console.error('Failed to copy: ', err)
```
TkDodo, line 53: not sure if we should console.error in production as well

packages/react-query-devtools/src/Explorer.tsx
```diff
@@ -62,6 +112,70 @@ export const Expander = ({ expanded, style = {} }: ExpanderProps) => (
   </span>
 )
 
+export const Copier = () => (
```
TkDodo, line 115: do these components need to be exported? Are they used outside this file?

StefanDjokovic:
Thanks, fixed the comments 🙏

StefanDjokovic:
Formatted the test file, sorry about that, didn't realise that 😅

TkDodo (approved):

TkDodo:
awesome work, thank you 🚀

TkDodo:
I can see some act warnings in the react17 tests - can you take a look at them maybe?

StefanDjokovic:
@TkDodo I wasn't able to reproduce the warnings or last Fail/Warnings locally.

I tried running npm run test:react:17 and they don't show up, is there anything else that needs to be done to have them? I saw the warnings before when I wasn't importing act from @testing-library/react, but they seem to be working fine now locally

PR #4477, opened 2022-11-07
fix(devtools): unset padding for resetting after closed

### LINKED ISSUE #4467: After closed devtools, padding of parent element remains
### Describe the bug

The padding of the parent element is reset with inline style when devtools is opened.
The inline style remains after closed devtools, so original padding specified with non-inline style of parent element is not applied.

image

https://codesandbox.io/s/async-sky-5ibueo?file=/index.html

### Steps to reproduce

1. Open devtools.
2. Close devtools.

### Expected behavior

No padding remains for resetting and the original padding is applied.

Every time

### Screenshots or Videos

_No response_

### Platform

All

### react-query version

4.14.5

### TypeScript version

_No response_

### Additional context

_No response_

minami-minami:
fixes #4467

### DIFF
--- a/packages/react-query-devtools/src/devtools.tsx
+++ b/packages/react-query-devtools/src/devtools.tsx
@@ -390,18 +390,7 @@ export function ReactQueryDevtools({
           <Logo aria-hidden />
           <ScreenReader text="Open React Query Devtools" />
         </button>
-      ) : (
-        <style
-          nonce={styleNonce}
-          dangerouslySetInnerHTML={{
-            __html: `
-          .ReactQueryDevtoolsParent.ReactQueryDevtoolsParent {
-            padding-bottom: 100px;
-          }
-        `,
-          }}
-        />
-      )}
+      ) : null}
     </Container>
   )
 }
--- a/packages/react-query-devtools/src/devtools.tsx
+++ b/packages/react-query-devtools/src/devtools.tsx
@@ -247,26 +247,37 @@ export function ReactQueryDevtools({
   }, [isResolvedOpen])
 
   React.useEffect(() => {
-    if (isResolvedOpen) {
-      const root = rootRef.current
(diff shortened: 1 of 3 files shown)


packages/react-query-devtools/src/devtools.tsx
```diff
@@ -277,6 +277,8 @@ export function ReactQueryDevtools({
         return () => {
           window.removeEventListener('resize', run)
           if (root?.parentElement && typeof previousValue === 'string') {
+            // unset padding for resetting
+            root.parentElement.style.padding = ''
```
TkDodo, line 281: what if the parentElement had a padding before? Should we not store the old padding and re-use that one?

tbh, I think meddling with the parentElement is not a good idea because it could seriously destroy the layout of the app. Maybe there is a way to achieve what we want _without_ setting some styles on the parent element?

minami-minami: You are right.
I will consider other approaches 👍

minami-minami: @TkDodo 
I think the reason for the padding on the parent element is to prevent the element from wrapping around behind the devtools.
It seems difficult to solve this without affecting the parent element, so I think it is a trade-off 😇

For now, we have responded to restore the original padding.

The only other thing I can think of is...
 - Adding a opt-out / opt-in flag
 - Giving up that prevent wrapping

Any good ideas?

TkDodo: reverting back to the previous style seems like a good idea 👍

TkDodo:
would it be possible to add a test for the new behaviour of restoring the old styles?

minami-minami:
@TkDodo  
Added test 👍

TkDodo (approved):

Issue #4475, opened 2022-11-07
eslint-plugin-query: exhaustive deps not reported when `queryKey` is an external const


TkDodo:
this code should fail, because `x` is not included in the dependencies, but it does not:

```
const x = 5
const queryKey = ['templatePreviews']

useQuery({ queryKey, queryFn: () => x })
```
⬆️  No eslint error here ❌ 

---

it is reported just fine if the queryKey is inlined:

```
const x = 5

useQuery({ queryKey: ['templatePreviews'], queryFn: () => x })
```

⬆️  `ESLint: The following dependencies are missing in your queryKey: x(@tanstack/query/exhaustive-deps)` ✅

TkDodo:
@Newbie012 FYI

TkDodo:
yep, definitely. I just left the feedback here when I tried to use the eslint-plugin on our codebase. It's just natural that new usage patterns arise and  it becomes more obvious which cases we want to / should handle :)

PR #4480, opened 2022-11-07
run exhaustive dep check on referenced query key

Newbie012:
fixes https://github.com/TanStack/query/issues/4475

image


Newbie012:
@TkDodo Something odd just happened. The Lint job has failed on eslint errors that occurred before I made this PR. I fixed them in https://github.com/TanStack/query/pull/4480/commits/f14f185249bf0d4e7d5190bf3e8ba2ce7db60b48.

TkDodo:
I merged the other PR first so now they are conflicts here

Newbie012:
There seems to be another flaky test.

About the issue I wrote above, it seems like I was wrong. I had 1 problem from this PR and a few *warnings* which won't fail in CI. What's the reasoning for not marking all rules as problems when running as CI?

TkDodo:
> About the issue I wrote above, it seems like I was wrong. I had 1 problem from this PR and a few warnings which won't fail in CI. What's the reasoning for not marking all rules as problems when running as CI?

eslint rules should be either errors or warnings, independent of environment. We can make some rules from warning to error if they are important to be fixed.

TkDodo (approved):

PR #4491, opened 2022-11-09
Update QueryClient.md

claasahl:
"logger" option is not present in types (v3.39.2)

### DIFF
--- a/docs/reference/QueryClient.md
+++ b/docs/reference/QueryClient.md
@@ -59,9 +59,6 @@ Its available methods are:
 - `mutationCache?: MutationCache`
   - Optional
   - The mutation cache this client is connected to.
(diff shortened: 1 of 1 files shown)


TkDodo:
but it is available in v4 and the docs are for v4

claasahl:
@TkDodo Thanks for the quick reply. Is there a stable version of v4 available?

https://www.npmjs.com/package/react-query?activeTab=versions suggests that v4 was not officially released yet. The latest version seems to be v3.39.2.

TkDodo:
it's `@tanstack/react-query` for almost half a year now!

https://tanstack.com/blog/announcing-tanstack-query-v4

claasahl:
@TkDodo thanks for the pointer 🙇

Issue #4476, opened 2022-11-07
eslint-plugin-query: false positive for factory functions that return an object


TkDodo:
```
const getQuery = () => ({
  queryKey: ['foo'],
  queryFn: () => Promise.resolve(5)
})

return useQuery(getQuery())
```

⬆️ `ESLint: Objects syntax for useQuery is preferred(@tanstack/query/prefer-query-object-syntax)`

what's more dangerous, it even auto fixes this to:

```
return useQuery({ queryKey: getQuery() })
```

which is wrong.

---

As a workaround, we have to create a new object and spread:

```
useQuery({ ...getQuery() })
```

---

The wrong auto-fix is a big problem I think because it will convert working code to non-working code :/

TkDodo:
@Newbie012 not sure if this easy to fix...

Newbie012:
I'll fix it

Newbie012:
I got it working, and many other complex cases (such as deep multiple returns).The code is currently unreadable, so I'll fix it and push a PR today or tomorrow.

PR #4497, opened 2022-11-11
fix(eslint-plugin): false positive for factory functions

Newbie012:
fixes https://github.com/TanStack/query/issues/4476

Honestly, I'm unsure how I feel about this fix. Since we need to support `@babel/eslint-parser`, we need to find a different way to get a function's "return type".

I'm not sure how I feel about my approach - The linter will check all possible return nodes, and if at least one node is not an `ObjectExpression`, the linter will report that node.

I'm open to suggestions.


TkDodo:
I think it's good 👍 . The only other thing would be to depend on typescript types, not sure if you think this would be generally better. But if, we'll do this for v5 only I guess...

Newbie012:
It would require much less code and probably a more precise check, but it will hurt the plugin's adoption, so I guess we'll stick with the current implementation for now.

PR #4502, opened 2022-11-13
test: add useQueries with suspense tests

dante01yoon:
this is my first pr in query. 

## changes
added tests
- shouldn't unmount before all promises fetched
- should suspend all queries in parallel - global configuration

### DIFF
--- a/packages/react-query/src/__tests__/suspense.test.tsx
+++ b/packages/react-query/src/__tests__/suspense.test.tsx
@@ -1205,7 +1205,6 @@ describe('useQueries with suspense', () => {
               await sleep(10)
               return '1'
             },
-            suspense: true,
           },
           {
             queryKey: key2,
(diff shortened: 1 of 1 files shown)


packages/react-query/src/__tests__/suspense.test.tsx
```diff
@@ -1121,4 +1121,120 @@ describe('useQueries with suspense', () => {
...
+            suspense: true,
+          },
+          {
+            queryKey: key2,
+            queryFn: async () => {
+              results.push('2')
+              await sleep(20)
+              return '2'
+            },
+            suspense: false,
```
TkDodo, line 1217: both queries manually set the `suspense` option, which overwrites the default setting. So the default options are not tested here ...

dante01yoon: I see. I'll remove that

PR #4513, opened 2022-11-15
fix(vue-query): export useQuery return type

juvirez:


Jarzka:
It seems that this change broke my build:

This worked previously:

```
  return useQuery({
    queryKey: ["queryId"],
    queryFn: () => {
      return getStuff(code)
    },
  });
```

Now I got an error:
```
TS2769: Object literal may only specify known properties, and 'queryKey' does not exist in type 'readonly unknown[]'. 
```

PR #4527, opened 2022-11-20
Use useMemo for the useMutation result

gromchen:
Otherwise, an infinite loop is very likely.
One may not use the result destructuring and provide the value to the `useEffect` dependencies.

```javascript
const mutation = useMutation(...)

useEffect(() => { ... }, mutation)
```

### DIFF
--- a/packages/react-query/src/useMutation.ts
+++ b/packages/react-query/src/useMutation.ts
@@ -118,7 +118,10 @@ export function useMutation<
     throw result.error
   }
 
-  return { ...result, mutate, mutateAsync: result.mutate }
+  return React.useMemo(
+    () => ({ ...result, mutate, mutateAsync: result.mutate }),
+    [mutate, result],
+  )
 }
 
 // eslint-disable-next-line @typescript-eslint/no-empty-function


TkDodo:
Why would you add the whole mutation to the dependency array? .mutate and .mutateAsync are stable...

gromchen:
That's how ESLint works.
Moreover, why `useMutation` result shouldn't be stable?
There is no way in React hooks to stop users from using the whole object as a hook result, so it's dangerous not to have it stable.

gromchen:
@TkDodo, what is a stopper here?

gromchen:
For example, I've just recently tried to destructure the result of `useQueryClient` for `setQueryData`. And I've got the error `TypeError: this is undefined`, so it's unclear when the result should or should not be destructured. Ideally, to minimize mistakes, both scenarios have to be supported.

TkDodo:
> That's how ESLint works.

This works just fine:

```
  const { mutate } = useMutation(() => Promise.resolve(5))

  React.useEffect(() => {
    mutate()
  }, [mutate])
```

I have to say I don't really understand why this violates the lint rule:

```
  const mutation = useMutation(() => Promise.resolve(5))

  React.useEffect(() => {
    mutation.mutate()
  }, [mutation.mutate])
```

This _should_ work imo - maybe a limitation ...

> For example, I've just recently tried to destructure the result of useQueryClient for setQueryData. And I've got the error TypeError: this is undefined, so it's unclear when the result should or should not be destructured. Ideally, to minimize mistakes, both scenarios have to be supported.

That is an entirely different issue. The `QueryClient` is a class - you can see that because you create it with `new QueryClient()`. Classes cannot be safely destructed. You can also try:

```
const { getDay } = new Date()
getDay()
```

and you'll get the same error.

> what is a stopper here?

I'm not convinced it solves the issue. Looking at the code, we have to depend on `result`, which comes from `getCurrentResult`. If the mutation is invoked, the result will likely change because it transitions to `isLoading: true`.

I've tried out the `useEffect` in the codesandbox preview build that has your changes, and it still runs the effect indefinitely, confirming my suspicion:

- https://codesandbox.io/s/tanstack-query-example-react-basic-typescript-forked-7l2js8?file=/src/index.tsx

You also haven't added a test that confirms that it now works....

gromchen:
>I've tried out the useEffect in the codesandbox preview build that has your changes, and it still runs the effect indefinitely, confirming my suspicion:
>https://codesandbox.io/s/tanstack-query-example-react-basic-typescript-forked-7l2js8?file=/src/index.tsx

Oh, that isn't good. It looks like the issue is more profound.
Let me close the PR, and I'm looking forward to using react query more confidently in future.

PR #4528, opened 2022-11-21
docs: Update Supported Query Syntax for Apollo Client V3 Reactive Variables 

zakuru:
Sorry if there is any confusion but it seems that In the V3 of Apollo client it is mentioned that all type of syntaxes are supported not only GQL thanks to Reactive Variable feature
https://www.apollographql.com/docs/react/local-state/reactive-variables/ 

`New in Apollo Client 3, reactive variables are a useful mechanism for representing local state outside of the Apollo Client cache. Because they're separate from the cache, reactive variables can store data of any type and structure, and you can interact with them anywhere in your application without using GraphQL syntax.`


zakuru:
Thinking more of it, this PR does not explain well the concept of Reactive Variables in the scope of that matrix
Please let me know
Thanks

PR #4537, opened 2022-11-22
docs: add react query suspense community contribution

kappys1:
Hello Tanner and TkDodo, here's the PR adding the Query React Suspense package to the Community section on Docs.

I tried to add an easy examples about how it works. Maybe it will change if it's required add some new feature like FallbackError but now it's not urgent because is possible to do it with error boundary.

it didn't generate a codesandbox but I created:
[example](https://codesandbox.io/s/react-query-suspense-ijcolz)

Please let me know if you'd like for me to make any changes.

### DIFF
--- a/docs/community/kappys-react-query-suspense.md
+++ b/docs/community/kappys-react-query-suspense.md
@@ -1,0 +1,86 @@
+---
+id: kappys-react-query-suspense
+title: React Query Suspense
+---
+
+Suspense your components until wherever call is completed just passing the query key.
+
+It's powerful because it isn't necessary pass isLoading props child by child or use the isLoading Hook. This does it by you
+
+## Installation
+You can install React Query Suspense via [NPM](https://www.npmjs.com/package/@kappys/react-query-suspense).
+
+```bash
+$ npm i @kappys/react-query-suspense
+# or
+$ pnpm add @kappys/react-query-suspense
+# or
+$ yarn add @kappys/react-query-suspense
+```
+
+## ⚡ Quick start
+
+Start wrapping the content that you want suspense until the call is ready to render.
+
+
+#### Simple way wrapping a component with `ReactQuerySuspense` waiting the example query key call.
+
+It's not necessary check isLoading or isSuccess, just pass by props the queryKey and forget all things.
+
+```ts
+import React from "react";
+import { ReactQuerySuspense, QueryKey } from '@kappys/react-query-suspense'
+
+export const SampleComponent: React.FC<React.PropsWithChildren<{
+  keys: QueryKey[];
+}>> = ({ children, keys }) => {
+
+  const key: QueryKey = ["example"];
+
+  return (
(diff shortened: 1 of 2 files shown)


TkDodo:
I've looked at the implementation but I don't currently feel comfortable promoting this lib via our docs. One urgent issue I am seeing is that an [Observer is created during render](https://github.com/kappys1/react-query-suspense/blob/02ec9d4b784e546eb8d108d7aac0ac413cb91c89/src/useReactQueryPromiseTracker.ts#L9-L12), which is not allowed in react as it's a side-effect, and the subscription is therefore also never cleaned up.

I also don't fully understand how this lib is working. I don't see that `suspense` is set to `true` anywhere, you can only pass a `queryKey` and no other options. It is mentioned that you don't need to check for `isLoading` or `isSuccess`, but on type level, data will still be potentially `undefined`. The examples also don't show how a query is used inside that `ReactQuerySuspense` boundary (which also doesn't add a react suspense boundary it seems).

I would encourage you to iterate a bit more with the idea and the api before we'll add it here.

kappys1:
@TkDodo I appreciate very much your comments! thanks!

You told me very useful feedback and very good improvements for this helper library

I'm working on this :)

TkDodo:
> Regarding the observer created during rendering

yeah sorry, I was actually wrong about that. creating the observer is fine, we also do it directly on useState:

https://github.com/TanStack/query/blob/9d9aea5fb12eb89dec54c619845b3d226b53cf2b/packages/react-query/src/useQueries.ts#L175-L177

It's subscribing to / unsubscribing from the observer that shouldn't happen during rendering - but this is not happening _at all_ in your code. I'm really having trouble understanding how this lib works. Maybe you can show how you're using it or show a test that shows this :)

kappys1:
Actually, I'm not unsubscribing the observer but I'm seeing that you neither are doing in `useQueries.ts`. Anyway I'm going to investigate how do it.
Also, I will explain you how it works better in some examples but the meaning of this library is easy,
It just was made to easy the way to suspense when someone are making a call and is loading without pass the `isLoading` attribute that you can get from `useQuery`. 
what is the goal here? it's not necesary has the `useQuery` in the same componet, pass parent-childen, redux, context, etc... because only adding the `queryKey` or multiple of them if you want wait more calls, you can do it. 
I know that it's possible if you use the observeQueries that you provide, but this one just a helper to do it easy.

To do it we have: 
- `useReactQueryPromiseTracker.ts `: this is the core: here we observe the calls status and provide if is loading or not. In the future I will provide more features like if is error, done, etc. 

- `useReactQuerySuspense.ts`: this one is the hook to consume the `useReactQueryPromiseTracker` where control if you want defer the call, etc etc.

- `reactQuerySuspense.tsx`: here you have the Component where you can use to suspense directly and add the Fallback meanwhile you a calling some api endpoint. this one consume directly `useReactQuerySuspense`.

In this library we provide the Tracker, hook and component to make the possibility to use as the developer wants, maybe wants to use directly the component or maybe use the hooks and do some logic in their component.

PR #4544, opened 2022-11-23
feat: queryClient.ensureQueryData

zorzysty:
Shorthand for calling `queryClient.getQueryData(...) ?? queryClient.fetchQuery(...)`


packages/query-core/src/queryClient.ts
```diff
@@ -118,6 +118,21 @@ export class QueryClient {
...
   }
 
+  ensureQueryData<
+    TQueryFnData = unknown,
+    TQueryKey extends QueryKey = QueryKey,
+  >(
+    queryKey: TQueryKey,
+    queryFn: QueryFunction<TQueryFnData, TQueryKey>,
+    filters?: QueryFilters,
+  ): Promise<TQueryFnData> {
```
TkDodo, line 128: I think we need the same interface here as `fetchQuery`. That means:

- 3 overloads (we'll remove them in v5)
  - if you want to do me a favour, you'll write the tests with the object-only syntax :)
- the same interface as `fetchQuery`: 4 generics and `FetchQueryOptions`
- we probably don't need the `filters` on `getQueryData`. they are only needed so that you can opt-out of exact filtering and add some other fuzzy filter options instead. It doesn't really fit the use-case for `ensureQueryData`.

zorzysty: Oh, I thought it was suppose to have a signature based on `getQueryData`. I'll change it, no problem.

docs/reference/QueryClient.md
```diff
@@ -178,6 +179,22 @@ const data = queryClient.getQueryData({ queryKey })
...
+
+`ensureQueryData` is an asynchronous function that can be used to get an existing query's cached data. If the query does not exist, `queryClient.fetchQuery` will be called and its results returned.
+
+```tsx
+const data = queryClient.ensureQueryData({ queryKey, queryFn })
+```
+
+**Options**
+
+- The options for `ensureQueryData` are exactly the same as those of [`fetchQuery`](#queryclientfetchquery).
```
TkDodo, line 192: except that `queryKey` is required maybe?

packages/query-core/src/queryClient.ts
```diff
@@ -118,6 +119,61 @@ export class QueryClient {
...
+  ): Promise<TData>
+  ensureQueryData<
+    TQueryFnData = unknown,
+    TError = unknown,
+    TData = TQueryFnData,
+    TQueryKey extends QueryKey = QueryKey,
+  >(
+    queryKey: TQueryKey,
+    queryFn: QueryFunction<TQueryFnData, TQueryKey>,
+    options?: FetchQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
```
TkDodo, line 150: in the overloads with multiple parameters, we could `omit` the extra parameters from the `options` because it doesn't make sense to pass `queryKey` or `queryFn` in `options` if we have them as a separate parameter.

I did that for `useQuery`, but for nothing else. Maybe it could be a separate PR. Or maybe it doesn't matter because it will be simplified with v5 anyways 😅

docs/reference/QueryClient.md
zorzysty: [b0537ed](https://github.com/TanStack/query/pull/4544/commits/b0537ed35af9dcdbb08f0765cb3bbc67499048f4)

packages/query-core/src/queryClient.ts
zorzysty: [5abe28e](https://github.com/TanStack/query/pull/4544/commits/5abe28e4c78fcf5a66aaa63dc73a6783f48cb809)
I don't think it can be changed for `arg2` and `arg3` without touching the implementation of `parseQueryArgs`. But for the second and third overload it seems to be working fine.

packages/query-core/src/queryClient.ts
```diff
@@ -118,6 +119,67 @@ export class QueryClient {
...
+    TQueryFnData = unknown,
+    TError = unknown,
+    TData = TQueryFnData,
+    TQueryKey extends QueryKey = QueryKey,
+  >(
+    queryKey: TQueryKey,
+    queryFn: QueryFunction<TQueryFnData, TQueryKey>,
+    options?: Omit<
+      FetchQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+      'queryKey'
```
TkDodo, line 155:
```suggestion
      'queryKey' | 'queryFn'
```

tannerlinsley (approved):

XpamAmAdEuS:
It would be great if we also have same shorthand for prefetchQuery.

TkDodo:
> It would be great if we also have same shorthand for prefetchQuery.

add `.catch(noop)` to it and you got it 🤷‍♂️

PR #4545, opened 2022-11-23
docs: fix docs for queryClient.getQueryData

zorzysty:


docs/reference/QueryClient.md
```diff
@@ -166,7 +166,7 @@ The options for `prefetchInfiniteQuery` are exactly the same as those of [`fetch
 `getQueryData` is a synchronous function that can be used to get an existing query's cached data. If the query does not exist, `undefined` will be returned.
 
 ```tsx
-const data = queryClient.getQueryData({ queryKey })
+const data = queryClient.getQueryData(queryKey)
```
TkDodo, line 169: this is on purpose. `getQueryData` has an overload that allows to pass in `filters` only, and `queryKey` is part of the filters.

I've updated the docs to use the object syntax only in preparation for v5, where we plan to remove all overloads:

- https://github.com/TanStack/query/discussions/4252

zorzysty: I can't see any overloads here in the main branch:
https://github.com/TanStack/query/blob/11f08c37d56c8aa5d9b28d8422842bc1a147c487/packages/query-core/src/queryClient.ts#L112-L121

And there is an actual TS error when I try to use the syntax described in the docs:
obraz

Am I missing something? I know it's going to be changed in v5, but currently docs should be valid for v4, right?

zorzysty: If I'm right and the overload doesn't exist, I'd be happy to add it :)

TkDodo: you're right, the overload is only for `getQueriesData`. There, it makes sense because you can fuzzily match. Not sure if we actually need the `filters` for `getQueryData`. I wouldn't add an overload just now. It will depend on what api we want for v5, it could be `queryClient.getQueryData(queryKey)` or `queryClient.getQueryData(filters)`, where `filters` contains the key.

I'll have to think about it so for now, fixing the docs is fine. Thanks for pointing this out.

PR #4557, opened 2022-11-26
docs: update some things referring to old react-query

henribru:


README.md
```diff
@@ -1,32 +1,32 @@
...
 </a><a href="https://discord.com/invite/WrRKjPJ" target="\_parent">
   <img alt="" src="https://img.shields.io/badge/Discord-TanStack-%235865F2" />
-</a><a href="https://github.com/tannerlinsley/react-query/actions?query=workflow%3A%22react-query+tests%22">
-<img src="https://github.com/tannerlinsley/react-query/workflows/react-query%20tests/badge.svg" />
-</a><a href="https://npmjs.com/package/react-query" target="\_parent">
-  <img alt="" src="https://img.shields.io/npm/dm/react-query.svg" />
+</a><a href="https://github.com/TanStack/query/actions?query=workflow%3A%22react-query+tests%22">
+<img src="https://github.com/TanStack/query/workflows/react-query%20tests/badge.svg" />
+</a><a href="https://www.npmjs.com/package/@tanstack/react-query" target="\_parent">
+  <img alt="" src="https://img.shields.io/npm/dm/@tanstack/react-query.svg" />
```
henribru, line 14: Maybe this should be kept for now just because the react-query package still has a lot more downloads?

henribru: Also not sure if it would be good to add shields for the other packages as well or if it would get too messy

TkDodo: I think it's fine. The numbers will align sooner or later :)

We can use one shield for the `query-core`, because each adapter implicitly uses that one, and if people do js-only, they will also use the core. The angular adapater (external) also depends on the query-core, so I think it will show the most accurate usage of the lib:

- https://img.shields.io/npm/dm/@tanstack/query-core.svg

numbers are pretty aligned with the ones from react-query, but this might change in the future :)

README.md
```diff
@@ -1,32 +1,32 @@
...
+  <img alt="" src="https://img.shields.io/github/stars/TanStack/query.svg?style=social&label=Star" />
 </a><a href="https://twitter.com/tannerlinsley" target="\_parent">
   <img alt="" src="https://img.shields.io/twitter/follow/tannerlinsley.svg?style=social&label=Follow" />
 </a> <a href="https://gitpod.io/from-referrer/">
   <img src="https://img.shields.io/badge/Gitpod-Ready--to--Code-blue?logo=gitpod" alt="Gitpod Ready-to-Code"/>
 </a>
 
-Enjoy this library? Try the entire [TanStack](https://tanstack.com)! [React Table](https://github.com/tannerlinsley/react-table), [React Form](https://github.com/tannerlinsley/react-form), [React Charts](https://github.com/tannerlinsley/react-charts)
+Enjoy this library? Try the entire [TanStack](https://tanstack.com)! [TanStack Table](https://github.com/TanStack/table), [React Form](https://github.com/tannerlinsley/react-form), [React Charts](https://github.com/tannerlinsley/react-charts)
```
TkDodo, line 29: let's do TanStack Table, Router, Virtual, React Charts and React Ranger here. Those are the ones listed on the main page as well: https://tanstack.com/

henribru: Sounds good, fixed in https://github.com/TanStack/query/pull/4557/commits/8063d340861cacb26b2c3009132047970f1b3fab

README.md
henribru: Sure, I can change it to `query-core`. Should I change the bundle size badge too?

TkDodo: > Sure, I can change it to query-core

👍 

> Should I change the bundle size badge too?

That's an interesting question. The bundle size for the query-core is smaller because it's missing the adapter specific logic. I wouldn't want that people feel deceived by that...

henribru: Yeah, fair point. The adapters aren't all the same size, but since it's not a huge difference using the React adapter is probably fine. Especially since it's by far the most used one anyway

henribru: Updated the downloads badge in https://github.com/TanStack/query/pull/4557/commits/b61010d74030a09aefc2d730643b7e933e731346

TkDodo (approved):

PR #4561, opened 2022-11-26
chore(react-query-devtools): upgrade match-sorter-utils

KevinVandy:
match-sorter-utils should no longer have the bad transpilation problem anymore.


TkDodo:
thanks. You have to please run `pnpm install` so that we get an updated lock file :)

KevinVandy:
> thanks. You have to please run `pnpm install` so that we get an updated lock file :)

whoops there we go

Issue #4556, opened 2022-11-26
Vue reactive call no longer preserves union type of vue-query's useQuery


henribru:
### Describe the bug

In the old, separate vue-query package, using vue's `reactive` on the return of `useQuery` preserved the union return type, but now all the union types seem to get combined into one type. This means you can't narrow types based on e.g. `isSuccess`. 

### Your minimal, reproducible example

I attempted to reproduce this in the Typescript Playground, but for whatever reason it just shows me `any` for the return type of the old `useQuery` ... https://www.typescriptlang.org/play?ts=4.8.4#code/JYWwDg9gTgLgBAbzlApgQwMY2ANxXAXzgDMoIQ4AiHAVxUoG4AoUSWROGgZxQEU6oAT0IkyFAOS0UAWgCOAweOato8JNz4K4aLpx78UQgHIoA7iNLk44gAIw0AOy72MAawD0UuQqVMmxGgcsYAgHEggIAAoASkQmOATkFBgaKDCABTFgHgA6VC4IABs8SMpiCMpopgI-DFDnOHlDYQBeJMxsEo0DIUiAbTKKgF0AGnCIaOjmJqEcgBM0exZiOEiZwRzsgGUaDAwULi5YhHjE9fnFtGra+vh1k3M21A7cFEjuhQf+wYhKUfHJtMFBclsAVmtPmZNlwdnsDkc4olGpDTCCrgQgA

I've made a local reproduction you can clone instead: https://github.com/henribru/vue-query-use-query-reactive/blob/master/src/main.ts

### Steps to reproduce

1. Open up src/main.ts in e.g. VS Code
2. Hover over `query.data` and notice it's been narrowed to string
3. Hover over `queryNew.data` and notice it hasn't been narrowed and is `string | undefined`
4. Hover over `query` and notice it's a union type
5. Hover over `queryNew` and notice it's not a union type

### Expected behavior

I expected `@tanstack/vue-query` to have the same behavior as `vue-query` so that I can discriminate between union types.

### How often does this bug happen?

Every time

### Screenshots or Videos

![image](https://user-images.githubusercontent.com/6639509/204086009-10a67754-55c5-404e-a1b9-c9653910dffb.png)
-
![image](https://user-images.githubusercontent.com/6639509/204086018-e4e355a2-c3cd-47ff-8d9a-399849772337.png)
-
[...]

PR #4562, opened 2022-11-26
fix(vue-query): use distributive omit to preserve union

henribru:
Fixes https://github.com/TanStack/query/issues/4556

Not sure if the tests are okay. I tried to base them on the type tests in the react-query package


DamianOsipiuk (approved):
Nice one! 🚀

PR #4564, opened 2022-11-28
docs(examples/react): fix prefetch pagination logic when trying to quickly get to the last page

wojciechSlodziak:
In the useQuery call we set "keepPreviousData", so we must make sure that for fetching the next page we don't base our logic on the data from potentially two pages away.
If user would click very fast to get to the last page we would potentially load the eleventh page and that could result in an error depending on the backend implementation.

### DIFF
--- a/examples/react/pagination/pages/index.js
+++ b/examples/react/pagination/pages/index.js
@@ -40,7 +40,7 @@ function Example() {
         fetchProjects(page + 1),
       )
     }
-  }, [data, page, queryClient])
+  }, [data, isPreviousData, page, queryClient])
 
   return (
     <div>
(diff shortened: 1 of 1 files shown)


examples/react/pagination/pages/index.js
```diff
@@ -35,7 +35,7 @@ function Example() {
 
   // Prefetch the next page!
   React.useEffect(() => {
-    if (data?.hasMore) {
+    if (!isPreviousData && data?.hasMore) {
```
TkDodo, line 38: if we do that, we need to add `isPreviousData` to the effect's dependency array.

wojciechSlodziak: I will update my PR, I am still new to this.

TkDodo (approved):

PR #4566, opened 2022-11-28
docs: Adds Using Remix to SSR guide.

plpaquin:
Hello,

I was setting up @tanstack/query in a Remix project and as a first time Query user I had to adapt the Next.js guide to my use case. I had to write a simple utility Hook to make it work so I thought I'd publish that as a NPM package and share my findings here.

Hope this helps,

Have a nice day.


TkDodo (commented):
❤️

PR #4574, opened 2022-11-30
docs: update code block example

ben-styling:
Just a small typo fix

### DIFF
--- a/docs/eslint/eslint-plugin-query.md
+++ b/docs/eslint/eslint-plugin-query.md
@@ -23,7 +23,7 @@ Add `@tanstack/eslint-plugin-query` to the plugins section of your `.eslintrc` c
 
 ```json
 {
-  "plugins": ["@tanstack/query"]
+  "plugins": ["@tanstack/eslint-plugin-query"]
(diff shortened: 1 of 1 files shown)


docs/eslint/eslint-plugin-query.md
```diff
@@ -23,7 +23,7 @@ Add `@tanstack/eslint-plugin-query` to the plugins section of your `.eslintrc` c
 
 ```json
 {
-  "plugins": ["@tanstack/query"]
+  "plugins": ["@tanstack/eslint-plugin-query"]
```
TkDodo, line 26: I think this is correct as it is. You can leave out the `eslint-plugin` prefix when adding plugins 🤷 

@Newbie012 ?

Newbie012: Correct. I believe there are examples that are using the plugin according to the current docs.

ben-styling: Sorry, you're right, I'm confusing plugins with extends 😅

PR #4582, opened 2022-12-01
fix: SSR Next.js typescript examples in Javascript

ArthurPedroti:
I update the _app.tsx and pages/posts.tsx example.

In _app.tsx examples, we have the AppProps and we need to pass a generic type for pageProps. I don't know if this is the best approach or if there's another type to use here.

Before:
image

After:
image

And in the pages/posts.tsx I only updated the legend of the file name, I think there's nothing else to correct.
image

If have a better type to use in AppProps, please show me so I can fix it. Thanks

### DIFF
--- a/docs/guides/ssr.md
+++ b/docs/guides/ssr.md
@@ -58,10 +58,10 @@ To support caching queries on the server and set up hydration:
 - Wrap your app component with `<Hydrate>` and pass it the `dehydratedState` prop from `pageProps`
 
 ```tsx
-// _app.jsx
+// _app.tsx
 import { Hydrate, QueryClient, QueryClientProvider } from '@tanstack/react-query'
 
-export default function MyApp({ Component, pageProps }) {
+export default function MyApp({ Component, pageProps }: AppProps<any>) {
   const [queryClient] = React.useState(() => new QueryClient())
 
   return (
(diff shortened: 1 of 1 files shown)


docs/guides/ssr.md
```diff
@@ -58,10 +58,10 @@ To support caching queries on the server and set up hydration:
 - Wrap your app component with `<Hydrate>` and pass it the `dehydratedState` prop from `pageProps`
 
 ```tsx
-// _app.jsx
+// _app.tsx
 import { Hydrate, QueryClient, QueryClientProvider } from '@tanstack/react-query'
 
-export default function MyApp({ Component, pageProps }) {
+export default function MyApp({ Component, pageProps }: AppProps<any>) {
```
TkDodo, line 64:
```suggestion
export default function MyApp({ Component, pageProps }: AppProps<{ dehydratedState: unknown }>) {
```

would work, but I'm not sure if we'd want to have docs in TS for just one page. If we do this, it should probably be for all examples of the documentation at the same time. This is also not specific to react-query - it's a nextjs typings "problem".

TkDodo:
thanks for the PR, but I'll close this for now for the reasons mentioned in the comment.

PR #4585, opened 2022-12-02
docs(useQuery): remove reference to `setQueryData` from onSuccess doc

jmlee2k:
The docs no longer state that `onSuccess` is called after `setQueryData` (which isn't true as of #2969).


jmlee2k:
Thanks for the work on this amazing library!

Issue #4584, opened 2022-12-02
refetchInterval not working


Willham12:
### Describe the bug

refetchInterval not working with 4.19.0, with 4.18.0 everything is fine

### Your minimal, reproducible example

https://github.com/TanStack/query/issues/new

### Steps to reproduce

```js
		useQuery([QueryKeys.job], job, {
			refetchInterval: 2000,
			enabled: paymentLinkResource !== undefined,
			onSuccess: () => { console.log("log")
			}});
```

### Expected behavior

refetch every 200ms

### How often does this bug happen?

Every time

### Platform

iOS

### Tanstack Query version

4.19.0

### TypeScript version

3.8

### Additional context

...

TkDodo:
provide a reproduction please

johnanthonyeletto:
Same issue here in React Native using 4.19.0. This query doesn't work:
```js
const { data } = useQuery(
        ['query-key'],
        () => {
           console.log("runQuery"); 
            return 1;
       },
        {
            refetchInterval: 1000,
        }
   );
```

TkDodo:
what does `typeof document` yield in react-native please? This is the only change we made:

- https://github.com/TanStack/query/commit/34e0ad239ba55b3aeb374c773c19d5b9094507a7

TkDodo:
I tried to create a basic reproduction in expo snack:
- https://snack.expo.dev/@tkdodo/greedy-yogurt

it works fine in the web view, but as soon as I try to create an ios view, I'm getting:


```
App.js (1235:889)
Unable to resolve module 'module://react-dom.js'
(Device)
  Evaluating module://react-dom.js
  Evaluating module://@tanstack/react-query.js
  Evaluating module://App.js.js
  Loading module://App.js
```

I don't know enough about react-native or expo snack to understand this error. Maybe someone else can take it from here?

Willham12:
Thanks for the quick fix ;D

PR #4587, opened 2022-12-02
add `eslint-plugin-no-typeof-window-undefined`

nirtamir2:
This PR adds [`eslint-plugin-no-typeof-window-undefined`](https://github.com/nirtamir2/eslint-plugin-no-typeof-window-undefined#readme) eslint plugin to prevent `typeof window === "undefined"` because Deno supports window (so this won't check well if the code is in browser / server-side) and replace it with `typeof document === "undefined"`

I run `pnpm test:eslint --fix` and it auto-fix some places the code uses `type window === "undefined"`, so make sure the conversion in those places works for you.

I'm following this tread https://twitter.com/TkDodo/status/1598227778435948545

## Notice
I run test and one of the tests (`should not use window to get the context when contextSharing is true and window does not exist`) failed, so I fix it by simulating `undefined` `document` (like another place in code does).
I found 2 more test cases with the same situation but they pass. (look for `// Mock a non web browser environment` in code) I'm not sure if it's better to use the same approach with deleting the document from `globalThis` and return it in the end of the test. Currently I did not touch them.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -6,6 +6,7 @@
     "plugin:@typescript-eslint/eslint-recommended",
     "plugin:@typescript-eslint/recommended",
     "plugin:import/typescript",
+    "plugin:no-typeof-window-undefined/recommended",
     "react-app",
     "prettier"
   ],
--- a/packages/react-query/src/__tests__/QueryClientProvider.test.tsx
+++ b/packages/react-query/src/__tests__/QueryClientProvider.test.tsx
@@ -246,6 +246,11 @@ describe('QueryClientProvider', () => {
     })
 
     test('should not use window to get the context when contextSharing is true and window does not exist', () => {
+      const { document } = globalThis
+
+      // @ts-expect-error
+      delete globalThis.document
+
       const queryCache = new QueryCache()
       const queryClient = createQueryClient({ queryCache })
 
@@ -269,6 +274,7 @@ describe('QueryClientProvider', () => {
 
       expect(queryClientFromHook).toEqual(queryClient)
 
+      globalThis.document = document
       windowSpy.mockRestore()
     })
   })
--- a/docs/guides/window-focus-refetching.md
+++ b/docs/guides/window-focus-refetching.md
@@ -39,7 +39,7 @@ In rare circumstances, you may want to manage your own window focus events that 
 ```tsx
 focusManager.setEventListener(handleFocus => {
   // Listen to visibilitychange and focus
-  if (typeof window !== 'undefined' && window.addEventListener) {
+  if (typeof document !== 'undefined' && window.addEventListener) {
     window.addEventListener('visibilitychange', handleFocus, false)
     window.addEventListener('focus', handleFocus, false)
   }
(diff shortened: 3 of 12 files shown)


.eslintrc
```diff
@@ -6,6 +6,7 @@
     "plugin:@typescript-eslint/eslint-recommended",
     "plugin:@typescript-eslint/recommended",
     "plugin:import/typescript",
+    "plugin:no-typeof-window-undefined/recommended",
```
nirtamir2, line 9: Not sure if it's better to specify this plugin config (with a single rule) or specify it in plugins and enable this rule specifically.

packages/react-query/src/__tests__/QueryClientProvider.test.tsx
```diff
@@ -246,6 +246,11 @@ describe('QueryClientProvider', () => {
     })
 
     test('should not use window to get the context when contextSharing is true and window does not exist', () => {
```
nirtamir2, line 248: This test failed after my changes with 
```
 TypeError: Cannot read properties of undefined (reading 'ReactQueryClientContext')
```

TkDodo:
I don't think the changes made by this eslint rule are correct here. Why should we check for the existence of `document` if we later access things on `window.`, for example:

```
if (typeof document !== 'undefined') {
  return window.matchMedia(query).matches
}
```

reading that code makes no sense to me 😅

nirtamir2:
Hi,
Thanks for the quick response.
I created this eslint plugin because I thought we could not depend on window existence to check if we are in a browser environment.
It seems like it's a thing we have to get used to, although less intuitive.

From a TypeScript perspective, the check for `typeof window !== "undefined"` narrow the window type, if it could be undefined.
But window's type is `Window & globalThis` and **cannot be** `undefined` in this scope regardless of the check.


> Why should we check for the existence of document if we later access things on window.

Because I believe the intent is to check if we are in a browser environment. It won't narrow window type in this case. If so, checking `typeof document === "undefined"` would be more correct (in environments like Deno).

What do you think is the best approach to handle this case? I can:
- Revert the code changes and run the lint rule in specific places (with eslint's overrides)
- Revert the code changes and write `// eslint-diable-next-line eslint-plugin-no-typeof-window-undefined` comments
- Use `isServer` from `@tanstack/query-core` or a similar function/variable that just encapsulates the `typeof document !== "undefined"`

What do you think?

TkDodo:
I think for code in the devtools, we can simply remove the `window` check. devtools don't render on the server, and the code in useEffect is not executed on the server at all.

However, we are getting some bug reports that intervals are not running in react-native since we merged this. Are you sure that `typeof document` returns `true` in react native environments?

see:

- https://github.com/TanStack/query/issues/4584

nirtamir2:
For react native:
`typeof document === "undefined"` return true
`typeof window === "undefined"` return false

TkDodo:
What would be a good way to check if you're in a react native browser environment other than checkout for window? Server environments like deno have window defined, but not document...

TkDodo:
gonna close this because of the react-native issue:
- https://github.com/TanStack/query/pull/4592

I don't think checking for `document` is a good idea anymore :)

Issue #3741, opened 2022-06-23
Query keys with object property set to undefined are not considered equal to missing property during invalidation


chekrd:
### Describe the bug

Query keys with object property set to undefined are not considered equal to missing property. Based on the docs, it should be the same: https://react-query.tanstack.com/guides/query-keys#query-keys-are-hashed-deterministically

### Your minimal, reproducible example

https://codesandbox.io/s/react-query-trz8sx?file=/src/App.tsx

### Steps to reproduce

1. Open the codesandbox link
2. Show codesandbox console
3. There are two queries rendered, the first fetch all items, and the second fetch filtered ones.
4. You can see two buttons with invalidation hooks. The first one invalidates both queries, the second one invalidates only non-filtered results. See the console for refetch logs.

### Expected behavior

I expect both keys should invalidate the same records no matter of filter: undefined as mentioned in react-query docs.

### How often does this bug happen?

Every time

### Platform

macOS v12.4
Chrome 103

### react-query version

Both v3.39.1 and v4.0.0-beta.23

### TypeScript version

v4.7.3

TkDodo:
query keys _are_ hashed deterministically, but when comparing queries, there is no hashing going on, because we have to partially compare them. For that, we use `Object.keys` to iterate over the objects. If you explicitly pass `filter:undefined`, we  cannot invalidate a query that has `filter: { done: true }`, why should we? That filter clearly doesn't match `filter: undefined`.

so yeah, for partial matching, there is a difference between a property being _present_ (independent of the value, even if it's _undefined_), and a property being _missing_.

RichardJuchelka:
Thanks for the quick response @TkDodo.
Does it make sense to differentiate between no prop and `undefined` prop? It still seems to be an issue to me. I have to filter out the `filter` prop in the key factory if not set. I can imagine another dev is not going to cover this for different entities. It can introduce stale cache errors (as I just found out). In my example you can see I didn't set `filter:undefined` explicitly, I just passed down the optional argument. I don't think creating a specific key is the right way. It would introduce copy-pasted hook for the same resource, with the required `filter` argument. 
What do you recommend in this situation?

TkDodo:
I would probably create two different functions to build keys that compose each other:

```
const todoListKeys = {
  list: () =>
    [
      {
        entity: "todos",
        scope: "list"
      }
    ] as const,
  filteredList: (filter: Filter) =>
    [{ ...todoListKeys.list()[0], filter }] as const
};
```

[...]

PR #4616, opened 2022-12-09
test(queryCache): add failing test for partially undefined key (#3741)

maxandron:
The current behavior when using functions such as invalidateQueries or setQueriesData is unexpected when objects with undefined properties are involved.

A key containing an object with an undefined property is hashed without the undefined property, yet the property is considered in the partialDeepEqual function.

This creates some confusing scenarios, as demonstrated in the discussion on TkDodo's blog
<https://github.com/TkDodo/blog-comments/discussions/71#discussioncomment-4348406> and in the referenced issue

This commit includes a failing test to demonstrate the expected behavior

### DIFF
--- a/packages/query-core/src/tests/queryCache.test.tsx
+++ b/packages/query-core/src/tests/queryCache.test.tsx
@@ -206,8 +206,13 @@ describe('queryCache', () => {
 
     test('should return all the queries when key contains object with an undefined property', async () => {
       const baseKey = queryKey()
-      await queryClient.prefetchQuery([...baseKey, { a: 1 }], () => 'data1')
-      expect(queryCache.findAll([...baseKey, { a: undefined }])).toHaveLength(1)
+      const createKey = (filter?: string) => [{ ...baseKey, filter }]
+      await queryClient.prefetchQuery(
+        createKey('filter1'),
+        () => 'filtered-data',
+      )
+      await queryClient.prefetchQuery(createKey(), () => 'all-data')
+      expect(queryCache.findAll(createKey())).toHaveLength(2)
     })
   })
 
--- a/packages/query-core/src/tests/queryCache.test.tsx
+++ b/packages/query-core/src/tests/queryCache.test.tsx
@@ -203,6 +203,12 @@ describe('queryCache', () => {
       await queryClient.prefetchQuery(key2, () => 'data2')
       expect(queryCache.findAll().length).toBe(2)
     })
+
+    test('should return all the queries when key contains object with an undefined property', async () => {
+      const baseKey = queryKey()
+      await queryClient.prefetchQuery([...baseKey, { a: 1 }], () => 'data1')
+      expect(queryCache.findAll([...baseKey, { a: undefined }])).toHaveLength(1)
+    })
   })
 
   describe('QueryCacheConfig.onError', () => {


packages/query-core/src/tests/queryCache.test.tsx
```diff
@@ -203,6 +203,12 @@ describe('queryCache', () => {
       await queryClient.prefetchQuery(key2, () => 'data2')
       expect(queryCache.findAll().length).toBe(2)
     })
+
+    test('should return all the queries when key contains object with an undefined property', async () => {
+      const baseKey = queryKey()
+      await queryClient.prefetchQuery([...baseKey, { a: 1 }], () => 'data1')
+      expect(queryCache.findAll([...baseKey, { a: undefined }])).toHaveLength(1)
+    })
```
TkDodo, line 211: thanks. If I get this right, the more realistic example would be something like this?

```
    test('should return all the queries when key contains object with an undefined property', async () => {
      const baseKey = queryKey()

      const createKey = (id?: number) => [{ ...baseKey, a: id }]

      await queryClient.prefetchQuery(createKey(1), () => 'data1')
      await queryClient.prefetchQuery(createKey(), () => 'data-nothing')
      expect(queryCache.findAll(createKey())).toHaveLength(2)
    })
```

so when we use the same `createKey` function to create our keys with optional filters, and then use it e.g. for invalidation, it doesn't match them all if we don't provide the param, but it should?

The thing is, there isn't really a way to make this work with either object or array keys. As soon as you add an entry to the array or object, it can't fuzzily match a non-existing entry

maxandron: Precisely, I'll edit the test accordingly.
It can also happen when simply using a custom hook with an optional parameter (a key factory is not really required).

In addition, any query created with `createKey()` will be hashed as if it were created with just `[{ ...baseKey }]`

A temporary workaround is to wrap the filter value with an object and replace it with an empty object if undefined:

```typescript
const baseKey = queryKey()
const createKey = (filter?: string) => [{...baseKey, filter: filter !== undefined ? { filter } : {}}]
// Instead of: const createKey = (filter?: string) => [{...baseKey, filter }]
await queryClient.prefetchQuery(createKey("filter1"), () => 'filtered-data')
await queryClient.prefetchQuery(createKey(), () => 'all-data')
expect(queryCache.findAll(createKey())).toHaveLength(2)
```

TkDodo: the root problem is that we somehow assume that passing an `undefined` filter to the `createKey` function would result in filter not being present in the queryKey. This might be amplified by the devtools showing hashed keys, and in the standard hashing process (JSON.stringify), `undefined` is removed.

But in an object, `undefined` as a value will be present because we add a valid key. It's the same for arrays:

```
const createKey = (filter?: string) => [...baseKey, filter]
```

this will have the exact same problem. And we can't just remove `undefined` from arrays because order matters.

maxandron: Yep, the dev tools showing the hashed keys amplifies the problem. It stems from the fact that the hash is not representative of the key (which is not what you would expect from a hash function)

That means the hash for two different keys can be the same (which means it's an inadequate hash function).

Perhaps it's more expected for tuples to be of the same length regardless of the values we pass inside, while it's unexpected for objects.
I may be wrong, as referencing an out-of-bounds element in an array also produces the value `undefined`.

So there are a few ways to go about this, I guess:

1. Don't change anything and add documentation to help users in this scenario (perhaps even a note on your blog can go a long way)
2. Ignore undefined in objects only (not in arrays) when doing comparisons (e.g., through a new opt-in option)
3. Fix the hash function to act similarly to the comparison
4. Make the dev tools show the key instead of the hash
5. Ignore the problem

What do you think?

Issue #4615, opened 2022-12-09
useQueries tries to inject default query client when query array is empty


stellR42:
### Describe the bug

I'm using `vue-query` in a packaged library so instead of using the `VueQueryPlugin` I define my own query client and pass it to every query call. The issue is that when using `useQueries`, when the query array is empty I get this error: ` No 'queryClient' found in Vue context, use 'VueQueryPlugin' to properly initialize the library.`

This seems to be because `vue-query` can't find a passed-in query client (obviously, since there are no queries) so it defaults to trying to inject the default query client that `VueQueryPlugin` would have created.

### Your minimal, reproducible example

https://codesandbox.io/s/angry-estrela-jsovt5?file=/src/App.vue

### Steps to reproduce

Error is displayed as soon as the sandbox loads.

### Expected behavior

It should bail out as soon as it sees that the query array is empty instead of making assumptions about what the defaults should be. Or maybe there can be an option to pass a query client to the `useQueries` function without relying on the query client in each query config.

### How often does this bug happen?

Every time

### Platform

Windows Chrome v102

### Tanstack Query version

   @tanstack/vue-query 4.19.1

TkDodo:
good point. I think the api for passing in a custom queryClient should actually be different for useQueries. In the react adapter, we have a similar thing with passing in a custom react context (this will change in v5 when we will _also_ pass in a custom queryClient).

The api we have in react is:

```
useQueries({ queries: [ query1, query2, ... queryN ], context })
```

where `query1` to `queryN` is: `Omit<UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>, 'context'>` so that you cannot pass in a context for each query, but instead, have to provide one at the very top for all queries.

Yes, this limits all queries to use the same context / queryClient, but I think this is alright.

As a non-breaking fix, we could allow both - `queryClient` at the top as well as `queryClient` on each query, but discourage the latter (with a warning), and remove it in v5. Thoughts @DamianOsipiuk ?

DamianOsipiuk:
Yup, definitely we should provide `queryClient` instance next-to `queries` , not individually.

PR #4625, opened 2022-12-12
feat(vue-query): allow to pass queryClient directly in useQueries options

DamianOsipiuk:
Closes: https://github.com/TanStack/query/issues/4615

### DIFF
--- a/packages/vue-query/src/useQueries.ts
+++ b/packages/vue-query/src/useQueries.ts
@@ -148,12 +148,13 @@ export function useQueries<T extends any[]>({
   const optionsQueryClient = unreffedQueries.value[0]?.queryClient as
     | QueryClient
     | undefined
-  const queryClient = queryClientInjected ?? optionsQueryClient ?? useQueryClient(queryClientKey)
+  const queryClient =
+    queryClientInjected ?? optionsQueryClient ?? useQueryClient(queryClientKey)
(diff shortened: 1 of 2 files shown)


packages/vue-query/src/useQueries.ts
```diff
@@ -146,7 +148,16 @@ export function useQueries<T extends any[]>({
...
-  const queryClient = optionsQueryClient ?? useQueryClient(queryClientKey)
+  const queryClient =
+    queryClientInjected ?? optionsQueryClient ?? useQueryClient(queryClientKey)
+  if (queryClientKey || optionsQueryClient) {
+    queryClient
+      .getLogger()
+      .error(
+        `Providing queryClient to individual queries in useQueries has been deprecated and will be removed in the next major version. You can still pass queryClient as an option directly to useQueries hook.`,
+      )
+  }
```
TkDodo, line 159: can we put that behind a `if (process.env.NODE_ENV !== 'production') {` check? This is how we usually do those logs:

https://github.com/TanStack/query/blob/1e0745a6cc80cbdbafc3153812b7982a8f549447/packages/query-core/src/queryClient.ts#L76-L80

DamianOsipiuk: Good catch!

TkDodo (approved):

PR #4626, opened 2022-12-13
refactor(solid-query): vite-plugin-solid backwards compatibility

ardeora:
Add backwards compatibility to the most popular vite-plugin-solid version (2.3.0) and will also help work with nx.

It also removes the version lock from all solid-query examples


TkDodo (approved):

ardeora:
> description says:
> 
> > It also removes the version lock from all solid-query examples
> 
> But I don't see that anywhere?

apologies! Yeah that was error on my part while copy pasting from the older PR. The version lock has already been removed in an earlier PR so we're good to go here!

TkDodo:
okay. if we merge this, it won't be release automatically because there are no changes in a `src` directory. We would need to do it manually by adding `RELEASE_ALL` somewhere in the body when squashing the commit. This would release all packages though.

Alternatively, we could just merge, but wait with a release until the next time something changes. Whatever you prefer.

oliverbutler:
Hey @ardeora, I've noticed I get an issue when building when this "solid" condition (for context when using @ts-rest's solid-query package I'm developing with solid-start).

But not when I remove the condition, this isn't an issue report, I'd just love to understand why this "solid" condition exists, and say, why it doesn't exist for the react-query package.

Happy to open a discussion/issue.

Screenshot 2022-12-18 at 19 48 07

PR #4636, opened 2022-12-13
fix(query-core): allow multiple roots to share same QueryClient

appden:
If two roots shared the `QueryClient` instance, then it would never unsubscribe from online/focus manager listeners and would also forever receive duplicate events because the unsubscriber functions would get "forgotten".

These issues would be especially noticeable on React Native apps where multiple roots that share the same `QueryClient` are common.

### DIFF
--- a/packages/query-core/src/queryClient.ts
+++ b/packages/query-core/src/queryClient.ts
@@ -62,6 +62,7 @@ export class QueryClient {
   private defaultOptions: DefaultOptions
   private queryDefaults: QueryDefaults[]
   private mutationDefaults: MutationDefaults[]
+  private mountCount: number
   private unsubscribeFocus?: () => void
   private unsubscribeOnline?: () => void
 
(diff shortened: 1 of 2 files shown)


packages/query-core/src/queryClient.ts
```diff
@@ -81,6 +83,9 @@ export class QueryClient {
   }
 
   mount(): void {
+    this.mountCount++
+    if (this.mountCount !== 1) return
+
```
TkDodo, line 88: could we alternatively just call `this.unmount()` at the beginning of the `mount` function here? Basically saying: "If you `mount` a second time, we `unmount` the first instance", as opposed to: "If you `mount` a second time, we don't do anything because you're already mounted" ....

I've tried your test with this approach and it passed.

appden: Thanks for taking a look at this! You're right, the test perhaps could be better, but that approach would break if we mounted two roots that share a client, then unmounted just one. In that circumstance, we'd have unsubscribed even though there is still one root mounted, hence the need for a count. 😀

appden: @TkDodo I just pushed up an additional test to cover the case I described above.

TkDodo (approved):

PR #4665, opened 2022-12-20
docs(vue import): fix import path for Vue plugin

lioneaglesolutions:
The docs were showing `vue-query` instead of `@tanstack/vue-query` for installing in Vue


DamianOsipiuk:
Could you also fix the other import in the same file?

lioneaglesolutions:
@DamianOsipiuk done!

PR #4672, opened 2022-12-22
Docs V2: fix basic-graphql-request example links

wbgneto:


TkDodo:
it doesn't make much sense to have the examples of `v2` point to `main`, which is currently `v4`.

Issue #4677, opened 2022-12-23
rename useErrorBoundary


TkDodo:
### context

A property that starts with the prefix `use` is unfortunately named, as `use` usually indicates a hook in React. `ErrorBoundary` might also be a quite react specific term.

Renaming the property to e.g. `throwError` would more accurately describe what is happening: the error will be thrown. In React, it will be picked up by the nearest error boundary.

### proposal

- rename `useErrorBoundary` to `throwError`

Moshyfawn:
Should this issue be assigned to be then?

PR #4697, opened 2022-12-23
refactor: rename useErrorBoundar option to throwError

Moshyfawn:
`throwError` name makes the most sense as it's a common name within request packages like [ky](https://github.com/sindresorhus/ky) _(ex: [`throwHttpErrors`](https://github.com/sindresorhus/ky#throwhttperrors))_

Closes: #4677


packages/react-query/src/utils.ts
```diff
@@ -2,7 +2,7 @@ export function shouldThrowError<T extends (...args: any[]) => boolean>(
   _useErrorBoundary: boolean | T | undefined,
```
Moshyfawn, line 2: What's the reason to prefix this `useErrorBoundary` arg with `_`? Should it be renamed to `_throwError` with `_` as well or `throwError` is fine?

TkDodo: I think this function used to look differently and there was likely a name clash. `throwError` is fine

TkDodo:
> throwHttpErrors

good point about that. Do you thin `throwErrors` would be better than `throwError` ?

Also, can you please add a paragraph to the migration guide:

https://github.com/TanStack/query/blob/cfb15ef582e080c82c149cdcba998d3d5e570ae7/docs/react/guides/migrating-to-react-query-5.md#L6-L9

Oh, and the PR needs to go to the `v5` branch please, not `main` 😅

Moshyfawn:
> > throwHttpErrors
> 
> good point about that. Do you thin `throwErrors` would be better than `throwError` ?

I'm not sure about the plural version as it might imply that there are multiple errors being thrown, which is not the case, I believe. I'm happy to change it to the plural version if you think it more fitting

Moshyfawn:
> Also, can you please add a paragraph to the migration guide:
> 
> https://github.com/TanStack/query/blob/cfb15ef582e080c82c149cdcba998d3d5e570ae7/docs/react/guides/migrating-to-react-query-5.md#L6-L9

Is it fine that the `id` is "migrating-to-**react**-query-5" and the guide title is "Migrating to **TanStack** Query v5"

Moshyfawn:
> Also, can you please add a paragraph to the migration guide:
> 
> https://github.com/TanStack/query/blob/cfb15ef582e080c82c149cdcba998d3d5e570ae7/docs/react/guides/migrating-to-react-query-5.md#L6-L9

Here you go 04b033bfe872cab747cb2d78bfad595ac3064cea

TkDodo:
> Is it fine that the id is "migrating-to-react-query-5" and the guide title is "Migrating to TanStack Query v5"?

hm, the id is the slug in the url. So maybe it should just be `migrate-to-v5`. I can do this directly on the v5 branch though.

> I'm not sure about the plural version as it might imply that there are multiple errors being thrown, which is not the case, I believe. I'm happy to change it to the plural version if you think it is more fitting.

> Although, you're able to reset the error boundary, at which point there can be another error thrown, which makes it multiple..

yeah let'a go with `throwErrors` please. It can be different kind of errors, like an error from the queryFn or an error from the selectFn. It can also be different errors at different times as you mentioned. I would read it as: "throw all errors that could potentially occur"

Moshyfawn:
> > Is it fine that the id is "migrating-to-react-query-5" and the guide title is "Migrating to TanStack Query v5"?
> 
> hm, the id is the slug in the url. So maybe it should just be `migrate-to-v5`. I can do this directly on the v5 branch though.
> 
> > I'm not sure about the plural version as it might imply that there are multiple errors being thrown, which is not the case, I believe. I'm happy to change it to the plural version if you think it is more fitting.
> 
> > Although, you're able to reset the error boundary, at which point there can be another error thrown, which makes it multiple..
> 
> yeah let'a go with `throwErrors` please. It can be different kind of errors, like an error from the queryFn or an error from the selectFn. It can also be different errors at different times as you mentioned. I would read it as: "throw all errors that could potentially occur"

Makes sense! Will di

TkDodo (approved):

TkDodo:
@allcontributors add @Moshyfawn for code

PR #4699, opened 2022-12-24
refactor(reactBatchedUpdates): remove the side effect that sets batchNotifyFn in react

### LINKED ISSUE #4685: remove unstable_batchedUpdates
### context

we currently use `unstable_batchedUpdates` as our `batchNotifyFn` in React and React Native:

https://github.com/TanStack/query/blob/357ec041a6fcc4a550f3df02c12ecc7bcdefbc05/packages/react-query/src/setBatchUpdatesFn.ts#L4

Apparently, this function is a "noop" in React18, as discussed here:

https://github.com/facebook/react/issues/24831#issuecomment-1296944806

### proposal

- remove the default side-effect that sets batchedUpdates
 - keep the option to set a batchUpdatesFn via the notifyManager
- this would also mean we would be completely side-effect free:

https://github.com/TanStack/query/blob/9b21609350d9ad41faf5c2fb0ac6c67e4324a637/packages/react-query/package.json#L24-L26

RodSarhan:
This PR attempts to fix #4685 of v5 milestones 

- set sideEffects to false in react-query/package.json
- remove the side effect import from index.ts
- delete reactBatchedUpdates.ts, reactBatchedUpdates.native.ts, setBatchUpdatesFn.ts files and references to them

### DIFF
--- a/packages/react-query/package.json
+++ b/packages/react-query/package.json
@@ -21,9 +21,7 @@
     },
     "./package.json": "./package.json"
   },
-  "sideEffects": [
-    "./src/setBatchUpdatesFn.ts"
-  ],
+  "sideEffects": false,
   "scripts": {
     "clean": "rm -rf ./build",
     "test:codemods": "../../node_modules/.bin/jest --config codemods/jest.config.js",
--- a/packages/react-query/src/index.ts
+++ b/packages/react-query/src/index.ts
@@ -1,8 +1,5 @@
 /* istanbul ignore file */
 
-// Side effects
-import './setBatchUpdatesFn'
-
 // Re-export core
 export * from '@tanstack/query-core'
 
--- a/packages/react-query/src/reactBatchedUpdates.native.ts
+++ b/packages/react-query/src/reactBatchedUpdates.native.ts
@@ -1,4 +1,0 @@
-// @ts-ignore
-// eslint-disable-next-line import/no-unresolved
-import { unstable_batchedUpdates } from 'react-native'
-export { unstable_batchedUpdates }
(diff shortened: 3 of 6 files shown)


TkDodo (approved):

TkDodo:
Thanks. I think we'll need to wait for the PR that removes the react17 test suite because in react17, the batching does make a difference :)

RodSarhan:
> Thanks. I think we'll need to wait for the PR that removes the react17 test suite because in react17, the batching does make a difference :)

Awesome thanks!
Yeah I was wondering if the solution should be backwards compatible, but looks like you decided to completely drop support for react 17

TkDodo:
react17 removal PR has been merged. can you solve the conflicts here please?

TkDodo:
Also, please add some lines to the migration guide, especially for react-native, where this might be needed to set manually.

RodSarhan:
The merge conflict should be resolved.
As for the migration guide, do we have any confirmation that this is still needed in react-native?

TkDodo:
> The merge conflict should be resolved.
As for the migration guide, do we have any confirmation that this is still needed in react-native?

Not really, but it depends on which react-native "architecture" you're using. I think what we can write is something along the lines of:

```
If your Framework supports a custom batching function, you can let TanStack Query know about it by calling, for example:

import { unstable_batchedUpdates } from 'react-native'
import { notifyManager } from '@tanstack/react-query
notifyManager.setBatchNotifyFunction(unstable_batchedUpdates)
```

solid does this with a `batch`  function now that is not unstable, we could also point to that:

https://github.com/TanStack/query/blob/b0d3c62070e8960ced79e6686c6c2f0154582a55/packages/solid-query/src/setBatchUpdatesFn.ts#L1-L4

TkDodo:
perfect, thank you 🙏

PR #4702, opened 2022-12-24
remove `remove` returned from useQuery

### LINKED ISSUE #4684: remove `remove` returned from `useQuery`
### context

`remove` removes the query from the queryCache without informing observers about it. It is best used to remove data imperatively that is no longer needed, e.g. when logging a user out.

It doesn't make much sense to do this while a query is still active, because it will just trigger a hard loading state with the next re-render. So when would you ever use the `remove` function returned from useQuery, because at that point, you still have an active observer ?

[anecdotal twitter thread](https://twitter.com/TkDodo/status/1587695241825554434)

### proposal

- remove `remove` in v5

Mamoanwar97:
This PR attempts to fix #4684 of v5 milestones, In this PR:

- remove `remove` method from `queryObserver`
- remove `remove` returned from `useQuery`
- remove `remove` returned from `useInfiniteQuery`
- update tests and docs accordingly


Mamoanwar97:
should I remove the `remove` method from the queryObserver class as well? (since it is not used anymore)
https://github.com/TanStack/query/blob/454482c4d1005d30235212a0cd44bac02be642da/packages/query-core/src/queryObserver.ts#L273

TkDodo:
> should I remove the remove method from the queryObserver class as well? (since it is not used anymore)

Yes please

Mamoanwar97:
> > should I remove the remove method from the queryObserver class as well? (since it is not used anymore)
> 
> Yes please

Done ✅

packages/react-query/src/__tests__/useQuery.test.tsx
```diff
@@ -862,71 +857,6 @@ describe('useQuery', () => {
     })
   })
 
-  it('should not get into an infinite loop when removing a query with cacheTime 0 and rerendering', async () => {
```
TkDodo, line 865: can we keep the test but implement the removing with `queryClient.removeQueries({ queryKey: key })` instead? I think the test itself is still valuable

packages/react-query/src/__tests__/useQuery.test.tsx
```diff
@@ -1201,96 +1131,6 @@ describe('useQuery', () => {
...
-        { notifyOnChangeProps: 'all' },
-      )
-
-      states.push(state)
-
-      const { remove, refetch } = state
-
-      return (
-        <div>
-          <button onClick={() => remove()}>remove</button>
```
TkDodo, line 1267: we could also keep this test and just do `queryClient.removeQueries({ queryKey: key })` here

TkDodo (commented):
Thank you 🙏 , please have a look at the comments.

We'd also need to add documentation to the migration guide, which is currently empty:

https://github.com/TanStack/query/blob/cfb15ef582e080c82c149cdcba998d3d5e570ae7/docs/react/guides/migrating-to-react-query-5.md#L6-L9

If you don't want to write something, let me know and I'll do it :)

TkDodo:
PR needs to go to the `v5` branch, not `main` please

packages/react-query/src/__tests__/useQuery.test.tsx
Mamoanwar97: Done ✅

packages/react-query/src/__tests__/useQuery.test.tsx
Mamoanwar97: Done ✅

TkDodo (approved):

Mamoanwar97:
> Thank you 🙏 , please have a look at the comments.
> 
> We'd also need to add documentation to the migration guide, which is currently empty:
> 
> https://github.com/TanStack/query/blob/cfb15ef582e080c82c149cdcba998d3d5e570ae7/docs/react/guides/migrating-to-react-query-5.md#L6-L9
> 
> If you don't want to write something, let me know and I'll do it :)

I tried to write something using the description you wrote in the issue itself, Let me know if you want to change something in the wording 😅

TkDodo:
@allcontributors add @Mamoanwar97 for code

Issue #4696, opened 2022-12-23
Eslint plugin removes generics from useQuery hooks


brianlovin:
### Describe the bug

Originally [shared on Twitter](https://twitter.com/brian_lovin/status/1606351401101074432).

The `@tanstack/eslint-plugin-query` package is removing generics from `useQuery` hooks when `--fix`ed:

![Screen Shot 2022-12-23 at 11 39 46@2x](https://user-images.githubusercontent.com/1923260/209392982-84c5c6a5-7982-4ab4-8cec-c66a2b5368a7.png)

### Your minimal, reproducible example

N/A

### Steps to reproduce

1. Have a `useQuery` hook with generics passed for `Data` and `Error`
2. Install the eslint plugin with default rules
3. `--fix` the file

### Expected behavior

I would expect my manually-typed generics for `Data` and `Error` to be considered when the eslint plugin switches me to `prefer-object-syntax` rules.

### How often does this bug happen?

Every time

### Screenshots or Videos

![Screen Shot 2022-12-23 at 11 39 46@2x](https://user-images.githubusercontent.com/1923260/209393202-19954465-1645-4aa5-b6e4-5135ca1a8c68.png)

### Platform

- macOS
- VSCode

### Tanstack Query version

4.2.3

### TypeScript version

4.8.2

brianlovin:
Note: after the initial `--fix` removes the generics, if I add them back in, they won't be cleared. They are only being cleared in the codemod that moves to object syntax.

TkDodo:
@Newbie012 fyi

Newbie012:
should be fixed in https://github.com/TanStack/query/pull/4703

PR #4703, opened 2022-12-24
fix(eslint-plugin): prefer object syntax should not remove generics

Newbie012:
fixes https://github.com/TanStack/query/issues/4696


packages/eslint-plugin-query/src/rules/prefer-query-object-syntax/prefer-query-object-syntax.test.ts
```diff
@@ -1,7 +1,11 @@
 import { rule, name } from './prefer-query-object-syntax'
-import { createRuleTester, normalizeIndent } from '../../utils/test-utils'
+import { normalizeIndent } from '../../utils/test-utils'
+import { ESLintUtils } from '@typescript-eslint/utils'
 
-const ruleTester = createRuleTester()
```
Newbie012, line 4: while `createRuleTester` is using `@typescript-eslint/parser`, it didn't understand type annotations. instead, it inferred them as `BinaryExpression` instead. Thus, I decided to drop it.

Issue #4700, opened 2022-12-24
eslint-plugin-query - @tanstack/query/prefer-query-object-syntax not identify queryKeys that created by function


nirtamir2:
### Describe the bug

This pass for:
```ts
useQuery(queryKeys.userById(userId), async () => await fetchUserById(userId))
```
But I expect eslint will warn me about `@tanstack/query/prefer-query-object-syntax` and will suggest changing it to:
```ts
useQuery({
  queryKey: queryKeys.userById(userId),
  queryFn : async () => await fetchUserById(userId)
})
```

### Your minimal, reproducible example

https://codesandbox.io/s/bold-tereshkova-ncsfj1?file=/src/index.jsx

### Steps to reproduce

Write
```ts
const queryKeys = {
userById: (userId: string) => ["users", {userId}]
} as const

async function fetchUserById(userId: string){
	return [{userId}]
}

useQuery(queryKeys.userById(userId), async () => await fetchUserById(userId))
```

And change eslintrc config to be
```js
module.exports = {
  extends: ["plugin:@tanstack/eslint-plugin-query/recommended"],
};
```

### Expected behavior

@tanstack/query/prefer-query-object-syntax will lint and fix it, instead of passing the lint rule

### How often does this bug happen?

Every time

### Platform

MacOS

### Tanstack Query version

^4.20.2

Newbie012:
should be fixed in https://github.com/TanStack/query/pull/4704

PR #4704, opened 2022-12-24
fix(eslint-plugin): allow query key factory style

Newbie012:
fixes https://github.com/TanStack/query/issues/4700


TkDodo:
ah, this conflicts now with the other PR I just merged 😅

Newbie012:
yea, it happens since because I add tests for each PR in the same file at the end 😅

Issue #4676, opened 2022-12-23
make TError default to Error instead of unknown


TkDodo:
### context

In JavaScript, you can `throw` anything, even literals like `5` or Promises (?? suspense). That is why the generic for the type `TError` defaults to `unknown`. This is in line with how TypeScript itself handles [errors in catch clauses](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables) since v4.4.

This is not very practical. Unless you explicitly throw a non-error, error will be at least of type `Error`. If you use `axios`, it might be of type `AxiosError`, but that is also not guaranteed. For example, if you have a runtime error in `select`, that will be caught, and it puts your query into `error` state. The error will be of type `Error` then.

Practically, it means that you either fall back to runtime checks:

```ts
const { error } = useQuery(...)
if (error instanceof Error) {
  // do something with error.message
}
```

or, you pass a generic to `useQuery`:

```ts
const { error, isError } = useQuery<Todos, Error>(...)
if (isError) {
  // do something with error.message
}
```

The second approach is bad for two reasons:

1. `TError` is the second generic, so you also have to annotate the first generic, which is the return type of the queryFn. This could actually be inferred.
2. `useQuery` has 4 generics, and if you only provide two of them, the other two will fall back to their default value instead of being inferred from their usage. That means `select` will not work as expected:

```ts
const { error, isError } = useQuery<Todos, Error>({
  queryKey: ['todos'],
  queryFn: fetchTodos,
  select: (data) => data[0],
})

if (isError) {
  // do something with error.message
}
```

[...]

PR #4706, opened 2022-12-25
feat(TError) : Make TError default to Error instead of unknown

judicaelandria:
This PR makes the TError default to Error instead of unknown, more info about this [here](https://github.com/TanStack/query/issues/4676)


packages/query-core/src/infiniteQueryObserver.ts
```diff
@@ -23,7 +23,7 @@ type InfiniteQueryObserverListener<TData, TError> = (
 
 export class InfiniteQueryObserver<
   TQueryFnData = unknown,
-  TError = unknown,
+  TError extends Error = Error,
```
TkDodo, line 26: why `extends Error` ? This will make it impossible for users to opt out of this, for example if they `throw "some error occurred"`, their `TError` is `string`, which doesn't `extend Error` ❗

packages/query-core/src/tests/mutations.test.tsx
```diff
@@ -325,7 +325,7 @@ describe('mutations', () => {
       context: undefined,
       variables: undefined,
       data: 'new',
-      error: undefined,
+      error: null,
```
TkDodo, line 328: I wouldn't expect any changes in tests from a type level change. Why is `error` `null` here instead of `undefined` ?

packages/react-query/src/__tests__/suspense.test.tsx
```diff
@@ -669,7 +669,7 @@ describe("useQuery's in Suspense mode", () => {
         {
           retry: false,
           suspense: true,
-          useErrorBoundary: (err) => err !== 'Local Error',
+          useErrorBoundary: (err) => err.message !== 'Local Error',
```
TkDodo, line 672: this can't work because we `return Promise.reject('Remote Error')`. We should probably `return Promise.reject(new Error('Remote Error'))`

packages/react-query/src/__tests__/useQueries.test.tsx
```diff
@@ -145,7 +145,7 @@ describe('useQueries', () => {
             return (i + 1) * count * 2
           },
         })),
-      })
+      }) as QueryObserverResult<unknown, Error>[]
```
TkDodo, line 148: why is this change necessary?

TkDodo (changes requested):

packages/query-core/src/tests/mutations.test.tsx
judicaelandria: because when we defaulted the type to Error, it's expecting a null value instead of undefined

packages/query-core/src/infiniteQueryObserver.ts
judicaelandria: yes, I got that but sometimes we need to narrow the type, I changed it, can you take a look? if it's not better can you suggest a solution?

packages/react-query/src/__tests__/useQueries.test.tsx
judicaelandria: we needed to infer the types here because the type of queries here is unknown altho the states is waiting a type Error as a default type

TkDodo: the Problem seems to be that result is inferred as:

```
const result: UseQueryResult<number, unknown>[]
```

while I think it should be:

```
const result: UseQueryResult<number, Error>[]
```

If we just bypass the issue here in the test, users might suffer from the same issue as well, so I think we should strive towards fixing the issue.

I tried to tinker a bit with the `useQueries` types but with no success. I've reverted some of the type assertions (https://github.com/TanStack/query/pull/4706/commits/9e6368b575ae70f52235eeb9412e85fdeb538743) to have the state we currently have on `main`. I was hoping that @artysidorenko could maybe have a look? I think he knows what to do to make types default to `Error` instead of `unknown`. I think all the issue left are with `useQueries`.

judicaelandria: yes you're totally right, we need to fix instead of bypassing the error, I'll to find a fix for the useQueries, but if the guy you mentioned can help that would be great

judicaelandria:
closes https://github.com/TanStack/query/issues/4676

packages/react-query/src/__tests__/useQueries.test.tsx
artysidorenko: Hi folks - I think [this](https://github.com/artysidorenko/react-query/commit/3fddbffec6f5301c92e99adff721ebc11d426927) should do it for the TError type - it was just missing the defaulting on dynamic arrays.

Also saw a small type complaint in the function body of `useQueries` because the queryClient.defaultQueryOptions method [didn't have any default type params](https://github.com/artysidorenko/react-query/commit/35b3aaa855e138341afb871999487766bd0affd2)

judicaelandria:
@TkDodo done with this one too

TkDodo (approved):
🚀

TkDodo:
can you please resolve the conflicts? We've merged the big syntax change (object syntax) PR and sadly it causes quite some conflicts. Should be straight forward to solve though - sorry!

judicaelandria:
> can you please resolve the conflicts? We've merged the big syntax change (object syntax) PR and sadly it causes quite some conflicts. Should be straight forward to solve though - sorry!



> can you please resolve the conflicts? We've merged the big syntax change (object syntax) PR and sadly it causes quite some conflicts. Should be straight forward to solve though - sorry!

done

TkDodo:
formatting is failing:

```
> query@ prettier /home/runner/work/query/query
> prettier "{packages,examples}/**/src/**/*.{md,js,jsx,ts,tsx,json,vue}" "--check"

Checking formatting...
[warn] packages/solid-query/src/createQueries.ts
```

judicaelandria:
> formatting is failing:
> 
> ```
> > query@ prettier /home/runner/work/query/query
> > prettier "{packages,examples}/**/src/**/*.{md,js,jsx,ts,tsx,json,vue}" "--check"
> 
> Checking formatting...
> [warn] packages/solid-query/src/createQueries.ts
> [warn] Code style issues found in the above file. Forgot to run Prettier?
> ```

ah yeah! sorry, forgot to run it, fixed

TkDodo:
seems like the same solid tests is always failing, and its one that was touched in this PR. do you maybe know why? Or should we ask @ardeora 😅

judicaelandria:
> seems like the same solid tests is always failing, and its one that was touched in this PR. do you maybe know why? Or should we ask @ardeora 😅 

oh that's odd!all the test is passing locally, i'll investigate

judicaelandria:
@TkDodo this a new error from the merge, it has something to do with the suspense
Screenshot 2023-01-05 at 9 17 52 PM

TkDodo:
> @TkDodo this a new error from the merge, it has something to do with the suspense

I think it's because the `ts-expects-error` was spanning the whole line, and now it's two lines because of syntax changes, and the second line is not "covered" anymore. It only errors in the local IDE, not in the `typecheck` task. But yeah, this test is accessing private fields, which it shouldn't - but I've already re-written it in a different branch.

Anyways, this is not the problem here. The problem was that we still had some tests that rejected a promise with a string. I fixed that here: b2d62d9266ac75d88606e889f4b5eac38cb0b76a

Issue #4674, opened 2022-12-23
require minimum of TS 4.7


TkDodo:
v5 will need TS4.7 or greater to support the object syntax properly.

incepter:
I believe the ts version in the repository of react-query should always use the latest version (or any the maintainers feel okay with!), since it won't affect the developers' land

TkDodo:
If we use the latest version and use features of the latest version (4.9), like `satisfies`, the .d.ts files that we ship as type definitions will not work with versions prior to 4.9. we could do differential transpiling and ship different types for lower TS versions but that effort is not worth it imo.

PR #4708, opened 2022-12-25
Feature/ts 4.7

TkDodo:
closes https://github.com/TanStack/query/issues/4674

### DIFF
--- a/docs/react/guides/migrating-to-react-query-5.md
+++ b/docs/react/guides/migrating-to-react-query-5.md
@@ -11,7 +11,7 @@ v5 is a major version, so there are some breaking changes to be aware of:
 
 Previously, remove method used to remove the query from the queryCache without informing observers about it. It was best used to remove data imperatively that is no longer needed, e.g. when logging a user out.
 
-But It doesn't make much sense to do this while a query is still active, because it will just trigger a hard loading state with the next re-render. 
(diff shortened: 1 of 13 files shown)


TkDodo:
@ardeora I fixed the TS version to 4.7.4 because some examples pulled in 4.8 and we should only have one TS version in the monorepo. Because of this, two type errors now show up in solid query. Can you maybe have a look?

```
Error: packages/solid-query/src/createBaseQuery.ts(48,15): error TS2367: This condition will always return 'false' since the types 'TData | undefined' and 'unique symbol' have no overlap.
Error: packages/vue-query/src/useBaseQuery.ts(132,9): error TS2352: Conversion of type 'ToRefs<{ readonly data: DeepReadonly<UnwrapRef<UnwrapRef<TData>>>; readonly error: DeepReadonly<UnwrapRef<UnwrapRef<TError>>>; ... 21 more ...; readonly fetchStatus: FetchStatus; } | { ...; } | { ...; } | { ...; }>' to type 'UseQueryReturnType<TData, TError, QueryObserverResult<TData, TError>>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
```

DamianOsipiuk:
@TkDodo Fixed the error in `vue-query`

ardeora:
@TkDodo Fixed the error in `solid-query` as well 💯

TkDodo:
thank you ❤️

PR #4714, opened 2022-12-27
Remove overloads

### LINKED ISSUE #4694: remove overloads
### context

`useQuery` and friends have many overloads in TypeScript - different ways how the function can be invoked. As an example, let's look at `useQuery`. You can call it in three different ways:

```
useQuery(queryKey, queryFn, options)
useQuery(queryKey, options)
useQuery(options)
```

The `queryKey` is the only required property. The queryFn is optional because you might have defined it globally. If you use the syntax where you only pass one object in, the queryKey is also required.

Not only is this tough to maintain, type wise, it also requires a runtime check to see which type the first and the second parameter are to correctly create options. Internally, we only work with one options object.

Further, we started to add more overloads for features that are only doable with overloads. For example, if you pass `initialData`, the return type of `data` will _not_ contain `undefined`. Because `initialData` can also be a function, we need another 3 overloads for each possibility to call `useQuery`, resulting in 9 total overloads.

The `useQuery.ts` file has 140 lines of code - only 3 of which are actual JavaScript.

Other functions suffer from the same problem, and overloads are also not consistent. For example, `queryCache.findAll` has overloads, but `mutationCache.findAll` does not.

### proposal

- _remove_ all overloads and only allow all functions to be called with a single signature - one object

### detailed changes

```diff
- useQuery(key, fn, options)
+ useQuery({ queryKey, queryFn, ...options })
// ⬆️ queryKey must be required

- useInfiniteQuery(key, fn, options)
+ useInfiniteQuery({ queryKey, queryFn, ...options })
// ⬆️ queryKey must be required

- useMutation(fn, options)
+ useQuery({ mutationFn, ...options })

- useIsFetching(key, filters)
+ useIsFetching({ queryKey, ...filters })

- useIsMutating(key, filters)
+ useIsMutating({ mutationKey, ...filters })
```

```diff
- queryClient.isFetching(key, filters)
+ queryClient.isFetching({ queryKey, ...filters })

- queryClient.ensureQueryData(key, fn, options)
+ queryClient.ensureQueryData({ queryKey, queryFn, ...options })
// ⬆️ queryKey must be required

- queryClient.getQueriesData(key, filters)
+ queryClient.getQueriesData({ queryKey, ...filters })

- queryClient.setQueriesData(key, updater, filters, options)
+ queryClient.setQueriesData({ queryKey, ...filters }, updater, options)

- queryClient.removeQueries(key, filters)
+ queryClient.removeQueries({ queryKey, ...filters })

- queryClient.resetQueries(key, filters, options)
+ queryClient.resetQueries({ queryKey, ...filters }, options)

- queryClient.cancelQueries(key, filters, options)
+ queryClient.cancelQueries({ queryKey, ...filters }, options)

- queryClient.invalidateQueries(key, filters, options)
+ queryClient.invalidateQueries({ queryKey, ...filters }, options)

- queryClient.refetchQueries(key, filters, options)
+ queryClient.refetchQueries({ queryKey, ...filters }, options)

- queryClient.fetchQuery(key, fn, options)
+ queryClient.fetchQuery({ queryKey, queryFn, ...options })

- queryClient.prefetchQuery(key, fn, options)
+ queryClient.prefetchQuery({ queryKey, queryFn, ...options })

- queryClient.fetchInfiniteQuery(key, fn, options)
+ queryClient.fetchInfiniteQuery({ queryKey, queryFn, ...options })

- queryClient.prefetchInfiniteQuery(key, fn, options)
+ queryClient.prefetchInfiniteQuery({ queryKey, queryFn, ...options })

```

```diff
- queryCache.find(key, filters)
+ queryCache.find({ queryKey, ...filters })
// ⬆️ queryKey must be required
// `exact` could be omitted from possible filters and always set to `true` !

- queryCache.findAll(key, filters)
+ queryCache.findAll({ queryKey, ...filters })
```

Mamoanwar97:
This PR attempts to fix #4694 of v5 milestones, In this PR:

- remove overloads from query-core
- remove overloads from react-query
- remove overloads from solid-query
- remove overloads from vue-query
- update tests, docs, and examples accordingly

### DIFF
--- a/docs/react/adapters/solid-query.md
+++ b/docs/react/adapters/solid-query.md
@@ -13,7 +13,7 @@ import { Switch, Match, For } from 'solid-js'
 const queryClient = new QueryClient()
 
 function Example() {
-  const query = createQuery(() => ['todos'], fetchTodos)
+  const query = createQuery({queryKey: () => ['todos'], queryFn: fetchTodos})
 
   return (
     <div>
@@ -69,10 +69,10 @@ Solid Query offers an API similar to  React Query, but there are some key differ
 
 ```tsx
 // ❌ react version
-useQuery(["todos", todo], fetchTodos)
+useQuery({queryKey: ["todos", todo], queryFn: fetchTodos})
 
 // ✅ solid version
-createQuery(() => ["todos", todo()], fetchTodos)
+createQuery({queryKey: () => ["todos", todo()], queryFn: fetchTodos})
 ```
 
 - Suspense works for queries out of the box if you access the query data inside a `<Suspense>` boundary.
@@ -81,7 +81,7 @@ createQuery(() => ["todos", todo()], fetchTodos)
 import { For, Suspense } from 'solid-js'
 
 function Example() {
-  const query = createQuery(() => ['todos'], fetchTodos)
+  const query = createQuery({queryKey: () => ['todos'], queryFn: fetchTodos})
   return (
     <div>
       {/* ✅ Will trigger loading fallback, data accessed in a suspense context. */}
@@ -120,13 +120,13 @@ function Example() {
   // )
 
   // ✅ solid version -- does not support destructuring outside reactive context
-  const query = createQuery(
-    () => ['repoData'],
-    () =>
+  const query = createQuery({
+    queryKey: () => ['repoData'],
+    queryFn: () =>
       fetch('https://api.github.com/repos/tannerlinsley/react-query').then(
         (res) => res.json(),
       ),
-  )
+  })
 
   // ✅ access query properties in JSX reactive context
   return (
@@ -161,7 +161,9 @@ const queryClient = new QueryClient()
 
 function Example() {
   const [enabled, setEnabled] = createSignal(false)
-  const query = createQuery(() => ['todos'], fetchTodos, {
+  const query = createQuery({ 
+    queryKey: () => ['todos'],
+    queryFn: fetchTodos,
     // ❌ passing a signal directly is not reactive
     // enabled: enabled(),
 
--- a/docs/react/adapters/solid-query.md
+++ b/docs/react/adapters/solid-query.md
@@ -13,7 +13,10 @@ import { Switch, Match, For } from 'solid-js'
 const queryClient = new QueryClient()
 
 function Example() {
-  const query = createQuery({queryKey: () => ['todos'], queryFn: fetchTodos})
+  const query = createQuery({
+    queryKey: () => ['todos'], 
+    queryFn: fetchTodos
+  })
 
   return (
     <div>
@@ -69,10 +72,16 @@ Solid Query offers an API similar to  React Query, but there are some key differ
 
 ```tsx
 // ❌ react version
-useQuery({queryKey: ["todos", todo], queryFn: fetchTodos})
+useQuery({
+  queryKey: ["todos", todo], 
+  queryFn: fetchTodos
+})
 
 // ✅ solid version
-createQuery({queryKey: () => ["todos", todo()], queryFn: fetchTodos})
+createQuery({
+  queryKey: () => ["todos", todo()], 
+  queryFn: fetchTodos
+})
 ```
 
 - Suspense works for queries out of the box if you access the query data inside a `<Suspense>` boundary.
@@ -81,7 +90,11 @@ createQuery({queryKey: () => ["todos", todo()], queryFn: fetchTodos})
 import { For, Suspense } from 'solid-js'
 
 function Example() {
-  const query = createQuery({queryKey: () => ['todos'], queryFn: fetchTodos})
+  const query = createQuery({
+    queryKey: () => ['todos'], 
+    queryFn: fetchTodos
+  })
+
   return (
     <div>
       {/* ✅ Will trigger loading fallback, data accessed in a suspense context. */}
@@ -113,11 +126,12 @@ export default function App() {
 
 function Example() {
   // ❌ react version -- supports destructing outside reactive context
-  // const { isLoading, error, data } = useQuery(['repoData'], () =>
-  //   fetch('https://api.github.com/repos/tannerlinsley/react-query').then(res =>
+  // const { isLoading, error, data } = useQuery({
+  //   queryKey: ['repoData'], () =>
+  //   queryFn: fetch('https://api.github.com/repos/tannerlinsley/react-query').then(res =>
   //     res.json()
   //   )
-  // )
+  // })
 
   // ✅ solid version -- does not support destructuring outside reactive context
   const query = createQuery({
--- a/docs/react/guides/migrating-to-react-query-5.md
+++ b/docs/react/guides/migrating-to-react-query-5.md
@@ -116,3 +116,7 @@ if you still need to remove a query, you can use `queryClient.removeQueries({que
 ### The minimum required TypeScript version is now 4.7
 
 Mainly because an important fix was shipped around type inference. Please see this [TypeScript issue](https://github.com/microsoft/TypeScript/issues/43371) for more information.
+
+### The `useErrorBoundary` prop has been renamed to `throwErrors`
+
+To make the `useErrorBoundary` prop more framework-agnostic and avoid confusion with the established React function prefix "`use`" for hooks and the "ErrorBoundary" component name, it has been renamed to `throwErrors` to more accurately reflect its functionality.
--- a/docs/react/guides/migrating-to-react-query-5.md
+++ b/docs/react/guides/migrating-to-react-query-5.md
@@ -79,13 +79,22 @@ now we only support the object format.
 + queryCache.findAll({ queryKey, ...filters })
 ```
 
-### `getQueryData` now accepts queryKey only as an Argument
+### `queryClient.getQueryData` now accepts queryKey only as an Argument
 
-`getQueryData` argument is changed to accept only a `queryKey`
+`queryClient.getQueryData` argument is changed to accept only a `queryKey`
 
 ```diff
-- queryClient.getQueryData({...filters})
+- queryClient.getQueryData(queryKey, filters)
 + queryClient.getQueryData(queryKey)
+```
+
+### `queryClient.getQueryState` now accepts queryKey only as an Argument
+
+`queryClient.getQueryState` argument is changed to accept only a `queryKey`
+
+```diff
+- queryClient.getQueryState(queryKey, filters)
++ queryClient.getQueryState(queryKey)
 ```
 
 ### The `remove` method has been removed from useQuery
--- a/docs/react/guides/migrating-to-react-query-5.md
+++ b/docs/react/guides/migrating-to-react-query-5.md
@@ -28,7 +28,7 @@ now we Only support the object format
 
 Previously, remove method used to remove the query from the queryCache without informing observers about it. It was best used to remove data imperatively that is no longer needed, e.g. when logging a user out.
 
-But It doesn't make much sense to do this while a query is still active, because it will just trigger a hard loading state with the next re-render. 
+But It doesn't make much sense to do this while a query is still active, because it will just trigger a hard loading state with the next re-render.
 
 if you still need to remove a query, you can use `queryClient.removeQueries({queryKey: key})`
 
@@ -39,3 +39,7 @@ if you still need to remove a query, you can use `queryClient.removeQueries({que
 - query.remove()
 + queryClient.removeQueries({ queryKey })
 ```
+
+### The minimum required TypeScript version is now 4.7
+
+Mainly because an important fix was shipped around type inference. Please see this [TypeScript issue](https://github.com/microsoft/TypeScript/issues/43371) for more information.
--- a/docs/react/guides/migrating-to-react-query-5.md
+++ b/docs/react/guides/migrating-to-react-query-5.md
@@ -11,7 +11,7 @@ v5 is a major version, so there are some breaking changes to be aware of:
 
 useQuery and friends used to have many overloads in TypeScript - different ways how the function can be invoked. Not only this was tough to maintain, type wise, it also required a runtime check to see which type the first and the second parameter, to correctly create options. 
 
-now we Only support the object format
+now we only support the object format.
 
 ```diff
 - useQuery(key, fn, options)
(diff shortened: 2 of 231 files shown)


docs/react/adapters/solid-query.md
```diff
@@ -69,10 +69,10 @@ Solid Query offers an API similar to  React Query, but there are some key differ
 
 ```tsx
 // ❌ react version
-useQuery(["todos", todo], fetchTodos)
+useQuery({queryKey: ["todos", todo], queryFn: fetchTodos})
 
 // ✅ solid version
-createQuery(() => ["todos", todo()], fetchTodos)
+createQuery({queryKey: () => ["todos", todo()], queryFn: fetchTodos})
```
TkDodo, line 75: I'm not sure if this is technically true - I think the whole object we pass into `createQuery` should be a function?

See the roadmap discussion with @ardeora here:
- https://github.com/TanStack/query/discussions/4252#discussioncomment-3779788

Also, for consistency, let's please use multi line whenever we use the object syntax:

```
useQuery({
  queryKey: ["todos", todo],
  queryFn: fetchTodos,
})
```

docs/react/adapters/solid-query.md
```diff
@@ -161,7 +161,9 @@ const queryClient = new QueryClient()
 
 function Example() {
   const [enabled, setEnabled] = createSignal(false)
-  const query = createQuery(() => ['todos'], fetchTodos, {
+  const query = createQuery({ 
+    queryKey: () => ['todos'],
+    queryFn: fetchTodos,
     // ❌ passing a signal directly is not reactive
     // enabled: enabled(),
```
TkDodo, line 168: I think this workaround is also unnecessary without the overloads. @ardeora can you have a look please

docs/react/guides/migrating-to-react-query-5.md
```diff
@@ -7,6 +7,23 @@ title: Migrating to TanStack Query v5
...
+```diff
+- useQuery(key, fn, options)
++ useQuery({ queryKey, queryFn, ...options })
+
+- useInfiniteQuery(key, fn, options)
++ useInfiniteQuery({ queryKey, queryFn, ...options })
+
+- useMutation(fn, options)
++ useMutation({ mutationFn, ...options })
+```
```
TkDodo, line 25: can we list all affected functions here please

examples/solid/basic-typescript/src/index.tsx
```diff
@@ -97,13 +99,11 @@ const getPostById = async (id: number): Promise<Post> => {
...
-    () => getPostById(postId),
-    {
-      enabled: !!postId,
-    },
-  )
+  return createQuery({
+    queryKey: () => ['post', postId],
+    queryFn: () => getPostById(postId),
+    enabled: !!postId,
+  })
```
TkDodo, line 106: also here, I think this should be:

```
createQuery(() => ({ queryKey, queryFn }))
```

packages/query-core/src/queryClient.ts
```diff
@@ -113,84 +107,29 @@ export class QueryClient {
   }
 
   getQueryData<TQueryFnData = unknown>(
-    queryKey: QueryKey,
-    filters?: QueryFilters,
+    filters: WithRequired<QueryFilters, 'queryKey'>,
   ): TQueryFnData | undefined {
-    return this.queryCache.find<TQueryFnData>(queryKey, filters)?.state.data
+    return this.queryCache.find<TQueryFnData>(filters)?.state.data
   }
```
TkDodo, line 113: I wasn't sure about `getQueryData`. I've only ever used it with a queryKey and no other filters. If the key is required, do the other filters even make sense 🤔. Maybe it would be better to have this as just a function that lets you pass in a `queryKey` only?

What do you think @ardeora @DamianOsipiuk ? On the other hand, one object is the consistent way 😅

packages/query-core/src/queryClient.ts
```diff
@@ -202,41 +141,28 @@ export class QueryClient {
...
       return undefined
     }
 
-    const parsedOptions = parseQueryArgs(queryKey)
-    const defaultedOptions = this.defaultQueryOptions(parsedOptions)
+    const defaultedOptions = this.defaultQueryOptions({ queryKey })
     return this.queryCache
       .build(this, defaultedOptions)
-      .setData(data, { ...options, manual: true })
+      .setData(data, { ...options, manual: true }) as TQueryFnData
```
TkDodo, line 155: why is this change needed?

packages/query-core/src/queryClient.ts
```diff
@@ -608,16 +348,21 @@ export class QueryClient {
   }
 
   setQueryDefaults(
-    queryKey: QueryKey,
-    options: QueryObserverOptions<unknown, any, any, any>,
+    options: WithRequired<
+      QueryObserverOptions<unknown, any, any, any>,
+      'queryKey'
+    >,
```
TkDodo, line 354: also here, not sure if these functions should be changed 🤔 . Would users still know that they'd be setting options for a specific key if we have:

```
queryClient.setQueryDefaults({
  queryKey: ['todos'],
  staleTime: 1000
})
```

?  Maybe it would be better to keep this as two arguments, and instead have the `queryKey` be _omitted_ from the options?

packages/react-query/src/types.ts
```diff
@@ -28,19 +29,19 @@ export interface UseBaseQueryOptions<
   TQueryData = TQueryFnData,
   TQueryKey extends QueryKey = QueryKey,
 > extends ContextOptions,
-    QueryObserverOptions<TQueryFnData, TError, TData, TQueryData, TQueryKey> {}
+    WithRequired<
+      QueryObserverOptions<TQueryFnData, TError, TData, TQueryData, TQueryKey>,
+      'queryKey'
+    > {}
```
TkDodo, line 35: why don't we make the `queryKey` required directly inside `QueryOptions` ?

https://github.com/TanStack/query/blob/2b01ac0828b8dcabfa1a1747f44aaeb11e9c28ad/packages/query-core/src/types.ts#L76

Is there ever a situation where we want the `queryKey` to be optional there? I think it was only optional because we had overloads and the key could be a separate parameter ...

packages/solid-query/src/utils.ts
```diff
@@ -1,61 +1,33 @@
...
-    return arg1 as any
-  }
-
-  if (typeof arg2 === 'function') {
-    return { ...arg3, queryKey: arg1(), queryFn: arg2 } as any
   }
-
-  return { ...arg2, queryKey: arg1() } as any
+  return arg as any
 }
```
TkDodo, line 18: not sure if this is still necessary. from what I can see, it tries to make the queryKey reactive ? I guess we'd now want that for the whole options object? Another question for @ardeora :)

packages/solid-query/src/createMutation.ts
```diff
@@ -95,8 +42,8 @@ export function createMutation<
   })
 
   createComputed(() => {
-    const newParsedOptions = parseMutationArgs(arg1, arg2, arg3)
-    setOptions(newParsedOptions)
+    const newParsedOptions = { ...mutationOptions }
+    setOptions(mutationOptions)
```
TkDodo, line 46: is it necessary to create a shallow copy of `mutationOptions` ?

packages/vue-query/src/useBaseQuery.ts
```diff
@@ -46,15 +45,14 @@ export function useBaseQuery<
...
-  arg3: UseQueryOptionsGeneric<TQueryFnData, TError, TData, TQueryKey> = {},
+  genericOptions: UseQueryOptionsGeneric<
+    TQueryFnData,
+    TError,
+    TData,
+    TQueryKey
+  >,
 ): UseQueryReturnType<TData, TError> {
-  const options = computed(() => parseQueryArgs(arg1, arg2, arg3))
+  const options = computed(() => parseQueryArgs(genericOptions))
```
TkDodo, line 55: should `parseQueryArgs` be renamed now, as it doesn't do any parsing anymore? Maybe `unrefQueryArgs` ? @DamianOsipiuk

TkDodo (commented):
awesome work, this is huge 🙌 . Please have a look at my comments

packages/react-query/src/types.ts
Mamoanwar97: `QueryOptions` is used everywhere and extended by many things .. As a safe option, I made the `queryKey` only required in the places we need it.

Here are some examples where I think we don't need the queryKey to be required:

default options

HydrateOptions

getNextPageParam

fetch-In query-core/query.ts

getDefaultState in query-core/query.ts

packages/solid-query/src/createMutation.ts
Mamoanwar97: I don't think so, I have just tried to remove it and the tests are passing .. will remove the shallow copy

packages/query-core/src/queryClient.ts
Mamoanwar97: for a reason, without the `as TQueryFnData`, typescript reads that `setData` is returning `unknown`

without-as-TQueryFnData

While inside `setData` it infers the first argument type and returns it so it should return the `TQueryFnData` 

setData

but for a reason that is not happening when we use setData inside setQueryData, although the type of `data` passed to the `setData` is `TQueryFnData` not `unknown`

typof data passed to setData

packages/vue-query/src/useBaseQuery.ts
Mamoanwar97: Done ✅

packages/solid-query/src/createMutation.ts
Mamoanwar97: Done ✅

packages/query-core/src/queryClient.ts
Mamoanwar97: Done ✅

docs/react/adapters/solid-query.md
Mamoanwar97: Done ✅

docs/react/guides/migrating-to-react-query-5.md
Mamoanwar97: Done ✅

examples/solid/basic-typescript/src/index.tsx
Mamoanwar97: It is my first time dealing with solid tbh, but I think the object syntax is working fine with the tests 

@ardeora is the object syntax fine? or should it be converted to an arrow function returning the object?

packages/query-core/src/queryClient.ts
TkDodo: ah yes, so would:

```
.setData<TQueryFnData>(data, { ...options, manual: true })
```

work as well?

packages/react-query/src/types.ts
TkDodo: that's fine, we can revisit this later if necessary. There are some functions that don't really need the full `QueryOptions` (like `getNextPageParam`).

packages/query-core/src/queryClient.ts
Mamoanwar97: Fixed ✅ 

I traced it and tried to find the difference between my branch and the main branch 

In the main branch:
- queryKey was passed to parseQueryArgs which returns a type of `QueryOptions<any, any, any, QueryKey>`
- when the `parsedOptions` is passed to `defaultQueryOptions` it returns ` DefaultedQueryObserverOptions<any, any, unknown, any, QueryKey>`

main-branch

but in my branch, because I was passing the `queryKey` only to `defaultQueryOptions` it inferred the type to be `DefaultedQueryObserverOptions<unknown, unknown, unknown, unknown, QueryKey>`

my-branch-before

so to fix it instead of `as` alias, I passed the generics to the `defaultQueryOptions` by myself

my-branch after

packages/query-core/src/queryClient.ts
DamianOsipiuk: I would opt for either:
- single argument with everything `optional`. As an alias to `getQueryCache().find()`
- single argument of `queryKey`. Then if you need more advanced search you would do `getQueryCache().find()`.

To be honest i would go with only `queryKey` as if i would like to use more filters i would expect to potentially get more than one result. Meaning `getQueriesData`.

packages/vue-query/src/useBaseQuery.ts
DamianOsipiuk: I would use `cloneDeepUnref` directly here and in other places.
Also with some minor changes to types of `cloneDeepUnref` and `cloneDeep` we could get rid of many `as` assertions all over the codebase.
But i guess i could do that in another PR.
This should be good enough. 👍

examples/solid/basic-typescript/src/index.tsx
ardeora: Thanks for working on this! Looks great. Yeah solid query `v5` will need more changes here. With the removal of overloads. We don't need to have `QueryKey` as a function anymore and would be more like what Dominik mentioned above. For example,

```ts
createQuery({
    queryKey: () => ['post', postId],
    queryFn: () => getPostById(postId),
    enabled: !!postId,
  })
```

will become 

```ts
createQuery(() => {
    queryKey: ['post', postId],
    queryFn: () => getPostById(postId),
    enabled: !!postId,
  })
```

But this change is definitely wayyy more involved since we need to change the typings in `types.ts` to get in line with with `react-query`. It would also change how `createQuery` tracks the signal changes inside the function. This would also mean changing the tests to reflect the same. @TkDodo  do you suggest we make those changes in a separate PR or should I add annotations on this PR itself?

packages/query-core/src/queryClient.ts
ardeora: Yeah I agree with @DamianOsipiuk. For example, if I use the filters like `exact: false` or `predicate`, I'm usually expecting more than one result

examples/solid/basic-typescript/src/index.tsx
TkDodo: I see 3 ways forward here:

1) you can checkout the PR with the gh cli and work on it directly
2) you can create a PR _to_ this PR :)
3) we can merge this to v5, as it will not be released, and you can work on it in a separate issue / PR

whatever you prefer is fine with me :)

packages/query-core/src/queryClient.ts
TkDodo: agree. It's weird that we would only get the first matching result then - that's what we have `getQueriesData` etc. for.

Ok then let's do one param only, that is _just_ the queryKey, and that is required then.

Mamoanwar97: sure 🙌 should I apply the same concept on `getQueryState` ? I think it is similar to `getQueryData`

TkDodo: yes, they are pretty much the same :)

Mamoanwar97: Done ✅ also updated the migration docs to include this change 🙌

DamianOsipiuk (approved):

examples/solid/basic-typescript/src/index.tsx
DamianOsipiuk: I vote for merging it if it's good enough for now.
This PR is massive and it would be hard to keep up to date, when other PRs start merging to v5.

TkDodo: yes, you're right @DamianOsipiuk . There are already conflicts, sorry for that. Once they are fixed, I'll merge this PR next

ardeora: Yeah, I agree! The scope for the other `solid-query` v5 changes can be addressed in a separate PR

ardeora (approved):

examples/solid/basic-typescript/src/index.tsx
Mamoanwar97: the merge conflicts are fixed now 🫡

Mamoanwar97: @TkDodo merge conflicts are fixed 🫡

TkDodo:
@allcontributors add @Mamoanwar97 for code

TkDodo:
@ardeora merged. Please create a follow up PR to address the open solid todos (or just an issue if you'd want someone else to tackle it). @DamianOsipiuk I don't think there is anything open for vue, but if there is, let's also create a follow-up

Issue #4688, opened 2022-12-23
Combine keepPreviousData with placeholderData 


TkDodo:
as per this RFC:
- https://github.com/TanStack/query/discussions/4426

PR #4715, opened 2022-12-27
feat: remove `keepPreviousData` in favor of `placeholderData`

DamianOsipiuk:
BREAKING CHANGE: removed `keepPreviousData` in favor of `placeholderData` identity function

Closes: https://github.com/TanStack/query/issues/4688

@TkDodo Questions:
- do we want to default it to `true`? Meaning enabling this behavior by default with option to opt-out? I think we could make that switch later on if we decide to do that.
- do we expose identity function for users? I do not think it will bring too much, as users could use any identity function they have already in the project.


TkDodo:
> do we want to default it to true

[Tanner's poll](https://twitter.com/tannerlinsley/status/1603066586738216960) wasn't very conclusive. I would keep the default `placeholderData` to `undefined`

> do we expose identity function for users? I do not think it will bring too much, as users could use any identity function they have already in the project.

I would say exposing a function called `keepPreviousData` that is just an identity function might help with the migration (code-mod?). It would also be tree-shakeable if you don's use it, so there isn't a lot of drawback to expose it I think ?

docs/react/guides/migrating-to-react-query-5.md
```diff
@@ -26,3 +26,39 @@ if you still need to remove a query, you can use `queryClient.removeQueries({que
...
+To achieve the same functionality as `keepPreviousData`, we have added previous query `data` as an argument to `placeholderData` function.
+Therefore you just need to provide an identity function to `placeholderData`
+
+```diff
+useQuery({
+  queryKey,
+  queryFn,
+- keepPreviousData: true,
++ placeholderData: (previousData) => previousData
+});
```
TkDodo, line 43: let's also add in the diff here that `isPreviousData` is now `isPlaceholderData`

packages/query-core/src/queriesObserver.ts
```diff
@@ -155,29 +155,14 @@ export class QueriesObserver extends Subscribable<QueriesObserverListener> {
...
-          // return previous data from one of the observers that no longer match
-          const previouslyUsedObserver = unmatchedObservers[index]
-          if (previouslyUsedObserver !== undefined) {
-            return {
-              defaultedQueryOptions: options,
-              observer: previouslyUsedObserver,
-            }
-          }
-        }
+      (options) => {
```
TkDodo, line 165: will `placeholderData => placeholderData` work with `useQueries` ? I think this was one of the open questions from the RFC.

packages/query-core/src/types.ts
```diff
@@ -27,7 +27,9 @@ export interface QueryFunctionContext<
 
 export type InitialDataFunction<T> = () => T | undefined
 
-export type PlaceholderDataFunction<TResult> = () => TResult | undefined
+export type PlaceholderDataFunction<TQueryFnData> = (
+  previousData: TQueryFnData | undefined,
+) => TQueryFnData | undefined
```
TkDodo, line 32: to be consistent, I think the generic should be named `TQueryData` here. It's what we use at the call-side. The only difference is that `TQueryFn` data is _exactly_ what the `queryFn` returns, while `TQueryData` might be a different structure - mainly for infinite queries. For normal queries, those two are identical.

packages/react-query/src/__tests__/useQueries.test.tsx
```diff
@@ -84,15 +84,15 @@ describe('useQueries', () => {
         queries: [
           {
             queryKey: [key1, count],
-            keepPreviousData: true,
+            placeholderData: (previousData?: number) => previousData,
```
TkDodo, line 87: is the type for `previousData` not inferred here?

TkDodo (commented):
very nice work 🚀

packages/react-query/src/__tests__/useQueries.test.tsx
DamianOsipiuk: This is potentially some kind of bug in TS.
When you check the type of `placeholderData`, it shows the proper type with `number` inferred.
But marks the function parameter as `any`, like it would lost the inferred context somehow.
Tried updating TS to latest with no success.

image

packages/query-core/src/types.ts
DamianOsipiuk: ✅

docs/react/guides/migrating-to-react-query-5.md
DamianOsipiuk: ✅

packages/query-core/src/queriesObserver.ts
DamianOsipiuk: So from what i found it worked this way:
- find all queries in the array that have the same `queryKey` as in previous pass - keep those observers intact
- find all queries in the array that were `present` in previous pass, but are no longer there (by `queryKey`) - unsubscribe from those observers
- find all queries in the array that were `not present` in previous pass, but were added to the current array (by `queryKey`) - subscribe to those observers

What `keepPreviousData` did before is to `reuse` observer that should be removed (unsubscribed) for queries that were added.
Although this seems like a good idea and it actually works for queries that have the same data structure, i do not think we should keep this behavior and therefore removed it (you also mentioned it in the RFC).

Since we do not actually know what kind of data structure new queries would have, we have no way of matching them with previous queries. Current method `might work`, but we can also end up with query returning `object` from `placeholder` and then `array` from `queryFn`.

packages/react-query/src/__tests__/useQueries.test.tsx
TkDodo: seems like this is specific to `useQueries` only? Maybe @artysidorenko can give us some help here :)

packages/query-core/src/queriesObserver.ts
TkDodo: yeah I think this is the right call 👍 

does it make sense to change the structure of `placeholderData` (at least on type level) then to `() => TQueryData` and _not_ pass `previousData` to it?

Also, it seems like we still have a lot of tests for the old keepPreviousData? I've also seen that they are flaky from time to time - are they still needed ?

packages/react-query/src/__tests__/useQueries.test.tsx
artysidorenko: Hi - yes unfortunately it's a limitation that's been appearing on all the `useQueries` callback parameters from the beginning ([sorta-vaguely-documented here in the tests](https://github.com/TanStack/query/blob/main/packages/react-query/src/__tests__/useQueries.test.tsx#L674-L717)). Similar to what you saw: it's really strange how the IDE/intellisense suggests the correct type on hover, but it still complains about `no-implicit-any`.

I wasn't really sure why it's happening, but you've motivated me to put together a simplified example and share it as a [TS issue](https://github.com/microsoft/TypeScript/issues/52047) - hopefully will provide some insight 🤞

TkDodo: thanks @artysidorenko - I've subscribed to this issue. Nothing more we can do here I think for now.

Andarist: The reported issue can be implemented using so-called "reverse mapped types". You can see the solution to the reported issue [here](https://github.com/microsoft/TypeScript/issues/52047#issuecomment-1367642953)

It would be fun for me to play around with your types in an attempt to make them work. I'm afraid though that your types are quite complex and I don't have time to figure out what they currently do on my own. Would somebody from your team like to jump on a call with me and spike this someday?

TkDodo: I would love to, but I don't think I understand those types enough to be of much help. @artysidorenko is the wizard who wrote them. I'd love to be a fly on the wall when you're talking about it 😁

Andarist: I concluded that the main problem here, for your real thing, is that you have multiple "generics" per a tuple element. Unfortunately, reverse mapped types can't handle this too well.

They can only infer directly into `T[K]` - but when we try to treat the whole tuple element as what we try to infer then we end up with the original problem as we can't relate different properties of what we have inferred as `T[K]` as easily.

I tried quite a bit to provide the argument's type for the `handler` in a different way. My main bets were different kinds of intersections*, other mapped types wrapping the argument position, and even circular constraints for `T`. None of those turned out to work here.

I somewhat strongly believe that this improvement would make it quite trivial to implement: https://github.com/microsoft/TypeScript/issues/51612

*I even tried to intersect multiple reverse mapped types. This didn't work out but it feels like **maybe** it wouldn't be that hard to make it work in the compiler. Such things often don't work because more advanced logic is missing intersections/unions handling and adding that is often not that hard.

packages/query-core/src/queriesObserver.ts
DamianOsipiuk: Removed tests and changed the `placeholderData` signature for `useQueries`.
When we remove `context` we can potentially move `UseQueryOptionsForUseQueries` to core to be shared across packages.

DamianOsipiuk:
@TkDodo @artysidorenko So when i tried to come up with identity function, it somehow messes up with the types returned from `useQuery` :(

```ts
// This works - const state: UseQueryResult<number, unknown>
const query1 = useQuery({
  queryKey: [key, count],
  queryFn: async () => {
    return Promise.resolve(count)
  },
  placeholderData: (previousData) => previousData,
})

// This is messed up - const query2: UseQueryResult<unknown, unknown>
function keepPreviousData<T>(previousData: T | undefined): T | undefined {
  return previousData
}
const query2 = useQuery({
  queryKey: [key, count],
  queryFn: async () => {
    return Promise.resolve(count)
  },
  placeholderData: keepPreviousData,
})
```

If i change this type definition from `QueryObserverOptions`:

```ts
placeholderData?: TQueryData | PlaceholderDataFunction<TQueryData>
```

to https://github.com/TanStack/query/pull/4715/files#diff-988cd43951b30185641ccb757eefd39c6e67027d7ccab1f445117b4c0401dc8bR251

```ts
type NonFunctionGuard<T> = T extends Function ? never : T

placeholderData?: NonFunctionGuard<TQueryData> | PlaceholderDataFunction<NonFunctionGuard<TQueryData>>
```

It works as intended.

Therefore i assume that Typescript might be confused cause in `TQueryData | PlaceholderDataFunction<TQueryData>` if fails to narrow down the type of `TQueryData` to not be a function and therefore falls back to unknown.
Not sure if this is another TS bug or limitation.

What do you think about this solution?

packages/react-query/src/__tests__/useQueries.test.tsx
artysidorenko: Hi @Andarist, thanks for looking into this 🙏 

> the main problem here, for your real thing, is that you have multiple "generics" per a tuple element

Yep, indeed, this is what I kept running into. In the end abandoned mapped types entirely, and replaced with [that bulky utility type `QueriesOptions`](https://github.com/TanStack/query/blob/9d9aea5fb12eb89dec54c619845b3d226b53cf2b/packages/react-query/src/useQueries.ts#L98), to walk through the argument array recursively and infer each of those "multi-generic" tuple elements, one at a time. It sort-of does the job, but as you notice it's not great readability-wise, there's a fair few sticking plasters to cover edge cases, and some downright weirdness, like the no-implicit-any issue here.

Let me know if you want to discuss further - my coverage isn't great at the moment (on vacation - coming to you from a hotel lobby in Mexico) but back home and more responsive from 13th Jan.

Andarist: I think I'm going to continue poking around with those reverse mapped types in the compiler. I have 2 ideas on how this could be solved and I will attempt to implement both of them in the coming months. Stay tuned 😉

Andarist: I **think** that i’ve found a patch for the compiler that would allow is to rewrite those types and make them easier and make those function arguments inferrable.

Let’s schedule this call to spike a potential rewrite 🚀

packages/query-core/src/types.ts
```diff
@@ -27,7 +27,15 @@ export interface QueryFunctionContext<
...
-export type PlaceholderDataFunction<TResult> = () => TResult | undefined
+type NonFunctionGuard<T> = T extends Function ? never : T
+
+export type PlaceholderDataFunction<TQueryData> = (
+  previousData: TQueryData | undefined,
+) => TQueryData | undefined
+
+export type QueriesPlaceholderDataFunction<TQueryData> = () =>
+  | TQueryData
+  | undefined
```
TkDodo, line 38: let's also document please that `useQueries` doesn't get previousData passed because the feature doesn't really work reliably with the dynamic nature of useQueries

TkDodo (approved):
changes look good to me - thanks for also using `keepPreviousData` internally for our tests 🎉

TkDodo:
oh, there are conflicts because of the overloads removal 😅

packages/query-core/src/types.ts
DamianOsipiuk: ✅

TkDodo:
can you please resolve the conflicts here, too?

DamianOsipiuk:
> can you please resolve the conflicts here, too?

Done ✅

docs/react/guides/migrating-to-v5.md
```diff
@@ -150,18 +150,60 @@ If you want to throw something that isn't an Error, you'll now have to set the g
...
+  queryKey,
+  queryFn,
+- keepPreviousData: true,
++ placeholderData: keepPreviousData
+});
+```
+
+There are some caveats to this change however, which you must be aware of:
+
+- `placeholderData` will always put you into `success` state, while `keepPreviousData` will give you the status of the previous query. That status could be `error` if we have data fetched successfully and then got a background refetch error. However, the error itself is not shared. This doesn't seem quite right in any case. Therefore we decided to stick with behavior of `placeholderData`.
```
TkDodo, line 189:
```suggestion
- `placeholderData` will always put you into `success` state, while `keepPreviousData` gave you the status of the previous query. That status could be `error` if we have data fetched successfully and then got a background refetch error. However, the error itself was not shared, so we decided to stick with behavior of `placeholderData`.
```

docs/react/guides/migrating-to-v5.md
```diff
@@ -150,18 +150,60 @@ If you want to throw something that isn't an Error, you'll now have to set the g
...
+- `keepPreviousData` will give you the `dataUpdatedAt` timestamp of the previous data, while with `placeholderData`, `dataUpdatedAt` will stay at `0`. This might be annoying if you want to show that timestamp continuously on screen. However you might get around it with `useEffect`.
```
TkDodo, line 190:
```suggestion
- `keepPreviousData` gave you the `dataUpdatedAt` timestamp of the previous data, while with `placeholderData`, `dataUpdatedAt` will stay at `0`. This might be annoying if you want to show that timestamp continuously on screen. However you might get around it with `useEffect`.
```

TkDodo (approved):

packages/react-query/src/__tests__/useQueries.test.tsx
TkDodo: @Andarist @artysidorenko let me know if I can help here. Which time would work for you ? I'm thinking that Discord might be a better way to communicate. Feel free to contact me there

artysidorenko: Sounds good - pinged you on discord just now.

Also keeping an eye on andarist's [TS PR currently in code review](https://github.com/microsoft/TypeScript/pull/52062). Looks like [the example given in the PR test](https://github.com/microsoft/TypeScript/pull/52062/files#diff-d9f6d4a38129503d13ff5d4c592a3402d123eea76e86254a5479ffebdc50b909) fits the useQueries use case pretty closely 🤞 once that's merged I think we'll be unblocked here.

jyuloeng:
Hi @TkDodo ~ Sorry to ask, but my project relies heavily on `keepPreviousData` now, so in v5, should I use such as
```ts
const query1 = useQuery({
  queryKey: [key, count],
  queryFn: async () => {
    return Promise.resolve(count)
  },
  placeholderData: (previousData) => previousData,
})
```
this method instead?

TkDodo:
@jyuloeng yes that's correct. For convenience, you can also do:

```
import { placeholderData } from '@tanstack/react-query`

placeholderData: keepPreviousData
```

it's just an identity function that we are exporting.

packages/query-core/src/queriesObserver.ts
arcanis: Sorry for the necropost, but I just hit the issue where `useQueries` doesn't support `keepPreviousData` due to not receiving the previous data, and it's not clear from the migration guide what's the expected workaround when the data shape doesn't change 🤔

Issue #4675, opened 2022-12-23
remove custom logger


TkDodo:
### context

In v3, we had a global logger that you could set via `setLogger` to your own logging mechanism. This was a side effect which we tried to get rid of, so in v4, you can now pass a `logger` prop to the `QueryClient`.

The logger mainly does one thing: It logs failed queries to the console. This is fine for development, but it was confusing to many that it also showed up in production.

That's why we've removed all logging in production in v4, and we've also started to use the logger more to show development specific warnings, for example:

- when you pass a `queryKey` that isn't an Array.
- when you return `undefined` from your queryFn.
- when you set multiple, conflicting query defaults.

Those logs are meant to help developers find potential issues in their code. They are not meant to be shown in production. I'd see them on the same level as the warning you get from React in development mode, for example:

- Each child in a list should have a unique "key" prop.

There is also no way to show these warnings in production. This also helps with bundle size because all logging happens behind an env check, so it is stripped from the final bundle. This means we can be as verbose as we want with those messages.

---

All of this means that the custom logger is kind of unnecessary. Why would you want to pass a custom logger only to log differently in development mode? The `logger` prop itself cannot be tree-shaken, so it will always be included in the final bundle even though we actually never use it.

### proposal

- remove the `logger` prop from `QueryClient`
- do not log failed queries to the console anymore, as failed queries are not a programmer "error" that we need to point people towards. The network tab shows failed requests just fine.

We should be able to just use the `console` for those development logs as it exists in all environments.

saurabhan:
You'd like to remove the logger prop from the QueryClient, but keep the defaultLogger functionality for logging programmer errors, right? just a quick question would returning defaultLogger from getLogger() function suffice or should I refactor the call everywhere?

TkDodo:
I would replace the logger completely and just use console.log instead. All logging should happen behind env checks.

saurabhan:
Got it.

TkDodo:
do you want to work on this @saurabhan ?

saurabhan:
Yes, I'd like to help with this.

bjesuiter:
How does the logging env check work? 
for example, when I'm running a npm module in deno the required envs might not be there as expected. 

Or another Szenario: running on deno deploy and imported via esm.sh

TkDodo:
we already have env checks around each log statement, e.g.:

[...]

PR #4721, opened 2022-12-28
[Refactor] Removed Custom Logger #4675

saurabhan:
Removed custom Logger and moved to console.log and console.error

Issue : [#4675](https://github.com/TanStack/query/issues/4675)


packages/query-core/src/query.ts
```diff
@@ -355,7 +350,7 @@ export class Query<
 
     if (!Array.isArray(this.options.queryKey)) {
       if (process.env.NODE_ENV !== 'production') {
-        this.logger.error(
+        console.error(
           `As of v4, queryKey needs to be an Array. If you are using a string like 'repoData', please change it to an Array, e.g. ['repoData']`,
         )
       }
```
TkDodo, line 356: just seeing this coincidentally: Can we move the `!Array.isArray` check inside the `process.env` check so that it can be reliably removed by bundlers. Maybe it works anyways because the if statement is empty, but I'd still rather have the order switched.

packages/query-core/src/queryClient.ts
```diff
@@ -68,13 +65,12 @@ export class QueryClient {
...
     this.queryDefaults = []
     this.mutationDefaults = []
 
-    if (process.env.NODE_ENV !== 'production' && config.logger) {
-      this.logger.error(
+    if (process.env.NODE_ENV !== 'production') {
+      console.error(
         `Passing a custom logger has been deprecated and will be removed in the next major version.`,
       )
     }
```
TkDodo, line 76: this is a deprecation warning for v4 that isn't necessary with v5 anymore. Please remove it.

packages/query-core/src/tests/queriesObserver.test.tsx
```diff
@@ -50,11 +50,11 @@ describe('queriesObserver', () => {
...
 
-    expect(mockLogger.error).toHaveBeenCalledTimes(2)
-    expect(mockLogger.error).toHaveBeenCalledWith(
+    expect(console.error).toHaveBeenCalledTimes(2)
+    expect(console.error).toHaveBeenCalledWith(
       'Passing a custom logger has been deprecated and will be removed in the next major version.',
     )
-    expect(mockLogger.error).toHaveBeenCalledWith(
+    expect(console.error).toHaveBeenCalledWith(
       'Passing a custom logger has been deprecated and will be removed in the next major version.',
```
TkDodo, line 58: I think we have to put a spy on the console and then check if his has been called. We don't want the console to be spammed with errors during tests, as this will make it harder to find actual errors.

we have something similar e.g. here:

https://github.com/TanStack/query/blob/c376158afaddf7cd92a85491c3f18c1b123efe4b/packages/react-query/src/__tests__/ssr-hydration.test.tsx#L58-L59

At the end of the test, mocks need to be restored:

https://github.com/TanStack/query/blob/c376158afaddf7cd92a85491c3f18c1b123efe4b/packages/react-query/src/__tests__/ssr-hydration.test.tsx#L134

this would be needed for all tests please

packages/query-core/src/tests/utils.ts
```diff
@@ -6,7 +6,7 @@ import * as utils from '../utils'
 
 export function createQueryClient(config?: QueryClientConfig): QueryClient {
   jest.spyOn(console, 'error').mockImplementation(() => undefined)
```
TkDodo, line 8: this needs to be removed from here I thinks

packages/query-core/src/tests/utils.ts
```diff
@@ -6,7 +6,7 @@ import * as utils from '../utils'
...
-  return new QueryClient({ logger: mockLogger, ...config })
+  return new QueryClient({ ...config })
```
TkDodo, line 9:
```suggestion
  return new QueryClient(config)
```

packages/query-persist-client-core/src/__tests__/utils.ts
```diff
@@ -3,13 +3,7 @@ import { QueryClient } from '@tanstack/query-core'
...
 export function createQueryClient(config?: QueryClientConfig): QueryClient {
   jest.spyOn(console, 'error').mockImplementation(() => undefined)
-  return new QueryClient({ logger: mockLogger, ...config })
-}
-
-export const mockLogger = {
-  log: jest.fn(),
-  warn: jest.fn(),
-  error: jest.fn(),
+  return new QueryClient({ ...config })
```
TkDodo, line 6: same here, remove the spy here, and the spread is unnecessary now

packages/react-query-persist-client/src/__tests__/utils.ts
```diff
@@ -5,13 +5,7 @@ import { QueryClient } from '@tanstack/query-core'
...
   jest.spyOn(console, 'error').mockImplementation(() => undefined)
-  return new QueryClient({ logger: mockLogger, ...config })
-}
-
-export const mockLogger = {
-  log: jest.fn(),
-  warn: jest.fn(),
-  error: jest.fn(),
+  return new QueryClient({ ...config })
 }
```
TkDodo, line 9: and here as well

packages/react-query/src/__tests__/utils.tsx
```diff
@@ -47,7 +47,7 @@ export const Blink = ({
 
 export function createQueryClient(config?: QueryClientConfig): QueryClient {
   jest.spyOn(console, 'error').mockImplementation(() => undefined)
-  return new QueryClient({ logger: mockLogger, ...config })
+  return new QueryClient({ ...config })
 }
```
TkDodo, line 51: once more 😅

packages/solid-query/src/__tests__/utils.tsx
```diff
@@ -31,7 +31,7 @@ export const Blink = (
 
 export function createQueryClient(config?: QueryClientConfig): QueryClient {
   jest.spyOn(console, 'error').mockImplementation(() => undefined)
-  return new QueryClient({ logger: mockLogger, ...config })
+  return new QueryClient({ ...config })
 }
```
TkDodo, line 35: same

rollup.config.ts
```diff
@@ -367,7 +369,7 @@ function cjs({
       forceDevEnv ? forceEnvPlugin('development') : undefined,
       replace({
         // TODO: figure out a better way to produce extensionless cjs imports
-        "require('./logger.js')": "require('./logger')",
+        // "require('./logger.js')": "require('./logger')",
```
TkDodo, line 372: feel free to remove those instead of commenting them out as those files no longer exist.

TkDodo (commented):
thanks, this looks great! Some work mainly around tests needs to be done, please see my comments.

packages/query-core/src/tests/queriesObserver.test.tsx
saurabhan: I have made the Recommended Changes and implemented console mock everywhere a console.error call was made in tests.
somehow some tests are failing and the console output is a mess

image

TkDodo: I think we're still missing this from the issue description:

> do not log failed queries to the console anymore, as failed queries are not a programmer "error" that we need to point people towards. The network tab shows failed requests just fine.

TkDodo: This will also make it easier for people to test erroneous queries because they don't need to turn off the logger anymore - and we also don't need to do that internally then for every failed query :)

saurabhan: got it 😅, will do the required changes, thanks for being patient.

packages/query-core/src/tests/utils.ts
saurabhan: Removing this somehow leads to consoleMock spy to not work, after removing this I had to add this in every test file to stop consoles from spilling while testing, not sure if it is the correct way,
```tsx
let consoleMock : jest.SpyInstance

  beforeEach(() => {
    consoleMock = jest.spyOn(console, 'error').mockImplementation(() => undefined)
  })

  afterEach(() => {
    consoleMock.mockRestore()
  })
```

not adding this results in console errors to be logged even if the tests are passing.

packages/query-core/src/tests/queryClient.test.tsx
```diff
@@ -124,6 +124,8 @@ describe('queryClient', () => {
     })
 
     test('should warn in dev if several query defaults match a given key', () => {
```
saurabhan, line 126: This test keeps on failing as the console to called lesser times than expected.

packages/query-core/src/tests/utils.ts
TkDodo: also not sure that this is correct. Tests shouldn't generally write to the console, so what did show up on the console without mocking it?

There are some tests that explicitly do something wrong where an error is expected where we do need to mock the console, but those should be the minority.

This implies that you remove the general error logging, otherwise, every erroneous query spams the console, and we do test error states quite a bit :)

TkDodo: yep, it's still here. That's the reason:

Screenshot 2022-12-30 at 11 02 11

packages/query-core/src/tests/queryClient.test.tsx
TkDodo: probably because we don't have this warning anymore?

https://github.com/TanStack/query/pull/4721/files/09fe1ea6045cb9495841f8c782a7da78715122f0#diff-bf605a2d8d2cf086c1781865ec3cb71ebc3726f953e6f992d520978690948bfcR224-R227

maybe we need to just decrease the count by one?

packages/query-core/src/tests/utils.ts
TkDodo: also for this test, we'd probably need to mock `console.warn` now:

Screenshot 2022-12-30 at 11 04 11

we don't need to worry about the tests that log further deprecation warnings, because those warnings will be removed when we remove the deprecated features, which is also planned for v5. So now we are in the situation where multiple features are worked on at the same time that kind of intersect 😅 . Maybe we should wait with merging the logger-removal feature until all the other features that use the logger for deprecation warnings have been merged.

saurabhan: Good call, as most of the consoles are for deprecations, until then I will dive a little more in the source to get better understanding of things :)

I will try removing the general error logging tonight and try again, from the above screenshot I guess that it is referncing to the query.ts 

```tsx
 if (!isCancelledError(error)) {
        // Notify cache callback
        this.cache.config.onError?.(error, this as Query<any, any, any, any>)

           console.error(error)
         }
      }
 ```
 and mutation.ts 
 
 ```tsx
          console.error(error)
        }
  ```
  
  also found out that some tests does not explicitly call console in them but the function they are using calls a console internally and that skips the mock implementation and gets logged to the console. like this test is queryObserver.test.tsx
  
  ```tsx
  test('should prefer isDataEqual to structuralSharing', async () =>{}
  ```
  
  wrapping this test in mockConsole solves the error being printed to the console as the test is passing.

**Comment by TkDodo** [comment]
@saurabhan are you going to keep working on this PR or do you want someone else to take over?

**Comment by saurabhan** [comment]
> 

I'm sorry, but I am on-call at work for the next week.
If anyone wants to pick this up in the meantime they can otherwise I will pick it up after my on-call ends.

Also, I tried but the mocking console inside the test still prints it to the terminal, writing it at the top of the tests like this
```tsx
  let consoleMock : jest.SpyInstance;
  beforeEach(() => {
   consoleMock =  jest.spyOn(console, 'error').mockImplementation(() => {})
  })
  
  afterEach(() => {
    consoleMock.mockRestore()
  })
  ```
  
  This works and stops the console from spilling over in the terminal. had to do this in every test file like mutations.test.tsx does not call the console explicitly but a function it calls from mutation.ts calls the console.error, doing the above stops console spill and tests pass as usual.

**Comment by TkDodo** [comment]
I updated with v5 and removed the error logging when a query errors. That means what is left is going through tests and mostly revert a lot of your changes, because no console logging actually happens when we have tests for error states. I started with that but ran out of time. To be continued.

**Comment by saurabhan** [comment]
Hi TkDodo, can you please have a look and tell me what all changes are to be made or reverted, I will get them done. Thanks

**Comment by TkDodo** [comment]
> Hi TkDodo, can you please have a look and tell me what all changes are to be made or reverted, I will get them done. Thanks

I haven't looked at the errorBoundary tests yet, but it seems like we need to add a `spy` to the console whenever we use errorBoundary.

Apart from that, we _don't need_ to change anything for queries that just reject the Promise and go into error state without error boundaries. There, we used to log to the console, but I have removed that.

for example, in this test:
https://github.com/TanStack/query/pull/4721/files#diff-3cd8cae55f836a38ea976d6cbf9f1b59e8fdfa4d3218b123cb1f9dccfc6824d2R2867-R2870

We shouldn't need the log spy because there is no error logging that we would expect.

**Comment by saurabhan** [comment]
I have removed consoleMock from the test and fixed some of the tests, also added consoleMock where Error boundary is being used.

**Comment by TkDodo** [comment]
@saurabhan can you please update with v5? seems like there are lots of conflicts

**Comment by saurabhan** [comment]
Fixed linting and Format, 🤞

File: packages/query-core/src/tests/queryObserver.test.tsx [hunk 12]
```diff
@@ -554,7 +550,8 @@ describe('queryObserver', () => {
     // Should not log an error
     queryClient.clear()
     await sleep(40)
-    expect(mockLogger.error).not.toHaveBeenNthCalledWith(1, 'reject 1')
+    expect(consoleMock).not.toHaveBeenNthCalledWith(1, 'reject 1')
+    consoleMock.mockRestore()
```
TkDodo, line 554: it doesn't make any sense to check this assertion via the logger. We're not logging rejections anymore, so it will always be zero. Previously, there was a case where we wanted to test that, according to the comment:

```
    // A second reject is triggered for the retry
    // but the retryer has already set isResolved to true
    // so it does nothing and no error is thrown
```

as it stands, `consoleMock` will _never_ be called with `reject 1` so the test will always succeed. I'm not sure if we can test this any better or if we should just remove the test.

packages/react-query/src/__tests__/suspense.test.tsx
```diff
@@ -332,9 +335,20 @@ describe("useQuery's in Suspense mode", () => {
     fireEvent.click(rendered.getByText('retry'))
 
     await waitFor(() => rendered.getByText('rendered'))
+
+    // check if the console.error function was called
+    if (consoleMock.mock.calls[0]) {
+      //test the error message that was passed to the console
+      expect(consoleMock.mock.calls[0][0]).toContain('Suspense Error Bingo')
+    }
```
TkDodo, line 343: why a conditional expect? If the console mock is never called, we just don't expect anything and the test will run through fine ...

packages/react-query/src/__tests__/suspense.test.tsx
```diff
@@ -817,9 +847,19 @@ describe("useQuery's in Suspense mode", () => {
     fireEvent.click(rendered.getByLabelText('fail'))
     // render error boundary fallback (error boundary)
     await waitFor(() => rendered.getByText('error boundary'))
+    // check if the console.error function was called
+    if (consoleMock.mock.calls[0]) {
+      //test the error message that was passed to the console
+      expect(consoleMock.mock.calls[0][0]).toContain('Suspense Error Bingo')
+    }
```
TkDodo, line 854: same here. expect calls shouldn't be done conditionally imo

packages/react-query/src/__tests__/suspense.test.tsx
```diff
@@ -876,9 +916,18 @@ describe("useQuery's in Suspense mode", () => {
     fireEvent.click(rendered.getByLabelText('fail'))
     // render error boundary fallback (error boundary)
     await waitFor(() => rendered.getByText('error boundary'))
+    if (consoleMock.mock.calls[0]) {
+      //test the error message that was passed to the console
+      expect(consoleMock.mock.calls[0][0]).toContain('Suspense Error Bingo')
+    }
```
TkDodo, line 922: there is even a lint rule for that: [no-conditional-expect](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-conditional-expect.md)

packages/react-query/src/__tests__/suspense.test.tsx
```diff
@@ -936,6 +985,12 @@ describe("useQuery's in Suspense mode", () => {
 
     // render error boundary fallback (error boundary)
     await waitFor(() => rendered.getByText('error boundary'))
+    if (consoleMock.mock.calls[0]) {
+      //test the error message that was passed to the console
+      expect(consoleMock.mock.calls[0][0]).toContain('Suspense Error Bingo')
+    }
```
TkDodo, line 991: `no-conditional-expect`

packages/react-query/src/__tests__/useMutation.test.tsx
```diff
@@ -722,10 +725,24 @@ describe('useMutation', () => {
...
+      // expect(consoleMock).toHaveBeenCalledTimes(1)
     })
+
+    // check if the console.error function was called
+    if (consoleMock.mock.calls[0]) {
+      //test the error message that was passed to the console
+      expect(consoleMock.mock.calls[0][0]).toContain(
+        'Expected mock error. All is well!',
+      )
+    }
```
TkDodo, line 737: `no-conditional-expect`

packages/react-query/src/__tests__/useQuery.test.tsx
```diff
@@ -1120,11 +1119,11 @@ describe('useQuery', () => {
       rendered.getByText('error')
     })
 
-    expect(mockLogger.error).toHaveBeenCalledWith(error)
     expect(states.length).toBe(2)
 
     expect(states[0]).toMatchObject({ status: 'loading', data: undefined })
-    expect(states[1]).toMatchObject({ status: 'error', error })
+    // expect(states[1]).toMatchObject({ status: 'error', error })
```
TkDodo, line 1125:
```suggestion
```

packages/vue-query/src/__tests__/useQueries.test.ts
```diff
@@ -159,6 +159,8 @@ describe('useQueries', () => {
   })
 
   test('should stop listening to changes on onScopeDispose', async () => {
+    const consoleMock = jest.spyOn(console, 'error')
+    consoleMock.mockImplementation(() => undefined)
```
TkDodo, line 163: why do we need a mock here? it doesn't look like the test would log anything on purpose?

packages/vue-query/src/__tests__/vueQueryPlugin.test.ts
```diff
@@ -296,6 +296,8 @@ describe('VueQueryPlugin', () => {
     })
 
     test('should delay useQueries subscription and not call fetcher if data is not stale', async () => {
+      const consoleMock = jest.spyOn(console, 'error')
+      consoleMock.mockImplementation(() => undefined)
```
TkDodo, line 300: why do we need a mock here? it doesn't look like the test would log anything on purpose?

TkDodo (changes requested):
Thank you 🙏 . Please have a look at my comments

packages/react-query/src/__tests__/suspense.test.tsx
saurabhan: There were stack traces being logged as well, so the mock console was being called many times because of that, i was checking only the first error with this.

TkDodo:
```
.mockImplementation(() => {})
```

yields a lint error:

```
ESLint: Unexpected empty arrow function.(@typescript-eslint/no-empty-function)
```

let's just do `.mockImplmentation(() => undefined)` ?

packages/react-query/src/__tests__/suspense.test.tsx
TkDodo: even then, the `if` statement is unnecessary.

I've tried and this seems to work well:

```
expect(consoleMock).toHaveBeenCalledWith(
  expect.objectContaining(new Error('Suspense Error Bingo')),
)
```

packages/vue-query/src/__tests__/vueQueryPlugin.test.ts
saurabhan: There was a console error logged by a vue function, i think I missed it while I updated with v5, will fix and push. Thanks.

packages/query-core/src/tests/queryObserver.test.tsx
saurabhan: As of now I have removed the consoleMock and expect,
I guess we can take an approach where we are checking if the observer's 'isResolved' is set to true before the retry function is called, this would make sure that it is not attempting to retry if the query is resolved.

TkDodo: yeah now the test has no `expect` calls so it will always succeed.

> I guess we can take an approach where we are checking if the observer's 'isResolved' is set to true before the retry function is called, this would make sure that it is not attempting to retry if the query is resolved.

👍

saurabhan: not sure about this, but I was not able to check the status of the query as it always comes out to be loading, I tried failureCount which returns the amount of failed tires, and asserted that.

```tsx
    expect(result.failureCount).toBe(1);
    expect(result.failureReason).toBe("reject 1");
```

I guess we can also assert count as it will only be incremented only once.

TkDodo: I've just removed the test for now 🤷

TkDodo (approved):

Issue #2810, opened 2021-10-19
ContextSharing not working for different modules


reh2ur:
**Describe the bug**
When using the `QueryClientProvider` in an npm package and requiring the query client via `useQueryClient` in the consumer app, I am always getting the following error:
```
Error: No QueryClient set, use QueryClientProvider to set one
```

It seems like the logic for getting the shared context is broken.

**Expected behavior**
Using the `useQueryClient` should check for the shared `ReactQueryClientContext` and use it if existing.

**Additional context**
With #1912 the context sharing was disabled by default, and seems to have been disabled at all.
It seems like the `QueryClientSharingContext` is always false, as it is initialized with that value:
https://github.com/tannerlinsley/react-query/blob/05a8057e2be64915d2a12b716d8131f113598dbd/src/react/QueryClientProvider.tsx#L12
https://github.com/tannerlinsley/react-query/blob/05a8057e2be64915d2a12b716d8131f113598dbd/src/react/QueryClientProvider.tsx#L34

Shouldn't the context be initialized with whether there is a shared context or a boolean value, passed to the `useQueryClient` hook?

TkDodo:
right, I think you'd need a `QueryClientProvider` to turn contextSharing on, which kinda defeats the purpose 🤔 . Can you create a codesandbox example that works with a version _before_ #1912 was merged and doesn't work with the latest version? Thanks.

reh2ur:
We are experimenting with extracting shared logic from multiple NextJS projects into an internally used library. I am trying to create a reusable Wrapper component that encapsulates UI elements as well as support for data fetching. This `<Wrapper />` would take care of rendering the `QueryClientProvider` with an appropriate client, so that the consumer could just call `useQuery` without having to deal with the actual client.

The library has `react-query` as a peer dependency, which is why the `useQuery` hooks of the consumer app and the `QueryClientProvider` of our shared module habe the same `react-query` instances.

vsylvestre:
@reh2ur I'd be curious to have more details as to how you fixed this, since I'm in a very similar situation.

[...]

Issue #4682, opened 2022-12-23
replace custom context with custom queryClient


TkDodo:
### context

in v4, we introduced the possibility to pass a custom context to all react-query hooks. This allowed for proper isolation when using MicroFrontends. See [the migration guide](https://tanstack.com/query/v4/docs/guides/migrating-to-react-query-4#custom-contexts-for-multiple-providers) for examples and details.

However, context is a react only feature. All that context does is give us access to the query client. We could achieve the same isolation by allowing to pass in a custom query client instead of a custom context that then serves the query client.

vue-query does have this api already, so it makes sense to streamline it in react as well.

see also:
- https://github.com/TanStack/query/discussions/4360

### proposal

- replace custom context param on react hooks with custom queryClient param

DamianOsipiuk:
Blocked by https://github.com/TanStack/query/issues/4694

Issue #4679, opened 2022-12-23
size improvements


TkDodo:
### context

The "big" bundle size is a constant topic when TanStack Query is compared to other libraries. I think there are a few things that can be done to improve the situation:

#### don't transpile optional chaining

Optional chaining is used a lot in the codebase, and it's supported in [93.44% of all browsers](https://caniuse.com/mdn-javascript_operators_optional_chaining). Transpiling it is quite costly:

```diff
- this.retryer?.continue()
+ this.retryer == null ? void 0 : this.retryer.continue()
```

#### switch to private class fields and methods

Private class fields are supported in [92.7% of all browsers](https://caniuse.com/mdn-javascript_classes_private_class_fields). They have a bunch of advantages:

- they are private at runtime, not just at compile time
- because of that, they can be minified by terser
- as per [this discussion](https://github.com/TanStack/query/discussions/4026), they would help us with circular references to private fields from within the QueryCache.

### proposal

- adapt our current supported browserslist:

```diff
- Chrome >= 73
+ Chrome >= 84
- Firefox >= 78
+ Firefox >= 90
- Edge >= 79
+ Edge >= 84
- Safari >= 12.0
+ Safari >= 15
- iOS >= 12.0
+ iOS >= 15
- opera >= 53
+ opera >= 70
```

Out of these changes, `Safari` would be the "newest" supported browser, with a release date of September 2021. This would still likely mean at least 1.5 years of browser support for Safari when we release v5.

If people want to support older browsers, they can always transpile the code themselves.

- use private class fields and methods instead of the `private` TypeScript keyword.

DamianOsipiuk:
@TkDodo After quick test i found out that the minimal set of browsers we need are:
```
Chrome >= 91
ChromeAndroid >= 91
Firefox >= 90
Edge >= 91
Safari >= 15
iOS >= 15
opera >= 77
```

Otherwise `babel` refuses to keep `optional chaining` and `private class fields` intact.

This covers `89.8 %` users globally according to [browsersl.ist](https://browsersl.ist/#q=Chrome+%3E%3D+91%2CChromeAndroid+%3E%3D+91%2CFirefox+%3E%3D+78%2CEdge+%3E%3D+91%2CSafari+%3E%3D+14%2CiOS+%3E%3D+14%2Copera+%3E%3D+77)

TkDodo:
[...]

PR #4723, opened 2022-12-28
refactor: remove contextSharing

### LINKED ISSUE #4681: remove contextSharing
### context

The `QueryClientProvider` has a prop `contextSharing: boolean`. The docs say:

> Set this to `true` to enable context sharing, which will share the first and at least one instance of the context across the window to ensure that if React Query is used across different bundles or microfrontends they will all use the same **instance** of context, regardless of module scoping.

To be honest - I don't really know how this property is working. There were some discussion on [this issue](https://github.com/TanStack/query/issues/2810) that suggest that it is not really useful.

For microfrontends, isolation is often preferred. With v4, we introduced the option to pass a [custom context](https://tanstack.com/query/v4/docs/guides/migrating-to-react-query-4#custom-contexts-for-multiple-providers), which allows for exactly that.

If you want your app to use the same client when it's composed of multiple packages, all you'd need to do is create one QueryClient in your app and let the different bundles pick those up. As long as they all use the same version of TanStack Query, this should work fine.

### proposal

- ✅ Deprecate `contextSharing` in v4
- Remove `contextSharing` from `QueryClientProvider` in v5

tludlow:
This PR attempts to close #4681, one of the v5 react query milestone issues.

I have:

- Removed `contextSharing` from react-query and updated / removed the tests.
- Removed `contextSharing` from solid-query and updated / removed the tests.
- Removed `contextSharing` from vue-query and updated / removed the tests.
- Removed the `contextSharing` documentation from the `QueryClientProvider` reference documentation
- Added a section to the v5 migration guide about the removal of this property. I think this documentation could probably be better, I used information from both the issue #4681 and the linked discussion in #2810 to try and explain this. I suppose I could add an example of what I have described in this documentation, are there any good examples of how to document a code example for multiple packages like this within other parts of the documentation?


packages/react-query/src/QueryClientProvider.tsx
```diff
@@ -3,46 +3,22 @@ import * as React from 'react'
...
 
-// If we are given a context, we will use it.
-// Otherwise, if contextSharing is on, we share the first and at least one
-// instance of the context across the window
-// to ensure that if React Query is used across
-// different bundles or microfrontends they will
-// all use the same **instance** of context, regardless
-// of module scoping.
 function getQueryClientContext(
   context: React.Context<QueryClient | undefined> | undefined,
```
Moshyfawn, line 11: Would it make sense to make the type this instead?
```suggestion
  context: React.Context<QueryClient> | undefined,
```

packages/react-query/src/QueryClientProvider.tsx
```diff
@@ -3,46 +3,22 @@ import * as React from 'react'
...
   }
-  if (contextSharing && typeof window !== 'undefined') {
-    if (!window.ReactQueryClientContext) {
-      window.ReactQueryClientContext = defaultContext
-    }
-
-    return window.ReactQueryClientContext
-  }
 
   return defaultContext
```
Moshyfawn, line 17: What do you think about making it a simple `OR` instead? 
```suggestion
  return context || defaultContext
```

packages/react-query/src/QueryClientProvider.tsx
```diff
@@ -80,19 +49,7 @@ export const QueryClientProvider = ({
...
-
-  const Context = getQueryClientContext(context, contextSharing)
+  const Context = getQueryClientContext(context)
 
-  return (
-    <QueryClientSharingContext.Provider value={!context && contextSharing}>
-      <Context.Provider value={client}>{children}</Context.Provider>
-    </QueryClientSharingContext.Provider>
-  )
+  return <Context.Provider value={client}>{children}</Context.Provider>
```
Moshyfawn, line 54: Maybe we can call the `Context` by it's name now that there's no name clashing?
```suggestion
  const QueryClientContext = getQueryClientContext(context)

  return <QueryClientContext.Provider value={client}>{children}</QueryClientContext.Provider>
```

packages/react-query/src/QueryClientProvider.tsx
```diff
@@ -55,23 +31,16 @@ type QueryClientProviderPropsBase = {
...
-  contextSharing?: boolean
-} & QueryClientProviderPropsBase
 
-export type QueryClientProviderProps =
-  | QueryClientProviderPropsWithContext
-  | QueryClientProviderPropsWithContextSharing
+type QueryClientProviderPropsWithContext = ContextOptions &
+  QueryClientProviderPropsBase
+
+export type QueryClientProviderProps = QueryClientProviderPropsWithContext
```
Moshyfawn, line 38: Would it make sense to simply define the `QueryClientProviderProps` like
```suggestion
export type QueryClientProviderProps = {
  client: QueryClient
  children?: React.ReactNode
} & ContextOptions
```

Moshyfawn (commented):
I have a couple of suggestions; please, let me know what you think ;)

packages/react-query/src/QueryClientProvider.tsx
tludlow: Yeah this makes sense, I'll do it for the solid-query implementation too

packages/react-query/src/QueryClientProvider.tsx
tludlow: Yeah this will then match the solid-query context naming.

Moshyfawn (approved):
Easier to read and less convoluted, IMO. Thx, @tludlow!

packages/react-query/src/QueryClientProvider.tsx
Moshyfawn: Exactly!

Moshyfawn:
Why does clicking "Review Changes" on a specific commit approves the whole PR.. I'll never now xD

packages/react-query/src/QueryClientProvider.tsx
tludlow: I think this will need to stay as it is due to the `defaultContext` type definition + default value above.

Moshyfawn: I'm just thinking, if someone passed their custom context down, what's the case where it's `undefined`?

tludlow: The `defaultContext` has a default value of `undefined` hence the `React.Context<QueryClient | undefined>` type and the context argument for `getQueryClientContext` is optional, hence the second undefined. I could remove the `undefined` as you said if I were to make the default value for the `defaultContext` be something else but I imagine its done this way for a reason that I'm not familiar with.

packages/react-query/src/QueryClientProvider.tsx
tludlow: Yeah I can do this for the react and solid implementations. If there are any conventions I'm breaking by doing this then I'd love to know from one of the maintainers :)

packages/solid-query/src/QueryClientProvider.tsx
```diff
@@ -55,56 +25,26 @@ export const useQueryClient = ({ context }: ContextOptions = {}) => {
...
-  props: QueryClientProviderProps,
-): JSX.Element => {
-  const mergedProps = mergeProps(
-    {
-      contextSharing: false,
-    },
-    props,
-  )
+export const QueryClientProvider = ({
+  client,
```
DamianOsipiuk, line 34: I think that `destructuring` props in solid will break the reactivity.
@ardeora

Moshyfawn: You might be right
https://www.solidjs.com/tutorial/props_split

tludlow: Oh, thanks for pointing this out. I've never used solid, I'll put it back to how it was :)

ardeora (approved):
Thanks for working on this! This looks perfect :D

packages/react-query/src/QueryClientProvider.tsx
TkDodo: keep in mind that we'll also replace the feature of passing down a custom context with passing down a custom queryClient:

- https://github.com/TanStack/query/issues/4682

so I think it probably doesn't matter :)

packages/react-query/src/QueryClientProvider.tsx
TkDodo: `context ?? defaultContext` is what I'd prefer :)

docs/react/guides/migrating-to-react-query-5.md
```diff
@@ -26,3 +26,9 @@ if you still need to remove a query, you can use `queryClient.removeQueries({que
 ### The minimum required TypeScript version is now 4.7
 
 Mainly because an important fix was shipped around type inference. Please see this [TypeScript issue](https://github.com/microsoft/TypeScript/issues/43371) for more information.
+
+### The `contextSharing` prop has been removed from QueryClientProvider
+
+You could previously use the `contextSharing` property to share the first (and at least one) instance of the query client context across the window. This ensured that if TanStack Query was used across different bundles or microfrontends then they will all use the same instance of the context, regardless of module scoping.
+
+However, isolation is often preferred for microfrontends. In v4 the option to pass a custom context to the `QueryClientProvider` was added, which allows exactly this. If you wish to use the same query client across multiple packages of an application, you can create a `QueryClient` in your application and then let the bundles share this through the `context` property of the `QueryClientProvider`.
```
TkDodo, line 34: we'll need to change this again after we've shipped #4682 but it's good for now 👍

TkDodo (approved):

packages/react-query/src/QueryClientProvider.tsx
Moshyfawn: Doesn't `??` transpile to a tone of code adding bundle size?

TkDodo: not with the plans we have in:
- #4679

Moshyfawn: Read through it before, but didn't register that part for some reason! Hell yea, stocked!

TkDodo:
can you please fix the conflicts ?

tludlow:
> can you please fix the conflicts ?

Fixed.

Note to self: don't use the github web merge conflicts tool

PR #4736, opened 2023-01-02
(feat: solid-query): Horizontal upgrade v5

ardeora:
This PR aims to upgrade `solid-query` to v5 proposed changes. This change is a horizontal upgrade where no additional features (like SSR support, devtools) are added. The main changes addressed here are:

## Leverage removal of function overloads
The removal of function overloads allows solid query to get rid of the object getter syntax to track changes of signals passed in to primitives. This will lead to better composability and will bring consistency between internal functions like `createQuery`, `createQueries`, `createMutation`, etc. and external functions (that can live outside solid's reactive scope) like `queryClient.setQueryData`, `queryCache.find`, etc. This update also puts us in a very nice spot for supporting SSR frameworks like Solid Start, Astro and opens room for a better devtools experience. Here is a before (v4) and after (v5) example of how `createQuery` will work

### Before (v4)
```tsx
//  🔸 Query keys in outside solid's reactive scope are arrays
await queryClient.prefetchQuery({ queryKey: [ 'todos' ] )}
 
function Component() {
  const [enabled, setEnabled] = createSignal(false)
  const [todo, setTodo] = createSignal(0)

  const todosListQuery = createQuery({
    // 🔸 Same query key in solid query primitives is a getter function
    queryKey: () => ['todos'],
    queryFn: fetchTodos,
   // 🔸 Object getter syntax can get unruly as query becomes more dynamic
    get enabled() {
      return enabled()
    }
  })

  const todoDetailsQuery = createQuery({
    queryKey: () => ['todo', todo()],
    queryFn: fetchTodo,
    get enabled() {
      return todo() > 0
    }
  })

  return (
    <div>
      <For each={todosListQuery.data}>
        {(todo) => (
          <button onClick={() => setTodo(todo.id)}>{todo.title}</button>
        )}
      </For>
      <Show when={todoDetailsQuery.data}>
         {(todo) => (
           <div>{todo.body}</div>
         )}
      </Show>
    </div>
  )
}
```

### After (v5)
```tsx
//  🟩 Query keys in outside solid's reactive scope are arrays
await queryClient.prefetchQuery({ queryKey: [ 'todos' ] )}
 
function Component() {
  const [enabled, setEnabled] = createSignal(false)
  const [todo, setTodo] = createSignal(0)

  const todosListQuery = createQuery(() => ({
    // 🟩 Consistency in query keys! They are arrays everywhere
    queryKey: ['todos'],
    queryFn: fetchTodos,
   //🟩 Pass in Signals directly!
    enabled: enabled()
  }))

  const todoDetailsQuery = createQuery(() => {
    queryKey:  ['todo', todo()],
    queryFn: fetchTodo,
    enabled: todo() > 0
  })

  return (
    <div>
      <For each={todosListQuery.data}>
        {(todo) => (
          <button onClick={() => setTodo(todo.id)}>{todo.title}</button>
        )}
      </For>
      <Show when={todoDetailsQuery.data}>
         {(todo) => (
           <div>{todo.body}</div>
         )}
      </Show>
    </div>
  )
}
```

## Use `notifyManager` for batching updates
`solid-query` had a hacky way of batching updates from queryObserver results. This update uses `notifyManager` across all functions to keep all `TanStack Query` products consistent. This also unblocks us on some PRs that were dependent on this https://github.com/TanStack/query/pull/4295

## Removal of `normalizedOptions` functions
A remnant of the time when we had dozens of function overloads!

## Test cleanup
Tests are all cleaned up

Please let me know if you have any other questions/feedback or concerns!

Thanks, 
Aryan


examples/solid/simple/.eslintrc
```diff
@@ -6,6 +6,7 @@
   },
   "rules": {
     "react/react-in-jsx-scope": "off",
-    "jsx-a11y/accessible-emoji": "off"
+    "jsx-a11y/accessible-emoji": "off",
+    "@tanstack/query/prefer-query-object-syntax": "off"
```
TkDodo, line 10: good point - we should probably remove that rule for v5, because TypeScript will enforce the correct usage for us.
- https://github.com/TanStack/query/issues/4739

packages/solid-query/package.json
```diff
@@ -16,7 +16,7 @@
   "exports": {
     ".": {
       "types": "./build/lib/index.d.ts",
-      "solid": "./build/solid/index.jsx",
+      "solid": "./build/solid/index.js",
```
TkDodo, line 19: I think we'd now need to change `"sideEffects": false,` here in `package.json` to point to:

```
  "sideEffects": [
    "./src/setBatchUpdatesFn.ts"
  ],
```

right?

TkDodo (commented):
Great work, thank you 🚀

DamianOsipiuk:
I have just one comment about the commit messages.
They should be in the format, where `feat/fix/chore` etc. are **before** the brackets.
Ex. `feat(solid-query): Horizontal upgrade v5`
Otherwise semantic-release bot would not pick it up.

ardeora:
> I have just one comment about the commit messages. They should be in the format, where `feat/fix/chore` etc. are **before** the brackets. Ex. `feat(solid-query): Horizontal upgrade v5` Otherwise semantic-release bot would not pick it up.

Apologies for this! I fixed the commit messages

examples/solid/simple/.eslintrc
ardeora: Heh! I wouldn't have necessarily removed the lint rule but since the only argument now is a function. The lint rule complains because it thinks we're passing it a `queryFn`

packages/solid-query/package.json
ardeora: Yeah I missed this. I'll fix it in a new commit

ardeora: Done!

TkDodo:
I usually adapt the commit message when squashing anyways 😅

TkDodo (approved):

Issue #4689, opened 2022-12-23
Add 'use client' in the correct places


TkDodo:
To be compatible with React Server Components

@Ephem we'd need to decide if this can go into v4 or if we also do this for v5

Ephem:
At least on the surface this is an easy fix with no downsides I can see that helps early adopters play with the new APIs without a bunch of wrapping, so I think we should add it in a v4 minor. I haven't investigated the exact places to add it yet though, but that shouldn't be that tricky.

Suman085:
I would be happy to contribute here if I get more clarity on the work that needs to be done here :)

TkDodo:
@Suman085 we'd need to add `"use client"` at the top of all files that use either state, context or effects in order to have them opt out of being server components in next13.

the devtools would also naturally need `"use client"`.

abhijithp05:
Please close this issue

PR #4738, opened 2023-01-02
feat(client components): add `use client` directive at the top of files having client components

girishsontakke:
This PR is addressing the improvement suggested in #4689.


TkDodo (commented):
Thanks. I think there are still a couple of places missing like:
- `useIsFetching`
- `useIsMutating`
- the devtools
- the `react-query-persist` package

TkDodo:
not sure where all the conflicts are coming from. can you update with main once more please?

girishsontakke:
> not sure where all the conflicts are coming from. can you update with main once more please?

Sure I have removed the conflicts.

TkDodo (commented):
there are more things that should be client components. Basically, everything that uses state, effects or context needs to be a client component. I haven't looked at the full list, but:

- `hydrate.tsx` is definitely a client component
- isRestoring.tsx` has react context

TkDodo:
> I think this also means we might want to split any file that has both client components and hooks into two separate ones, like Hydrate.tsx and QueryClientProvider.tsx?

oh yeah, good idea. I think in v5 the `Hydrate` is no longer a problem because we have removed the hook already, but it makes sense for `QueryClientProvider`

Ephem:
I took the liberty of removing the `'use client'` from the hooks files again, but didn't split the files, I think we can do that later. I think `isRestoring.tsx` is another annoying candidate btw:

```tsx
'use client'
import * as React from 'react'

const IsRestoringContext = React.createContext(false)

export const useIsRestoring = () => React.useContext(IsRestoringContext)
export const IsRestoringProvider = IsRestoringContext.Provider
```

Any such context/useContext pattern will need to be split into two files to get good error messages for the `useContext` part..

Ephem (approved):

Issue #3706, opened 2022-06-13
fetchQuery and useQuery options 


artem-malko:
### Describe the bug

Actually, I'm not sure, that it is a bug.

So, if I have a fetchQuery and my own query (which uses useQuery) for the one entity (newsItem in the example), and my own query has specific options (fetchQuery uses default options, own query overrides cacheTime and staleTime). My own query executes on the NewsItem component mount. But, if a data for newsItem is fetched via fetchQuery, my own query is not working for that case. So, in a queryCache that query has options from default, not from my own query.

I understand, that my own query won't start its fetcher, if the data is in the cache. But it was quite unexpected, that options from my own query are not used. Check out the example from codesandbox.

### Your minimal, reproducible example

https://codesandbox.io/s/react-query-fetchquery-usequery-g18l4y

### Steps to reproduce

1. Go to the codesandbox example
2. Click to the button with text: "fetch newsitem with id ..." In the ReactQuery devtools you will see, that there is a query with cacheTime and staleTime from default settings.
3. Click to the button with text "Toggle newsItemComp". The NewsItem will be shown.
4. Options from the query "useNewsItem", which is used there won't be applied.

### Expected behavior

As a user, I expected that useQuery will rewrite options of the query in the reactQuery cache but i am seeing, that options from fetchQuery is used (default options).

### How often does this bug happen?

Every time

### Platform

It doesn't depend

### react-query version

4.0.0-beta.23

### TypeScript version

4.7.3

artem-malko:
> hmmm, staleTime for example is definitely applied, because if you click the toggle button, you can see that the query shows up as fresh in the devtools, not stale.

Yes, it works in case of my own query usage. But, if I click to the "fetch newsitem with id ..." button, and after that show the NewsItem component — staleTime won't be 99999, it will be 0, cause of default options from fetcher.

> I think this is a bug, but not sure what problems we would introduce if we also updated cache level options in that effect 🤔

Yeah, I have same thoughts) But, I described a real case: fetcher uses default options, query uses its own options. It is not expected, that data is refetched, when component is mounted, case in its query there are options like:

```ts
{
  cacheTime: 99999,
  staleTime: 99999,
}
```
These numbers are just for an example)

I think, that the latest execution (fetcher or useQuery) should update query options in a queryCache. And it's ok for another story, when useQuery is used before fetcher. What do you think?

artem-malko:
[...]

PR #4749, opened 2023-01-03
fix: set options for a query on setOptions in a queryObserver

artem-malko:
This PR attempts to close https://github.com/TanStack/query/issues/3706

### Opened questions

* How to write a test for my fix? It looks quite complex, so I need an e2e test?

@TkDodo fyi, cause you have the context of the issue)


TkDodo:
> How to write a test for my fix? It looks quite complex, so I need an e2e test?

All our tests are "e2e" tests because they mount a query on a page and then have buttons to interact with them. I think you could just write what your sandbox has and assert accordingly.

However, what bothers me is that it seems like we have a test that tests exactly the opposite of what we're trying to achieve, and that's obviously failing now:

https://github.com/TanStack/query/blob/eeec5f77bc9a703ffb6a6d283dcedada34aa3c75/packages/react-query/src/__tests__/useQuery.test.tsx#L2379-L2401

TkDodo:
fyi, the test was introduced here:
- https://github.com/TanStack/query/pull/942

artem-malko:
Hm, I'm quite confused) 

First of all, that tests checks only queryFunc, not options. So, as I understand, we can save queryFunc between executions, but all other options can be updated?

TkDodo:
> First of all, that tests checks only queryFunc, not options. So, as I understand, we can save queryFunc between executions, but all other options can be updated?

I think the test only checks for queryFn because queryFn is one of the valid options that live on the query itself. It could just as well check any other option.

The PR description says:

> Currently the query configuration is dependent on the order in which hooks are rendered and this could result in unexpected behaviour. This MR will make sure the query configuration is only changed on execution. Related to https://github.com/TanStack/query/discussions/932

artem-malko:
I'm still confused a little( Looks like current behaviour (which is a bug for me) is a correct from react-query side =(

artem-malko:
I'll close this request. Looks like it is quite outdated and it is not so useful for the community) I'll watch for that issue by using v5 and I'll be back with a feedback)

PR #4759, opened 2023-01-05
docs(guides refetchOnWindowFocus): more clear instructions for the use

thienna:
I read the current documentation that there is no mention of the default value of refetchOnWindowFocus, and the current way of writing it makes me re-verify that only refetch when the data is stale, it took me 15 minutes to read the code to verify how it works, so I want to update the docs for more clear.

### DIFF
--- a/docs/react/guides/window-focus-refetching.md
+++ b/docs/react/guides/window-focus-refetching.md
@@ -3,7 +3,7 @@ id: window-focus-refetching
 title: Window Focus Refetching
 ---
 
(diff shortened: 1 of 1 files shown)


TkDodo:
It is covered in the [important defaults page](https://tanstack.com/query/v4/docs/react/guides/important-defaults):

> Stale queries are refetched automatically in the background when:
> - New instances of the query mount
> - The window is refocused
> - The network is reconnected
> - The query is optionally configured with a refetch interval

I don't mind merging this addition, but it's important to understand the general concepts. As long as data is considered "fresh" (NOT stale), you will always only read data from the cache and none of the events will trigger a background refetch.

PR #4768, opened 2023-01-05
feat(svelte-query): Svelte Query Adapter for TanStack Query

lachlancollins:
## Overview
This PR is a continuation of #4344. It adds an official Svelte adapter which is based on the brilliant work of @drejohnson and @amen-souissi.

I'll be actively editing this description as progress is made. I'd also greatly appreciate any and all contributions over at https://github.com/lachlancollins/query.

## API

Based on the discussion at the old PR, the functions which export a store now have the `create` prefix to be less React hooks-like. Additionally, the `<QueryClientProvider>` and `<Hydrate>` components have been added to be more consistent with the other adapters.

- `createQuery`
- `createQueries`
- `createInfiniteQuery`
- `createMutation`
- `useQueryClient`
- `useIsFetching`
- `useIsMutating`
- `useHydrate`
- `<QueryClientProvider>`
- `<Hydrate>`

## Package

`svelte-package` outputs all required files to a (customisable) `/build` folder, which contains everything that needs to be published to npm. I'm fairly sure this differs from how the other packages are published, so let me know how I can ensure this integrates with the CI scripts properly.

## To-do

If you would like to help with one of these tasks, please comment here or create an issue on the fork :)

- [x] Docs overview
- [x] Build using `svelte-package` (allows using .svelte files)
- [x] Implement QueryClientProvider component
- [x] Implement Hydrate component
- [x] Rename store functions with create prefix
- [x] Update examples to SvelteKit 1.0
- [x] Add playground and star-wars examples
- [x] Use vitest instead of jest (setup easier for vite environment)
- [x] Add createQuery and createMutation tests
- [x] Integrate into publish scripts


lachlancollins:
Hi @TkDodo, I'm making some progress getting Rollup to output raw Svelte files (for QueryClientProvider.svelte), but I think there might need to be a change to the `buildConfigs` function. I compared then copied parts of the config from [tanstack/table](https://github.com/TanStack/table/blob/main/rollup.config.ts) which uses an `external` field (for cjs, esm, mjs) in addition to the `globals` field (for umd). This stops Rollup from trying to include `node_modules` in the build output. I don't know the impact that this would have on the other libraries, but I can try to copy the logic from tanstack/table and test it out. Would you be happy if I opened a separate PR to implement this?

My alternative to this is considering using `svelte-package`. This seems to be by far the easiest way to build Svelte libraries, however it would need to be able to run separately to the rollup build. If this isn't feasible, I'll keep chipping away at the Rollup config, or abandon adding QueryClientProvider.svelte.

TkDodo:
We're also bundling the eslint-plugin without rollup so if using svelte-package is the easiest for you let's do that. You can append it to the build script

lachlancollins:
Hi @TkDodo, I think this is ready to be reviewed!

I've rewritten some of the larger React examples (`playground` and `star-wars`) - however, I'm not well-versed in testing Svelte components. Right now I have basic tests for `createQuery` and `createMutation`, and can confirm all the examples work as expected.

I personally feel this is enough since the majority of the code is just types and wrappers around `query-core` (which is obviously well-tested). However, I can totally understand if you'd prefer to be safe - is this sufficient for now, or would you like more tests? If so, is anyone following this PR able to offer some help?

TkDodo:
@lachlancollins are you on discord? if so, please join the TanStack server and reach out to me. Thank you.

https://discord.com/invite/WrRKjPJ

docs/svelte/overview.md
```diff
@@ -1,8 +1,69 @@
 ---
 id: overview
-title: Svelte Query (Coming Soon)
+title: Svelte Query
```
TkDodo, line 3: @DamianOsipiuk should this `ref` the react docs and replace some parts like the vue docs do?

https://github.com/TanStack/query/blob/fa04a1d26b2900163ffb2e2ae238735fed4dddaa/docs/vue/overview.md#L1-L6

examples/svelte/hydration/src/routes/+page.ts
```diff
@@ -0,0 +1,11 @@
+import { dehydrate, QueryClient } from '@tanstack/svelte-query'
+import { getPosts, limit } from '$lib/data.js'
+
+/** @type {import('./$types').PageLoad} */
+export async function load() {
+  const queryClient = new QueryClient()
+  await queryClient.prefetchQuery(['posts', limit], () => getPosts(limit))
```
TkDodo, line 7: it would be good if all examples were to use the object syntax already, like:

```
queryClient.prefetchQuery({
  queryKey: ['posts', limit],
  queryFn: () => getPosts(limit)
})
```

this will be the only valid syntax in v5, and we've done this for all docs and examples of the other frameworks already.

examples/svelte/star-wars/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "devDependencies": {
+    "@sveltejs/adapter-auto": "^1.0.0",
+    "@sveltejs/kit": "^1.0.0",
+    "autoprefixer": "^10.4.13",
+    "postcss": "^8.4.20",
+    "svelte": "^3.54.0",
+    "svelte-check": "^2.9.2",
+    "tailwindcss": "^3.2.4",
+    "tslib": "^2.4.1",
+    "typescript": "^4.9.3",
```
TkDodo, line 24: please make all examples use, strictly:

```
"typescript": "4.7.4",
```

package.json
```diff
@@ -6,13 +6,14 @@
...
     "test": "pnpm run test:ci",
-    "test:ci": "pnpm run test:format && pnpm run test:eslint && pnpm run test:jest --collectCoverage false && pnpm run typecheck",
+    "test:ci": "pnpm run test:format && pnpm run test:eslint && pnpm run test:jest --collectCoverage false && pnpm run test:vitest && pnpm run typecheck",
     "test:react:17": "pnpm --filter \"./packages/react-*\" --parallel --no-bail run test:jest --collectCoverage false",
     "test:eslint": "pnpm --filter \"./packages/**\" --parallel --no-bail run test:eslint",
     "test:format": "pnpm run prettier --check",
     "test:jest": "pnpm --filter \"./packages/**\" --parallel --no-bail run test:jest",
     "test:size": "pnpm run build && bundlewatch",
-    "build": "rollup --config rollup.config.js && pnpm run typecheck && pnpm pnpm --filter \"eslint-plugin-query\" run build && pnpm run build:copyTypes",
+    "test:vitest": "pnpm --filter \"@tanstack/svelte-query\" run test:vitest",
```
TkDodo, line 15: I think it would be better to have one test task that every package should implement, otherwise, they won't run in parallel. Naming the task `test:jest` was probably a bad idea, so how about we rename `test:jest` to `test:lib` and then have every package specify that task and implement it with either `jest` or `vitest` ?

package.json
```diff
@@ -37,6 +38,7 @@
     "@testing-library/react": "^13.0.0",
     "@testing-library/react-17": "npm:@testing-library/react@12.1.4",
     "@testing-library/react-hooks": "^7.0.2",
+    "@testing-library/svelte": "^3.2.2",
```
TkDodo, line 41: this should probably, if possible, just be a dependency of the svelte-query package

packages/svelte-query/src/app.html
```diff
@@ -0,0 +1,12 @@
...
+  <head>
+    <meta charset="utf-8" />
+    <link rel="shortcut icon" type="image/svg+xml" href="%sveltekit.assets%/emblem-light.svg" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    %sveltekit.head%
+  </head>
+  <body>
+    <div>%sveltekit.body%</div>
+  </body>
+</html>
```
TkDodo, line 12: is this file needed? where is it used?

packages/svelte-query/src/routes/+page.svelte
```diff
@@ -0,0 +1,4 @@
+<h1>Svelte Query</h1>
+<p>This routes foulder is unused</p> 
+<p>Documentation should be written in docs/svelte</p>
+<p>Examples should be added to examples/svelte</p>
```
TkDodo, line 4: can we remove this if it is unused?

not really sure why we have a `lib` directory under `src`, as well as `routes`. Is this some sort of bootstrapping? Is it strictly needed?

if possible, I'd prefer to have the same structure as the other lib packages

packages/svelte-query/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": "./.svelte-kit/tsconfig.json",
```
TkDodo, line 2: I't not sure this path exists? Shouldn't this extend our base config? E.g. like the react adapter:

https://github.com/TanStack/query/blob/f61b908014bf936ebd6f51758c746cd07eedf460/packages/react-query/tsconfig.json#L1-L13

note that we use `tsc -b` at the top level to create the `.d.ts` files for everything, so I think the configs need to match

TkDodo (commented):
please have a look at my comments

TkDodo:
@ardeora I've seen the typing errors on solid-query before:

```
Error: packages/solid-query/src/createQueries.ts(157,33): error TS2698: Spread types may only be created from object types.
Error: packages/solid-query/src/createQueries.ts(157,63): error TS2339: Property 'queryKey' does not exist on type 'ArrType<readonly [...T extends [] ? [] : T extends [infer Head] ? [GetOptions<Head>] : T extends [infer Head, ...infer Tail] ? [...Tail] extends [] ? [] : [...Tail] extends [...] ? [...] : [......] extends [...] ? [......] extends [] ? [] : [......] extends [...] ? [...] : [......] extends [...] ? [......] extends [...'.
```

is this because of maybe a different TypeScript version being introduced? I think we should only use `4.7.4` everywhere. Would that fix it?

package.json
lachlancollins: Yep that makes sense - for now I've just renamed it to "test:jest" to avoid changing the other packages as little as possible in this PR

packages/svelte-query/src/app.html
lachlancollins: See below comment

packages/svelte-query/src/routes/+page.svelte
lachlancollins: Turns out I could delete this without consequences! The `svelte-package` command runs within a `@sveltejs/kit` project, so it means there are some structural patterns it needs to conform to.

I _can_ set the package to build `./src` instead of `.src/lib`, but that also builds and exports the `__tests__` folder.

packages/svelte-query/tsconfig.json
lachlancollins: This is another sveltekit specific thing, the `./.svelte-kit` folder is generated by vite when the project runs or builds. Also, `svelte-package` doesn't need tsc to generate types - everything is handled by the single `build` script.

package.json
TkDodo: perfect, I can do the rename separately

TkDodo:
> setting everything to exact 4.7.4 actually adds a vue-query error

I think @DamianOsipiuk has fixed this on the v5 branch - maybe we can backport the fix to main?

TkDodo:
it could be that:

```
"typescript": "^4.7.4",
```

doesn't really give you max 4.7.4. Can we try the strict one without the `^`:

```
"typescript": "4.7.4",
```

lachlancollins:
> it could be that:
> 
> ```
> "typescript": "^4.7.4",
> ```
> 
> doesn't really give you max 4.7.4. Can we try the strict one without the `^`:
> 
> ```
> "typescript": "4.7.4",
> ```

@tkdoko I fixed it - at some point the pnpm-lock updated every package, so I reset it to upstream and re-installed the new packages.

TkDodo:
we're getting there. The preview sandbox shows the following error:

```
node_modules/esbuild/lib/main.js:1336:21: ERROR: [plugin: vite:dep-scan] Failed to resolve entry for package "@tanstack/svelte-query". The package may have incorrect main/module/exports specified in its package.json.
node_modules/esbuild/lib/main.js:1336:21: ERROR: [plugin: vite:dep-scan] Failed to resolve entry for package "@tanstack/svelte-query". The package may have incorrect main/module/exports specified in its package.json.
```

I think the `package.json` is not correct, have a look at how other packages have entry points specified:

https://github.com/TanStack/query/blob/d7d98a9789426a8e1b964166ea9cba9971da83b3/packages/vue-query/package.json#L17-L28

I'm not sure if svelte-kit produces the output in the same branches, but it would be good if we could customize it like that.

for example, for our `size` testing, we check all file matching `packages/*/build/umd/*.production.js`

TkDodo:
and prettier needs to run on some files :)

lachlancollins:
> we're getting there. The preview sandbox shows the following error:
> 
> ```
> node_modules/esbuild/lib/main.js:1336:21: ERROR: [plugin: vite:dep-scan] Failed to resolve entry for package "@tanstack/svelte-query". The package may have incorrect main/module/exports specified in its package.json.
> node_modules/esbuild/lib/main.js:1336:21: ERROR: [plugin: vite:dep-scan] Failed to resolve entry for package "@tanstack/svelte-query". The package may have incorrect main/module/exports specified in its package.json.
> ```
> 
> I think the `package.json` is not correct, have a look at how other packages have entry points specified:
> 
> ...
> 
> I'm not sure if svelte-kit produces the output in the same branches, but it would be good if we could customize it like that.
> 
> for example, for our `size` testing, we check all file matching `packages/*/build/umd/*.production.js`

Hi @TkDodo, I've made a bit of progress here.

By default, `svelte-package` generates its own `package.json` inside the build dir which autogenerates all the main/module/exports fields. However, it seems like this wouldn't integrate too easily into the CI. So, I've copied the important fields from the build `package.json` into the main `package.json`, and set up rimraf to delete the build `package.json`.

The next "issue" is that `svelte-package` only builds ESM. From what I can tell, there isn't really any need for CJS or UMD in the Svelte ecosystem (anymore?) - at least nobody has been complaining about the output of `svelte-package`. There are definitely some Svelte packages I've used in the past which refused to work in SSR that were built through Rollup.

TkDodo:
okay, if ESM is all we need, then so be it 😅

packages/svelte-query/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "author": "Dre Johnson",
+  "license": "MIT",
+  "repository": "tanstack/query",
+  "homepage": "https://tanstack.com/query",
+  "funding": {
+    "type": "github",
+    "url": "https://github.com/sponsors/tannerlinsley"
+  },
+  "type": "module",
+  "module": "build/lib/index.js",
```
TkDodo, line 14: do we not need a separate entry for the `types` that point to the `.d.ts` files?

lachlancollins: Added it now!

JosephPallipadan (approved):

lachlancollins:
@TkDodo I think this should pass CI now

docs/svelte/overview.md
DamianOsipiuk: Ideally it should if content is appropriate.

TkDodo: Can we do this in a follow up @lachlancollins ?

PR #4769, opened 2023-01-05
Feature/size improvements

### LINKED ISSUE #4679: size improvements

TkDodo:
fixes #4679 

draft PR because tests are very likely failing


TkDodo:
@DamianOsipiuk could you have a look at the failing vue-query test? I don't really understand what's happening there 🤔

Mamoanwar97:
> @DamianOsipiuk could you have a look at the failing `vue-query` test? I don't understand what's happening there 🤔

Before using `ECMAScript Private class feature` here:

before

the tests that are failing in `vue` right now, are trying to access `defaultOptions` and making sure that the `queryClient` is called with the correct config

vue-test

I tried to fix these tests but unfortunately, I failed 😬

DamianOsipiuk:
@TkDodo Fixed Vue tests

PR #4779, opened 2023-01-08
chore: Restructure package scripts

lachlancollins:
## Purpose

When I added the svelte-query adapter, there many things about Svelte that didn't integrate nicely into the existing tools, including as rollup (svelte-query uses svelte-package), jest (svelte-query uses vitest) and tsc (svelte-query uses svelte-check). Because I wanted to avoid touching other packages as much as possible, I adapted it as best as I could for the time, but planned to come back and restructure some of the package scripts.

## Changes

- Most scripts can now run on windows in addition to linux
- Removed pnpm `--no-bail` flag as this is true by default
- `test:jest` has been renamed to `test:lib`, with `test:lib:dev` for watch mode and `test:lib:publish` for the CI step without coverage reports
- `typecheck` has been renamed to `build:types`, and a separate `test:types` has been added (which runs `tsc` for most packages, and `svelte-check` for svelte-query)
- `rimraf` added to replace `rm -rf`
- `../../node_modules/.bin/` were removed as these aren't required (maybe this was from pre-pnpm?)
- `prettier-plugin-svelte` has been added to parse svelte files (I ran it when writing my PR, but didn't add it to the project)
- Unused jest config removed from `query-broadcast-client-experimental` (jest throws an error when no tests are present)
- `build:copyTypes` removed from root and moved into `react-query-devtools`
- `react-query` codemod tests merged with the main lib tests (adding the --collectCoverage flag was causing an error when pnpm tried to run on `pnpm run test:codemod && jest --config ./jest.config.ts`)


lachlancollins:
@TkDodo I think this works even though the tests are occasionally failing on `react-query` - the first commit failed, the second commit passed, and the third one failed again but on a different test

TkDodo (approved):
amazing, I wanted to have this for a long time but couldn't figure out how to make tsc work in a mode where it doesn't also produce .d.ts files ❤️ 

I've tried everything locally and it works fine, it also looks like we don't need to change anything in the publish script.

are we sure that tasks that run with `pnpm --filter` are still being run in parallel ? I can't really tell locally or on CI. But it also might not matter given that the times look identical 🤷 

@DamianOsipiuk are you also fine with merging this?

DamianOsipiuk (approved):
Looks good 🚀

lachlancollins:
> are we sure that tasks that run with `pnpm --filter` are still being run in parallel ? I can't really tell locally or on CI. But it also might not matter given that the times look identical 🤷

@TkDodo yep quite sure, pnpm runs [4 tasks simultaneously by default](https://pnpm.io/cli/recursive#--workspace-concurrency). You can see below how it starts the solid-query test, reports progress for some other packages, then reports solid-query as done later.

```
packages/solid-query test:types$ tsc
.../query-broadcast-client-experimental test:types: Done
packages/svelte-query test:types$ svelte-kit sync && svelte-check --tsconfig ./tsconfig.json
packages/query-persist-client-core test:types: Done
packages/vue-query test:types$ tsc
packages/svelte-query test:types: ====================================
packages/svelte-query test:types: Loading svelte-check in workspace: /home/runner/work/query/query/packages/svelte-query
packages/svelte-query test:types: Getting Svelte diagnostics...
packages/solid-query test:types: Done
```

PR #4788, opened 2023-01-10
fix(svelte-query): Remove dependency on SvelteKit

lachlancollins:
I've realised that the only part of sveltekit needed to build the svelte-query package are just re-exported from vite-plugin-svelte. This PR removes sveltekit as a dependency, and the build output remains entirely the same.

I've also moved the code from src/lib to src for consistency with the other packages in the repo, and updated vite-plugin-svelte in the examples.


packages/svelte-query/tsconfig.json
```diff
@@ -1,14 +1,28 @@
 {
-  "extends": "./.svelte-kit/tsconfig.json",
```
TkDodo, line 2: should we now extend our `base.tsconfig.json` file? What I'd like to have is consistent typescript settings enabled that are not enabled by default, like:

```
"noUncheckedIndexedAccess": true,
```

our base config ensures that.

TkDodo (commented):
> No release should be needed, however I think the publish script will think it is needed because the files in src have moved.

we can merge the PR without a fix / feat commit message then it will not trigger a release :)

packages/svelte-query/tsconfig.json
lachlancollins: I've tried this, but it seems to mess with `svelte-package`'s logic which determines where the `.d.ts` files get built to, and leads to the structure below. Is it okay if for now I copy the rules that don't break it from tsconfig.base.json?

image

TkDodo: yep 👍

lachlancollins: I got a bit closer to making this work with `tsconfig.base.json` - `tsc` reports no errors (but doesn't pass svelte files), while `svelte-check` gives the error below. The inconsistency between the two means `svelte-check` is probably not reading the config identically to `tsc` which is causing problems.

```
Error: File 'c:/Users/XYZ/Documents/GitHubProjects/tanstack-query/packages/query-core/src/index.ts' is not under 'rootDir' 'c:/Users/XYZ/Documents/GitHubProjects/tanstack-query/packages/svelte-query/src'. 'rootDir' is expected to contain all source files.
```

lachlancollins: Okay fairly certain this will not work for now, but the new rules are working :)

lachlancollins:
> we can merge the PR without a fix / feat commit message then it will not trigger a release :)

@TkDodo I'm now thinking it _might_ be best to just do a patch release, since the devDependencies have changed and the src structure has changed (which gets included in the NPM package). But not too fussed either way!

Issue #4510, opened 2022-11-14
SSR + suspense + solid-query does not seem to work


marbemac:
### Describe the bug

SSR + suspense does not seem to work with `solid-query` (or I'm doing something wrong on the implementation). I've included a minimum reproducible repo that compares solid-query vs basic `createResource` (which I looks like solid-query is using in some capacity under the hood, so *maybe* this is fixable!).

### Your minimal, reproducible example

https://codesandbox.io/p/github/marbemac/solid-query-ssr-bug

### Steps to reproduce

Turn off javascript and load up the solid-query route in the attached reproducible example. Under the hood it's just using a simple query, with a suspense wrapper (https://github.com/marbemac/solid-query-ssr-bug/blob/main/client/pages/solid-query.tsx), rendering on the server with `renderToStringAsync`.

### Expected behavior

As a user, I'd like to leverage suspense + solid-query in SSR applications.

### How often does this bug happen?

Every time

### Platform

- macOs

### react-query version

4.15.1

### TypeScript version

4.8.4

### Additional context

- https://github.com/marbemac/solid-query-ssr-bug
- https://codesandbox.io/p/github/marbemac/solid-query-ssr-bug

TkDodo:
I think I remember @ardeora mentioning somewhere that there is no complete story for solid-query in SSR yet and that we wanted to wait until solid-start is stable. But I could be wrong / misremember 😅

OrJDev:
https://twitter.com/aryan__deora/status/1592025888761446400?s=46&t=HoDG03U8-68q6IdSB-sZGQ

marbemac:
Resolved by the variety of improvements Aryan and I made to solid-query over the last year.

PR #4803, opened 2023-01-12
feat(solid-query):  SSR support

ardeora:
This change introduces SSR support for `solid-query` primitives:

- Supports SSR
  - createQuery
  - createInfiniteQuery
  - createMutation

- Currently unsupported
  - createQueries

# How does this work?
SolidJS [resources](https://www.solidjs.com/docs/latest/api#createresource) run isomorphically whether they are on the server or on the client. If they run on the server, their value is serialised and sent down the wire before they run on the client again. 

We leverage this functionality in `createBaseQuery` where we add a new resource called `queryResource`. When a query is resolved on the server we populate the `queryResource` with the dehydrated queryClient. Now when `createQuery` is picked back up again on the client, it first checks to see if the `queryResource` has any dehydrated queryClient. If it does, it loads the query before setting up the queryObservers and then everything just works!

## Example of createQuery working with SSR

FmCN9x7WIAEwINb

## createQueries
Currently createQueries doesn't have SSR/Suspense support. This is something I want to tackle for the v5 release but need some more brain cells than I currently have to figure out 😅 Hopefully I'll have something in the coming weeks.

Please let me know if you have any comments, feedback or concerns and I would be sure to work on them!

Thanks,
Aryan


packages/solid-query/src/createBaseQuery.ts
```diff
@@ -35,46 +42,99 @@ export function createBaseQuery<
...
 
-  const [dataResource, { refetch, mutate }] = createResource<TData | undefined>(
+  const [dataResource, { mutate, refetch }] = createResource<TData | undefined>(
     () => {
       return new Promise((resolve) => {
-        if (!(state.isFetching && state.isLoading)) {
-          if ((unwrap(state.data) as TData | typeof emptyData) === emptyData) {
-            resolve(undefined)
+        if (isServer) {
+          if (!(state.isFetching && state.isLoading)) {
```
TkDodo, line 84: `isFetching && isLoading` could be unified to just check for `isInitialLoading`

TkDodo (commented):
general question: does this have the same feature / limitation as react-query where the query will run on the client again immediately because the hydrated data is considered stale (given the default staleTime of zero)?

if so, it might make sense to clarify this in the docs.

packages/solid-query/src/createBaseQuery.ts
ardeora: Yeah since staleTime is zero when the query is mounted on the client it will do a refetch. You could opt out of it by having a non zero staletime. Or maybe just setting refetchOnMount to false as well. I'll update the docs with this caveat and also change the checks to `isInitialLoading`

TkDodo: non-zero staleTime is the best way imo 👍

ardeora: I'll look more into this! I'm not fully sure how it works but maybe using `isRestoring` for the first render on the client might help? Just spitballing here. I'll test more and report back

TkDodo: `useIsRestoring` is currently only used for hydrating from the persist plugin, not when hydrating via the `Hydration` boundary.

packages/solid-query/src/createBaseQuery.ts
```diff
@@ -35,46 +42,99 @@ export function createBaseQuery<
...
+      }
+      queryResolver = resolve
+    })
+  })
+
+  // If queryResource is defined,
+  // This means that the query was fetched on the server!
+  // We need to hydrate the queryClient with the query from
+  // the server before we set up the observer and its results.
+  if (queryResource()) {
```
marbemac, line 72: Minor, but could use the native `onHydrated` option on `createResource` up on line 59? E.g.

```ts
createResource<DehydratedState | undefined>(
  () => {
      if (!isServer) {
        resolve(undefined)
      }
      queryResolver = resolve
    })
  },
  {
    onHydrated(_k, info) {
      // We need to hydrate the queryClient with the query from
      // the server before we set up the observer and its results.
      hydrate(queryClient(), info.value)
    },
  },
)
```

examples/solid/solid-start/src/routes/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+        </button>
+        <button
+          onClick={() => {
+            setPostId((id) => (id === 100 ? 100 : id + 1))
+          }}
+        >
+          Next Page
+        </button>
+      </div>
+      <Suspense>
```
marbemac, line 51: I'm not sure that this `Suspense` boundary is accomplishing anything? Since the resource is created outside of the boundary, up on line 16, I believe that the suspense boundary that is catching the resource is actually the root one in `root.tsx`.

Can see this more clearly when you adjust `entry-server.tsx` to use `renderStream`, and turn off javascript to see which boundary actually catches the query on the server render:

Screenshot 2023-01-13 at 1 41 41 PM

A better example might be to move the query into a separate posts component, so that one can see the streaming magic and suspense boundary at work, which is SO cool :):

Screenshot 2023-01-13 at 1 44 20 PM

And streaming shows off nicely - look at all that blue:

Screenshot 2023-01-13 at 1 47 16 PM

Example adjustment to this file:

```tsx
import { Title } from 'solid-start'
import { createQuery } from '@tanstack/solid-query'
import { createSignal, For, Suspense } from 'solid-js'
import type { Accessor } from 'solid-js'

interface PostData {
  userId: number
  id: number
  title: string
  body: string
}

export default function Home() {
  const [postId, setPostId] = createSignal(1)

  return (
    <main>
      <Title>Solid Query v5</Title>
      <h1>Solid Query v5</h1>
      <p>
        This is just a dev release for demo. Please dont use it in production
        yet. Also this package will be deprecated as soon as features are live
        on @tanstack/query
      </p>
      <div>
        <button
          onClick={() => {
            setPostId((id) => (id === 1 ? 1 : id - 1))
          }}
        >
          Previous Page
        </button>
        <button
          onClick={() => {
            setPostId((id) => (id === 100 ? 100 : id + 1))
          }}
        >
          Next Page
        </button>
      </div>

      <Suspense
        fallback={<div style={{ 'margin-top': '20px' }}>Loading post...</div>}
      >
        <Posts postId={postId} />
      </Suspense>
    </main>
  )
}

function Posts({ postId }: { postId: Accessor<number> }) {
    queryKey: ['posts', postId()],
    staleTime: Infinity,
    queryFn: async () => {
      const response = await fetch(
        'https://jsonplaceholder.typicode.com/posts/' + postId(),
      ).then((res) => res.json())

      // simulate extra latency to make things like streaming behavior more clear
      await new Promise((r) => setTimeout(r, 1000))

      return [response] as PostData[]
    },
  }))

  return (
    <For each={query.data}>
      {(post) => (
        <div>
          <h3>{post.title}</h3>
          <p>{post.body}</p>
        </div>
      )}
    </For>
  )
}
```

packages/solid-query/src/createBaseQuery.ts
```diff
@@ -35,46 +42,99 @@ export function createBaseQuery<
...
+            if (
+              (unwrap(state.data) as TData | typeof emptyData) === emptyData
+            ) {
+              resolve(undefined)
+            }
+            resolve(unwrap(state.data))
           }
-          resolve(unwrap(state.data))
         }
       })
```
marbemac, line 104: Any way that we could expose the createResource `deferStream` option? Seems to be working very nicely (and could make for a neat extra route in the solid-start example, to show off deferred and non-deferred streaming), but I'm not sure what pattern ya'll are using to expose adapter specific options up on `createQuery`.

marbemac (commented):
Awesome work! <3 this so much. BTW, this would close #4510.

packages/solid-query/src/createBaseQuery.ts
marbemac: Scratch that - putting it here results in the `hydrate` call not happening until after the `observer.subscribe`, which is not what we want.

packages/solid-query/src/createBaseQuery.ts
marbemac: However, I did notice this warning. Does not seem to cause any issues in this simple example, but I suspect it will cause random bugs in a more complex / real world application unless resolved. The same query done with a regular createResource (rather than through solid-query) does not result in the hydration warning.

Screenshot 2023-01-13 at 3 40 51 PM

ardeora: > However, I did notice this warning. Does not seem to cause any issues in this simple example, but I suspect it will cause random bugs in a more complex / real world application unless resolved. The same query done with a regular createResource (rather than through solid-query) does not result in the hydration warning.
> 
> Screenshot 2023-01-13 at 3 40 51 PM

Hello! Thank you so much for taking out time to review the changes. Really appreciate it ❤️  Interesting! I have once seen this issue when I didn't use a signal in the JSX. But I cannot seem to recreate this issue again. I have a Stackblitz example if you would like to fork to recreate the warning? I can then take a look to see why that happens. Maybe just an overly prescriptive warning? Will definitely confirm what's causing it. https://stackblitz.com/edit/github-bo2rcr?file=src%2Froutes%2Findex.tsx

ardeora: > Any way that we could expose the createResource `deferStream` option? Seems to be working very nicely (and could make for a neat extra route in the solid-start example, to show off deferred and non-deferred streaming), but I'm not sure what pattern ya'll are using to expose adapter specific options up on `createQuery`.

Yeah we can definitely expose `deferStream` as an option to `createQuery`/`createinfiniteQuery`. I tried futzing around with streaming SSR but could not figure out how `deferStream` works. Do you have an example how `deferStream` works in resources? That would be immensely helpful

packages/solid-query/src/createBaseQuery.ts
ardeora: Yeah I initially had the hydration on the `onHydrate` callback but seems to run as an effect after flushing the function so had to do this hacky resolver way

examples/solid/solid-start/src/routes/index.tsx
ardeora: Yeah this was from when i was trying to futz around with Streaming SSR but could not figure it out haha. I'll fix the example to follow the above changes you requested!

packages/solid-query/src/createBaseQuery.ts
marbemac: > Do you have an example how deferStream works in resources? That would be immensely helpful

For sure - try adjusting the example to https://github.com/TanStack/query/pull/4803#discussion_r1069858512 + updating `entry-server.tsx` to use `renderStream` rather than `renderAsync`. Then if you turn off javascript in your browser you should see the difference quite clearly when toggling "deferStream" true/false for that `post` query.

> Hello! Thank you so much for taking out time to review the changes. Really appreciate it ❤️ Interesting! I have once seen this issue when I didn't use a signal in the JSX. But I cannot seem to recreate this issue again.

Oh no problem, thanks for getting the ball rolling! This DX is going to be so great, I'm pretty excited about it.

Ah, you'll see the hydration key issue if you adjust to use `renderStream` & set `deferStream` to true (and make sure javascript back on in browser). 

It think it's something in the solid-query implementation, because if one adjusts to just use `createResource` directly to fetch the post, ala the example below, the hydration key warning goes away 🤔.

```ts
function Post({ postId }: { postId: Accessor<number> }) {
  const [postData] = createResource(async () => {
    const response = await fetch(
      'https://jsonplaceholder.typicode.com/posts/' + postId(),
    ).then((res) => res.json())

    // simulate extra latency to make things like streaming behavior more clear
    await new Promise((r) => setTimeout(r, 1000))

    return [response] as PostData[];
  }, { 
    deferStream: true,
  })

  return (
    <div>
      <For each={postData()}>
        {(post) => (
          <div>
            <h3>{post.title}</h3>
            <p>{post.body}</p>
          </div>
        )}
      </For>
    </div>
  )
}
```

ardeora: A little update on this. I think I figured out why we get the `hydration key` warning. It's because (I think) it's trying to find a dom node where `queryResource` is being used in the component tree. But it's only used internally by createQuery!

ardeora: Oooh! I think I fixed it 🥳 🥳 🥳

marbemac: Unfortunately, I'm still seeing the hydration error. BUT, I think I know why now - mind if I open a separate PR with an adjusted approach and explanation (primarily, consolidates to a single `createResource` call)? Can then compare/contrast/experiment with it, and either incorporate the changes here, roll w the other PR, or find another approach if it's not valid - I don't mind either way.

ardeora: Which version of @tanstack/query are you using? Is it the one with the Code Sandbox link above?

ardeora: Also, yes! PRs are always welcome 😄

ardeora: I think I know now. Try removing the fallback on your suspense boundaries. You wont see the hydration warning anymore 😄 That's weird, it's expecting the first pass render to have the data resource be undefined on the client when hydrating!

packages/solid-query/src/createBaseQuery.ts
ardeora: Okay I think I got it working so that a query is not refetched when a query is mounted from the server. This change [ad45d25](https://github.com/TanStack/query/pull/4803/commits/ad45d259812fd98baed974612206df8e7703bb3d)
makes solid-query work exactly the same between CSR and SSR'd apps!

packages/solid-query/src/createBaseQuery.ts
ardeora: I have a discussion on the solid-start discussion channel. If you take a look this hydration key warning is being experienced by a lot of folks (for people not using tanstack query). It doesn't affect anything. This maybe something solid-start needs to fix

packages/solid-query/src/createBaseQuery.ts
```diff
@@ -35,46 +42,107 @@ export function createBaseQuery<
...
 
-  batch(() => {
-    mutate(() => unwrap(state.data))
-    refetch()
-  })
+  if (!isServer) {
+    batch(() => {
+      mutate(() => unwrap(state.data))
+      refetch()
+    })
```
marbemac, line 120: After some experimenting, I believe this is at least one of the causes of the hydration key error - attempting to avoid this in #4840, but atm my alternative approach is causing most of the `expect(states.length).toBe(3)` assertions in the tests to be one larger than before (e.g. `expect(states.length).toBe(4)`). Haven't had time to dig into the tests quite yet.

packages/solid-query/src/createBaseQuery.ts
marbemac: I'll join it :).

Popped the WIP up over here if you're interested -> #4840.

PR #4805, opened 2023-01-12
fix(focusManager): stop listening for `focus` events

ilyasmez:
### Why?

As discussed in #4797, `focus` event has many caveats, as it's triggered in the following scenarios:
- When interacting with an iframe.
- When opening/closing a browser dialog (`alert`, `confirm`).
- When moving between the page and DevTools.
- When opening/closing a file picker.

I think the original idea behind `refetchOnWindowFocus` was to refetch the queries when the user "comes back" to the page, and that's achievable using `visibilitychange`.

### Before
You can see that the query is being refetched in many unexpected scenarios, while the user didn't leave the page.
<video src="https://user-images.githubusercontent.com/30146019/212169502-a4078286-eab6-4d1e-8914-4a24f968aab1.mp4" />

### After
The query is refetched only when the user stop using the app, and comes back.
<video src="https://user-images.githubusercontent.com/30146019/212169521-78ba9e63-c320-4b79-aa87-bd8caf2ef4f7.mp4" />

### What does this PR do
- Remove the `focus` event listener in `focusManager`.
- Update docs (also removed the old pitfalls and caveats, as they will be fixed with this change).
- Update tests (I went with consistency instead of trying to rewrite some of them).

### DIFF
--- a/docs/react/guides/important-defaults.md
+++ b/docs/react/guides/important-defaults.md
@@ -15,8 +15,6 @@ Out of the box, TanStack Query is configured with **aggressive but sane** defaul
   - The network is reconnected
   - The query is optionally configured with a refetch interval
 
-If you see a refetch that you are not expecting, it is likely because you just focused the window and TanStack Query is doing a [`refetchOnWindowFocus`](../guides/window-focus-refetching). During development, this will probably be triggered more frequently, especially because focusing between the Browser DevTools and your app will also cause a fetch, so be aware of that.
-
 > To change this functionality, you can use options like `refetchOnMount`, `refetchOnWindowFocus`, `refetchOnReconnect` and `refetchInterval`.
 
 - Query results that have no more active instances of `useQuery`, `useInfiniteQuery` or query observers are labeled as "inactive" and remain in the cache in case they are used again at a later time.
--- a/docs/react/guides/window-focus-refetching.md
+++ b/docs/react/guides/window-focus-refetching.md
@@ -48,36 +48,19 @@ In rare circumstances, you may want to manage your own window focus events that 
 
 ```tsx
 focusManager.setEventListener((handleFocus) => {
-  // Listen to visibilitychange and focus
+  // Listen to visibilitychange
   if (typeof window !== 'undefined' && window.addEventListener) {
     window.addEventListener('visibilitychange', handleFocus, false)
-    window.addEventListener('focus', handleFocus, false)
   }
 
   return () => {
     // Be sure to unsubscribe if a new handler is set
     window.removeEventListener('visibilitychange', handleFocus)
-    window.removeEventListener('focus', handleFocus)
   }
 })
 ```
 
 [//]: # 'Example3'
-
-## Ignoring Iframe Focus Events
-
-A great use-case for replacing the focus handler is that of iframe events. Iframes present problems with detecting window focus by both double-firing events and also firing false-positive events when focusing or using iframes within your app. If you experience this, you should use an event handler that ignores these events as much as possible. I recommend [this one](https://gist.github.com/tannerlinsley/1d3a2122332107fcd8c9cc379be10d88)! It can be set up in the following way:
-
-[//]: # 'Example4'
-
-```tsx
-import { focusManager } from '@tanstack/react-query'
-import onWindowFocus from './onWindowFocus' // The gist above
-
-focusManager.setEventListener(onWindowFocus) // Boom!
-```
-
-[//]: # 'Example4'
 [//]: # 'ReactNative'
 
 ## Managing Focus in React Native
@@ -105,7 +88,7 @@ useEffect(() => {
 
 ## Managing focus state
 
-[//]: # 'Example5'
+[//]: # 'Example4'
 
 ```tsx
 import { focusManager } from '@tanstack/react-query'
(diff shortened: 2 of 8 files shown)


docs/react/guides/important-defaults.md
```diff
@@ -15,8 +15,6 @@ Out of the box, TanStack Query is configured with **aggressive but sane** defaul
   - The network is reconnected
   - The query is optionally configured with a refetch interval
 
-If you see a refetch that you are not expecting, it is likely because you just focused the window and TanStack Query is doing a [`refetchOnWindowFocus`](../guides/window-focus-refetching). During development, this will probably be triggered more frequently, especially because focusing between the Browser DevTools and your app will also cause a fetch, so be aware of that.
-
```
TkDodo, line 19: wow this is great! I didn't know that it would also fix this issue ❤️

docs/react/guides/window-focus-refetching.md
```diff
@@ -105,7 +88,7 @@ useEffect(() => {
 
 ## Managing focus state
 
-[//]: # 'Example5'
+[//]: # 'Example4'
```
TkDodo, line 91: maybe we should switch to named examples @DamianOsipiuk 😅. I see the vue-query docs only overwrite/use the `'Example'` slot, but this could become tedious with numbers

TkDodo (changes requested):

TkDodo (approved):
accidentally pressed request changes - i wanted to approve 😅

docs/react/guides/window-focus-refetching.md
DamianOsipiuk: As you know naming things is hard, so i just went with numbers 🙈 
But yeah, probably we should name them properly.

DamianOsipiuk:
@TkDodo I believe we should add a note about this to the migration guide.

TkDodo:
aah yes, forgot about that 👍 .
I'll do it on the v5 branch directly

lbhbrave:
Hello guys. We have also encountered ux experience issues caused by focus recently and would like to know if this pull request can be merged into version V3. The risk is too high for us to perform a major version upgrade

TkDodo:
> if this pull request can be merged into version V3

it's a breaking change so no. also, it was merged to v5. seems like you haven't upgraded in over a year ...

lbhbrave:
Yes. we are keeping using v3. And we fix this by git-patch. Maybe it's time
to upgrade. Thx~

Dominik Dorfmeister ***@***.***> 于2023年7月15日周六 16:19写道：

gabrielbrattgard:
@TkDodo @ilyasmez  I might be wrong here but this only seems to work when switching between tabs and NOT between browser windows or returning from another desktop application. Is that intentional?

TkDodo:
@gabrielbrattgard it does whatever the browser [visibilitychange](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event) does so yes, that's an purpose. You can totally opt back into the `focus` event though with `focusManager.setEventListener`

mrlubos:
This is great @ilyasmez!

RiccardoRomagnoli:
@ilyasmez @TkDodo There is a case where without focus we can not archive the intended outcome.

With this change, now if you have two browser windows (one of them is our app), not maximised and you switch between one and the other (by click, by tab, eetc) the query will now not refetch anymore.

ilyasmez:
> There is a case where without focus we can not archive the intended outcome.
> 
> With this change, now if you have two browser windows (one of them is our app), not maximised and you switch between one and the other (by click, by tab, eetc) the query will now not refetch anymore.

@RiccardoRomagnoli, do you really want the page to refetch when the user _interacts_ with it? This might result in content changing under their cursors (e.g. the item they wanted to click is not there anymore, or the position of the item changes). I believe that interacting with the page is already too late.
What you need is probably the [broadcastQueryClient](https://tanstack.com/query/v3/docs/framework/react/plugins/broadcastQueryClient), it will update all the open windows immediately as they happen, which is a much superior experience than waiting for the user to interact with the page.

HemalR:
> > There is a case where without focus we can not archive the intended outcome.
> > With this change, now if you have two browser windows (one of them is our app), not maximised and you switch between one and the other (by click, by tab, eetc) the query will now not refetch anymore.
> 
> @RiccardoRomagnoli, do you really want the page to refetch when the user _interacts_ with it? This might result in content changing under their cursors (e.g. the item they wanted to click is not there anymore, or the position of the item changes). I believe that interacting with the page is already too late. What you need is probably the [broadcastQueryClient](https://tanstack.com/query/v3/docs/framework/react/plugins/broadcastQueryClient), it will update all the open windows immediately as mutations happen, which is a much superior user experience than waiting for the user to interact with the page, imo.

This behaviour can be desirable in certain scenarios. E.g. A notification badge counter that can update when a user refocuses on to the browser window.

My workaround fix from the comments on this thread plus experimenting with other ways:

```
<script lang="ts">
	import { browser } from '$app/environment';
	import { QueryClient, QueryClientProvider, focusManager } from '@tanstack/svelte-query';
	import { onMount } from 'svelte';
	import '../app.css';

	let { children } = $props();

		defaultOptions: {
			queries: {
				enabled: browser,
				refetchOnWindowFocus: true,
				staleTime: 0 // Set this to 0 to always consider data stale
			}
		}
	});

	onMount(() => {
		focusManager.setEventListener((handleFocus) => {
			const focus = () => {
				console.log('Focus from parent');
				handleFocus(true);
				queryClient.refetchQueries(); // Without this, the above 'focus from parent' will be logged but queries will only refetch the first time the window is refocused from a different program. After that, data will not be refetched.
			};
			window.addEventListener('visibilitychange', focus);
			window.addEventListener('focus', focus);
			return () => {
				window.removeEventListener('visibilitychange', focus);
				window.removeEventListener('focus', focus);
			};
		});
	});
</script>

	{@render children()}
</QueryClientProvider>
```

If anybody can explain the weird reason why the queries only refetch if manually triggered I'd love to know and improve what I'm doing here.

PR #4809, opened 2023-01-13
docs(svelte-query): Expand SSR docs

### LINKED ISSUE #4798: Svelte executes queryFn on server and ignores result
### Describe the bug

When using svelte-query with sveltekit, `createQuery` executes on the server. This causes any client-specific APIs, such as a relative fetch or `window` access, to throw an error. Furthermore, since async SSR must be done in the `load` function in sveltekit, the result of `queryFn` is dropped.

https://codesandbox.io/p/sandbox/lucid-architecture-sqwk6e?file=%2Fsrc%2Froutes%2F%2Bpage.svelte

### Steps to reproduce

1. Open the dev logs
2. Reload the page
3. Observe the error being thrown on the server
4. Observe that even though the error is caught, and the string `'error'` is returned, the client renders `undefined` until it finishes fetching

### Expected behavior

I expect `queryFn` only to be called on the client but it is called on the server.

Every time

### Platform

- OS: Linux
- Browser: Firefox

### TanStack Query version

4.22.1

### Additional context

Workaround is to add
```ts
import { browser } from '$app/environment';
// ...
queryFn: () => {
    if (!browser) return null;
    // ...
}
```

lachlancollins:
Fixes #4798. Expanded the SSR docs to explain what to do when you aren't explicitly fetching a query on the server. I've added this fix to the svelte-query examples.


TkDodo:
prettier formatting is complaining :)

lachlancollins:
> prettier formatting is complaining :)

Thanks, fixed now :)

PR #4819, opened 2023-01-14
feat: replace custom context prop with custom queryClient instance

DamianOsipiuk:
BREAKING CHANGE: custom context has been removed from hooks in favor of passing custom queryClient instance

Closes: #4682

TODO:
- [x] Migration guide entry
- [x] Add replaceable content section in migration guide for framework specific section

### DIFF
--- a/packages/solid-query/src/createMutation.ts
+++ b/packages/solid-query/src/createMutation.ts
@@ -1,3 +1,4 @@
+import type { QueryClient } from '@tanstack/query-core'
 import { MutationObserver } from '@tanstack/query-core'
 import { useQueryClient } from './QueryClientProvider'
 import type {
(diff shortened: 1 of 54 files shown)


TkDodo (commented):
thank you, this looks good.
we'd need an entry in the migration guide please, and maybe it makes sense to keep one test somewhere that tests with a custom queryClient being passed around?

packages/solid-query/src/createMutation.ts
```diff
@@ -17,11 +18,12 @@ export function createMutation<
   TContext = unknown,
 >(
   options: CreateMutationOptions<TData, TError, TVariables, TContext>,
+  queryClient?: () => QueryClient,
 ): CreateMutationResult<TData, TError, TVariables, TContext> {
-  const queryClient = useQueryClient({ context: options().context })
+  const client = useQueryClient(queryClient?.())
```
DamianOsipiuk, line 23: @ardeora In createBaseQuery we are wrapping useQueryClient with createMemo.
However in createMutation and createQueries we do not do this.
Should we add createMemo to those hooks as well?

DamianOsipiuk:
@TkDodo PR should be ready now

TkDodo (approved):

packages/solid-query/src/createMutation.ts
ardeora: This is fine for now 😄 The memoed queryClient isn't really fully reactive yet.  I'll create a separate PR to wire in the reactivity for the other queryClients too

ardeora:
This looks awesome! I love this change!

ardeora (approved):

PR #4825, opened 2023-01-15
test: stabilize various tests

zorzysty:
Bumping node and vitest versions fixes  `FATAL ERROR: v8::FromJust Maybe value is Nothing.`, like the one here: https://github.com/TanStack/query/actions/runs/3923037978/jobs/6706379076


packages/svelte-query/package.json
```diff
@@ -34,7 +34,7 @@
     "tslib": "^2.4.1",
     "typescript": "^4.7.4",
     "vite": "^4.0.0",
-    "vitest": "^0.26.3"
+    "vitest": "^0.27.1"
```
TkDodo, line 37: the `"@vitest/coverage-istanbul"` dependency matched the version of `vitest`. not sure if that's required but maybe we should update it as well?

zorzysty: good catch
[7d3ad6a](https://github.com/TanStack/query/pull/4825/commits/7d3ad6a5b1ed99cdc226a763a180ba30074f07cd)

PR #4829, opened 2023-01-16
Rename cacheTime to gcTime

### LINKED ISSUE #4678: rename cacheTime
### context

Almost everyone gets `cacheTime` wrong ([exhibit A](https://twitter.com/TkDodo/status/1575893647392145408)). It sounds like "the amount of time that data is cached for", but that is not correct.

`cacheTime` does nothing as long as a query is still in used. It only kicks in as soon as the query becomes unused. After the time has passed, data will be "garbage collected" to avoid the cache from growing (see also [this explanation](https://twitter.com/TkDodo/status/1576078947887915013)).

RTK Query has the same feature - their prop is called [keepUnusedDataFor](https://redux-toolkit.js.org/rtk-query/api/createApi#keepunuseddatafor). I think this is quite descriptive but also a bit long.

### proposal

- rename `cacheTime` to `gcTime`

`gc` is referring to "garbage collect" time. It's a bit more technical, but also a quite [well known abbreviation](<https://en.wikipedia.org/wiki/Garbage_collection_(computer_science)>) in computer science.

Also, it is not something that most people will have to customize. The default of 5 minutes is usually fine. A rename will reduce the chance that this property is mixed up with `staleTime`.

Lastly, if someone doesn't immediately know what `gcTime` stands for, they will (hopefully) consult the docs. This is a lot better than _thinking_ they know what `cacheTime` does.

Here is an old discussion on that topic:

- https://github.com/TanStack/query/discussions/1217

Mamoanwar97:
closes #4678

### DIFF
--- a/docs/react/guides/caching.md
+++ b/docs/react/guides/caching.md
@@ -26,7 +26,7 @@ Let's assume we are using the default `gcTime` of **5 minutes** and the default 
     - Note that regardless of whether both `fetchTodos` query functions are identical or not, both queries' [`status`](../reference/useQuery) are updated (including `isFetching`, `isLoading`, and other related values) because they have the same query key.
   - When the request completes successfully, the cache's data under the `['todos']` key is updated with the new data, and both instances are updated with the new data.
 - Both instances of the `useQuery({ queryKey: ['todos'], queryFn: fetchTodos })` query are unmounted and no longer in use.
-  - Since there are no more active instances of this query, a cache timeout is set using `gcTime` to delete and garbage collect the query (defaults to **5 minutes**).
+  - Since there are no more active instances of this query, a garbage collection timeout is set using `gcTime` to delete and garbage collect the query (defaults to **5 minutes**).
 - Before the cache timeout has completed, another instance of `useQuery({ queryKey: ['todos'], queyFn: fetchTodos })` mounts. The query immediately returns the available cached data while the `fetchTodos` function is being run in the background. When it completes successfully, it will populate the cache with fresh data.
 - The final instance of `useQuery({ queryKey: ['todos'], queryFn: fetchTodos })` unmounts.
 - No more instances of `useQuery({ queyKey: ['todos'], queryFn: fetchTodos })` appear within **5 minutes**.
--- a/docs/react/guides/caching.md
+++ b/docs/react/guides/caching.md
@@ -14,7 +14,7 @@ This caching example illustrates the story and lifecycle of:
 - Inactive Queries
 - Garbage Collection
 
-Let's assume we are using the default `cacheTime` of **5 minutes** and the default `staleTime` of `0`.
+Let's assume we are using the default `gcTime` of **5 minutes** and the default `staleTime` of `0`.
 
 - A new instance of `useQuery({ queryKey: ['todos'], queryFn: fetchTodos })` mounts.
   - Since no other queries have been made with the `['todos']` query key, this query will show a hard loading state and make a network request to fetch the data.
@@ -26,7 +26,7 @@ Let's assume we are using the default `cacheTime` of **5 minutes** and the defau
     - Note that regardless of whether both `fetchTodos` query functions are identical or not, both queries' [`status`](../reference/useQuery) are updated (including `isFetching`, `isLoading`, and other related values) because they have the same query key.
   - When the request completes successfully, the cache's data under the `['todos']` key is updated with the new data, and both instances are updated with the new data.
 - Both instances of the `useQuery({ queryKey: ['todos'], queryFn: fetchTodos })` query are unmounted and no longer in use.
-  - Since there are no more active instances of this query, a cache timeout is set using `cacheTime` to delete and garbage collect the query (defaults to **5 minutes**).
+  - Since there are no more active instances of this query, a cache timeout is set using `gcTime` to delete and garbage collect the query (defaults to **5 minutes**).
 - Before the cache timeout has completed, another instance of `useQuery({ queryKey: ['todos'], queyFn: fetchTodos })` mounts. The query immediately returns the available cached data while the `fetchTodos` function is being run in the background. When it completes successfully, it will populate the cache with fresh data.
 - The final instance of `useQuery({ queryKey: ['todos'], queryFn: fetchTodos })` unmounts.
 - No more instances of `useQuery({ queyKey: ['todos'], queryFn: fetchTodos })` appear within **5 minutes**.
--- a/docs/react/reference/useQuery.md
+++ b/docs/react/reference/useQuery.md
@@ -93,7 +93,7 @@ const {
   - If set to `Infinity`, the data will never be considered stale
 - `gcTime: number | Infinity`
   - Defaults to `5 * 60 * 1000` (5 minutes) or `Infinity` during SSR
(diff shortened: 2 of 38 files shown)


packages/query-core/src/removable.ts
```diff
@@ -11,18 +11,18 @@ export abstract class Removable {
...
     }
   }
 
-  protected updateCacheTime(newCacheTime: number | undefined): void {
-    // Default to 5 minutes (Infinity for server-side) if no cache time is set
-    this.cacheTime = Math.max(
-      this.cacheTime || 0,
-      newCacheTime ?? (isServer ? Infinity : 5 * 60 * 1000),
+  protected updateGcTime(newGcTime: number | undefined): void {
+    // Default to 5 minutes (Infinity for server-side) if no garbage collector time is set
```
TkDodo, line 22:
```suggestion
    // Default to 5 minutes (Infinity for server-side) if no gcTime is set
```

packages/query-core/src/tests/hydration.test.tsx
```diff
@@ -119,13 +119,13 @@ describe('dehydration and rehydration', () => {
     queryClient.clear()
   })
 
-  test('should use the cache time from the client', async () => {
+  test('should use the garbage collector time from the client', async () => {
```
TkDodo, line 122:
```suggestion
  test('should use the garbage collection time from the client', async () => {
```

packages/query-core/src/tests/query.test.tsx
```diff
@@ -28,25 +28,25 @@ describe('query', () => {
     queryClient.clear()
   })
 
-  test('should use the longest cache time it has seen', async () => {
+  test('should use the longest garbage collector time it has seen', async () => {
```
TkDodo, line 31:
```suggestion
  test('should use the longest garbage collection time it has seen', async () => {
```

packages/query-core/src/tests/queryClient.test.tsx
```diff
@@ -578,31 +578,31 @@ describe('queryClient', () => {
       expect(second).toBe(first)
     })
 
-    test('should be able to fetch when cache time is set to 0 and then be removed', async () => {
+    test('should be able to fetch when garbage collector time is set to 0 and then be removed', async () => {
```
TkDodo, line 581:
```suggestion
    test('should be able to fetch when garbage collection time is set to 0 and then be removed', async () => {
```

packages/query-core/src/tests/queryClient.test.tsx
```diff
@@ -578,31 +578,31 @@ describe('queryClient', () => {
...
+        gcTime: 0,
       })
       expect(result).toEqual(1)
       await waitFor(() =>
         expect(queryClient.getQueryData(key1)).toEqual(undefined),
       )
     })
 
-    test('should keep a query in cache if cache time is Infinity', async () => {
+    test('should keep a query in cache if garbage collector time is Infinity', async () => {
```
TkDodo, line 597:
```suggestion
    test('should keep a query in cache if garbage collection time is Infinity', async () => {
```

packages/solid-query/src/__tests__/createQuery.test.tsx
```diff
@@ -3983,7 +3983,7 @@ describe('createQuery', () => {
 
     const query = queryCache.find({ queryKey: key })
     // @ts-expect-error
-    expect(query!.cacheTimeout).toBe(undefined)
+    expect(query!.gcTimeout).toBe(undefined)
```
TkDodo, line 3986: same here, this doesn't actually do anything.

TkDodo (commented):
thank you ❤️ . Just some minor comments from my side:

docs/react/guides/caching.md
```diff
@@ -26,7 +26,7 @@ Let's assume we are using the default `cacheTime` of **5 minutes** and the defau
     - Note that regardless of whether both `fetchTodos` query functions are identical or not, both queries' [`status`](../reference/useQuery) are updated (including `isFetching`, `isLoading`, and other related values) because they have the same query key.
   - When the request completes successfully, the cache's data under the `['todos']` key is updated with the new data, and both instances are updated with the new data.
 - Both instances of the `useQuery({ queryKey: ['todos'], queryFn: fetchTodos })` query are unmounted and no longer in use.
-  - Since there are no more active instances of this query, a cache timeout is set using `cacheTime` to delete and garbage collect the query (defaults to **5 minutes**).
+  - Since there are no more active instances of this query, a cache timeout is set using `gcTime` to delete and garbage collect the query (defaults to **5 minutes**).
```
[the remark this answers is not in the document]
Mamoanwar97: Done ✅

docs/react/reference/useQuery.md
```diff
@@ -91,9 +91,9 @@ const {
   - Defaults to `0`
   - The time in milliseconds after data is considered stale. This value only applies to the hook it is defined on.
   - If set to `Infinity`, the data will never be considered stale
-- `cacheTime: number | Infinity`
+- `gcTime: number | Infinity`
   - Defaults to `5 * 60 * 1000` (5 minutes) or `Infinity` during SSR
-  - The time in milliseconds that unused/inactive cache data remains in memory. When a query's cache becomes unused or inactive, that cache data will be garbage collected after this duration. When different cache times are specified, the longest one will be used.
+  - The time in milliseconds that unused/inactive cache data remains in memory. When a query's cache becomes unused or inactive, that cache data will be garbage collected after this duration. When different garbage collector times are specified, the longest one will be used.
```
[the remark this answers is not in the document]
Mamoanwar97: Done ✅

packages/query-core/src/removable.ts
Mamoanwar97: Done ✅

packages/query-core/src/tests/queryClient.test.tsx
Mamoanwar97: Done ✅

packages/query-core/src/tests/queryClient.test.tsx
Mamoanwar97: Done ✅

packages/react-query/src/__tests__/useQuery.test.tsx
```diff
@@ -4029,7 +4029,7 @@ describe('useQuery', () => {
 
     const query = queryCache.find({ queryKey: key })
     // @ts-expect-error
-    expect(query!.cacheTimeout).toBe(undefined)
+    expect(query!.gcTimeout).toBe(undefined)
```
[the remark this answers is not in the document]
Mamoanwar97: I tried spying on `setTimeout` here and in solidQuery tests and the tests passed. 

but for some reason in react-query tests,  jest is raising a warning and recommends using `jest.useFakeTimers()` when spying on `setTimeout`, but in solid query tests, which have the same implementation jest is not raising any warning.

when I tried to add `jest.useFakeTimers()` to the test in react-query, the test failed and useFakerTimers affected other tests to fail as well.

packages/solid-query/src/__tests__/createQuery.test.tsx
Mamoanwar97: solved ✅

PR #4837, opened 2023-01-18
docs(useMutation): clarify `mutationFn` option default

louis-young:
This pull request clarifies the `useMutation` `mutationFn` option default as being the global mutation configuration `mutationFn`. 

### Notes

- This change came out of a conversation had in the TanStack Discord server between another member, Dominik, and me. 
- Please let me know if an issue would be useful for tracking.

### DIFF
--- a/docs/react/reference/useMutation.md
+++ b/docs/react/reference/useMutation.md
@@ -43,8 +43,7 @@ mutate(variables, {
 **Options**
 
 - `mutationFn: (variables: TVariables) => Promise<TData>`
-  - Optional
-  - Defaults to the global mutation configuration `mutationFn`, which is `undefined` by default.
+  - **Required, but only if no default mutation function has been defined**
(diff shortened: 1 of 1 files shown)


docs/react/reference/useMutation.md
```diff
@@ -43,7 +43,8 @@ mutate(variables, {
 **Options**
 
 - `mutationFn: (variables: TVariables) => Promise<TData>`
-  - **Required**
+  - Optional
+  - Defaults to the global mutation configuration `mutationFn`, which is `undefined` by default.
```
TkDodo, line 47: Hmm, we will die at runtime if the default of `undefined` is used. To align with the useQuery docs, I would say:

```suggestion
  - **Required, but only if no default mutation function has been defined**
```

louis-young: Awesome suggestion, nice one 👍

PR #4838, opened 2023-01-19
fix: simplify vue hooks, fix types

DamianOsipiuk:


packages/react-query/src/__tests__/useInfiniteQuery.test.tsx
```diff
@@ -846,14 +846,12 @@ describe('useInfiniteQuery', () => {
...
-        abortListeners.push(abortListener)
-        signal.onabort = onAbort
-        signal.addEventListener('abort', abortListener)
-      }
+      const onAbort = jest.fn()
+      const abortListener = jest.fn()
+      onAborts.push(onAbort)
+      abortListeners.push(abortListener)
+      signal.onabort = onAbort
+      signal.addEventListener('abort', abortListener)
```
TkDodo, line 854: do you know why eslint didn't report this issue? It does show up as red squiggles in my editor, but the pipeline is still green ...

Screenshot 2023-01-20 at 13 32 36

packages/react-query/src/__tests__/useQuery.test.tsx
```diff
@@ -5729,6 +5730,7 @@ describe('useQuery', () => {
           queryFn: async ({ signal }) => {
             count++
             await sleep(10)
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
             return `${signal ? 'signal' : 'data'}${count}`
```
TkDodo, line 5734: here we could remove the `signal` because destructuring it from `ctx` "counts" as the signal being consumed:

```suggestion
            return `signal${count}`
```

packages/solid-query/src/__tests__/createQuery.test.tsx
```diff
@@ -5792,6 +5793,7 @@ describe('createQuery', () => {
           queryFn: async ({ signal }) => {
             count++
             await sleep(10)
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
             return `${signal ? 'signal' : 'data'}${count}`
```
TkDodo, line 5797: same here:

```suggestion
            return `signal${count}`
```

TkDodo (approved):

packages/react-query/src/__tests__/useInfiniteQuery.test.tsx
DamianOsipiuk: No idea, it failed for me, when i ran `test` script locally.

PR #4840, opened 2023-01-19
feat(solid-query): SSR streaming support

marbemac:
Working off of the progress in https://github.com/TanStack/query/pull/4803. 

This refactors `createBaseQuery` to an alternative approach that consolidates to a single `createResource` call. Instead of coordinating two createResource calls - one that sends down the dehydrated query state, and one that sends down the query data - this PR uses a single createResource that tracks the entire query state, and leverages solid's native `onHydrate` functionality to hydrate that state on the client. This means we also get support for SSR streaming (and optional defer for specific requests just by setting the `deferStream` option)!

I also expanded the `solid-start-streaming` example to better showcase the various ssr patterns (and make it easier to experiment in real world).


packages/solid-query/src/createBaseQuery.ts
```diff
@@ -45,40 +50,109 @@ export function createBaseQuery<
...
+        if (info.value) {
+          hydrate(queryClient(), {
+            queries: [
+              {
+                queryKey: defaultedOptions.queryKey,
+                queryHash: defaultedOptions.queryHash,
+                state: info.value,
+              },
+            ],
+          })
```
marbemac, line 137: Here I'm assuming that hydrating mutations from server isn't really a use case - or is it?

packages/solid-query/src/createBaseQuery.ts
```diff
@@ -45,40 +50,109 @@ export function createBaseQuery<
...
+            () => mutate,
+            () => refetch(),
+          )
+        }
+      },
+    },
+  )
 
-  onCleanup(() => unsubscribe())
+  onCleanup(() => {
```
marbemac, line 150: Observing some strange behavior on the server with `onCleanup`. Pop a log in here, and load one of the pages that contains two queries... note how `onCleanup` is only triggered once on the server, despite two queries. Not sure why.

packages/solid-query/src/createBaseQuery.ts
```diff
@@ -45,40 +50,109 @@ export function createBaseQuery<
...
-        // ref: https://www.solidjs.com/docs/latest/api#updating-stores
-        unwrappedResult.data = emptyData as any as undefined
-      }
-      setState(unwrap(unwrappedResult))
-      mutate(() => unwrap(result.data))
-      refetch()
-    })()
-  })
+        if (!unsubscribe) {
+          unsubscribe = createClientSubscriber(
```
marbemac, line 141: For queries that were created in SSR, `onHydrate` will be called on the client, at which point we can hydrate the `queryClient` for that particular query, and then lazy subscribe to the observer.

packages/solid-query/src/createBaseQuery.ts
```diff
@@ -45,40 +50,109 @@ export function createBaseQuery<
...
+            unsubscribe = createClientSubscriber(
+              () => mutate,
+              () => refetch(),
+            )
+          }
+
+          if (!state.isInitialLoading) {
+            resolve(state)
           }
-          resolve(unwrap(state.data))
```
marbemac, line 55: TBH, wasn't sure when/where/why to `unwrap` versus not `unwrap`... any rules of thumb?

packages/solid-query/src/__tests__/createQuery.test.tsx
```diff
@@ -1579,7 +1579,6 @@ describe('createQuery', () => {
 
     // Fetch query
     expect(states[0]).toMatchObject({
-      data: undefined,
```
marbemac, line 1582: Is it important that this property be defined on the original state object (e.g. when we `states.push({ ...state })`)? When accessing through the proxy, e.g. `state.data`, it will come back as undefined as expected.

ardeora:
Thanks for working on this! I really like it. Any idea why `refetchOnWindowFocus` doesn't seem to work?

marbemac:
> Thanks for working on this! I really like it. Any idea why refetchOnWindowFocus doesn't seem to work?

No problem, it's been fun digging into tanstack query a bit more. Hmm, `refetchOnWindowFocus: true` is the default behavior right - are you saying it's not re-doing the queries for you after refocusing? Seems to work for me (do note that there's a staleTime of 10s on both queries).

ardeora:
Sorry I dont know what just happened. I accidentally edited your comment. But take a look at the codesandbox example created above! Doesn't seem to work on my device :/ 

https://user-images.githubusercontent.com/45807386/213616643-73dbcca5-2cf8-4fc7-849a-b3e467acb49b.mov

marbemac:
Ah, maybe that behavior is due to https://github.com/TanStack/query/commit/096da932ec5c8865125c65d39f631084c94555f4? I branched off of your ssr branch, and then merged in latest v5 before starting my work. Perhaps trying switching back and forth between browser tabs to see if the refetch on visibility change occurs?

ardeora:
> Ah, maybe that behavior is due to [096da93](https://github.com/TanStack/query/commit/096da932ec5c8865125c65d39f631084c94555f4)? I branched off of your ssr branch, and then merged in latest v5 before starting my work. Perhaps trying switching back and forth between browser tabs to see if the refetch on visibility change occurs?

That seems to be it! If I change tabs I can see it refetch on coming back to the tab

packages/solid-query/src/createBaseQuery.ts
```diff
@@ -45,40 +49,100 @@ export function createBaseQuery<
...
+
+  const createClientSubscriber = (refetch: (info?: unknown) => void) => {
+    return observer.subscribe((result) => {
+      notifyManager.batchCalls(() => {
+        const unwrappedResult = { ...unwrap(result) }
+        setState(unwrappedResult)
+        refetch()
+      })()
+    })
+  }
```
ardeora, line 77:
```suggestion
  const createClientSubscriber = (refetch: (info?: unknown) => void, hydrated: boolean = false) => {
    if (hydrated) {
        defaultedOptions.refetchOnMount = false
        observer.setOptions(defaultedOptions)
     }
    return observer.subscribe((result) => {
      notifyManager.batchCalls(() => {
        const unwrappedResult = { ...unwrap(result) }
        setState(unwrappedResult)
        refetch()
      })()
    })
  }
```

Can we also add this to the client subscriber? This would allow us to not refetch a query that was rendered on the server, even if a staleTime wasn't set. Similar to how CSR apps behave

packages/solid-query/src/createBaseQuery.ts
```diff
@@ -45,40 +49,100 @@ export function createBaseQuery<
...
+                queryHash: defaultedOptions.queryHash,
+                state: info.value,
+              },
+            ],
+          })
+        }
 
-  onCleanup(() => unsubscribe())
+        if (!unsubscribe) {
+          unsubscribe = createClientSubscriber((i?: unknown) => refetch(i))
```
ardeora, line 134:
```suggestion
          unsubscribe = createClientSubscriber((i?: unknown) => refetch(i), true)
```
This would change to a true flag for hydrated queries only

packages/solid-query/src/__tests__/createQuery.test.tsx
ardeora: This fine for now! I think libraries built on top of solid-query would need to know about this caveat. That's worth a consideration

packages/solid-query/src/createBaseQuery.ts
marbemac: For sure - I'll give it a try now. Would also be nice if we could figure out an actual "ssr" test that covers that hydrate code branch, and this stale behavior. Will see if I can figure that out.

marbemac: 👍 done in 9fc9e3d7142b2b812f4950b678aad6b080c7fc03. Haven't figured out the ssr test... was looking for inspiration, but can't even find tests in the solidjs repo that cover ssr related functionality.

ardeora: Great work! I can add the tests for SSR in a separate PR. Could you resolve the conflicts before I run the CI workflow?

marbemac: 👍 done

ardeora:
@TkDodo  @marbemac made some awesome additions to createBaseQuery and this should be a good way to integrate both CSR and SSR in solid-query! Would love it if you could review it and merge the changes whenever convenient 😄 I'll close my PR once this gets merged

ardeora:
@marbemac I made a PR with some fixes with this approach. 

The PR fixes two issues:

1. If we just use SSR (no Streaming) with Solid Start, when the query is hydrated, The data is updated due to it being a part of the resource. But the state store will be in its initial state. Where stuff like isFetching will be true, dataUpdatedAt will be 0, etc. This happens because there is no reactivity on the server! So even though we update the state store in the server subscription handler. The state variable never updates on server rendered DOM elements. So after the query is hydrated we can safely update the store
2. The second issue will be noticed if we have two queries with the same key but one of them has initialData set. When the second query mounts, it will already have data set but it will still wait for the fetcher to resolve instead of deduping the data.

This also was the reason why the unsubscribe function wasn't working correctly on the server. It is fixed now.

packages/solid-query/src/createBaseQuery.ts
ardeora: The PR I just created will solve this issue

marbemac:
@ardeora great! I'll take a look in a few hours.

ardeora:
Hello @marbemac! Could you rebase again from the v5 branch? I'll create another PR to fix the svelte-query tests then :D If they haven't been fixed already and get this PR merged

marbemac:
Sure thing! Can get that done today.

marbemac:
@ardeora should be all set

ardeora:
@marbemac Thanks again for all the hard work on this. The changes have been merged now. v5 is going to be 💯

PR #4848, opened 2023-01-21
fix: mutate callbacks called if mutation started after unmount

janlat:
https://github.com/TanStack/query/discussions/4804

### DIFF
--- a/packages/react-query/src/__tests__/useMutation.test.tsx
+++ b/packages/react-query/src/__tests__/useMutation.test.tsx
@@ -1044,7 +1044,7 @@ describe('useMutation', () => {
     )
   })
 
-  test('should go to error state if onSuccess callback errors', async () => {
+  it('should go to error state if onSuccess callback errors', async () => {
     const error = new Error('error from onSuccess')
     const onError = jest.fn()
 
@@ -1079,7 +1079,7 @@ describe('useMutation', () => {
     expect(onError).toHaveBeenCalledWith(error, 'todo', undefined)
   })
 
-  test('should go to error state if onError callback errors', async () => {
+  it('should go to error state if onError callback errors', async () => {
     const error = new Error('error from onError')
     const mutateFnError = new Error('mutateFnError')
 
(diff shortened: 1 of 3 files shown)


TkDodo:
the interesting part here is that this test in the query-core now fails:

```
const mutation = new MutationObserver(queryClient, {
  mutationFn: async () => {
    return 'update'
  },
  onMutate: (text) => text,
})

await mutation.mutate()
expect(mutation.getCurrentResult().data).toEqual('update')
```

with:

```
Error: expect(received).toEqual(expected) // deep equality

Expected: "update"
Received: undefined
```

we:
- create a mutation
- call `.mutate`
- have no result on the observer after that

This is because with this change, we would actually have to call `mutation.subscribe(listener)` to be have our observer be attached to the mutation.

I think the better fix might be to only call the callbacks if we still have listeners at that point?

I did that here: 2ad4a753bff3e472383cda56a937f86ae79a4c65

TkDodo:
@janlat @louis-young what do you think?

louis-young:
Interesting. Well spotted! I _think_ I'd prefer not to have to explicitly subscribe to the mutation; I think attacking this from a "should we call the callbacks" on the observer angle is sensible and probably safer, especially considering that's the specific behaviour we're trying to adjust. Good idea, I like it 👍

packages/react-query/src/__tests__/useMutation.test.tsx
```diff
@@ -1154,4 +1154,60 @@ describe('useMutation', () => {
...
+          <button onClick={() => setShow(false)}>hide</button>
+          {show && <Component />}
+        </div>
+      )
+    }
+
+    function Component() {
+      const mutation = useMutation({
+        mutationFn: async (text: string) => {
+          await sleep(10)
```
louis-young, line 1177: Do we need to sleep here? I think this might be redundant now we're no longer using an explicit sleep (nice change by the way) to wait for the unmount and mutation to happen.

```suggestion
```

TkDodo: not necessarily, but asynchronous mutation / query functions are preferred in most test because otherwise, render cycles might be batched together and you'd expect to have a loading state where you'll have none because of that.

louis-young: I see, that makes sense. Thanks 🙂

PR #4850, opened 2023-01-22
remove-useless-ts-ignore

therealrinku:
devtools.md useless ts-ignore remove

### DIFF
--- a/docs/react/devtools.md
+++ b/docs/react/devtools.md
@@ -123,7 +123,6 @@ function App() {
   const [showDevtools, setShowDevtools] = React.useState(false)
 
   React.useEffect(() => {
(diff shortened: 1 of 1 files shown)


docs/react/devtools.md
```diff
@@ -123,7 +123,6 @@ function App() {
   const [showDevtools, setShowDevtools] = React.useState(false)
 
   React.useEffect(() => {
-    // @ts-ignore
```
TkDodo, line 126: it's not unnecessary if you copy-paste this code into your typescript codebase. The error is:

```
TS2339: Property 'toggleDevtools' does not exist on type 'Window & typeof globalThis'.
```

PR #4852, opened 2023-01-22
docs(svelte-query): Correctly type layout data when using prefetchQuery

Renni771:
A minor issue but the types for some `+layout.svelte` data prop was incorrect. This PR addresses that issue.


lachlancollins (approved):
Agh whoops, I wonder if I should open an issue in sveltekit for detecting wrong imports like this, it already has detection of similar issues.

yaser432 (approved):
- [ ]

PR #4861, opened 2023-01-23
docs(svelte-query): Fix variable reference issue on sveltekit doc

nicolasmendonca:
Just adding my two cents here. Here's a small documentation fix around the sveltekit data load function.

`posts` wasn't defined nor extracted from `data`. So I just added `data.posts` to make it make sense.

btw I love TanStack Query! 🙌🏼 Congrats to the team!


lachlancollins (approved):
@nicolasmendonca thank you!

PR #4877, opened 2023-01-25
fix(svelte-query): Fix createQueries result type

### LINKED ISSUE #4832: Wrong return type in createQueries / Svelte
### Describe the bug

The createQueries function is returning an array typed as CreateQueriesResult<T> which is equivalent as Readable<Readable<T>[]> but is actually returning Readable<T[]>.
In other words, typescript accuses the return of being a store with the value of an array of stores whose value is the query return but in reality the return has the format of a store with the value of an array of query returns

https://codesandbox.io/s/xenodochial-haze-yfolbg

### Steps to reproduce

Just open the project

### Expected behavior

That the returned type corresponds with the type accused by the typescript

Every time

### Screenshots or Videos

Typescript doesn't yell at me for this but I get a runtime error because productStore is not a store
image
image
Typescript complains that I'm not using it as a store but it works in the browser
image

### Platform

- OS: ARCH LINUX
- BROWSER: CHROME

### TanStack Query version

4.22.0

### TypeScript version

4.9.4

### Additional context

_No response_

lachlancollins:
Fixes #4832. The results from createQueries were being incorrectly typed as each being svelte stores, as they were using the CreateQueryResult type. This changes each result to directly use QueryObserverResult. I've also added a very simple test.

Also @TkDodo I noticed that svelte-query alone takes an array of queries directly into createQueries (e.g. `createQueries([...])`) while the other packages all use an object with the queries key (e.g. `useQueries({ queries: [...] })`). Is this something I should change in v4, or wait until v5?

### DIFF
--- a/packages/svelte-query/src/__tests__/CreateMutation.svelte
+++ b/packages/svelte-query/src/__tests__/CreateMutation.svelte
@@ -1,5 +1,9 @@
 <script lang="ts">
-  import { createMutation, QueryClient, type CreateMutationOptions } from '../index'
+  import {
+    createMutation,
+    QueryClient,
+    type CreateMutationOptions,
+  } from '../index'
   import { setQueryClientContext } from '../context'
 
   export let options: CreateMutationOptions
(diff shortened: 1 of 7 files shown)


TkDodo:
> Is this something I should change in v4, or wait until v5?

If we have shipped it with that api, we should only change it in v5 please.

PR #4884, opened 2023-01-26
perf(persist): subscriber calls persistQueryClientStore only on cache-related events

emzet93:
By default `persistQueryClientSubscribe` was calling `persistQueryClientSave` on every single change in query cache.
It was also triggered by events like `observerOptionsUpdated` which are emitted very often, usually on every single re-render of component/hook using `useQuery`.

For performance improvement, we can limit list of events triggering `persistQueryClientSave` only for those, which actually have influence on the cache itself.


packages/query-persist-client-core/src/persist.ts
```diff
@@ -121,16 +121,24 @@ export async function persistQueryClientSave({
 export function persistQueryClientSubscribe(
   props: PersistedQueryClientSaveOptions,
 ) {
+  // List of events which should trigger persistQueryClientSave
+  // All events connected with observers can be ignored for performance reason
+  const cacheRelatedEvents = ['added', 'removed', 'updated']
```
emzet93, line 126: I wanted to set type for this array to make typescript yell about potential typos, but I didn't find a way to import that properly.

TkDodo: we can do `as const` :

```suggestion
  const cacheRelatedEvents = ['added', 'removed', 'updated'] as const
```

emzet93: won't work unfortunately:
```
TS2345: Argument of type '"added" | "removed" | "updated" | "observerAdded" | "observerRemoved" | "observerOptionsUpdated"' is not assignable to parameter of type '"added" | "removed" | "updated"'.   Type '"observerAdded"' is not assignable to type '"added" | "removed" | "updated"'.
```
It interprets `cacheRelatedEvents.includes(event.type)` as invalid as we basically compare different types

louis-young (commented):
Thanks for putting this together 🚀 

Is it worth reusing a predicate function here? Perhaps it's a hasty abstraction, but something along the lines of `isCacheableEvent`.

With regards to the type: as Dominik said, an `as const` assertion will narrow the type from a string array to an array of the literal string values, preventing any human error 🙂

packages/query-persist-client-core/src/persist.ts
emzet93: sth like that would work:
```
export const isCacheableEvent = (event: QueryCacheNotifyEvent) => {
  const cacheRelatedEvents: QueryCacheNotifyEvent['type'][] = [
    'added',
    'removed',
    'updated',
  ]
  return cacheRelatedEvents.includes(event.type)
}```

Although it has few flaws:
1. It would require similar function for mutation cache. Which maybe isn't that bad, `cacheRelatedEvents` is the same for mutation and query now but probably doesn't have to be forever.
2. Not sure where should I define this functions and where export it from
3. Currently `QueryCacheNotifyEvent` interface is not exported, so we would have to either export that or define `isCacheableEvent` in the same file 🤔

cc @louis-young

Comment (louis-young) [reply packages/query-persist-client-core/src/persist.ts:126]: Hmm, I think we need to use a partial type here. We could extend the `NotifyEventQueryAdded...` interfaces from a base `NotifyEventQueryType` interface that specifies the type as a string union of the possible event types and extract the members of that union to use as the type here. There are other ways to attack this too.

If we didn't want to do anything drastic, I don't know if we want to export `QueryCacheNotifyEvent` (the type on which `type` exists), but if we did then we could do something like this:

```ts
  const cacheRelatedEvents: Array<Partial<QueryCacheNotifyEvent['type']>> = [
    'added',
    'removed',
    'updated',
  ]
  ```

Comment (louis-young) [reply packages/query-persist-client-core/src/persist.ts:126]: I didn't see your reply as we posted it at almost the same time!

I think the update types are the same for both queries and mutations, so if we went down the road of my first suggestion (using a string union of the possible types) then we could use that as a source of truth for both queries and mutations. 

Dominik probably has a few clever ideas up his sleeve so I'll defer to him 🙂

Comment (emzet93) [reply packages/query-persist-client-core/src/persist.ts:126]: Yeah so basically we have similar ideas 😃 The solution you pasted here is more or less the same I suggested above. I think we don't even need to mark it as Partial.

But still it depends on decision if those types should be exported or not

Comment (emzet93) [reply packages/query-persist-client-core/src/persist.ts:126]: post at the same time again 😅
Yep, so let's wait for Dominik to share his opinion here. Thanks!

Comment (louis-young) [reply packages/query-persist-client-core/src/persist.ts:126]: We did indeed! Yeah, you're right. There's no need to use a partial array type 👍 I agree, I think either approach still depends on _some_ type being exported.

Comment (TkDodo) [reply packages/query-persist-client-core/src/persist.ts:126]: if we can export the `QueryCacheNotifyEvent`, we'd also need to do the same for `MutationCacheNotifyEvent` and then have our array kinda be both?

so:

```
  const cacheRelatedEvents:
    | Array<QueryCacheNotifyEvent['type']>
    | Array<MutationCacheNotifyEvent['type']> = ['added', 'removed', 'updated']
```

?

Comment (emzet93) [reply packages/query-persist-client-core/src/persist.ts:126]: Yep exactly, that works.
Or we could create two separate arrays for query and mutation if there is a chance that event types for those two will be different in the future.

Comment (louis-young) [reply packages/query-persist-client-core/src/persist.ts:126]: I think I'd be in favour of sharing the event types between queries and mutations. I'd propose something like this:

1. Define a string union of shared event types:

```ts
export type NotifyEventType =
  | 'added'
  | 'removed'
  | 'updated'
  | 'observerAdded'
  | 'observerRemoved'
  | 'observerResultsUpdated'
  | 'observerOptionsUpdated'
```

2. Create a `NotifyEvent` interface that we extend from to type the event types of all of the different query and mutation events:

```ts
interface NotifyEvent {
  type: NotifyEventType
}

interface NotifyEventQueryAdded extends NotifyEvent {
  type: 'added'
  query: Query<any, any, any, any>
}
```

3. Write a predicate that receives the event type as an argument:

```ts
const isCacheableEventType = (eventType: NotifyEventType) => {
  const CACHEABLE_EVENT_TYPES: Array<NotifyEventType> = [
    'added',
    'removed',
    'updated',
  ]

  return CACHEABLE_EVENT_TYPES.includes(eventType)
}
```

4. Consume said predicate:

```ts
  const unsubscribeQueryCache = props.queryClient
    .getQueryCache()
    .subscribe((event) => {
      if (isCacheableEventType(event.type)) {
        persistQueryClientSave(props)
      }
    })
```

What do you think? I'd be happy to implement this if you like 🙂

emzet93: I like your solution a lot! I just checked and everything works perfectly.
I have it almost implemented so I can push it if it's ok with you.

I placed `NotifyEvent` and `NotifyEventType` in `query-core/types`.
But I'm not sure where in the files we should put `isCacheableEventType` function. Is `query-core/utils` a good place?

Thanks for any suggestions :)

louis-young: Awesome! Of course, please feel free 🙂 

I think I'll leave the decision to Dominik as he has a _much_ better idea of where things should live. 

Thanks for helping out 🚀

packages/query-persist-client-core/src/persist.ts
```diff
@@ -52,6 +53,20 @@ export interface PersistQueryClientOptions
     PersistedQueryClientSaveOptions,
     PersistQueryClienRootOptions {}
 
+/**
+ * Checks if emitted event is about cache change and not about observers.
+ * Useful for persist, where we only want to trigger save when cache is changed.
+ */
+function isCacheableEventType(eventType: NotifyEventType) {
```
emzet93, line 60: @TkDodo I placed this function here as in internal one for now. Let me know if there is a better place for that.

louis-young: I think we should check that Dominik is happy with [this idea](https://github.com/TanStack/query/pull/4884#discussion_r1088716259), but I think we could start on the tests for the new behaviour if you're up for it 🙂

emzet93: sure I am. Although no idea where to start

louis-young: It looks like the relevant tests are here: 

https://github.com/TanStack/query/blob/main/packages/query-persist-client-core/src/__tests__/persist.test.ts#L22

I think this is probably going to be non-trivial to test. I'd be happy to take a look but I won't have time for a little while, although I think Dominik said he'd be willing to lend a hand 🤞

emzet93: awesome, thanks!

packages/query-persist-client-core/src/persist.ts
```diff
@@ -52,6 +53,20 @@ export interface PersistQueryClientOptions
...
+  const CACHEABLE_EVENT_TYPES: Array<NotifyEventType> = [
+    'added',
+    'removed',
+    'updated',
+  ]
```
TkDodo, line 65: could we move this out of the `function isCacheableEventType` ? There's no need to re-create this every time.

packages/query-persist-client-core/src/persist.ts
```diff
@@ -52,6 +53,20 @@ export interface PersistQueryClientOptions
...
+
+  return CACHEABLE_EVENT_TYPES.includes(eventType)
```
TkDodo, line 67: I'd also prefer if we'd name this just `const cacheableEventTypes`.

TkDodo (commented):
the only thing missing for me is tests. What we would need is some sort of test that writes a custom persister that basically just tracks with `jest.fn` how often we call the persistence write function, then create an observer, mount and unmount, maybe set some options etc and then make sure that we call the write function not too often :)

emzet93:
> the only thing missing for me is tests. What we would need is some sort of test that writes a custom persister that basically just tracks with `jest.fn` how often we call the persistence write function, then create an observer, mount and unmount, maybe set some options etc and then make sure that we call the write function not too often :)

Thanks, it all makes sense. I'm not sure about my availability this week, but I hope I'll find some time to add it :)

packages/query-persist-client-core/src/persist.ts
raegen: Good choice, literally opened the rest of the conversation to argue for exporting the event type. The events of the mentioned type (`QueryCacheNotifyEvent`) are not internal, but rather exposed in the API through `Subscribable<QueryCacheListener>.subscribe` and as such the only thing not exporting it does is cause ugly, redundant code in consumer. It's generally a simple matter: if consumer can access something -> export its type.
Cheers and kudos for persistent client, it's a huge value idea.

TkDodo (approved):

PR #4888, opened 2023-01-26
Update isLoading behaviour

### LINKED ISSUE #4855: Change status 'loading' -> 'pending' and update isLoading behaviour
- Rename status: loading to status: pending and the derived boolean isLoading to isPending
- Introduce a new, derived boolean flag isLoading that is implemented as isPending && isFetching
- Keep `isInitialLoading` as an alias for `isLoading`, to avoid a breaking change for existing v4 users 
- Rename `isLoading` in mutations to `isPending`

For the reasoning of this issue checkout the [first point in the v5 roadmap](https://github.com/TanStack/query/discussions/4252) or [checkout this video](https://youtube.com/watch?v=vvq8iZCZ-J8&feature=shares&t=6633)

prateek3255:
- Rename status: loading to status: pending and the derived boolean isLoading to isPending
- Introduce a new, derived boolean flag isLoading that is implemented as isPending && isFetching
- Keep isInitialLoading as an alias for isLoading, to avoid a breaking change for existing v4 users
- Rename isLoading in mutations to isPending

Fixes #4855


docs/react/guides/dependent-queries.md
```diff
@@ -34,14 +34,14 @@ const {
 The `projects` query will start in:
 
 ```tsx
-status: 'loading'
+status: 'pending'
```
TkDodo, line 37:
```suggestion
status: 'pending'
isLoading: false
```

docs/react/guides/dependent-queries.md
```diff
@@ -34,14 +34,14 @@ const {
...
 fetchStatus: 'idle'
 ```
 
 As soon as the `user` is available, the `projects` query will be `enabled` and will then transition to:
 
 ```tsx
-status: 'loading'
+status: 'pending'
```
TkDodo, line 44:
```suggestion
status: 'pending'
isLoading: true
```

docs/react/guides/migrating-to-v5.md
```diff
@@ -296,4 +296,16 @@ The `Hydrate` component has been renamed to `HydrationBoundary`. The `Hydrate` c
...
+The `loading` status has been renamed to `pending`, and similarly the derived `isLoading` flag has been renamed to `isPending`.
+
+For mutations as well the `status` has been changed from `loading` to `pending` and the `isLoading` flag has been changed to `isPending`.
+
+Lastly the a new derived `isLoading` flag has been added to the queries that is implemented as `isPending && isFetching`. This means that `isLoading` and `isInitialLoading` have the same thing, but `isInitialLoading` is deprecated now and will be removed in the next major version.
+
+To understand the reasoning behing this change checkout the [v5 roadmap discussion](https://github.com/TanStack/query/discussions/4252).
+ 
+
+
```
TkDodo, line 310:
```suggestion
```

packages/query-core/src/queryObserver.ts
```diff
@@ -521,16 +521,17 @@ export class QueryObserver<
...
     const result: QueryObserverBaseResult<TData, TError> = {
       status,
       fetchStatus,
-      isLoading,
+      isPending,
       isSuccess: status === 'success',
       isError,
-      isInitialLoading: isLoading && isFetching,
+      isInitialLoading: isPending && isFetching,
+      isLoading: isPending && isFetching,
```
TkDodo, line 534: let's assign this to a `const isLoading = isPending && isFetching`, and then return `isInitialLoading: isLoading` as well as `isLoading` please

packages/query-core/src/types.ts
```diff
@@ -379,7 +379,12 @@ export interface QueryObserverBaseResult<TData = unknown, TError = Error> {
   isFetchedAfterMount: boolean
   isFetching: boolean
   isLoading: boolean
+  isPending: boolean
   isLoadingError: boolean
+  /**
+   * @deprecated isInitialLoading is being deprecated in favor of isLoading
+   * and will be removed in the next major version.
+   */
   isInitialLoading: boolean
```
TkDodo, line 388: we should add the deprecation warning to the api reference as well please

TkDodo (commented):
nice work 🚀

TkDodo:
also, I think we'd have to update the `examples` as well?

prateek3255:
Hey @TkDodo Made the requested changes and updated the examples as well, although there are some type errors in the svelte tests after merging v5 which I am not able to pin point.

vezaynk:
```
Introduce a new, derived boolean flag isLoading that is implemented as isPending && isFetching
```

This undoes the breaking change between 3->4 (iirc) which made

```ts
useQuery({ ..., enabled: false }).isLoading === true
``` 

Is the idea to revert the behavior back? 🤔

prateek3255:
> Is the idea to revert the behavior back? 🤔

Yes kind of because the query isn't fetching in this case so it doesn't exactly make sense for `isLoading` to be true. For the same reason the status `loading` has been replaced by `pending` and `isLoading` has been renamed to `isPending`. So in the above case `isPending` would be true instead of `isLoading`. And now `isLoading` will only only be true when the query is in pending state and is fetching data.

To know more in detail why this change was made checkout the first point in the v5 roadmap: https://github.com/TanStack/query/discussions/4252

docs/react/reference/useQuery.md
```diff
@@ -226,11 +228,13 @@ const {
...
-  - Is the same as `isFetching && !isLoading`
-- `isInitialLoading: boolean`
+  - Is `true` whenever a background refetch is in-flight, which _does not_ include initial `pending`
+  - Is the same as `isFetching && !isPending`
+- `isLoading: boolean`
   - Is `true` whenever the first fetch for a query is in-flight
-  - Is the same as `isFetching && isLoading`
+  - Is the same as `isFetching && isPending`
+- `isInitialLoading: boolean`
+  - An alias for `isLoading`, will be deprecated in the next major version.
```
TkDodo, line 237:
```suggestion
- `isInitialLoading: boolean`
  - **deprecated**
  - An alias for `isLoading`, will be removed in the next major version.
```

TkDodo (approved):

PR #4898, opened 2023-01-30
fix(core): make sure queries that never fetched can be garbage collected

### LINKED ISSUE #4887: Queries are persisted with cacheTime of 0
### Describe the bug

`cacheTime` and `queryKey` don't seem to behave how i would expect when using a persister.

https://codesandbox.io/s/persister-cachetime-55qnu4?file=/src/index.jsx:1108-1561

### Steps to reproduce

1. Open codesandbox link
2. Read `BUG:` comments in `useQuery` hook options
3. Refresh page repeatedly using the refresh button in the browser preview
    - Note that the even though we've specified `cacheTime: 0` that the resolved value is persisted and used in the new page render
    - Note that changing `queryKey` value will persist the previous value indefinitely. This creates cache bloat - it's effectively creating a memory leak as these values will continue to be deserialized on subsequent application loads

### Expected behavior

- `cacheTime: 0` does not serialize the values into persisted storage. 
- Queries that specify `cacheTime: 0` do not use cached values in subsequent unmounts/mounts

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome 109.0.5414.119

### TanStack Query version

4.23.0

### TypeScript version

_No response_

### Additional context

_No response_

TkDodo:
This is important if data comes into the cache via a non-fetching method (like setQueryData or hydration) AND there is no active observer for them. Without this, they would never be gc'ed, but they should be after the default cacheTime

closes #4887

### DIFF
--- a/packages/svelte-query/src/__tests__/createQuery.test.ts
+++ b/packages/svelte-query/src/__tests__/createQuery.test.ts
@@ -1,11 +1,11 @@
 import { describe, it, expect } from 'vitest'
-import { render, screen } from '@testing-library/svelte'
+import { render, waitFor } from '@testing-library/svelte'
 import CreateQuery from './CreateQuery.svelte'
 import { sleep } from './utils'
 
 describe('createQuery', () => {
   it('Render and wait for success', async () => {
-    render(CreateQuery, {
+    const rendered = render(CreateQuery, {
       props: {
         options: {
           queryKey: ['test'],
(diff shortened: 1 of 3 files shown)


packages/svelte-query/src/__tests__/createQuery.test.ts
```diff
@@ -17,14 +17,12 @@ describe('createQuery', () => {
...
+    await waitFor(() => {
+      expect(rendered.getByText('Loading')).toBeInTheDocument()
+    })
 
-    expect(screen.queryByText('Success')).toBeInTheDocument()
-    expect(screen.queryByText('Loading')).not.toBeInTheDocument()
-    expect(screen.queryByText('Error')).not.toBeInTheDocument()
+    await waitFor(() => {
+      expect(rendered.getByText('Success')).toBeInTheDocument()
+    })
```
TkDodo, line 26: @lachlancollins FYI, I've seen this test fail randomly on CI. It's usually better to write tests that do not rely on timing (`await sleep(20)`). I've re-written the test to match how we'd like all other tests to look as well. I'm aware many of the tests in react-query also need to catch-up 😅

lachlancollins: Ahh I see, thanks!

PR #4902, opened 2023-01-30
fix(core): make sure pausedMutations always execute in serial

### LINKED ISSUE #4896: race condition in offline -> online state transition with dependent mutations
### Describe the bug

The react-query does not respect dependencies between offline-created mutations during their synchronization. They are fired all at once, which might cause race conditions in some cases.

The problem does not exist after state hydration, where `queryClient.resumePausedMutations` function makes the mutations fire one after another.

https://codesandbox.io/s/beautiful-voice-ykc42f?file=/src/App.tsx

### Steps to reproduce

1. Open codesandbox link.
2. Disable network access using browser debugger to enter offline mode.
3. Create new post.
4. Submit it.
5. Restore network.

The draft post is created, but not submitted.

### Expected behavior

A draft post should be created and submitted.
The submission mutation should wait for the creation mutation to finish.

None

### Screenshots or Videos

_No response_

### Platform

windows 10, Brave 1.47.186 Chromium: 109.0.5414.119

### TanStack Query version

4.23.0

### TypeScript version

4.9.3

### Additional context

Created after submission of [this stackoverflow question](https://stackoverflow.com/questions/75109195/react-query-how-to-avoid-race-condition-and-ensure-synchronization-order-of-dep?noredirect=1#comment132573636_75109195).

TkDodo:
when resumePausedMutations is called multiple times, we are starting the second paused mutation even though the first one has not yet finished. This can happen when a focus event and an online event happen in short succession

closes #4896

### DIFF
--- a/packages/query-core/src/mutationCache.ts
+++ b/packages/query-core/src/mutationCache.ts
@@ -79,6 +79,7 @@ export class MutationCache extends Subscribable<MutationCacheListener> {
 
   private mutations: Mutation<any, any, any, any>[]
   private mutationId: number
+  private isResuming = false
 
   constructor(config?: MutationCacheConfig) {
     super()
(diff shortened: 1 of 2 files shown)


jligeza:
Doesn't work for me. When turning on network and focusing the tab, the offline-created mutations seem to be stuck in a permanent paused state.

TkDodo:
you are right, something is off here. I think I was too fast with merging this :/

I'll revert and re-open this PR

PR #4906, opened 2023-01-30
fix(core): make sure pausedMutations always execute in serial

TkDodo:
when resumePausedMutations is called multiple times, we are starting the second paused mutation even though the first one has not yet finished. This can happen when a focus event and an online event happen in short succession

closes https://github.com/TanStack/query/issues/4896


TkDodo:
@jligeza I've reverted, and here is the same PR again in draft version. I won't get around to taking another look soon, but maybe you want to look into or know why this happened 🤔

jligeza:
Works for me.

PR #4922, opened 2023-02-02
feat(types): cheaper `WithRequired`

Andarist:
To the best of my understanding - this should work just fine but I didn't test it "thoroughly" with TanStack Query.

I'm opening this PR mostly because I got curious if it will pass your CI jobs. If it does though - it might be worth merging. This reduces 3 mapped types to just one :)


DamianOsipiuk:
@Andarist Any idea how much faster this one performs?
Or maybe do you have an idea how to measure it?

Seems to be working as intended, but looks weird 😂

Andarist:
Hard to tell really. It's "theoretically" faster by the sole virtue of using fewer mapped types. The best way to test this would be to run `tsc --extendedDiagnostics` with both versions in a project that is using this library a lot.

DamianOsipiuk:
So i ran a couple of runs on `vue-query` which seems to be using more of `WithRequired`  with added diagnostics, and clearing builds between runs.

It seems like it improves the time by `~0.05s` and memory usage by `~5000K`. 

Of course those tests were not very scientific 🧑‍🔬 🔫  But it seems it does make a difference.
I would be curious about performance benefit in a large app.

@TkDodo Should we merge it?

Andarist:
It also might be worth checking out the functionality of this code in pre-4.8 - since this is a version that changed some rules around `T & {}`

TkDodo:
Good point, we want to support 4.7 and onwards in v5

PR #4939, opened 2023-02-04
Fix #1484 mutation not calling onSettled if onError throws

arthurgeron:
When onError throws it aborts the current scope and dispatches another "error" state without finishing (e.g. calling onSetlled), it'd also never report the exception by calling onError with it or log it to console, which causes the mutation to freeze and always report a loading state as reported [here](https://github.com/TanStack/query/discussions/1484)

Bug can be reproduced [here](https://codesandbox.io/p/sandbox/serene-lovelace-3tch9c?file=%2Fsrc%2Findex.jsx&selection=%5B%7B%22endColumn%22%3A1%2C%22endLineNumber%22%3A11%2C%22startColumn%22%3A1%2C%22startLineNumber%22%3A11%7D%5D)


arthurgeron:
CodeSandbox that reproduces the bug:
https://codesandbox.io/p/sandbox/serene-lovelace-3tch9c?file=%2Fsrc%2Findex.jsx&selection=%5B%7B%22endColumn%22%3A1%2C%22endLineNumber%22%3A11%2C%22startColumn%22%3A1%2C%22startLineNumber%22%3A11%7D%5D

TkDodo:
Not sure about your bug reproduction, can't get it to work (`filter` is not defined). I simplified it a bit:

- `mutationFn` throws `new Error('foo')`
- `onError` throws `new Error('bar')`

we're winding up correctly in an error state due to the existing `finally` clause, and the error displayed is `'foo'`. This was a conscious decision because when you have two separate errors, we can only "store" one of them, and `onError` is a method to execute side-effects which mostly shouldn't have any errors. It is not meant to "transform" errors any further.

https://codesandbox.io/p/sandbox/icy-wave-wguwm9?file=%2Fsrc%2Findex.jsx&selection=%5B%7B%22endColumn%22%3A1%2C%22endLineNumber%22%3A11%2C%22startColumn%22%3A1%2C%22startLineNumber%22%3A11%7D%5D

arthurgeron:
I've tested the link here in incognito and it reproduced the bug correctly, the mutation fails but gets stuck in a loading state, you can see it in this video:
[0fd0cd0a-98a7-4879-a200-c49ba13dfb7b.webm](https://user-images.githubusercontent.com/3487334/218355130-2664adb5-f33c-47eb-b292-a44ce5a35aaf.webm)

> we're winding up correctly in an error state due to the existing `finally` clause, and the error displayed is `'foo'`.
Yes the error state is the ideal, but in reality it winds up in a loading state which does not change unless another mutation is triggered

TkDodo:
can you please try to make it fail in my reduced example please

arthurgeron-work:
> can you please try to make it fail in my reduced example please

I've reproduced it in your example but quite inconsistently, we could be dealing with two different bugs here: 

1- The one I reported which might only exist in react-query and not tan stack it might also only be a issue in react query and not tan stack. That might be the case because it's easily reproduced as you can see in the  [codesandbox](https://codesandbox.io/s/react-query-mutation-forked-7bcdc2?file=/src/ReactQueryExample.js) I've forked from the gallery and modified, no timeouts needed. Still it might be important to fix it for users that haven't migrated;

2 - The one I reproduced in the first sandbox and again [here](https://codesandbox.io/p/sandbox/angry-cori-1hm3te?file=%2Fsrc%2Findex.jsx&selection=%5B%7B%22endColumn%22%3A21%2C%22endLineNumber%22%3A25%2C%22startColumn%22%3A21%2C%22startLineNumber%22%3A25%7D%5D) but with no throws in onError, `filter` does not exist and trying to parse it inside a JSON on reject seems to cause it to hang in loading state, which does not happen if you defined a value to filter or remove the timeout.

TkDodo:
> but with no throws in onError, filter does not exist and trying to parse it inside a JSON on reject seems to cause it to hang in loading state

yeah I see it now - the promise never gets rejected because accessing `filter`, which doesn't exist, throws and then `promise.reject` is never executed, so the promise still hangs in loading state.

if you'd use an async function instead and added some `sleep` in there instead of creating a new Promise that never resolves or rejects (because of the Error), there wouldn't be a problem.

I'm gonna close this PR because I don't think there is an issue in the current version of Tanstack query.

PR #4945, opened 2023-02-06
feat(react-query-devtools): allow custom theme

yeion7:
This PR adds support to customize the devtools theme, for that I implemented a few things

* All components used in `ReactQueryDevtoolsPanel` read theme from context
* Added a new optional prop to `ReactQueryDevtools`, that allows to provide a `theme`, by default is the current theme
* Expose `defaultTheme` and `ThemeProvider`

### DIFF
--- a/packages/react-query-devtools/src/devtools.tsx
+++ b/packages/react-query-devtools/src/devtools.tsx
@@ -37,7 +37,7 @@ import {
   ActiveQueryPanel,
 } from './styledComponents'
 import ScreenReader from './screenreader'
-import { ThemeProvider, defaultTheme as theme } from './theme'
+import { ThemeProvider, defaultTheme, useTheme } from './theme'
 import { getQueryStatusLabel, getQueryStatusColor } from './utils'
 import Explorer from './Explorer'
 import Logo from './Logo'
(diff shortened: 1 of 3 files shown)


yeion7:
Easier to review with `hide whitespace` 

image

yeion7:
Example with a custom theme: 

https://codesandbox.io/s/tanstack-query-example-react-basic-typescript-forked-19r2fy?file=/src/index.tsx:3385-3408

TkDodo (changes requested):
Thank you, it's a good idea, but I don't think we should expose the theme as-is. We haven't given much thought into naming, and I'm not sure if everything is using the theme variables yet.

For this to work, I think we should only have semantic tokens, not color tokens like `gray` or `grayAlt`. As a user, you can't know where `gray` is used, and if you want to make a light theme, it's weird to set this to something other than a gray color...

also, we use the semantic values quite arbitrarily. For example `danger` is used for everything that "should be red", which feels wrong.  And `warning` is mostly used for `stale` queries. `paused` is semantic though, so it's kind of a mix of names...

if you want to work on improving this, please do. But if we expose a theme and make it public for consumers, a bit more thought should go into this.

Additional idea: Maybe we should expose a dark and a light theme and toggle automatically per default depending on prefersColorScheme 🤔

TkDodo:
we've since re-written the devtools in solid to be framework agnostic so I think the PR in this state is not going to make it.

PR #4949, opened 2023-02-07
prepare-codemods-package

balazsmatepetro:
FYI @TkDodo


packages/react-query/package.json
```diff
@@ -33,17 +33,11 @@
   "files": [
     "build/lib/*",
     "build/umd/*",
-    "src",
-    "codemods",
-    "!codemods/jest.config.js",
-    "!codemods/**/__testfixtures__",
-    "!codemods/**/__tests__"
```
TkDodo, line 40: one remaining question is - how can we distribute the codemods alongside the framework adapters like react? I don't think we'd want to actually release the codemods as a separate npm package 🤔 . I have to admit I kinda forgot what the idea here was 😅

TkDodo: could we just link to `../packages/codemods` here instead of removing it?

balazsmatepetro: No, we don't want to release it as a separate npm package. The idea is to link the corresponding codemods (and necessary utils) to the packages. E.g. the current codemods can be simply linked to the `react-query` package (or any other packages that use JSX) because they are fully compatible.

The `vue` and `svelte` codemods might require some additional effort because I'm not sure if `jscodeshift` is compatible with them.

Long story short: this PR is about moving the common tooling into one place, so if somebody is also interested in contributing to the codemods, this PR unblocks them.

I took the remove overloads codemod as we discussed, it's in progress, so let's sync when you're back to work. 😉

Issue #2142, opened 2021-04-17
broadcastQueryClient doesn't keep cache between tabs


DoneDeal0:
**Describe the bug**
I'm trying broadcastQueryClient feature to make the cache persist across tabs. When opening a new tab, the server endpoint is called, while the react-query should prevent this and return the cache data instead.

My usecase:
When the app mounts, I make an api call to retrieve the user's info and store them in a global state. There are sensitive informations, so local-storage is not an option. I have a search result page. Users open a new tab each time they click on a product. Each new tabs remounts the whole app, and the `/get-user-infos` route is called repeatedly. I need to cache the result across tabs.

**To Reproduce**
index.js
```js
import { QueryClient, QueryClientProvider } from "react-query";
import { broadcastQueryClient } from "react-query/broadcastQueryClient-experimental";

const queryClient = new QueryClient();
broadcastQueryClient({
  queryClient,
  broadcastChannel: "myapp",
});

ReactDOM.render(
  <QueryClientProvider client={queryClient}>
    <App />
  </QueryClientProvider>,
  document.getElementById("root")
);
```
App.js

```js
export default function App() {
  const { user, loading } = getUserInfos();
  return (
    <BrowserRouter>
      {loading ? "loading..." : <div>hello {user.name}! </div>}
      <Switch>
        <Route to="/" render={Home} />
        <Route to="/user" render={User} />
      </Switch>
    </BrowserRouter>
  );
}
```
service.js (where the api cal is made)

```js
/* eslint-disable react-hooks/rules-of-hooks */
import { useQuery } from "react-query";
import { api } from "./config";

const _getUserInfos = async () => {
  try {
    const res = api.get("/get-user-infos");
    return res;
  } catch (err) {
    return err;
  }
};

export const getUserInfos = () => {
  const { data, isLoading } = useQuery("contact", () => _getUserInfos(), {
    staleTime: 1000 * 60 * 60 * 24, // 24 hours
    cacheTime: 1000 * 60 * 60 * 24, // 24 hours
  });
  return { user: data && data.data.user, loading: isLoading };
};
```

The new tab is open like this on `<Home/>`
```js
  const onOpen = () => {
    const tab = window.open("/user");
    tab.focus();
  };
```

The cache works fine as long as I use react-router-dom on the same page.

The server:
```js
const app = require("express")();
const cors = require("cors");

app.use(cors({ credentials: true, origin: ["http://localhost:3000"] }));

[...]

PR #4950, opened 2023-02-07
feat(query-broadcast-client-experimental): enable hydrating the cache on initialization

ronaldcurtis:
Fixes issue: https://github.com/TanStack/query/issues/2142

This enables new windows/tabs to request a snapshot of the cache from other windows in order to hydrate its own cache.

It solves the problem of new windows making unnecessary network requests upon initialization.


ronaldcurtis:
Just realized that another possible workaround exists. That is, you can use [PersistQueryClientProvider](https://tanstack.com/query/v4/docs/react/plugins/persistQueryClient#persistqueryclientprovider) instead to hydrate the cache across windows on the same domain 🎉

PR #4958, opened 2023-02-09
Change the name of second parameter of onError handler to `variables`

codeisneverodd:
The second parameter of onError handler is named `newTodo`. 
Unfortunately, it's not `newTodo` but `variables` from `mutate`.
This naming can confuse readers.


TkDodo:
you can name it however you want. In this example, it's just a `string`, so `variables` (plural) might be misleading as well. We kept the same name as the thing that gets passed into `onMutate`, where it's also named `newTodo`.

PR #4974, opened 2023-02-12
Set `fetchStatus` to `idle` when dehydrating

prateek3255:
Fixes a bug which caused the `fetchStatus` to be `fetching` forever with persisters when a successful query was refetching in background and the app was quit before the fetch succeeds

### DIFF
--- a/packages/query-core/src/hydration.ts
+++ b/packages/query-core/src/hydration.ts
@@ -59,7 +59,7 @@ function dehydrateMutation(mutation: Mutation): DehydratedMutation {
 // in the html-payload, but not consume it on the initial render.
 function dehydrateQuery(query: Query): DehydratedQuery {
   return {
-    state: query.state,
+    state: { ...query.state, fetchStatus: 'idle' },
     queryKey: query.queryKey,
     queryHash: query.queryHash,
   }
--- a/packages/query-core/src/hydration.ts
+++ b/packages/query-core/src/hydration.ts
@@ -1,6 +1,7 @@
 import type { QueryClient } from './queryClient'
 import type { Query, QueryState } from './query'
 import type {
+  FetchStatus,
   MutationKey,
   MutationOptions,
   QueryKey,
@@ -59,7 +60,7 @@ function dehydrateMutation(mutation: Mutation): DehydratedMutation {
 // in the html-payload, but not consume it on the initial render.
 function dehydrateQuery(query: Query): DehydratedQuery {
   return {
-    state: { ...query.state, fetchStatus: 'idle' },
+    state: query.state,
     queryKey: query.queryKey,
     queryHash: query.queryHash,
   }
@@ -142,10 +143,17 @@ export function hydrate(
   queries.forEach((dehydratedQuery) => {
     const query = queryCache.get(dehydratedQuery.queryHash)
 
+    // Reset fetch status to idle in the dehydrated state to avoid
+    // query being stuck in fetching state upon hydration
+    const dehydratedQueryState = {
+      ...dehydratedQuery.state,
+      fetchStatus: 'idle' as FetchStatus,
+    }
+
     // Do not hydrate if an existing query exists with newer data
     if (query) {
       if (query.state.dataUpdatedAt < dehydratedQuery.state.dataUpdatedAt) {
-        query.setState(dehydratedQuery.state)
+        query.setState(dehydratedQueryState)
       }
       return
     }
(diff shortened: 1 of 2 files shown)


packages/query-core/src/tests/hydration.test.tsx
```diff
@@ -426,4 +427,18 @@ describe('dehydration and rehydration', () => {
...
+    await queryClient.prefetchQuery(['string'], () => fetchData('string', 10))
+
+    queryClient.refetchQueries(['string'])
+
+    const dehydrated = dehydrate(queryClient)
+    const stringified = JSON.stringify(dehydrated)
+
+    const parsed = JSON.parse(stringified) as DehydratedState
+    expect(parsed.queries[0]?.state.fetchStatus).toBe('idle')
+  })
```
TkDodo, line 443: I can see that we are not awaiting `refetchQueries`, but this could still result in race conditions if CI is slow because it relies on timing (10ms sleep). I think a better way would be to have the queryFn return a Promise that we only resolve after we have called dehydrate, and maybe even expect that the `dehydrated` json contains a query in loading and fetching state?

prateek3255: @TkDodo I have updated the code to ensure that we manually resolve Promise after dehydration, can you also check if it looks good? Also I am not sure how the dehydrated json would contain a query in loading and fetching state since we set the `fetchStatus` to `idle` when dehydrating.

TkDodo: oh, the change is made in `dehydrate`. For some reason, I thought we apply the `fetchStatus:'idle'` overwrite when writing back to the queryCache, so when `hydrating`. Wouldn't that be better ?

prateek3255: Yeah you are right, doing it while hydration makes more sense. Have updated the code and test can you check again?

packages/query-core/src/hydration.ts
```diff
@@ -142,10 +143,17 @@ export function hydrate(
   queries.forEach((dehydratedQuery) => {
     const query = queryCache.get(dehydratedQuery.queryHash)
 
+    // Reset fetch status to idle in the dehydrated state to avoid
+    // query being stuck in fetching state upon hydration
+    const dehydratedQueryState = {
+      ...dehydratedQuery.state,
+      fetchStatus: 'idle' as FetchStatus,
```
TkDodo, line 150:
```suggestion
      fetchStatus: 'idle' as const,
```

packages/query-core/src/hydration.ts
```diff
@@ -142,10 +143,17 @@ export function hydrate(
...
+    }
+
     // Do not hydrate if an existing query exists with newer data
     if (query) {
       if (query.state.dataUpdatedAt < dehydratedQuery.state.dataUpdatedAt) {
```
TkDodo, line 155:
```suggestion
      if (query.state.dataUpdatedAt < dehydratedQueryState.dataUpdatedAt) {
```

packages/query-core/src/tests/hydration.test.tsx
```diff
@@ -426,4 +427,46 @@ describe('dehydration and rehydration', () => {
...
+    await queryClient.prefetchQuery(['string'], () => customFetchData())
+
+    queryClient.refetchQueries(['string'])
+
+    const dehydrated = dehydrate(queryClient)
+    resolvePromise('string')
+    const stringified = JSON.stringify(dehydrated)
+
+    // ---
+    const parsed = JSON.parse(stringified) as DehydratedState
```
TkDodo, line 460: isn't this just the `dehydrated` const from line 455 ?

packages/query-core/src/hydration.ts
TkDodo: the problem here is that `'foo' as FetchStatus` would also be fine for typescript, so if we ever rename `idle` to something else, TS will not notify us about this change here. `const assertions` solve this: https://tkdodo.eu/blog/the-power-of-const-assertions

prateek3255: Yes makes sense, updated.

packages/query-core/src/tests/hydration.test.tsx
prateek3255: Yeah you are right can expect the the value from dehydrated itself don't really need the assertion here. Moved the expect fetching to the right place

TkDodo (approved):

PR #4977, opened 2023-02-12
chore: update dependencies

DamianOsipiuk:


package.json
```diff
@@ -23,71 +23,72 @@
...
     "solid-js": "^1.5.7",
     "solid-testing-library": "^0.3.0",
     "stream-to-array": "^2.3.0",
-    "ts-jest": "^27.1.1",
-    "ts-node": "^10.7.0",
-    "typescript": "4.7.4",
-    "vue": "^3.2.33"
+    "ts-jest": "^29.0.5",
+    "ts-node": "^10.9.1",
+    "typescript": "4.9.5",
```
TkDodo, line 90: One remark: I think we agreed on coding with the TS version that is the lowest that we'll also support, which is 4.7.4. This will avoid accidentally using a newer TS feature like `satisfies` that would then break for users.

TkDodo:
@DamianOsipiuk vue-query tests seem to be failing now regularly, can you have a look please?

PR #4979, opened 2023-02-13
fix return type of MutateSyncFunction

ozum:
According to [docs](https://tanstack.com/query/v4/docs/vue/reference/useMutation), `MutateSyncFunction` should return `Promise<TData>`.


TkDodo:
where does it say so?
- `mutate` returns `void`
- `mutateAsync` returns a Promise

ozum:
> where does it say so?
> 
> * `mutate` returns `void`
> * `mutateAsync` returns a Promise

@TkDodo, I checked again, and you are right. I confused them I guess. I'm closing this PR.

PR #4984, opened 2023-02-13
feat(useInfiniteQuery): new maxPages option

### LINKED ISSUE #4975: windowSize for useInfiniteQuery
as discussed on [twitter](https://twitter.com/ArnBZN/status/1624732689227030531), as a potentially different api to replace `refetchPages`.

@arnaudbzn

arnaudbzn:
closes #4975 

- [x] New `maxPages` option support for infinite queries (`infiniteQueryBehavior` logic is impacted)
- [x] Unit tests with 100% code coverage for the modified code
- [x] New basic example: `infinite-query-with-max-pages`

### Todos
- [x] Fix the issue with the `pnpm-lock.yaml`
- [x] Documentation
- [x] Add the example in the `docs/config.json`

### Comments
- if `setQueryData` is used to add some pages, these pages are added to the query data even if the `maxPages` limit is reached. As discussed in the corresponding issue, this behavior should be fine.


packages/query-core/src/tests/infiniteQueryBehavior.test.tsx
```diff
@@ -121,4 +121,138 @@ describe('InfiniteQueryBehavior', () => {
...
+      data: { pages: [0, 1] },
+    })
+
+    queryFnSpy.mockClear()
+
+    // Refetch the infinite query
+    await observer.refetch()
+
+    // Only 2 pages should be refetched
+    expect(queryFnSpy).toHaveBeenNthCalledWith(1, {
```
TkDodo, line 242: does this really check if we have called it only 2 times? If I understand `toHaveBeenNthCalledWith`, it only checks if the 1st call has been called with the certain params and the 2nd call as well. But there could be a 3d call as well and this test would pass.

I think we should add:

`expect(queryFnSpy).toHaveBeenCalledTimes(2)`

packages/query-core/src/infiniteQueryBehavior.ts
```diff
@@ -53,10 +53,23 @@ export function infiniteQueryBehavior<
...
-          newPageParams = previous
-            ? [param, ...newPageParams]
-            : [...newPageParams, param]
-          return previous ? [page, ...pages] : [...pages, page]
+          const { maxPages } = context.options
+
+          const limitedPages = new LimitedLengthArray(maxPages, [...pages])
+          const limitedNewPageParams = new LimitedLengthArray(
+            maxPages,
+            newPageParams,
```
TkDodo, line 61: we're not shallow copying `newPageParams` here 🤔

packages/query-core/src/utils.ts
```diff
@@ -330,3 +330,31 @@ export function keepPreviousData<T>(
 ): T | undefined {
   return previousData
 }
+
+export class LimitedLengthArray<T> {
+  private items: T[]
+  private length: number | undefined
```
TkDodo, line 336: in v5, we're using private class fields please:

```
#items: T[],
#length: number | undefined
```

But I'm wondering if a class is the right abstraction / worth the overhead. Wouldn't two immutable util functions achieve the same goal? Something like:

```
const addToEnd = (items, item, max = 0) => {
  const new = [...items, item]
  return max > new.length ? new.slice(1) : new
}

const addToStart = (items, item, max = 0) => {
  const new = [item, ...items]
  return max > new.length ? new.slice(0, -1) : new
}
```

TkDodo (commented):
love it ❤️

packages/query-core/src/tests/infiniteQueryBehavior.test.tsx
arnaudbzn: Good catch! I'll do.

packages/query-core/src/infiniteQueryBehavior.ts
arnaudbzn: I've tried to stay consistent with the existing code where `newPageParams` value is directly updated.

packages/query-core/src/tests/infiniteQueryBehavior.test.tsx
arnaudbzn: Done in 156751db8643d3cb477e9050df48b3d664f44012

packages/query-core/src/utils.ts
arnaudbzn: Ok let me refactor that with util functions.

packages/query-core/src/infiniteQueryBehavior.ts
arnaudbzn: Refactored with the util functions in 434af38abd73ebc4a38a72ca87fbca8f359dbb7a

packages/query-core/src/utils.ts
arnaudbzn: Done in 434af38abd73ebc4a38a72ca87fbca8f359dbb7a

TkDodo (approved):
nice one 🚀

TkDodo:
let's finish up on docs please, the example is also good the way it is imo, and then let's ship it.

Btw, I've seen that this could also go into v4 actually, because it's non-breaking. Should we do that or stick with v5? I mean, v5 needs "new cool features", too 😅

arnaudbzn:
Ok for the docs and example, remaining todos should be completed later today.

About v4 or v5: 
As a v4 user, I'll love to have this feature now!
But it makes sense to ship it with the v5 to move forward the community to the latest version 🏄🏻‍♂️

We could call this feature "Eternal Query"⚡️: infinite query with high scalability (without degrading the client memory/perf when scrolling a high number of pages or refetching the query).

TkDodo:
> We could call this feature "Eternal Query"⚡️: infinite query with high scalability

wow, you're a marketing genius 🚀

arnaudbzn:
Docs updated:
- reference: useInfiniteQuery page
- guide: Infinite Queries page: new example

TkDodo:
now I remember why we'd want this feature v5: my plan was to remove the `refetchPages` option everywhere. The main case for having this option was: If I have so many pages in my cache I don't want to refetch them all. But not refetching all pages might lead to UI inconsistencies. Also, this option is available on e.g. `queryClient.refetchQueries`, but it only does something for infinite queries, not "normal" queries. That's a weird api imo.

So, could you remove that in this PR as well please?

arnaudbzn:
I can for sure, it makes sense.
But it's maybe cleaner in terms of git/gh history to do that on another PR after the merge of this one?

TkDodo (approved):

TkDodo:
I'll merge this, but for a follow-up: an entry in the migration guide where we show off `new features` might also be interesting :)

arnaudbzn:
Great, I'll create a new PR for that.

PR #4990, opened 2023-02-16
feat: remove refetchPage

arnaudbzn:
Breaking change related to PR #4984 
`refetchPage` introduced is removed in favor of the new`maxPages` option

- [x] Remove `refetchPage` related code
- [x] Remove related tests
- [x] Update docs
- [x] Update migration guide breaking changes section
- [x] Update migration guide new featureq section (maxPages)
- [x] Fix types


packages/query-core/src/types.ts
```diff
@@ -338,31 +338,17 @@ export interface ResultOptions {
...
 
-export interface RefetchQueryFilters<TPageData = unknown>
-  extends QueryFilters,
-    RefetchPageFilters<TPageData> {}
+export interface RefetchQueryFilters extends QueryFilters {}
 
-export interface ResetQueryFilters<TPageData = unknown>
-  extends QueryFilters,
-    RefetchPageFilters<TPageData> {}
+export interface ResetQueryFilters extends QueryFilters {}
```
TkDodo, line 351: do we need those extra interfaces if they are now exactly the same as `QueryFilters` ? I think I only made them for that feature ...

TkDodo (approved):

packages/query-core/src/types.ts
arnaudbzn: Sure let me remove this one.

arnaudbzn: Done in #b338f7d5f163edc313ab9b44c682ab2aa01f3440
Still some type checks failed in vue tests..

arnaudbzn: All checks have passed.

arnaudbzn:
@TkDodo Migration guide updated with a "New Features" section.

TkDodo (approved):

cah4a:
Thank you for your great work. 
What if I still want to refetch just one page?

TkDodo:
> What if I still want to refetch just one page?

there's no good way to make that work

PR #4995, opened 2023-02-17
feat(svelte-query): use store for reactivity in options

### LINKED ISSUE #4851: [svelte] keepPreviousData not working
### Describe the bug

The `keepPreviousData` options does not seem to work in the svelte-query. Even when set to true the data is empty until the new query is done.

I made the query reactive using the documentation here: https://tanstack.com/query/v4/docs/svelte/reactivity

I think it is because the current syntax will rerun the `createQuery` method  and create a new query object each time any used variable changes, instead of subscribing to change within the already created query object, but I do not know enough about the inner workings of query-core and svelte-query to know if this is indeed an issue.

https://codesandbox.io/s/nifty-cloud-pxwjho?file=/App.svelte

### Steps to reproduce

1. Create a reactive query with a queryKey with a variable and keepPreviousData set to true
2. Render the results to the DOM
3. Update the variable
4. All results are removed immediately removed from the DOM
5. And rendered again when the new query result is available

### Expected behavior

As a user, I expected the previous data be visible in the DOM during fetching of the updated data

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: Windows
- Browsers: Chrome 109.0, Firefox 109.0

### TanStack Query version

4.22.3

### TypeScript version

4.9.4

### Additional context

_No response_

chrislcs:
To fix #4851

This adds the option to use a writable store for the query options, so that svelte query can react appropriately when these options change. The current method of reactivity recreates the entire query each time an option is changed, which causes issues.


chrislcs:
@lachlancollins With a bit of a delay, here is the PR for using writable stores for the query options

bevm0:
What if the query options are always converted to a store, and then the query provides a method to update those options and trigger the refetch?

e.g. using the revised documentation
```ts
/** regular object with the query options */
 const queryOptions = {
    queryKey: ['refetch'],
    queryFn: async () => await fetch(endpoint).then((r) => r.json()),
    refetchInterval: 1000,
  }

  const query = createQuery(queryOptions)

  function updateRefetchInterval(event) {
    /** internally invokes the "update" method for Svelte stores and then reacts appropriately */
    $query.updateOptions({
        refetchInterval: event.target.valueAsNumber
    })
  }
```

Extra context: 
I wrote a private Svelte wrapper around @tanstack/react-query and I don't have access to the original query options once I've created all of the hooks, but I'd like to trigger the reactivity for the query without having to reconstruct it.

lachlancollins (commented):
@chrislcs I couldn't have picked a worse time to review this - everything looks good, the PR was closed because the target branch changed. Could you please change this to target `alpha`?

TkDodo:
sorry about that 🙈

chrislcs:
> What if the query options are always converted to a store, and then the query provides a method to update those options and react accordingly, e.g. by calling the new queryFn?
> 
> example using the revised documentation
> 
> ```ts
> /** regular object with the query options */
>  const queryOptions = {
>     queryKey: ['refetch'],
>     queryFn: async () => await fetch(endpoint).then((r) => r.json()),
>     refetchInterval: 1000,
>   }
> 
>   const query = createQuery(queryOptions)
> 
>   function updateRefetchInterval(event) {
>     /** internally invokes the "update" method for Svelte stores and then reacts appropriately */
>     $query.updateOptions({
>         refetchInterval: event.target.valueAsNumber
>     })
>   }
> ```
> 
> Extra context: I wrote a private Svelte wrapper around @tanstack/react-query and I don't have access to the original query options once I've created all of the hooks, but I'd like to trigger the reactivity for the query by updating its query params

@bevm0 This could be another addition, separate from this PR. Internally the options already get converted into a store, so to add this functionality you'd need to add an `updateOptions` function to the return query object. Something like this (as part of the `createBaseQuery` function):

```ts
  const { subscribe } = derived(result, ($result) => {
    $result = observer.getOptimisticResult(get(defaultedOptionsStore))
    return !get(defaultedOptionsStore).notifyOnChangeProps
      ? observer.trackResult($result)
      : $result
  })

  function updateOptions(
    updatedOptions: Partial<
      CreateBaseQueryOptions<TQueryFnData, TError, TData, TQueryData, TQueryKey>
    >,
  ) {
    optionsStore.update((currentOptions) => ({
      ...currentOptions,
      ...updatedOptions,
    }))
  }

  return { subscribe, updateOptions }
```

Up to the svelte-query maintainers if this is desired functionality.

PR #5004, opened 2023-02-19
Feature/infinite queries

TkDodo:


packages/vue-query/src/useInfiniteQuery.ts
```diff
@@ -72,13 +76,21 @@ export function useInfiniteQuery<
...
+    TQueryFnData,
+    TQueryKey,
+    TPageParam
+  >,
   queryClient?: QueryClient,
 ): UseInfiniteQueryReturnType<TData, TError> {
   const result = useBaseQuery(
     InfiniteQueryObserver as typeof QueryObserver,
-    options,
+    options as any,
```
TkDodo, line 93: @DamianOsipiuk I had to stick an `as any` here but I don't really understand why. Note that the previous types had an issue I think:

`UseInfiniteQueryOptions<TQueryFnData, TError, TData, TQueryKey>` wasn't really right as it was missing a generic, so it was passing `TQueryKey` in the "wrong spot". I only noticed wen adding `TPageParam`. Could you have a look please?

lachlancollins (approved):

packages/query-core/src/types.ts
```diff
@@ -124,6 +122,26 @@ export interface QueryOptions<
   maxPages?: number
 }
 
+export interface DefaultPageParam<TPageParam = unknown> {
+  defaultPageParam: TPageParam
+}
+
+export interface InfiniteQueryOptions<
```
DamianOsipiuk, line 129: I feel like this name does not convey the purpose of the interface.
Maybe something along `InfiniteQueryPageParamsOptions` would be better?
Since current name could be mistaken for a full set of options, which is not.

packages/vue-query/src/useInfiniteQuery.ts
DamianOsipiuk: @TkDodo Fixed

DamianOsipiuk (commented):
Hmmm, do we still need `pageParams` in return of `useInfiniteQuery`? 🤔 

For what it can be usefull?

packages/vue-query/src/useInfiniteQuery.ts
TkDodo: 🙏

packages/query-core/src/types.ts
TkDodo: agree. renamed here: 736edbd9

TkDodo:
@DamianOsipiuk 

> Hmmm, do we still need pageParams in return of useInfiniteQuery? 🤔

> For what it can be usefull?

I outlined the problems in the [initial RFC](https://github.com/TanStack/query/discussions/4854). We still need the `pageParams` to trigger refetches in some cases. As an example, suppose we have the following query:

```
useInfiniteQuery({
  queryKey,
  queryFn: ({ pageParam }) => Promise.resolve({ num: pageParam }),
  defaultPageParam: 0
  getNextPage: (lastPage) => lastPage.num + 1,
  getNextPage: (firstPage) => firstPage.num - 1,
})
```

this will start out with:

```
{
  pages: [{ num: 0 }],
  pageParam: [0]
}
```

now let's fetch some more in the forward direction:

```
{
  pages: [{ num: 0 }, { num: 1 }, { num: 2 }],
  pageParam: [0, 1, 2]
}
```

so far, so good. If a refetch happens now, it always starts at the beginning and then refetches all pages in order. so we:

- take `0` from the `pageParams` to initiate the refetches
- for every subsequent page, we just call `getNextPageParam` again to re-compute it on the fly.

Now it seems, we could just re-use the `defaultPageParam` and start with that, but we cannot, for two reasons:

1) what if we have fetched in the previous direction before, and our state is:

```
{
  pages: [{ num: -1 }, { num: 0 }, { num: 1 }, { num: 2 }],
  pageParam: [-1, 0, 1, 2]
}
```

we now have to start the fetch chain with `-1` and not with `0`. We only have that info in `pageParams`.

2) in the same way, if you use the new `pageSize` option and set it to 2, you might be in a situation where you no longer have the "initial" page in the cache, we might only have:

```
{
  pages: [{ num: 1 }, { num: 2 }],
  pageParam: [1, 2]
}
```

now, we would have to start with `1`.

Grabbing this info from the `pageParam` works well, but yeah, it's the only reason we have this extra nesting structure :/ It happens right here and is the only place where we need `pageParams`:

https://github.com/TanStack/query/blob/736edbd90297eae277d7dafcad8342686da7c030/packages/query-core/src/infiniteQueryBehavior.ts#L121

Alternatively, we would have to store the information "what's the pageParam of the current first page" somewhere on the `query`. Would that be better?
I also thought about making `defaultPageParam` a function that is executed with `allPages | undefined`, but is there a guarantee that users can always compute that value just from looking at the pages ?

DamianOsipiuk:
@TkDodo But it seems that it's now more like an implementation detail, rather than something users could use.

If it's not usefull for users it would be good to hide this somewhere. And what we really need is something like `firstPageParam` internally.
Removing `pageParams` would make using `select` and cache manipulation easier.

But question is, how would we store this firstPageParam with persisters? Or do we even need to do that?

TkDodo:
> But it seems that it's now more like an implementation detail, rather than something users could use.

yes, I agree.

> Removing pageParams would make using select and cache manipulation easier.

cache manipulation, yes. select, kinda. It should be possible to transform to _anything_ via select. it works at runtime, but not on type level. I've added some failing type level tests. I think we should fix them first before changing the structure.

> But question is, how would we store this firstPageParam with persisters? Or do we even need to do that?

yes, we would need to store it. it could be part of the `query.state`, as we store more than just the `data` anyways:

https://github.com/TanStack/query/blob/765f6196e900c1dd826b040773fcf97839ca60d0/packages/query-core/src/query.ts#L37-L50

I will need to look into it. We don't have an `infinite` version of that query. Maybe storing it on `fetchMeta` works, idk 😅

TkDodo:
hmm, it's _mostly_ an implementation detail, but when working directly with the cache, you'd still need to set it. As an example, assume you have 3 pages:

```
{
  pages: [{}, {}, {}]
  pageParams: [5, 10, 15]
}
```

if we take away `pageParams` from `data` and store them somewhere else (on queryState), what happens if you call `setQueryData` and remove the first page? You will also need to adapt `pageParams` so that it only contains `[10, 15]`. We can only do that _because_ `pageParams` are part of our Infinite Data structure. So it seems that exactly the use-cases where pageParams becomes annoying are actually the ones that might need it ...

TkDodo:
@ardeora I'll merge this PR into alpha because it contains some good improvements on the API. If we find a way to get rid of the `pageParams` as part of `data`, we can still do this separately.

PR #5014, opened 2023-02-21
docs:eslint:plugin-name-typo-fix

IamSAL:
[Usage guide](https://tanstack.com/query/v4/docs/react/eslint/eslint-plugin-query)

Guide about modifying .eslintrc is updated to include valid eslint plugin name.


TkDodo:
see: https://eslint.org/docs/latest/use/configure/plugins#configure-plugins

To configure plugins inside of a configuration file, use the plugins key, which contains a list of plugin names. The eslint-plugin- prefix can be omitted from the plugin name.

```
{
    "plugins": [
        "plugin1",
        "eslint-plugin-plugin2"
    ]
}
```

PR #5027, opened 2023-02-23
feat(vue-query): useQuery support ref args

aa900031:
useMutation is support ref args already, but useQuery is not. so I set useQuery every arguments with MaybeRef just like useMutation


aa900031:
Hi @DamianOsipiuk! Please check this PR, Welcome to discussion. Thank you.

packages/vue-query/src/useBaseQuery.ts
```diff
@@ -51,12 +51,16 @@ export function useBaseQuery<
...
   arg1:
-    | TQueryKey
-    | UseQueryOptionsGeneric<TQueryFnData, TError, TData, TQueryKey>,
+    | MaybeRef<TQueryKey>
+    | MaybeRef<UseQueryOptionsGeneric<TQueryFnData, TError, TData, TQueryKey>>,
   arg2:
-    | QueryFunction<TQueryFnData, UnwrapRef<TQueryKey>>
-    | UseQueryOptionsGeneric<TQueryFnData, TError, TData, TQueryKey> = {},
-  arg3: UseQueryOptionsGeneric<TQueryFnData, TError, TData, TQueryKey> = {},
+    | MaybeRef<QueryFunction<TQueryFnData, UnwrapRef<TQueryKey>>>
```
Mini-ghost, line 57: I'm not sure if setting the type of `queryFn` as `MaybeRef` is a wise choice. One reason is that it might cause more conflicts when users upgrade to `v5`.

Do you have any thoughts on this?

aa900031: I agree that is not wise choice too, but currently `mutationFn` type is already as ｀`MaybeRef`

aa900031: I check `parseQueryArgs` again,  I've realized that it doesn't support `queryFn` as a `MaybeRef`. I'll have to remove it in this case.

Mini-ghost: https://github.com/TanStack/query/blob/857ecd91b6d86306bacd3785e3e2f71d7287ad06/packages/vue-query/src/useBaseQuery.ts#L188-L190

Currently, it should be possible to support `queryFn` as a `ref`. Regarding the transition to `v5`, I'm not sure if I'm overthinking it.

aa900031: Yes you're right !

However, I've also noticed another thing, it ensures that `queryFn` is a function in here:

https://github.com/TanStack/query/blob/857ecd91b6d86306bacd3785e3e2f71d7287ad06/packages/vue-query/src/types.ts#L43-L50

Since it's being handled here, should follow suit as well?

Mini-ghost: Oops! ! you are right ! I missed this part.

aa900031: I just noticed that the [test](https://cloud.nx.app/runs/OilmEaAu0p/task/%40tanstack%2Fvue-query%3Abuild%3Atypes) aren't passing. I found that in `src/__tests__/useQuery.test.ts` at line 262, it's passing a ref to `queryFn`. It looks like we should reintroduce `MaybeRef` in this case.

Mini-ghost: If that's the case, the types here might not be quite accurate, right? 🤯

https://github.com/TanStack/query/blob/857ecd91b6d86306bacd3785e3e2f71d7287ad06/packages/vue-query/src/types.ts#L43-L50

It would be nice to have feedback from @DamianOsipiuk  to know which is expect.

DamianOsipiuk: Although in theory `queryFn` is an option like any other, so it should be properly handled when it's changed to another one, I would strongly discourage it.
So if types currently do not allow it, let's leave it this way.

aa900031: @DamianOsipiuk Thank you for comment, we don't change any queryFn type currently.

aa900031: @DamianOsipiuk Please merge this PR, this is important for use queryOptions as component props ex:

```ts
const Comp = defineComponent({
  props: {
    queryOptions: Object,
  },
  setup: (props) => {
    return useQuery(
      computed(() => ({
        ...props.queryOptions,
        queryFn: () => {
           // ...
        }
      }))
    )
  }
})
```

PR #5038, opened 2023-02-25
fix(react-query-devtools): do not hide sort options when filter is used

### LINKED ISSUE #4989: Devtools (React) "Filter" field, when used, makes "Sort" dropdown, "Asc / Desc" and "offline" buttons disappear
### Describe the bug

The "Filter" field of Devtools makes the other fields (Sort, asc/desc, offline) disappear when used.
Even though the are used to sort the filtered results, one can only change the sorting and the rest of the options when the "filter" field is cleared.

Attached video below

### Steps to reproduce

1. Open Tanstack (React) Query Devtools
2. Add a filter on the "filter" field
3. The rest of the fields disappear

### Expected behavior

As a user, I expect the fields to not disappear when populating some of them with data.

Every time

### Screenshots or Videos

https://user-images.githubusercontent.com/2103443/219364287-b6d127ee-8793-4261-a4cb-71ec48957281.mov

### Platform

Mac OS, Chrome 110

### TanStack Query version

v4.24.4

### TypeScript version

v4.9.5

### Additional context

_No response_

emzoumpo:
Sort options (asc / desc, and sort function) were hidden when filter was set.
Sort options are now shown along with the filter, even when a filter is set.

Closes #4989

### DIFF
--- a/packages/react-query-devtools/src/devtools.tsx
+++ b/packages/react-query-devtools/src/devtools.tsx
@@ -621,106 +621,94 @@ export const ReactQueryDevtoolsPanel = React.forwardRef<
                       width: '100%',
                     }}
                   />
-                  {!filter ? (
-                    <>
-                      <Select
-                        aria-label="Sort queries"
-                        value={sort}
-                        onChange={(e) => setSort(e.target.value)}
-                        style={{
-                          flex: '1',
-                          minWidth: 75,
-                          marginRight: '.5em',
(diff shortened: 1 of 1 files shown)


emzoumpo:
@TkDodo the failed action doesn't seem to be the result of my changes, locally the tests are passing.
test-fail-svelte

TkDodo (approved):

PR #5046, opened 2023-02-27
remove overloads codemod

balazsmatepetro:


TkDodo:
can you add documentation to the upgrade guide please:

https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5

TkDodo:
not sure if this is possible but it would be much easier to look at the test cases if we had input and output co-located for each situation. Something like:

```
useIsMutating(['foo', 'bar'])
```

will become

```
useIsMutating({
  mutationKey: ['foo', 'bar']
})
```

right now, I have to know that line 18 in the input file corresponds to lines 36-38 in the output file ...

balazsmatepetro:
> can you add documentation to the upgrade guide please:
> 
> https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5

Of course, that will be added in a separate commit.

balazsmatepetro:
> not sure if this is possible but it would be much easier to look at the test cases if we had input and output co-located for each situation. Something like:

Unfortunately not, at least I haven't found a way yet. 😞 I will try to come up with something nicer but probably not in this PR.

TkDodo (approved):

PR #5050, opened 2023-02-27
feat(svelte-query): use store for reactivity in options

### LINKED ISSUE #4851: [svelte] keepPreviousData not working

chrislcs:
New version of the accidentally closed PR: https://github.com/TanStack/query/pull/4995

---

To fix #4851

This adds the option to use a writable store for the query options, so that svelte query can react appropriately when these options change. The current method of reactivity recreates the entire query each time an option is changed, which causes issues.


lachlancollins (approved):
LGTM - please fix the merge conflict then it'll be right to go!

chrislcs:
Conflicts should be solved

PR #5051, opened 2023-02-28
feat: typed meta option for queries and mutations

TkDodo:
adds queryMeta and mutationMeta to the `Register` interface


TkDodo:
not sure if this is possible but it would be much easier to look at the test cases if we had input and output co-located for each situation. Something like:

```
useIsMutating(['foo', 'bar'])
```

will become

```
useIsMutating({
  mutationKey: ['foo', 'bar']
})
```

right now, I have to know that line 18 in the input file corresponds to lines 36-38 in the output file ...

PR #5057, opened 2023-03-01
chore: turn on no-unnecessary-type-assertion

TkDodo:


packages/solid-query/src/createInfiniteQuery.ts
```diff
@@ -31,7 +31,7 @@ export function createInfiniteQuery<
 ): CreateInfiniteQueryResult<TData, TError> {
   return createBaseQuery(
     createMemo(() => options()),
-    InfiniteQueryObserver as typeof QueryObserver,
+    InfiniteQueryObserver,
```
TkDodo, line 34: @ardeora eslint now reported this type assertion as unnecessary, but if I remove it, I'm getting a type error. can you maybe have a look?

TkDodo: sorry, nevermind 😅 . There's the same issue in react-query. I think it's a false positive of the lint rule

PR #5072, opened 2023-03-03
fix(queryClient): do not fetch while promise finished

nnnnoel:


TkDodo:
Can you provide some context what this is about please? Ideally, a test case that fails on main but works on this PR. Thanks

nnnnoel:
@TkDodo when initialize same multiple queries, queryClient fetches multiple time for same queries.
```tsx
const Component1 = () => {
  const query1 = useQuery(['fetch1'], () => fetch());

  return <div />;
}

const Component2 = () => {
  const query1 = useQuery(['fetch1'], () => fetch());

  return <div />;
}

const ParentComponent = () => {
  return (
    <>
      <Component1 />
      <Component2 />
    </>
  );
}
```

I thought @tanstack/query can be nice project for data caching, but cannot get the impact because of these issue.
same query must be fetch only one time on mount at the same time.

nnnnoel:
I can add test cases for this PR. thanks for idea!

TkDodo:
react-query should absolutely deduplicate in the above scenario. Again, please show a test case or a codesandbox reproduction that shows the behavior that you think is wrong.

nnnnoel:
@TkDodo OMG That's my fault.
in my project, queries fetch twice but it's script evaluation time was too long.
so it goes beyond first query's staleTime.

It just happening, I'm so sorry to your time spent.

Thank you!

Issue #5069, opened 2023-03-03
eslint plugin exhaustive deps objects false positive


DianomiJH:
### Describe the bug

I recently installed the plugin and started using the recommended rules but it flagged a bunch of false positives. 

When using an object of data within useQuery, we sometimes pass the whole object to the queryKey, but the eslint plugin says some items are missing from the queryKey.

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/vibrant-voice-fktju6?welcome=true&file=%2FREADME.md

### Steps to reproduce

open the terminal and run `npm run lint` and see errors in the terminal. 

### Expected behavior

I'd expect this to work the same as a useEffect dependency array, allowing passing the object to the array and removing warnings, But I'm seeing it asking for additional items in the array.

### How often does this bug happen?

Every time

### Screenshots or Videos

![image](https://user-images.githubusercontent.com/84312830/222692268-9f9fdc6b-df2a-4803-80a4-c2f68433091f.png)

### Platform

macOS
in vscode (and codesandbox)

### TanStack Query version

v4.22.0

### TypeScript version

^4.9.4

edit:
I also notice, if I do add state.id to the query key, the warning goes away, but I'm still using 2 other items from state in the query function and am not getting any errors anymore! (maybe this is more problematic than the original issue??)

TkDodo:
yeah I think it's two bugs actually:

1) if you use `state.id` and `state.name` in the queryFn, but you only have `state.id` in the queryKey, it doesn't error even though it should.

2) if you use `state.id` and `state.name` in the queryFn, and you have the complete `state` in the queryKey, it errors even though it shouldn't.

@Newbie012 FYI. I can try to write the failing test cases later today when I have some more time.

DianomiJH:
I think there might be another issue here too, if you use optional chaining in the queryKey but don't use it in the queryFn it complains at you (or maybe it was the other way around, I'm on mobile right now and can't check)

TkDodo:
here is a failing test case for situation 1 I described (should error, but doesn't):

```
{
  name: 'should fail if queryFn is using multiple object props when only one of them is in the queryKey',
  code: normalizeIndent`
    const state = { foo: 'foo', bar: 'bar' }

    useQuery({
        queryKey: ['state', state.foo],
        queryFn: () => Promise.resolve({ foo: state.foo, bar: state.bar })
    })
  `,
  errors: [
    {
      messageId: 'missingDeps',
      data: { deps: 'state.bar' },
[...]

PR #5079, opened 2023-03-05
fix(eslint-plugin): improve object property checks

Newbie012:
fixes https://github.com/TanStack/query/issues/5069

### DIFF
--- a/packages/eslint-plugin-query/src/rules/exhaustive-deps/exhaustive-deps.test.ts
+++ b/packages/eslint-plugin-query/src/rules/exhaustive-deps/exhaustive-deps.test.ts
@@ -87,7 +87,7 @@ ruleTester.run('exhaustive-deps', rule, {
         type Result = {};
         function MyComponent(props) {
             useQuery({
-              queryKey: ["foo", dep1],
+              queryKey: ["foo", dep],
               queryFn: () => api.get<Result>(dep),
             });
         }
@@ -234,6 +234,17 @@ ruleTester.run('exhaustive-deps', rule, {
             queryFn: async () => id,
           })
         }
+      `,
+    },
+    {
+      name: 'should not fail if queryKey is having the whole object while queryFn uses some props of it',
+      code: normalizeIndent`
+        const state = { foo: 'foo', bar: 'bar' }
+    
+        useQuery({
+            queryKey: ['state', state],
+            queryFn: () => Promise.resolve({ foo: state.foo, bar: state.bar })
+        })
       `,
     },
   ],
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin-query/src/rules/exhaustive-deps/exhaustive-deps.test.ts
```diff
@@ -87,7 +87,7 @@ ruleTester.run('exhaustive-deps', rule, {
         type Result = {};
         function MyComponent(props) {
             useQuery({
-              queryKey: ["foo", dep1],
+              queryKey: ["foo", dep],
```
Newbie012, line 90: Not sure how I missed it.

packages/eslint-plugin-query/src/rules/exhaustive-deps/exhaustive-deps.test.ts
```diff
@@ -236,6 +236,17 @@ ruleTester.run('exhaustive-deps', rule, {
...
     },
+    {
+      name: 'should not fail if queryKey is having the whole object while queryFn uses some props of it',
+      code: normalizeIndent`
+        const state = { foo: 'foo', bar: 'bar' }
+    
+        useQuery({
+            queryKey: ['state', state],
+            queryFn: () => Promise.resolve({ foo: state.foo, bar: state.bar })
+        })
```
TkDodo, line 247: will this fail if state has more options than what is used in the `queryFn`, like:

```
const state = { foo: 'foo', bar: 'bar', baz: 'baz' }
```

as stated in one of my comments, this would mean duplicate cache entries if we only use `foo` and `bar` in the `queryFn`, but `baz` changes values. It would be cool if those cases error as well, like:

```
const useFoo = (id) => useQuery({ queryKey: ['foo', id], queryFn: () => fetchFoo() })
```

also here, `id` is unused and probably _shouldn't_ be in the key. It's a different problem that "exhaustiveness", because it goes in the other direction. Probably nothing that blocks this PR, but I'd be curious if this is something that we already cover, or maybe should cover ...

Newbie012: Currently, the plugin doesn't error on "unusedDeps". I wrote a small sketch that should mitigate most of the cases, but it currently doesn't support your example (but it could be).

Although, what if `state` can be one nullable/optional? or what if it's a union of multiple types?

TkDodo: yeah it gets difficult for sure. gonna merge this one 👍

Issue #5074, opened 2023-03-04
jest -> vitest


TkDodo:

crutchcorn:
Per our conversation here:

https://twitter.com/TkDodo/status/1631952196974592000?t=nqJfxD8QT7aOnDw8MZzCYQ&s=19

(Mostly linking to preserve for my memory's sake) 

I'd like to see what progress I can make on this issue

snnyvrz:
I'm just a comment away? I want to know if I can help

ydubinskyi:
Would like to contribute to this

PR #5081, opened 2023-03-06
WIP: Jest to vitest migration

ZackDeRose:
Addresses #5074

## Local Benchmarks

+ on this branch (w/ vitest): `pnpm run test:lib  76.13s user 14.84s system 428% cpu 21.220 total` => 21.22s
+ in main branch (w/ jest): `pnpm run test:lib  147.31s user 20.49s system 512% cpu 32.763 total` => 32.763s
+ (on this branch with parallel option: `pnpm run --parallel test:lib  80.45s user 16.30s system 613% cpu 15.769 total` => 15.769s)

Calculated via:

```terminal
git clean -dfx && pnpm install && pnpm build && time pnpm run test:lib
```

## Things left:
+ [ ] Update all packages to set `test.globals: true` in the `vitest.config.ts` files (I switched to this method for the later packages, I should have used this approach all the way through)
+ [ ] Remove remaining traces of jest
+ [ ] Added a skip on the `react-query > ssr-hydration` suite - I'll need to understand what's going on here a bit better to restore this test to the original behavior (@TkDodo / @crutchcorn 👀)
+ [ ] The `vue-query` package needs some review. I'm not familiar with the `__mocks__` directories and what they're supposed to be doing. I added my own mock implementation in `packages/vue-query/test-setup.ts` that will mock the `vue-demi` package using vitest's `vi.mock()` method.

Please let me know if there's anything more you'd like here.

## Task Pipelines && Caching

Interestingly, I noticed these tests all have a dependency on the workspace-level build being run - so if you run:

```terminal
> git clean -dfx && pnpm install && pnpm run test:lib
```

some packages will fail due to the way the workspace is set up (the `package.json`s are pointing to built locations for the `main`/`module` properties)

This will work though:

```terminal
> git clean -dfx && pnpm install && pnpm build && pnpm run test:lib
```

I expect in ci and most devs' environments, it's likely not a significant issue as likely a build has occurred at some point (though it might not always be current!).

It would be awesome to introduce task caching and a task dependency pipeline to this workspace. Would be happy to work on a separate effort to introduce this via Nx. Please let me know if there's interest here, I believe we could get a PoC up without too much effort.


packages/react-query/src/__tests__/ssr-hydration.test.tsx
```diff
@@ -43,7 +44,8 @@ function PrintStateComponent({ componentName, result }: any): any {
   return `${componentName} - status:${result.status} fetching:${result.isFetching} data:${result.data}`
 }
 
-describe('Server side rendering with de/rehydration', () => {
+// TODO: fix these - probably gotta ask about this before I can fix it
+describe.skip('Server side rendering with de/rehydration', () => {
```
ZackDeRose, line 48: This is the last big hurdle I believe.

packages/vue-query/src/__mocks__/useBaseQuery.ts
```diff
@@ -1,5 +1,7 @@
+import { vi } from 'vitest'
```
ZackDeRose, line 1: Note - I'm not entirely sure how this mock is getting used (or if it is used?)

packages/vue-query/src/__mocks__/useQueryClient.ts
```diff
@@ -1,4 +1,5 @@
 import { QueryClient } from '../queryClient'
+import { vi } from 'vitest'
```
ZackDeRose, line 2: Note - I'm not entirely sure how this mock is getting used (or if it is used?)

packages/vue-query/test-setup.ts
```diff
@@ -0,0 +1,12 @@
+import { vi } from 'vitest'
```
ZackDeRose, line 1: note: this is how I ended up mocking the `vue-demi` package in our tests (ended up replacing the file: `packages/vue-query/__mocks__/vue-demi.ts` as I didn't fully understand how these `__mocks__` directories are getting used.

pnpm-lock.yaml
```diff
@@ -473,7 +477,7 @@ importers:
       expo-status-bar: 1.1.0
       react: 17.0.1
       react-dom: 17.0.1_react@17.0.1
-      react-native: 0.64.3_gpe6tsd6gj6gjuc7hywxyjd2qm
+      react-native: 0.64.3_react@17.0.1
```
ZackDeRose, line 480: not aware of y'all's strat for lock files. Let me know if this needs any adjustment.

TkDodo:
> some packages will fail due to the way the workspace is set up (the package.jsons are pointing to built locations for the main/module properties)

is this the issue you were talking about, because it shows up in CI as well:

> packages/query-persist-client-core test:lib: Error: Failed to resolve entry for package "@tanstack/query-core". The package may have incorrect main/module/exports specified in its package.json.

we don't really want to run a build before running tests because that also takes some time. I'm wondering why this was working before with jest 🤔

ZackDeRose:
> is this the issue you were talking about, because it shows up in CI as well:

yes, precisely.

> I'm wondering why this was working before with jest 🤔

I was assuming there was some build happening in the ci prior to the tests running. I wonder if there exists a vitest config for letting vitest import via the `paths` in `tsconfig.base.json`. I'll see if I can figure that out.

TkDodo:
> I was assuming there was some build happening in the ci prior to the tests running.

no, there definitely wasn't.

ZackDeRose:
> > I was assuming there was some build happening in the ci prior to the tests running.
> 
> no, there definitely wasn't.

I believe I have a fix: https://www.npmjs.com/package/vite-tsconfig-paths

I'll revert the changes around the `package.json` files and see if that fixes our problems.

ZackDeRose:
> > > I was assuming there was some build happening in the ci prior to the tests running.
> > 
> > 
> > no, there definitely wasn't.
> 
> I believe I have a fix: https://www.npmjs.com/package/vite-tsconfig-paths
> 
> I'll revert the changes around the `package.json` files and see if that fixes our problems.

✅ Did revert the changes to `package.json`
👀 Found that we can use a `resolve.alias` option in `vitest.config.ts` files, so opted to use this rather than add another dependency

TkDodo:
there are a ton of "invalid hook call" errors on the Test React17 check, which is weird.

I just realized: on the `alpha` branch, we don't have this check anymore because we dropped react17 support. If it's easier, maybe basing this PR to `alpha` would be easier? But I would expect conflicts 😅 so not sure

ZackDeRose:
> there are a ton of "invalid hook call" errors on the Test React17 check, which is weird.

Yeah, I think that's pretty bad news actually for getting us into main. I think vitest's mock method must be slightly different than jest's, because the method used to cover react 17 support definitely is not working right now. Absolutely seeing react 18 and react-dom 18 in the stack right before react 17 and react 17.

I'm gonna run an experiment on the pipeline here to see if we can fix it by overwriting these deps instead of mocking them out.

> I just realized: on the `alpha` branch, we don't have this check anymore because we dropped react17 support. If it's easier, maybe basing this PR to `alpha` would be easier? But I would expect conflicts 😅 so not sure

I hope it doesn't come to that, but honestly, it might! 🤣 I'll check in tomorrow - hopefully we'll have good news from that experiment.

ZackDeRose:
Quick update @TkDodo - looks like I need an approver to approve the new workflow I submitted. I wanted to try running those react-17 tests by overwriting the contents in `node_modules` via an extra couple steps in the workflow (as you can see in the commit). I think that may work (and if it does, I can obviously update the rest of the workspace to reflect this way of testing react 17) - but let me know if there's other things I'm missing about why it won't work (or if it's not worth the shifting).

TkDodo:
> but let me know if there's other things I'm missing about why it won't work (or if it's not worth the shifting).

I'm fine with having `vitest` only on `alpha`, whatever suits you better. No matter how we approach it, getting this migration into `alpha` might not be easy - even if we get it to `main` first, we have to get `main` back to `alpha`, too 😅 

I think if we add additional tests on `main` in the future it will not be a problem to merge them to `alpha` and change some vitest specific things like mocks, and that would be the only drawback of not having it in main imo

and it looks like test pipeline is 🟢 , but react17 is still 🔴  :(

TkDodo:
seems like code-coverage was only collected for the `vue` package, which is why codecov reports a big decrease

ZackDeRose:
> No matter how we approach it, getting this migration into alpha might not be easy - even if we get it to main first, we have to get main back to alpha, too 😅

I think probably best to sunset this PR in favor of a PR into the `alpha` branch. I should have most of this knowledge primed for any needed adjustments having been through it this time. Let me know if you agree this is the best next step @TkDodo.

> seems like code-coverage was only collected for the `vue` package, which is why codecov reports a big decrease

Hrmm. I noticed that `svelte-query` (which was already using vitest) was doing some things for coverage. Was it the `svelte` package that had coverage reporting on? Either way I can look to match their coverage reporting so we can get that number back up.

TkDodo:
> Let me know if you agree this is the best next step

yeah if it's not too much work for you, I'd agree ❤️ . you can also try to merge alpha into this PR to see how severe conflicts are?

yeah `svelte-query` is the one that already uses `vite` and coverage collection works there

PR #5092, opened 2023-03-07
fix: queryClient consistent as a second parameter

DamianOsipiuk:
- `useQueries` accept `queryClient` as second parameter just like other hooks
- solid `useIsFetching` and `useIsMutating` accept `queryClient` as second parameter

### DIFF
--- a/packages/solid-query/src/createQueries.ts
+++ b/packages/solid-query/src/createQueries.ts
@@ -150,18 +150,18 @@ export type QueriesResults<
 export function createQueries<T extends any[]>(
   queriesOptions: () => {
     queries: readonly [...QueriesOptions<T>]
-    queryClient?: QueryClient
   },
+  queryClient?: () => QueryClient,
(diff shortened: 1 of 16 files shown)


packages/solid-query/src/createQueries.ts
```diff
@@ -150,18 +150,18 @@ export type QueriesResults<
 export function createQueries<T extends any[]>(
   queriesOptions: () => {
     queries: readonly [...QueriesOptions<T>]
-    queryClient?: QueryClient
   },
+  queryClient?: () => QueryClient,
```
ardeora, line 154:
```suggestion
  queryClient?: Accessor<QueryClient>,
```
Just a small change 😄

ardeora (changes requested):
Everything else looks great! Thanks for making these changes and making the API consistent

ardeora (approved):

PR #5093, opened 2023-03-07
fix(solid-query): ssr fixes

marbemac:
~### 1. Don't use `onCleanup` on the server~

~It's still not safe to rely on `onCleanup` on the server, so I've re-added the `!isServer` check for the time being. I was able to isolate + reproduce the `onCleanup` behavior, and opened an issue in the solidjs repo - we'll see what Ryan + team thinks -> https://github.com/solidjs/solid/issues/1599.~

Nevermind! Ryan fixed the https://github.com/solidjs/solid/issues/1599 in solid v1.6.13 - I verified it solves the problem here in solid-query, and bumped the minimum solid-js version to 1.6.13.

**Before**

Any page with more than one query on it cannot be rendered in SSR (streaming or regular async render). Try firing up the example and loading `http://127.0.0.1:3000/deferred` - it will load indefinitely.

Screenshot 2023-03-07 at 3 49 57 PM

**After**

Routes with more than one query on them can be rendered in SSR.

### 2. Update to prefer the query resource value, if present

In particular, this makes it possible to rely on props like `isFetching`, etc, while rendering on the server. It also ensures that the query state on the client starts with the correct values (e.g. if the query was already fetched on the server, `isFetched` should start off as `true` on the client).

I'm still unsure how to add tests for SSR related behavior, so for now my best "testing" method has been to create examples and manually verify. I added a `http://127.0.0.1:3000/hydration` route to the `solid-start-streaming` example. 

**Before**

Note the mis-match on values like `isFetching` - the server and initial render still think it is true, when in fact it is done fetching and should be false. Also.. `isRefetching` set to `true` across the board - weird.

Screenshot 2023-03-07 at 3 38 56 PM

**After**

Note how after these changes, the values on the server, and on initial render, are more along the lines of what one would expect.

Screenshot 2023-03-07 at 3 39 46 PM

I'm not sure if there is a better way to solve this issue, or if something else is going on that is better fixed in query-core. There is still one weird bit of behavior - on first refetch note how `isFetched` briefly is set to false. Overall though this change feels like a solid step forward.. just not QUITE there yet across the board.

---

In case @ardeora or anybody else wants to mess around with the behavior, here's he best way I've found to get the example going for local dev:

1. Set `@tanstack/solid-query` to `workspace:*` in the example package.json (https://github.com/TanStack/query/blob/alpha/examples/solid/solid-start-streaming/package.json#L18)
2. Set `main` in the solid-query `package.json` to `"main": "src/index.js",`. Remove the `exports`, `module`, and `types` properties.
3. `pnpm i`
4. `cd examples/solid/solid-start-streaming && pnpm dev`
5. Make changes to the solid-query package as needed, the example will hot-reload

### DIFF
--- a/packages/solid-query/package.json
+++ b/packages/solid-query/package.json
@@ -54,7 +54,7 @@
     "@tanstack/query-core": "workspace:*"
   },
   "peerDependencies": {
-    "solid-js": "^1.6.2"
+    "solid-js": "^1.6.13"
   },
   "peerDependenciesMeta": {}
 }
--- a/packages/solid-query/src/createBaseQuery.ts
+++ b/packages/solid-query/src/createBaseQuery.ts
@@ -172,7 +172,7 @@ export function createBaseQuery<
   )
 
   onCleanup(() => {
-    if (unsubscribe) {
+    if (!isServer && unsubscribe) {
       unsubscribe()
       unsubscribe = null
     }
--- a/packages/solid-query/src/createBaseQuery.ts
+++ b/packages/solid-query/src/createBaseQuery.ts
@@ -209,10 +209,8 @@ export function createBaseQuery<
       target: QueryObserverResult<TData, TError>,
       prop: keyof QueryObserverResult<TData, TError>,
     ): any {
-      if (prop === 'data') {
-        return queryResource()?.data
-      }
-      return Reflect.get(target, prop)
+      const val = queryResource()?.[prop]
+      return val !== void 0 ? val : Reflect.get(target, prop)
     },
   }
 
--- a/packages/solid-query/src/createBaseQuery.ts
+++ b/packages/solid-query/src/createBaseQuery.ts
@@ -63,7 +63,20 @@ export function createBaseQuery<
   ) => {
     return observer.subscribe((result) => {
       notifyManager.batchCalls(() => {
-        const unwrappedResult = { ...unwrap(result) }
+        const query = observer.getCurrentQuery()
+        const unwrappedResult = {
+          ...unwrap(result),
+
+          // hydrate() expects a QueryState object, which is similar but not
+          // quite the same as a QueryObserverResult object. Thus, for now, we're
+          // copying over the missing properties from state in order to support hydration
+          dataUpdateCount: query.state.dataUpdateCount,
+          fetchFailureCount: query.state.fetchFailureCount,
+          fetchFailureReason: query.state.fetchFailureReason,
+          fetchMeta: query.state.fetchMeta,
+          isInvalidated: query.state.isInvalidated,
+        }
+
         if (unwrappedResult.isError) {
           if (process.env['NODE_ENV'] === 'development') {
             console.error(unwrappedResult.error)
--- a/packages/solid-query/src/createBaseQuery.ts
+++ b/packages/solid-query/src/createBaseQuery.ts
@@ -177,9 +177,17 @@ export function createBaseQuery<
     }
   })
 
-  createComputed(() => {
-    observer.setOptions(client().defaultQueryOptions(options()))
-  })
+  createComputed(
+    on(
+      () => client().defaultQueryOptions(options()),
+      () => observer.setOptions(client().defaultQueryOptions(options())),
+      {
+        // Defer because we don't need to trigger on first render
+        // This only cares about changes to options after the observer is created
+        defer: true,
+      },
+    ),
+  )
 
   createComputed(
     on(
--- a/packages/solid-query/src/createBaseQuery.ts
+++ b/packages/solid-query/src/createBaseQuery.ts
@@ -210,7 +210,7 @@ export function createBaseQuery<
       prop: keyof QueryObserverResult<TData, TError>,
     ): any {
       const val = queryResource()?.[prop]
-      return val !== void 0 ? val : Reflect.get(target, prop)
+      return val !== undefined ? val : Reflect.get(target, prop)
     },
   }
 
--- a/packages/solid-query/src/createBaseQuery.ts
+++ b/packages/solid-query/src/createBaseQuery.ts
@@ -172,7 +172,7 @@ export function createBaseQuery<
   )
 
   onCleanup(() => {
-    if (!isServer && unsubscribe) {
+    if (unsubscribe) {
       unsubscribe()
       unsubscribe = null
     }
--- a/packages/solid-query/src/createBaseQuery.ts
+++ b/packages/solid-query/src/createBaseQuery.ts
@@ -18,7 +18,6 @@ import {
   createResource,
   on,
   onCleanup,
-  onMount,
 } from 'solid-js'
 import { createStore, unwrap } from 'solid-js/store'
 import { useQueryClient } from './QueryClientProvider'
@@ -176,10 +175,6 @@ export function createBaseQuery<
       unsubscribe()
       unsubscribe = null
     }
-  })
-
-  onMount(() => {
-    observer.setOptions(defaultedOptions, { listeners: false })
   })
 
   createComputed(() => {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -71,7 +71,7 @@ importers:
       rollup-plugin-visualizer: ^5.9.0
       rollup-preset-solid: ^1.4.0
       semver: ^7.3.8
-      solid-js: ^1.5.7
+      solid-js: ^1.6.13
       solid-testing-library: ^0.3.0
       stream-to-array: ^2.3.0
       ts-node: ^10.9.1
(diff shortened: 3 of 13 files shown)


packages/solid-query/package.json
```diff
@@ -54,7 +54,7 @@
     "@tanstack/query-core": "workspace:*"
   },
   "peerDependencies": {
-    "solid-js": "^1.6.2"
+    "solid-js": "^1.6.13"
```
marbemac, line 57: Bumped for this fix -> https://github.com/solidjs/solid/commit/60f8624db81e5491bdd3f4b69870899976212941

pnpm-lock.yaml
```diff
@@ -249,15 +249,15 @@ importers:
...
       vite: ^3.0.0
     dependencies:
-      '@tanstack/query-sync-storage-persister': link:../../../packages/query-sync-storage-persister
-      '@tanstack/react-query': link:../../../packages/react-query
-      '@tanstack/react-query-devtools': link:../../../packages/react-query-devtools
-      '@tanstack/react-query-persist-client': link:../../../packages/react-query-persist-client
+      '@tanstack/query-sync-storage-persister': 4.26.1
+      '@tanstack/react-query': 4.26.1_biqbaboplfbrettd7655fr4n2y
+      '@tanstack/react-query-devtools': 4.26.1_brdhmlf72zuns3lsk66phyptty
+      '@tanstack/react-query-persist-client': 4.26.1_vohxj3xpfxn57cw32hqo6tozmq
```
marbemac, line 255: I'm not sure why there are so many changes in the lock file that don't seem related to the packages I bumped. All I did was run `pnpm install` after bumping. For reference, I'm on pnpm `v7.25.0`.

packages/solid-query/src/createBaseQuery.ts
```diff
@@ -209,10 +209,8 @@ export function createBaseQuery<
       target: QueryObserverResult<TData, TError>,
       prop: keyof QueryObserverResult<TData, TError>,
     ): any {
-      if (prop === 'data') {
-        return queryResource()?.data
-      }
-      return Reflect.get(target, prop)
+      const val = queryResource()?.[prop]
+      return val !== void 0 ? val : Reflect.get(target, prop)
```
ardeora, line 213: TIL About the `void 0` operation 😅  Is there a reason why we dont use just `undefined` here?
```suggestion
      return val !== undefined ? val : Reflect.get(target, prop)
```

ardeora (commented):
Thank you so much for working on the example and the fix. This looks great! Just have a simple comment about `void 0` being used. Everything else looks 🔥  Could we also rebase from the alpha branch for me to merge it to the `alpha` branch?

ardeora:
@marbemac Also the `isFetched` being false on the first refetch is because when we call `hydrate` it doesnt seem to set the `dataUpdatedCount` property when dehydrating the query on the server. This property is not available on the queryObserver result so we might have to manually set it? https://github.com/TanStack/query/blob/fe0ae802d8e7521ba6d55239e7daf03ec0dd20fa/packages/query-core/src/queryObserver.ts#L573

marbemac:
Howdy @ardeora - apologies for the delay here! I was out for a bit, but will take a look today.

marbemac:
> Also the isFetched being false on the first refetch is because createBaseQuery doesnt seem to set the dataUpdateCount property when dehydrating the query on the server. This property is not available on the queryObserver result so we might have to manually set it?

Hmm, begs the question.. why is the query state that we're hydrating missing some properties? This it the first time I'm actually digging into query-core a bit, and there might be a larger issue?

I didn't realize that `QueryObserverResult` and `QueryState` are managed separately. Basically, we're hydrating the query and setting the `state`, which is expecting `QueryState`, to a `QueryObserverResult` object in https://github.com/TanStack/query/blob/alpha/packages/solid-query/src/createBaseQuery.ts#L149 (since this is what `observer.subscribe` passes us).

Looking at `QueryState`, 9 of 12 properties overlap with `QueryObserverResult`.

```ts
interface QueryState<TData = unknown, TError = DefaultError> {
  data: TData | undefined // present on QueryObserverResult
  dataUpdateCount: number // NOT present on QueryObserverResult
  dataUpdatedAt: number // present on QueryObserverResult
  error: TError | null // present on QueryObserverResult
  errorUpdateCount: number // present on QueryObserverResult
  errorUpdatedAt: number // present on QueryObserverResult
  fetchFailureCount: number // present BUT with a different name (failureCount) on QueryObserverResult
  fetchFailureReason: TError | null // present BUT with a different name (failureReason) on QueryObserverResult
  fetchMeta: FetchMeta | null // NOT present on QueryObserverResult
  isInvalidated: boolean // NOT present on QueryObserverResult (perhaps is present but with a different name? not sure what this is)
  status: QueryStatus // PRESENT on QueryObserverResult
  fetchStatus: FetchStatus // PRESENT on QueryObserverResult
}
```

For the immediate issue at hand, the `isFetching` issue is indeed occurring because `dataUpdateCount` is not present on `QueryObserverResult` (interestingly, `errorUpdateCount` IS present on `QueryObserverResult`).

 is the intent to ultimately simplify and consolidate `QueryState` into `QueryObserverResult`? Or does managing `QueryState` as a separate thing serve some other functional purpose?

For the time being, I'll see if I can get the actual state object hydrating (rather than `QueryObserverResult`).

marbemac:
Was wondering why the typings are not catching this mismatch re passing `QueryObserverResult` where `QueryState` is expected.

Should we consider typing the `dehydratedState` argument? If it had been typed we would have seen this issue much sooner -> https://github.com/TanStack/query/blob/alpha/packages/query-core/src/hydration.ts#L116

packages/solid-query/src/createBaseQuery.ts
```diff
@@ -178,13 +190,17 @@ export function createBaseQuery<
     }
   })
 
-  onMount(() => {
-    observer.setOptions(defaultedOptions, { listeners: false })
-  })
```
marbemac, line 183: From what I can tell, this isn't doing anything. Removed to simplify and reduce the number of calls to `queryObserver.createResult`. LMK if there's a reason for this and I can revert np!

ardeora (approved):

rollup.config.ts
```diff
@@ -35,12 +35,12 @@ const babelPlugin = (type: 'legacy' | 'modern') =>
...
-            edge: 79,
-            safari: 12,
-            ios: 12,
-            opera: 53,
+            chrome: '73',
+            firefox: '78',
+            edge: '79',
+            safari: '12',
+            ios: '12',
+            opera: '53',
```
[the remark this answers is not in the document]
ardeora: > > Also the isFetched being false on the first refetch is because createBaseQuery doesnt seem to set the dataUpdateCount property when dehydrating the query on the server. This property is not available on the queryObserver result so we might have to manually set it?
> 
> Hmm, begs the question.. why is the query state that we're hydrating missing some properties? This it the first time I'm actually digging into query-core a bit, and there might be a larger issue?
> 
> I didn't realize that `QueryObserverResult` and `QueryState` are managed separately. Basically, we're hydrating the query and setting the `state`, which is expecting `QueryState`, to a `QueryObserverResult` object in https://github.com/TanStack/query/blob/alpha/packages/solid-query/src/createBaseQuery.ts#L149 (since `observer.subscribe` passes us a `QueryObserverResult`).
> 
> **Should we be subscribing to the query cache instead (since it passes the current query, including state, to the listeners) https://github.com/TanStack/query/blob/alpha/packages/query-core/src/queryObserver.ts#L665?**
> 
> Looking at `QueryState`, 9 of 12 properties overlap with `QueryObserverResult`.
> 
> ```ts
> interface QueryState<TData = unknown, TError = DefaultError> {
>   data: TData | undefined // present on QueryObserverResult
>   dataUpdateCount: number // NOT present on QueryObserverResult
>   dataUpdatedAt: number // present on QueryObserverResult
>   error: TError | null // present on QueryObserverResult
>   errorUpdateCount: number // present on QueryObserverResult
>   errorUpdatedAt: number // present on QueryObserverResult
>   fetchFailureCount: number // present BUT with a different name (failureCount) on QueryObserverResult
>   fetchFailureReason: TError | null // present BUT with a different name (failureReason) on QueryObserverResult
>   fetchMeta: FetchMeta | null // NOT present on QueryObserverResult
>   isInvalidated: boolean // NOT present on QueryObserverResult (perhaps is present but with a different name? not sure what this is)
>   status: QueryStatus // PRESENT on QueryObserverResult
>   fetchStatus: FetchStatus // PRESENT on QueryObserverResult
> }
> ```
> 
> For the immediate issue at hand, the `isFetching` issue is indeed occurring because `dataUpdateCount` is not present on `QueryObserverResult` (interestingly, `errorUpdateCount` IS present on `QueryObserverResult`).
> 
> is the intent to ultimately simplify and consolidate `QueryState` into `QueryObserverResult`? Or does managing `QueryState` as a separate thing serve some other functional purpose? Huge caveat here that I probably don't know what I'm talking about, JUST getting into query-core codebase for the first time 😅.
> 
> For the time being, I'll see if I can get the actual state object hydrating (rather than `QueryObserverResult`).

One reason why they both are not exactly the same, because things like placeholderData or using the `select` function (to transform data) will only affect the data on the observer level. The queryState data is not always equal to the data on the QueryObserverResult.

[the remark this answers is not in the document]
marbemac: Hmm yeah I can see why query result and query state might each have their own `data` properties. Still not clear why most of the other properties are duplicated between the two, some with slightly different names. A topic for another time though, thanks for merging!

PR #5095, opened 2023-03-07
fix(query-sync-storage-persister): Relax undefined check against storage to handle null values

### LINKED ISSUE #4957: query-sync-storage-persister raises on Android WebView if `setDomStorageEnabled` is not configured
### Describe the bug

An Android WebView that is not configured with `settings.setDomStorageEnabled(true)` will not expose `window.localStorage` for use.

`query-sync-storage-persister` guards against the value for `storage` being `undefined`:

https://github.com/TanStack/query/blob/b82f05e3028718f03eb132a4dd92dc3d7623f7b5/packages/query-sync-storage-persister/src/index.ts#L45

I think this is necessary for old browsers that predate local storage, where `window.localStorage === undefined`. Given the persister is typically initialized with `createSyncStoragePersister({ storage: window.localStorage })` that makes sense.

However, in the Android WebView case, it appears that `window.localStorage === null` - the property is present on `window` but the value is `null` rather than `undefined`.

As a result, the conditional linked above passes (`typeof null` is `"object"` 🤷 ), and `removeClient` can later raise with `TypeError: Cannot read properties of null (reading 'removeItem')`:

https://github.com/TanStack/query/blob/b82f05e3028718f03eb132a4dd92dc3d7623f7b5/packages/query-sync-storage-persister/src/index.ts#L81-L83

It looks like `persistClient` doesn't hit this because `trySave` has some error handling around it:

https://github.com/TanStack/query/blob/b82f05e3028718f03eb132a4dd92dc3d7623f7b5/packages/query-sync-storage-persister/src/index.ts#L46-L53

I can guard against this when initializing `createSyncStoragePersister`, but I think it probably makes sense to relax the conditional in `createSyncStoragePersister` to just `if (storage) { ... }` unless there's a good reason not to.

None sorry, but the bug is fairly trivial

### Steps to reproduce

To reproduce you'd need an app installed with a `WebView` configured in a particular way (unsure if it would need to be configured as `settings.setDomStorageEnabled(false)`, or if that's the default).

Whatever triggers `removeClient` would then throw rather than noop.

### Expected behavior

If `window.localStorage` is `null`, initializing the query persister via `createSyncStoragePersister({ storage: window.localStorage })` should behave the same as if `window.localStorage` is `undefined` - queries aren't persisted but no exceptions are raised.

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: Android
- Browser: Chrome Mobile WebView
- Version: Appears to affect all versions (have seen Sentry reports from `Chrome Mobile WebView 72.0.3626` all the way through to `Chrome Mobile WebView 110.0.5481`)

### TanStack Query version

@tanstack/query-sync-storage-persister@4.24.4

### TypeScript version

4.7.4

### Additional context

_No response_

nfm:
Fixes #4957.

On Android WebView if `setDomStorageEnabled` is not configured, `window.localStorage` will be `null` (rather than `undefined`).

Relax the existing conditional to check if `storage` is present to handle both `null` and `undefined` values to remove a gotcha here.


packages/query-sync-storage-persister/src/index.ts
```diff
@@ -42,7 +43,7 @@ export function createSyncStoragePersister({
   deserialize = JSON.parse,
   retry,
 }: CreateSyncStoragePersisterOptions): Persister {
-  if (typeof storage !== 'undefined') {
+  if (storage) {
```
DamianOsipiuk, line 46:
```suggestion
  if (storage != null) {
```

This is actually a good place to use `!= null`. It will guard against `null` and `undefined` but be true for other potentially `falsy` values like `0`

TkDodo: yeah but we have types for that 😅. It's not allowed to pass in a storage that's a string or a boolean. In those cases, I would prefer to make the check that has the least amount of code :)

TkDodo:
@nfm doesn't the async persister have the same guard?

packages/query-sync-storage-persister/src/index.ts
nfm: I'm easy either way - let me know what the preference is and I can change it (or leave it as is).

DamianOsipiuk: Leave it as-is

nfm:
> @nfm doesn't the async persister have the same guard?

Thanks @TkDodo - looks like it does 👍 I haven't used the async persister but it looks like it's susceptible to the same issue. I'll push a fix for this too.

nfm:
It looks like this failed a test but I'm not sure why - from what I can gather it looks unrelated:

```
 FAIL   react-query  src/__tests__/useQuery.test.tsx (33.03 s)
  ● useQuery › should call onError after a query has been fetched with an error
```

Please let me know if I've done something silly here 🙃

TkDodo (approved):

TkDodo:
just our standard test flakiness ... working on it 😅

TkDodo (approved):

PR #5097, opened 2023-03-08
Jest to vitest migration alpha

ZackDeRose:
Mostly a clone of #5081, but for the `alpha` branch (see discussion in that pr w/ @TkDodo)

Addresses #5074


packages/query-async-storage-persister/src/__tests__/asyncThrottle.test.ts
```diff
@@ -105,11 +106,10 @@ describe('asyncThrottle', () => {
     expect(mockFunc.mock.calls[1]?.[0]).toBe(2)
   })
 
-  test('"onError" should be called when "func" throw error', (done) => {
+  test('"onError" should be called when "func" throw error', () => {
     const err = new Error('error')
     const handleError = (e: unknown) => {
       expect(e).toBe(err)
-      done()
```
TkDodo, line 112: how does vitest know to wait for this expectation to occur ?

packages/query-core/src/tests/query.test.tsx
```diff
@@ -200,7 +191,6 @@ describe('query', () => {
     expect(queryFn).toHaveBeenCalledTimes(1)
     const args = queryFn.mock.calls[0]![0]
     expect(args).toBeDefined()
-    // @ts-expect-error page param should be undefined
     expect(args.pageParam).toBeUndefined()
```
TkDodo, line 194: it shouldn't be necessary to remove this 🤔 . With this ts-expect-error, we were testing that `args.pageParam` is not accessible on TS level. Is this now because `vi.fn` is not typed anymore ?

packages/react-query/src/__tests__/useMutationState.test.tsx
```diff
@@ -275,10 +221,65 @@ describe('useMutationState', () => {
...
+          {mounted && <IsMutating />}
+        </div>
+      )
+    }
+
+    const { getByText } = renderWithClient(queryClient, <Page />)
+    fireEvent.click(getByText('unmount'))
+
+    // Should not display the console error
+    // "Warning: Can't perform a React state update on an unmounted component"
```
TkDodo, line 280: feel free to delete this whole test. this warning is no longer a thing in react18, so the test doesn't test anything

packages/react-query/src/__tests__/useQueries.test.tsx
```diff
@@ -931,4 +873,63 @@ describe('useQueries', () => {
...
+    }
+
+    const { getByText } = renderWithClient(queryClient, <Page />)
+    fireEvent.click(getByText('unmount'))
+
+    // Should not display the console error
+    // "Warning: Can't perform a React state update on an unmounted component"
+
+    await sleep(20)
+    QueriesObserverSpy.mockRestore()
```
TkDodo, line 933: same here - we don't need that test

packages/react-query/src/__tests__/useQuery.test.tsx
```diff
@@ -4057,24 +4057,22 @@ describe('useQuery', () => {
 
     await waitFor(() => rendered.getByText('fetched data'))
 
-    jest.useFakeTimers({
-      legacyFakeTimers: true,
-    })
-    const setTimeoutSpy = jest.spyOn(globalThis.window, 'setTimeout')
+    // vi.useFakeTimers()
```
TkDodo, line 4060: is the `vi.useFakeTimers` call is not needed, can we remove it please?

packages/solid-query/package.json
```diff
@@ -51,7 +51,8 @@
     "solid-jest": "^0.2.0"
   },
   "dependencies": {
-    "@tanstack/query-core": "workspace:*"
+    "@tanstack/query-core": "workspace:*",
+    "vite-plugin-solid": "^2.3.9"
```
TkDodo, line 55: shouldn't this be a `devDependency` ?

packages/vue-query/src/useBaseQuery.ts
```diff
@@ -131,10 +122,10 @@ export function useBaseQuery<
           )
           if (optimisticResult.isStale) {
             stopWatch()
-            resolve(observer.fetchOptimistic(defaultedOptions.value))
+            resolve(observer.fetchOptimistic(defaultedOptions.value) as any)
           } else {
             stopWatch()
-            resolve(optimisticResult)
+            resolve(optimisticResult as any)
```
TkDodo, line 128: @DamianOsipiuk FYI

packages/vue-query/tsconfig.json
```diff
@@ -4,10 +4,9 @@
     "composite": true,
     "rootDir": "./src",
     "outDir": "./build/lib",
-    "tsBuildInfoFile": "./build/.tsbuildinfo"
+    "tsBuildInfoFile": "./build/.tsbuildinfo",
+    "module": "esnext"
```
TkDodo, line 8: doesn't this change the output of the type definitions for the `vue` package?

TkDodo (commented):
Thank you 🙏 ❤️ . Please have a look at my comments:

packages/vue-query/src/useBaseQuery.ts
```diff
@@ -68,15 +67,7 @@ export function useBaseQuery<
   const client = queryClient || useQueryClient()
 
   const defaultedOptions = computed(() => {
-    const defaulted = client.defaultQueryOptions(
```
DamianOsipiuk, line 71: This change causes causes the need to add `any` later in this file, which is not a good thing.
If you get ts errors in this line for some reason, please add `ts-expect-error` in this place and i will take a look at that later on

packages/vue-query/tsconfig.json
DamianOsipiuk: @ZackDeRose Could you please explain why this change is necessary?

packages/vue-query/src/__mocks__/useBaseQuery.ts
```diff
@@ -1,5 +1,7 @@
+import { vi } from 'vitest'
+
 const { useBaseQuery: originImpl, unrefQueryArgs: originalParse } =
-  jest.requireActual('../useBaseQuery')
+  (await vi.importActual('../useBaseQuery')) as any
 
-export const useBaseQuery = jest.fn(originImpl)
```
DamianOsipiuk, line 4: If `vi.mock` automatically wraps all `original implementations` with spy functions, then we could remove it.
Jest used the `__mock__` directory for mocking modules, without necessity to write the same mock in multiple files.

packages/vue-query/src/__mocks__/useQueryClient.ts
```diff
@@ -1,9 +1,10 @@
...
+import { vi } from 'vitest'
 
 const queryClient = new QueryClient({
   defaultOptions: {
     queries: { retry: false, gcTime: Infinity },
   },
 })
 
-export const useQueryClient = jest.fn(() => queryClient)
+export const useQueryClient = vi.fn(() => queryClient)
```
DamianOsipiuk, line 10: This will be used by `automocking` - https://vitest.dev/guide/mocking.html#automocking-algorithm

packages/query-async-storage-persister/src/__tests__/asyncThrottle.test.ts
ZackDeRose: I believe vitest will wait for any unresolved promises created by the test's callback closure to resolve before continuing. I affirmed this assumption by adding a false assertion (`expect(true).toBe(false)`) where the `done()` was and ensuring the test fails).

If we want we can make the test callback `async` and explicitely `await` the last line of this test if we think this behavior might confuse folks!

packages/query-core/src/tests/query.test.tsx
ZackDeRose: 🤔 taking a look. The `vi.fn` function does support generics, they're just structured differently than `jest.fn` generics...

ZackDeRose: Oof looks like they're just backwards - let me comb out the places I removed generics quick....

packages/react-query/src/__tests__/useMutationState.test.tsx
ZackDeRose: :+1:

packages/react-query/src/__tests__/useQueries.test.tsx
ZackDeRose: :+1:

packages/solid-query/package.json
ZackDeRose: :+1: fixing!

packages/vue-query/tsconfig.json
ZackDeRose: Hey folks - yes I needed top-level await for those mock files. The equivalent of jest's synchronous `requireActual` is vitest's async `importActual`. I'll revert this and see if I can fix those mocks some other way.

packages/vue-query/src/__mocks__/useBaseQuery.ts
ZackDeRose: makes sense - thanks!

packages/vue-query/tsconfig.json
ZackDeRose: Looks like the error I was getting is only in the ts language server, so this is safe to revert. See comment in the mock file!

packages/vue-query/src/__mocks__/useBaseQuery.ts
```diff
@@ -1,5 +1,8 @@
+import { vi } from 'vitest'
+
 const { useBaseQuery: originImpl, unrefQueryArgs: originalParse } =
-  jest.requireActual('../useBaseQuery')
+  // @ts-expect-error - vitest uses esmodules; tsconfig is not set to use them
+  (await vi.importActual('../useBaseQuery')) as any
```
ZackDeRose, line 5: After reverting my change to the `tsconfig.json` this error shows up in the IDE via the ts language service, but running `pnpm test:ci` passes just fine even with the error marked in the IDE.

I've added this `// @ts-expect-error` so we don't see this via the IDE, but lmk if there's a preferred fix I'm not aware of.

ZackDeRose:
Setting sights on coverage reports now

ZackDeRose:
let's see if that fixes coverage - confirming I see the reports generated now on my local filesystem

packages/vue-query/src/useBaseQuery.ts
ZackDeRose: The error is actually coming from eslint, not tsc.

I fixed here now by switching the `as` to  a `:` (as you can see), but let me know if you prefer me to revert back to `as` and use a `// eslint-ignore` comment to get this through eslint

packages/query-async-storage-persister/src/__tests__/asyncThrottle.test.ts
TkDodo: yeah that looks good 👍

packages/vue-query/src/useBaseQuery.ts
TkDodo: this looks better than before imo 👍

TkDodo (approved):

TkDodo:
ran into an error:

```
packages/react-query test:lib: FATAL ERROR: v8::FromJust Maybe value is Nothing.
packages/react-query test:lib:  1: 0xb09980 node::Abort() [node]
packages/react-query test:lib:  2: 0xa1c235 node::FatalError(char const*, char const*) [node]
packages/react-query test:lib:  3: 0xcf758a v8::Utils::ReportApiFailure(char const*, char const*) [node]
packages/react-query test:lib:  4: 0xb0d7dd node::fs::FileHandle::CloseReq::Resolve() [node]
```

I've re-run to see if this was just flaky

TkDodo:
hmm the error seems to persist :/

ZackDeRose:
interesting - I'm getting passing `pnpm test:ci` locally :\

investigating now

ZackDeRose:
Was able to reproduce this `FATAL ERROR: v8::FromJust Maybe value is Nothing.` error with node v16.4.2

I'll add a bump this in the ci files to 16.19 (most recent v16), I don't seem to get the error on that version of node, let me know if that acceptable or if we need to stay on this version

DamianOsipiuk (approved):

PR #5107, opened 2023-03-09
fix: Fix `getNextPageParam` types

mrousavy:
When using `useInfiniteQuery`, it can happen that `getNextPageParam` gets called with the first argument being `undefined`. The types however say that it cannot be undefined, so this is the fix to that.

Maybe I'm also using it wrong?

```tsx
interface Result {
  data: Transaction[]
  nextCursor: string | undefined
}
function getTransactions(address: string): Promise<Result> {
  ...
}

function App() {
  const { data, status, fetchNextPage, refetch } = useInfiniteQuery(
    'transactions',
    ({ pageParam }) => getTransactions(address),
    {
      getNextPageParam: (lastPage) => lastPage?.nextCursor, // <-- lastPage here can be undefined, but the type is 'Result<Transaction[]>' (no ' | undefined')
      initialData: { pages: [], pageParams: [] },
    }
  )

  ...
}
```

### DIFF
--- a/packages/query-core/src/types.ts
+++ b/packages/query-core/src/types.ts
@@ -34,12 +34,12 @@ export type QueryKeyHashFunction<TQueryKey extends QueryKey> = (
 ) => string
 
 export type GetPreviousPageParamFunction<TQueryFnData = unknown> = (
-  firstPage: TQueryFnData,
+  firstPage: TQueryFnData | undefined,
   allPages: TQueryFnData[],
 ) => unknown
 
 export type GetNextPageParamFunction<TQueryFnData = unknown> = (
-  lastPage: TQueryFnData,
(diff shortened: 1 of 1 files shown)


TkDodo:
This is probably due to `initialData` having everything empty. Not sure if we should widen the type for this. empty arrays should probably not be allowed as initialData instead ... I'd use `placeholderData` for that

mrousavy:
I just tried again without `initialData`, and it still gets called with an undefined firstPage param 🤔

TkDodo:
Can you show a codesandbox reproduction please?

TkDodo:
I'm gonna close this PR for now, please file an issue with a reproduction because it shouldn't be called with undefined.

PR #5116, opened 2023-03-11
`test:lib` task-caching w/ Nx

ZackDeRose:
<div align="left">
      <a href="https://www.youtube.com/watch?v=5M9-l5IciLU" target="blank">
         
      </a>
</div>

# What This Does

+ Implements local task-caching [currently] only for the `test:lib` task (with Nx)
+ I also update the root `test-lib` script to run with nx's run-many command, as well as the `Test` workflow in `.github/workflows/pr.yml`
  + While not directly affected, the `Run Test` step of the `ci` pipeline is adjusted as well as `test:ci` will call `pnpm run test:lib`
  + My testing shows a 20s time boost to the `Test` job of the `pr` pipeline (see https://github.com/ZackDeRose/query/pull/1#issuecomment-1464847177 for my findings)
+ Note that any `test:lib:dev` commands are not affected and work exactly as they did before

## What this PR doesn't do, but could!

I wanted to make this PR bite-sized (as opposed to my last one!) but would be happy to expand this with more task-caching features:

+ I believe most of the "testing" tasks could be cached in a similar way (this would mainly take a few lines to the `nx.json` file, and some coordination to make sure I correctly understand what files are created by those tasks, and which files should/shouldn't invalidate the cache)
+ builds similarly should be cacheable without too much more work
+ We could introduce distributed caching (w/ [Nx Cloud](https://nx.dev)) - this would mainly be there to enable caching across github action runs [open source projects get a free unlimited license to Nx Cloud]
  + I'd recommend a public read-only key along with a private write key to be stored in github secrets (happy to help set that up)
  + alternatively, we can persist the `node_modules/.cache` between ci runs via the workflow yml files, but that's much more manual process (and we'd have to eventually introduce pruning before this cache just gets too large)
+ We can introduce distributed task execution to further parallelize the ci workflows

@TkDodo (or others!) let me know if any of the above would be appealing! I'd be happy to work on introducing any of these (either in this PR or in future PRs)

[Also note that I work full-time for Nx!]

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -1,0 +1,16 @@
+{
+  "tasksRunnerOptions": {
+    "default": {
+      "runner": "nx/tasks-runners/default",
+      "options": {
+        "cacheableOperations": ["test:lib"]
+      }
+    }
+  },
+  "defaultBase": "main",
+  "targetDefaults": {
+    "test:lib": {
+      "outputs": ["{projectRoot}/coverage"]
+    }
+  }
+}
--- a/nx.json
+++ b/nx.json
@@ -8,9 +8,32 @@
     }
   },
   "defaultBase": "main",
+  "namedInputs": {
+    "globalBuildAffectingConfig": [
+      "{workspaceRoot}/babel.config.js",
+      "{workspaceRoot}/rollup.config.js",
+      "{workspaceRoot}/rollup.config.ts",
+      "{workspaceRoot}/tsconfig.json",
+      "{workspaceRoot}/tsconfig.base.json"
+    ],
+    "globalNonBuildAffectingConfig": [
+      "{workspaceRoot}/.eslintrc",
+      "{workspaceRoot}/jest-preset.js"
+    ],
+    "default": ["{projectRoot}/**/*", "globalBuildAffectingConfig", "globalNonBuildAffectingConfig"],
+    "public": [
+      "default",
+      "!{workspaceRoot}/.eslintrc",
+      "!{workspaceRoot}/jest-preset.js",
+      "!{projectRoot}/**/?(*.)+(spec|test).[jt]s?(x)?(.snap)",
+      "!{projectRoot}/.eslintrc",
+      "!{projectRoot}/jest.config.js"
+    ]
+  },
   "targetDefaults": {
     "test:lib": {
-      "outputs": ["{projectRoot}/coverage"]
+      "outputs": ["{projectRoot}/coverage"],
+      "inputs": ["default", "^public"]
     }
   }
 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,6 @@ on:
       - 'main'
       - 'alpha'
       - 'beta'
-
 jobs:
   test-and-publish:
     if: github.repository == 'TanStack/query' && (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/alpha' || github.ref == 'refs/heads/beta')
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ on:
       - 'main'
       - 'alpha'
       - 'beta'
-  workflow_dispatch:
+
 jobs:
   test-and-publish:
     if: github.repository == 'TanStack/query' && (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/alpha' || github.ref == 'refs/heads/beta')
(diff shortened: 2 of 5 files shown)


nx.json
```diff
@@ -0,0 +1,39 @@
...
+      "{workspaceRoot}/jest-preset.js"
+    ],
+    "default": ["{projectRoot}/**/*", "globalBuildAffectingConfig", "globalNonBuildAffectingConfig"],
+    "public": [
+      "default",
+      "!{workspaceRoot}/.eslintrc",
+      "!{workspaceRoot}/jest-preset.js",
+      "!{projectRoot}/**/?(*.)+(spec|test).[jt]s?(x)?(.snap)",
+      "!{projectRoot}/.eslintrc",
+      "!{projectRoot}/jest.config.js"
```
TkDodo, line 30: will this have to change on `alpha` where we are not using `jest` anymore?

I think, again, maybe it would be better to just introduce those changes on `alpha` ?

TkDodo:
love the idea of nx-cloud ❤️

nx.json
ZackDeRose: Most of these would stay the same - we'd really just want to change out the `jest-preset.js` and `jest.config.js` files for `vitest.config.ts` and `test-setup.ts`.

The surface area here in is quite small, which is why I thought to adjust in `main` directly.

Happy to submit to `alpha` if that's preferred though! Just let me know which you'd prefer :)

TkDodo: okay let's keep it on main, and I'll merge it back to alpha and solve the conflicts

ZackDeRose:
Hey @TkDodo! Looks like this failed in CI due to some flakiness (the error comes right after a `await sleep(20)` - and I confirmed after pulling down the new changes that this passes locally.)

Can we give this another try? Also let me know if anything else you want to see adjusted in this branch!

TkDodo:
Yeah I'm trying to fix flakiness as I see it. Some tests are not written in a robust way 😅

ZackDeRose:
> Yeah I'm trying to fix flakiness as I see it

That's great. If there's an effort to go after these please let me know! I think this one for example, we should use `waitFor` instead!

TkDodo:
yeah all the newer tests are written with `waitFor` instead of sleeps:

https://github.com/TanStack/query/blob/4aaaa51b375bee287677e68ae201f473ef809c6c/packages/react-query/src/__tests__/useQuery.test.tsx#L6057-L6063

ZackDeRose:
Yeah I noticed that! I like that alot. Seems like there might some potential for flakiness based on this timing out, but I've actually never had a test fail on me because of that in any of the testing I've looked at for this project so far!

TkDodo:
Just watched your video - amazing, thank you so much for this ❤️ . I'll merge the PR and then merge it to `alpha`. Whatever you wanna do next, I'm game. adding caching for the lint task, or going nx cloud for what we have already 🎉

TkDodo:
please let me know if I fixed that correctly on alpha 😅 

https://github.com/TanStack/query/commit/ac31fc0e4ee51110226d31810b55ec593f3c4426

TkDodo:
@ZackDeRose I can now see some post-install task failing in our pipelines:
Screenshot 2023-03-14 at 09 34 16

it just logs this error and it still succeeds, but the install times seem higher now (27s vs 11s before)

TkDodo:
okay never mind the longer times, this was just a cache miss on pnpms side. It's back to 11s with a warm cache. But do you know what the error is?

ZackDeRose:
> okay never mind the longer times, this was just a cache miss on pnpms side. It's back to 11s with a warm cache. But do you know what the error is?

Ah yes - the Nx Daemon is a daemon process that runs in the background to eagerly determine Nx project configuration whenever files change. Looks like the daemon stopped because of an issue parsing pnpm's lockfile.

I'll bring this to our core engineers (@), as we should be handling this better - but the daemon failing shouldn't be considered a failing event. If it ever does fail, it should just restart itself the next time Nx would need it!

ZackDeRose:
@TkDodo - I'll be tracking this here: https://github.com/nrwl/nx/issues/15656

JamesHenry:
Just FYI `NX_DAEMON=false` environment variable could be used in the immediate term if the issues is daemon specific

ZackDeRose:
Thanks @JamesHenry! I'll get something up with this added now :)

ZackDeRose:
@TkDodo some quick fixes:

#5128 to fix the error messages re: the nx daemon
#5129 quick adjustment to alpha branch

> Whatever you wanna do next, I'm game. adding caching for the lint task, or going nx cloud for what we have already 🎉

Let's go for adding the rest of our tasks first! I think that should actually be pretty easy! I'll send you a PR soon.

Let's plan for Nx cloud following that because we may have to coordinate that a bit!

Issue #4693, opened 2022-12-23
Investigation: fine-grained async control over caching behaviour


TkDodo:
Related:
- https://github.com/TanStack/query/discussions/2649

One approach would be to have another async function that runs _before_ the `queryFn` (and maybe one that runs _after_ it ?) that would skip calling the `queryFn` if something other than `undefined` was returned. Here, people could read from an async storage (and potentially write to it in the "after" hook). The query could also stay in `fetchStatus: 'idle'` because we haven't triggered the `queryFn` yet.

Combined with the possibility to customize the in-memory cache (see #4687), people would be free to keep the size of the in-memory cache small.

incepter:
Here are the cache insights from my lib:


The `CacheConfig` accepts a:

-  a `load` function that can be sync or async, and returns a record of a CachedState shape, in the case or react-query, it would be just the cached state
- an `onCacheLoads` callback that may be used to imperatively set the state from cached data.
- a `getDeadline` function that receives the Succeeded state, and returns the state time.

The `run` of my `producer` (query for react-query) first calculates the hash of the run (the query key in your case), then checks if there is a non expired cache entry, then set's it as current state (if not already).


I hope these insights provide any help.

Im not linking my lib to avoid being targeted as an ad.

TkDodo:
One thing we could do is add an `onQuery` callback to the `QueryCache` itself. This would be similar to the `onMutate` callback of `useMutation` - it would run _before_ the `queryFn`. 

It could be async, and when something other than `undefined` is returned, we would use that as the result for this query instead of calling the `queryFn`.

> The query could also stay in fetchStatus: 'idle' because we haven't triggered the queryFn yet.

Not sure this is possible or desired.

> and maybe one that runs after it ?

This is what the `onSuccess` callback could be used for.

In combination with #4687 , we could so something like this:

```
const queryClient = new QueryClient({
  queryCache: new QueryCache({
    cache: createMyCustomCache(),
    onQuery: async (query) => {
      const data = await readFromAsyncStorage(query.queryKey)
      return data ?? undefined
    }
    onSuccess: (data) => {
      writeToAsyncStorage(query.queryKey, data)
    }
  })
})
```

DamianOsipiuk:
@TkDodo I think this topic connect to the one we have discussed previously.

Potentially we could introduce a `persister` argument to the query `options`.
If provided, `persister` could be run instead of a `queryFn`.
[...]

PR #5125, opened 2023-03-13
feat: fine grained storage persister

DamianOsipiuk:
Ref: https://github.com/TanStack/query/issues/4693

This is a POC of a `persister` that works for both `sync` and `async` storage

Things that are missing from the currently implemented client persister:
- `throttling`
  - do we need it when persisting one query at a time?
- `retry`
  - TODO
  - needed for `out of space` issues that were reported previously
- `single key storage`
  - do we really need it? you can search in local storage, but may trigger some people OCD
- `persisting paused mutations`
  - offline mode?

How to run?
- `cd examples/vue/persister`
- `pnpm i`
- `pnpm run dev`
- list is persisted to `localStorage` and pages to `idb`

Seems to be working pretty well...

### DIFF
--- a/examples/vue/persister/src/persister.ts
+++ b/examples/vue/persister/src/persister.ts
@@ -1,0 +1,128 @@
+import type {
+  QueryClient,
+  QueryFunctionContext,
+  QueryKey,
+  QueryState,
+} from '@tanstack/query-core'
+import { hashQueryKey } from '@tanstack/query-core'
+
+export type Promisable<T> = T | PromiseLike<T>
+
+export interface PersistedQuery {
+  buster: string
+  timestamp: number
+  queryHash: string
+  queryKey: QueryKey
+  state: QueryState
+}
+
+export type PersistRetryer = (props: {
+  persistedQuery: PersistedQuery
+  error: Error
+  errorCount: number
+}) => Promisable<PersistedQuery | undefined>
+
+export interface AsyncStorage {
+  getItem: (key: string) => Promise<string | undefined | null>
+  setItem: (key: string, value: string) => Promise<void>
+  removeItem: (key: string) => Promise<void>
+}
+
+export interface StoragePersisterOptions<QC extends QueryClient> {
+  // TODO: if we decide to move this to an API layer, we could make this work without passing queryClient
+  /**
+   * Query Client instance
+   */
+  queryClient: QC
+  /** The storage client used for setting and retrieving items from cache.
+   * For SSR pass in `undefined`.
+   */
+  storage: AsyncStorage | Storage | undefined | null
+  /**
+   * How to serialize the data to storage.
+   * @default `JSON.stringify`
+   */
+  serialize?: (client: PersistedQuery) => string
+  /**
+   * How to deserialize the data from storage.
+   * @default `JSON.parse`
+   */
+  deserialize?: (cachedString: string) => PersistedQuery
+  /**
+   * A unique string that can be used to forcefully invalidate existing caches,
+   * if they do not share the same buster string
+   */
+  buster?: string
+  /**
+   * The max-allowed age of the cache in milliseconds.
+   * If a persisted cache is found that is older than this
+   * time, it will be discarded
+   */
+  maxAge?: number
+}
+
+export function createPersister<T, QC extends QueryClient>(
+  queryFn: () => Promise<T>,
+  {
+    queryClient,
+    storage,
+    buster = '',
+    maxAge = 1000 * 60 * 60 * 24,
+    serialize = JSON.stringify,
+    deserialize = JSON.parse,
+  }: StoragePersisterOptions<QC>,
+) {
+  return async (context: QueryFunctionContext) => {
+    const queryHash = hashQueryKey(context.queryKey)
+    const queryState = queryClient.getQueryState(context.queryKey)
+
+    if (!queryState?.data && storage != null) {
+      const storedData = await storage.getItem(queryHash)
+      if (storedData) {
+        const persistedQuery = deserialize(storedData) as PersistedQuery
+
+        if (persistedQuery.timestamp) {
+          const expired = Date.now() - persistedQuery.timestamp > maxAge
+          const busted = persistedQuery.buster !== buster
+          if (expired || busted) {
+            await storage.removeItem(queryHash)
+          } else {
+            queryClient.getQueryCache().build(
+              queryClient,
+              {
+                queryKey: context.queryKey,
+                queryHash: queryHash,
+              },
+              persistedQuery.state,
+            )
+            return Promise.resolve(persistedQuery.state.data as T)
+          }
+        } else {
+          await storage.removeItem(queryHash)
+        }
+      }
+    }
+
+    const queryFnResult = await queryFn()
+
+    if (storage != null) {
+      // TODO: if we decide to move this to an API layer, we could make this work without additional timeout
+      setTimeout(() => {
+        const newState = queryClient.getQueryState(context.queryKey)
+
+        storage.setItem(
+          queryHash,
+          serialize({
+            state: newState!,
+            queryKey: context.queryKey,
+            queryHash: queryHash,
+            timestamp: Date.now(),
+            buster: buster,
+          }),
+        )
+      }, 0)
+    }
+
+    return Promise.resolve(queryFnResult)
+  }
+}
--- a/examples/vue/persister/package.json
+++ b/examples/vue/persister/package.json
@@ -8,10 +8,12 @@
     "serve": "vite preview"
   },
   "dependencies": {
-    "vue": "3.2.41",
-    "@tanstack/vue-query": "^4.14.1",
+    "@tanstack/query-core": "^4.14.1",
     "@tanstack/query-persist-client-core": "^4.14.1",
-    "@tanstack/query-sync-storage-persister": "^4.14.1"
+    "@tanstack/query-sync-storage-persister": "^4.14.1",
+    "@tanstack/vue-query": "^4.14.1",
+    "idb-keyval": "^6.2.0",
+    "vue": "3.2.41"
   },
   "devDependencies": {
     "@vitejs/plugin-vue": "3.2.0",
--- a/examples/vue/persister/src/Post.vue
+++ b/examples/vue/persister/src/Post.vue
@@ -1,8 +1,10 @@
 <script lang="ts">
+import { get, set, del } from 'idb-keyval'
 import { defineComponent } from 'vue'
-import { useQuery } from '@tanstack/vue-query'
+import { useQuery, useQueryClient } from '@tanstack/vue-query'
 
 import { Post } from './types'
+import { createPersister } from './persister'
 
 const fetcher = async (id: number): Promise<Post> =>
   await fetch(`https://jsonplaceholder.typicode.com/posts/${id}`).then(
@@ -19,9 +21,18 @@ export default defineComponent({
   },
   emits: ['setPostId'],
   setup(props) {
+    const queryClient = useQueryClient()
+
     const { isLoading, isError, isFetching, data, error } = useQuery({
       queryKey: ['post', props.postId],
-      queryFn: () => fetcher(props.postId),
+      queryFn: createPersister(() => fetcher(props.postId), {
+        storage: {
+          getItem: (key: string) => get(key),
+          setItem: (key: string, value: string) => set(key, value),
+          removeItem: (key: string) => del(key),
+        },
+        queryClient,
+      }),
     })
 
     return { isLoading, isError, isFetching, data, error }
--- a/examples/vue/persister/src/Posts.vue
+++ b/examples/vue/persister/src/Posts.vue
@@ -1,8 +1,9 @@
 <script lang="ts">
 import { defineComponent } from 'vue'
-import { useQuery } from '@tanstack/vue-query'
+import { useQuery, useQueryClient } from '@tanstack/vue-query'
 
 import { Post } from './types'
+import { createPersister } from './persister'
 
 const fetcher = async (): Promise<Post[]> =>
   await fetch('https://jsonplaceholder.typicode.com/posts').then((response) =>
@@ -19,9 +20,14 @@ export default defineComponent({
   },
   emits: ['setPostId'],
   setup() {
+    const queryClient = useQueryClient()
+
     const { isLoading, isError, isFetching, data, error, refetch } = useQuery({
       queryKey: ['posts'],
-      queryFn: fetcher,
+      queryFn: createPersister(fetcher, {
+        storage: localStorage,
+        queryClient,
+      }),
     })
 
     return { isLoading, isError, isFetching, data, error, refetch }
--- a/examples/vue/persister/src/main.ts
+++ b/examples/vue/persister/src/main.ts
@@ -9,17 +9,17 @@ const vueQueryOptions: VueQueryPluginOptions = {
   queryClientConfig: {
     defaultOptions: {
       queries: {
-        gcTime: 1000 * 60 * 60 * 24,
-        staleTime: 1000 * 60 * 60 * 24,
+        cacheTime: 1000 * 60 * 60 * 24,
+        staleTime: 1000 * 10,
       },
     },
   },
-  clientPersister: (queryClient) => {
-    return persistQueryClient({
-      queryClient,
-      persister: createSyncStoragePersister({ storage: localStorage }),
-    })
-  },
+  // clientPersister: (queryClient) => {
+  //   return persistQueryClient({
+  //     queryClient,
+  //     persister: createSyncStoragePersister({ storage: localStorage }),
+  //   })
+  // },
 }
 
 createApp(App).use(VueQueryPlugin, vueQueryOptions).mount('#app')
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -842,17 +842,21 @@ importers:
 
   examples/vue/persister:
     specifiers:
+      '@tanstack/query-core': ^4.14.1
       '@tanstack/query-persist-client-core': ^4.14.1
       '@tanstack/query-sync-storage-persister': ^4.14.1
       '@tanstack/vue-query': ^4.14.1
       '@vitejs/plugin-vue': 3.2.0
+      idb-keyval: ^6.2.0
       typescript: 4.8.4
       vite: 3.2.2
       vue: 3.2.41
     dependencies:
+      '@tanstack/query-core': 4.24.4
       '@tanstack/query-persist-client-core': 4.24.4
       '@tanstack/query-sync-storage-persister': 4.24.4
       '@tanstack/vue-query': 4.24.4_vue@3.2.41
+      idb-keyval: 6.2.0
       vue: 3.2.41
     devDependencies:
       '@vitejs/plugin-vue': 3.2.0_vite@3.2.2+vue@3.2.41
@@ -11308,6 +11312,12 @@ packages:
     dependencies:
       safer-buffer: 2.1.2
 
+  /idb-keyval/6.2.0:
+    resolution: {integrity: sha512-uw+MIyQn2jl3+hroD7hF8J7PUviBU7BPKWw4f/ISf32D4LoGu98yHjrzWWJDASu9QNrX10tCJqk9YY0ClWm8Ng==}
+    dependencies:
+      safari-14-idb-fix: 3.0.0
+    dev: false
+
   /ieee754/1.2.1:
     resolution: {integrity: sha512-dcyqhDvX1C46lXZcVqCpK+FtMRQVdIMN6/Df5js2zouUsqG7I6sFxitIC+7KYK29KdXOLHdu9zL4sFnoVQnqaA==}
 
@@ -15412,6 +15422,9 @@ packages:
     dependencies:
       mri: 1.2.0
     dev: true
+
+  /safari-14-idb-fix/3.0.0:
+    resolution: {integrity: sha512-eBNFLob4PMq8JA1dGyFn6G97q3/WzNtFK4RnzT1fnLq+9RyrGknzYiM/9B12MnKAxuj1IXr7UKYtTNtjyKMBog==}
 
   /safe-buffer/5.1.2:
     resolution: {integrity: sha512-Gd2UZBJDkXlY7GbJxfsE8/nvKkUEU1G38c1siN6QP6a9PT9MmHB8GnpscSmMJSoF8LOIrt8ud/wPtojys4G6+g==}


examples/vue/persister/src/persister.ts
```diff
@@ -0,0 +1,128 @@
...
+    const queryHash = hashQueryKey(context.queryKey)
+    const queryState = queryClient.getQueryState(context.queryKey)
+
+    if (!queryState?.data && storage != null) {
+      const storedData = await storage.getItem(queryHash)
+      if (storedData) {
+        const persistedQuery = deserialize(storedData) as PersistedQuery
+
+        if (persistedQuery.timestamp) {
+          const expired = Date.now() - persistedQuery.timestamp > maxAge
```
TkDodo, line 85: I think we would also need to check if we have data in the cache already that is newer than data that we have in storage, no? For example, if dat is put into the cache via `queryClient.setQueryData` in the meantime. This check is basically built into the `hydrate()` function, so maybe we should be using that one instead of calling `.build()` directly?

also - that data wouldn't be persisted, as we are strictly bound to the `queryFn` right now, so if the queryFn doesn't run, there's no persistence. This can also happen if you prefetch on the server, then put data into the cache via `hydrate`.

examples/vue/persister/src/persister.ts
```diff
@@ -0,0 +1,128 @@
...
+          } else {
+            queryClient.getQueryCache().build(
+              queryClient,
+              {
+                queryKey: context.queryKey,
+                queryHash: queryHash,
+              },
+              persistedQuery.state,
+            )
+            return Promise.resolve(persistedQuery.state.data as T)
```
KarthikMAM, line 98: If you do a return here... wouldn't that mean if data is in cache it will never be refreshed breaking `stale-while-revalidate` use-cases?

TkDodo: this branch is only executed if we don't have data yet, see [this line](https://github.com/TanStack/query/pull/5125/files#diff-2a40d1fa40a7ae31330acd1ba80cfde7a697273dd9ae0d159bcba7c91e5a3959R79). It's basically a way to lazily seed the cache

examples/vue/persister/src/persister.ts
```diff
@@ -0,0 +1,128 @@
...
+          }
+        } else {
+          await storage.removeItem(queryHash)
+        }
+      }
+    }
+
+    const queryFnResult = await queryFn()
```
TkDodo, line 106: we need to pass the `context` to the `queryFn`.

examples/vue/persister/src/persister.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    if (storage != null) {
+      // TODO: if we decide to move this to an API layer, we could make this work without additional timeout
+      setTimeout(() => {
```
TkDodo, line 110: could we not just attach it to the promise?

```
return Promise.resolve(queryFnResult).then(...)
```

?

TkDodo (commented):
I like that this solution just works with the current API, basically as a higher order query function. If we do this, we should likely implement it with `hydrate` and `dehydrate` though, for the issues I mentioned in the comments.

The disadvantage is that you _cannot_ set this globally - you have to wrap every QueryFunction with it. It's also not doable for folks using code generation (graphql-codegen) or other frameworks like trpc because you don't provide a `queryFn` manually there. We could argue we don't care because it's the problem of libs providing those abstractions...

Not sure how to solve this most elegantly. One way would be to provide the option to set a `queryFnWrapper`, that must basically return a QueryFunction, and that we invoke around it. Could be a global option on the QueryCache, something like:

```
new QueryCache({
  queryFnWrapper: createPersister(opts)
})
```

or, it could be a `useQuery` default option:

```
new QueryClient({
  defaultOptions: {
    queries: {
      queryFnWrapper: createPersister(opts)
    }
  }
})
```

what do you think @DamianOsipiuk ?

TkDodo:
@DamianOsipiuk another thing that comes to my mind is that for infinite queries, we execute the `queryFunction` multiple times (once for each page), but ideally, we'd only write to / read from the persisters once for the whole cache entry ...

examples/vue/persister/src/persister.ts
DamianOsipiuk: Correct me if i'm wrong, but query subscription and in turn `queryFn` would not be run if we are `restoring`?

https://github.com/TanStack/query/blob/beta/packages/react-query/src/useBaseQuery.ts#L62-L64
```ts
const unsubscribe = isRestoring
  ? () => undefined
  : observer.subscribe(notifyManager.batchCalls(onStoreChange))
```


This means that when we are in this piece of logic it's been already restored and we have something in the cache.
So initial condition for this branch would prevent restoring from persister and skip to `queryFn`:
https://github.com/TanStack/query/pull/5125/files/9ed0d9a942151772b5ac0f07ca9da539bde55824#diff-2a40d1fa40a7ae31330acd1ba80cfde7a697273dd9ae0d159bcba7c91e5a3959R79
```ts
const queryState = queryClient.getQueryState(context.queryKey)

if (!queryState?.data && storage != null) {
//      ^ do not restore if we already have something
```

examples/vue/persister/src/persister.ts
DamianOsipiuk: A good question here is if we should
- `restore` and still `fetch from server`
- `restore` and `wait for next fetch trigger`

Cause i think in this case we would just restore during initial component mount and skip `queryFn` which is not entirely correct as we would like to get latest data while revalidating. So instead of return `setQueryData` in here?

examples/vue/persister/src/persister.ts
DamianOsipiuk: I have used `setTimeout` to make sure that result of `queryFn` will be populated in the queryCache for easier serialization of persister entry. But we can build it manually i guess.

packages/query-core/src/types.ts
```diff
@@ -118,6 +124,7 @@ export interface QueryOptions<
   networkMode?: NetworkMode
   gcTime?: number
   queryFn?: QueryFunction<TQueryFnData, TQueryKey, TPageParam>
+  persister?: QueryPersister<TQueryFnData, TQueryKey>
```
DamianOsipiuk, line 127: @TkDodo Regarding setting persister as default globally, we could use this new option, and instead of wrapping `queryFn` just provide it on the side when necessary, or globally.

examples/vue/persister/src/persister.ts
TkDodo: > but query subscription and in turn queryFn would not be run if we are restoring?

yes, that's true.

> This means that when we are in this piece of logic it's been already restored and we have something in the cache.

okay yeah that's also correct. It means that restoring would only work on empty caches though?

I'm just wondering why we wouldn't use the `hydrate` utility that we have already, which does a check that is slightly similar: It will always restore if data is newer, so that after hydration, you'll have the most up-to-date data, no matter where it came from ..,.

packages/query-core/src/query.ts
```diff
@@ -392,6 +392,13 @@ export class Query<
...
+      if (this.options.persister) {
+        return this.options.persister(
+          this.options.queryFn,
+          queryFnContext as QueryFunctionContext<TQueryKey>,
+        )
+      }
+
       return this.options.queryFn(
         queryFnContext as QueryFunctionContext<TQueryKey>,
       )
```
TkDodo, line 404: note: I think we would also need something for `infiniteQueries`, as they execute the QueryFn in a different branch:

https://github.com/TanStack/query/blob/8d5e08fdfffc42cadddd61865bc12f754b4b3564/packages/query-core/src/infiniteQueryBehavior.ts#L39-L45

examples/vue/persister/src/persister.ts
DamianOsipiuk: We are inside a `queryFn` here, so we can't set it directly in queryCache, cause it will not render it until `queryFn` resolves. That is why we resolve the Promise with data from the persister.

As for
> It means that restoring would only work on empty caches though?

It there any way that `persisted data` will be newer than something that got added to the cache after client was instantiated in browser?
I do not think there will be any way that could happen.

Additionally by skipping this logic when we have data in `queryCache` already, we would not read it on every `queryFn` run, which is a perf improvement.

packages/query-core/src/query.ts
DamianOsipiuk: I think i have found a proper place to hook it up for infinite queries.
The only problem that is left is whenever you try to use new `createPersister` and `context` in `queryFn` types break for `queryFn`.

examples/vue/persister/src/persister.ts
TkDodo: > It there any way that persisted data will be newer than something that got added to the cache after client was instantiated in browser?

only if you get it into the storage by other means, so yeah, I get your point, and also 👍 about not running it in every queryFn

examples/vue/persister/src/persister.ts
TkDodo: > restore and still fetch from server
restore and wait for next fetch trigger

this should depend on the fact if the data that was restored is still considered `fresh` or not - depending on `staleTime`. If the data is considered outdated - call the `queryFn` immediately (but still return stale data). If it's fresh, just return it ...

packages/query-persist-client-core/src/createPersister.ts
```diff
@@ -0,0 +1,147 @@
...
+}: StoragePersisterOptions<QC>) {
+  return async (
+    queryFn: (context: QueryFunctionContext) => T | Promise<T>,
+    context: QueryFunctionContext,
+  ) => {
+    const queryHash = hashKey(context.queryKey)
+    const queryState = queryClient.getQueryState(context.queryKey)
+
+    // Try to restore only if we do not have any data in the cache and we have persister defined
+    if (!queryState?.data && storage != null) {
```
TkDodo, line 87: I think we need to check for `queryState?.data !== undefined` explicitly, because having falsy values in `data` for a successful query is a legit use-case.

packages/query-persist-client-core/src/createPersister.ts
```diff
@@ -0,0 +1,147 @@
...
+  queryHash: string
+  queryKey: QueryKey
+  state: QueryState
+}
+
+export interface AsyncStorage {
+  getItem: (key: string) => Promise<string | undefined | null>
+  setItem: (key: string, value: string) => Promise<void>
+  removeItem: (key: string) => Promise<void>
+}
```
TkDodo, line 21: we already have this interface define in `query-async-storage-persister`:

https://github.com/TanStack/query/blob/1d3ee672d2a1536e3463243813c91902aa97d973/packages/query-async-storage-persister/src/index.ts#L8-L12

I think the best way forward would be to have the async-storage-persister package re-use the one from the core you've defined here ?

packages/query-persist-client-core/src/createPersister.ts
```diff
@@ -0,0 +1,147 @@
...
+     queryKey: ['myKey'],
+     queryFn: fetcher,
+     persister: createPersister({
+       storage: localStorage,
+       queryClient,
+     }),
+   })
+   ```
+ */
+export function createPersister<T, QC extends QueryClient>({
```
TkDodo, line 71: let's rename this to `experimental_createPersister` to be on the safe side.

packages/query-persist-client-core/src/createPersister.ts
```diff
@@ -0,0 +1,147 @@
...
+            dataUpdatedAt: Date.now(),
+            status: 'success',
+            error: null,
+            errorUpdateCount: 0,
+            errorUpdatedAt: 0,
+            fetchFailureCount: 0,
+            fetchFailureReason: null,
+            fetchMeta: null,
+            fetchStatus: 'idle',
+            isInvalidated: false,
```
TkDodo, line 135: I don't think these values are necessarily correct. For every fetch other than the first, it would re-set `dataUpdateCount` to `0`. I think we would need a way to get the actual values from the current Query.

And while we can get the current Query values easily, they will not have the most updated values yet, because we are technically still inside the QueryFn, so the last transition hasn't happened yet 🤔

packages/query-persist-client-core/src/createPersister.ts
```diff
@@ -0,0 +1,147 @@
...
+                exact: true,
+              })
+            }, 0)
+            // We must resolve the promise here, as otherwise we will have `loading` state in the app until `queryFn` resolves
+            return Promise.resolve(persistedQuery.state.data as T)
+          }
+        } else {
+          await storage.removeItem(queryHash)
+        }
+      }
```
TkDodo, line 112: if anything fails in this path (like `getItem` not working for some reason), we'll put the Query into an error state, and the retry will try that again, but we will never reach the actual `queryFn`.

I think we'd want a try/catch around this block - similar to how its done in `persist.ts`:

https://github.com/TanStack/query/blob/1d3ee672d2a1536e3463243813c91902aa97d973/packages/query-persist-client-core/src/persist.ts#L84-L107

packages/query-persist-client-core/src/createPersister.ts
```diff
@@ -0,0 +1,147 @@
...
+        if (persistedQuery.timestamp) {
+          const expired = Date.now() - persistedQuery.timestamp > maxAge
+          const busted = persistedQuery.buster !== buster
+          if (expired || busted) {
+            await storage.removeItem(queryHash)
+          } else {
+            // Just after restoring we want to get fresh data from the server
+            // Maybe add an option for this?
+            setTimeout(() => {
+              queryClient.invalidateQueries({
```
TkDodo, line 101:
```suggestion
              queryClient.refetchQueries({
```

packages/query-core/src/infiniteQueryBehavior.ts
```diff
@@ -114,6 +114,17 @@ export function infiniteQueryBehavior<TQueryFnData, TError, TData, TPageParam>(
...
+        context.fetchFn = () => {
+          return context.options.persister?.(fetchFn as any, {
+            queryKey: context.queryKey,
+            meta: context.options.meta,
+            signal: context.signal,
+          })
+        }
+      } else {
+        context.fetchFn = fetchFn
+      }
```
TkDodo, line 127: this is a good solution because it will only wrap the whole `fetchFn` once, and not the fetch we do for every page 👍

packages/query-persist-client-core/src/createPersister.ts
```diff
@@ -0,0 +1,147 @@
...
+            errorUpdatedAt: 0,
+            fetchFailureCount: 0,
+            fetchFailureReason: null,
+            fetchMeta: null,
+            fetchStatus: 'idle',
+            isInvalidated: false,
+          },
+          queryKey: context.queryKey,
+          queryHash: queryHash,
+          timestamp: Date.now(),
```
TkDodo, line 139: why do we need an extra `timestamp` if we have `state.dataUpdatedAt` ?

packages/query-persist-client-core/src/createPersister.ts
```diff
@@ -0,0 +1,147 @@
...
+        const persistedQuery = deserialize(storedData)
+
+        if (persistedQuery.timestamp) {
+          const expired = Date.now() - persistedQuery.timestamp > maxAge
+          const busted = persistedQuery.buster !== buster
+          if (expired || busted) {
+            await storage.removeItem(queryHash)
+          } else {
+            // Just after restoring we want to get fresh data from the server
+            // Maybe add an option for this?
```
TkDodo, line 99: I think we'd want to compare the timestamp of the persisted Query with the current date by using the `timeUntilStale` util and the `staleTime` of the Query:

https://github.com/TanStack/query/blob/13b17ee45de3299dd3d3a2dfc3d23ad2f024882a/packages/query-core/src/utils.ts#L88-L90

with that, we would know if we should refetch or not, based on weather the data we restore from storage is considered fresh or not ...

Question is really where to get the `staleTime` from, because that's an observer level setting, and we don't really have access to the QueryObserver from here 🤔 . Or do we ?

TkDodo (commented):
This looks great already, thank you so much. I started to remember what we wanted this util for and left some more comments:

emzet93:
@DamianOsipiuk @TkDodo this feature looks amazing and will take persisting data with this lib to the next level. I have one question about this implementation:
As far as I understand, it reads persisted data on demand, on the query level using a wrapper over `queryFn`?
If so, are those two concerns valid?
- query function actually needs to be called to bring persisted data into query cache, right? So if we use async storage (or even sync too) for persistence, query will end up in hard loading state always, even if we have persisted data for it? It will just resolve much faster because it reads from storage and not from network?
- after app is reopened, calling `getQueryData` on client won't return persisted data if query function was not called by some observer in the meantime?

Again - it's amazing work and I would like to understand what is happening here better 🙏

TkDodo:
@emzet93 

> query function actually needs to be called to bring persisted data into query cache

yes, there will be a brief moment where the Query is in hard loading state. This is almost not different to how `PersistQueryClient` works though. There, restoring also runs in an `effect`, so there will be at least one loading render cycle. The only difference is that we are not "fetching", so the states will be:
- `status: 'pending', fetchStatus: 'idle'`
while with the persist wrapper over `useQuery`, you would be in:
- `status: 'pending', fetchStatus: 'fetching'`

We can think about putting the query into `idle` fetchStatus while we are restoring, too to distinguish from a "real" fetch.

> after app is reopened, calling getQueryData on client won't return persisted data if query function was not called by some observer in the meantime?

that's correct, but it's also on purpose. The idea is to load queries into the in-memory cache on demand only, to avoid excessive memory usage. If you want it available after "page load", you should still use the `<PersistQueryClient>`.

Note that it's also possible to mix the two approaches: You can have most queries restore on page load with the `PersistQueryClient`, but have some (maybe heavy) queries be stored separately with the fine grained persisters.

emzet93:
> @emzet93
> 
> > query function actually needs to be called to bring persisted data into query cache
> 
> yes, there will be a brief moment where the Query is in hard loading state. This is almost not different to how `PersistQueryClient` works though. There, restoring also runs in an `effect`, so there will be at least one loading render cycle. The only difference is that we are not "fetching", so the states will be:
> 
> * `status: 'pending', fetchStatus: 'idle'`
>   while with the persist wrapper over `useQuery`, you would be in:
> * `status: 'pending', fetchStatus: 'fetching'`
> 
> We can think about putting the query into `idle` fetchStatus while we are restoring, too to distinguish from a "real" fetch.
> 
> > after app is reopened, calling getQueryData on client won't return persisted data if query function was not called by some observer in the meantime?
> 
> that's correct, but it's also on purpose. The idea is to load queries into the in-memory cache on demand only, to avoid excessive memory usage. If you want it available after "page load", you should still use the `<PersistQueryClient>`.
> 
> Note that it's also possible to mix the two approaches: You can have most queries restore on page load with the `PersistQueryClient`, but have some (maybe heavy) queries be stored separately with the fine grained persisters.

thanks for the reply! Currently I'm preventing rendering of my whole React Native app until hydration from storage is completed, which ensures that when I read sth from query client, hydration is completed. It won't be possible with grained persisters, however mixing two approaches could be a great way to go.

Also I'm not sure if read it properly: when grained persister saves the query data to the storage? Is after every successful refetch? Or is it saved in single batch together with all/global persister?

That's pretty important point from react native pov. Slow hydration is not that big deal imho - it's not that bad if user needs to wait half a second on app start. Saving data feels more critical. Now cache is saved as a whole after every change in query cache. It can be pretty expensive, especially on hermes engine which has serious issues with serialising data. Also it happens when user is actively using the app and it often results in visible FPS drops, so graining persisters could be a huge improvement here.

TkDodo:
> when grained persister saves the query data to the storage? Is it after every successful refetch? Or is it saved in single batch together with all/global persister?

it's after every fetch:

https://github.com/TanStack/query/pull/5125/files#diff-49e8b295893ae938f8fc349ec5c9d037944289d845c64c2ed9e38f9bbd63eba1R116-R120

> Now cache is saved as a whole after every change in query cache. It can be pretty expensive, especially on hermes engine which has serious issues with serialising data.

yes, this is one of the issues we're trying to solve here. The saving and restoring always works on query level, so every one is independent.

emzet93:
> https://github.com/TanStack/query/pull/5125/files#diff-49e8b295893ae938f8fc349ec5c9d037944289d845c64c2ed9e38f9bbd63eba1R116-R120

Perfect! Thanks for the explanations

emzet93:
@TkDodo:
also what do you think about adding the api allowing to restore persisted data to query cache for specific query, so it could be called in advance and allow skipping hard loading state when persisted query is mounted? I think that it should be possible if we pass a persister instance to it, so based on example from this PR it would be sth like:
```

const persister = createPersister({
        storage: {
          getItem: (key: string) => get(key),
          setItem: (key: string, value: string) => set(key, value),
          removeItem: (key: string) => del(key),
        },
        queryClient,
      });
      

await queryClient.restoreQuery(['posts'], persister);
```
then once `useQuery` for this key is mounted, it will have data already in first render.

Do you guys see any issues/limitations here?

If not then I would be happy to contribute and try to implement it on top of this PR once it's ready.

TkDodo:
yeah I think this should be possible, but you'd need exact keys, not fuzzy ones. @DamianOsipiuk what do you think ?

emzet93:
Ah yeah, this is true. But it still can be useful for queries with static keys I think.

TkDodo:
I don't think we know all stored keys because they would be stored under different locations. It's not like we're persisting them in the same item. To read from an external storage like localstorage, we need the exact key, because the queryHash (computed from the queryKey) is used as identifier:

```
const storedData = await storage.getItem(queryHash)
```

packages/query-core/src/types.ts
```diff
@@ -118,6 +132,7 @@ export interface QueryOptions<
   networkMode?: NetworkMode
   gcTime?: number
   queryFn?: QueryFunction<TQueryFnData, TQueryKey, TPageParam>
+  persister?: QueryPersister<TQueryFnData, TQueryKey, TPageParam>
```
DamianOsipiuk, line 135: @emzet93 @TkDodo Just want to move the discussion to a thread, instead of separate comments. (GitLab has that covered 🙄)

> also what do you think about adding the api allowing to restore persisted data to query cache for specific query, so it could be called in advance and allow skipping hard loading state when persisted query is mounted? I think that it should be possible if we pass a persister instance to it, so based on example from this PR it would be sth like:

I believe we could add it. Even better `persister` could be passed as a `default option` for query, so the only thing you would need to specify is `queryKey`.
The issue i see is - how do we skip `queryFn` and `saving to storage` when restoring this way? Cause we only need the first part of `createPersister`?

We could potentially extract `restoring` to a separate function, so you could use only that piece?
Thoughts?

packages/query-persist-client-core/src/createPersister.ts
DamianOsipiuk: Replicated from other persister. But yeah, we don't i guess

packages/query-persist-client-core/src/createPersister.ts
```diff
@@ -0,0 +1,158 @@
...
+     queryFn: fetcher,
+     persister: createPersister({
+       storage: localStorage,
+       queryClient,
+     }),
+   })
+   ```
+ */
+export function experimental_createPersister<T, QC extends QueryClient>({
+  queryClient,
```
DamianOsipiuk, line 71: @TkDodo I think we would need a way to get access to `queryClient` somehow from `Query` context to not require it being passed by the user.
Otherwise how would you provide it as a default query option - chicken and egg problem.
```ts
new QueryClient({
  defaultOptions: {
    queries: {
      staleTime: 3600,
      persister: createPersister({
        storage: {
          getItem: (key: string) => get(key),
          setItem: (key: string, value: string) => set(key, value),
          removeItem: (key: string) => del(key),
        },
        queryClient, // ? but... how?
      }),
    }
  },
})
```

packages/query-core/src/types.ts
emzet93: 'Even better persister could be passed as a default option for query, so the only thing you would need to specify is queryKey.'

Setting persister as default will probably cause every single query to be persisted, right (if not overwritten on query level)? I imagine that it might not always be desired, so we would still need possibility to pass it directly to 'restoreQuery', in case persisted is not defined on defaults.

'We could potentially extract restoring to a separate function, so you could use only that piece?'

That sounds good to me 👍

'I think this would belong to userland. User can get all keys of storage and call multiple times restore function?'

Agree that it should be userland thing. And I had the same idea for calling multiple restore. But @TkDodo made me aware that getting all keys from storage will give us all query hashes, not query keys. Restore should probably takes key, not hash, so it won't help us because I assume that we can convert hash back to key.

packages/query-persist-client-core/src/createPersister.ts
TkDodo: I think we can:

- get the `query` from the `QueryClient` (by finding it via the `queryKey`)
- then, loop through all observers and take the smallest staleTime we find

similar to how we do it when there's a window focus refetch:

https://github.com/TanStack/query/blob/9c721e548f500be924da8ef6da041a919f513140/packages/query-core/src/query.ts#L273-L278

- if we have no observer, because we wind up executing the queryFunction via prefetching or so, there won't be a refetch.

I think this would be fine?

packages/query-core/src/types.ts
TkDodo: > We could potentially extract restoring to a separate function, so you could use only that piece? Thoughts?

Yes 👍 

> Setting persister as default will probably cause every single query to be persisted, right (if not overwritten on query level)?

Maybe we could make persister be an optional function? This is a common pattern in RQ, something like:

```
persister: Persister | undefined | ((query: Query) => Persister | undefined)
```

with that, you could define a global persister for all queries that still excludes certain queries, e.g. depending on the `queryKey` or depending on the `meta` field.

packages/query-persist-client-core/src/createPersister.ts
TkDodo: true. it would work if it were a function I think:

```
const queryClient: QueryClient = new QueryClient({
  defaultOptions: {
    queries: {
      staleTime: 3600,
      persister: () => createPersister({
        storage: {
          getItem: (key: string) => get(key),
          setItem: (key: string, value: string) => set(key, value),
          removeItem: (key: string) => del(key),
        },
        queryClient,
      }),
    }
  },
})
```

this would go well with my proposal of making persisters an optional function.

---

Oh, I just saw that we also expose a `setDefaultOptions` function from the `queryClient`:

https://github.com/TanStack/query/blob/9c721e548f500be924da8ef6da041a919f513140/packages/query-core/src/queryClient.ts#L617-L619

so you can really do:

```
queryClient.setDefaultOptions({
  queries: {
    staleTime: 3600,
    persister: createPersister({
      storage: {
        getItem: (key: string) => get(key),
        setItem: (key: string, value: string) => set(key, value),
        removeItem: (key: string) => del(key),
      },
      queryClient,
    }),
  }
})
```

and that should work without problems

packages/query-core/src/types.ts
emzet93: Yeah that sounds perfect 👍
Two questions though:
- it should still be possible to pass optional `persister` to `queryClient.restoreQuery` in case no default persister is set? Or should setting default persister be mandatory in order to use `restoreQuery`?
- if we set default persister, it still should be possible to disable it for specific query by setting `persister: null/undefined` on `useQuery` level, right?

TkDodo: > it should still be possible to pass optional persister to queryClient.restoreQuery in case no default persister is set? Or should setting default persister be mandatory in order to use restoreQuery?

yes this should work.

> if we set default persister, it still should be possible to disable it for specific query by setting persister: null/undefined on useQuery level, right?

yes, the options are shallow merged:

https://github.com/TanStack/query/blob/13b17ee45de3299dd3d3a2dfc3d23ad2f024882a/packages/query-core/src/queryClient.ts#L469-L474

so if you have `persister: undefined`, it would overwrite it.

emzet93: Great! So it sounds like a good way to go.

packages/query-persist-client-core/src/createPersister.ts
DamianOsipiuk: I managed to solve it differently

TkDodo (approved):
I think this looks very good now 👍

packages/query-persist-client-core/src/createPersister.ts
TkDodo: nice, I like it 🙌

packages/query-persist-client-core/src/createPersister.ts
```diff
@@ -0,0 +1,161 @@
...
+  /**
+   * Prefix to be used for storage key.
+   * Storage key is a combination of prefix and query hash in a form of `prefix-queryHash`.
+   * @default 'tanstack-query'
+   */
+  prefix?: string
+  /**
+   * Filter function returning whether current query should be run trough persister.
+   */
+  queryFilter?: (queryKey: QueryKey, query: Query) => boolean
```
DamianOsipiuk, line 57: @TkDodo Since `experimental_createPersister` already returns `persister` function it would be hard to add predicate option, so instead i have added `queryFilter` as an option that will be evaluated before `restoring` or `persisting`

docs/react/plugins/createPersister.md
```diff
@@ -0,0 +1,119 @@
...
+    },
+  },
+})
+```
+
+## API
+
+### `experimental_createPersister`
+
+Call this function to create a `persister` that you can pass to `defaultOptions` of `QueryClient` or to any `useQuery` hook.
```
DamianOsipiuk, line 53: added some docs

DamianOsipiuk:
> I think this looks very good now 👍

@TkDodo If you could take one last look, we should be ready to merge.

packages/query-persist-client-core/src/createPersister.ts
TkDodo: that's a good idea, but why not support full QueryFilters?

https://github.com/TanStack/query/blob/13b17ee45de3299dd3d3a2dfc3d23ad2f024882a/packages/query-core/src/utils.ts#L13

TkDodo: what do you think?

https://github.com/TanStack/query/pull/5125/commits/36112eaa84714b1cfc1493a9f152d7282df34c6c

DamianOsipiuk: Nice 👍

TkDodo:
@badsyntax the `experimental_createPersister` is a wrapper around the `queryFn`, so it needs the `queryFn` to run in order to restore from storage. That doesn't happen per default when you are offline, because `networkMode` is `'online'`.

What we have done is to actually set `networkMode: 'offlineFirst'` per default if you pass a persister:

https://github.com/TanStack/query/blob/1dc5758518594c82f5399deeb30e393b27d0a91a/packages/query-core/src/queryClient.ts#L513-L518

but if you pass `networkMode: 'online'` manually, there's nothing we can do. I'd say don't do that 😅

badsyntax:
@TkDodo thanks for the quick reply. ok that makes sense! i'll keep it as `offlineFirst`. The only real difference i see now is that while offline those requests are still sent with `offlineFirst`. I guess this doesn't matter when network is really offline, but does present some problems when "mocking" offline state. (EG in my React Native app you can simulate offline mode by simply setting `onlineManager.setOnline(false)` but this approach will no longer work).

TkDodo:
yes, that's what `offlineFirst` does, but that's also what makes the `createPersister` work: A request is made even though you are offline. That request is "intercepted" by the persister, and it will try to read from storage first. So if you want to get data from storage while being offline, you need at least `offlineFirst`.

PR #5131, opened 2023-03-14
Feature/unify dehydrate options

TkDodo:


Ephem (commented):
As we discussed I think we could consider renaming the options `filterQueries` and `filterMutations`? Other than that this looks good to me besides the migration docs listing wrong exports.

Ephem:
Latest notification made me realise I had totally missed your comment here, sorry! I think those are valid points, sounds good to me!

docs/react/reference/hydration.md
```diff
@@ -50,7 +47,7 @@ const dehydratedState = dehydrate(queryClient, {
 
 ### limitations
 
-Some storage systems (such as browser [Web Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API)) require values to be JSON serializable. If you need to dehydrate values that are not automatically serializable to JSON (like `Error` or `undefined`), you have to serialize them for yourself. Since only successful queries are included per default, to also include `Errors`, you have to provide `shouldDehydrateQuery`, e.g.:
+Some storage systems (such as browser [Web Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API)) require values to be JSON serializable. If you need to dehydrate values that are not automatically serializable to JSON (like `Error` or `undefined`), you have to serialize them for yourself. Since only successful queries are included per default, to also include `Errors`, you have to provide `shouldDehydrateQuery `, e.g.:
```
TkDodo, line 50:
```suggestion
Some storage systems (such as browser [Web Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API)) require values to be JSON serializable. If you need to dehydrate values that are not automatically serializable to JSON (like `Error` or `undefined`), you have to serialize them for yourself. Since only successful queries are included per default, to also include `Errors`, you have to provide `shouldDehydrateQuery`, e.g.:
```

PR #5134, opened 2023-03-15
Local Task Caching for all Revelant tasks

ZackDeRose:
## What this does

+ Makes all relevant tasks cacheable
+ From the perspective of calling the existing pnpm scripts, all scripts should result in the same behavior as before; with the exceptions that:
  1. Nx is now being used as the task runner 
  2. Some commands that were sequential are now parallelized where it made sense to parallelize
  3. Some commands are cacheable where it made sense to cache task results


nx.json
```diff
@@ -3,11 +3,24 @@
...
+        ]
       }
     }
   },
   "defaultBase": "main",
+  "pluginsConfig": {
+    "@nrwl/js": {
+      "analyzeSourceFiles": false
+    }
+  },
```
ZackDeRose, line 23: *short explanation*: this exlicitely states our depency graph strategy - which is to analyze deps based on package.json files. This was the behavior we are already using.

*long explanation*: I added the `@nrwl/workspace` briefly to this PR, and while I ended up not needing it, it appears that this changes the default behavior to determine deps by source file analysis. I want to leave this here so this isn't a surprise in case we end up introducing other nx plugins later on!

nx.json
```diff
@@ -34,6 +51,33 @@
...
+        "{projectRoot}/build/.tsbuildinfo"
+      ],
+      "inputs": ["default", "^public"],
+      "dependsOn": ["^test:types"]
+    },
+    "build:types": {
+      "outputs": ["{projectRoot}/build"],
+      "inputs": ["default", "^public"],
+      "dependsOn": ["^build:types"]
+    },
```
ZackDeRose, line 70: Interestingly, `test:types` and `build:types` scripts tend to do the same things:

`test:types`: `tsc`
`build:types`: `tsc --build`

If I'm not mistaken, the build option is true by default, so these are equivalent. If we want to not write to disk on the `test:types` I think we can use: `test:types --build=false`, but we'll need to make sure that the builds we depend on happen first for this to work correctly!

ZackDeRose: @TkDodo - if we want to make the adjustment above, can you create a new issue for this? And feel free to assign to me.

I'd like for this PR to be a like-for-like as much as possible so we don't change any behaviors!

package.json
```diff
@@ -6,23 +6,30 @@
...
     "cipublish": "ts-node scripts/publish.ts",
     "validatePackages": "ts-node scripts/validate-packages.ts"
   },
+  "nx": {
+    "includedScripts": [
+      "test:format",
+      "build",
+      "test:build"
+    ]
+  },
```
ZackDeRose, line 32: Adding this property creates an "include list" of the `package.json` scripts that nx will include in it's task graph.

My intention here is actually to make sure we exclude:
+ `test:eslint`
+ `test:types`
+ `build`
+ `build:types`

so that calling these do not infinitely recurse when we call `pnpm build` for example!

project.json
```diff
@@ -0,0 +1,28 @@
+{
```
ZackDeRose, line 1: this file creates the concept of a "project" for nx at the source root, so we can define and run some root-level scripts

scripts/project.json
```diff
@@ -0,0 +1,8 @@
+{
```
ZackDeRose, line 1: this defines a `scripts` project in our workspace for nx inside the `scripts` directory.

With this, we can now call: `nx test:eslint scripts` to lint our scripts, and what's cooler, we can now just "add this to the pile" by calling `nx run-many --target=test:lint` inside of our package scripts/ci stuff :)

package.json
```diff
@@ -6,23 +6,30 @@
...
     "test:lib:dev": "pnpm --filter \"./packages/**\" run test:lib:dev",
-    "test:build": "pnpm run build && bundlewatch && pnpm run validatePackages",
-    "test:types": "pnpm --filter \"./packages/**\" run test:types",
-    "build": "rollup --config rollup.config.js && pnpm --filter \"./packages/**\" run build && pnpm run build:types",
-    "build:types": "pnpm --filter \"./packages/**\" run build:types",
+    "test:build": "nx test:build root",
+    "test:types": "nx run-many --target=test:types --parallel=5",
+    "build": "nx run-many --target=build --projects=root,@tanstack/eslint-plugin-query",
+    "build:types": "nx run-many --target=build:types --parallel=5",
     "watch": "concurrently --kill-others \"rollup --config rollup.config.js -w\" \"pnpm run build:types --watch\"",
```
ZackDeRose, line 19: We might be able to use `nx watch` here for the watch command - but I'll keep that separate from this pr

project.json
```diff
@@ -0,0 +1,28 @@
...
+    },
+    "test:build": {
+      "executor": "nx:run-commands",
+      "options": {
+        "commands": ["bundlewatch", "pnpm run validatePackages"],
+        "parallel": true
+      }
+    }
+  },
+  "implicitDependencies": ["**/*"]
```
ZackDeRose, line 27: This marks our `root` nx project as dependent on all packages. We use this to inform our caching - as you can see in the `nx.json` file, the `inputs` for the `build` target is the `public` files of all dependencies.

project.json
```diff
@@ -0,0 +1,28 @@
...
+      "options": {
+        "commands": [
+          "rollup --config rollup.config.js",
+          "nx run-many --target=build:types",
+          "nx build @tanstack/eslint-plugin-query",
+          "nx build @tanstack/svelte-query"
+        ],
+        "parallel": true
+      }
+    },
```
ZackDeRose, line 18: Build should do these 4 things, and it can actually do all 4 in independently if we want, so that's what we're doing here.

Interestingly `@tanstack/svelte-query` has it's own build step, but all other packages are built in 1 go with the `rollup` command it appears.

@TkDodo - let me know if there's a preference here, we might be able to get some boost by parallelizing the package builds and running them individually (this also allows the cache to be more granular - as it stands we need to "invalidate" the build cache for all projects whenever a single project's public code is touched 🥶

If we do want to break it up, let's do it in another PR! As I mention elsewhere, I'd like to keep this one as "like-for-like" as possible!

TkDodo (approved):

TkDodo:
I think I spotted an issue with the `test:format` task. I've formatted everything correctly, run the task. Then made a change in a tsx file, run the task again, but it just reads from the cache:

```
> nx test:format root


> nx run root:test:format  [existing outputs match the cache, left as is]
```

I then tried with `--skip-nx-cache` and the error showed up:

```
> nx test:format root "--skip-nx-cache"


> nx run root:"test:format"

> query@ prettier /Users/dodo/work/TanStackQuery
> prettier --plugin-search-dir . "{packages,examples}/**/src/**/*.{md,js,jsx,ts,tsx,json,vue,svelte}" "--check"
Checking formatting...
[warn] packages/react-query-devtools/src/devtools.tsx
 >  NX   ERROR: Something went wrong in run-commands - Command failed: pnpm run prettier --check
```

ZackDeRose:
I must've misconfigured the cache here - on it now!

ZackDeRose:
fix up: https://github.com/TanStack/query/pull/5147

^ demo video to show this working now :)

We should definitely be vigilant against false positives like this!

PR #5142, opened 2023-03-16
Update imports

ravshansbox:


TkDodo:
this PR goes backwards. The `react-query` repo is the old one, the `@tanstack/react-query` package is the correct one. You can see this on npm as well:

- ✅ https://www.npmjs.com/package/@tanstack/react-query
- ❌ https://www.npmjs.com/package/react-query

> You're looking at the v3 version of react-query. Starting with v4, react-query is now available as [@tanstack/react-query](https://www.npmjs.com/package/@tanstack/react-query)

PR #5143, opened 2023-03-16
adding the nx-cloud task-runner

ZackDeRose:
# What does it do?

+ sets up the Distributed cache, with the main goal being that we can benefit from Nx Cloud in CI!

See in action here:

Screen Shot 2023-03-16 at 11 00 59 PM

https://github.com/TanStack/query/actions/runs/4444450790/jobs/7802666221


nx.json
```diff
@@ -11,7 +11,8 @@
           "test:format",
           "test:build",
           "build"
-        ]
+        ],
+        "accessToken": "OTgzYjM0YmUtNzlmNi00MWFhLWIyZmEtODA0NTgwODNiZjY4fHJlYWQtd3JpdGU="
```
TkDodo, line 15: is that access token something we should provide as an env variable instead?

packages/query-core/src/tests/hydration.test.tsx
```diff
@@ -467,3 +467,4 @@ describe('dehydration and rehydration', () => {
     expect(hydrationCache.find(['string'])?.state.fetchStatus).toBe('idle')
   })
 })
+console.log('here')
```
TkDodo, line 470:
```suggestion

```

TkDodo:
@ZackDeRose I was just clicking `re-run` on the `tests` task after it ran successfully, so I was thinking that re-running it should result in a cache hit, but it didn't. Instead, it re-ran all the tests: https://github.com/TanStack/query/actions/runs/4466159848/jobs/7844081201

nx.json
ZackDeRose: Yes, I think what we normally recommend is creating a "public" key that only allows read access to include here.

Then, we'd create a private key (or possible a set of private keys) with read-write access. We'd set this in our github secrets and use it in CI, and optionally trusted maintainers could add a private key (either shared or one unique to them) and set it via an environment variable.

I'll confirm with the cloud team the recommendation for open-source repos tomorrow.

packages/query-core/src/tests/hydration.test.tsx
ZackDeRose: 🤦‍♂️ I'm sure this is me testing the cache-busting and accidentally commiting the change lol

ZackDeRose:
> @ZackDeRose I was just clicking `re-run` on the `tests` task after it ran successfully, so I was thinking that re-running it should result in a cache hit, but it didn't. Instead, it re-ran all the tests: https://github.com/TanStack/query/actions/runs/4466159848/jobs/7844081201

I just tried by creating an empty commit - and I was able to get the cache hits: https://github.com/TanStack/query/actions/runs/4466470752

I'm wondering if the rerun happened too soon for some reason - let me investigate a bit further tomorrow and I'll let you know what I find.

ZackDeRose:
Leaving for later: https://cloud.nx.app/orgs/6412ca9d1c251d000efa21ba/workspaces/6412c827e6da5d7b4a0b1fe3/compare-tasks?baseId=6418278434279d0aeae94330

Comparing to the previous run: https://cloud.nx.app/runs/Nzfif6k5wi
Screen Shot 2023-03-20 at 2 35 34 AM

Looks like the cache-miss was unexpected. I did run locally after blowing away my local cache as soon as the original task completed, and I got all remote cache hits locally. Not sure why we couldn't catch them running in Github actions, I'll demo tomorrow with the team and see what we come up with.

ZackDeRose:
## Regarding Results Not Caching When ReRunning

I created my own fork so I could try recreating this issue - I hadn't been able to recreate since I bumped the nx-cloud runner version (see https://github.com/ZackDeRose/query/pull/3 for that investigation).

I have turned on the `NX_VERBOSE_LOGGING` environment variable in the `pr.yml` file - this will add the hash information to the logs so we can debug any cache issues we see in the future.

## Regarding Nx Cloud Tokens

+ This PR changes the publicly listed key in the `nx.json` file to a read-only key that is safe for the public repo.
+ Locally, if trusted contributors want to create their own read && write key, they can do so via the https://nx.app portal (I've sent you an invite on Nx Cloud to the organization I created, @TkDodo). We can create an `nx-cloud.env` file at the root of the workspace with the contents being:

```
NX_CLOUD_ACCESS_TOKEN=your_key_here
```
(this file is added to the .gitignore list in this PR)

+ In the `pr` workflow, I've set this `NX_CLOUD_ACCESS_TOKEN` to be added to our workflow's env settings, and verified this working with my own fork. We'll need to added this as a secret in this repo as well for cloud caching to work in our PR's

## Regarding the Caching of Tasks

Investigating the timing of some of our tasks - I don't think it makes sense to continue caching the `test:build` and `test:format` tasks, as they make up such a small part of our "task profile" in terms of where we could potentially save time. `test:lib` and `test:eslint` appear to make up the majority of our time:

Screen Shot 2023-03-21 at 11 23 27 PM

`build`, `build:types`, `test:types`, and `rollup` are not trivial, so I'd like to continue caching them.

If we have concerns about caching the build for release, we can also add the `--skip-nx-cache` flag to the `scripts/cipublish.ts` script where `pnpm build` is called. Let me know if you'd like me to make that change @TkDodo!

## The Nx Cloud GitHub App

We can add the Nx Cloud github app here: https://github.com/apps/nx-cloud

This will add a comment to our PRs going forward with updates on the tasks (I already added to my fork so we can see it here: https://github.com/ZackDeRose/query/pull/3).

## What's Next?

Updating our `pr.yml` file to use Distributed Task Execution!

TkDodo:
> If we have concerns about caching the build for release, we can also add the --skip-nx-cache flag to the scripts/cipublish.ts script where pnpm build is called.

yeah I think that's a good idea. What takes most time in the build script is that we re-run the tests again, so after remote caching, that should be instant, given that the tests have executed already on the PR level. That happens outside the `cipublish` script call, so in that script, we just run the build once for affected packages and ship it. I would be fine doing this without caching.

TkDodo:
@ZackDeRose I've added the nx cloud github app to TanStack/query.

ZackDeRose:
> > If we have concerns about caching the build for release, we can also add the --skip-nx-cache flag to the scripts/cipublish.ts script where pnpm build is called.
> 
> yeah I think that's a good idea. What takes most time in the build script is that we re-run the tests again, so after remote caching, that should be instant, given that the tests have executed already on the PR level. That happens outside the `cipublish` script call, so in that script, we just run the build once for affected packages and ship it. I would be fine doing this without caching.

Cool, I've added the `--skip-nx-cache` to the `scripts/publish.ts` file now accordingly.

> @ZackDeRose I've added the nx cloud github app to TanStack/query.

Cool! All that's left then is to create a read/write key via the Nx Cloud dashboard. @TkDodo, I had sent you an invite to the org I had set up initially for this repo - let me know if you weren't able to receive it. Once you're added, you can create the key and add it to this github project's secrets so the key can be used in our workflow files.

TkDodo (approved):

PR #5147, opened 2023-03-17
fix: test:format task giving false positives

ZackDeRose:
<div align="left">
      <a href="https://www.youtube.com/watch?v=QetA8uiMcwM" target="blank">
         
      </a>
</div>

## What it does:

+ fixes issue with the `test:format` target, where touching some files was giving us a false cache hit
+ moves the `test:format` from the `nx.json` file specifically to the root `project.json` file, as the `test:format` task only exists at the root level.

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -65,9 +65,6 @@
       "inputs": ["default", "^public"],
       "dependsOn": ["^build:types"]
     },
-    "test:format": {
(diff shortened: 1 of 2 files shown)


TkDodo:
caching is hard. I appreciate the fast fix ❤️

PR #5150, opened 2023-03-17
refactor(react-query-devtools): move devtools components in files

gpichot:
This is a small PR that moves stuff around to prepare the ground to add support for a timeline panel.

I also added display names to component to make it clearer what are the components in React Devtools.


gpichot:
@TkDodo Yes, sure, it's done.
I may do something wrong, though, I tried to run an example (examples/react/rick-morty), but it doesn't seem to use local packages for alpha branch on my laptop.

I did a:
```
npx nx reset
rm -rf node_modules # in both root and example directories
pnpm install # at root
pnpm run watch # at root
# In example directory
pnpm run dev
```
But it still links local react-query package to the locally installed 4.26.1 version. Do you have any idea why?

TkDodo:
@gpichot can you have a look at the conflicts please?

gpichot:
@TkDodo Done

PR #5153, opened 2023-03-17
Feature/query options

TkDodo:


packages/react-query/src/useQuery.ts
```diff
@@ -7,6 +7,28 @@ import type {
...
+  TError = unknown,
+  TData = TQueryFnData,
+  TQueryKey extends QueryKey = QueryKey,
+>(
+  options: DefinedInitialDataOptions<TQueryFnData, TError, TData, TQueryKey>,
+): DefinedInitialDataOptions<TQueryFnData, TError, TData, TQueryKey>
+
+export function queryOptions(options: unknown) {
+  return options
+}
```
TkDodo, line 30: should we move this to the core, @DamianOsipiuk ?

The options are currently based on `UseQueryOptions`, which only exists in react-query. I think each lib has their own impl over `QueryObserverOptions` 🤔

DamianOsipiuk: Yeah this would not work as in `Vue` we have to wrap every option as a possible `ref`
https://github.com/TanStack/query/blob/alpha/packages/vue-query/src/useQuery.ts#L16

TkDodo: okay then let's let every framework export their own `queryOptions` ?

DamianOsipiuk: Yes, i think that would be necessary.

packages/react-query/src/useQuery.ts
```diff
@@ -7,6 +7,28 @@ import type {
 } from './types'
 import { useBaseQuery } from './useBaseQuery'
 
+export function queryOptions<
+  TQueryFnData = unknown,
```
DamianOsipiuk, line 11: Since this is `query` specific, we would need similar helper for `infinite`, `mutation` and `queries`?

TkDodo: probably not necessary for mutations, but for infinite: yes

TkDodo: We don't even have the initialData overload for infiniteQuery 😅

TkDodo: I would add a separate helper in a follow-up. This is not breaking so we can add it at any time I think

jasikpark:
from testing this, it seems like `queryOptions` needs to take the return type of `select` into account?

eg

```tsx

export function fetchConfigOverridesQueryOptions(hostId: string, { suspense } = { suspense: false }) {
  return queryOptions({
    queryKey: hostQueryKeys.configOverrides(hostId),
    queryFn: () => Hosts.getConfigOverrides(hostId),
    suspense,
    select: mapOverridesToForm,
  });
}
```

where 

```tsx
export function mapOverridesToForm({ data: { overrides } }: Successful<Hosts.HostConfigOverrideResult>): FormData {
```

means that I get this error with a vendorized version of this PR:

```
Type 'UndefinedInitialDataOptions<{ data: HostConfigOverrideResult; metadata?: Metadata | undefined; }, unknown, FormData, (string | readonly [...string[], "detail", string])[]>' is not assignable to type 'UseQueryOptionsForUseQueries<{ data: HostConfigOverrideResult; metadata?: Metadata | undefined; }, Error, { data: HostConfigOverrideResult; metadata?: Metadata | undefined; }, (string | readonly [......[]])[]>'.
  Type 'UndefinedInitialDataOptions<{ data: HostConfigOverrideResult; metadata?: Metadata | undefined; }, unknown, FormData, (string | readonly [...string[], "detail", string])[]>' is not assignable to type 'Omit<UseQueryOptions<{ data: HostConfigOverrideResult; metadata?: Metadata | undefined; }, Error, { data: HostConfigOverrideResult; metadata?: Metadata | undefined; }, (string | readonly [......[]])[]>, "placeholderData">'.
    Types of property 'refetchInterval' are incompatible.
      Type 'number | false | ((data: FormData | undefined, query: Query<{ data: HostConfigOverrideResult; metadata?: Metadata | undefined; }, unknown, { ...; }, (string | readonly [......[]])[]>) => number | false) | undefined' is not assignable to type 'number | false | ((data: { data: HostConfigOverrideResult; metadata?: Metadata | undefined; } | undefined, query: Query<{ data: HostConfigOverrideResult; metadata?: Metadata | undefined; }, Error, { ...; }, (string | readonly [......[]])[]>) => number | false) | undefined'.
        Type '(data: FormData | undefined, query: Query<{ data: HostConfigOverrideResult; metadata?: Metadata | undefined; }, unknown, { data: HostConfigOverrideResult; metadata?: Metadata | undefined; }, (string | readonly [......[]])[]>) => number | false' is not assignable to type 'number | false | ((data: { data: HostConfigOverrideResult; metadata?: Metadata | undefined; } | undefined, query: Query<{ data: HostConfigOverrideResult; metadata?: Metadata | undefined; }, Error, { ...; }, (string | readonly [......[]])[]>) => number | false) | undefined'.
          Type '(data: FormData | undefined, query: Query<{ data: HostConfigOverrideResult; metadata?: Metadata | undefined; }, unknown, { data: HostConfigOverrideResult; metadata?: Metadata | undefined; }, (string | readonly [......[]])[]>) => number | false' is not assignable to type '(data: { data: HostConfigOverrideResult; metadata?: Metadata | undefined; } | undefined, query: Query<{ data: HostConfigOverrideResult; metadata?: Metadata | undefined; }, Error, { ...; }, (string | readonly [......[]])[]>) => number | false'.
            Types of parameters 'data' and 'data' are incompatible.
              Type '{ data: HostConfigOverrideResult; metadata?: Metadata | undefined; } | undefined' is not assignable to type 'FormData | undefined'.
                Type '{ data: HostConfigOverrideResult; metadata?: Metadata | undefined; }' is missing the following properties from type 'FormData': lighthouse, listen, tun, loggi
```

Though this is only when passed to `useQueries`!

jasikpark:
if I pass the configs inline,

```tsx
 const [{ data: savedData, isPending }, { data: featureFlags }] = useQueries({
    queries: [
      {
        queryKey: hostQueryKeys.configOverrides(hostId),
        queryFn: () => Hosts.getConfigOverrides(hostId),
        suspense: true,
        select: mapOverridesToForm,
      },
      {
        queryKey: orgQueryKeys.featureFlags(),
        suspense: true,
        queryFn: ({ signal }) => Org.getFeatureFlags({ signal }),
        // 10 minutes
        staleTime: 600000,
      },
    ],
  });
```

they typecheck, except `signal` is untyped

TkDodo:
hmm yeah, seems to be an issue with useQueries typings. can you file an issue?

https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgVwM4FMCKz1WO1OAXzgDMoIQ4ByAARgEMA7VRgYwGsB6KdBtmAFoAjjigBPagFgAULLhsILeL1TIANvAC8KDNlz5UACgTy450QYIAuOAG0z586ZlO3cSxIDS6cbbvUpBAQ1AC6ADSO7p7iAGJMtiZwqMAA5kwM6sQAlHBaAHxwAAoUIMAYAHSqEOoAbuhGgcHU2ZGu7kRtbhFmRNkA3EA

jasikpark:
made the issue, thx: https://github.com/TanStack/query/issues/5342

jasikpark:
is the resolution for "useQueries" with this PR still tbd? 

my issue got closed - idk whether there's plans to fix the useQueries types?

TkDodo:
it's labelled as a known limitation

jasikpark:
well, the second type error is covered by that, but the first problem in https://github.com/TanStack/query/pull/5153#issuecomment-1522517570 is that the output of the queryOptions function doesn't typecheck to useQueries. I'll wait to file an issue til I can try the prereleased version instead of my copy of this PR though.

PR #5161, opened 2023-03-19
Preserve 'use client' directives

### LINKED ISSUE #4933: Doesn't work out of the box with next 13
### Describe the bug

when adding react-query to next 13 i get following error when running 
```
error - ./node_modules/@tanstack/react-query/build/lib/Hydrate.mjs
Attempted import error: 'useRef' is not exported from 'react' (imported as 'React').

Import trace for requested module:
./node_modules/@tanstack/react-query/build/lib/Hydrate.mjs
./node_modules/@tanstack/react-query/build/lib/index.mjs
./src/app/layout.tsx
```

code 
```
import {QueryClient, QueryClientProvider} from "@tanstack/react-query";

/* ... */
          {children}
      </QueryClientProvider>
/* ... */
```

https://github.com/JonasWijne/next-test-react-query

### Steps to reproduce

1. git clone https://github.com/JonasWijne/next-test-react-query.git
2. cd  next-test-react-query
3. yarn
4. yarn dev
5. open browser to http://localhost:3000

### Expected behavior

as a user i expect it to compile/dev serve

Every time

### Screenshots or Videos

image

### Platform

- OS: MacOS
- browser: chrome

### TanStack Query version

4.24.4

### TypeScript version

4.9.5

### Additional context

_No response_

Ephem:
Fixes #4933

* Pulls in `rollup-plugin-preserve-directives` and adds it to all un-bundled builds to preserve 'use client' directives
  * See rollup issue https://github.com/rollup/rollup/issues/4699
* Adds `'use client'` to all files that should have it (I think)

This makes everything work as documented with Next app router, that is, developers should no longer need to:

* Wrap our imports in `"use client"` themselves (just import and use `Hydrate` directly for example)
* Import from `@tanstack/query-core` in RSCs (can now import directly from `@tanstack/react-query`)

> For anyone reading along: Note that React Query-support for the Next app router is still experimental!

I tested it like this (basically quick version of the docs):

```jsx
// Providers.jsx
// This wraps the entire app in root layout.jsx
"use client";

import { useState } from "react";
import { QueryClient, QueryClientProvider } from "@tanstack/react-query";

export default function Providers({ children }) {
  const [queryClient] = useState(() => new QueryClient());

  return (
    <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
  );
}
```

```tsx
// page.jsx
import { dehydrate, Hydrate, QueryClient } from "@tanstack/react-query";
import ClientComponent from "./ClientComponent";

export default async function ReactQuery() {

  await queryClient.fetchQuery({
    queryKey: ["1"],
    queryFn: () => "Server",
  });

  return (
    <Hydrate state={dehydrate(queryClient)}>
      <ClientComponent />
    </Hydrate>
  );
}
```

```jsx
// ClientComponent.jsx
"use client";

export default function ClientComponent() {
  const { data } = useQuery({
    queryKey: ["1"],
    queryFn: () => "Client",
    // Set staleTime to see that we get the "Server" output,
    // can later switch tabs to trigger revalidation and see we
    // get the "Client" output
    staleTime: 10000,
  });

  return <div>{data}</div>;
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,6 +92,7 @@
     "react-dom-17": "npm:react-dom@^17.0.2",
     "rimraf": "^3.0.2",
     "rollup": "^2.70.2",
+    "rollup-plugin-preserve-directives": "^0.1.0",
     "rollup-plugin-size": "^0.2.2",
     "rollup-plugin-terser": "^7.0.2",
     "rollup-plugin-visualizer": "^5.6.0",
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "react-dom-17": "npm:react-dom@^17.0.2",
     "rimraf": "^3.0.2",
     "rollup": "^2.70.2",
-    "rollup-plugin-preserve-directives": "^0.1.0",
+    "rollup-plugin-preserve-directives": "0.1.0",
     "rollup-plugin-size": "^0.2.2",
     "rollup-plugin-terser": "^7.0.2",
     "rollup-plugin-visualizer": "^5.6.0",
--- a/packages/react-query/src/errorBoundaryUtils.ts
+++ b/packages/react-query/src/errorBoundaryUtils.ts
@@ -1,3 +1,4 @@
+'use client'
 import type {
   DefaultedQueryObserverOptions,
   Query,
--- a/packages/react-query/src/reactBatchedUpdates.ts
+++ b/packages/react-query/src/reactBatchedUpdates.ts
@@ -1,2 +1,3 @@
+'use client'
 import * as ReactDOM from 'react-dom'
 export const unstable_batchedUpdates = ReactDOM.unstable_batchedUpdates
--- a/packages/react-query/src/useBaseQuery.ts
+++ b/packages/react-query/src/useBaseQuery.ts
@@ -1,3 +1,4 @@
+'use client'
 import * as React from 'react'
 import { useSyncExternalStore } from './useSyncExternalStore'
 
--- a/packages/react-query/src/useInfiniteQuery.ts
+++ b/packages/react-query/src/useInfiniteQuery.ts
@@ -1,3 +1,4 @@
+'use client'
 import type {
   QueryObserver,
   QueryFunction,
--- a/packages/react-query/src/useIsFetching.ts
+++ b/packages/react-query/src/useIsFetching.ts
@@ -1,3 +1,4 @@
+'use client'
 import * as React from 'react'
 import type { QueryKey, QueryFilters } from '@tanstack/query-core'
 import { notifyManager, parseFilterArgs } from '@tanstack/query-core'
--- a/packages/react-query/src/useIsMutating.ts
+++ b/packages/react-query/src/useIsMutating.ts
@@ -1,3 +1,4 @@
+'use client'
 import * as React from 'react'
 import { useSyncExternalStore } from './useSyncExternalStore'
 
--- a/packages/react-query/src/useMutation.ts
+++ b/packages/react-query/src/useMutation.ts
@@ -1,3 +1,4 @@
+'use client'
 import * as React from 'react'
 import { useSyncExternalStore } from './useSyncExternalStore'
 
--- a/packages/react-query/src/useQueries.ts
+++ b/packages/react-query/src/useQueries.ts
@@ -1,3 +1,4 @@
+'use client'
 import * as React from 'react'
 import { useSyncExternalStore } from './useSyncExternalStore'
 
--- a/packages/react-query/src/useQuery.ts
+++ b/packages/react-query/src/useQuery.ts
@@ -1,3 +1,4 @@
+'use client'
 import type { QueryFunction, QueryKey } from '@tanstack/query-core'
 import { parseQueryArgs, QueryObserver } from '@tanstack/query-core'
 import type {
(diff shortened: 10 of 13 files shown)


TkDodo:
I've checked the output of the codesandbox preview build:

- ✅ `/build/umd/index.development.js` doesn't have any `'use client'`
- ✅ `/build/umd/index.production.js` doesn't have any `'use client'`
- ✅ `/build/lib/Hydrate.esm.js` has `'use client'` at the top
- ✅ `/build/lib/Hydrate.js` has `'use client'` at the top, followed by `'use strict'`

This looks like what we want 👍 . Shipping it 🚢

thebuilder:
I'm seeing a lot of warnings like this:

```
Module level directives cause errors when bundled, "use client" in "node_modules/.pnpm/@tanstack+react-query@4.28.0_biqbaboplfbrettd7655fr4n2y/node_modules/@tanstack/react-query/build/lib/useQueries.mjs" was ignored.
```

Occurs when bundling client output with Vite. Not sure what I need to change in my setup, to support the "use client" directive?

TkDodo:
I'm seeing the same warnings in rollup:

```
packages/react-query/src/useMutation.ts (1:0)
1: 'use client'
   ^
2: import * as React from 'react'
3: import { useSyncExternalStore } from './useSyncExternalStore'
(!) Module level directives cause errors when bundled, 'use client' was ignored.
packages/react-query/src/useInfiniteQuery.ts (1:0)
```

it does make sense because when we bundle together into one file, we shouldn't keep `'use client'`.

thebuilder:
Won't that always be the case when consuming `react-query`? It will always be bundled into the application?

TkDodo:
I don't know. I guess `nextJs` might consume the `.mjs` files separately and then decide based on that what is a client component and what isn't. I honestly don't know enough about how next does this 🤷

thebuilder:
Great - Not just me that's confused about this new `"use client"` thing. Add it to the pile of problems we already have with CommonJS, ESM and module publishing. :(

Is this the right solution for the the react-query library? If the Next.js 13 app folder is still beta, should the ecosystem figure out the right way of handling `use client` in libraries as part of this?

Ephem:
@thebuilder It is confusing, and sorry for adding a bunch of warnings to your build! I suspect you are not using Server Components at the moment, so you can safely just ignore these warnings.

If you want to hide the annoying output, you can add a [Rollup onwarn](https://rollupjs.org/configuration-options/#onwarn) filter to hide them. I get why Rollup warns for this, but now that directives is becoming more common, it's also unfortunate.

The reason you are seeing this is because you are using the unbundled `.mjs`-version of React Query, which now includes `'use client'` at the top of some files. It's good that you are using this version, because that means your bundler can remove any parts of React Query you are not actively using, so it doesn't get included in your final app-bundle!

Thing is though, Server Components can use this `'use client'` directive to do _even smarter_ code splitting. So yes, React Query will always get bundled into the application one way or another in the end, but for the unbundled builds that are a sort of "intermediary step", we want to provide all the "metadata" we can to bundlers to do smart things.

This is all based on a specification any bundler can implement too, so it's not Next.js specific even if they were first to implement it. It's definitely something that is up to us as an ecosystem to figure out, and different libraries might have different tradeoffs.

Ephem:
Oh, and just to answer your question specifically to avoid confusion:

> Not sure what I need to change in my setup, to support the "use client" directive?

Since you are not using Server Components, you don't need to "support" the directive! Your bundler is warning because it doesn't know what to do with this. It's at the top of a file it wants to include in a bundle, and it can't safely hoist that directive to the entire bundle, so it's removing it instead, which is fine since you don't need it. It doesn't want to just remove things without warning though, which is why you are seeing this.

thebuilder:
Thanks for the explanation. Since RSC is still only supported in a few select frameworks, like Next, it obviously is not something we can use with Vite or other frameworks. 

I feel like this is something that Vite/Rollup should handle out of the box. My build output is going to be filled with warnings, especially if more libraries start adding 'use client'. Until that happens, it seems a bit premature to start adding it to library code?

Ephem:
I think about it the other way around though. Unless libraries start using this, what's the incentive to change/fix the behaviour? 😃

Again, sorry about the noisy warnings, adding an `onwarn` handler is a quick way to get rid of them. If it helps I can put together the specific config you need to add tomorrow?

thebuilder:
Maybe it should be added directly to the https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-react package instead?

Ephem:
Absolutely! I only suggest adding the `onwarn` as a stopgap. I think ultimately Rollup should only output this warning once per build instead of once per file and I think any React plugin that explicitly does not support RSCs could hide that warning by default.

duypham90:
I got many error warning when build production with React + Vite v4.2.1
Not use NextJS

Screen Shot 2023-03-31 at 21 09 59

vadymstebakov:
> I got many error warning when build production with React + Vite v4.2.1 Not use NextJS
> 
> Screen Shot 2023-03-31 at 21 09 59

I hot the same issue with React + Vite v4.1.4

vaynevayne:
same as  `vite 4.2.1`

BATCOH:
Here is the `vite.config.ts` snippet that disables "use client" warnings
```typescript
import { defineConfig } from "vite";

// https://vitejs.dev/config/
export default defineConfig({
  build: {
    rollupOptions: {
      /**
       * Ignore "use client" waning since we are not using SSR
       * @see {@link https://github.com/TanStack/query/pull/5161#issuecomment-1477389761 Preserve 'use client' directives TanStack/query#5161}
       */
      onwarn(warning, warn) {
        if (
          warning.code === "MODULE_LEVEL_DIRECTIVE" &&
          warning.message.includes(`"use client"`)
        ) {
          return;
        }
        warn(warning);
      },
    },
  },
});
```

PR #5165, opened 2023-03-20
[WIP] React Query Optimizations - Maintaining Object References

kevupton:
## Issue:

Right now it is impossible to use the object returned by the `useQuery` in dependency arrays for react. This is because each rerender we are recreating the object and then returning a new instance. 

## Suggestion

We give the user the ability to turn off optimistic results, by using an `optimistic: false` flag on the options.

```ts
const query = useQuery({ optimistic: false });
useEffect(() => {
  // do something in here
}, [query]);
```

## Additional Optimization Improvement

We can also improve performance by caching the options object so that we do not recreate it each time, as the `options._defaulted` prevents it from be defaulted again. 

### useQueryOptions

```ts
import {useQueryOptions} from '@tanstack/react-query';

export const useAccountOptions = (id: string) => {
  return useQueryOptions(
    () => ({
      enabled: !!id,
      queryKey: ['account', id],
      queryFn: async () => ...,
    }),
    [id]  // only update the options object whenever the id changes.
  );
};
```

Usage: 
```ts
const options = useAccountOptions(id);
const query = useQuery(options);
```

### Notes:

`useQueryOptions` means that the options object is more accessible and can be used by `useQueries` much more easily, or passed around and used as needed.


TkDodo:
without optimistic results, you'll get the first render cycle in a state that has `isFetching: false`. Also, you will not get notified about observer option changes because `observer.setOptions` assumes updates have been propagated optimistically already:

https://github.com/TanStack/query/blob/fe0ae802d8e7521ba6d55239e7daf03ec0dd20fa/packages/react-query/src/useBaseQuery.ts#L89-L93

I would generally say that if you want referential stability, you should probably be using `data` returned from `useQuery`, as it is structurally shared out of the box:

```
const { data } = useQuery(...)

useEffect(() => {
  // do something in here with data
}, [data]);
```

kevupton:
@TkDodo 
I ran some tests and the first render cycle does appear to be fetching:

image

Using this code:
```ts
const useTest = () => {
  const query = useQueryOptions(() => ({
    queryKey: ["test"],
    queryFn: async () => {
      await new Promise((res) => setTimeout(res, 30000));
      return true;
    },
  }));

  return useQuery(query);
};
```

I believe the reason is because we are already optimistically creating the results when we initially create the query.

Here the optimistic results are created in this function
https://github.com/TanStack/query/blob/main/packages/query-core/src/queryObserver.ts#L94

Which eventually lead to:
https://github.com/TanStack/query/blob/main/packages/query-core/src/queryObserver.ts#L597

Which from what I can see is doing the same thing as calling 
```ts
observer.getOptimisticResult()
```

So the only reason I thought to leave the optimistic results there, was in the scenario after the very first render, as I am not sure about the unknown possibilities the getOptimisticResults would influence after the first render.

TkDodo:
could we instead just structurally share the optimistically created result with the current result? Something like, in `getOptimisticResult`:

```diff
- return this.createResult(query, options)
+ return replaceData(
+   this.#currentResult,
+   this.createResult(query, options),
+   options,
+ )
```

This should lead to `currentResult` being returned in case there are no changes. We'd need to keep the `useMemo` on the `trackResult` call though

TkDodo:
> So the only reason I thought to leave the optimistic results there, was in the scenario after the very first render, as I am not sure about the unknown possibilities the getOptimisticResults would influence after the first render.

So I went ahead and just used the return value of `useSyncExternalStore` in `useBaseQuery`, thus removing the call to `getOptimisticResult` altogether, which is something that I would generally really appreciate.

However, it leads to 14 test failures out of 128 of the `useQuery.test.tsx` test suite.

Probably, in one of those failing test lies the answer to why we can't easily do this if you want to look at this a bit further 😅

TkDodo:
right, the reason is probably that `setOptions` is only called in the constructor and then again in an effect. So if your query changes queryKeys, without the call to `getOptimisticResult`, it would have one render cycle where the options are from an old observer.

kevupton:
yeh ok awesome, will take a look into this one more and see what we can do 🙏

packages/react-query/src/useBaseQuery.ts
```diff
@@ -118,7 +118,9 @@ export function useBaseQuery<
...
 
   // Handle result property usage tracking
-  return !defaultedOptions.notifyOnChangeProps
-    ? observer.trackResult(result)
-    : result
+  return React.useMemo(() => {
+    return !defaultedOptions.notifyOnChangeProps
+      ? observer.trackResult(result)
+      : result
+  }, [defaultedOptions.notifyOnChangeProps, observer, result])
```
incepter, line 125: I don't believe this is an optimization ?
I mean, the initial code has a ternary expression that uses memoized result if the render wasn't abount a new value;
The target code has array and function creation every time + hook call that compares the previous and new array values.

packages/react-query/src/useQueryOptions.ts
```diff
@@ -0,0 +1,31 @@
+import type {DefaultedQueryObserverOptions, QueryClient, QueryKey, QueryObserverOptions} from "@tanstack/query-core";
+import {useQueryClient} from "./QueryClientProvider";
+import type React from "react";
+import type { DependencyList } from "react";
+import {useMemo} from "react";
+
+export function useQueryOptions<
```
incepter, line 7: How are these two APIs used ? The PR only contains declarations and exports.

Aren't useQueryOptions a combination of what the user gives and default options? in this case, how this hook is adding any value ?

kevupton: sorry this was in progress at the time, based on the above feedback. I have updated it now. It ties into a larger picture here.

packages/react-query/src/useBaseQuery.ts
kevupton: the goal of this PR is to return an object that is the same when the input is the same.

kevupton:
@incepter @TkDodo Let me know what you guys think so far. 
There are a few larger changes here, but in doing them, I have been able to reduce the number of rendering and state updates, plus able to maintain a object equality when there have been no changes.

TkDodo:
@kevupton sorry for letting this one slide. I'm generally interested in getting rid of `getOptimisticUpdate`, if we can do it in a backwards compatible way. We've released v5 in the meantime - would you want to take another stab at it? I can see there are quite some conflicts, probably mainly because we switched to private class fields

TkDodo:
I think it would be easier to start from scratch here. Let me know if you're interested in completing this.

kevupton:
@TkDodo No problem, apologies for the delayed response. 
I agree would it be best to go from scratch. I will have a go again and see how I go 🙏

PR #5166, opened 2023-03-20
update build task

ZackDeRose:
# What it does:

+ updates the build task to use `dependsOn` rather than `run-commands`.
  + I ended up using this for @tanstack/table, and I think I like this better
+ The splitting out of the rollup step to its own cacheable task makes the caching a bit better too (I believe prior, we were stashing everything in the `build` directory for both `build` and `build:types` steps)

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -10,7 +10,8 @@
           "build:types",
           "test:format",
           "test:build",
-          "build"
+          "build",
+          "rollup"
         ]
       }
     }
--- a/nx.json
+++ b/nx.json
@@ -56,12 +56,18 @@
       "inputs": ["default", "^public"]
     },
     "test:types": {
-      "outputs": ["{projectRoot}/build"],
+      "outputs": [
+        "{projectRoot}/build/**/*.d.ts",
+        "{projectRoot}/build/.tsbuildinfo"
+      ],
       "inputs": ["default", "^public"],
       "dependsOn": ["^test:types"]
     },
     "build:types": {
-      "outputs": ["{projectRoot}/build"],
+      "outputs": [
+        "{projectRoot}/build/**/*.d.ts",
+        "{projectRoot}/build/.tsbuildinfo"
+      ],
       "inputs": ["default", "^public"],
       "dependsOn": ["^build:types"]
     },
--- a/nx.json
+++ b/nx.json
@@ -8,8 +8,6 @@
           "test:eslint",
           "test:types",
           "build:types",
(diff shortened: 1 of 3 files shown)


TkDodo:
> adding rollup to cacheable operations

is this a thing worth doing? I just fear that if we get the cache config wrong, we might be shipping cached packages that are not made from the latest source ...

TkDodo:
also, codesandbox preview build is now failing 🤔

ZackDeRose:
> > adding rollup to cacheable operations
> 
> is this a thing worth doing? I just fear that if we get the cache config wrong, we might be shipping cached packages that are not made from the latest source ...

Prior to this PR, we are still caching the build. This change is really to split out rollup so that the rollup command is "atomically cacheable", meaning we can run `nx rollup` and cache the result of that command alone. Without this change, we can cache the result of `pnpm build`, but not any of the individual steps.

Let me know if we want to remove the build from the list of cacheable operations! As we've mentioned, caching is hard, so if we're not confident yet we can keep caching only for the pr workflows or only for the "validation" / "test:*" tasks!

I do think ideally we'll want to eventually cache the build though!

> also, codesandbox preview build is now failing 🤔

Hrmm, not able to recreate the same failure locally while following along with the commands in the log here: https://ci.codesandbox.io/status/TanStack/query/pr/5166/builds/358582

I'll try to dig deeper tomorrow. Do we know how to find/set the node version for code sandbox?

ZackDeRose:
RE: code sandbox failures - leaving this for future me:

csb's node version: 16.18.1
setup steps:
1. clones repo
2. `npm run install:csb`
3. `npm run build` < this doesn't complete successfully

TkDodo:
> Let me know if we want to remove the build from the list of cacheable operations! As we've mentioned, caching is hard, so if we're not confident yet we can keep caching only for the pr workflows or only for the "validation" / "test:*" tasks!

> I do think ideally we'll want to eventually cache the build though!

given that this is a caching library, we believe in caching 😆 . If you're confident it'll work, I'm fine with it :)

TkDodo:
it might be worth noticing that we usually only release packages that have affected changes in the `src` directory, but there are ways to bypass this by having `RELEASE_ALL` in the commit message. This will right now release all packages. I just want to make sure that in those cases, we won't get cached output released ...

ZackDeRose:
closing this PR as I've pulled these changes into: #5143

let's continue the conversation there @TkDodo

PR #5180, opened 2023-03-23
Update useQueries.md

mbosecke:
Specifying that the return value is ordered, to help users understand how to match up the results with their inputs:

https://stackoverflow.com/questions/72045306/usequeries-how-to-get-response-for-a-certain-query

### DIFF
--- a/docs/react/reference/useQueries.md
+++ b/docs/react/reference/useQueries.md
@@ -25,4 +25,4 @@ The `useQueries` hook accepts an options object with a **queries** key whose val
 
 **Returns**
 
-The `useQueries` hook returns an ordered array with all the query results.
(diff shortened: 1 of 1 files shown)


docs/react/reference/useQueries.md
```diff
@@ -25,4 +25,4 @@ The `useQueries` hook accepts an options object with a **queries** key whose val
 
 **Returns**
 
-The `useQueries` hook returns an array with all the query results.
+The `useQueries` hook returns an ordered array with all the query results.
```
TkDodo, line 28: order how? can we elaborate on that?

mbosecke: I have fixed it with this, hopefully it's a bit more clear:

The `useQueries` hook returns an array with all the query results. The order returned is the same as the input order.

PR #5195, opened 2023-03-27
refactor: change QueryMeta and MutationMeta to an interface

umangjun92:
Changes the `QueryMeta` and `MutationMeta` types to interfaces, so that they can have their shapes overridden in the implementing codebase for safer typing and autocompletion.

```
interface Meta {
  persist?: boolean;
}

  interface QueryMeta extends Meta {}
}

useQuery(queryKey, {
  meta: {
    persist: false,
  },
});
```

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -48,7 +48,9 @@ export interface InfiniteData<TData> {
   pageParams: unknown[]
 }
 
-export type QueryMeta = Record<string, unknown>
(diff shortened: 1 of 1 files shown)


TkDodo:
We won't be making a new release of `v3` since it's under the `react-query` name, as opposed to `@tanstack/react-query`. FYI, we've already solved this in v5.

PR #5212, opened 2023-04-01
removing docs from default inputs

ZackDeRose:
## What it does:

+ removes docs from the `default` named inputs. This way, changes to `.md` files won't invalidate the nx cache!


nx.json
```diff
@@ -36,7 +36,8 @@
     "default": [
       "{projectRoot}/**/*",
       "globalBuildAffectingConfig",
-      "globalNonBuildAffectingConfig"
+      "globalNonBuildAffectingConfig",
+      "!{workspaceRoot}/**/*"
```
TkDodo, line 40: how does this remove `docs` specifically? According to the nx docs:

> {workspaceRoot} is a key word that is replaced by the root path of your workspace.

so `workspaceRoot` is our root directory - the one above `packages`, right? So that glob pattern would exclude .. everything 😅 ?

ZackDeRose: ackkk my mistake - sorry I lost touch on this one - fixing now

TkDodo (approved):

PR #5215, opened 2023-04-01
switching to an affected strategy

ZackDeRose:
## What It Does:

+ Switches from using only the cache to not rerun tasks to using `nx affected` in addition to the cache
  + package scripts have been updated accordingly
  + pr workflow updated to set the `--base` option for nx affected to the PR's target branch
  + ci workflow updated to set the `--base` option for nx affected to the commit before the triggering push
  + ci-v3 workflow updated to set the `--base` option for nx affected to the commit before the triggering push


.github/workflows/ci-v3.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           node-version: ${{ matrix.node }}
       - name: Install dependencies
         uses: bahmutov/npm-install@v1
-      - run: yarn test:ci
+      - run: yarn test:ci --base=${{ github.event.before }}
```
TkDodo, line 26: I don't think we need to change something here. This will only run on the v3 branch which doesn't know nx

```suggestion
      - run: yarn test:ci
```

TkDodo (approved):

TkDodo:
I don't think the forwarding of `--base` works?

```
> query@ test:lib /home/runner/work/query/query
> nx affected --target=test:lib --parallel=5


 >  NX   Affected criteria defaulted to --base=main --head=HEAD

fatal: ambiguous argument 'main': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
nx.js affected
```

ZackDeRose:
> I don't think the forwarding of `--base` works?
> 
> ```
> > query@ test:lib /home/runner/work/query/query
> > nx affected --target=test:lib --parallel=5
> 
> 
>  >  NX   Affected criteria defaulted to --base=main --head=HEAD
> 
> fatal: ambiguous argument 'main': unknown revision or path not in the working tree.
> Use '--' to separate paths from revisions, like this:
> 'git <command> [<revision>...] -- [<file>...]'
> nx.js affected
> ```

Yes, something went wrong! lol

`pnpm` should forward the option - I believe this is actually me not getting the github context stuff correct. Let me see...

ZackDeRose:
Alright - got the fix now. For the ci workflow, we'll want to make sure to test it.

(The issues with the PR workflow were due to this PR coming in from a fork. After checking out with default options, we were in a detatched head state, so Nx was unable to find the main branch in the linked list of commits, so affected command failed. With pushes to main/alpha/etc., this **should** be a non-issue since we'll always be inside this repo, never a fork. All the same, don't wanna bring things down!)

TkDodo (approved):

PR #5218, opened 2023-04-01
test(eslint-plugin): add test

Newbie012:


packages/eslint-plugin-query/src/rules/exhaustive-deps/exhaustive-deps.test.ts
```diff
@@ -38,6 +38,20 @@ ruleTester.run('exhaustive-deps', rule, {
...
+      code: `
+        import useApi from './useApi'
+
+        const useFoo = () => {
+          const api = useApi();
+          return useQuery({
+            queryKey: ['foo'],
+            queryFn: () => api.fetchFoo(),
+          })
+        }
```
TkDodo, line 52: how is this test succeeding? By what metrics is `api` excluded ?

PR #5219, opened 2023-04-02
feat(useQueries): combine 

TkDodo:


TkDodo:
@ardeora @lachlancollins could you please take a look at the solid and svelte errors we're getting here? Thanks 🙏

ardeora:
> @ardeora @lachlancollins could you please take a look at the solid and svelte errors we're getting here? Thanks 🙏

I'll work on adding the combine option to `solid-query` today! Thanks for the reminder 😄

TkDodo:
@ardeora this solid tests seems to fail consistently, I can also reproduce it locally:

```
 ❯ src/__tests__/createQueries.test.tsx  (6 tests | 1 failed) 150ms
   ❯ src/__tests__/createQueries.test.tsx > useQueries > should return the correct states
     → Cannot redefine property: status
```

TkDodo:
ah damn, this seems to have been me - adding property tracking to queriesObserver in the last commit. Hmm, let me see if I can move property tracking to the react scope only

PR #5227, opened 2023-04-05
fix(query-core): correct placeholderData prevData value with select fn

ardeora:
Howdy! 
This PR fixes an issue in v5 when we use placeholderData fn in conjunction with the select fn. The current bug 
is that the previousData available in the params of the placeholderData fn is the selected data from the previous query and not the cached data from the previous query. This causes an issue that would be better explained with the following example:

 https://codesandbox.io/s/gallant-gates-dslrlm?file=/src/App.js

I have also added one test case that shows the desired (fixed) behavior. Please let me know if I can add any other test cases to make this bulletproof! 

Thanks,
Aryan

### DIFF
--- a/packages/query-core/src/queryObserver.ts
+++ b/packages/query-core/src/queryObserver.ts
@@ -64,7 +64,6 @@ export class QueryObserver<
     TQueryData,
     TQueryKey
   >
-  #previousQueryResult?: QueryObserverResult<TData, TError>
   #selectError: TError | null
   #selectFn?: (data: TQueryData) => TData
   #selectResult?: TData
@@ -414,9 +413,6 @@ export class QueryObserver<
     const queryInitialState = queryChange
       ? query.state
       : this.#currentQueryInitialState
-    const prevQueryResult = queryChange
-      ? this.#currentResult
-      : this.#previousQueryResult
 
     const { state } = query
     let { error, errorUpdatedAt, fetchStatus, status } = state
@@ -490,7 +486,7 @@ export class QueryObserver<
           typeof options.placeholderData === 'function'
             ? (
                 options.placeholderData as unknown as PlaceholderDataFunction<TQueryData>
-              )(prevQueryResult?.data as TQueryData | undefined)
+              )(prevResultState?.data)
             : options.placeholderData
         if (options.select && typeof placeholderData !== 'undefined') {
           try {
@@ -619,7 +615,6 @@ export class QueryObserver<
       | undefined
     this.#currentQuery = query
     this.#currentQueryInitialState = query.state
-    this.#previousQueryResult = this.#currentResult
 
     if (this.hasListeners()) {
       prevQuery?.removeObserver(this)
--- a/packages/query-core/src/queryObserver.ts
+++ b/packages/query-core/src/queryObserver.ts
@@ -483,7 +483,6 @@ export class QueryObserver<
         prevResult?.isPlaceholderData &&
         options.placeholderData === prevResultOptions?.placeholderData
       ) {
-        console.log('memoized placeholder data')
         placeholderData = prevResult.data
       } else {
         placeholderData =
--- a/packages/query-core/src/queryObserver.ts
+++ b/packages/query-core/src/queryObserver.ts
@@ -64,10 +64,12 @@ export class QueryObserver<
     TQueryData,
     TQueryKey
   >
-  #previousQueryResult?: QueryObserverResult<TData, TError>
   #selectError: TError | null
   #selectFn?: (data: TQueryData) => TData
   #selectResult?: TData
+  // This property keeps track of the last defined query data.
+  // It will be used to pass the previous data to the placeholder function between renders.
+  #lastDefinedQueryData?: TQueryData
   #staleTimeoutId?: ReturnType<typeof setTimeout>
   #refetchIntervalId?: ReturnType<typeof setInterval>
   #currentRefetchInterval?: number | false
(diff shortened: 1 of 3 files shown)


TkDodo (approved):

TkDodo:
your fix looks good, but interestingly, in the sandbox, even with the fixed version, we see a render cycle where data is undefined:

https://codesandbox.io/s/admiring-mahavira-w3zgzd

I'm not really sure why that would happen :/

TkDodo:
hm I think I got something: the `prevResultState` tracks the previous render basically, so once after we've selected and kept previous data, the `prevResultState` is now `pending` with empty data. With the next render. The `prevQueryResult` that we had, on the other hand, was only changing when a query update came in.

So I think maybe we still need to keep the `prevQueryResult` value, but make sure it doesn't store "selected" results, but raw results only?

ardeora:
@TkDodo Thanks for the code review! I think I have a fix in place that makes this example works now 🚀  

https://codesandbox.io/s/romantic-moon-lrkjge?file=/package.json

Please let me know if there is anything else that I can improve on this PR. 

Thanks so much!
Hope you have a great weekend!

ardeora:
Also added a test to `react-query` that uses select and placeholderData function together!

TkDodo:
I can sadly still see a difference to v4:

if we click twice on `next`, the `data` will switch to `Loading...` on the second click.

I think it's because if I go from id1 to id2, it keeps data of id1. But then I switch to id3, it tries to keep data from id2, which is still `undefined`, so we get a loading state ...

This wasn't the case in v4. I can click on next as often as I want, and it will never go into an intermediate loading state ...

TkDodo:
we can see the issue if we log in the `placholderData` function:

```
placeholderData: (prev) => {
  console.log(prev);
  return prev;
},
```

if I click the button once, it will log the prev data correctly. If I click it again while we're still loading (and showing previous data), I will get `undefined` passed.

---

I think the `prevQueryResult` is only used for passing to `placeholderData`. I compared this with `main`, and this is the same property that we used for `keepPreviousData`:

https://github.com/TanStack/query/blob/317fe0822606590c08a9e3ce0952ab785ae34912/packages/query-core/src/queryObserver.ts#L475-L478

so, I don't really understand why this can work on main but not on alpha 🤔

ardeora:
It doesn't work on main too! If you remove keepPreviousData and just keep `select` and `placeholderData` you will see the same issue on v4 too!

ardeora:
https://github.com/TanStack/query/blob/317fe0822606590c08a9e3ce0952ab785ae34912/packages/query-core/src/queryObserver.ts#L475-L478
The code snippet you shared above it works because with keepPreviousData the status is set to success! 
https://github.com/TanStack/query/blob/317fe0822606590c08a9e3ce0952ab785ae34912/packages/query-core/src/queryObserver.ts#L513 and the placeholderData is never run! So we just keep using the previously selected data between renders!

TkDodo:
> If you remove keepPreviousData and just keep select and placeholderData you will see the same issue on v4 too! 😅

What I mean is that on main, it works with `keepPreviousData: true`:

https://codesandbox.io/s/focused-dawn-pnis3t?file=/src/App.js

Here, you can click the next button as often as you want and you will never see a `loading` state.

If we do the same thing in v5, which means, replacing:

```diff
- keepPreviousData: true
+ placeholderData: (prev) => prev
```

which should yield the same result, but it doesn't, as when you click twice, it will go to a loading state. Or is that not what you are observing, too?

ardeora:
> > If you remove keepPreviousData and just keep select and placeholderData you will see the same issue on v4 too! 😅
> 
> What I mean is that on main, it works with `keepPreviousData: true`:
> 
> https://codesandbox.io/s/focused-dawn-pnis3t?file=/src/App.js
> 
> Here, you can click the next button as often as you want and you will never see a `loading` state.
> 
> If we do the same thing in v5, which means, replacing:
> 
> ```diff
> - keepPreviousData: true
> + placeholderData: (prev) => prev
> ```
> 
> which should yield the same result, but it doesn't, as when you click twice, it will go to a loading state. Or is that not what you are observing, too?

Yes I see the same result as you! It's just that in v4 we never run select/placeholderData functions again and just keep passing the previous result.data to the next render! See https://github.com/TanStack/query/pull/5227#issuecomment-1512056418

With just placeholderData we cant use the previousResult.data because it was the last selected value. So we somehow need to keep track of the raw value somewhere to make this work

TkDodo:
> With just placeholderData we cant use the previousResult.data because it was the last selected value. So we somehow need to keep track of the raw value somewhere to make this work

oh, I see, so this is still select specific? If that's the case, yes, storing the last select result and re-using that should fix it, if we have both placeholderData and select defined.

---

I don't know why, but I also can't get it to work with just `placeholderData`, without select:

https://codesandbox.io/s/beautiful-pine-70shth?file=/src/App.js

clicking next twice will show the loading state.

and interestingly, this exact example works if I switch back to a previous alpha version:

https://codesandbox.io/s/upbeat-alex-gkmwcl?file=/src/App.js

🤔

TkDodo:
I have to say this is pretty puzzling. From what I can tell, on current alpha, every scenario works without select.

But I don't think we can just solve it by storing the previous select result, because we can't pass the previous select result to `placeholderData`. The idea is that it always runs through the placeholderData function, and then through select..

TkDodo:
>  The current bug
is that the previousData available in the params of the placeholderData fn is the selected data from the previous query and not the cached data from the previous query.

okay, I see now, we already store the selected result on alpha, and that is actually the source of the bug. What we could do on `main` is decide based on the flag and just return that previously selected data. But now, we can't do that: we have to gain access to the "last non-undefined raw data" somehow, pass that through `placeholderData` function, and then (potentially) through select again

ardeora:
Yep! Makes sense. Thanks for confirming! This might require a lot more refactor. I'll take a look at this again tomorrow morning/afternoon to see if I can make that change 😅

TkDodo:
I'll test it later today. Thank you for working on this ❤️

TkDodo:
this is great, thank you 🙏

PR #5240, opened 2023-04-07
fix: do not mutate input

TkDodo:
params passed to function from consumers should be treated as readonly and thus shouldn't be modified


TkDodo:
@DamianOsipiuk I think the test failure here is legit, but I'm not sure how to fix it, as it's vue related 😅 . Can you please have a look?

DamianOsipiuk (approved):

Issue #5059, opened 2023-03-01
SSR & SvelteKit prefetching example doesn't work as expected


universaltonton:
### Describe the bug

The docs (https://tanstack.com/query/latest/docs/svelte/ssr#using-prefetchquery) says that when using `prefetchQuery` in `+page.ts`, a refetch won't be triggered when calling `createQuery` in `+page.svelte` (with the same query key of course).

After following the documentation, it seems it doesn't work that way and a refetch does happen.

The example provided in documentation also clearly show this behaviour as shown in the screenshots below (the 3rd "posts" in the 2nd screenshots comes from `refetchOnWindowFocus: true`).

If I set `staleTime: Infinity` then the 2nd fetch doesn't happen but then `refetchOnWindowFocus` doesn't get triggered anymore.

### Your minimal, reproducible example

https://codesandbox.io/s/github/tanstack/query/tree/main/examples/svelte/ssr?from-embed

### Steps to reproduce

Just follow the example at https://tanstack.com/query/latest/docs/svelte/examples/svelte/ssr

### Expected behavior

The query should not run a 2nd time

### How often does this bug happen?

Every time

### Screenshots or Videos

![image](https://user-images.githubusercontent.com/126693611/222182664-35c5ae6a-ba5f-4ccb-8f76-bae9b33fdbd1.png)

![image](https://user-images.githubusercontent.com/126693611/222182730-622e3165-2fe8-4ffe-af10-3c8bc0c369b7.png)

### Platform

- OS: macOS 13.0.1 
- Browser : all of them

### TanStack Query version

4.24.10

### TypeScript version

4.9.3

liborvanek:
I believe this is expected behaviour – according to the [documentation](https://tanstack.com/query/v4/docs/react/guides/ssr#staleness-is-measured-from-when-the-query-was-fetched-on-the-server), queries will be refetched in the background on page load because the `staleTime` is set to `0` by default. 
To avoid immediate refetch, set a higher value.

universaltonton:
Well that's kind of misleading then: the documentation for React and for Svelte seems to be contradictory.

The Svelte doc says:

_Svelte Query supports prefetching queries on the server. Using this setup below, you can fetch data and pass it into QueryClientProvider before it is sent to the user's browser. Therefore, this data is already available in the cache, and no initial fetch occurs client-side._

The code from the doc doesn't set `staleTime` at all.

lachlancollins:
[...]

PR #5243, opened 2023-04-08
docs(svelte-query): Fix SSR example and update svelte templates

lachlancollins:
This uses the solution posted by @bevm0 on #5059. By using the load function's fetch, the server result is properly shared with the client code. The general template generated from `create-svelte` has also been updated since a lot has changed since v1.0.


examples/svelte/ssr/src/lib/data.ts
```diff
@@ -1,13 +1,22 @@
 import type { Post } from './types'
+import type { RequestEvent } from '@sveltejs/kit'
 
-export const getPosts = async (limit: number) => {
-  const response = await fetch('https://jsonplaceholder.typicode.com/posts')
+export const getPosts = async (
+  limit: number,
+  customFetch: RequestEvent['fetch'] = fetch,
```
bevm0, line 6:
```suggestion
  customFetch: typeof fetch = fetch,
```
I think SvelteKit's fetch function satisfies this type, so maybe it's more succinct to do this instead of adding an extra import?

On a side note, should any of the documentation be updated as well?

lachlancollins: Thanks, I've fixed the type and updated the docs :)

PR #5253, opened 2023-04-11
refactor: enhance error undefined message

DylanJu:
In production, it's hard to debug if got undefined data because error message don't tell which query is the problem.
To find the query or api, a developer should look many network logs. It's hard things and makes spend a lot of time.

when undefined error occurs production only, there is no clue which one is broken.

If throw "undefined" error with which query is the problem it makes easy to debug and save a lot of time.
Actually, this is from my experience. https://discord.com/channels/719702312431386674/1094894443447980152/1094894443447980152


packages/query-core/src/query.ts
```diff
@@ -463,7 +463,7 @@ export class Query<
               `Query data cannot be undefined. Please make sure to return a value other than undefined from your query function. Affected query key: ${this.queryHash}`,
             )
           }
-          onError(new Error('undefined') as any)
+          onError(new Error(`${this.queryHash} data is undefine11d`) as any)
```
TkDodo, line 466:
```suggestion
          onError(new Error(`${this.queryHash} data is undefined`) as any)
```

TkDodo:
you need to adapt the test that checks for this message please

DylanJu:
fixed [5d26428](https://github.com/TanStack/query/pull/5253/commits/5d264285a78ee2daf630f9a5bb17f485ea5dab0d)

PR #5255, opened 2023-04-11
perf(useQueries): optimize excessive invocations of getQueries in useQueries

AEmad01:
This fixes an issue with excessive invocations of `getQueries` inside `useQueries`. Currently it is being invoked on every single iteration on `optimisticResult.find` which is also the same size as the queries. `getQueries` also maps over all queries so in theory, this function is O(n*n) which can be optimized by moving the `getQueries` call just outside `.find`, a 2 line fix.

Measurement:
When using `useQueries` with a high number of queries, (20,000). The performance profiler indicates that we spend 4.86 seconds in `useQueries`, which is painfully slow.

Screenshot 2023-04-11 at 20 31 11

The flamegraph right underneath useQueries, indicates that we call `getQueries` 20,000 times in this example, and also indicates the exact line that causes the huge execution time, although a bit misleading because `.map` cannot be this slow, unless we use it in another loop which is what happens as we call it inside a `.find`.

Screenshot 2023-04-11 at 20 30 58

After the fix:

There is only a single invocation of `getQueries` and the execution time is down to 272ms, for 20k queries.

Screenshot 2023-04-11 at 20 25 56
 
and the profiler also confirms the execution time of `getQueries`
Screenshot 2023-04-11 at 20 56 06


TkDodo (approved):
thanks 🙏

PR #5256, opened 2023-04-11
Adding DTE to the PR workflow

ZackDeRose:
## What it does

Uses distributed task execution for the PR workflow - this way we can get better utilization out of the machines we use for the pr workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Run Tests
         uses: nick-fields/retry@v2.8.3
         with:
-          command: pnpm run test:ci --base=${{ github.event.before }}
+          command: pnpm run test:ci --base=${{ github.event.before }} --skip-nx-cache
           timeout_minutes: 10
           max_attempts: 3
       - name: Publish
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -56,6 +56,8 @@ jobs:
           echo "HEAD: ${{ env.NX_HEAD }}"
       - name: Run All Checks
         uses: nick-fields/retry@v2.8.3
+        env:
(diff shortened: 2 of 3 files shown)


ZackDeRose:
Hey @TkDodo! Anything else needed in this PR?

I'd estimate this is roughly a 10-25% improvement to the pr workflow timings - which is still good, but I think adding the task caching + filtering to only run affected tasks was an order of magnitude more impactful

.github/workflows/pr.yml
```diff
@@ -5,9 +5,13 @@ env:
   NX_VERBOSE_LOGGING: true
```
TkDodo, line 5: can we disable the verbose logging again? it's a lot of logs - is it necessary ?

.github/workflows/ci.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       - name: Run Tests
         uses: nick-fields/retry@v2.8.3
         with:
-          command: pnpm run test:ci --base=${{ github.event.before }}
+          command: pnpm run test:ci --base=${{ github.event.before }} --skip-nx-cache
```
TkDodo, line 42: I remember discussing this for the actual build, which we do:

https://github.com/TanStack/query/blob/eff7f03c44c91d80be1375e9686cf41f9d48a17d/scripts/publish.ts#L360

but this is just running tests. I think we don't need to skip cache here

.github/workflows/pr.yml
ZackDeRose: For sure!

ZackDeRose:
## Time Investigation

With DTE: 2m48s long stick (https://github.com/TanStack/query/actions/runs/4805517871/jobs/8552011511?pr=5256)
Without DTE: 3m21s long stick (https://github.com/ZackDeRose/query/actions/runs/4805602190/jobs/8552179929)

Result: 16.4% time improvement ((201 - 168) / 201)

(used with cold caches to preserve like-for-like comparisons)

PR #5274, opened 2023-04-14
docs: update migrating-to-v5

arthurdenner:
I think the rocket emoji is causing a visual bug too but since I edited this through GitHub's UI, I couldn't confirm.

image


TkDodo:
the visual bug comes from two headings being directly underneath each other. I've seen this before - could be something how TanStack docs / remix treat this, idk.

Simple fix would be to add a line of text between the headings, like "We have some great new features for v5" or something like that 😅

PR #5276, opened 2023-04-15
Add `Hydrate` + `useHydrate` to Solid Query

oscartbeaumont:
The title pretty much says it all. These are direct clones of the functionality that already exists in React Query and Svelte Query.

### DIFF
--- a/packages/solid-query/src/Hydrate.ts
+++ b/packages/solid-query/src/Hydrate.ts
@@ -1,0 +1,27 @@
+import type { JSX } from 'solid-js'
+import type { HydrateOptions } from '@tanstack/query-core'
+import { hydrate } from '@tanstack/query-core'
+import { useQueryClient } from './QueryClientProvider'
+import type { ContextOptions } from './types'
+
+export function useHydrate(
+  state: unknown,
+  options: HydrateOptions & ContextOptions = {},
+) {
+  const queryClient = useQueryClient({ context: options.context })
+
+  if (state) {
+    hydrate(queryClient, state, options)
+  }
+}
+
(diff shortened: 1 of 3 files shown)


ardeora:
Hello! Thanks for working on this! So in solid we're using resources to track the data of the query observers in v4. In v5 the whole queryObserver result is tracked under a resource! This gives us the ability to hydrate and dehydrate queries using the native solid-js resource! So I don't think there would be a need for `Hydrate` and `useHydrate` going forward! The v4 branch can certainly be updated to use the improvements made in the v5 branch! Would you want to update this PR to follow the patterns defined in the `alpha` branch? The hydration logic is shown here -> https://github.com/TanStack/query/blob/alpha/packages/solid-query/src/createBaseQuery.ts#L157-L186

TkDodo:
FYI, we've removed `useHydrate` in v5 and only have a `HydrationBoundary` component now. Might not make sense to introduce it like that and then break it again in v5 ?

oscartbeaumont:
That's completely fair.

I wanted these utilities because I was trying to implement my own SSR stuff on v4 due to problems I had getting the v5 beta working but after doing this PR I ended up sorting out the issues I had with v5 and now I can definitely see how these utilities don't make sense.

The stuff you guys have done with SSR + streaming in v5 is incredible!

For context, I was having problems with the beta `5.0.0-alpha.18` release as it doesn't work due to missing files in the `dist` folder which the `package.json` refers to. I *think* this has been fixed `main` because it built fine locally. I have switched to using release `5.0.0-alpha.16` which I saw @ardeora using in one of his repos.

PR #5280, opened 2023-04-17
Update vue basic example link to v5

shonya3:
v4 was showing "Initializing Sandbox Containe" forever, so i tried to update link to v5, and it works fine.


TkDodo:
it works fine for me now. could've been an upstream issue with codesandbox. linking to v5 is definitely not correct, as that version is currently in alpha.

PR #5284, opened 2023-04-17
feat(solid-query): Better queryKey types support

ardeora:
This change adds narrowed types for queryKey in createQuery.

Before:
```ts
const [count, setCount] = createSignal(1);
  queryKey: ['pokemon', count()]
  queryFn: async ({ queryKey }) => {
     const [_, id] = queryKey;
     //              ^? queryKey: (string|number)[]
     return fetch(...)
  }
}))
```

After:
```ts
const [count, setCount] = createSignal(1);
  queryKey: ['pokemon', count()]
  queryFn: async ({ queryKey }) => {
     const [_, id] = queryKey;
     //              ^? queryKey: ['pokemon', number]
     return fetch(...)
  }
}))
```


ardeora:
Need more work! Will create a PR later

PR #5287, opened 2023-04-19
feat(solid-query): Add `reconcile` option

ardeora:
This pull request proposes a new feature for `solid-query` that allows for more precise updates to query data. The new `reconcile` option enables updates to be made between state updates, while the structuralSharing option is removed since it doesn't work well with `solid-js'` data diffing patterns.

### Before
shapes

Consider this following example that fetches data for a users list with createQuery.
```tsx
interface User {
  id: number
  name: string
  last_online: string
}

const usersList = createQuery(() => ({
  queryKey: ['users_list'],
  queryFn: async () => {
    const users = await fetch(`${API_URL}/users`).then(res => res.json()) 
    return users as User[]
  },
}))

return (
  <div>
    <button onClick={usersList.refetch}>Refetch</button>
    <For each={usersList.data}>
      {user => (
        <div>
          <div>{user.name}</div>
          <div>{user.last_online}</div>
        </div>
      )}
    </For>
  </div>
)
```
Currently, every time a user clicks the `Refetch` button, the `queryFn` is re-run, and a new array reference of the updated users list is returned. As a result, new DOM nodes are created every time the query data updates, which can quickly become expensive.

With the proposed feature, whenever the query data updates, the previous data will be transformed into the shape of the new data while preserving referential equality of the data array between updates. This will allow for fine-grained updates to the already created DOM nodes.

### After

shapes (1)

```tsx
interface User {
  id: number
  name: string
  last_online: string
}

const usersList = createQuery(() => ({
  queryKey: ['users_list'],
  queryFn: async () => {
    const users = await fetch(`${API_URL}/users`).then(res => res.json()) 
    return users as User[]
  },
  reconcile: 'id', // Not needed since reconcile defaults to `id`
}))

return (
  <div>
    <button onClick={usersList.refetch}>Refetch</button>
    <For each={usersList.data}>
      {user => (
        <div>
          <div>{user.name}</div>
          <div>{user.last_online}</div>
        </div>
      )}
    </For>
  </div>
)
```

The diagram illustrates how the reconcile option can enable fine-grained updates. Whenever the `queryFn` is re-run, the `id` reconcile key can help solid-query identify which objects in the array have been updated, allowing for precise updates to those objects while preserving their referential equality.

This feature can be compared to the `key` prop used in React, which allows for efficient updates of lists by identifying unique keys for each element. The `reconcile` option in solid-query works similarly but operates at a more granular level, allowing for precise updates to individual elements within an array.

Overall, this feature can significantly improve the efficiency of applications and reduce the cost of updates by enabling precise and efficient updates to query data.

## API reference
The `reconcile` option in solid-query can be used in multiple ways.

### Option 1: `reconcile` key
```ts
interface Fruit {
  name: string;
  emoji: string;
}

const fruits = createQuery(() => ({
  queryKey: ['fruits'],
  queryFn: async () => {
    const fruitsResponse = await fetch(`${API_URL}/fruits`).then(res => res.json()) 
    return fruitsResponse as Fruit[]
  },
  reconcile: 'name',
}))
```
Similar to the example above. The `reconcile` option can take any `string` value that will act as a key for the data returned by `createQuery`. In the example above, the reconcile option is set to `name`, which means that `solid-query` will use the `name` property of each fruit object to reconcile updates to the query data.

## Option 2: `reconcile` function
```ts
import { produce } from "solid-js/store";

interface Fruit {
  name: string;
  emoji: string;
}

const fruits = createQuery(() => ({
  queryKey: ['fruits'],
  queryFn: async () => {
    const fruitsResponse = await fetch(`${API_URL}/fruits`).then(res => res.json()) 
    return fruitsResponse as Fruit[]
  },
  reconcile(oldData, newData) {
    if (!oldData) return newData;
    return produce((draft: Fruit[]) => {
      draft.forEach((fruit) => {
        const newFruit = newData.find(f => f.name === fruit.name)
        if (newFruit) {
          fruit.emoji = newFruit.emoji
        }
      })
    })(oldData)
  },
}))
```
Another way to use the `reconcile` option in solid-query is by providing a callback function that receives both the previous data and the new data returned by createQuery. This function allows users to provide their own reconciliation logic based on the data being updated.

In the code snippet above, the `reconcile` function takes two arguments: oldData and newData. If there is no `oldData`, the function returns the `newData` as is. If there is `oldData`, the function uses the `produce` function from the `solid-js/store` library to create a mutable copy of the old data. It then iterates over each fruit object in the old data, checks if there is a matching fruit object in the new data based on the name property, and updates the emoji property of the old object if a match is found.

## Option 3: disable `reconcile` logic
```ts
interface Fruit {
  name: string;
  emoji: string;
}

const fruits = createQuery(() => ({
  queryKey: ['fruits'],
  queryFn: async () => {
    const fruitsResponse = await fetch(`${API_URL}/fruits`).then(res => res.json()) 
    return fruitsResponse as Fruit[]
  },
  reconcile: false,
}))
```
If users don't want to perform fine-grained updates to the data, they can disable the `reconcile` option by setting it to `false`. This will bypass any reconciliation logic and perform immutable updates between query updates.

I believe this change makes `solid-query` even more powerful and intuitive by allowing users to optimize the performance of their applications while providing a smooth user experience.

As always, we encourage everyone to test out this new feature and provide feedback and concerns in the comments. Thank you for your contributions!

### DIFF
--- a/packages/solid-query/src/QueryClient.ts
+++ b/packages/solid-query/src/QueryClient.ts
@@ -1,0 +1,84 @@
+import type {
+  QueryClientConfig as QueryCoreClientConfig,
+  DefaultOptions as CoreDefaultOptions,
+  QueryObserverOptions as QueryCoreObserverOptions,
+  InfiniteQueryObserverOptions as QueryCoreInfiniteQueryObserverOptions,
+  DefaultError,
+  QueryKey,
+} from '@tanstack/query-core'
+import { QueryClient as QueryCoreClient } from '@tanstack/query-core'
+
+export interface QueryObserverOptions<
+  TQueryFnData = unknown,
+  TError = DefaultError,
+  TData = TQueryFnData,
+  TQueryData = TQueryFnData,
+  TQueryKey extends QueryKey = QueryKey,
+  TPageParam = never,
+> extends Omit<
+    QueryCoreObserverOptions<
+      TQueryFnData,
+      TError,
+      TData,
+      TQueryData,
+      TQueryKey,
+      TPageParam
+    >,
+    'structuralSharing'
+  > {
+  /**
+   * Set this to a reconciliation key to enable reconciliation between query results.
+   * Set this to `false` to disable reconciliation between query results.
+   * Set this to a function which accepts the old and new data and returns resolved data of the same type to implement custom reconciliation logic.
+   * Defaults reconciliation key to `id`.
+   */
+  reconcile?:
+    | string
+    | false
+    | ((oldData: TData | undefined, newData: TData) => TData)
+}
+
+export interface InfiniteQueryObserverOptions<
+  TQueryFnData = unknown,
+  TError = DefaultError,
+  TData = TQueryFnData,
+  TQueryData = TQueryFnData,
+  TQueryKey extends QueryKey = QueryKey,
+  TPageParam = unknown,
+> extends Omit<
+    QueryCoreInfiniteQueryObserverOptions<
+      TQueryFnData,
+      TError,
+      TData,
+      TQueryData,
+      TQueryKey,
+      TPageParam
+    >,
+    'structuralSharing'
+  > {
+  /**
+   * Set this to a reconciliation key to enable reconciliation between query results.
+   * Set this to `false` to disable reconciliation between query results.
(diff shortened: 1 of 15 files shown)


ardeora:
Here is an example of it in action! https://stackblitz.com/edit/github-ubelji-feyyo3?file=src%2FApp.tsx

PR #5291, opened 2023-04-19
fix(eslint-plugin): ignore queryClient from exhaustive-check

Newbie012:
as discussed in https://twitter.com/dw60010009/status/1648703272763670528


dcheng1290 (approved):

packages/eslint-plugin-query/src/rules/exhaustive-deps/exhaustive-deps.test.ts
```diff
@@ -298,6 +298,23 @@ ruleTester.run('exhaustive-deps', rule, {
...
+      code: `
+        const CONST_VAL = 1
+        function useHook() {
+          const queryClient = useQueryClient()
+          const kueryKlient = useQueryClient()
+          useQuery({
+            queryKey: ["foo"],
+            queryFn: () => {
+                doSomething(queryClient)
+                doSomethingSus(kueryKlient)
```
TkDodo, line 312: I just want to see if invoking a method on the queryClient is also fine:

```suggestion
                doSomething(queryClient)
                queryClient.invalidateQueries()
                doSomethingSus(kueryKlient)
```

Newbie012: We already have a test case for that, but why not 🤷‍♂️

PR #5301, opened 2023-04-20
feat(eslint-plugin): prefer-object-syntax support fetchQuery

gterral:
This pull request updates the `prefer-object-syntax` eslint rule to add support for the fetchQuery method 

I tried to follow the [Contributing guide](https://github.com/TanStack/query/blob/main/CONTRIBUTING.md), please let me know if I need to change something.

Related discussion => https://github.com/TanStack/query/discussions/5300

Thank you for the awesome library ❤️


packages/eslint-plugin-query/src/rules/prefer-query-object-syntax/prefer-query-object-syntax.ts
```diff
@@ -7,6 +7,7 @@ import { objectKeys } from '../../utils/object-utils'
 const QUERY_CALLS = {
   useQuery: { key: 'queryKey', fn: 'queryFn', type: 'query' },
   createQuery: { key: 'queryKey', fn: 'queryFn', type: 'query' },
+  fetchQuery: { key: 'queryKey', fn: 'queryFn', type: 'query' },
```
TkDodo, line 10: the difference is: `fetchQuery` comes from `queryClient`, so it would need to be `queryClient.fetchQuery`. I'd want to avoid false-positives from other, random methods called `fetchQuery`. @Newbie012 FYI

Newbie012: this rule was initially written in mind to support only `useQuery`. Since then, other hooks needed to be supported, so I refactored the rule in a way of identifying only hooks (which is not the case for `queryClient.fetchQuery`. The identification of `fetchQuery` should be different than the other ways:

1. Look for `fetchQuery` call expression.
2. Track down its source and check if it came from `useQueryClient` (I already wrote a function that tells if a given reference is a queryClient. It's not exactly the same issue as here, but it's a good starting point):

https://github.com/TanStack/query/blob/b369b6c9fd289b66646c53b793785c9c0899497d/packages/eslint-plugin-query/src/rules/exhaustive-deps/exhaustive-deps.utils.ts#L31-L40

3. Apply rule checks if positive.

This way you don't have any false positives. Doesn't matter if you have a different name than `queryClient` or destructure `fetchQuery` from it.

gterral: Very clear thanks! I will try to give it a try in the upcoming days!

TkDodo: @gterral any updates here? This is pretty stale and  I'd close it if you're not working on this anymore, since the rule won't exist in v5.

gterral: Yep, sorry didn't took the time to work on it yet as this wasn't really urgent on my side. 
(And just noticed it's been almost 2 months already 😱). 

I don't think I will able to work on it during the upcoming week, so I can understand if you prefer to close the issue rather keeping it open, but if you do let it open, I promise that I will do it at one point! 

Hopefully in the 3/4 weeks. (July should be quieter on my side).

TkDodo: it could very well be that we have v5 out by then 😅

gterral: Ok!

Sorry and good luck for the v5 release then ;)!

PR #5304, opened 2023-04-21
fix(useQuery): add support for  RN such as useSyncExternalStore in ESM mode

PFanz:
The original code imports the useSyncExternalStore from the web version in ESM mode, which causes the useQuery to fail to send requests in React Native.
image

Modify the ESM code to be placed in a unified directory without the .esm suffix, so that React Native can correctly find the .native.js file.

### DIFF
--- a/packages/react-query/package.json
+++ b/packages/react-query/package.json
@@ -12,11 +12,11 @@
   },
   "types": "build/lib/index.d.ts",
   "main": "build/lib/index.js",
-  "module": "build/lib/index.esm.js",
+  "module": "build/esm/index.js",
   "exports": {
     ".": {
       "types": "./build/lib/index.d.ts",
-      "import": "./build/lib/index.mjs",
+      "import": "./build/esm/index.js",
       "default": "./build/lib/index.js"
     },
     "./package.json": "./package.json"
--- a/rollup.config.ts
+++ b/rollup.config.ts
@@ -302,18 +302,18 @@ function esm({
 }: Options): RollupOptions {
   const bundleOutput: OutputOptions = {
     format: 'esm',
-    file: `${packageDir}/build/lib/${outputFile}.esm.js`,
+    file: `${packageDir}/build/esm/${outputFile}.js`,
     sourcemap: true,
     banner,
   }
 
   const normalOutput: OutputOptions = {
     format: 'esm',
-    dir: `${packageDir}/build/lib`,
+    dir: `${packageDir}/build/esm`,
     sourcemap: true,
     banner,
     preserveModules: true,
-    entryFileNames: '[name].esm.js',
+    entryFileNames: '[name].js',
   }
 
   return {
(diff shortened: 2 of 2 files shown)


TkDodo:
Idk, @DamianOsipiuk has spent quite some time to get esm support "right". I'd need him to verify that this change is not breaking for any other bundlers

PFanz:
> 

In addition to ESM files, there are also similar issues with CJS files. I suggest separating files of different types.

packages/react-query/package.json
```diff
@@ -12,11 +12,11 @@
   },
   "types": "build/lib/index.d.ts",
   "main": "build/lib/index.js",
-  "module": "build/lib/index.esm.js",
+  "module": "build/esm/index.js",
   "exports": {
     ".": {
       "types": "./build/lib/index.d.ts",
-      "import": "./build/lib/index.mjs",
```
DamianOsipiuk, line 19: `mjs` is required for modern node versions and should be left as-is

rollup.config.ts
```diff
@@ -302,18 +302,18 @@ function esm({
 }: Options): RollupOptions {
   const bundleOutput: OutputOptions = {
     format: 'esm',
-    file: `${packageDir}/build/lib/${outputFile}.esm.js`,
```
DamianOsipiuk, line 305: All build artifacts were placed in one directory due to `exports` and `typescript` interop.
There were some problems loading types with separate folders setup.

DamianOsipiuk (commented):
When we did those changes initially we checked with React Native users and it seemed to work fine for them.

Therefore i'm wondering.
- Do you have some non-standard setup in place?
- Are you able to update your build libraries to newer versions?
- Did it start breaking after some updates and have worked before?

PFanz:
> When we did those changes initially we checked with React Native users and it seemed to work fine for them.当我们最初进行这些更改时，我们与React Native用户进行了检查，似乎对他们来说效果很好。
> 
> Therefore i'm wondering.所以我想知道。
> 
> * Do you have some non-standard setup in place?你有一些非标准的设置吗？
> * Are you able to update your build libraries to newer versions?您是否能够将构建库更新到较新的版本？
> * Did it start breaking after some updates and have worked before?它是否在一些更新后开始中断，并且以前工作过？

Not have some non-standard setup。
In the current build, useBaseQuery.js imports useSyncExternalStore which has no file extension, so it automatically uses the useSyncExternalStore.native.js file in RN projects. However, in useBaseQuery.mjs and useBaseQuery.esm.js, useSyncExternalStore.mjs and useSyncExternalStore.esm.js are imported, which causes useSyncExternalStore to always use the file built from useSyncExternalStore.ts instead of useSyncExternalStore.native.ts.

packages/react-query/package.json
PFanz: Emmm, in that case, a better build approach may be needed here.

TkDodo:
This will implicitly be fixed with v5, because we're not using the uSES shim anymore.

PR #5309, opened 2023-04-22
Select with a transform may break when refetchInterval is set to a function

Nick-Lucas:
#5310 

This may affect more than just refetchInterval, it appears to be from usage of a `data` variable alongside select changing the type of that variable


TkDodo:
as we found out, this is a TypeScript limitation

PR #5314, opened 2023-04-23
docs(community): adding zodios usage and example

enyelsequeira:
# Changes

I think zodios is a great example that could be included in the `Community Resources`. It is easy to use and all with typesafety in mind.

### DIFF
--- a/docs/config.json
+++ b/docs/config.json
@@ -207,7 +207,7 @@
             },
             {
               "label": "Zodios",
-              "to": "react/community/zodios-by-ecrybe"
+              "to": "react/community/zodios-by-ecyrbe"
             }
           ]
         },
--- a/docs/config.json
+++ b/docs/config.json
@@ -204,6 +204,10 @@
             {
               "label": "Angular Query",
               "to": "react/community/angular-query"
+            },
+            {
+              "label": "Zodios",
+              "to": "react/community/zodios-by-ecrybe"
             }
           ]
         },
--- a/docs/react/community/zodios.md
+++ b/docs/react/community/zodios.md
@@ -1,5 +1,5 @@
 ---
-id: zodios by ecrybe
+id: zodios-by-ecrybe
 title: Zodios
 ---
 
--- a/docs/react/community/zodios.md
+++ b/docs/react/community/zodios.md
@@ -1,0 +1,170 @@
+---
+id: zodios by ecrybe
+title: Zodios
+---
+
+Zodios is a REST API toolbox with end-to-end typesafety, frontend and backend. It comes with some packages and plugins to make the data fetching a lot easier and what is best typesafe all the way to the query without having to remember query keys!
+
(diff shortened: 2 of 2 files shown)


TkDodo (changes requested):
you need to add an entry to `config.json`:

https://github.com/TanStack/query/blob/765f6196e900c1dd826b040773fcf97839ca60d0/docs/config.json#L190-L208

TkDodo:
@ecyrbe FYI :)

enyelsequeira:
> you need to add an entry to `config.json`:
> 
> https://github.com/TanStack/query/blob/765f6196e900c1dd826b040773fcf97839ca60d0/docs/config.json#L190-L208

Understood it has been added and changed id of .md

ecyrbe:
Hello, i don't think this is the way to integrate zodios in tanstack query docs.

Because Zodios can change interface, have breaking changes, etc and tanstack query will have a hard time to maintain these docs in sync.

A simple link to some Zodios docs would be enough in my point of view.

What do you think ?

enyelsequeira:
> Hello, i don't think this is the way to integrate zodios in tanstack query docs.
> 
> Because Zodios can change interface, have breaking changes, etc and tanstack query will have a hard time to maintain these docs in sync.
> 
> A simple link to some Zodios docs would be enough in my point of view.
> 
> What do you think ?

Hey, so I think every resource out there will have breaking changes. When the moment comes someone would just have to update the documentation for it. I think I’ve seen some other pull request updating docs as something changes as well. Just my thoughts though.

docs/config.json
```diff
@@ -204,6 +204,10 @@
             {
               "label": "Angular Query",
               "to": "react/community/angular-query"
+            },
+            {
+              "label": "Zodios",
+              "to": "react/community/zodios-by-ecrybe"
```
ecyrbe, line 210:
```suggestion
              "to": "react/community/zodios-by-ecyrbe"
```

docs/react/community/zodios.md
```diff
@@ -0,0 +1,170 @@
+---
+id: zodios-by-ecrybe
```
ecyrbe, line 2:
```suggestion
id: zodios-by-ecyrbe
```

ecyrbe (commented):
Some minor typo 😁

docs/config.json
enyelsequeira: oops, my mistake 😅, spelling has been fixed :)

TkDodo:
I've discussed with @ecyrbe and we've decided to go a little different approach: We'll draft one "Awesome React Query Collection" (similar to how [trpc](https://trpc.io/docs/community/awesome-trpc) does it) page where we'll list all community packages with links to docs, github and also when this might be a good fit. This seems better than a separate page for each, with a "quickstart" guide that might get outdated fast

PR #5318, opened 2023-04-24
feat: rename throwErrors to throwOnError

TkDodo:
to be aligned with the options we already have on imperative methods, like `invalidateQueries`


jasikpark (commented):
lgtm

PR #5322, opened 2023-04-25
docs(svelte-query): Use API wrapper to simplify SSR logic

lachlancollins:
Wrapper optionally accepts a fetch function. On the client this should be empty, while on server you need to pass the SvelteKit fetch function.


TkDodo:
@lachlancollins feel free to merge if you're happy with it :)

PR #5343, opened 2023-04-28
Add .vscode/settings.json

### LINKED ISSUE #5334: VSCode ESLint plugin thinks files are not included in TS project configs
### Describe the bug

In VSCode, a red squiggly line appears in the beginning of the first line of each file and when you hover over it, it displays the following error message:

```
Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
The file does not match your project config: packages/query-core/src/infiniteQueryBehavior.ts.
The file must be included in at least one of the projects provided.
```

What this usually means is that the file is not included in (any of) the `tsconfig.json` file(s) referenced by the `parserOptions.project` field. But everything seems to be correctly configured and `pnpm test:eslint` runs just fine, so it's a bit confusing to me why VSCode's ESLint plugin throws this error.

I also opened an issue at [vscode-eslint](https://github.com/microsoft/vscode-eslint) (https://github.com/microsoft/vscode-eslint/issues/1645) since it could be a bug with the plugin and not necessarily something that needs to be tweaked in Tanstack Query's project configuration.

N/A

### Steps to reproduce

1. Use VSCode with an empty profile.
2. Clone the repo into a [Node.js & TypeScript Dev Container](https://github.com/microsoft/vscode-dev-containers/tree/main/containers/typescript-node).
3. Install the VSCode ESLint plugin.
4. Open a terminal and run `pnpm install`.
5. Open any TypeScript file, e.g. `packages/query-core/src/infiniteQueryBehavior.ts`.
6. See red squiggly line in the beginning of the first line.

### Expected behavior

As a developer, I expected the VSCode ESLint plugin to work with this project, but I am seeing that it doesn't.

Every time

### Screenshots or Videos

CleanShot 2023-04-28 at 10 30 39@2x

### Platform

- OS: macOS
- Browser: N/A
- Version: N/A

VSCode version information:

```
Version: 1.77.3
Commit: 704ed70d4fd1c6bd6342c436f1ede30d1cff4710
Date: 2023-04-12T09:19:37.325Z
Electron: 19.1.11
Chromium: 102.0.5005.196
Node.js: 16.14.2
V8: 10.2.154.26-electron.0
OS: Darwin x64 22.4.0
Sandboxed: No
```

VSCode ESLint plugin version: 2.4.0

### Tanstack Query adapter

None

### TanStack Query version

v4.28.0

### TypeScript version

v4.8.3

### Additional context

I had set the `typescript.tsdk` setting to `4.8.3`, a.k.a. the same version of TypeScript as specified in the repo's root package.json, so it shouldn't be related to mismatching TypeScript versions or something like that.

CleanShot 2023-04-28 at 11 09 35@2x

DanielGiljam:
With settings that make VSCode ESLint work with the project.

Fixes #5334.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,7 +35,6 @@ size-plugin.json
 stats-hydration.json
 stats.json
 stats.html
-.vscode/settings.json
 
 *.log
 .DS_Store
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "eslint.workingDirectories": [{ "pattern": "./packages/*/" }]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,4 @@
 {
-  "eslint.workingDirectories": [{ "pattern": "./packages/*/" }]
+  "eslint.workingDirectories": [{ "pattern": "./packages/*/" }],
+  "typescript.tsdk": "node_modules/typescript/lib"
 }


TkDodo:
I don't understand why we would need to checkin a `.vscode` directory. Note that I don't use `vscode`, and IntelliJ also creates a directory of local settings that we also don't ever check into version control.

In all projects I've ever worked with, those local settings were gitignored because everyone tends to set it up for themselves differently ...

DanielGiljam:
The reason why we'd want to check in `.vscode/settings.json` is so that the VSCode ESLint extension would work for people who open up the project in VSCode. Currently, you're met with cryptic linting errors when opening the project in VSCode and linting doesn't work.

I've seen projects where they check in project settings for popular editors/IDEs.

Personally, I don't see the point of not doing it, since if someone wants to have their own custom project settings, this does not prevent that (they can configure their clone to ignore the file, if it bothers them that it then always shows up as "changed" in `git status`). This just ensures that out-of-the-box, the project plays nicely with VSCode. (If the project requires some project-specific settings to work nicely with IntelliJ, then I would advocate for checking in those settings as well.)

TkDodo:
> they can configure their clone to ignore the file, if it bothers them that it then always shows up as "changed" in git status

it would probably bother me as a maintainer if people then start just doing `git add .` and always sending PRs that will change those files.

regarding this comment:

- https://github.com/microsoft/vscode-eslint/issues/1645#issuecomment-1527492537

it's worth noticing that I'm not getting any error when I'm running the above mentioned command in the command line on the `alpha` branch, but it does error on `main`.

Maybe there is something we've fixed / configured differently in the meantime? Can you please checkout the `alpha` branch and see if it works there out of the box for you, too ?

RobertDS07:
@TkDodo I've tested here and in alpha is fixed! But I got another error:

image

TkDodo:
errors are not expected, but I don't get this error 🤷

TkDodo:
closing this because I don't think committing editor specific directories to version control is the way to go.

Issue #5336, opened 2023-04-28
When passing a custom behavior in `options.behavior` to the `InfiniteQueryObserver` constructor, an invalid `"observerOptionsUpdated"` event is fired


DanielGiljam:
### Describe the bug

When passing a custom behavior in `options.behavior` to the `InfiniteQueryObserver` constructor, an invalid `"observerOptionsUpdated"` event is fired.

This issue is related to https://github.com/TanStack/query/issues/5331.

The cause for this issue is that now the following block in `QueryObserver#setOptions` will run (`!shallowEqualObjects(...)` will evaluate to `true`)…

https://github.com/TanStack/query/blob/1614c31f84744c62b5bee8f2ff97fc15a1652e86/packages/query-core/src/queryObserver.ts#L169-L175

…since `prevOptions.behavior` now refers to a different object than `this.options.behavior` and `shallowEqualObjects` returns false if any of the first argument's properties differ from the second argument's corresponding properties (it ignores additional properties on the second argument which don't exist on the first argument).

`this.currentQuery` has not been assigned yet when `setOptions` is called for the first time from the constructor, so `event.query` ends up being `undefined`, which is unexpected since the `QueryCacheNotifyEvent` interface defines `query` as  always being defined.

### Your minimal, reproducible example

https://codesandbox.io/s/dank-resonance-7jsurp?file=/src/index.ts

### Steps to reproduce

1. Open the [reproducible example](https://codesandbox.io/s/dank-resonance-7jsurp?file=/src/index.ts) in CodeSandbox.
4. Witness error "Cannot read properties of undefined (reading 'queryHash')".

### Expected behavior

As a developer, I expect `query` to always be defined as the `QueryCacheNotifyEvent` interface suggests, but I am seeing that in this particular case, it will be `undefined`.

### How often does this bug happen?

Every time

### Platform

- OS: N/A
- Browser: N/A
- Version: N/A

### Tanstack Query adapter

None

### TanStack Query version

v4.27.0

TkDodo:
instead of opening new issues about `behavior`, can you please answer the question what you are trying to do with it first?

- https://github.com/TanStack/query/issues/5331#issuecomment-1527234478

PR #5344, opened 2023-04-28
Support user-specified `options.behavior` in `InfiniteQueryObserver`

### LINKED ISSUE #5331: `InfiniteQueryObserver` overrides `behavior` passed through `options`
### Describe the bug

When passing a `behavior` passed through `options` to `InfiniteQueryObserver` like this:

```js
new InfiniteQueryObserver(client, {behavior: customBehavior});
```

…it's overridden in `InfiniteQueryObserver`'s `setOptions` method:

https://github.com/TanStack/query/blob/b369b6c9fd289b66646c53b793785c9c0899497d/packages/query-core/src/infiniteQueryObserver.ts#L70-L87

https://codesandbox.io/s/old-cdn-u1sims?file=/src/App.js

### Steps to reproduce

1. Open the [reproducible example](https://codesandbox.io/s/old-cdn-u1sims?file=/src/App.js) in CodeSandbox.
2. Open the dev console.
3. Witness that the `customBehavior` is not applied.
    - It doesn't say "Hello from my custom behavior!" in the dev console.

### Expected behavior

As a developer, I expect my custom behavior to be applied, instead I'm seeing that my custom behavior is not applied.

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: 112.0

### Tanstack Query adapter

None

### TanStack Query version

v4.28.0

### TypeScript version

_No response_

### Additional context

_No response_

DanielGiljam:
Fixes #5331 and #5336.

### DIFF
--- a/packages/query-core/src/infiniteQueryBehavior.ts
+++ b/packages/query-core/src/infiniteQueryBehavior.ts
@@ -7,11 +7,11 @@ import type {
   RefetchQueryFilters,
 } from './types'
 
-export function infiniteQueryBehavior<
-  TQueryFnData,
-  TError,
-  TData,
->(): QueryBehavior<TQueryFnData, TError, InfiniteData<TData>> {
+export function infiniteQueryBehavior<TQueryFnData, TError, TData>(
+  behaviorFromOptions:
+    | QueryBehavior<TQueryFnData, TError, InfiniteData<TData>>
+    | undefined,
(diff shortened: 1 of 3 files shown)


TkDodo:
As said on the issue, `behaviour` is not an API we'd expect users to use. That's why it's undocumented.

I'd rather focus on fixing the issue you have with the behaviours we provide rather than make this API public.

PR #5350, opened 2023-04-30
chore: Validate "exports" field in package.json

lachlancollins:
According to [publint](https://publint.dev/@tanstack/svelte-query@4.29.5):

`This entrypoint has types at build/lib/index.d.ts but it is not exported from pkg.exports. Consider adding it to pkg.exports["."].types to be compatible with TypeScript's "moduleResolution": "bundler" compiler option. ([More info](https://publint.dev/rules.html#types_not_exported))`

This _was_ actually working with bundler mode because the `.d.ts` files are adjacent to the `.js` files. However, this is obviously more correct, and the other query packages do this already.

I've added a check to validate-packages.ts to look for exports['.']['default'], and also moved the entries logic into config.ts. In the future, it might be possible to add publint to this file - however I ran into a CJS/ESM error which I will investigate at a later time.

### DIFF
--- a/scripts/config.ts
+++ b/scripts/config.ts
@@ -61,7 +61,7 @@ export const packages: Package[] = [
     name: '@tanstack/svelte-query',
     packageDir: 'svelte-query',
     srcDir: 'src',
-    entries: [],
+    entries: ['module', 'svelte', 'types'],
   },
   {
     name: '@tanstack/vue-query',
--- a/scripts/config.ts
+++ b/scripts/config.ts
@@ -3,57 +3,77 @@ import type { BranchConfig, Package } from './types'
 
 // TODO: List your npm packages here. The first package will be used as the versioner.
 export const packages: Package[] = [
-  { name: '@tanstack/query-core', packageDir: 'query-core', srcDir: 'src' },
+  {
+    name: '@tanstack/query-core',
+    packageDir: 'query-core',
+    srcDir: 'src',
+    entries: ['main', 'module', 'types'],
+  },
   {
     name: '@tanstack/query-persist-client-core',
     packageDir: 'query-persist-client-core',
     srcDir: 'src',
+    entries: ['main', 'module', 'types'],
   },
   {
     name: '@tanstack/query-async-storage-persister',
     packageDir: 'query-async-storage-persister',
     srcDir: 'src',
+    entries: ['main', 'module', 'types'],
   },
   {
     name: '@tanstack/query-broadcast-client-experimental',
     packageDir: 'query-broadcast-client-experimental',
     srcDir: 'src',
+    entries: ['main', 'module', 'types'],
   },
   {
     name: '@tanstack/query-sync-storage-persister',
     packageDir: 'query-sync-storage-persister',
     srcDir: 'src',
+    entries: ['main', 'module', 'types'],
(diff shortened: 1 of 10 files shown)


lachlancollins:
@TkDodo any idea why these checks failed?

TkDodo:
Yeah alpha is broken currently since I merged main back to it, where we added distributed task execution in nx. @ZackDeRose is looking into it. Sorry for the inconvenience 🙈

lachlancollins:
@TkDodo no stress! The Nx distributed agents speed gains are 💯. Are checks on main also broken?

TkDodo:
@lachlancollins upgrade to node18 fixed the issue

TkDodo:
build validation fails now with:

> Validating packages...
Error: Missing entry for "types" in svelte-query/package.json!

lachlancollins:
I'm on it!

lachlancollins:
Hi @TkDodo , could you please review my changes to validate-packages.ts? I've added a check to make sure packages have `exports['.']['default']` field - eslint-plugin-query was failing this test so I've added exports to it. If this is too far outside the scope of this PR let me know and I'll split it. Or I could rename this PR more generally related to package.json :)

TkDodo:
@lachlancollins I don't think the eslint-plugin needs an `exports` entry. Maybe we could add `exports` to `entries` in `scripts/config.ts` so that it becomes:

```
entries: ['main', 'module', 'types', 'exports'],
```

for most packages, but leave it out for the eslint one?

scripts/config.ts
```diff
@@ -3,57 +3,77 @@ import type { BranchConfig, Package } from './types'
...
     name: '@tanstack/solid-query',
     packageDir: 'solid-query',
     srcDir: 'src',
+    entries: ['main', 'module', 'types'],
   },
   {
     name: '@tanstack/svelte-query',
     packageDir: 'svelte-query',
     srcDir: 'src',
+    entries: [],
```
TkDodo, line 64: are you sure that we don't need anything other than `exports` ? No `main`, no `types` ?

lachlancollins:
@TkDodo I'd have to do something like `exportsEntries: ['default', 'import', '...']`, since exports is a bit more complex to check. Would be easier to replace all this logic with publint!

Is there any harm in adding exports to the eslint package? It is the "replacement" to main/module/types and is used by all the other packages.

Also, I'm happy to add back module and types to svelte-query, but I don't think it's even possible to build svelte with a bundler that doesn't support exports :P

TkDodo:
> Is there any harm in adding exports to the eslint package?

I suppose not.

> Also, I'm happy to add back module and types to svelte-query, but I don't think it's even possible to build svelte with a bundler that doesn't support exports

Alright, just wanted to double check 👍

TkDodo:
not sure why codesandbox build is failing now and I can't re-trigger that. can you add an empty commit to re-run it?

lachlancollins:
@TkDodo what gets run on codesandbox? It seems like `nx run root:rollup` completes successfully, then whatever happens immediately after causes the error.

lachlancollins:
@TkDodo I saw webpack mentioned on the svelte subreddit so I panicked and added back the non-exports fields 😅

PR #5353, opened 2023-04-30
docs: deprecate callbacks on useQuery

TkDodo:


NMinhNguyen:
Are there plans to mark these APIs as `@deprecated` in TypeScript so that users of this library get IDE hints?

PR #5355, opened 2023-05-01
feat: queryOptions helper for solid-query

OrJDev:
Regarding this pr: https://github.com/TanStack/query/pull/5153

This adds `queryOptions` helper function for solid-query as-well, the use is about the same:

```ts
const queryOpts = queryOptions(() => ({
  queryKey: ['myQuery', 'key', 1],
}))

createQuery(queryOpts)
createQuery(() => queryOpts())
```


ardeora:
This is incredible! Thanks for adding the tests too! This looks great

PR #5347, opened 2023-04-29
feat(devtools): Add framework agnostic devtools draft

ardeora:
This is a initial draft of the framework agnostic Tanstack Query devtools. Will add description as it gets more fleshed out.

PR #5362, opened 2023-05-02
feat(svelte-query-devtools): Svelte Adapter for new Devtools

lachlancollins:
Implements #5347 for Svelte 🚀

### DIFF
--- a/packages/svelte-query-devtools/src/Devtools.svelte
+++ b/packages/svelte-query-devtools/src/Devtools.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
   import { onMount } from 'svelte'
+  import { DEV, BROWSER } from 'esm-env'
   import type { QueryClient } from '@tanstack/svelte-query'
   import { useQueryClient, onlineManager } from '@tanstack/svelte-query'
   import type {
@@ -7,7 +8,6 @@
     DevtoolsPosition,
     DevToolsErrorType,
   } from '@tanstack/query-devtools'
-  import { TanstackQueryDevtools } from '@tanstack/query-devtools'
 
   export let initialIsOpen = false
   export let buttonPosition: DevtoolsButtonPosition = 'bottom-left'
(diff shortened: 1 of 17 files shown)


TkDodo:
wohoo 🚀 you guys are on fire 🔥

ardeora:
This is amazing! Let me know if I can help in anyway possible 🙏

lachlancollins:
@ardeora everything works fine in dev mode in SvelteKit, then as soon as I build for prod I get `ReferenceError: window is not defined`. I've tried several things to only get it to run in dev on the browser with no luck yet. Would you have any suggestions?

ardeora:
> @ardeora everything works fine in dev mode in SvelteKit, then as soon as I build for prod I get `ReferenceError: window is not defined`. I've tried several things to only get it to run in dev on the browser with no luck yet. Would you have any suggestions?

I think this is expected right? We shouldn't be including the devtools in the prod build! If you see this part you'll see that we conditionally remove the devtools panel on build! https://github.com/TanStack/query/blob/alpha/packages/react-query-devtools/src/index.ts#L5-L10 

Is there a way to do this in Svelte where it treeshakes the devtools in prod environment?

lachlancollins:
> Is there a way to do this in Svelte where it treeshakes the devtools in prod environment?

That's what I'm trying to do, my first two methods didn't work - I'll search around for a diff solution!

ardeora:
> > Is there a way to do this in Svelte where it treeshakes the devtools in prod environment?
> 
> That's what I'm trying to do, my first two methods didn't work - I'll search around for a diff solution!

Oh yeah I'm not super familiar with Svelte but I'll try to find solutions that might make this work. I'm curious what are the two methods you already tried?

lachlancollins:
> Oh yeah I'm not super familiar with Svelte but I'll try to find solutions that might make this work. I'm curious what are the two methods you already tried?

On mobile so apologies for formatting:

1. Tried process.env.NODE_ENV like react devtools but didn't expect it to work
2. Tried esm-env (which is also used by SvelteKit) to recognise when running on browser/server/dev/prod
3. Tried using {#if dev && browser}<SvelteQueryDevtools />{/if} within the svelte-basic example

lachlancollins:
`esm-env` definitely seems like the best way to get it working (https://kit.svelte.dev/docs/packaging#best-practices). Maybe I need a dynamic import - even importing the component without rendering it was triggering the error.

lachlancollins:
@ardeora @TkDodo this is ready to go - is there anything I need to set for it to be published automatically?

packages/svelte-query-devtools/src/Devtools.svelte
```diff
@@ -0,0 +1,43 @@
...
+      const { TanstackQueryDevtools } = await import('@tanstack/query-devtools')
+      const devtools = new TanstackQueryDevtools({
+        client,
+        queryFlavor: 'Svelte Query',
+        version: '5',
+        onlineManager,
+        buttonPosition,
+        position,
+        initialIsOpen,
+        errorTypes,
```
ardeora, line 31: Super everything is looking great!! I was wondering if the props are reactive. Say if there is a button in the parent conatiner that changes buttonPosition from 'top-left' -> 'top-right'. Will the devtools button switch position?

lachlancollins: They are now!

ardeora (approved):

PR #5364, opened 2023-05-03
fix(devtools): Update ESM export extension

ardeora:
This updates the esm export to have an `mjs` extension to allow compatibility with Next.js

### DIFF
--- a/packages/query-devtools/package.json
+++ b/packages/query-devtools/package.json
@@ -12,11 +12,11 @@
   },
   "types": "build/types/index.d.ts",
   "main": "build/umd/index.js",
-  "module": "build/esm/index.js",
+  "module": "build/esm/index.mjs",
   "exports": {
     ".": {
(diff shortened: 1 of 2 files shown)


ardeora:
Hello! @lachlancollins  I'm still seeing this error

```
       /Users/aryandeora/Workspace/query/packages/svelte-query-devtools/src/Devtools.svelte:4:36
       Error: Cannot find module '@tanstack/svelte-query' or its corresponding type declarations. (ts)
         import { DEV, BROWSER } from 'esm-env'
         import type { QueryClient } from '@tanstack/svelte-query'
         import { useQueryClient, onlineManager } from '@tanstack/svelte-query'
```

I think this might be because svelte-query-devtools is being built before svelte-query. Any idea how this could be solved?

you'll have to run `pnpm run test:ci` to see this error. You can make a PR to my branch or create a new PR and I can rebase after

PR #5366, opened 2023-05-04
feat(devtools): Make queryclient reactive

ardeora:
This PR also adds reactivity to make the queryclient prop reactive in the devtools

### DIFF
--- a/packages/react-query-devtools/src/devtools.tsx
+++ b/packages/react-query-devtools/src/devtools.tsx
@@ -40,11 +40,12 @@ export function ReactQueryDevtools(
   props: DevtoolsOptions,
 ): React.ReactElement | null {
   const queryClient = useQueryClient()
+  const client = props.client || queryClient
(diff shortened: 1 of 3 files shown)


packages/react-query-devtools/src/devtools.tsx
```diff
@@ -40,11 +40,12 @@ export function ReactQueryDevtools(
   props: DevtoolsOptions,
 ): React.ReactElement | null {
   const queryClient = useQueryClient()
+  const client = props.client || queryClient
```
TkDodo, line 43: `useQueryClient` can handle this internally, so we can do:

```
const client = useQueryClient(props.client)
```

`useQueryClient` will read from context or return the passed prop if it isn't undefined:

https://github.com/TanStack/query/blob/e61c7849b1f4dd5bd7909dc2204e7c152c1b9479/packages/react-query/src/QueryClientProvider.tsx#L10-L22

PR #5375, opened 2023-05-06
Update query-functions.md

ChenKun1997:


TkDodo:
the types are correct as-is. `meta` is required, but can be undefined.

PR #5385, opened 2023-05-08
Batshit batching docs

yornaath:
As discussed with @TkDodo in the batching issue #365 . 
Here are the docs for how to set up batching of requests using `@yornaath/batshit`


docs/react/community/batching-requests-using-bathshit.md
```diff
@@ -0,0 +1,75 @@
...
+}
+```
+
+> Note: Developer ux remains the same as if the hook fetched the items individually as single requests and can be used arbitrarily across your codebase.
+
+Finally lets define our user details component that consumes the hook we made.
+
+```ts
+const UserDetails = (props: {userId: number}) => {
+  const {isFetching, data} = useUser(props.userId)
```
TkDodo, line 58:
```suggestion
  const {isLoading, data} = useUser(props.userId)
```

docs/react/community/batching-requests-using-bathshit.md
```diff
@@ -0,0 +1,75 @@
...
+  return <>
+    {
+      isFetching ?
```
TkDodo, line 61:
```suggestion
      isLoading ? 
```

TkDodo (commented):
thanks. FYI, we will soon re-structure the community resources. We generally don't want detailed information like QuickStarts in our docs because it will be cumbersome to keep them up-to-date as 3rd party libs progress. Instead, we'll keep a list of tools with pitches of when to use them and link to their pages only.

PR #5402, opened 2023-05-11
fix(perf): optimized perf of listeners removal

Andarist:
prompted by https://twitter.com/BenLesh/status/1656747203531079680

### DIFF
--- a/packages/query-core/src/focusManager.ts
+++ b/packages/query-core/src/focusManager.ts
@@ -66,7 +66,7 @@ export class FocusManager extends Subscribable {
   }
 
   onFocus(): void {
-    this.listeners.forEach((listener) => {
+    this.listeners.forEach(({ listener }) => {
       listener()
     })
   }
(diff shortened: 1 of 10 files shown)


benlesh (approved):
LGTM.

Although if you can weather the idempotent add, it would be more efficient to not have to wrap it in an object.

Andarist:
I assumed that it could be considered a breaking change for the current major so I didnt go for it here

JacobMGEvans:
Awesome Twitter to PR community involvement.

TkDodo:
> I assumed that it could be considered a breaking change for the current major so I didnt go for it here

I don't think it would break something, but if we want to be on the safe side, we could merge this to `main` and then remove the object wrapping on the `alpha` branch?

Andarist:
> I don't think it would break something, but if we want to be on the safe side, we could merge this to main and then remove the object wrapping on the alpha branch?

That sounds like a good incremental approach.

TkDodo:
> That sounds like a good incremental approach.

did that here: https://github.com/TanStack/query/commit/049e0c6e56c6314f4042e9c4849244a19e311161

PR #5407, opened 2023-05-13
docs: deprecate callbacks on `useQuery`

NMinhNguyen:
See https://github.com/TanStack/query/pull/5353#issuecomment-1545557244


NMinhNguyen:
Wasn't sure if the commit should be `feat` or something else

alicenstar:
@NMinhNguyen @TkDodo Can I ask what the reason for deprecating these callbacks is? And if they'll be replaced with something else/what you recommend using instead?

NMinhNguyen:
> @NMinhNguyen @TkDodo Can I ask what the reason for deprecating these callbacks is? And if they'll be replaced with something else/what you recommend using instead?

This should hopefully answer your question: https://tkdodo.eu/blog/breaking-react-querys-api-on-purpose

Issue #5416, opened 2023-05-16
eslint-plugin-query: false positive for out-of-scope constants with arrow functions


TkDodo:
interestingly, we have a test for this:

```
{
  name: 'should ignore constants defined out of scope (react hook)',
  code: `
    const CONST_VAL = 1
    function useHook() {
      useQuery({
        queryKey: ["foo"],
        queryFn: () => CONST_VAL
      });
    }
  `,
},
```
✅ 

which passes, but it starts to break as soon as we use an arrow function:

```
{
  name: 'should ignore constants defined out of scope (react hook arrow fn)',
  code: `
    const CONST_VAL = 1
    const useHook = () => {
      useQuery({
        queryKey: ["foo"],
        queryFn: () => CONST_VAL
      });
    }
  `,
},
```
❌ `message: 'The following dependencies are missing in your queryKey: CONST_VAL',`

@Newbie012 can you please take a look

am-maneaters:
another potential test to add:
```
{
  name: 'should ignore constants defined out of scope (react hook inline function)',
  code: `
    const CONST_VAL = 1
    const useHook = function () {
      useQuery({
        queryKey: ["foo"],
        queryFn: () => CONST_VAL
      });
    }
  `,
},
```

Newbie012:
oops

raised a [fix](https://github.com/TanStack/query/pull/5426)

PR #5426, opened 2023-05-17
fix(eslint-plugin): fix component and hook lookup

Newbie012:
fixes https://github.com/TanStack/query/issues/5416


Newbie012:
codesandbox action seems to be broken

PR #5439, opened 2023-05-20
chore: Bump eslint version to 8.34.0

lachlancollins:
I'm running into a weird eslint error when working on svelte-query which suggests it's using eslint@7.32.0. This is required by the react-native example, while the workspace root uses 8.34.0. I've bumped the dependency across all files that weren't using 8.34.0, which fixes the error.

### DIFF
--- a/examples/react/basic-typescript/package.json
+++ b/examples/react/basic-typescript/package.json
@@ -21,7 +21,7 @@
     "@types/react": "^17.0.3",
     "@types/react-dom": "^17.0.3",
     "@vitejs/plugin-react": "^2.0.0",
-    "eslint": "7.x",
+    "eslint": "^8.34.0",
     "eslint-config-prettier": "^8.3.0",
     "typescript": "4.7.4",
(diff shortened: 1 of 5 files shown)


lachlancollins:
@TkDodo I've noticed this duplicated dependency issue is also occurring for vite. See the below error when I open `vitest.config.ts` for solid-query, which is fixed by updating vite and vitest. Do you think now is a good time to bump some other dev dependencies?

`Type 'Plugin_2' is not assignable to type 'PluginOption'.
  Type 'import(".../tanstack-query/node_modules/.pnpm/vite@3.2.2/node_modules/vite/dist/node/index").Plugin' is not assignable to type 'import(".../tanstack-query/node_modules/.pnpm/vite@4.0.4_@types+node@18.13.0/node_modules/vite/dist/node/index").Plugin'.`

TkDodo:
> Do you think now is a good time to bump some other dev dependencies?

yeah we can bump everything on `alpha` that makes our live easier 👍

PR #5440, opened 2023-05-20
feat(infiniteQuery): allow prefetching arbitrary amounts of pages

TkDodo:


packages/query-core/src/types.ts
```diff
@@ -359,20 +359,28 @@ export interface FetchQueryOptions<
   staleTime?: number
 }
 
-export interface FetchInfiniteQueryOptions<
+type FetchInfiniteQueryPages<TQueryFnData = unknown, TPageParam = unknown> =
+  | { pages?: never; getNextPageParam?: never }
+  | {
+      pages: number
+      getNextPageParam: GetNextPageParamFunction<TPageParam, TQueryFnData>
+    }
```
TkDodo, line 367: if you pass `pages`, you also need to pass `getNextPageParam`, because it's needed to fetch more than one page.

One thing I'm unsure about is the generic name `pages`. Maybe it should be more specific, like `pagesAmount` or so ? Thoughs?

PR #5441, opened 2023-05-20
chore: Replace cp with cpy-cli

lachlancollins:
`cp` isn't supported on windows, so the `react-query` and `react-query-devtools` always fail to build. This adds a node script to copy files instead.


TkDodo:
would it make sense to use something like https://github.com/sindresorhus/cpy-cli ?

lachlancollins:
> would it make sense to use something like [sindresorhus/cpy-cli](https://github.com/sindresorhus/cpy-cli) ?

Definitely, I tried finding something like this but obviously didn't look hard enough!

PR #5453, opened 2023-05-23
fix(useIsMutating): fix vue warning

juvirez:
`onScopeDispose() is called when there is no active effect scope to be associated with`


packages/vue-query/src/useMutationState.ts
```diff
@@ -24,9 +24,8 @@ export function useIsMutating(
     status: 'pending' as const,
   }))
 
-  const length = computed(
-    () => useMutationState({ filters: unreffedFilters }, client).value.length,
-  )
+  const mutationState = useMutationState({ filters: unreffedFilters }, client)
```
DamianOsipiuk, line 27: Could you explain why this change is necessary?

juvirez: [Effective scope](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0041-reactivity-effect-scope.md) does not exist inside the `computed`. You cannot create reactive variables inside `computed`.
All reactive variables should be created synchronously from setup function, not from some callbacks.
If you try to create reactive variable without reactive context, you will get a Vue warning:
`onScopeDispose() is called when there is no active effect scope to be associated with`.

PR #5457, opened 2023-05-24
refactor(query-core): Make MutationMeta and QueryMeta generic

vojvodics:
This is a rework of #5412 for v5

The basic idea is that we can infer types from meta fields passed in as functions:
<figure>
Screenshot 2023-05-24 at 10 51 26
<figcaption>Without inference</figcaption>
</figure>

<figure>
Screenshot 2023-05-24 at 10 51 58
<figcaption>With inference</figcaption>
</figure>

This, however, requires some setup from the user:

```

interface MyMeta<
  TData = unknown,
  TError = unknown,
  TVariables = unknown,
  TContext = unknown
> {
  errorMessage: string | ((error: TError, variables: TVariables, context: TContext) => string);
  // I can also infer the type of data correctly
  successMessage: string | ((data: TData, variables: TVariables, context: TContext) => string);
}

  interface Register<
  TData = unknown,
  TError = unknown,
  TVariables = unknown,
  TContext = unknown
> {
    mutationMeta: MyMeta<TData, TError, TVariables, TContext>;
  }
}
```

Also, as mentioned in #5412 - the inference will only work if used in useMutation/useQuery.

Because all generic parameters are optional, this should work exactly the same as before and you don't need to use generics if you don't have to.


TkDodo:
what's problematic about using functions inside `meta` is that you really can't close over anything in your component, because there is only one `meta` object per query, and you'd likely read stale values:

```
function MyComponent(props) {

  useQuery({
    queryKey,
    queryFn,
    meta: () => props.foo
  })
}
```

so, I'm unsure if we should make it generic just to support that case 🤔

vojvodics:
yeah, I see the reason to discourage the usage of functions in meta. That's a problem even without generics.
I'd argue that the example you provided is more of a case in favor of generics - I would have done precisely that instead of reading the values from the query.

TkDodo:

PR #5459, opened 2023-05-25
Add refetchInterval to QueryOptions

rasmusrim:
This property is missing in the types, but is there when I do a console.log.

### DIFF
--- a/packages/query-core/src/types.ts
+++ b/packages/query-core/src/types.ts
@@ -68,6 +68,7 @@ export interface QueryOptions<
    */
   retry?: RetryValue<TError>
   retryDelay?: RetryDelayValue<TError>
+  refetchInterval?: number;
   networkMode?: NetworkMode
   cacheTime?: number
   isDataEqual?: (oldData: TData | undefined, newData: TData) => boolean


TkDodo:
it's part of QueryObserverOptions, and it can be different per observer, so it's technically not part of a Query. It might be there at runtime, but please don't rely on it.

busticated:
docs say:

> You can generally pass everything to queryOptions that you can also pass to [useQuery](https://tanstack.com/query/latest/docs/framework/react/reference/useQuery). Some options will have no effect when then forwarded to a function like queryClient.prefetchQuery, but TypeScript will still be fine with those excess properties.

https://tanstack.com/query/latest/docs/framework/react/reference/queryOptions

not sure how i'm supposed to use `refetchInterval` in `useQueries()` given that i can't set it on my query 🤔

busticated:
```ts
export function myQuery(
  id?: string,
  token?: string,
) {
  const api = new APIClient({ token });

  return {
    enabled: !!(token && id),
    queryKey: ['my', id],
    queryFn: async () => {
      return api.detail(id);
    },
  };
}

export function myQueryWithRefetch(
  id?: string,
  token?: string,
) {
  const query = myQuery(id, token);
  // @ts-expect-error - see: https://github.com/TanStack/query/pull/5459
  query.refetchInterval = 1000 * 10;
  return query;
}

export function useMyQueriesData(id: string) {
  const { token } = useAPIToken();

  return useQueries({
    queries: [
      myQueryWithRefetch(id, token),
      myOtherQuery(id, token),
    ],
    combine: (results) => {
      const [one, two] = results;
      const combined = combineQueryStates(results);
      combined.data.one = one?.data || null;
      combined.data.two = two?.data || null;
      return combined;
    },
  });
}
```

Issue #5443, opened 2023-05-21
Potential race condition exists between an observer subscribing to a query and when that query's response comes back


ericbiewener:
### Describe the bug

I am seeing a bug in my application when using mocked network requests (i.e. query fetch requests resolve immediately), resulting from a race condition between these network requests resolving and a component's observer subscribing to the query's changes. Unfortunately, I can not reproduce it outside of our proprietary codebase, so simply demonstrating the theoretical possibility will need to suffice.

Consider two components, `A` and `B`, both of which call `useQuery` with the same `queryKey` (thus sharing the same underlying `Query` instance).

1. `A` calls `useQuery`. This creates a new `Query` instance and the query begins fetching. A receives `isLoading: true` from the hook.
2. `B` calls `useQuery`, sharing the same `Query` instance as `A`, and therefore not executing a new fetch. `B` receives `isLoading: true` from the hook.
3. The `useEffect` hooks for `A` execute, resulting in the query observer for `A` calling `addObserver` on its query (via `useSyncExternalStore`), thus subscribing to query state notifications.
4. The network request resolves, updating the `Query` state and notifying `A` of changes.
5. The `useEffect` hooks for `B` execute, so now `B` is subscribed to query state notifications, _but it does not get notified of the query state updates from the previous step_. **This is the race condition.**
6. `A` re-renders with the updated query state, `B` does not.

This race condition exists because React Query does not control how the JS event loop will schedule the `useEffect` calls vs. the resolution of the network request, since both are happening asynchronously. Of course, the moment any real network latency is added to the network call, the execution of the `useEffect` hooks first is effectively guaranteed. Furthermore, while I can reproduce this dependably in Chrome, I am not able to do so in Firefox, which hints at the role that the JS engine is playing in regards to how the work is being scheduled on the event loop.

Our codebase currently has nearly 1700 testcafe tests, and this edge case bug appeared on only a few, all of which were in the same test suite, thus indicating just how much of an edge case it is.

FWIW, component `B` in our case is a child of component `A`, and is rendered as a Next.js `dynamic` component. I don't see any reason to believe that is relevant beyond it having some random effect on the scheduling of work on the event loop, but in theory so could anything else.

**Solution**
[...]

PR #5464, opened 2023-05-25
fix: add subscription right after observer creation in react-query

artem-malko:
This PR attempts to close https://github.com/TanStack/query/issues/5443

### Opened questions

* It is really hard to write any tests for such problem(
* Is it clear about useLazyRef? Looks like a hack

Please, read the [issue](https://github.com/TanStack/query/issues/5443) before code review.

@TkDodo and @ericbiewener you have the full context)

### DIFF
--- a/packages/react-query/src/useBaseQuery.ts
+++ b/packages/react-query/src/useBaseQuery.ts
@@ -14,6 +14,7 @@ import {
   useClearResetErrorBoundary,
 } from './errorBoundaryUtils'
 import { ensureStaleTime, shouldSuspend, fetchOptimistic } from './suspense'
+import { useLazyRef } from './useLazyRef'
 
 export function useBaseQuery<
   TQueryFnData,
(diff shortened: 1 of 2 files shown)


artem-malko:
@TkDodo if there is a way to check react-query in my private project from that pull request, I'd like to use it)

packages/react-query/src/useLazyRef.ts
```diff
@@ -0,0 +1,26 @@
...
+  const ref = React.useRef<Value | typeof none>(none)
+
+  // if it's not initialized (1st render)
+  if (ref.current === none) {
+    // we initialize it
+    ref.current = initializer()
+  }
+  // now we return the initialized ref
+  // "as" is used to not leak any info about type of "none" outside of useLazyRef
+  return ref as React.MutableRefObject<Value>
```
TkDodo, line 25: I don't understand why this is preferred to `useState` ? It just seems like a more complicated, harder-to-type way of achieving the exact same thing?

packages/react-query/src/useBaseQuery.ts
```diff
@@ -65,37 +66,75 @@ export function useBaseQuery<
...
+      TQueryFnData,
+      TError,
+      TData,
+      TQueryData,
+      TQueryKey
+    >(queryClient, defaultedOptions)
+
+    initiallyUnsubscribe.current = observer.subscribe(() => {
+      needsToFlushStoreChange.current = true
+    })
```
TkDodo, line 90: I thought the agreement was to use layout effect instead ? I'm pretty sure it's against the rules of react to create a subscription inside an initializer, which should be pure.

TkDodo (changes requested):

TkDodo:
superseded by:
- https://github.com/TanStack/query/pull/5474

packages/react-query/src/useBaseQuery.ts
artem-malko: It was an attempt to create a subscription on the first render) But yeah, you're right, it's quite tricky

PR #5472, opened 2023-05-26
fix: Downgrade chalk to v4

lachlancollins:
v5 is ESM only, which would require additional tweaks to support. Might consider later on!


TkDodo:
yeah I've been there before with an accidental upgrade that didn't work 😅

PR #5474, opened 2023-05-26
fix(react-query): fix missed updates between creation and subscription

TkDodo:


packages/react-query/src/useBaseQuery.ts
```diff
@@ -77,10 +77,17 @@ export function useBaseQuery<
...
-        isRestoring
+      (onStoreChange) => {
+        const unsubscribe = isRestoring
           ? () => undefined
-          : observer.subscribe(notifyManager.batchCalls(onStoreChange)),
+          : observer.subscribe(notifyManager.batchCalls(onStoreChange))
+
+        // Update result to make sure we did not miss any query updates
+        // between creating the observer and subscribing to it.
+        observer.updateResult()
```
artem-malko, line 87: Looks interesting, I'll try to test it in my project

TkDodo: codesandbox preview seems to be having issues right now. You could patch-package the fix in your project, or we just ship it and trust the test coverage.

artem-malko: The main problem, that I cannot patch) It's to long story to describe)

Let's ship it, cause looks like this fix might help.

packages/react-query/src/__tests__/useQuery.test.tsx
```diff
@@ -5887,4 +5887,37 @@ describe('useQuery', () => {
...
+      const { data, status } = useQuery({
+        enabled: false,
+        queryKey: key,
+        queryFn: async () => {
+          await sleep(10)
+          return 5
+        },
+      })
+
+      // this simulates a synchronous update between the time the query is created
```
artem-malko, line 5905: Please, correct me if I'm wrong, It's not important that the update was synchronous or asynchronous? You've used the synchronous update just because it's easer to test?

TkDodo: Yes, exactly

artem-malko (approved):

artem-malko:
@TkDodo good news, we've tested the latest version of react-query and everything is ok right now! THX a lot)

@ericbiewener I can recommend you to update react-query to [4.29.12](https://github.com/TanStack/query/releases/tag/v4.29.12)

PR #5475, opened 2023-05-26
refactor(query-core): optimization addEventListener events with forEach

li-jia-nan:
optimization addEventListener events with forEach


li-jia-nan:
@TkDodo hello, can you review this? thx

KODerFunk:
I agree with @QzCurious that this is not an optimization. 
1. Runtime performance will be worse due to loops. 
2. The array pre-allocates memory. 
3. More lines of code. 
4. To understand the events exposed, you need to raise your eyes higher.

TkDodo:
open a revert PR if you want. I don't really care that much, I find both ways equally readable.

PR #5476, opened 2023-05-26
chore: Update to pnpm v8

lachlancollins:
The main branch is already using lockfile version 6.0 (pnpm v8), while the alpha branch has lockfile version 5.4 (pnpm v7). Can't think of any reason against updating (see changes [here](https://github.com/orgs/pnpm/discussions/4967)), and it should mean github actions run faster since it doesn't need to regenerate the lockfile back to 5.4.

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,5 +1,5 @@
 {
-  "installCommand": "install:csb",
+  "installCommand": "corepack enable && install:csb",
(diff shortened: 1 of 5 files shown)


lachlancollins:
@TkDodo I spoke too soon - codesandbox doesn't use pnpm 8 yet. If I allow it to regen the pnpm lockfile it works, which isn't necessary for github actions. I don't quite understand why codesandbox hasn't been throwing errors since it's been using lockfile v6.0 for a while!

PR #5480, opened 2023-05-27
refactor: move queryOptions to its own file

TkDodo:


TkDodo:
I tried this out and the errors are way better if you have `'use client'` in hooks and you try to use them in a server component. I'll just refactor `queryOptions` to a separate file instead

Ephem (approved):
> I tried this out and the errors are way better if you have `'use client'` in hooks and you try to use them in a server component. I'll just refactor `queryOptions` to a separate file instead

Sad to hear that, it's annoying, but I'm also happy I wasn't hallucinating about that part. 😅

So I guess the title of this PR is no longer up to date, it's now extracting `queryOptions` so it's usable on the server? I think it's looking good!

TkDodo:
so with what we've learned at react summit:

- keep `'use client'` for hooks and for the `HydrationBoundary`
- add `import client-only` to the `QueryClientProvider`

is that about right?

Ephem:
Yeah I think that's a nice approach. The reasoning for `QueryClientProvider` having `import 'client-only'` instead of `'use client'` is that because it needs a `queryClient` as props and that can only be provided to it from another client component?

Ephem:
Yep, just checking that we're on the same page. It has to be a _Client Component_, but it can never be a _Client Boundary_ so to speak.

Hmm, the options to `HydrationBoundary` are indeed a problem.. I don't think I noticed when they were introduced so I don't know what use cases they support (why would you need to set specific `queryOptions` for these queries for example other than those that are automatically included in the dehydration? why the custom context support?), but if we need to keep these around, maybe we should bring back `<Hydrate>` as a `import "client-only"` version that supports options, and make `<HydrationBoundary>` not accept options and be the `'use client'` version?

Just thinking out loud.

TkDodo:
> why would you need to set specific queryOptions for these queries

you can apparently set `defaultOptions` with which those queries or mutations should be hydrated:

https://github.com/TanStack/query/blob/4fa1cdbea369a713a50e0fece3f15fb23b6c6513/packages/query-core/src/hydration.ts#L18-L23

> why the custom context support?

in v5, it's a custom queryClient (no more context). We added support to pass `queryClient` everywhere (even `useQuery`) so that you can use it in a situation where you don't want to rely on the queryClient from react context, e.g. in microfrontends.

> maybe we should bring back <Hydrate> as a import "client-only" version that supports options, and make <HydrationBoundary> not accept options and be the 'use client' version?

if you currently try to pass `options` to `<HydrationBoundary>`, it will fail if its in a server component (unless your options are in fact serializable, which they could very well be). In that case, you would need to move it to a client component. The error is not that nice, but I think most users won't use that, so I don't really see the problem tbh 😅

Ephem:
Oh, I just noticed I hadn't answered this, I thought I did. 😅

> so I don't really see the problem tbh 

Yeah, I think you're right, I'm overthinking it. 😄

TkDodo:
I'll merge this (moving queryOptions to its own file) and I'll do the actual `use client` / `import client-only` changes in a different PR that goes to `main`.

PR #5487, opened 2023-05-29
feat: Split rollup config for Nx caching

lachlancollins:
Adds a `build:rollup` script to each package that previously used rollup, and imports the config from `scripts/getRollupConfig.mjs`.

This separation allows Nx to cache changes to each build separately, which speeds up development as Nx knows which packages depend on which builds 🚀 

Without cache (noting it will obviously be faster with cache):
- Before: 28s
- After: 25s


lachlancollins:
Hi @ZackDeRose, I'm a bit stumped here, wondering if you could help out 😕 

When I run `pnpm build --skip-nx-cache` locally, it runs `nx run @tanstack/query-devtools:build` and `nx run @tanstack/query-core:build` first, since other packages depend on these. When it runs with Nx Cloud on GitHub, it seems like the ordering isn't preserved, and there are build errors because `query-core` and `query-devtools` are missing.

Do you have any suggestions for enforcing build order in the Nx config?

PR #5492, opened 2023-05-30
fix(svelte): correct workspace version for svelte-query-devtools

RomainLanz:
Hey there! 👋🏻 

When cloning the `alpha` branch, we cannot install dependencies because all Svelte packages use a no-longer-available version of `@tanstack/svelte-query-devtools`.

```
 ERR_PNPM_NO_MATCHING_VERSION_INSIDE_WORKSPACE  In examples/svelte/auto-refetching: No matching version found for @tanstack/svelte-query-devtools@5.0.0-alpha.39 inside the workspace
```

This PR corrects the version required by those packages.

### DIFF
--- a/examples/svelte/auto-refetching/package.json
+++ b/examples/svelte/auto-refetching/package.json
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@tanstack/svelte-query": "^5.0.0-alpha.38",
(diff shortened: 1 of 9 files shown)


lachlancollins:
Hi @RomainLanz , I actually fixed this in #5495, but thank you for pointing it out!

PR #5493, opened 2023-05-30
refactor(vue-query): use shallowRef to improve performance

RomainLanz:
Hey there! 👋🏻 

This PR refactors the `useQueries` composable to use a `shallowRef` instead of a `ref` to define the state.

If an object is assigned as a ref's value, the object is made deeply reactive. That is the case in the `useQueries` composable. Meaning if we had a huge `data` object, all its properties would be reactive, causing performance issues.

A [`shallowRef`](https://vuejs.org/api/reactivity-advanced.html#shallowref), in comparison, does not convert the object. The inner value is stored and exposed as-is.

This PR is in draft because I believe we can do the same optimization in the `useQuery` composable. I need to play with it and ensure nothing is breaking before making changes.

[Related PR on VueUse repository to add a `toShallowRef` helper](https://github.com/vueuse/vueuse/pull/3058) that could be used in `useQuery`.

Linked Discussion: https://github.com/TanStack/query/discussions/5400


TkDodo:

GreatAuk:
`useQuery` return a readonly data, use `ref` seems is meaningless.

RomainLanz:
> `useQuery` return a readonly data, use `ref` seems is meaningless.

It has to be a `ref` (or `shallowRef`) to ensure a rerender when data changes.

Alexbrazdasilva:
@RomainLanz 

I arrived long after the party ended, but...

It's possible to define the elements of long arrays as skippable using `markRaw`. This way, the data remains reactive, but the elements of the long array do not. This mimics the behavior of `shallowRef` without using `shallowRef`.

Does this solve the performance problem? In the `onMount` event, it might not, but in subsequent renders, it might.

I did it like this:

```ts
const { data, isLoading } = useQuery({
  queryKey: ["city", "select-city"],
  queryFn: () => $api.get("city"),
  staleTime: 10 * 60 * 1000, // 10 minutes
  select(data) {
    return safeArray(data.data).map(markRaw);
  },
});
```

PR #5511, opened 2023-06-02
chore: Fix GitHub Workflows

lachlancollins:
`nx affected` continues to break workflows for no obvious reason with `vue-query:build`, while `nx run-many` does not have this issue. As a temporary fix, use `run-many`, and can add back `affected` in a PR that proves it works.


lachlancollins:
Nevermind, I surrender, vue-query has defeated me

PR #5514, opened 2023-06-02
chore: Update to Nx 16.3.1

lachlancollins:


nx.json
```diff
@@ -1,5 +1,8 @@
 {
   "$schema": "./node_modules/nx/schemas/nx-schema.json",
+  "affected": {
+    "defaultBase": "alpha"
```
TkDodo, line 4: we have to keep this in mind when we go to `beta` and onwards :)

PR #5517, opened 2023-06-02
feat: Remove UMD build output

lachlancollins:
If you're not using a module bundler or package manager, you can also use this library via an ESM-compatible CDN such as [ESM.sh](https://esm.sh/). Simply add a `<script type="module">` tag to the bottom of your HTML file:

```html
<script type="module">
  import React from 'https://esm.sh/react@18.2.0'
  import ReactDOM from 'https://esm.sh/react-dom@18.2.0'
  import { QueryClient } from 'https://esm.sh/@tanstack/react-query@alpha'
</script>
```


package.json
```diff
@@ -97,7 +95,7 @@
   "bundlewatch": {
     "files": [
       {
-        "path": "packages/*/build/umd/*.production.js"
+        "path": "packages/*/build/lib/*.js"
```
TkDodo, line 98: I'm not sure if the report we're getting is that helpful now:

Screenshot 2023-06-05 at 09 22 19

PR #5518, opened 2023-06-02
chore: Re-enable Nx Distributed Task Execution

lachlancollins:


lachlancollins:
Hi @ZackDeRose, I was wondering if you could help with this issue - I've run out of things to try. For some reason, `vue-query:build` runs fine without distributed task execution, but fails on the distributed agents, even if there's only a single agent. Strangely, it fails because of a type error, when `vue-query:test:types` passes without issue. Both depend on `query-core:build`, and both have the same inputs, so I really don't know what else I can do to fix this.

ZackDeRose:
(also fwiw, I think the solution in #5635 is a good work around in the meantime - where you mark one of the tasks as dependent on the other)

PR #5557, opened 2023-06-11
Fix: Update useAppState.ts

ursrudra:
Updating the cleanup function as the removeEventLister method is not available on AppState


TkDodo (commented):
can you please remove the added semicolons? In v5 (alpha), we're running the examples against prettier, too, and this will just cause failures there. Thanks.

ursrudra:
> can you please remove the added semicolons? In v5 (alpha), we're running the examples against prettier, too, and this will just cause failures there. Thanks.

Removed now, Thank you

PR #5576, opened 2023-06-13
fix(vue-query): do not subscribe on server - nuxt2 memory leak fix

DamianOsipiuk:
This should fix the memory leak on Nuxt2.
Nux2 has a different behavior than Nux3 which properly executes `onScopeDispose`.
And since we need to `await query.suspense()` to get data on server side, subscriptions should not be necessary on server environments.


zhuscat:
@DamianOsipiuk I found https://github.com/TanStack/query/commit/3e82c581da7b917e2be959feb6c8c7a726f6b14e reverts this commit. Is this intentional? And I found that `onScopeDispose` is not called on Nuxt 3. Will this cause memory leak?

PR #5580, opened 2023-06-14
Pass the last combined result in combine method

RomainLanz:
Hey there! 👋🏻 

This PR that was initiated based on a productive discussion with @TkDodo on the Discord server.

The objective of this PR is to enhance the functionality, specifically by enabling them to utilize the most recent combined result in their combination logic. This enhancement aims to provide more flexibility and control in handling scenarios where certain conditions are not met, allowing developers to conveniently revert to the previously computed data.

```ts
export function useAppData() {
  return useQueries({
    queries: [
      fooListQueryDefinition(),
      barListQueryDefinition({ sort: "asc" }),
    ],
    combine(results, combinedResult = []) {
      //                    👆
      const isLoading = results.some((result) => result.isLoading);
      const isError = results.some((result) => result.isError);
      const isSuccess = results.every((result) => result.isSuccess);
      const isRefetching = results.some((result) => result.isRefetching);

      if (!isSuccess || isRefetching) {
        return { isLoading, isError, isSuccess, isRefetching, data: combinedResult };
        //                                                                👆
      }

      const fooList = Object.values(results[0].data);
      const barList = Object.values(results[1].data);

      const data = fooList.map((f) => ({
        ...f,
        bar: barList.find((b) => b.id === f.bar_id) || null,
      }));

      return {
        isLoading,
        isError,
        isSuccess,
        isRefetching,
        data,
      };
    },
  });
}
```


TkDodo (changes requested):
Thanks. Can you add a test and update the documentation please?

RomainLanz:
Hey @TkDodo! 👋🏻 

I have updated the documentation.
I am not sure how to write tests inside the `queriesObserver.test`. In fact, there is no test to ensure `combine` works from here.

Those tests are directly written in framework packages like `react-query` and `vue-query` using the `useQueries` hook.

TkDodo:
yeah, you can add one to the react adapter for `useQueries` ?

RomainLanz:
Sure.

I have added the test and modified a bit the `queriesObserver`.
At line `55-61` the default value is set to `[]`. Meaning the first run of `combine` will have a `#combinedResult` set to `[]`.

I have checked that if the value of `#combinedResult` is exactly equal to `[]` it should use `undefined`.
I am not sure if that's the best way to do it, maybe we can leave the default behavior and ask the user to properly handle that.

TkDodo:
> At line 55-61 the default value is set to []. Meaning the first run of combine will have a #combinedResult set to [].

not sure I completely follow. when `this.#combineResult` is called the first time, `this.#combinedResult` is still undefined.

```
  #setResult(value: QueryObserverResult[]) {
    this.#result = value
    this.#combinedResult = this.#combineResult(value)
  }
```

why would it be empty array ?

RomainLanz:
Because you are calling `this.#setResult([])` in the constructor.

TkDodo:
yes, I set `this.#result`, but you're checking for `Array.isArray(this.#combinedResult)`

As far as I can see, at the point of the check, `this.#combinedResult` is still `undefined` ...

packages/vue-query/src/useQueries.ts
```diff
@@ -155,7 +155,7 @@ export function useQueries<
     ...options
   }: {
     queries: MaybeRefDeep<UseQueriesOptionsArg<T>>
-    combine?: (result: UseQueriesResults<T>) => TCombinedResult
+    combine?: (result: UseQueriesResults<T>, combinedResult: TCombinedResult | undefined) => TCombinedResult
```
TkDodo, line 158: you'd also need to change the interface of `useQueries` in the react adapter please

RomainLanz:
> As far as I can see, at the point of the check, `this.#combinedResult` is still `undefined` ...

I know, and it does not make sense to me either. I was surprised when I ran the test and saw that I was receiving `[]` in the first pass.

TkDodo:
there are build and type errors in the tests, which I can also see locally. Something doesn't really work here with the types...

TkDodo:
as discussed, the circular type issues are not really resolvable so we'll not add this feature.

PR #5598, opened 2023-06-17
feat: react-query-next-experimental package

TkDodo:


TkDodo:
@juliusmarminge @KATT @Ephem I put up a draft PR with a new experimental package for v5 that contains the `lib` code from the poc: https://github.com/trpc/use-client-stream-hydration/tree/main/src/app/lib

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
```diff
@@ -0,0 +1,192 @@
...
+    onEntries: (entries: TShape[]) => void
+    /**
+     * **Server method**
+     * onFlush is called on the server when the cache is flushed
+     */
+    onFlush?: () => TShape[]
+  }) {
+    // unique id for the cache provider
+    const id = `_${React.useId()}`
+    const idJSON = JSON.stringify(id)
```
TkDodo, line 84: since we put `idJSON` on the `window`, should we prefix it with something static like `RQ_` ?

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
```diff
@@ -0,0 +1,192 @@
...
+        } as unknown as TShape[]
+      }
+      return []
+    })
+    const count = React.useRef(0)
+    const onDehydrateRef = React.useRef(props.onFlush)
+    onDehydrateRef.current = props.onFlush
+    useServerInsertedHTML(() => {
+      // This only happens on the server
+      stream.push(...(onDehydrateRef.current?.() ?? []))
```
TkDodo, line 111: why do we need to put `props.onFlush` into a `ref` ? Can't we just access props inside `useServerInsertedHTML` ?

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
```diff
@@ -0,0 +1,192 @@
...
+            __html: html.join(''),
+          }}
+        />
+      )
+    })
+    // </server stuff>
+
+    // <client stuff>
+    const onEntriesRef = React.useRef(props.onEntries)
+    onEntriesRef.current = props.onEntries
```
TkDodo, line 142: I think we should writing to this `ref` into a `useEffect`. We do the same thing e.g. here:

https://github.com/TanStack/query/blob/a6186e50488bd3e14b676bf4aedd0953be3f0574/packages/react-query-persist-client/src/PersistQueryClientProvider.tsx#L22-L26

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+    // Client: consume cache:
+    const onEntries = React.useCallback(
```
TkDodo, line 145: since `onEntries` is only used once (in the effect), we could move it inside the effect to avoid `useCallback`, which means the dependencies of the effect would become `[id, transformer]`

TkDodo:
@Ephem there seems to be a build error related to the rollup plugin `preserve-directives`:

```
> rollup --config rollup.config.js


./src/index.ts → ./build/lib, ./build/lib...
[!] (plugin preserve-directives) SyntaxError: Unexpected token (43:9)
src/ReactQueryStreamedHydration.tsx (43:9)
```

Tbh, I just copied the rollup config from other packages, where it works fine ... 🤔

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
KATT: Just habit from using stuff in hooks, it can prob be removed

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
KATT: Yeah, I think react/next uses to underscores and a namespace

Originally we had a fixed namespace like `__cache` where I added stuff, like the structure was `window.__cache[useId()] = []`

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
KATT: I've never understood the benefit of doing a useEffect thing without dep array over just doing it inline

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
```diff
@@ -0,0 +1,89 @@
...
+  /**
+   * Track which queries were already passed to the client so we don't pass them again
+   */
+  // const [passedKeys] = useState(() => new Set<string>());
+
+  const cache = queryClient.getQueryCache()
+
+  if (typeof window === 'undefined' && !isSubscribed.current) {
+    // Do we need to care about unsubscribing? I don't think so to be honest
+    cache.subscribe((event) => {
```
KATT, line 40: I thiiink this is fine as its only mounted once on the server and the GC should clean it up once it is rendered, but might be worth looking into? There is no cleanup callback thingie on the server afaik

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
KATT: Sounds good to me!

KATT: I prefer inlining as much as possible, I've had that debate countless times lol

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
```diff
@@ -0,0 +1,194 @@
...
+      onEntries(...stream)
+
+      // Register our own consumer
+      win[id] = {
+        push: onEntries,
+      }
+
+      return () => {
+        // Cleanup after unmount
+        win[id] = {
```
KATT, line 174: This should probably be changed to an empty array instead

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
TkDodo: namespaced here: fede312c

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
TkDodo: refactored here: c428e49e

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
TkDodo: I think it's about the rules of react that you shouldn't write to refs because of concurrent features. If it's all the same to you: 40efb37b

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
TkDodo: inlined here: 73367547

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
```diff
@@ -0,0 +1,194 @@
...
+      [transformer],
+    )
+
+    React.useEffect(() => {
+      const win = window as any
+      // Register cache consumer
+      const stream: Array<Serialized<TShape>> = win[id] ?? []
+
+      if (!Array.isArray(stream)) {
+        throw new Error(`${id} seem to have been registered twice`)
```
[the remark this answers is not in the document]
TkDodo: removed: 9ce95136

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
TkDodo: okay: a83ca67b

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
TkDodo: I think it's only troublesome if it cannot be GC'd for some reason, e.g. a running timer. `cache.subscribe` per se doesn't create those, so I also think it's fine.

TkDodo:
@KATT I think this would be ready to ship. The example is not yet working because the package it depends on (the new one) doesn't exist yet on npm. The plan is to release this, then make a follow-up PR that adds the dependency.

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
```diff
@@ -0,0 +1,184 @@
...
+          transformer.deserialize(serialized),
+        )
+        onEntriesRef.current(entries)
+      }
+
+      const win = window as any
+      // Register cache consumer
+      const winStream: Array<Serialized<TShape>> = win[id] ?? []
+
+      onEntries(...winStream)
```
Ephem, line 159: This is the only time anything hydrates, right? Can we always do that in an effect? Is the effect for an entry always guaranteed to run before React hydrates the corresponding DOM? Imagine for example if this provider and a section that has a query flushes to the stream at the same time, will the effect have a chance to run before that section is hydrated by React then? If not, there will be hydration mismatches.

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
```diff
@@ -0,0 +1,87 @@
...
+ * - dehydrating the query client on the server
+ */
+export function ReactQueryStreamedHydration(props: {
+  children: React.ReactNode
+  queryClient?: QueryClient
+  transformer?: HydrationStreamProviderProps<DehydratedState>['transformer']
+}) {
+  const queryClient = useQueryClient(props.queryClient)
+
+  // <server only>
```
Ephem, line 23: What's this? This code doesn't really only run on the server pass right, it just happens to not do anything in the end when it runs on the client? If the code itself still runs I find this comment a bit confusing.

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
Ephem: Side effect in render which directly mutates state and also doesn't unsubscribe? Hat-trick, I like it. 😉 

I don't see `isSubscribed.current` ever being set to `true` though, forgot?

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const isSubscribed = React.useRef(false)
+  /**
+   * We need to track which queries were added/updated during the render
+   */
+  const [trackedKeys] = React.useState(() => new Set<string>())
```
Ephem, line 28: Could this be a ref instead? Just thinking it would make it clearer that it's not normal state but being mutated directly.

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
```diff
@@ -0,0 +1,87 @@
...
+            const shouldDehydrate =
+              trackedKeys.has(query.queryHash) &&
+              // !passedKeys.has(query.queryHash) &&
+              query.state.status !== 'pending'
+
+            // passedKeys.add(query.queryHash);
+            return shouldDehydrate
+          },
+        })
+        trackedKeys.clear()
```
Ephem, line 67: Not a 100% on how the internals work here, so I'll ask. Is the `added` event triggered when the query is added to the cache, even if it doesn't have data? That's why we check if it's pending here, so we don't dehydrate pending queries?

IF there are pending queries, they will still get cleared by `trackedKeys.clear()` even though they haven't dehydrated, but, when data comes in a new `update` event will be triggered and they will get re-added as a `trackedKey` and flushed next time?

I guess the case where this could happen is if you prefetch a query inside of render but don't await it in the same Suspense boundary or something like that?

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
Ephem, line 28: What happens if there are already queries in the cache, for example is some queries have been seeded/prefetched/are currently being prefetched (pending)? Are you then supposed to seed the browser RQ cache yourself too by having a `<Hydrate>` above these new providers in the tree (not possible for pending queries), or should this Set be initialised with any existing queries?

Ephem (commented):
Nice work on this! I left a few comments, but I think the big one is probably the hydrating in `useEffect` one, I have a feeling there are timing based issues there that can lead to hydration mismatches. Just doing it in render like we currently do elsewhere isn't great either of course so 🤷.

And of course the whole how this encourages fetch-in-render and client waterfalls over fetch-and-render, but I've talked that one to death already. 😄

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
TkDodo: I think using state for this is actually the more ergonomic approach, we're also doing this in e.g. `useQuery`:

https://github.com/TanStack/query/blob/3643b87c34d86117b150ca845a058991ff132ab0/packages/react-query/src/useBaseQuery.ts#L68-L74

I also have a blogpost on that topic: https://tkdodo.eu/blog/use-state-for-one-time-initializations

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
TkDodo: you're right - we're never setting the `isSubscribed` ref. This bears the question: do we even need it? Why would that render function run more than once on the server?

Ephem: If it renders inside a Suspense boundary that suspends I'm not sure what happens or if behaviour is guaranteed, so it does feel safer and with no downsides to use it I think? 😄

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
Ephem: Oh yeah, I totally spaced actually and forgot refs don't support initializer functions.. 😬 Should also have said this was very much a NIT to start with.

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
TkDodo: hm, here's my thinking:

1) the way I understand suspense is that it unmounts the component, then re-mounts it. At that point, the state of the ref would be lost anyways.
2) The `ReactQueryStreamedHydration` likely goes at the very top of your component tree, like in the example:
https://github.com/TanStack/query/blob/6ebe4f347a05462ef8c32d360fa183a4079b7515/examples/react/nextjs-suspense-streaming/src/app/providers.tsx#L22-L27
3) Even if we were to double subscribe - it would just add the same hash to the `trackedKeys` Set, which doesn't do anything.

> with no downsides to use it I think?

The downside is that we have some code that we have to maintain and ship to the browser, even though we're not really sure if we need it 😅 .

I've removed it here: 4ef87c15

But if you think that we should re-add it, we can. But yes, in that case, we would need to set the value to `true` in that if branch.

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
TkDodo: I've moved the comment further down: ba57ad84, but I think the intent was that `trackedKeys`, while it does run on the client as well, is only used on the server:
- the write is in the `typeof window === 'undefined'` check
- the read is in the `onFlush` callback, which also happens on the server only.

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
```diff
@@ -0,0 +1,77 @@
...
+      onFlush={() => {
+        /**
+         * Dehydrated state of the client where we only include the queries that were added/updated since the last flush
+         */
+        const dehydratedState = dehydrate(queryClient, {
+          shouldDehydrateQuery(query) {
+            return (
+              trackedKeys.has(query.queryHash) &&
+              query.state.status !== 'pending'
+            )
```
TkDodo, line 54: instead of checking for `pending` here, I think we should merge with the `defaultShouldDehydrateQuery` function, which only dehydrates successful queries.

without extra transformers like superjson, errors couldn't be serialized otherwise.

if necessary, we need to expose `dehydrateOptions` (as well as hydrate options maybe?

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
TkDodo: I think they would be hydrated automatically, too. Given that the `ReactQueryStreamedHydration` is very high up your component tree, it would subscribe very early to the QueryCache, so it would also track those queries.

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
TkDodo: I'm thinking maybe like this?
9d67d8bf

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
```diff
@@ -0,0 +1,184 @@
...
+   */
+  onEntries: (entries: TShape[]) => void
+  /**
+   * **Server method**
+   * onFlush is called on the server when the cache is flushed
+   */
+  onFlush?: () => TShape[]
+}
+
+export function createHydrationStreamProvider<TShape>() {
```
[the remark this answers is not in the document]
KATT: It doesn't have to be, it is just guaranteed to get its own context that is bound to the type

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
KATT: > This is the only time anything hydrates, right? 

Yes

> Can we always do that in an effect? Is the effect for an entry always guaranteed to run before React hydrates the corresponding DOM? Imagine for example if this provider and a section that has a query flushes to the stream at the same time, will the effect have a chance to run before that section is hydrated by React then? If not, there will be hydration mismatches.


I am not entirely sure to be honest. All of this stuff is experimental

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
KATT: What @TkDodo said, if you do something that changes the react query client cache, we'll subscribe to it and flush it next chance we get

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
KATT: LGTM

examples/react/nextjs-suspense-streaming/src/app/page.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const query = useQuery({
+    queryKey: ['wait', props.wait],
+    queryFn: async () => {
+      const path = `/api/wait?wait=${props.wait}`
+      const url = baseUrl + path
+
+      console.log('fetching', url)
+      const res: string = await (
+        await fetch(url, {
+          cache: 'no-store',
```
KATT, line 27: Do we need this `no-store`-directive even if it's a client component?

juliusmarminge: I dont think it does anything on the client

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
Ephem: Hm, oh yeah, it does loose it's values on re-render. Fair enough!

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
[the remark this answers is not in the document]
Ephem: It seems like an extra bit of indirection, but I don't mind it really, mostly wanted to make sure I understood why. 😃

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
Ephem: 👍

packages/react-query-next-experimental/src/ReactQueryStreamedHydration.tsx
Ephem: But what if you prime the cache before the rendering pass? Like if you await some data in a Server Component (maybe you need to generate metadata anyway), or you have some existing mechanism of prefetching that prefetches _before_ rendering?

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
Ephem: > I am not entirely sure to be honest. All of this stuff is experimental

Me neither and oh yeah, no worries, that's why I'm trying to think this through. 😃

Okay, so I looked at this closer and realized that it's only the first time this happens in an effect, after that, `window[id].push()` calls it directly when the script is evaluated which should always be _before_ hydrating that Suspense boundary.

So we could still theoretically have the problem that this effect hasn't run before the first Suspense boundary hydrates right (unless effects for previous boundaries _are_ guaranteed to run before hydration of child boundary)?

But at least, after that, right now the hydration happens before hydrating the Suspense boundary that is streaming in. Since this only happens on SSR/first page load, it should be pretty unlikely that we stream in a query that is already in the cache, at least with some NEW data. Even if we do, I think it's fine in this case (but wont be for the `<HydrateBoundary>` case since that will run on page navigations too).

PR #5601, opened 2023-06-18
fix: Update eslint-plugin-query build config

lachlancollins:
Updated to tsup v7, and fixed the `external` field to reduce build size from 284KB to 16KB.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
     "semver": "^7.5.1",
     "solid-js": "^1.6.13",
     "stream-to-array": "^2.3.0",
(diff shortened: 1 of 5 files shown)


Newbie012:
Are you sure about making `@typescript-eslint/utils` as external? Does the plugin work as expected without the need to install the package separately?

lachlancollins:
@Newbie012 seems to work fine in one of my projects. `@typescript-eslint/utils` will be installed as it's a dependency of `@typescript-eslint/eslint-plugin`, which is a devDependency of this package.

PR #5608, opened 2023-06-19
fix(devtools): padding for action buttons

TkDodo:
global `button` styles would otherwise affect buttons

### DIFF
--- a/packages/query-devtools/src/Devtools.tsx
+++ b/packages/query-devtools/src/Devtools.tsx
@@ -1664,6 +1664,7 @@ const getStyles = () => {
       max-width: 160px;
       border: 1px solid ${colors.darkGray[200]};
       cursor: pointer;
+      padding: 0;
       &:hover {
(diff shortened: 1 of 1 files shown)


TkDodo:
actually, there are more global styles that affect other parts of the devtools, e.g. the filters input:

Screenshot 2023-06-19 at 17 26 38

could we apply some sort of reset so that things inside the devtools are not affected by global styles of the site @ardeora ?

ardeora:
We can certainly add some sort of reset scoped to the devtools! Would you have a sample CSS files that you're currently using that is overriding the styles? So that I can crosscheck if everything in there is taken into account?

TkDodo:
@ardeora hmm, I've just seen that we have one that does:

```
input {
  background-color: #fff !important;
```

and the `!important` is of course the problem 😅 - it's what makes the filter ugly. No sure if this is worth / possible to fix in the devtools...

PR #5611, opened 2023-06-19
 [alpha][fix-5538] : sync Observer 'current' properties when optimistic reading occurs

incepter:
This is a follow-up to https://github.com/TanStack/query/pull/5573 for the v5-alpha version

### DIFF
--- a/packages/vue-query/src/useQueries.ts
+++ b/packages/vue-query/src/useQueries.ts
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import { isServer, QueriesObserver } from '@tanstack/query-core'
+import { QueriesObserver } from '@tanstack/query-core'
 import type {
   QueriesPlaceholderDataFunction,
   QueryKey,
@@ -11,7 +11,7 @@ import type { Ref } from 'vue-demi'
 import { computed, onScopeDispose, readonly, ref, watch } from 'vue-demi'
 
 import { useQueryClient } from './useQueryClient'
-import { cloneDeepUnref, noop } from './utils'
+import { cloneDeepUnref } from './utils'
 import type { UseQueryOptions } from './useQuery'
 import type { QueryClient } from './queryClient'
 import type { DistributiveOmit, MaybeRefDeep } from './types'
@@ -182,22 +182,21 @@ export function useQueries<
   )
   const state = ref(getCombinedResult()) as Ref<TCombinedResult>
 
-  const unsubscribe = ref(noop)
+  const unsubscribe = ref(() => {
+    // noop
+  })
 
   watch(
     client.isRestoring,
     (isRestoring) => {
       if (!isRestoring) {
         unsubscribe.value()
-        // Nuxt2 memory leak fix - do not subscribe on server
-        if (!isServer) {
-          unsubscribe.value = observer.subscribe(() => {
-            const [, getCombinedResultRestoring] = observer.getOptimisticResult(
-              defaultedQueries.value,
-            )
-            state.value = getCombinedResultRestoring()
-          })
-        }
+        unsubscribe.value = observer.subscribe(() => {
+          const [, getCombinedResultRestoring] = observer.getOptimisticResult(
+            defaultedQueries.value,
+          )
+          state.value = getCombinedResultRestoring()
+        })
         // Subscription would not fire for persisted results
         const [, getCombinedResultPersisted] = observer.getOptimisticResult(
           defaultedQueries.value,
--- a/packages/vue-query/src/useQueries.ts
+++ b/packages/vue-query/src/useQueries.ts
@@ -215,7 +215,10 @@ export function useQueries<
         defaultedQueries.value,
         options as QueriesObserverOptions<TCombinedResult>,
       )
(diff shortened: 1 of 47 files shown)


incepter:
When I run tests locally, I get a weird behavior of failures, but this one seems to be consistent: https://cloud.nx.app/runs/W8l5UK3y5A 

- The test in `vueQueryPlugin.test.ts` seems to regress, not sure why .. I tried seeing the observer part and it seems to work okay.
- `useQuery` tests seem weird, some tests when ran independently pass, and do not pass under `pnpm test` locally.


/cc @TkDodo do you know why tests did not get executed for this PR ? I think we should give it extra care.

Also, when running tests locally from my IDE, I needed to rebuild the query-core manually to take changes.

TkDodo:
> do you know why tests did not get executed for this PR ? I think we should give it extra care.

because GitHub sees you as a first time contributor and thus doesn't run the pipeline until someone approves :/ Have a look at the nx cloud report now: https://github.com/TanStack/query/pull/5611#issuecomment-1597122428

> Also, when running tests locally from my IDE, I needed to rebuild the query-core manually to take changes.

yeah this is a recent change by @lachlancollins in the alpha branch, I also stumbled upon it. 

You'd need to use the `pnpm run watch` command in the root to automatically pick up changes.

TkDodo (approved):

TkDodo:
@incepter ignore the vue-query tests, seems to fail on alpha everywhere. but the react-query test is probably an issue:
- https://cloud.nx.app/runs/cpF8yZtCAu

incepter:
Hey @TkDodo , sorry for the late reply.

You are right, in v5, placeHolderData acts like keepPreviousData too and has been enhanced.

The fix I found is to completely remove the need of testing on placeholder data, just assign the current properties whenever we have a new result.

The Tests fail in my windows machine which makes me always question whether the failure is real or not. Im not sure if we can improve this behavior

TkDodo:
thanks. The missing errors are just flakiness I guess. I'll likely merge both solutions (main + alpha) today

TkDodo:
@incepter the vue test fails locally for me as well on this PR.

@DamianOsipiuk could you maybe have a look at this here?

TkDodo:
the test in question works fine locally on the `alpha` branch, so maybe we are regressing here after all

incepter:
> the test in question works fine locally on the `alpha` branch, so maybe we are regressing here after all

Yeah it is surely a regression, since leaving the properties alone and not assigning anything in the observer would make the test pass.

I fact, this test was about to drive me crazy today since I do not know vue. But managed to finally find the culprit:

When we assign the `current` properties of the observer, say via `combineResult`, then calling `updateResult` won't actually result in notifying the component, because `updateResult` would call again `createResult` and get an object that shallow equal the previous result, then it won't get notified because the updateResult function just quits when objects are shallowly equal.

I think I've hit the end of the road this way. Any solution would I think of using the globals to make and would not be predictable and easy.

Do you guys think of a solution ? I think the v5 would follow new mechanics, I will be thinking of a new solution that would work and not break anything.

incepter:
Adding this seems to fix the [failing vue test](https://github.com/TanStack/query/pull/5611/commits/3930efcf4b18d7a1341621f081dd4319768f89a8). Im not sure of the used approach but it seems that it s broken in that function and we could solve it this way.

TkDodo:
> can I somehow run the CI before the approval ? It kind of annoys me not to be able to have the same behavior locally with the tests.

sorry, it's a github thing for first time contributors :/ I also think its annoying

packages/vue-query/src/useQueries.ts
```diff
@@ -214,7 +214,10 @@ export function useQueries<
         defaultedQueries.value,
         options as QueriesObserverOptions<TCombinedResult>,
       )
-      state.value = observer.getCurrentResult()
+      const [, getCombinedResultPersisted] = observer.getOptimisticResult(
+        defaultedQueries.value,
+      )
+      state.value = getCombinedResultPersisted()
```
TkDodo, line 220: @DamianOsipiuk can you have look here please?

DamianOsipiuk: Looks good 👍

DamianOsipiuk: Although this test relies on `useQueries`, which has different mechanism of data retrieval.
I wonder if this would still work for regular `useQuery`. Need to add a test for that since there we rely on `getCurrentResult` which appears to have changed behavior.

TkDodo: so - can we ship this or should we wait?

DamianOsipiuk: I would rather add the test first.

DamianOsipiuk (approved):
Seems to be working fine

packages/vue-query/src/useQueries.ts
incepter: Sorry guys I was on vacation this week and had too much family work to do all of a sudden .. Please let me know if there is anything I can do

DamianOsipiuk: @TkDodo it looks good from my side.

PR #5614, opened 2023-06-20
refactor(types): export query-cache related types

### LINKED ISSUE #5599: QueryCacheNotifyEvent types not exported
### Describe the bug

When subscribing to the query cache, like:

```ts
const queryCache: QueryCache = new QueryCache();
queryCache.subscribe(event => void work(event));
```

There's no way to type the function signature of `work` since the `QueryCacheNotifyEvent` types are not exported.

see above

### Steps to reproduce

see above

### Expected behavior

Would love to define a function like

```ts
const work = (event: QueryCacheNotifyEvent) => { ... }
```

Every time

### Screenshots or Videos

_No response_

### Platform

not important

### Tanstack Query adapter

react-query

### TanStack Query version

@tanstack/react-query@4.24.9

### TypeScript version

Version 4.9.5

### Additional context

Might be cool to get a linter installed that ensures that any public-facing types (used in other exported functions) are also exported.

indrajitbnikam:
Closes #5599 

export queryCache-related types for better DevX.

### DIFF
--- a/packages/query-core/src/index.ts
+++ b/packages/query-core/src/index.ts
@@ -1,7 +1,8 @@
 /* istanbul ignore file */
 
 export { CancelledError } from './retryer'
-export { QueryCache, QueryCacheNotifyEvent } from './queryCache'
+export { QueryCache } from './queryCache'
+export type { QueryCacheNotifyEvent } from './queryCache'
 export { QueryClient } from './queryClient'
 export { QueryObserver } from './queryObserver'
 export { QueriesObserver } from './queriesObserver'
--- a/packages/query-core/src/index.ts
+++ b/packages/query-core/src/index.ts
@@ -1,7 +1,7 @@
 /* istanbul ignore file */
 
 export { CancelledError } from './retryer'
-export { QueryCache } from './queryCache'
+export { QueryCache, QueryCacheNotifyEvent } from './queryCache'
 export { QueryClient } from './queryClient'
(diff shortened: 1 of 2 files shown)


TkDodo (changes requested):
Thanks.

- I would only export the `QueryCacheNotifyEvent` - not every sub-type of it.
- This won't be available in user-land without a re-export from the `index.ts` file

indrajitbnikam:
> This won't be available in user-land without a re-export from the index.ts file

So you want me to export `QueryCacheNotifyEvent` from the `query-core/src/QueryCache.ts` and the re-export from `index.ts` like below.

```TypeScript
export { QueryCache, QueryCacheNotifyEvent } from './queryCache'
```

Is my understanding correct?

indrajitbnikam:
Hi @TkDodo, I've addressed your review comments.

packages/query-core/src/index.ts
```diff
@@ -1,7 +1,7 @@
 /* istanbul ignore file */
 
 export { CancelledError } from './retryer'
-export { QueryCache } from './queryCache'
+export { QueryCache, QueryCacheNotifyEvent } from './queryCache'
```
TkDodo, line 4: I think this needs to be:

```suggestion
export { QueryCache } from './queryCache'
export type { QueryCacheNotifyEvent } from './queryCache'
```

PR #5617, opened 2023-06-21
Allow MutationObserver to use the same mutation similar to queries

devotox:


TkDodo:
can you please provide some information what this PR does? I don't understand it from just the title ...

PR #5619, opened 2023-06-22
fix: prevent uncaught exception in suspense()

ssendev:
`useQuery().suspense().catch(noop)` results in an uncaught exception and crashes node. This fixes it.


TkDodo:
Thanks. can you please add a test case that fails on the current `main` code but works with your fix? That way, we can ensure that we don't re-introduce the issue. Thank you

PR #5623, opened 2023-06-23
fix(core): Add pageParam type to QueryFunction generic types

mjlyons:
Although QueryFunctionContext allows setting the pageParam type through generics, QueryFunction uses QueryFunctionContext without exposing this typing functionality. This forces any queryFn typed as a QueryFunction to have an `any` pageParam. This change adds TPageParam to QueryFunction type, which is passed through to QueryFunctionContext.

This should be backward compatible and only affects typing (not runtime code).


TkDodo:
yeah I think we can merge this, but we've also fixed this holistically on v5 where the type of `TPageParam` will be inferred correctly from the new `defaultPageParam` option.

mjlyons:
@TkDodo wanted to check if there was anything you'd like me to do here? Noticed the Reach 17 test suite is failing, but appears to be unrelated.

PR #5626, opened 2023-06-25
refactor(rect-query): prefer client-only import over 'use client'

TkDodo:
for things that cannot work on the server, like the QueryClientProvider or the hooks; this will give better error message when attempted to be used in a server component

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "react-dom-17": "npm:react-dom@^17.0.2",
     "rimraf": "^3.0.2",
     "rollup": "^2.70.2",
-    "rollup-plugin-preserve-directives": "0.1.0",
+    "rollup-plugin-preserve-directives": "0.2.0",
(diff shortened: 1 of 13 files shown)


TkDodo:
@Ephem this is a follow-up to:
- https://github.com/TanStack/query/pull/5480

Ephem (approved):
Looks good to me!

I noticed the build was failing, I'm guessing the failing test is just a flaky one since it's a vue-test?

image

Also noticed it complained about the package `rollup-plugin-preserve-directives` was missing, do you know why? Seemed to happen after the bump right?

TkDodo:
I had to bump the package to fix the build error

Ephem:
Oh, there was that `.type`-error before the bump, I missed that. Now it's complaining it can't find the package though. 🤔

TkDodo:
the only remaining failure is the flaky vue test

PR #5630, opened 2023-06-26
Revert "refactor(rect-query): prefer client-only import over 'use client'"

TkDodo:
Reverts TanStack/query#5626


TkDodo:
I'll try the real approach

PR #5633, opened 2023-06-27
Feature/react server

TkDodo:


packages/react-query/src/shared.ts
```diff
@@ -0,0 +1,13 @@
...
+// Re-export core
+export * from '@tanstack/query-core'
+
+// things that can be imported in server components and client components
+export * from './types'
+export type { QueriesResults, QueriesOptions } from './useQueries'
+export { queryOptions } from './queryOptions'
+export type { QueryClientProviderProps } from './QueryClientProvider'
+export type { QueryErrorResetBoundaryProps } from './QueryErrorResetBoundary'
+export { HydrationBoundary } from './HydrationBoundary'
```
shuding, line 12: This will not work and cause an error in the `react-server` condition, so might need to be moved out from "shared".

Other setups look good though!

PR #5634, opened 2023-06-27
refactor: Sort imports with eslint

lachlancollins:
I personally find imports sorted in a consistent order easier to approach, just like how prettier formats code in a consistent way. The rules here are totally flexible but would mean all packages follow the same structure.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -5,6 +5,7 @@ const config = {
   extends: [
     'plugin:@typescript-eslint/eslint-recommended',
     'plugin:@typescript-eslint/recommended',
+    'plugin:import/recommended',
     'plugin:import/typescript',
     'prettier',
   ],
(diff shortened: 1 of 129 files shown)


TkDodo:
I agree, but it's gonna cause so many conflicts when merging back to `alpha` 😭 

Maybe we should this once v5 is released and we have only one branch to maintain?

lachlancollins:
> I agree, but it's gonna cause so many conflicts when merging back to `alpha` 😭
> 
> Maybe we should this once v5 is released and we have only one branch to maintain?

@TkDodo I was just going to merge main into alpha and only keep the changes to `.eslintrc.cjs`, then re-run eslint fix. I'm hoping this would actually mean there would be fewer conflicts between v4 and v5!

TkDodo:
okay, let me just ship the fix for 5538 first. It has two PRs - one for main and one for alpha.

PR #5659, opened 2023-07-02
feat: Support Svelte v4

lachlancollins:
I'm sure this already worked in svelte v4, but this adds it as a peer dependency and updates dev dependencies. I also moved `jsdom` out of the svelte `package.json` to the root `package.json` since it is used by all packages for vitest.


lachlancollins:
Pending https://github.com/testing-library/svelte-testing-library/pull/244

PR #5666, opened 2023-07-03
fix(eslint-plugin-query): Use tsup to generate types

lachlancollins:
The recent tsup 7.1.0 introduces separate CJS/ESM type declaration generation (https://github.com/egoist/tsup/pull/934). The reasoning behind why this is important is explained [here](https://twitter.com/atcb/status/1673491213415247872). I'm hoping this should allow the package to pass [ATTW](https://arethetypeswrong.github.io/)!

I was previously unable to make tsup generate dts correctly for this package, but I tracked the issue to the `external` setting. `@typescript-eslint/utils` should actually be a dependency of this package, since it is directly used by the plugin. Anyone who installs this package as a devDependency has all dependencies also installed as dev-only, so this doesn't change anything for consumers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
   },
   "namespace": "@tanstack",
   "devDependencies": {
+    "@arethetypeswrong/cli": "^0.4.2",
     "@babel/core": "^7.21.8",
     "@babel/preset-env": "^7.21.5",
     "@babel/preset-react": "^7.18.6",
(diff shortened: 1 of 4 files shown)


lachlancollins:
I've also raised an issue r.e. the unexpected impact of the `external` option in tsup here: https://github.com/egoist/tsup/issues/943

PR #5694, opened 2023-07-10
fix(solid-query): hydrate preloaded data correctly

marbemac:
Make sure that on the server we always resolve the resource data to a hydrate-able value, otherwise solid will silently call `onHydrated` on the client with an empty object which can cause a variety of problems.

The specific code branch we were missing was when `createQuery` is called and `observer.getOptimisticResult(defaultedOptions)` returns an already loaded query (if, for example, `preloadQuery` was called prior to the components mounting - for example with router data loaders).

---

As always, these SSR / hydration cases are difficult to test, so I added an example to the `solid-start-streaming` to manually verify for now. This example simply calls `preloadQuery` on the server, and then renders a component that uses that same query.

**Before 1**

This is with `deferStream: false` on the user query - note how the content ends up getting rendered twice, and in the dev console you can see that `onHydrated` is called with an empty object. Things work as expected after the changes in this PR.

Screenshot 2023-07-10 at 11 59 11 AM

**Before 2**

This is with `deferStream: true` on the user query - note how it ends up calling the user query on the server AND on the client (see logs in dev console), and in the dev console you can see that `onHydrated` is called with an empty object (which causes the client to think that it needs to refetch, hence the client call).

Screenshot 2023-07-10 at 11 58 26 AM

**After**

Screenshot 2023-07-10 at 11 57 58 AM

### DIFF
--- a/packages/solid-query/package.json
+++ b/packages/solid-query/package.json
@@ -36,7 +36,7 @@
     "test:eslint": "eslint --ext .ts,.tsx ./src",
     "test:types": "tsc --noEmit",
     "test:lib": "vitest run --coverage",
-    "test:lib:dev": "pnpm run test:lib --watch",
+    "test:lib:dev": "vitest watch --coverage",
     "test:build": "publint --strict",
     "build": "rollup --config rollup.config.js"
   },
--- a/packages/solid-query/src/createBaseQuery.ts
+++ b/packages/solid-query/src/createBaseQuery.ts
@@ -59,14 +59,15 @@ const hydrateableObserverResult = <
   query: Query<TQueryFnData, TError, TData, TQueryKey>,
   result: QueryObserverResult<T2, TError>,
 ): HydrateableQueryState<T2, TError> => {
-  const { refetch, ...rest } = unwrap(result)
+  // Including the extra properties is only relevant on the server
+  if (!isServer) return result as HydrateableQueryState<T2, TError>
 
   return {
-    ...rest,
+    ...unwrap(result),
 
     // cast to refetch function should be safe, since we only remove it on the server,
     // and refetch is not relevant on the server
-    refetch: (isServer ? undefined : refetch) as HydrateableQueryState<
+    refetch: undefined as unknown as HydrateableQueryState<
       T2,
       TError
     >['refetch'],
@@ -79,8 +80,8 @@ const hydrateableObserverResult = <
     isInvalidated: query.state.isInvalidated,
 
     // Unsetting these properties on the server since they might not be serializable
-    fetchFailureReason: isServer ? null : query.state.fetchFailureReason,
-    fetchMeta: isServer ? null : query.state.fetchMeta,
+    fetchFailureReason: null,
+    fetchMeta: null,
   }
 }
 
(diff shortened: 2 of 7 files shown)


packages/solid-query/package.json
```diff
@@ -36,7 +36,7 @@
     "test:eslint": "eslint --ext .ts,.tsx ./src",
     "test:types": "tsc --noEmit",
     "test:lib": "vitest run --coverage",
-    "test:lib:dev": "pnpm run test:lib --watch",
+    "test:lib:dev": "vitest watch --coverage",
```
marbemac, line 39: `vitest --watch` is a thing, `vitest run --watch` isn't - they have a dedicated `watch` command instead.

packages/solid-query/src/createBaseQuery.ts
```diff
@@ -40,6 +42,49 @@ function reconcileFn<TData, TError>(
...
 }
 
+type HydrateableQueryState<TData, TError> = QueryObserverResult<TData, TError> &
+  QueryState<TData, TError>
+
+/**
+ * Solid's `onHydrated` functionality will silently "fail" (hydrate with an empty object)
+ * if the resource data is not serializable.
+ */
+const hydrateableObserverResult = <
```
marbemac, line 52: Pulled this out into it's own function since it's used in multiple spots now.

packages/solid-query/src/createBaseQuery.ts
```diff
@@ -71,41 +120,19 @@ export function createBaseQuery<
...
-          // fetchFailureReason: query.state.fetchFailureReason,
-          // fetchMeta: query.state.fetchMeta,
-          isInvalidated: query.state.isInvalidated,
-        }
+        const unwrappedResult = hydrateableObserverResult(query, result)
 
         if (unwrappedResult.isError) {
-          if (process.env['NODE_ENV'] === 'development') {
-            console.error(unwrappedResult.error)
-          }
```
marbemac, line 102: Wasn't sure if this log is supposed to be here or an accidental leftover - lmk if it should stay and I can add it back.

packages/solid-query/src/createBaseQuery.ts
```diff
@@ -71,41 +120,19 @@ export function createBaseQuery<
...
           reject(unwrappedResult.error)
-        }
-        if (unwrappedResult.isSuccess) {
-          // Use of any here is fine
-          // We cannot include refetch since it is not serializable
-          resolve(unwrappedResult as any)
+        } else {
```
marbemac, line 134: IMHO we should always fall back to calling `resolve()`. I ran into an issue where neither `reject` nor `resolve` were called, resulting in a hung promise (and this was on the server, resulting in the entire page being hung). LMK if there's a reason to sometimes not resolve the promise.

packages/solid-query/src/createBaseQuery.ts
```diff
@@ -159,8 +186,10 @@ export function createBaseQuery<
             unsubscribe = createClientSubscriber()
           }
         }
+
         if (!state.isLoading) {
-          resolve(state)
+          const query = observer.getCurrentQuery()
+          resolve(hydrateableObserverResult(query, state))
```
marbemac, line 192: Here is the code path that wasn't covered - this is triggered if the data is already pre-loaded prior to `createQuery`. We need to "sanitize" the state on the server before resolving, so that it can be correctly hydrated back to the client.

TkDodo:

PR #5702, opened 2023-07-12
Add stability gotcha to mutations.md

alexburner:
From TkDodo on Stack Overflow: https://stackoverflow.com/a/68924092

### DIFF
--- a/docs/react/guides/mutations.md
+++ b/docs/react/guides/mutations.md
@@ -387,6 +387,21 @@ export default function App() {
 
 We also have an extensive [offline example](../examples/react/offline) that covers both queries and mutations.
 
+### Return value stability for React dependencies
+
+The mutate function itself is stable, but the object returned from useMutation is not. If you destruct, you can add it to your dependency array:
+
+[//]: # 'Example12'
+
+```tsx
+const { mutate } = useMutation(…)
+
+// This will only be called on mount
+useEffect(() => { mutate() }, [mutate])
+```
+
+[//]: # 'Example12'
+
 [//]: # 'Materials'
 
 ## Further reading


docs/react/guides/mutations.md
```diff
@@ -387,6 +387,21 @@ export default function App() {
...
+[//]: # 'Example12'
+
+```tsx
+const { mutate } = useMutation(…)
+
+// This will only be called on mount
+useEffect(() => { mutate() }, [mutate])
+```
+
+[//]: # 'Example12'
```
TkDodo, line 403: since this is react specific, please wrap the whole block including the heading in `[//]: # 'Info2'` (instead of `[//]: # 'Example12'`), and then overwrite it with an empty implementation for the vue docs, as we do here:

https://github.com/TanStack/query/blob/fa04a1d26b2900163ffb2e2ae238735fed4dddaa/docs/vue/guides/mutations.md#L30-L35

TkDodo: @alexburner FYI. If you don't want to work on this anymore, I'll close the PR.

alexburner: Ah yeah I got confused by your comment and wandered away

I do think this is an important gotcha to document, if anyone wants to put a ribbon on this I'd appreciate it

LucasUnplugged: @TkDodo is the convention here that `Info` is the React keyword, and `Example` is the Vue keyword? And the number is just sequential?

If so, I'm I understanding correctly that this should be?

```suggestion
### Return value stability for React dependencies

The mutate function itself is stable, but the object returned from useMutation is not. If you destruct, you can add it to your dependency array:

[//]: # 'Info2'

```tsx
const { mutate } = useMutation(…)

// This will only be called on mount
useEffect(() => { mutate() }, [mutate])
`` `

[//]: # 'Info2'

[//]: # 'Example12'
[//]: # 'Example12'
```

**NOTE: the extra space in the last ``` is purely because of the nesting of code within a suggestion, and should be removed**

CC: @alexburner

alexburner:
Yeehaw! Looks like we got there: https://tanstack.com/query/v5/docs/framework/react/guides/render-optimizations#referential-identity

PR #5703, opened 2023-07-13
feat(vue-query): support invoke in vue `runWithContext()`

Mini-ghost:
Currently, we cannot use `useQuery` and `useMutation` outside of `setup`, maybe a nuxt plugin or middleware or vue router `beforeEach`.

In Vue 3.3, we can use `app.runWithContext` to look up injections from the values provided by the current app without being restricted to use in `setup`

This is Vue document example:

```ts
import { inject } from 'vue'

app.provide('id', 1)

const injected = app.runWithContext(() => {
  return inject('id')
})

console.log(injected) // 1
```
[Application API | Vue.js # app.runWithContext()](https://vuejs.org/api/application.html#app-runwithcontext)

So I adjusted the order of throwing errors, we can try to `inject` first, and if the `queryClient` cannot be found, it will be determined which error message to display according to the result of `getCurrentInstance()`.

This change will not break projects before Vue 3.3, it is backward compatible


DamianOsipiuk:
@Mini-ghost Could you please add a test to verify that `vue-query` composable executed in `app.runWithContext` would not leak, by checking if `onScopeDispose` is properly called?

Mini-ghost:
@DamianOsipiuk I tried it and noticed that `onScopeDispose` is not called after `app.unmount()`. I'm not sure if this is an intended feature for vue or maybe a bug that needs to be tracked down.

### Reproduction
1. Open reproduction link: [onScopeDispose will not be called in runWithContext - stackblitz](https://stackblitz.com/edit/onscopedispose-will-not-be-called-in-runwithcontext?file=src%2Fmain.ts&terminal=dev)
2. Open `Devtool > Console`
3. Check if `cope dispose` is printed

posva (commented):
You need to add a second parameter to inject to avoid Vue warning

PR #5734, opened 2023-07-18
feat: add new function type to notifyOnChangeProps

paulobmarcos:
Following up with the implementation of [Discussion #5588](https://github.com/TanStack/query/discussions/5588).

**Summary:** 

With `notifyOnChangeProps` we can make a component only re-render if any of the listed properties change. This is very useful and we can completely stop all re-renders if needed with an empty array `[]`.

However, this still requires one extra re-render to change the `notifyOnChangeProps` value and there are use cases where the extra re-render could be considered expensive (React-Native and transitions).

**Proposal:**
The `notifyOnChangeProps` would be extended to also accept a callback. This way, we could change the return value of `notifyOnChangeProps` without an extra re-render.

CC: @TkDodo @ferrannp


docs/react/react-native.md
```diff
@@ -76,3 +76,52 @@ export function useRefreshOnFocus<T>(refetch: () => Promise<T>) {
...
+
+This custom hook provides a `notifyOnChangeProps` option that will return an empty array whenever a screen goes out of focus - effectively stopping any re-renders on that scenario. Whenever the screens gets in focus again, the behavior goes back to normal.
+
+```tsx
+import React from 'react'
+import { UseQueryOptions } from '@tanstack/react-query'
+import { useFocusEffect } from '@react-navigation/native'
+
+export function useFocusNotifyOnChangeProps(
+  notifyOnChangeProps: UseQueryOptions['notifyOnChangeProps']
```
paulobmarcos, line 92: What is the best way to get the `notifyOnChangeProps` type?

docs/react/react-native.md
```diff
@@ -76,3 +76,52 @@ export function useRefreshOnFocus<T>(refetch: () => Promise<T>) {
...
+
+  useFocusEffect(
+    React.useCallback(() => {
+      focusedRef.current = true
+
+      return () => {
+        focusedRef.current = false
+      }
+    }, [])
+  )
```
paulobmarcos, line 109: We can probably simply this further with `navigation.isFocused()`. Something we have to try out.

docs/react/react-native.md
```diff
@@ -76,3 +76,52 @@ export function useRefreshOnFocus<T>(refetch: () => Promise<T>) {
...
+    }
+
+    if (typeof focusedNotifyOnChangePropsRef.current === 'function') {
+      return focusedNotifyOnChangePropsRef.current()
+    }
+
+    return focusedNotifyOnChangePropsRef.current
+  }, [])
+}
+```
```
paulobmarcos, line 123: Would be good to give an example on how this would actually be used:

```tsx
const MyExampleComponent = () => {
  const notifyOnChangeProps = useFocusNotifyOnChangeProps('all');
  const { dataUpdatedAt } = useQuery({
    queryKey: ['myExampleKey'],
    queryFn: async () => {
      const response = await fetch('https://api.github.com/repos/tannerlinsley/react-query');
      return response.json();
    },
    notifyOnChangeProps,
  });

  return <div>DataUpdatedAt: {dataUpdatedAt}</div>;
};
```

packages/query-core/src/types.ts
```diff
@@ -200,9 +200,13 @@ export interface QueryObserverOptions<
...
    * When set to `['data', 'error']`, the component will only re-render when the `data` or `error` properties change.
    * When set to `'all'`, the component will re-render whenever a query is updated.
+   * When set to a function, the function will be executed to compute the list of properties.
    * By default, access to properties will be tracked, and the component will only re-render when one of the tracked properties change.
    */
-  notifyOnChangeProps?: Array<keyof InfiniteQueryObserverResult> | 'all'
+  notifyOnChangeProps?:
+    | Array<keyof InfiniteQueryObserverResult>
+    | 'all'
+    | (() => Array<keyof InfiniteQueryObserverResult> | 'all')
```
paulobmarcos, line 209: @TkDodo quick question - it would be totally fine to pass in a different function reference on every re-render, right?

I believe it would be same thing as passing in a new reference of `['data', 'error']`. 

```tsx
const { dataUpdatedAt } = useQuery({
    queryKey: ['myExampleKey'],
    queryFn: async () => {
      const response = await fetch('https://api.github.com/repos/tannerlinsley/react-query');
      return response.json();
    },
    notifyOnChangeProps: ['data', 'error']
  });
```

docs/react/react-native.md
TkDodo: yeah that's the only way right now, unless you want to extract and export the type separately?

https://github.com/TanStack/query/blob/dcda9200c9d29ae0b56b1d78ded0d9d721ba6b77/packages/query-core/src/types.ts#L266

docs/react/react-native.md
TkDodo:
```
const notifyOnChangeProps = useFocusNotifyOnChangeProps('all');
```

`'all'` is actually not the default, but `undefined` is, which does smart tracking of used properties.

packages/query-core/src/types.ts
TkDodo: yes, referential stability doesn't matter here.

packages/query-core/src/queryObserver.ts
```diff
@@ -636,16 +636,23 @@ export class QueryObserver<
...
 
+      if (!this.options.notifyOnChangeProps && !this.trackedProps.size) {
+        return true
+      }
+
       const { notifyOnChangeProps } = this.options
+      const notifyOnChangePropsValue =
+        typeof notifyOnChangeProps === 'function'
+          ? notifyOnChangeProps()
+          : notifyOnChangeProps
```
TkDodo, line 647: hmm, why isn't it enough to move `const notifyOnChangePropsValue = ...` to the beginning, and then use `notifyOnChangePropsValue` everywhere instead of `notifyOnChangeProps`. That way, we could keep the one if statement like it was before:

```
      if (
        notifyOnChangePropsValue === 'all' ||
        (!notifyOnChangePropsValue && !this.trackedProps.size)
      ) {
        return true
      }
```

or am I missing something?

docs/react/react-native.md
paulobmarcos: Changed to `undefined` 👍

packages/query-core/src/queryObserver.ts
paulobmarcos: No specific reason - changed to one single condition 👍

docs/react/react-native.md
```diff
@@ -76,3 +76,64 @@ export function useRefreshOnFocus<T>(refetch: () => Promise<T>) {
 ```
 
 In the above code, `refetch` is skipped the first time because `useFocusEffect` calls our callback on mount in addition to screen focus.
+
+## Disable re-renders on out of focus Screens
+
+In some situations, including performance concerns, you may want to stop re-renders when a React Native screen gets out of focus. To achieve this we can use `useFocusEffect` from `@react-navigation/native` together with the `notifyOnChangeProps` query option.
```
incepter, line 82: I have a question if you may, when the screen gets out of focus, does effects clean up get called ?

TkDodo (approved):

docs/react/react-native.md
paulobmarcos: Hi @incepter!

Yes, [useFocusEffect](https://reactnavigation.org/docs/use-focus-effect/) runs the cleanup whenever a screen/component gets out of focus :+1:

ferrannp:
Coming from the past. This seems wrong again https://tanstack.com/query/latest/docs/framework/react/react-native#disable-queries-on-out-of-focus-screens @TkDodo ?

PR #5653, opened 2023-06-30
docs(query-keys): clarify query key refetching

meatnordrink:
Update the query-key documentation to note that changing data included in a query key will automatically refetch the query data.

It took me awhile to find this information, and I expected to find it here. I consulted with another engineer on our team, and they agreed that they would expect to see it stated more clearly here. So we felt this addition would be helpful for people trying to onboard to React Query.

PR #5735, opened 2023-07-18
Docs: Fix unintentional visible comment in query-keys.md

meatnordrink:
It appears that, somewhere in the course of [this](https://github.com/TanStack/query/pull/5653/files) PR, a comment got bumped such that it's now visible in the production docs. This fixes that, as well as excludes a parenthetic that I believe was accidentally included from an italicized section.

### DIFF
--- a/docs/react/guides/query-keys.md
+++ b/docs/react/guides/query-keys.md
@@ -91,12 +91,15 @@ function Todos({ todoId }) {
 
 [//]: # 'Example5'
 
-Note that query keys act as dependencies for your query functions. Adding dependent variables to your query key will ensure that queries are cached independently, and that any time a variable changes, *queries will be refetched automatically (depending on your `staleTime` settings).* (See the [exhaustive-deps](../eslint/exhaustive-deps) section for more information and examples.)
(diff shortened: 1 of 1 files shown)


docs/react/guides/query-keys.md
```diff
@@ -91,7 +91,8 @@ function Todos({ todoId }) {
 
 [//]: # 'Example5'
 
-Note that query keys act as dependencies for your query functions. Adding dependent variables to your query key will ensure that queries are cached independently, and that any time a variable changes, *queries will be refetched automatically (depending on your `staleTime` settings).* (See the [exhaustive-deps](../eslint/exhaustive-deps) section for more information and examples.)
+Note that query keys act as dependencies for your query functions. Adding dependent variables to your query key will ensure that queries are cached independently, and that any time a variable changes, *queries will be refetched automatically* (depending on your `staleTime` settings). (See the [exhaustive-deps](../eslint/exhaustive-deps) section for more information and examples.)
+
```
meatnordrink, line 95: The meaningful change here is the addition of this space, keeping the `[//]: # 'Materials'` from being visible (as it is currently in production).

Issue #3572, opened 2022-04-29
Weird results when keepPreviousData is used with transitions


faradaytrs:
### Describe the bug

Sandbox link 

When I click decrease (in the second part of the example), I get unexpected results on example 2.2;
Because other suspense boundaries depend on the same state, useQuery waits for them to render and only starts fetching after transition is complete.
Can somebody explain why we get this result?

If i remove other two Suspense boundaries, everything seems normal.
With keepPreviousData on, it starts fetching only when transition is complete, not sure if it's bug

One more strange thing I noticed is fetching triggered by refetchOnWindowFocus doesn't cause fallback in suspense? Is it right behavior?

### Your minimal, reproducible example

https://codesandbox.io/s/funny-leftpad-wf7kbz?file=/src/App.js

### Steps to reproduce

Click decrease:
See that full update requires 10 seconds instead of 5.

### Expected behaviour

1. I expected 2.2 to work similarly with 2.1 in sandbox
2. I expect refetchOnWindowFocus trigger fallback when used with suspense: true and current behaviour can be opt in using useDeferredValue or keepPreviousValue

### How often does this bug happen?

Every time

### Platform

Mac OS Monterey
Latest Edge

### react-query version

4.0.0-beta.7

faradaytrs:
I moved this from discussions because I believe at least one is a bug. 
https://github.com/tannerlinsley/react-query/discussions/3567

TkDodo:
> One more strange thing I noticed is fetching triggered by refetchOnWindowFocus doesn't cause fallback in suspense? Is it right behavior?

This is on purpose because `suspense` replaces a "hard loading state" where we have no data yet. For background refetches, the stale data is shown and the query will stay in `success` state.

---

I have looked at the sandbox, but I honestly have no idea why this is happening. It looks like a bug, but I would appreciate some help.

TkDodo:
Thought about this a bit: `keepPreviousData` becomes unnecessary when you use `suspense` + `startTransition`. The screen will be kept in tact while the transition is going on. If we just remove `keepPreviousData` from your example, it works as expected.

`keepPreviousData` is meant to give you the "suspense" experience without actually using suspense.

recurrence:
I'm seeing a somewhat similar issue with infinite queries.  If I provide initialData, then previousData seems to be ignored.  Hence, any change to the query keys renders the initial data while the update is fetching.  This is very jarring as a user experience with initial data popping in every time the query key changes.

[...]

PR #5739, opened 2023-07-18
feat(react-query): useSuspenseQuery

TkDodo:


TkDodo:
@manudeli FYI

packages/react-query/src/useSuspenseQuery.ts
```diff
@@ -0,0 +1,30 @@
...
+  options: Omit<
+    UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+    'enabled' | 'suspense' | 'throwOnError' | 'placeholderData'
+  >,
+  queryClient?: QueryClient,
+): Omit<DefinedUseQueryResult<TData, TError>, 'isPlaceholderData'> {
+  return useBaseQuery(
+    {
+      ...options,
+      enabled: true,
```
manudeli, line 23:
```suggestion
```

Could I know why you forced `enabled: true` without exposing `enabled`? I thought enabled also could be exposed. I'd like to show you how exposing `enabled` in @suspensive/react-query. 
like this code https://github.com/suspensive/react/blob/main/packages/react-query/src/useSuspenseQuery.ts

Is it too many type overriding function?

If we force `enabled: true`, existing useQuery users can fetch according to the `enabled` option, but if it is a removed feature, it will be regrettable when using useSuspenseQuery. What's your Idea of removing?

manudeli (commented):
I am curious why you want to force enabled:true

packages/react-query/src/useSuspenseQuery.ts
TkDodo: The main reason is that with suspense, we'd expect `data` to be defined, which is not the case if you disable a Query. We could add a type overload for `enabled`, but I'm not sure if it's needed (or if it's possible if we switch to a react specific cache implementation).

we can always add `enabled` back again in a minor if necessary, but I'd like to start out with the most basic version.

manudeli: > we can always add enabled back again in a minor if necessary, but I'd like to start out with the most basic version.

Cool, You mean we want to avoid unnecessary BREAKING CHANGE. I understood your intention of why you didn't expose enabled right now.

manudeli:
@TkDodo 
I made https://github.com/TanStack/query/pull/5755 to add and update types for useSuspenseQuery. Review it please
1. UseSuspenseQueryOptions
2. UseSuspenseQueryResult

manudeli:
> @TkDodo I made #5755 to add and update types for useSuspenseQuery. Review it please
> 
> 1. UseSuspenseQueryOptions
> 2. UseSuspenseQueryResult
> 
> If you accept this change I also want to add UseSuspenseInfiniteQueryOptions, UseSuspenseInfniteQueryResult

I checked these types is merged, I'll add UseSuspenseInfiniteQueryOptions, UseSuspenseInfniteQueryResult right now.

docs/react/guides/suspense.md
```diff
@@ -98,10 +96,53 @@ const App: React.FC = () => {
...
+
+```tsx
+import { useSuspenseQuery } from '@tanstack/react-query'
+
+const { data } = useSuspenseQuery({ queryKey, queryFn })
+```
+
+This has the same effect as setting the `suspense` option to `true` in the query config, but it works better in TypeScript, because `data` is guaranteed to be defined (as errors and loading states are handled by Suspense- and ErrorBoundaries).
+
+On the flip side, you therefore can't conditionally enable / disable the Query. `placeholderData` also doesn't exist for this Query. To prevent the UI from being replaced by a fallback during an update, wrap your updates that change the QueryKey into [startTransition](https://react.dev/reference/react/Suspense#preventing-unwanted-fallbacks).
```
ojj1123, line 111: I have a few questions in this description.
Could you explain why `placeholderData` is missing from the `useSuspenseQuery` option?
I'm a little confused as to what the suspense mode and placeholderData options are related to.

incepter: I don't have much context, i guess because placeholder is only used when state is 'pending', which with useSuspenseQuery will suspend the current component and thus won't be used. Under the hood it os passed to a code-path where it is supported, but it is just ignored in this particular case. But @TkDodo will have an accurate answer

manudeli:
@TkDodo
I made https://github.com/TanStack/query/pull/5766 to add and update types for useSuspenseInfiniteQuery. Review it please

1. UseSuspenseInfiniteQueryOptions
2. UseSuspenseInfiniteQueryResult

docs/react/guides/suspense.md
TkDodo: @incepter is right, that's the reason. `placeholderData` is meant to be shown while you are initially fetching. With suspense, that is what `Suspense` boundaries and fallbacks can do.

Also, in v5, we are using `placeholderData: (previous) => previous` as a replacement for `keepPreviousData`. It allows us to show data of a previous query key while the new query key is fetching for the first time. This is useful for e.g. pagination.

With suspense, this is what `transitions` are doing: wrap the thing that triggers the change for the query key into a `startTransition` to get the same behaviour, so you don't need `placeholderData`. It also doesn't work well with suspense, see this issue:

- https://github.com/TanStack/query/issues/3572

ojj1123: https://github.com/TanStack/query/issues/3572#issuecomment-1304749251
> `keepPreviousData` becomes unnecessary when you use `suspense` + `startTransition`.
> ...
> `keepPreviousData` is meant to give you the "suspense" experience without actually using suspense.

Thanks! @TkDodo 
I understand why you removed the `placeholderData` option.

manudeli:
@ChanhyukPark-Tech (one member of [@suspensive/react-query](https://tanstack.com/query/latest/docs/react/community/suspensive-react-query) team, same as me)
I'd like to request that you help us(me and tkdodo) making official interfaces of @tanstack/react-query about suspense, by Review, Pull Request, etc. your participation will be of great help in making this to a reality.

ChanhyukPark-Tech (commented):
It's cool👍

PR #5748, opened 2023-07-20
fix potential typo

DaleSeo:
I'm not a native English speaker, so I may be mistaken, but it sounds off.

### DIFF
--- a/docs/react/guides/queries.md
+++ b/docs/react/guides/queries.md
@@ -117,7 +117,7 @@ TypeScript will also narrow the type of `data` correctly if you've checked for `
 
 ### FetchStatus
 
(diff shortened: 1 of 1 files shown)


docs/react/guides/queries.md
```diff
@@ -117,7 +117,7 @@ TypeScript will also narrow the type of `data` correctly if you've checked for `
 
 ### FetchStatus
 
-In addition to the `status` field, the `result` object, you will also get an additional `fetchStatus`property with the following options:
+In addition to the `status` field, from the `result` object, you will also get an additional `fetchStatus`property with the following options:
```
TkDodo, line 120: I'd just do:

```suggestion
In addition to the `status` field, you will also get an additional `fetchStatus` property with the following options:
```

PR #5754, opened 2023-07-20
feat(svelte-query): Export context functions

frederikhors:


lachlancollins (approved):
@frederikhors I figured this might be a good idea after I saw the answer in #5378! If you also find time to contribute a `svelte-query-persist-client` package based on your implementation I would greatly appreciate it :)

frederikhors:
> @frederikhors I figured this might be a good idea after I saw the answer in #5378! If you also find time to contribute a `svelte-query-persist-client` package based on your implementation I would greatly appreciate it :)

My time is limited but I'll definitely try.

PR #5755, opened 2023-07-20
fix(react-query): types for useSuspenseQuery

manudeli:
# Add types for useSuspenseQuery like useQuery
1. UseSuspenseQueryOptions
2. UseSuspenseQueryResult

### DIFF
--- a/packages/react-query/src/types.ts
+++ b/packages/react-query/src/types.ts
@@ -12,6 +12,7 @@ import type {
   QueryKey,
   QueryObserverOptions,
   QueryObserverResult,
+  QueryObserverSuccessResult,
   WithRequired,
 } from '@tanstack/query-core'
 
@@ -34,6 +35,16 @@ export interface UseQueryOptions<
 > extends WithRequired<
     UseBaseQueryOptions<TQueryFnData, TError, TData, TQueryFnData, TQueryKey>,
     'queryKey'
+  > {}
+
+export interface UseSuspenseQueryOptions<
+  TQueryFnData = unknown,
+  TError = DefaultError,
+  TData = TQueryFnData,
+  TQueryKey extends QueryKey = QueryKey,
+> extends Omit<
+    UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+    'enabled' | 'suspense' | 'throwOnError' | 'placeholderData'
   > {}
 
 export interface UseInfiniteQueryOptions<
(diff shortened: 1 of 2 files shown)


packages/react-query/src/types.ts
```diff
@@ -36,6 +37,16 @@ export interface UseQueryOptions<
...
 
+export interface UseSuspenseQueryOptions<
+  TQueryFnData = unknown,
+  TError = DefaultError,
+  TData = TQueryFnData,
+  TQueryKey extends QueryKey = QueryKey,
+> extends Omit<
+    UseQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+    'enabled' | 'suspense' | 'throwOnError' | 'placeholderData'
+  > {}
```
manudeli, line 48: I thought It would be better that useSuspenseQuery have its `QueryOptions` type and `Result` type like `UseQueryOptions`, `UseQueryResult` for useQuery

If useSuspenseQuery should be provided as an API of the same hierarchy as useQuery, I thought it would be better to provide the types(`UseSuspenseQueryOptions`, `UseSuspenseQueryResult`) in the same hierarchy.

packages/react-query/src/types.ts
```diff
@@ -65,6 +76,11 @@ export type UseQueryResult<
   TError = DefaultError,
 > = UseBaseQueryResult<TData, TError>
 
+export type UseSuspenseQueryResult<
+  TData = unknown,
+  TError = DefaultError,
+> = Omit<QueryObserverSuccessResult<TData, TError>, 'isPlaceholderData'>
```
manudeli, line 82: I thought If we don't have `enabled` option and have default `suspense` option, we don't have to use `DefinedUseQueryResult` containing unnecessary `QueryObserverRefetchErrorResult`.

```ts
type DefinedQueryObserverResult<TData = unknown, TError = DefaultError> = QueryObserverRefetchErrorResult<TData, TError> | QueryObserverSuccessResult<TData, TError>;
```

We just need only `QueryObserverSuccessResult` for useSuspenseQuery.
So we could type-narrow `isSuccess: false`, `error`, `isError: true` etc.

```ts
interface QueryObserverRefetchErrorResult<TData = unknown, TError = DefaultError> extends QueryObserverBaseResult<TData, TError> {
    data: TData;
    error: TError;
    isError: true;
    isPending: false;
    isLoadingError: false;
    isRefetchError: true;
    isSuccess: false;
    status: 'error';
}
interface QueryObserverSuccessResult<TData = unknown, TError = DefaultError> extends QueryObserverBaseResult<TData, TError> {
    data: TData;
    error: null;
    isError: false;
    isPending: false;
    isLoadingError: false;
    isRefetchError: false;
    isSuccess: true;
    status: 'success';
}
```

manudeli (commented):
I commented as my intention

manudeli:
GitHub workflows need to be fixed in [base branch](https://github.com/TanStack/query/pull/5739) first, and update branch is needed.

image

TkDodo (approved):

PR #5761, opened 2023-07-21
feat(codemod): add codemod that renames the `Hydrate` component usages to `HydrationBoundary` usages

balazsmatepetro:
Closes: https://github.com/TanStack/query/issues/5759


TkDodo:
can you add documentation to the upgrade guide please?

maybe it would also be good to have one way to run all migrations for v5, with one command?

balazsmatepetro:
> can you add documentation to the upgrade guide please?
> 
> maybe it would also be good to have one way to run all migrations for v5, with one command?

Sure, I will!

Issue #5031, opened 2023-02-24
The queryClient is made reactive (!?) in the options causing severe performance issues in vue


simonwep:
### Describe the bug

Hey! We make extensive use of the query client by, for example, using a query in a table to resolve a value for each cell. We also initialize our own instance of `QueryClient` and pass it as option to each `useXXX`-call like so:

```ts
const queryClient = new QueryClient({ ... });
const query = useQuery(..., ..., {
    queryClient,
});
```

We have to do this since we use the query-client outside of vue component-instances and, as of right now, the query client is by default [injected](https://github.com/TanStack/query/blob/8b51e8b690f68190c06c865a84fb6a22c5c29d34/packages/vue-query/src/useQueryClient.ts#L14) which would throw an error in this instance.

For some reason the queryClient seems to be made reactive as this causes severe performance issues in our case since vue traverses the whole client for reactive changes on _every change_.

### Your minimal, reproducible example

See below

### Steps to reproduce

I unfortunately couldn't manage to reproduce it as this seems to only happen if there is _lots_ of data so please bear with me. I seem to have found out what _might_ cause performance issues with the new option of passing the `queryClient` to, for example, `useQuery`:

All arguments passed to `useQuery` (and other utility functions) are [wrapped in a computed](https://github.com/TanStack/query/blob/8b51e8b690f68190c06c865a84fb6a22c5c29d34/packages/vue-query/src/useBaseQuery.ts#L57), [cloned](https://github.com/TanStack/query/blob/8b51e8b690f68190c06c865a84fb6a22c5c29d34/packages/vue-query/src/useBaseQuery.ts#L172) and [refs unwrapped](https://github.com/TanStack/query/blob/8b51e8b690f68190c06c865a84fb6a22c5c29d34/packages/vue-query/src/utils.ts#L59). This **includes** the `queryClient` we pass as option as well. Now, the query client is rather large and, I'm not sure, may contain many reactive values / data-structures which are cloned _every time_ they change (because of the [computed](https://github.com/TanStack/query/blob/8b51e8b690f68190c06c865a84fb6a22c5c29d34/packages/vue-query/src/useBaseQuery.ts#L57)).

When we mark the `queryClient` as [raw](https://vuejs.org/api/reactivity-advanced.html#markraw) all problems are gone, the same if we do _not_ pass the `queryClient` directly as option to one of the query functions.

[...]

PR #5764, opened 2023-07-21
perf(vue-query): avoid unnecessary deep watch 

Mini-ghost:
According to the follow issue:
- #5031

I have spent some time identifying the root cause of this performance issue, and here are the reasons I found along with the proposed solution.

In `useBaseQuery`, `useIsFetching`, `useMutating`, `useMutation`, and `useQueries` of vue-query, deep watching the parsed `options/filters` is performed. When `queryClient` is included in these options, the watch will traverse each property of the object deeply, which is causing the performance overhead.

https://github.com/vuejs/core/blob/3be4e3cbe34b394096210897c1be8deeb6d748d8/packages/runtime-core/src/apiWatch.ts#L441-L443

Currently, there are two feasible solutions:

1. During the installation of the vue-query plugin, mark `queryClient` as raw. The Vue watch API will skip objects marked as raw during traversal.
2. Remove `queryClient` before performing the deep watch to avoid unnecessary tracking.

In this PR, I have chosen the second solution.

While the first solution is simple and effective, it may not directly help scenarios like #5031. Users adopting such an approach would still need to manually mark `queryClient` as raw:

```ts
const queryClient = markRaw(new QueryClient({ ... }));
const query = useQuery(..., ..., {
    queryClient,
});
```

I'm not sure which one is better. If you think the first approach is more concise and powerful, I can modify the PR to use the first approach.

Below are two solutions along with the performance comparison with the previous version.

|![current verson][current-version-src]|![first solution][first-solution-src]|![second solution][second-solution-src]|
| :--: | :--: | :--: |
| current verson | first solution | **second solution** |

- Browser: Chrome v115
- Version: @tanstack/vue-query v4.29.25, vue v3.3.4
- Test Code:

```ts

Array.from({ length: 2000 }, (_, index) => {
  useQuery({
    queryClient,
    queryKey: ['TODO', index],
    queryFn() {
      return new Promise(resolve => setTimeout(resolve, 1000, index));
    },
  });
});
```

---

## Update 2023/07/24

This is the third solution: using `computed()`. Computed properties recompute and produce completely new data after any of their collected dependencies are updated. In this case, there is no need for deep `watch()` to trigger updates.

Moreover, this approach helps reduce the performance overhead of `watch()` when traversing objects.

solution 3

In @tanstack/vue-query, deep cloning has already been used to allow computed properties to collect all reactive dependencies. Therefore, I believe this method might be better than the previous two approaches.

https://github.com/TanStack/query/blob/118801df342881021c2bb6c8add5bb11b6dd944a/packages/vue-query/src/utils.ts#L30-L62

I'm not sure if the team would prefer to adopt a specific solution. If there are any suggestions or other approaches, I am more than willing to assist. Thank you for your excellent work.

[current-version-src]: https://github.com/TanStack/query/assets/39984251/3207446e-2975-4f8b-a756-40427446dea7
[first-solution-src]: https://github.com/TanStack/query/assets/39984251/eaa9f2c0-0b72-44b5-9913-75584331333a
[second-solution-src]: https://github.com/TanStack/query/assets/39984251/00e3552c-0610-4e51-9113-5a90c3ee4ca8

### DIFF
--- a/packages/vue-query/src/useBaseQuery.ts
+++ b/packages/vue-query/src/useBaseQuery.ts
@@ -65,6 +65,8 @@ export function useBaseQuery<
       : 'optimistic'
 
     if ('queryClient' in defaulted) {
+      // @ts-ignore
+      // The `queryClient` property does exist; it lacks type definition.
       delete defaulted.queryClient
     }
 
--- a/packages/vue-query/src/useBaseQuery.ts
+++ b/packages/vue-query/src/useBaseQuery.ts
@@ -64,6 +64,10 @@ export function useBaseQuery<
       ? 'isRestoring'
       : 'optimistic'
 
+    if ('queryClient' in defaulted) {
+      delete defaulted.queryClient
+    }
+
     return defaulted
   })
 
--- a/packages/vue-query/src/useBaseQuery.ts
+++ b/packages/vue-query/src/useBaseQuery.ts
@@ -64,12 +64,6 @@ export function useBaseQuery<
       ? 'isRestoring'
       : 'optimistic'
 
-    if ('queryClient' in defaulted) {
-      // @ts-ignore
-      // The `queryClient` property does exist; it lacks type definition.
-      delete defaulted.queryClient
-    }
-
     return defaulted
   })
 
@@ -94,14 +88,10 @@ export function useBaseQuery<
     { immediate: true },
   )
 
-  watch(
-    defaultedOptions,
-    () => {
-      observer.setOptions(defaultedOptions.value)
-      updateState(state, observer.getCurrentResult())
-    },
-    { deep: true },
-  )
+  watch(defaultedOptions, () => {
+    observer.setOptions(defaultedOptions.value)
+    updateState(state, observer.getCurrentResult())
+  })
 
   onScopeDispose(() => {
     unsubscribe()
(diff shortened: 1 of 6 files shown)


Mini-ghost:
This is the third solution: using `computed()`. Computed properties recompute and produce completely new data after any of their collected dependencies are updated. In this case, there is no need for deep `watch()` to trigger updates.

Moreover, this approach helps reduce the performance overhead of `watch()` when traversing objects.

solution 3

在 @tanstack/vue-query 中已經使用深拷貝讓計算屬性可以收集到所有響應依賴，因此我認為這個方法可能會比上面兩個做法都要更好。

https://github.com/TanStack/query/blob/118801df342881021c2bb6c8add5bb11b6dd944a/packages/vue-query/src/utils.ts#L30-L62

由於已經開啟這個 PR 所以如果有希望採取任何做法，我都很願意協助，感謝您出色的工作。


上面的內容翻譯成英文

Mini-ghost:
This is the third solution: using `computed()`. Computed properties recompute and produce completely new data after any of their collected dependencies are updated. In this case, there is no need for deep `watch()` to trigger updates.

Moreover, this approach helps reduce the performance overhead of `watch()` when traversing objects.

solution 3

In @tanstack/vue-query, deep cloning has already been used to allow computed properties to collect all reactive dependencies. Therefore, I believe this method might be better than the previous two approaches.

https://github.com/TanStack/query/blob/118801df342881021c2bb6c8add5bb11b6dd944a/packages/vue-query/src/utils.ts#L30-L62

Since I have already opened this PR, I am willing to assist with any chosen approach. Thank you for your excellent work.

PR #5766, opened 2023-07-21
fix(react-query): types for useSuspenseInfiniteQuery

manudeli:
related with https://github.com/TanStack/query/pull/5739

# Add types for useSuspenseInfiniteQuery like useSuspenseQuery
1. UseSuspenseInfiniteQueryOptions
2. UseSuspenseInfiniteQueryResult

### DIFF
--- a/packages/react-query/src/types.ts
+++ b/packages/react-query/src/types.ts
@@ -6,6 +6,7 @@ import type {
   DefinedQueryObserverResult,
   InfiniteQueryObserverOptions,
   InfiniteQueryObserverResult,
+  InfiniteQueryObserverSuccessResult,
   MutateFunction,
   MutationObserverOptions,
   MutationObserverResult,
@@ -66,6 +67,25 @@ export interface UseInfiniteQueryOptions<
     'queryKey'
   > {}
 
+export interface UseSuspenseInfiniteQueryOptions<
+  TQueryFnData = unknown,
+  TError = DefaultError,
+  TData = TQueryFnData,
+  TQueryData = TQueryFnData,
+  TQueryKey extends QueryKey = QueryKey,
+  TPageParam = unknown,
+> extends Omit<
+    UseInfiniteQueryOptions<
+      TQueryFnData,
+      TError,
+      TData,
+      TQueryData,
+      TQueryKey,
+      TPageParam
+    >,
+    'enabled' | 'suspense' | 'throwOnError' | 'placeholderData'
+  > {}
+
 export type UseBaseQueryResult<
   TData = unknown,
   TError = DefaultError,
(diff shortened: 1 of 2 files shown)


packages/react-query/src/types.ts
```diff
@@ -66,6 +67,25 @@ export interface UseInfiniteQueryOptions<
...
+    UseInfiniteQueryOptions<
+      TQueryFnData,
+      TError,
+      TData,
+      TQueryData,
+      TQueryKey,
+      TPageParam
+    >,
+    'enabled' | 'suspense' | 'throwOnError' | 'placeholderData'
+  > {}
```
manudeli, line 87: I added UseSuspenseInfiniteQueryOptions, cause of same reason with [why we need UseSuspenseQueryOptions](https://github.com/TanStack/query/pull/5755#discussion_r1269831942)

packages/react-query/src/types.ts
```diff
@@ -96,6 +116,11 @@ export type DefinedUseInfiniteQueryResult<
   TError = DefaultError,
 > = DefinedInfiniteQueryObserverResult<TData, TError>
 
+export type UseSuspenseInfiniteQueryResult<
+  TData = unknown,
+  TError = DefaultError,
+> = Omit<InfiniteQueryObserverSuccessResult<TData, TError>, 'isPlaceholderData'>
```
manudeli, line 122: I thought If we don't have `enabled` option and have default `suspense` option, we don't have to use `DefinedUseInfiniteQueryResult` containing unnecessary `InfiniteQueryObserverRefetchErrorResult`.

```ts
type DefinedInfiniteQueryObserverResult<TData = unknown, TError = DefaultError> = InfiniteQueryObserverRefetchErrorResult<TData, TError> | InfiniteQueryObserverSuccessResult<TData, TError>;
```

We just need only InfiniteQueryObserverSuccessResult for useSuspenseInfiniteQuery.
So we could remove `isSuccess: false`, `error`, `isError: true` etc. to do  type-narrow

```ts
interface InfiniteQueryObserverRefetchErrorResult<TData = unknown, TError = DefaultError> extends InfiniteQueryObserverBaseResult<TData, TError> {
    data: TData;
    error: TError;
    isError: true;
    isPending: false;
    isLoadingError: false;
    isRefetchError: true;
    isSuccess: false;
    status: 'error';
}
interface InfiniteQueryObserverSuccessResult<TData = unknown, TError = DefaultError> extends InfiniteQueryObserverBaseResult<TData, TError> {
    data: TData;
    error: null;
    isError: false;
    isPending: false;
    isLoadingError: false;
    isRefetchError: false;
    isSuccess: true;
    status: 'success';
}
```

This change is needed by same reason with https://github.com/TanStack/query/pull/5755#discussion_r1269836999

manudeli (commented):
I commented as my intention

packages/react-query/src/useSuspenseInfiniteQuery.ts
```diff
@@ -41,5 +43,5 @@ export function useSuspenseInfiniteQuery<
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
     InfiniteQueryObserver as typeof QueryObserver,
     queryClient,
-  ) as DefinedUseInfiniteQueryResult<TData, TError>
+  ) as InfiniteQueryObserverSuccessResult<TData, TError>
```
manudeli, line 46: I think alternative solution is below suggestion

```suggestion
  ) as unknown as UseSuspenseInfiniteQueryResult<TData, TError>
```

PR #5785, opened 2023-07-25
Introduce minor performance gain using `next/script` component 🚄

sanjaiyan-dev:
This pull request aims to improve performance by replacing the traditional script tag with the `next/script` component and setting the strategy to beforeInteractive. This change is expected to yield a minor performance boost in the package.

Note: Apologies if there are any mistakes in the code or the pull request. Please review and provide feedback.

### DIFF
--- a/packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
+++ b/packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
@@ -1,6 +1,7 @@
 'use client'
 
 import { useServerInsertedHTML } from 'next/navigation'
+import Script from 'next/script'
 import * as React from 'react'
 
 const serializedSymbol = Symbol('serialized')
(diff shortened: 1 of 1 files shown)


TkDodo:
@juliusmarminge can you have a look here please?

TkDodo:
Thanks for the PR. We've discussed and decided to not merge this for 2 reasons:

1) we'd like to keep dependencies to `next` to a minimum. If react ships a hook that lets us inject something into the stream, we can make this package a non-nextjs specific solution that other SSR providers can also profit from (we'd obviously rename it then).

2) `beforeInteractive` should be used for critical things, and I don't think that suspense boundaries would qualify for that.

sanjaiyan-dev:
> Thanks for the PR. We've discussed and decided to not merge this for 2 reasons:
> 
> 1) we'd like to keep dependencies to `next` to a minimum. If react ships a hook that lets us inject something into the stream, we can make this package a non-nextjs specific solution that other SSR providers can also profit from (we'd obviously rename it then).
> 
> 2) `beforeInteractive` should be used for critical things, and I don't think that suspense boundaries would qualify for that.

Ok :)

PR #5786, opened 2023-07-25
react ssr streaming example

marbemac:
Inspired by @tannerlinsley's work with the tanstack router streaming example, and the recent nextjs streaming package, I thought I'd put together a streaming example that doesn't require nextjs.

Marking as a draft because I've mostly done this as an exercise to see how easy/hard this is to implement with the current RQ API. In part I'm hoping that this might provide a basis for ideas re how to make this simpler - perhaps via some additions to react-query itself?

I have also implemented `deferQuery` - similar to solidjs's deferQuery option - when set, that query will pause the stream. The use case here is critical queries that, for example, fetch data required for meta tags.

### DIFF
--- a/examples/react/ssr-streaming/src/entry-client.tsx
+++ b/examples/react/ssr-streaming/src/entry-client.tsx
@@ -1,0 +1,17 @@
+import ReactDOM from 'react-dom/client'
+import { QueryClientProvider } from '@tanstack/react-query'
+
+import { Root } from './Root'
+import { createQueryClient } from './query-client'
+
+import { hydrateStreamingData } from './hydrate-streaming-data'
+
+const queryClient = createQueryClient()
+hydrateStreamingData({ queryClient })
+
+ReactDOM.hydrateRoot(
+  document,
+  <QueryClientProvider client={queryClient}>
+    <Root />
+  </QueryClientProvider>,
+)
--- a/examples/react/ssr-streaming/src/query-client.ts
+++ b/examples/react/ssr-streaming/src/query-client.ts
@@ -1,0 +1,63 @@
+import { QueryCache, QueryClient } from '@tanstack/react-query'
+
+type CreateQueryClientOpts = {
+  trackedQueries?: Set<string>
+  blockingQueries?: Map<string, Promise<void>>
+}
+
+export const createQueryClient = ({
+  trackedQueries,
+  blockingQueries,
+}: CreateQueryClientOpts = {}) => {
+  const blockingQueryResolvers = new Map<string, () => void>()
+
+  const queryCache = blockingQueries
+    ? new QueryCache({
+        onSettled(data, error, query) {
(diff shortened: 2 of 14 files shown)


examples/react/ssr-streaming/src/entry-client.tsx
```diff
@@ -0,0 +1,17 @@
+import ReactDOM from 'react-dom/client'
+import { QueryClientProvider } from '@tanstack/react-query'
+
+import { Root } from './Root'
+import { createQueryClient } from './query-client'
+
+import { hydrateStreamingData } from './hydrate-streaming-data'
+
+const queryClient = createQueryClient()
+hydrateStreamingData({ queryClient })
```
marbemac, line 10: This must be called once on the client.

examples/react/ssr-streaming/src/query-client.ts
```diff
@@ -0,0 +1,63 @@
...
+  if (trackedQueries && import.meta.env.SSR) {
+    // Do we need to care about unsubscribing? I don't think so to be honest
+    queryClient.getQueryCache().subscribe((event) => {
+      // console.log('META', event.query.meta);
+      const defer = event.query.meta?.['deferStream']
+
+      switch (event.type) {
+        case 'added':
+        case 'updated':
+          trackedQueries.add(event.query.queryHash)
```
marbemac, line 48: Similar track queries approach as the next streaming package.

examples/react/ssr-streaming/src/query-client.ts
```diff
@@ -0,0 +1,63 @@
...
+
+          if (event.type === 'added' && defer) {
+            blockingQueries?.set(
+              event.query.queryHash,
+              new Promise((r) =>
+                blockingQueryResolvers.set(event.query.queryHash, r),
+              ),
+            )
+          }
```
marbemac, line 57: Here is where that `meta.deferStream` option is used. If the query is marked as defer, then a promise is pushed into the blocking queries - that promise is awaited in the stream, and will resolve via the `onSettled` query cache hook higher up in this file.

examples/react/ssr-streaming/src/server.ts
```diff
@@ -0,0 +1,106 @@
...
+      const appStream = await renderToReadableStream(reactApp, {
+        bootstrapModules: ['/@vite/client', '/src/entry-client.tsx'],
+        bootstrapScriptContent: [addFastRefreshPreamble()].join('\n'),
+      })
+
+      const injector = createDataInjector({
+        blockingQueries,
+        queryClient,
+        trackedQueries,
+      })
```
marbemac, line 64: The server must pipe through this `TransformStream`.

TkDodo:
Hi, Thanks for this example, but I'm unsure how to proceed with this. I think as an example visible to users, it would be quite confusing

PR #5794, opened 2023-07-27
 feat(react-query): v5: Export `DefinedInitialDataOptions` and `UndefinedInitialDataOptions`

david-crespo:
I'm upgrading to v5 in my app. Love it. The only issue I've run into is that I have to use the unexported type `UndefinedInitialDataOptions` to make this helper for the options. I have some wrappers for `useQuery` that constrain the args to the methods and params of my API, and I want them to take all the options for `useQuery` except `queryKey` and `queryFn`.

```ts
/**
 * `queryKey` and `queryFn` are always constructed by our helper hooks, so we
 * only allow the rest of the options.
 */
type UseQueryOtherOptions<T, E = DefaultError> = Omit<
  UndefinedInitialDataOptions<T, E>,
  'queryKey' | 'queryFn'
>
```

### DIFF
--- a/packages/react-query/src/index.ts
+++ b/packages/react-query/src/index.ts
@@ -11,6 +11,10 @@ export { useQuery } from './useQuery'
 export { useSuspenseQuery } from './useSuspenseQuery'
 export { useSuspenseInfiniteQuery } from './useSuspenseInfiniteQuery'
 export { queryOptions } from './queryOptions'
+export type {
+  DefinedInitialDataOptions,
+  UndefinedInitialDataOptions,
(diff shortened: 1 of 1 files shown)


TkDodo:
thanks. there's a slight issue around these types, and we'd need the same for the `InfiniteQuery` helpers. I'll make a separate PR and I'll include the export once I'm sure things work as they should.

TkDodo:
decided not to rename, but your changes are included here:

- https://github.com/TanStack/query/pull/5803

david-crespo:
Thank you!

PR #5804, opened 2023-07-29
refactor(query-core): Make MutationMeta and QueryMeta generic

vojvodics:
Reopening https://github.com/TanStack/query/pull/5457 to the beta branch


TkDodo:
Hm, I would still advise to _not_ put functions into `meta`. with v5, `meta` will be serialized across boundaries with our hydration utils when:
- using a persister
- using SSR

it would also be necessary that these functions are pure and do not reference anything from their closures...

vojvodics:
Thanks for the explanation! I agree that it's easy to make a mistake if you're not careful and don't understand all the problems that come with it. 
Using generics in meta doesn't restrict you to only functions, but can still provide type safety (e.g. using fields of the property as keys in meta). 

Feel free to close the PR if you think this is not something you should support.

Issue #5436, opened 2023-05-19
[v5 beta] Type inference for `useQuery` breaks when using `queryOptions` util and `select`


jakst:
### Describe the bug

When using the new `queryOptions` util from react-query v5, type inference seems to be a bit fragile. For example when spreading the resulting options object and adding a select statement to the query, the resulting data type is incorrect. More interestingly, the type seems to be inferred correctly if we also set the type `refetchInterval` option. I haven't tested around exactly which other options trigger this behaviour, but I believe this data point should be helpful enough in fixing the issue.

```typescript
import { queryOptions } from '@tanstack/react-query'

const options = queryOptions({
  queryKey: ["key"],
  queryFn: () => Promise.resolve(1),
});

// Type inference breaks, so TS complains about return type of select()
const query1 = useQuery({
  ...options,
  select: (data) => data > 1
});

// Type inference works like it should, so query2.data has type boolean
const query2 = useQuery({
  ...options,
  select: (data) => data > 1,
  refetchInterval: 1,
});

// When inlining the query options, types are correct even without refetchInterval
const query3 = useQuery({
  queryKey: ["key"],
  queryFn: () => Promise.resolve(1),
  select: (data) => data > 1,
});
```

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/festive-platform-0krnf3?file=%2Fsrc%2FApp.tsx%3A21%2C28

### Steps to reproduce

* Open the codesandbox
* See the ts errors in the editor

### Expected behavior

The correct data type should be inferred from the return type of the select function in useQuery.

### How often does this bug happen?

Every time

### Platform

v5.0.0-alpha.34

### Tanstack Query adapter

react-query

### TanStack Query version

v5.0.0-alpha.34

### TypeScript version

v5.0.4

TkDodo:
yeah that's interesting. I'd expect `select` to _not_ work in these cases, because the return type of `queryOptions` will be of type `<number, Error, number, QueryKey>`. So `select` wants to "change" the 3rd generic, but it can't because it has already been inferred.

I'm pretty puzzled why adding `refetchInterval: 1` changes something. I mean, `refetchInterval` can be a function as well, and it uses the 3rd generic (`TData`), too, but I'm unsure how this can change the typings on `options`...

maybe @DamianOsipiuk or @ecyrbe have any ideas?

gloomweaver:
[...]

PR #5805, opened 2023-07-29
fix(types): infer the narrowest type with `queryOptions`

Andarist:
fixes https://github.com/TanStack/query/issues/5436

### DIFF
--- a/packages/react-query/src/__tests__/useQuery.types.test.tsx
+++ b/packages/react-query/src/__tests__/useQuery.types.test.tsx
@@ -45,6 +45,19 @@ describe('initialData', () => {
       })
     })
 
+    it('it should be possible to define a different TData than TQueryFnData using select with queryOptions spread into useQuery', () => {
+      doNotExecute(() => {
+        const options = queryOptions({
+          queryKey: ['key'],
+          queryFn: () => Promise.resolve(1),
+        })
+        useQuery({
+          ...options,
+          select: (data) => data > 1,
+        })
+      })
+    })
+
     it('TData should always be defined when initialData is provided as a function which ALWAYS returns the data', () => {
       doNotExecute(() => {
         const { data } = useQuery({
--- a/packages/react-query/src/queryOptions.ts
+++ b/packages/react-query/src/queryOptions.ts
@@ -24,18 +24,26 @@ export function queryOptions<
   TError = DefaultError,
   TData = TQueryFnData,
   TQueryKey extends QueryKey = QueryKey,
->(
-  options: UndefinedInitialDataOptions<TQueryFnData, TError, TData, TQueryKey>,
-): UndefinedInitialDataOptions<TQueryFnData, TError, TData, TQueryKey>
+  TOptions extends UndefinedInitialDataOptions<
+    TQueryFnData,
+    TError,
+    TData,
+    TQueryKey
+  > = UndefinedInitialDataOptions<TQueryFnData, TError, TData, TQueryKey>,
+>(options: TOptions): TOptions
 
 export function queryOptions<
   TQueryFnData = unknown,
   TError = DefaultError,
   TData = TQueryFnData,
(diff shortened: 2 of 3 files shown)


TkDodo (approved):
❤️🙏

TkDodo:
@Andarist it seems that with this change, we lost excess property checks on the `queryOptions` function. For example, if I misspell a property (e.g. `stallTime` instead of `staleTime`), I'm not getting any errors:

[TypeScript playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgRwK4FMoE8DyYbAQB2AznAL5wBmUEIcA5AAIwCGpbAxgNYD0U6VpxgBaNJiwM4vXnBhYw6EgC44AI3RsAUKEiw4AKjisy47AGFo6arXrM2HIXzNYRnK1JlyFS1Ru1aLngExCQAFAhacCgY2ADS6FiqANoMRKggGlAMALoANFExEgBiRKphAJRwALwAfHAACrbAJOgAdAIkEAA2AG7oYQCsFQXRXiwkIugAHorCU1C0UHAkbN3dACqg1gAmEEpEDPAzLfDKvIWrrOtbIOiqg4ZwAIwADO8F5BVaQA)

We don't have a test for this, but I tried it out with the version from before that PR, and there, it used to work 🤔

Andarist:
This gets tricky - the inferred type is **always** exempt from the excess property check. We need to remember that the excess property check is more like a lint rule than anything else. 

If you want to eat cake and have it too then perhaps this self-validation technique is your best bet: [TS playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgPIAczAPYgM7IDeAUMsgI4Cu0AngNIQ0D8AXMpSANYjYDuIAblIVqUGgDEQrdlx78hZPGDgAbCABVgAWwjSQlLQCNoQgL7FiYGuhQA1VcAAmcSAB51APmQBeIsIDadMigyJyM2DDI6gC6bEEQAB6QII4EYTQRaJg4+MhMUYHRyGwgEABuJsSmQsSOEAgqcFAoMBwIWLgitBgd+O7IicmpWb14HgAUCGz2Kk4uEO4eAJRs6jVU3dm4eOMkZBtiDDRs-gBE+kbQp9EANMIHEiBs40s+XgAKUNhawHgQAHTNPDYFQVcYAViWd0UyhUKk0OjYAEYAAwou6mJZCIA). 

Other than that... a bizarre thing like this could work: [TS playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBGIEkwDMCmAnVZipgXhgGUBPAWwCMQAbACgEoBuAKCmIAdcA5RFDLHfDFYcQyODzSZsqZi3a4ASqlIgAbqm5gkkzABMAPABUAfIIDeTGDADaAaRgBLMDADWqYqJiGYAQwhe7AF0YVAAPKCxdf25tPmkYAH44VHV0GAAuGFtAzMMg5gBfWV1UYCofTBhkAFdsKAdwGABHaoxiAHk2evAIfUsYAEVW9GJbd0EY3ikcABp+obaAMWcCScl+VDmrQigfKlRDB1JcVYk42aZjGmBMiysWtrHiBMyFkafme+HiZZfB7+Wnxg0D2ByOqD+O1Bh2OhTot2a3yeryR7gYiKWYBRmPRIP2MNQmSh+PBMCKTCYDxGnW6kBodwx73cmWsACIwNUKBhWYEtoyfliYPR8KYAAroECkBwQVAAOkwEGo6hoAFY6Hy8WDjpkAIwABj1cwKjCYQA). This infers each property separately~ to ensure that they were present in the input. Thanks to that we can keep the excess property check "as is" because the target object that is meant to report it is not the target of inference on its own. More work/thought would have to be put into this one to handle the optionality of input properties correctly.

TkDodo:
uuh, let me try out the `validate` approach, that looks great. I'll also add some tests for this. Thank you 🙏

PR #5811, opened 2023-07-31
type pageParams and allow null cursors

juliusmarminge:
- allows `null` to be returned from `getNextPageParam` and used to determine ifthere is a next pag ( same for previous )
- allows typing the returned `pageParams`. the `useInfiniteQuery` currently still returns this as unknown due to order of generic. Leaving this as a follow-up as discussed with TKDodo.

### DIFF
--- a/docs/react/guides/migrating-to-v5.md
+++ b/docs/react/guides/migrating-to-v5.md
@@ -337,6 +337,23 @@ useInfiniteQuery({
 })
 ```
 
+### The returned type from `useInfiniteQuery`'s `pageParams` is now typed
+
+Previously, the type of `useInfiniteQuery.data.pageParams` was `unknown[]`. In v5, this is now typed to match your query:
+
+```diff
+const query = useInfiniteQuery({
+  queryKey,
+  queryFn: ({ pageParam }) => fetchSomething(pageParam),
+  defaultPageParam: 0,
+})
+if (query.isSuccess) {
+  const pageParams = query.data.pageParams;
+- //    ^? unknown[]
++ //    ^? number[] 
+}
+```
+
 ### Manual mode for infinite queries has been removed
 
 Previously, we've allowed to overwrite the `pageParams` that would be returned from `getNextPageParam` or `getPreviousPageParam` by passing a `pageParam` value directly to `fetchNextPage` or `fetchPreviousPage`. This feature didn't work at all with refetches and wasn't widely known or used. This also means that `getNextPageParam` is now required for infinite queries.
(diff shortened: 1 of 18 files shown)


docs/react/reference/QueryClient.md
```diff
@@ -23,6 +23,7 @@ await queryClient.prefetchQuery({ queryKey: ['posts'], queryFn: fetchPosts })
 
 Its available methods are:
 
+- [`QueryClient`](#queryclient)
```
juliusmarminge, line 26: i didn't add this lol

docs/react/guides/migrating-to-v5.md
```diff
@@ -341,6 +341,10 @@ useInfiniteQuery({
 
 Previously, we've allowed to overwrite the `pageParams` that would be returned from `getNextPageParam` or `getPreviousPageParam` by passing a `pageParam` value directly to `fetchNextPage` or `fetchPreviousPage`. This feature didn't work at all with refetches and wasn't widely known or used. This also means that `getNextPageParam` is now required for infinite queries.
 
+### Returning `null` from `getNextPageParam` or `getPreviousPageParam` now indicates that there is no further page available
+
+In v4, you needed to explicitly return `undefined` to indicate that there is no further page available. We've widened this check to include `null`.
```
juliusmarminge, line 346: technically not though ;)

https://github.com/TanStack/query/blob/100ae76c402f862f4702eac30369340afd5b57c1/packages/query-core/src/infiniteQueryBehavior.ts#L194

TkDodo: yeah but only for `hasNextPage` and `hasPreviousPage` checks. The actual fetching happened here:

https://github.com/TanStack/query/blob/100ae76c402f862f4702eac30369340afd5b57c1/packages/query-core/src/infiniteQueryBehavior.ts#L76

and there, we only checked for `undefined` :) It was pretty inconsistent and buggy. If you returned `null`, it would:
- tell you that there is no next page
- but it would still continue to fetch if you called `fetchNextPage` 🤷 

btw, I adapted that code path, too to include null now.

juliusmarminge: oh damn 😅

packages/react-query/src/__tests__/useInfiniteQuery.type.test.tsx
```diff
@@ -206,6 +216,13 @@ describe('getNextPageParam / getPreviousPageParam', () => {
       const result: Expect<
         Equal<InfiniteData<string> | undefined, (typeof infiniteQuery)['data']>
       > = true
+      // TODO: Order of generics prevents this from working
+      // const result: Expect<
+      //   Equal<
+      //     InfiniteData<string, number> | undefined,
```
TkDodo, line 222: can we keep these commented in with an assertion towards:

```
InfiniteData<string, unknown>
```

?

juliusmarminge: https://github.com/TanStack/query/pull/5811/commits/d881d1397ac3bd7a83d7a129555501af3af60441

TkDodo (approved):

PR #5812, opened 2023-07-31
Enhancing Dehydration and Hydration Mechanisms for Query Options

pmzi:
I'm proposing a fix concerning the dehydration process of queries that currently does not consider individual query options. This omission results in a problematic behavior, particularly when a query with unique options like `cacheTime` or others is dehydrated and subsequently rehydrated. The query loses its initial options settings, reverting to the defaults defined in `defaultOptions`.

Here's an illustrative example:

```ts
useQuery({
    queryKey: [queryKeys.cities, request],
    queryFn: async () => await citiesApi.getCities(),
    cacheTime: ONE_HOUR * 2,
    staleTime: ONE_HOUR,
  });
```

In this scenario, if the query undergoes dehydration and hydration, both `cacheTime` and `staleTime` values are mistakenly reset to their `defaultOptions`. This becomes evident when incorporating `PersistQueryClientProvider` into an application. Here, `cacheTime` and `staleTime` will unexpectedly reset to default values following a page refresh.

Additionally, we've defined a custom option `isPersistedCache` to persist specific queries:

```ts
useQuery({
    queryKey: [queryKeys.cities, request],
    queryFn: async () => await citiesApi.getCities(),
    cacheTime: ONE_HOUR * 2,
    staleTime: ONE_HOUR,
    isPersistedCache: true,
  });
```

Subsequently, when setting up the `PersistQueryClientProvider`:

```tsx
<PersistQueryClientProvider
      client={queryClient}
      persistOptions={{
        persister,
        dehydrateOptions: {
          dehydrateQueries: true,
          shouldDehydrateQuery: query =>
            Boolean((query.options as QueryOptions).isPersistedCache),
        },
      }}
    >
....

</PersistQueryClientProvider>
```

We also extended the `QueryOptions` interface accordingly:

```ts
declare module "@tanstack/react-query" {
  export interface QueryOptions {
    isPersistedCache?: boolean;
  }
}
```

However, due to the existing limitation that the dehydration and hydration processes do not account for custom options, the proposed solution is not functional under the current system.

### DIFF
--- a/packages/query-core/src/hydration.ts
+++ b/packages/query-core/src/hydration.ts
@@ -33,6 +33,7 @@ interface DehydratedQuery {
   queryHash: string
   queryKey: QueryKey
   state: QueryState
+  options: QueryOptions
 }
 
 export interface DehydratedState {
@@ -60,6 +61,7 @@ function dehydrateMutation(mutation: Mutation): DehydratedMutation {
 function dehydrateQuery(query: Query): DehydratedQuery {
   return {
     state: query.state,
+    options: JSON.parse(JSON.stringify(query.options)), // To remove properties that are functions or other non-serializable types
     queryKey: query.queryKey,
     queryHash: query.queryHash,
   }
@@ -162,6 +164,7 @@ export function hydrate(
       client,
       {
         ...options?.defaultOptions?.queries,
+        ...dehydratedQuery.options,
         queryKey: dehydratedQuery.queryKey,
         queryHash: dehydratedQuery.queryHash,
       },
--- a/packages/query-core/src/tests/hydration.test.tsx
+++ b/packages/query-core/src/tests/hydration.test.tsx
@@ -106,7 +106,7 @@ describe('dehydration and rehydration', () => {
     hydrationClient.clear()
   })
 
-  test('should be able to provide default options for the hydrated queries', async () => {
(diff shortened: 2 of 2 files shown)


TkDodo:
sorry, we can't dehydrate options because options are non-serializable (they contain functions).

regarding your example:

`staleTime` is an observer level property. it doesn't default to anything until you actually add an observer (via `useQuery`). So when the query is rendered, it will have a correct staleTime.

`cacheTime` can be an issue because the default would mean that queries are removed from the cache 5 minutes after you restored them from localstorage, unless you are using them (mounted via `useQuery`).

That's why the docs mention that `cacheTime` should be higher than `maxAge` for those scenarios:

https://tanstack.com/query/v4/docs/react/plugins/persistQueryClient#how-it-works

the idea is to set a global cacheTime when using a persister to avoid this.

---

regarding `isPersistedCache`: In v5, we'll persist `meta` as well for this reason. `isPersistedCache` should be passed in `meta` to `useQuery`.

pmzi:
I still don't understand why we are not dehydrating and hydrating options. We can still remove the non-serializable values (as I did). This way, the `isPersistedCache` works as expected too...

pmzi:
Also, dehydrating and hydrating will solve the issue with setting a default cacheTime (which is not ideal at all and will be applied to all of the queries).

PR #5813, opened 2023-07-31
chore(.eslintignore): add eslint ignoring rule

manudeli:
I added eslint ignoring rule like [nextjs](https://github.com/vercel/next.js/blob/canary/.eslintignore)

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,2 @@
+node_modules
+build
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,1 @@
-node_modules
-build
+*/**/build
--- a/.gitignore
+++ b/.gitignore
@@ -8,11 +8,10 @@ yarn.lock
 
 # builds
 types
-build
-*/build
-dist
-lib
-es
+*/**/build
+*/**/dist
+*/**/lib
+*/**/es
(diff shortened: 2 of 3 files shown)


.eslintignore
```diff
@@ -0,0 +1,2 @@
+node_modules
+build
```
TkDodo, line 2: we don't have a top level build directory, and `node_modules` is always excluded. I don't think this does anything

manudeli: Oh, I didn't think about top level. I excluded top level `build` directory in `.eslintignore`. and remove `node_modules`

.gitignore
```diff
@@ -39,9 +38,7 @@ stats.html
 
 *.log
 .DS_Store
-node_modules
```
TkDodo, line 42: can you please leave node_modules in gitignore

manudeli: Ah, I should have left a comment explaining the change. You can check duplicate node_modules in `.gitignore`. so I removed it. https://github.com/TanStack/query/pull/5813/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947L5

removing `dist` have same reason

TkDodo (approved):

PR #5815, opened 2023-07-31
chore(rollup.config.ts): use banner to add 'use client' directive

manudeli:
add [rollup banner](https://rollupjs.org/configuration-options/#output-banner-output-footer) to add ['use client' directive for nextjs](https://nextjs.org/docs/getting-started/react-essentials#library-authors)

### references
[react-wrap-balancer](https://github.com/shuding/react-wrap-balancer/blob/main/tsup.config.ts#L10-L13)

### DIFF
--- a/packages/react-query-devtools/src/Explorer.tsx
+++ b/packages/react-query-devtools/src/Explorer.tsx
@@ -1,4 +1,3 @@
-'use client'
 import * as React from 'react'
 
 import superjson from 'superjson'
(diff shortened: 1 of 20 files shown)


TkDodo:
thanks for the PR but we've moved away from rollup in v5 so I don't think this PR provides enough value and will cause conflicts on the beta branch.

manudeli:
> thanks for the PR but we've moved away from rollup in v5 so I don't think this PR provides enough value and will cause conflicts on the beta branch.

I'm making this tsup version too. Wait a moment

PR #5816, opened 2023-07-31
chore(tsup.config.js): use banner to add 'use client' directive for nextjs

manudeli:
related https://github.com/TanStack/query/pull/5815

add tsup banner to add ['use client' directive for nextjs](https://nextjs.org/docs/getting-started/react-essentials#library-authors) easily

### reference
[react-wrap-balancer](https://github.com/shuding/react-wrap-balancer/blob/main/tsup.config.ts#L10-L13)


TkDodo:
hm, what does this banner do exactly, and why is this better than adding `use client` manually where we want it?

FYI, we actually want to move away from having `use client` on all files. Some files should use `import client-only` instead with a separate server entry point in package.json

I tried this here:
- https://github.com/TanStack/query/pull/5633

Question is: how does the `banner` decide to which files a banner should go?

TkDodo:
@manudeli I created a new PR for the `import client-only` feature:

- #5817 

I think this would conflict with the `banner` ?

manudeli:
These days, I'm too busy with my company to see this Pull Request. But at the beginning of September, maybe I will be able to focus this issue again. sorry for the late reply

TkDodo:
gonna close this because I don't think we want use-client everywhere.

PR #5817, opened 2023-07-31
Feature/client only

TkDodo:
follow up to:
- https://github.com/TanStack/query/pull/5633


packages/react-query/src/shared.ts
TkDodo, line 1: @shuding you left a comment on the previous PR about something not working in the `react-server` condition here?

- https://github.com/TanStack/query/pull/5633#pullrequestreview-1502351656

can you elaborate what exactly you think won't work? was this about `type` exports ?

FYI, this PR just cherry-picks the commits from the previous PR because I fcked up the alpha -> beta merge in between 🙈

integrations/next-js-app/app/client-component.tsx
```diff
@@ -0,0 +1,23 @@
+'use client'
```
TkDodo, line 1: @shuding I've added an integration test to see if bundling with the poison pill works. What I'm seeing is that if I comment out the `'use client'`, it fails during build with:

```
TypeError: (0 , server_namespaceObject.useQuery) is not a function
           at ClientComponent (~/TanStackQuery/integrations/next-js-app/.next/server/app/page.js:2059:54)
```

this is not the error I would've expected 🤔

TkDodo: I've updated NextJs to the latest version, and the error is even more obscure now without `'use client'`:

```
- error ./app/client-component.tsx:1:0
Module not found: Can't resolve '@tanstack/react-query'
> 1 | import { useQuery } from '@tanstack/react-query'
```

packages/react-query/package.json
```diff
@@ -22,6 +22,7 @@
     ".": {
       "import": {
         "types": "./build/modern/index.d.ts",
+        "react-server": "./build/modern/server.js",
         "default": "./build/modern/index.js"
       },
       "require": {
```
phryneas, line 28: Do you know for sure that the server will always use ESM? We have seen some bundlers still use `require` here, so that might be the reason why it's not picking up the import.

TkDodo: does that mean we have to add the `react-server` entry to `require`, too ?

integrations/next-js-app/app/client-component.tsx
huozhi: There's a react-server condition `"react-server": "./build/modern/server.js",` added in `server.ts`, where it doesn't contain `useQuery` hook, then it led to the error `(server_namespaceObject.useQuery) is not a function`

packages/react-query/package.json
phryneas: That might be the case.

integrations/next-js-app/app/client-component.tsx
sebmarkbage: In other words, since you have a special implementation for the Server, that doesn't import any `client-only` things, it doesn't trigger that error because all the code you've exported for the server works. It's just that you don't have an useQuery export there so it's undefined which may error late for CJS path or early for ESM paths (missing export).

`client-only` is more about if you don't provide a custom RSC build at all.

In theory we could do some assumption here for ESM by checking the exports in the client ESM path by loading that to see if the name exists there to give a better error that this is the reason.

You could also export a getter on the CJS builds that throws a helpful message when accessed. For ESM you can export a fake function that throws but the downside of that is that it might not error early but when you actually try to run it which doesn't catch refactor errors without testing everything.

TkDodo: okay that makes sense @sebmarkbage, thanks. From what I've seen, I am getting the desired error messages if I just revert the "split" into `index.ts` and `server.ts`, and just always export everything also for `react-server`, because then, the files that are imported on the server that have `import 'client-only'` will error correctly:

At least that's what I'm getting with the approach:
Screenshot 2023-09-15 at 09 58 49

TkDodo: oh wait a second, that take us right back to where I can't import e.g. `import { queryOptions } from '@tanstack/react-query'` in a server component, even though that function is perfectly valid to be imported on the server, because this import brings in the whole bundle, and it errors with client-only then too ...

TkDodo: I'm at a loss right now. Basically, what I'd want is the following. For simplicitly, let's suppose we have two exports:

```
// index.ts

export { queryOptions } from './queryOptions'
export { useQuery } form './useQuery'
```

`queryOptions` can be used on the server and on the client, `useQuery` can only be on the client. How would I need to structure my exports so that I get:

- I can `import { queryOptions } from '@tanstack/react-query'` everywhere, without error
- If I `import { useQuery } from '@tanstack/react-query'` without being in the `use client` tree, I'd get the error that says:

> You're importing a component that imports client-only. It only works in a Client Component but none of its parents are marked with "use client", so they're Server Components by default.

Is that possible ?

huozhi: You can leave the entry as a shared component where it doesn't have any directives or restriction imports (no `server-only` or `client-only`), then only add client only inside `useQuery.ts` file. Then when it's get imported `queryOptions` is always shared component so it can work anywhere, `useQuery` is also shared but since it's marked as `"client-only"` it will error when used inside server components

TkDodo: > then only add client only inside useQuery.ts file

but that's what I have, I think. The entry bundle doesn't have any restrictions

https://github.com/TanStack/query/pull/5817/files#diff-ee5b50d37cfa7591a7855271d0b54f2cdf8fcdd03e84078a43d60099a4b4bda4

only useQuery has:

https://github.com/TanStack/query/pull/5817/files#diff-58bd4875e5c02519414a5c8bb09db72d1a7aa89b8ccec640147cbe91c622bffdR1

> Then when it's get imported queryOptions is always shared component so it can work anywhere

What I think I'm seeing is that if I import `queryOptions` from the entry point bundle, it will also try to import `useQuery`, at which point it fails.

huozhi: Feel like could because in server components everything is together, the "client-only" ones get bundled and get error earlier in build time. The most ideal would be separate them. Could you test if it will work when you put these exports separately?

TkDodo: yeah I'm pretty sure that's what's happening because when I use a full bundle as server entry, I get errors pointing to `QueryClientProvider` even though I only import `queryOptions`.

> The most ideal would be separate them. Could you test if it will work when you put these exports separately?

Not sure I understand what you mean exactly?

huozhi: Yea, `@tanstack/react-query` will be all bundled as server components, so any `"client-only"` will error in server components bundle during build time, much earlier than runtime. One way to solve it is to separate the shared components exports and client-only components exports, then group the client exports in one file. So you won't have chance to import anything from it for server components. Def something needs to be improved, so far separating can be a workaround

TkDodo: Yeah I don't think I want the workaround for now, gotta wait and see. I think I'll just stick to `'use client'` where necessary for now, I'm afraid the error of `useQuery` not found is something that will be reported waaaay too often as a bug on our end.

PR #5822, opened 2023-08-01
feat(suspense): return tuple for useSuspenseQuery and useSuspenseInfiniteQuery

TkDodo:


packages/react-query/src/useSuspenseInfiniteQuery.ts
```diff
@@ -44,4 +44,6 @@ export function useSuspenseInfiniteQuery<
     InfiniteQueryObserver as typeof QueryObserver,
     queryClient,
   ) as InfiniteQueryObserverSuccessResult<TData, TError>
+
+  return [query.data, query]
```
manudeli, line 48: I guess you want to make library user could rename guaranteed TData easily by use tuple returning like useState?
but I want to mention there is some trade offs.

1. making difference of interface with useQuery will raise barrier to entry a bit for useSuspenseQuery user.
2. it is expected to cause a BREAKING CHANGE if we add the enabled option so that TData cannot be guaranteed.

```suggestion
  return query
```

Could I know why you want to make this change more?

juliusmarminge: > it is expected to cause a BREAKING CHANGE if we add the enabled option in minor version so that TData cannot be guaranteed.

not breaking if introduced `enabled` is true by default (and thus data should always be defined)

juliusmarminge (approved):
🔥

packages/react-query/src/useSuspenseInfiniteQuery.ts
TkDodo: if we add enabled (which we likely won't), we would do so that not specifying enabled will still result in guaranteed TData, and only if you add it (and make it false), it will be undefined. We do the same with `initialData`.

This change is mainly about developer ergonomics. With a normal query, you usually want access to data, error, isLoading etc. because you need to handle those states. With `useSuspenseQuery`, you mainly want access to `data`, and not the rest of the query.

The entry barrier should be fine, as this is a new hook. Also, TypeScript helps :)

manudeli: > > it is expected to cause a BREAKING CHANGE if we add the enabled option in minor version so that TData cannot be guaranteed.
> 
> not breaking if introduced `enabled` is true by default (and thus data should always be defined)

first of all, I don't think `enabled` option has been determined to be always true yet. if then, I think we shouldn't make returning type as tuple with hugging chance to make BREAKING CHANGE If `enabled` can accept a boolean

Also we also have below way that `useQuery` users are familiar with already too. Isn't it better?

```ts
const {data: post, ...others} = useSuspenseQuery({
  queryKey: ['post', postId],
  queryFn: () => fetchPost(postId),
})
```

TkDodo: > to make BREAKING CHANGE If enabled can accept a boolean

not sure what enabled has to do here:

```
const [data] = useSuspenseQuery({
  queryKey: ['foo'],
  queryFn: () => Promise.resolve('foo')
})
```

`data` will be of type `string` here, while:

```
const [data] = useSuspenseQuery({
  queryKey: ['foo'],
  queryFn: () => Promise.resolve('foo'),
  enabled: someCondition
})
```

`data` will be of type `string | undefined`

---

> Also we also have below way that useQuery users are familiar with already too.

Implementation detail, but object spread has the disadvantage that it triggers getters on the result object, which interferes with tracked queries, so I usually advise against that: https://tkdodo.eu/blog/react-query-render-optimizations#tracked-queries

manudeli: Yeah, I worried about same point. if we make enabled option making `ReturnType<typeof useSuspenseQuery>[0]` have type `TData | undefined`

image

image

All this cases cannot have advantage of isSuccess flag too.

TkDodo: The `isSuccess` check is a good point. However, I really doubt that we will add `enabled` to `useSuspenseQuery`. `enabled` has two use-cases:

1) dependent queries. With `suspense`, queries are always run in serial, because react suspends on the first query, then runs and continues on the second query. so our [dependent queries example](https://tanstack.com/query/v4/docs/react/guides/dependent-queries) literally becomes:

```
// Get the user
const [user] = useSuspenseQuery({
  queryKey: ['user', email],
  queryFn: getUserByEmail,
})

const userId = user.id

// Then get the user's projects
const [projects] = useQuery({
  queryKey: ['projects', userId],
  queryFn: getProjectsByUser,
})
```

there is no enabled needed at all for this.

2) disabling a query from running until a condition is met, like user-input

suspense evolves a lot component composition, so the [lazy queries](https://tanstack.com/query/v4/docs/react/guides/disabling-queries#lazy-queries) example could become:

```
function Todos() {
  const [filter, setFilter] = React.useState('')

  return (
      <div>
        // 🚀 applying the filter will enable and execute the query
        <FiltersForm onApply={setFilter} />
        {filter && <TodosTable filter={filter} />
      </div>
  )
}

function TodosTable({ filter }) {
    const [data] = useSuspenseQuery({
      queryKey: ['todos', filter],
      queryFn: () => fetchTodos(filter),
  })
}
```

TkDodo:
I've decided to *not* move forward with this, because the equivalent changes to `useSuspenseQuery` would be too hard to achieve right now. It's better to have 3 suspense hooks that have the same interface.

manudeli:
> I've decided to _not_ move forward with this, because the equivalent changes to `useSuspenseQuery` would be too hard to achieve right now. It's better to have 3 suspense hooks that have the same interface.

Not now, but I think it was a good idea.  Thanks for reconsidering👍

PR #5833, opened 2023-08-03
fix: Change pinned peerDeps to a range

dbasilio:
Peer Dependencies should be ranges to allow consumers to change versions more freely.

I only found two instances of peerDependencies that references other workspaces. Deps and devDeps were left alone. Follow up to https://github.com/TanStack/query/discussions/5832


TkDodo:
yeah we have more on the `beta` branch. Thanks 👍

PR #5836, opened 2023-08-04
Feature/suspense

TkDodo:


TkDodo:
> Why remove suspense boolean from useQuery and useInfiniteQuery. Sometimes People wanna a conditional suspend query vie option enable or just do not wanna throw error but need a suspend query.

Because the goal is to offer a first class suspense integration, and we can't really do that with a flag. We might want to take advantage of the upcoming `use` hook internally, or we'll want to wrap updates to the Query Key in transitions (or use deferred values automatically).

Yes, in this first iteration, we are still passing the flag around internally, but it's now an an internal API that we can change in minors as well.

TkDodo:
it's also easier for TypeScript to set `data` to "always defined" with a separate hook, which is a pretty big ask from suspense users

liaoliao666:
I do not think so. Even if the `use` hook that can also be used conditionally or throw errors.

And If people wanna `data` to be defined, There is already a `useSuspenseQuery` hook to do it. No need to remove `suspense` from `useQuery`.

Issue #5719, opened 2023-07-15
Infinite render loop with v4.29.{22,23}


ddolcimascolo:
### Describe the bug

Renovate is trying to update react-query from v4.29.19 to latest (this currently pulls v4.29.23) in our application monorepo containing multiple large React  apps and a helper package. We have lots of tests entering infinite render loop, with the usual `Maximum call stack exceeded` error that React is throwing.

I can't give you more details yet, I'm opening the issue early to gather community feedback. I will be able to provide more information on Monday, but as I'm not expecting any breaking change in a patch release, I though it would be useful to keep you posted.

I'm not even sure this is a problem with react-query, it's just that this _happens_ while upgrading this library.

Cheers,
David 

### Your minimal, reproducible example

To be done

### Steps to reproduce

To be done

### Expected behavior

Tests pass as with v4.29.19

### How often does this bug happen?

Every time

### Platform

Linux
JSDOM through Jest tests

### Tanstack Query adapter

react-query

### TanStack Query version

v4.29.23 

### Additional context

The problem is in unit tests only.

iscekic:
I can confirm we're having a similar issue with next `13.4.10` and react-query `4.29.25` when using `useInfiniteQuery`.

Downgrading to `13.4.9` and `4.29.19` fixes the issue.

TkDodo:
@iscekic can you show a reproduction please?

TkDodo:
and, can you narrow down in which RQ version the issue first occurred ? We do one patch release per issue, so if you can tell me exactly which release between `4.29.19` and `4.29.25` is the first to break, I can look into it.

iscekic:
I don't have a minimal repro, but I did manage to fix the issue using the latest version.

Seems like the culprit was the `getNextPageParam` fn, which didn't have a stable reference - adding `useCallback` fixed it.

TkDodo:
react-query doesn't require `getNextPageParam` to have a stable reference though, so I'd still appreciate the reproduction

ddolcimascolo:
Hi guys, same on my side after spending the morning trying to do a minimal reproduction in a codesandbox. We did have a non-stable function passed as an event handler in one of our components. Wrapping it in a `useCallback` fixed it.

I didn't manage to reproduce in isolation though, so I'd accept that the issue was on my side, though it's still strange that a patch release did introduce the behavior...

ddolcimascolo:
[...]

PR #5839, opened 2023-08-04
Fix: Infinite render loops in react-query v4.29.22 and up

Komoszek:
Fixes https://github.com/TanStack/query/issues/5719. Shalow `!==` would always return `true`, because  https://github.com/TanStack/query/blob/9b8b5f0a81bc68552bbafbc3991e7e89a893ecbe/packages/query-core/src/queryObserver.ts#L243C19-L243C19 AFAIK always gives new object.

### DIFF
--- a/packages/query-core/src/queryObserver.ts
+++ b/packages/query-core/src/queryObserver.ts
@@ -834,7 +834,7 @@ function shouldAssignObserverCurrentProperties<
 
   // if the newly created result isn't what the observer is holding as current,
   // then we'll need to update the properties as well
-  if (observer.getCurrentResult() !== optimisticResult) {
+  if (!shallowEqualObjects(observer.getCurrentResult(), optimisticResult)) {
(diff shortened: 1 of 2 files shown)


TkDodo:
thanks. codesandbox doesn't work so I can't verify the fix. Can you add the failing reproduction as a test case please?

It was this one: https://codesandbox.io/s/confident-visvesvaraya-t8hzwk?file=/src/App.js

TkDodo:
@incepter fyi

TkDodo:
so this does indeed seem to fix the reported reproduction, here's a fork of the reproduction depending on the codesandbox preview build:

https://codesandbox.io/s/keen-fast-rxg5rk?file=/package.json

Komoszek:
@TkDodo I've added the test case like you've asked

TkDodo (approved):

PR #5842, opened 2023-08-05
perf(vue-query): avoid unnecessary deep watch

Mini-ghost:
The reason for this PR is the same as RP #5764, although the `beta` version does not have the `queryClient` performance issue, deep watch still makes the options traversed.

Also, I'm not sure why the watch source in `useQueries` is an array, and I didn't find multiple watch sources in it, so I removed the array.

### DIFF
--- a/packages/vue-query/src/useBaseQuery.ts
+++ b/packages/vue-query/src/useBaseQuery.ts
@@ -137,7 +137,7 @@ export function useBaseQuery<
 
         run()
 
-        stopWatch = watch(defaultedOptions, run, { deep: true })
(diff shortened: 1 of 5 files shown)


Mini-ghost:
This PR contains tweaks that were missing in PR #5764
https://github.com/TanStack/query/blob/46e7b0282077ccb2b11871dba6b6a19e8d8731b6/packages/vue-query/src/useBaseQuery.ts#L130

Mini-ghost:
@DamianOsipiuk Maybe this PR should be synced to the `main` branch?

PR #5844, opened 2023-08-05
Simplify load method, fix syntax highlighting

danawoodman:
SvelteKit now automatically infers the type of an exported load method, so no need to manually define it.

Also, Github has an ongoing issue with the Svelte syntax highlighting so using `html` is a good fallback.

[View changes here](https://github.com/danawoodman/query/blob/main-1/docs/svelte/ssr.md)


danawoodman:
Hello? 🤔👀

lachlancollins (changes requested):
Hi @danawoodman, thanks for the submission! I'm happy with the auto types for load functions, but I would like to keep the codeblocks as svelte. The docs are primarily meant to be read on tanstack.com, and the change from html to svelte was so we could add proper prism svelte highlighting.

Progress here has been stalled - you can see my PRs here:
- https://github.com/TanStack/tanstack.com/pull/88
- https://github.com/pngwn/prism-svelte/pull/20

I might just go ahead and copy the [prism-svelte](https://github.com/pngwn/prism-svelte) code into the [tanstack.com](https://github.com/TanStack/tanstack.com) repo so this can highlight correctly.

danawoodman:
@lachlancollins do you want me to undo the svelte=>html changes?

danawoodman:
@lachlancollins sorry for delay, I've reverted the code block changes

lachlancollins (approved):

PR #5846, opened 2023-08-06
fix(vue-query): ensure the suspense is resolved

### LINKED ISSUE #5845: `suspense()` won't resolve when initial `enabled` is set to `false`
### Describe the bug

When `enabled` is initially set to `false`, the `suspense()` function will never resolve. One scenario where this is relevant is when you want `useQuery` to execute only on the client site.

```ts
const { data,  } = useQuery({
  queryKey: ['TEST'],
  queryFn: async () => { 
    // .... 
  },

  //  Disabled when on the server site or due to other factors
  enabled: process.client,
});

if (process.client) {
  await suspense();
} else {
  onServerPrefetch(suspense);
}
```

If you execute this code, you'll notice that the web request never completes because `suspense()` never gets resolved.

https://stackblitz.com/edit/nuxt-starter-et73sx?file=app.vue,plugins%2Fvue-query.ts

### Steps to reproduce

1. Open reproduction
2. Click Open in New Tab ()
3. **page request never ends `suspense()` is never resolved.**

### Expected behavior

If the query is disabled, `suspense()` should resolve immediate

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: 4.32.6

### Tanstack Query adapter

vue-query

### TanStack Query version

v4.32.6

### TypeScript version

_No response_

### Additional context

_No response_

Mini-ghost:
resolve #5845

### DIFF
--- a/packages/vue-query/src/useBaseQuery.ts
+++ b/packages/vue-query/src/useBaseQuery.ts
@@ -109,10 +109,10 @@ export function useBaseQuery<
           //noop
         }
         const run = () => {
+          const optimisticResult = observer.getOptimisticResult(
+            defaultedOptions.value,
+          )
           if (defaultedOptions.value.enabled !== false) {
-            const optimisticResult = observer.getOptimisticResult(
-              defaultedOptions.value,
-            )
             if (optimisticResult.isStale) {
               stopWatch()
               observer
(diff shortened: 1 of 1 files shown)


Mini-ghost:
@DamianOsipiuk 

This unit test seems to be in conflict with the problem I hope to solve at present, I would greatly appreciate hearing your thoughts. Thank you.

https://github.com/TanStack/query/blob/46e7b0282077ccb2b11871dba6b6a19e8d8731b6/packages/vue-query/src/__tests__/useQuery.test.ts#L314-L332

PR #5848, opened 2023-08-07
fix(pnpm-lock.yaml): update

manudeli:
When I work on beta branch and main branch, I used `pnpm install`. this is result. so I add this Pull Request

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1380,6 +1380,15 @@ importers:
         specifier: npm:react-dom@^17.0.2
         version: /react-dom@17.0.2(react@18.2.0)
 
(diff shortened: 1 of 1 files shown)


TkDodo:
Did you set the correct node version with nvm? They are different between main and beta...

manudeli:
I checked `pnpm -v` too. but there is change on `pnpm install`

image

manudeli:
[b1f0196](https://github.com/TanStack/query/commit/b1f019640347044478a30978eba765f893c5a818)
This change force pushed used node-version(.nvmrc): v16.19.0, pnpm(package.json packageManager): v8.5.1

PR #5851, opened 2023-08-08
fix(node, pnpm): update version to lts

manudeli:
I update version of pnpm, node to fix https://github.com/TanStack/query/pull/5848

When I change node version in `.nvmrc` from 16 to 18, in my local machine `corepack` could find and use `package.json's packageManager(pnpm@8.6.12)'s version` automatically.
and current LTS node version is 18. so why not update node version.

and we could check no error in ci(`pnpm test`) too. (In my local machine, error is occured, I don't know why.. sorry, but It's resolved by this change)


TkDodo:
we have fixed this on `beta` already, where we are using the node version from `nvmrc` in our ci pipelines. merging this will just cause conflicts. beta will land soon.

Issue #5852, opened 2023-08-08
Wallaby.js test runner giving errors when attempting to run Vitest unit tests


arnoud-dv:
### Describe the bug

[Wallaby.js](https://wallabyjs.com/docs/index.html) is a plugin that runs unit tests while typing in various IDE's. It supports Vitest but on the beta branch I get errors while attempting to run the tests through Wallaby.js

While not free I like using it as it gives a fast feedback cycle while developing and writing unit tests. Note there is a [free license](https://wallabyjs.com/oss/) available for OSS projects.

### Your minimal, reproducible example

https://github.com/TanStack/query/tree/beta

## Visual Studio Code
- Install the Wallaby.js extension
- CTRL/CMD + Shift + P
  - choose *Wallaby.js: select configuration*
  - choose *Automatic Configuration <custom directory>*
  - select a package, for example query-core
  - Wallaby.js will attempt to run unit tests but gives the following error:
  
![image](https://github.com/TanStack/query/assets/6420061/1a8d440f-1577-47ec-bc84-ea78f024d034)

## Webstorm
- Install the Wallaby.js extension
- Add a run configuration as in the screenshot:
![image](https://github.com/TanStack/query/assets/6420061/ae9efd96-e294-4e2d-a57c-1f51b15d617d)
- Wallaby.js will attempt to run unit tests but gives the following error:
![image](https://github.com/TanStack/query/assets/6420061/dcf9029a-f697-4b92-9ed7-a2f33fcf3b69)

### Expected behavior

I expected Wallaby.js to be able to run the Tanstack Query unit tests.

### How often does this bug happen?

Every time

### Platform

- OS: Linux Fedora but probably all operating systems

### Tanstack Query adapter

None

### TanStack Query version

v5.0.0-beta.15

### TypeScript version

5.0.4

### Additional context

I was able to fix this and will submit a PR

webprofijt:
Noticed jsdom is a root dependency which didn't seem right to me, just having it as a sub dependency would be preferable I think. Removing it helped fix this issue.

Also added @vitest/runner as dependency as mentioned here:

              @lucastraba - it sounds like you must be using `pnpm`, if so, it's expected that you will need to install `@vitest/runner`. If you're not using `pnpm`, could you please let us know what package manager you are using? By default, you don't need to install `@vitest/runner`.

_Originally posted by @smcenlly in https://github.com/wallabyjs/public/issues/3146#issuecomment-1577583039_

TkDodo:
[...]

PR #5853, opened 2023-08-08
fix: Wallaby test runner

arnoud-dv:
remove jsdom workspace dependency
add @vitetest/runner workspace dependency

closes https://github.com/TanStack/query/issues/5852


TkDodo:
I tried out wallaby but I can't get it to run in IntelliJ:

- If I set `packages/query-core` as root path as you've shown in the screenshot in the issue, I'm getting:

```
Wallaby.js has been stopped because your project is not associated with a valid GitHub project.
```

- If I set the root of the monorepo as root path, I get:

```
0 failing tests, 0 passing  Launch Coverage/Test Explorer

ReferenceError: describe is not defined
	at packages/codemods/src/v5/remove-overloads/__tests__/remove-overloads.test.js:4:1
	at packages/codemods/src/v5/rename-hydrate/__tests__/rename-hydrate.test.js:4:1

ReferenceError: describe is not defined
	at packages/codemods/src/v5/rename-properties/__tests__/rename-properties.test.js:4:1

ReferenceError: describe is not defined
	at packages/codemods/src/v4/__tests__/key-transformation.test.js:4:1
	at packages/codemods/src/v4/__tests__/replace-import-specifier.test.js:4:1
```

I'd love to use wallaby, we can merge this PR because everything still works outside of it, but I'd appreciate a working setup with it. Is it possible that we're missing some sort of wallbay.js config file?

TkDodo:
There are way more errors in the wallaby console, but they all seem to have the same root cause:

```
Runtime error: ReferenceError: describe is not defined
    at ./packages/query-core/src/tests/focusManager.test.tsx:6:1
```

webprofijt:
> There are way more errors in the wallaby console, but they all seem to have the same root cause:
> 
> ```
> Runtime error: ReferenceError: describe is not defined
>     at ./packages/query-core/src/tests/focusManager.test.tsx:6:1
> ```

Hi Dominik, there is this [troubleshooting section](https://wallabyjs.com/docs/integration/vitest.html#troubleshooting) at the Wallaby site where as a troubleshooting step they suggest trying to run `npx vitest`. I guess the pnpm equivalent is `pnpm exec vitest`. When I run either command I also get describe is not defined errors.

I merged the latest beta into my local branch and there was also an update of Wallaby and sadly it doesn't work for me now either on the package level where it did before. When I have some time left I'll try to look into this further.

PR #5903, opened 2023-08-23
fix: basic-graphql-request link in v3 docs

cseas:
The example link on this page doesn't work currently:
https://tanstack.com/query/v3/docs/react/graphql


cseas:
Looks like this didn't work. The relative path in the code now points to the correct file, but the url finally getting rendered still isn't correct.

Now it is: https://tanstack.com/query/v3/docs/react/examples/basic-graphql-request
Should be: https://tanstack.com/query/v3/docs/react/examples/react/basic-graphql-request

Issue #5909, opened 2023-08-24
@typescript-eslint/eslint-plugin: Missing peerDependencies.


justsaul:
### Describe the bug

While running linter I've encountered:

```
Error: Failed to load plugin '@tanstack/query' declared in '--config » eslint-config': @tanstack/eslint-plugin-query tried to access eslint, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
```

This issue should only concern users of package managers that don't rely on node_module resolution algo (yarn with nodeLinker strategy PnP in strict mode). This way, eslint plugin doesn't gain access to eslint, because its not declared as peerDependency. 

General solution is to define peerDependency within @typescript-eslint/eslint-plugin package.json such as:

```JSON
{
  "peerDependencies":  {
    "eslint": "^6 || ^7 || ^8"  // range of eslint matches '@typescript-eslint' package
  }
}
```

### Your minimal, reproducible example

-

### Steps to reproduce

- Define plugin
- run linter

### Expected behavior

Linter should exectute.

### How often does this bug happen?

Every time

### Platform

macOS

### Tanstack Query adapter

None

### TanStack Query version

v4.32.0

### TypeScript version

v4.8.3

### Additional context

Let me know if you agree with this, can create PR.

TkDodo:
Sounds good 👍

PR #5917, opened 2023-08-27
fix(eslint-plugin-query): provide missing peerDependencies

justsaul:
Exposing eslint as peerDependency This is needed for package managers that don't rely on hoisting heuristic to determine dependencies.

Addressing: [issue](https://github.com/TanStack/query/issues/5909#issue-1865041635)

### DIFF
--- a/packages/eslint-plugin-query/package.json
+++ b/packages/eslint-plugin-query/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tanstack/eslint-plugin-query",
-  "version": "4.32.5",
+  "version": "4.32.6",
   "description": "ESLint plugin for TanStack Query",
   "author": "Eliya Cohen",
   "license": "MIT",
@@ -51,5 +51,8 @@
     "@typescript-eslint/utils": "^5.41.0",
     "eslint": "^8.26.0",
     "tsup": "^6.3.0"
+  },
+  "peerDependencies": {
+    "eslint": "^6.0.0 || ^7.0.0 || ^8.0.0"
   }
 }


packages/eslint-plugin-query/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tanstack/eslint-plugin-query",
-  "version": "4.32.5",
+  "version": "4.32.6",
```
TkDodo, line 3: this will be incremented automatically by our build script, so please revert that:

```suggestion
  "version": "4.32.5",
```

packages/eslint-plugin-query/package.json
```diff
@@ -51,5 +51,8 @@
     "@typescript-eslint/utils": "^5.41.0",
     "eslint": "^8.26.0",
     "tsup": "^6.3.0"
+  },
+  "peerDependencies": {
+    "eslint": "^6.0.0 || ^7.0.0 || ^8.0.0"
```
TkDodo, line 56: shouldn't adding this result in a change in the our lockfile ?

TkDodo (changes requested):

packages/eslint-plugin-query/package.json
justsaul: I'm not too sure how pnpm works, but:
- i've ran  `pnpm i --frozen-lockfile` before & after changes;
- from what i've observed - pnpm doesn't store `peerDependencies` for packages that are defined within repository. For example [packages/react-query doesn't have peerDependencies in lock file but has them defined within package.json](https://github.com/TanStack/query/blob/857ecd91b6d86306bacd3785e3e2f71d7287ad06/pnpm-lock.yaml#L1272C3-L1283C21). Which makes sense - you lock it from consumer end.

TkDodo: you're right, there's no change. For the future, you'd want to run `pnpm i` to update to potentially update the lockfile.

Issue #5920, opened 2023-08-28
v5 docs: incorrect/misleading suspense docs


ryanliljestrom:
### Describe the bug

The third paragraph for the [v5 suspense docs](https://tanstack.com/query/v5/docs/react/guides/suspense) include the following when talking about throwing errors to the nearest error boundary:

```
If you wish to disable this, you can set the throwOnError option to false. If you wish that errors are not thrown at all, you can set the throwOnError option to false as well!
```

I'm guessing this is a copy/paste issue from the [v4 docs](https://tanstack.com/query/v4/docs/react/guides/suspense), when `useErrorBoundary` was re-named to `throwOnError`

### Your minimal, reproducible example

https://tanstack.com/query/v5/docs/react/guides/suspense

### Steps to reproduce

1. Go to https://tanstack.com/query/v5/docs/react/guides/suspense
2. See the third paragraph, starts with `In addition to queries...`. The last two sentences are the source of the confusion

### Expected behavior

I'm guessing the second sentence can be removed, or both of them re-worked

### How often does this bug happen?

Every time

### Platform

- Browser

### Tanstack Query adapter

None

### TanStack Query version

v5

ryanliljestrom:
Yup, can do later this PM

PR #5923, opened 2023-08-28
docs: update suspense.md

rliljest:
Fixes https://github.com/TanStack/query/issues/5920


docs/react/guides/suspense.md
```diff
@@ -11,7 +11,7 @@ React Query can also be used with React's Suspense for Data Fetching API's. For
...
 
-In addition to queries behaving differently in suspense mode, mutations also behave a bit differently. By default, instead of supplying the `error` variable when a mutation fails, it will be thrown during the next render of the component it's used in and propagate to the nearest error boundary, similar to query errors. If you wish to disable this, you can set the `throwOnError` option to `false`. If you wish that errors are not thrown at all, you can set the `throwOnError` option to `false` as well!
+In addition to queries behaving differently in suspense mode, mutations also behave a bit differently. By default, instead of supplying the `error` variable when a mutation fails, it will be thrown during the next render of the component it's used in and propagate to the nearest error boundary, similar to query errors.
```
TkDodo, line 14: Ooh, I didn't see that this is about mutations. As far as I know, mutations don't even support suspense, so the whole paragraph is weird ...

TkDodo, line 14: let's just do this:

```suggestion
If you want mutations to propagate errors to the nearest error boundary (similar to queries), you can set the `throwOnError` option to `true` as well.
```

PR #5928, opened 2023-08-30
feat(svelte-query-persist-client): Svelte implementation of client persistence

aadito123:
I only learned enough Svelte to write this package so please let me know if there's anything wrong. I migrated the React tests to Svelte and all the tests pass.

Also made changes to the base svelte-query package so that `isRestoring` is respected.

Only thing missing is the persistence across multiple `QueryClient`s. It is there in the React and Solid implementations because they have reactive wrappers around the `QueryClient` whereas the Svelte one does not. Not sure if there is a way to make it reactive in this situation without some API changes.

Addresses #5463


aadito123:
Pushed the formatting fix. The other test that failed was in the react-query package.

aadito123:
hey @lachlancollins, do you think you could take a look at this?

packages/svelte-query/src/context.ts
```diff
@@ -19,3 +21,17 @@ export const getQueryClientContext = (): QueryClient => {
...
+/** Retrieves a `isRestoring` from Svelte's context */
+export const getIsRestoringContext = (): Readable<boolean> => {
+  const isRestoring = getContext(_isRestoringContextKey)
+  if (!isRestoring) return readable(false)
+  return isRestoring as Readable<boolean>
+}
+
+/** Sets a `isRestoring` on Svelte's context */
+export const setIsRestoringContext = (isRestoring: Readable<boolean>): void => {
+  setContext(_isRestoringContextKey, isRestoring)
```
lachlancollins, line 36: Just want to check that this works correctly - is it okay to pass in a store directly into setContext like this?

lachlancollins (commented):
Amazing work, thanks so much for working on this @aadito123 ! I've had a quick look and everything overall seems good, but I'll have a closer look shortly.

packages/svelte-query/src/context.ts
lachlancollins: @aadito123 no that's awesome, I didn't realise it was a thing :)

frederikhors:
Any docs @aadito123? Thank you very much for this!

aadito123:
@frederikhors Nothing right now but I would be happy to work on them. In the meantime, you can look at the [test cases](https://github.com/TanStack/query/tree/beta/packages/svelte-query-persist-client/src/__tests__) for usage examples.

frederikhors:
Thanks. I'll try it asap (I'm still on V4 for now).

PR #5929, opened 2023-08-30
fix(vue-query): Make queryKey watchers sync

### LINKED ISSUE #5910: Vue - Query key doesn't match the current value of ref used inside
### Describe the bug

Hi,

I was trying to build a filter for some data using tanstack query.
It's supposed to only refetch the data with the given values whenever the user clicks a button so not automatically.
Therfore I set enabled to false to only execute the query when manually calling `refetch()`.

When testing it I encountered some wired behavior: The query key seems to always lack behind one iteration from the actual value of my filter.
I created a minimal example of the issue to make it reproducable and easier to explain:

```vue
<script setup lang="ts">
import { useQuery } from '@tanstack/vue-query';
import { nextTick, onMounted, ref } from 'vue';

const id = ref(0);

const {data, refetch} = useQuery({
  queryKey: ['multiply', id] as const,
  queryFn: async ({queryKey}) => {
    console.log("Current value of ref:", id.value);
    console.log("Current value of query key:", queryKey[1]); // Will always be one step behind
    
    return Promise.resolve(id.value * 2)
  },
  enabled: false
});
onMounted(() => refetch());

async function increment() {
  console.log('Increment clicked');
  id.value++;
  // await nextTick(); // Awaiting the next tick before refetching fixes the issue
  refetch();
}
</script>

<template>
  <div>Id times two: {{ data }}</div>
  <button @click="increment">Increment id</button>
</template>
```

In the above example I want to update data with the new value of id multiplied by two whenever the Increment button is clicked. (This could of course be done without manually calling `refetch()` but I needed a simple example)

Whenever the increment button is clicked the id gets incremented correctly but the query used during the refetch will still use the old value of id. Therfore the result is assigned to the old key:

image

Meanwhile data is always undefined as the key changes right after the data was fetched.
If I manually await the next tick inside vues' reactivity system the value is updated correctly before the refetch is triggered and everything works as expected.
That's why I guess it could be something related to the reactivity chain of the queryKey. Maybe a lazily evaluated computed or something similar.

https://github.com/markbrockhoff/tanstack-query-bug-repro

### Steps to reproduce

1. Clone the [reproduction repo](https://github.com/markbrockhoff/tanstack-query-bug-repro)
2. Install the dependencies: `pnpm i`
3. Start the project: `pnpm dev`
4. Click the "Increment" button and look at the logs inside the console

### Expected behavior

I'd expect the changes to reactive parts of the queryKey to take affect immediately so the refetch uses the correct queryKey without manually awaiting the nextTick before triggering it.

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: macos
- Browser: Chrome

### Tanstack Query adapter

vue-query

### TanStack Query version

v4.34.0

### TypeScript version

5.2.2

### Additional context

_No response_

markbrockhoff:
Make watchers of vue composables like useQuery sync to so changes to reactive variables inside the queryKey take effect immediately.

Closes #5910


packages/vue-query/src/__tests__/useQuery.test.ts
```diff
@@ -192,26 +187,26 @@ describe('useQuery', () => {
 
     const enabled = computed(() => !!data.value)
 
-    const { fetchStatus, status } = useQuery(
+    const dependentQueryFn = jest.fn()
```
DamianOsipiuk, line 190: I believe this mock fn could invoke `simpleFetcher` so it could be actually resolved

packages/vue-query/src/__tests__/useQuery.test.ts
```diff
@@ -192,26 +187,26 @@ describe('useQuery', () => {
...
     )
 
     expect(data.value).toStrictEqual(undefined)
     expect(fetchStatus.value).toStrictEqual('idle')
+    expect(dependentQueryFn).not.toHaveBeenCalled()
 
     await flushPromises()
 
     expect(data.value).toStrictEqual('Some data')
-    expect(fetchStatus.value).toStrictEqual('fetching')
```
DamianOsipiuk, line 209: i would prefer to keep the fetch status check, before and after to make sure dependent query state transitions are correct

markbrockhoff: No problem :)

PR #5931, opened 2023-08-30
chore: add return type

duguyihou:


TkDodo:
okay but why? the function is only used internally, and the return type can easily be inferred by the compiler ...

PR #5935, opened 2023-08-30
fix(vue-query): ensure the `queryFn` parameter has the correct type

Mini-ghost:
Ensure that the `queryFn` parameter has the correct data type to resolve  the following issues (discussions):
- #5930 

## Demo code

```ts
import { useQuery, UseQueryOptions } from '@tanstack/vue-query';
import { Ref, toRef } from 'vue';

type MaybeRef<T> = T | Ref<T>;
type MaybeRefOrGetter<T> = MaybeRef<T> | (() => T);

const fetchUser = (userId: number) =>
  fetch(`/api/user/${userId}`).then(res => res.json());

type User = any;
export function useUser(
  userId: MaybeRefOrGetter<number>,
  options?: UseQueryOptions<User, any, User, readonly [string, MaybeRef<number>]>
) {
  return useQuery({
    queryKey: ['user', toRef(userId)],
    queryFn({ queryKey }) {
      return fetchUser(queryKey[1]);
    },
    ...options,
  });
}
```

## Before

```bash
(parameter) queryKey: readonly [string, MaybeRef<number>]
```

截圖 2023-08-31 上午3 29 56

## After

```bash
(parameter) queryKey: readonly [string, number]
```

截圖 2023-08-31 上午3 39 00

### DIFF
--- a/packages/vue-query/src/types.ts
+++ b/packages/vue-query/src/types.ts
@@ -8,7 +8,7 @@ import type {
 import type { Ref, UnwrapRef } from 'vue-demi'
 import type { QueryClient } from './queryClient'
 
-export type MaybeRef<T> = Ref<T> | T
+type Primitive = string | number | boolean | bigint | symbol | undefined | null
 
 export type MaybeRefDeep<T> = MaybeRef<
   T extends Function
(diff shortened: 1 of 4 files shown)


Mini-ghost:
@callumacrae Oops! This is obviously my mistake. I will fix it as soon as possible and hope it solves your issue.

PR #5947, opened 2023-09-02
Update error queries.md

HectorShaw:
TSX marks "error" as unknown a simple solution to this  is to use type assertation.


TkDodo:
we're defaulting to error in v5. in v4, the better check would be:

```diff
- if (isError)
+ if (error instanceof Error)
```

PR #5950, opened 2023-09-03
Update devtools.md

callmedukey:
installing it as a regular dependency doesn't work during development with next 13 app dir. Import trace errors thrown all around.


TkDodo:
> installing it as a regular dependency doesn't work during development with next 13 app dir. Import trace errors thrown all around.

that's weird. Can you show a reproduction please? Is this a nextJs issue ?

PR #5640, opened 2023-06-27
adding the ts matrix

ZackDeRose:

PR #5953, opened 2023-09-04
fix some false positive for prefer-query-object-syntax rule

### LINKED ISSUE #5028: ESLint plugin does not work well with TypeScript `satisfies`
### Describe the bug

Hey there! 👋🏻 

The ESLint rule `prefer-query-object-syntax` does not handle the case correctly when `satisfies` is used.

```ts
const queryDefinition = {
    queryKey: ['my-key'],
    queryFn: fetchSomething,
} satisfies UseQueryOptions<{ data: any }>;

function useSomething() {
    return useQuery(queryDefinition);
                    // 👆 Not Happy and make it { queryFn: queryDefinition } when fixing
}
```

https://codesandbox.io/p/sandbox/withered-cookies-zogzv2

### Steps to reproduce

I tried to reproduce the issue in a Codesandbox, but it seems to not handle the `satisfies` keyword. Therefore, ESLint cannot have the error.

You can activate the rule and copy-paste the code shared above to give it a try.

### Expected behavior

ESLint should not catch this code as a mistake.

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- IDE: JetBrains PHP Storm

### TanStack Query version

4.24.6

### TypeScript version

4.9.5

### Additional context

_No response_

eg-bernardo:
Fixes #5028, and cases where the query options object is not defined in the same file as the `useQuery` call.

To be able to test the issue with `satisfies` I had to update `typescript` and `@typescript-eslint` packages, but I have only updated `devDependencies` in `eslint-plugin-query`.

I have removed a test case for invalid code:
```
useQuery(queryKey);
```
Because I think in most cases for `useQuery(some_variable)` when `some_variable` is not defined in the current file, it will be a query options object.

### DIFF
--- a/packages/eslint-plugin-query/src/rules/prefer-query-object-syntax/prefer-query-object-syntax.test.ts
+++ b/packages/eslint-plugin-query/src/rules/prefer-query-object-syntax/prefer-query-object-syntax.test.ts
@@ -55,6 +55,13 @@ ruleTester.run(name, rule, {
     {
       code: normalizeIndent`
         import { useQuery } from "@tanstack/react-query";
+        import { postsQuery } from "somewhere-else";
+        const usePosts = useQuery(postsQuery);
+      `,
+    },
+    {
+      code: normalizeIndent`
+        import { useQuery } from "@tanstack/react-query";
         const getQuery = () => ({ queryKey: ['foo'], queryFn: () => Promise.resolve(5) })
         useQuery(getQuery())
       `,
@@ -91,6 +98,13 @@ ruleTester.run(name, rule, {
             return { queryKey: ['foo'], queryFn: () => Promise.resolve(5) };
           }
         }
+        useQuery(getQuery())
+      `,
+    },
+    {
+      code: normalizeIndent`
+        import { useQuery } from "@tanstack/react-query";
+        import { getQuery } from "somewhere-else";
         useQuery(getQuery())
       `,
     },
@@ -149,17 +163,6 @@ ruleTester.run(name, rule, {
       output: normalizeIndent`
         import { useQuery } from "@tanstack/react-query";
         useQuery({ queryKey: ['data'] });
-      `,
-    },
-    {
-      code: normalizeIndent`
-        import { useQuery } from "@tanstack/react-query";
-        useQuery(queryKey);
-      `,
-      errors: [{ messageId: 'preferObjectSyntax' }],
-      output: normalizeIndent`
-        import { useQuery } from "@tanstack/react-query";
-        useQuery({ queryKey });
       `,
     },
     {
--- a/packages/eslint-plugin-query/src/rules/prefer-query-object-syntax/prefer-query-object-syntax.test.ts
+++ b/packages/eslint-plugin-query/src/rules/prefer-query-object-syntax/prefer-query-object-syntax.test.ts
@@ -158,6 +158,16 @@ ruleTester.run(name, rule, {
         const usePosts = () => useMutation(postsQuery);
       `,
     },
+    {
+      code: normalizeIndent`
+        import { useQuery } from "@tanstack/react-query";
+        const queryOptions = {
+          queryKey: ["queryKey"],
+          queryFn: () => Promise.resolve([]),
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin-query/src/rules/prefer-query-object-syntax/prefer-query-object-syntax.test.ts
```diff
@@ -3,7 +3,14 @@ import { normalizeIndent } from '../../utils/test-utils'
...
 const ruleTester = new ESLintUtils.RuleTester({
-  parser: '@typescript-eslint/parser',
+  /** HACK
+   * We are using a newer version of typescript than the one at the repository root,
+   * to be able to test code that uses newer typescript features.
+   * To prevent RuleTester from using the typescript version at the root
+   * we resolve '@typescript-eslint/parser' here instead of letting it do it internally.
+   * Once the version at the root is updated to >=4.9 this can be replaced by just "@typescript-eslint/parser"
+   */
+  parser: require.resolve('@typescript-eslint/parser') as any,
```
TkDodo, line 13: where are we using an older version of `@typescript-eslint/parser` ?

eg-bernardo: here: https://github.com/TanStack/query/blob/ef5125f68bc41c6c01219f7cb4b098b8d35a2045/package.json#L61

I came to the conclusion that I also needed to update typescript to >=4.9 to be able to test `satisfies` and I didn't wan't to change the typescript version in the respository root. But maybe I can update @typescript-eslint in the root. I will give it another look.

eg-bernardo: I have tried updating only `@typescript-eslint/*` packages to oldest version that suports `satisfies` (5.45.0), but that is not enough, `typescript` >=4.9 is also needed, otherwise the test that uses `satisfies` fails with `A fatal parsing error occurred: Parsing error: ',' expected.`

I didn't want to update the typescript version in the root, because I assumed that it is probably there to ensure that packages stay compatible with that version of typescript.

TkDodo: ah yes, I see. what we'd want is to develop with TS 5.2, but have a type-testing suite that works against all sorts of TS versions.

@ZackDeRose started with this some time back, but we never followed through:

- https://github.com/TanStack/query/pull/5640

I'll take a look at setting this up later this week, then we should be good without this workaround 👍

Newbie012:
My personal take:
I think it would be wrong to remove the 2 tests, since `useQuery(["query-key"])` is considered a valid syntax in older versions.
Last thing, consider adding another 2 separate tests - one with only `as const` and one with only `satisfies`.

Other than that, LGTM.

eg-bernardo:
I think githubs diff is not being very helpful, because I have only removed one test and added two here: https://github.com/TanStack/query/pull/5953/commits/27b7153e3300539c42cf209214463e21e61f935d and then added another here https://github.com/TanStack/query/pull/5953/commits/331fa0888341a9de1dad8c18e5ad99ea5f6d7285. It's clearer in the diffs for the individual commits.

The test for `useQuery(["queryKey"])` is still there https://github.com/TanStack/query/blob/331fa0888341a9de1dad8c18e5ad99ea5f6d7285/packages/eslint-plugin-query/src/rules/prefer-query-object-syntax/prefer-query-object-syntax.test.ts#L177

eg-bernardo:
This PR is only relevant for v4 since `prefer-query-object-syntax` is no longer required on v5.

Also it was blocked because it required typescript 4.9:
https://github.com/TanStack/query/pull/5953#discussion_r1316096252

and you said:
https://github.com/TanStack/query/pull/5953#discussion_r1316164878

But it seems v4 still targets typescript 4.7.

TkDodo:
you're right, I'll close this then.

PR #5954, opened 2023-09-04
fix(vue-query): ensure built-in types are not being unwrapped

Mini-ghost:
## Resolve the following error

This error was mentioned in PR  https://github.com/TanStack/query/pull/5935#issuecomment-1705020256

```ts
import { useQuery, UseQueryOptions } from https://github.com/TanStack/query/tree/master/packages/vue-query;
import { Ref, toRef } from https://github.com/vuejs/core;

type MaybeRef<T> = T | Ref<T>;
type MaybeRefOrGetter<T> = MaybeRef<T> | (() => T);

const fetchUser = (userId: Date) =>
  fetch(`/api/user/${userId.toISOString()}`).then(res => res.json());

type User = any;
export function useUser(
  userId: MaybeRefOrGetter<Date>,
  options?: UseQueryOptions<User, any, User, readonly [string, Ref<Date>]>
) {
  return useQuery({
    queryKey: ['user', toRef(userId)],
    queryFn({ queryKey }) {
      return fetchUser(queryKey[1]);
    },
    ...options,
  } satisfies typeof options);
}
```

```
example.ts:185:24 - error TS2345: Argument of type '{ toString: {}; toDateString: {}; toTimeString: {}; toLocaleString: {}; toLocaleDateString: {}; toLocaleTimeString: {}; valueOf: {}; getTime: {}; getFullYear: {}; getUTCFullYear: {}; getMonth: {}; ... 32 more ...; [Symbol.toPrimitive]: {}; }' is not assignable to parameter of type 'Date'.
  Types of property 'toString' are incompatible.
    Type '{}' is not assignable to type '() => string'.
      Type '{}' provides no match for the signature '(): string'.
```

In that PR, not avoiding unwrapping **built-in types** resulted in errors, such as Ref<Date>, occurring.

## Before
截圖 2023-09-04 下午7 35 04

## After

截圖 2023-09-04 下午7 36 39

### DIFF
--- a/packages/vue-query/src/types.ts
+++ b/packages/vue-query/src/types.ts
@@ -9,6 +9,7 @@ import type { Ref, UnwrapRef } from 'vue-demi'
 import type { QueryClient } from './queryClient'
 
 type Primitive = string | number | boolean | bigint | symbol | undefined | null
+type Builtin = Primitive | Function | Date | Error | RegExp;
 
 export type MaybeRefDeep<T> = MaybeRef<
   T extends Function
@@ -22,7 +23,7 @@ export type MaybeRefDeep<T> = MaybeRef<
 
 export type MaybeRef<T> = Ref<T> | T
 
-export type DeepUnwrapRef<T> = T extends Primitive
+export type DeepUnwrapRef<T> = T extends Builtin
   ? T
   : T extends Ref<infer U>
   ? DeepUnwrapRef<U>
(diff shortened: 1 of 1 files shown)


callumacrae:
what about user-defined classes?

Mini-ghost:
When dealing with user-defined classes, we can draw inspiration from Vue's handling method, as demonstrated here:

https://github.com/vuejs/core/blob/0e8bbe873e579f3d3a74c44af28f7df9e7a06978/packages/reactivity/src/ref.ts#L459-L472

A possible implementation might resemble the following:

```ts
declare module '@tanstack/vue-query' {
   export interface RefUnwrapBailTypes {
     UserDefinedClass: UserDefinedClass;
   }
}
```

However, it's important to note that @tanstack/query-core will convert the `queryKey` into a string using `JSON.stringify` to use it as a key for cache access. This approach might encounter issues with types like `Map`, `Set`, or user-defined classes that cannot be stringified.

Given this consideration, I'm not sure if `queryKey` should support such a wide range of types.

PR #5963, opened 2023-09-05
Update dependent-queries.md

jomefavourite:
This is to show how the useQueries hook can also be dependent on a query.

---

## useQueries dependent Query

Dynamic parallel query - useQueries can depend on a previous query also, here's how to achieve this:

```tsx
// Get the users Data
const { data: users } = useQuery({
  queryKey: ['users'],
  queryFn: getUsersData,
})

const usersId = users?.filter(user => user?.id)

// Then get the user's messages
const usersMessages = useQueries({
  queries: users
    ? usersId.map(id => {
        return {
          queryKey: ['messages', id],
          queryFn: () => getMessagesByUsers(id),
        };
      })
  : [], // if users is underfined an empty array will be returned
})
```

**Note** that useQueries return an array of query results

### DIFF
--- a/docs/react/guides/dependent-queries.md
+++ b/docs/react/guides/dependent-queries.md
@@ -58,7 +58,7 @@ fetchStatus: 'idle'
 
 Dynamic parallel query - `useQueries` can depend on a previous query also, here's how to achieve this:
 
-[//]: # 'Example'
+[//]: # 'Example2'
 
 ```tsx
 // Get the users ids
@@ -80,5 +80,7 @@ const usersMessages = useQueries({
   : [], // if users is undefined, an empty array will be returned
 })
 ```
+
+[//]: # 'Example2'
 
 **Note** that `useQueries` return an **array of query results**
--- a/docs/react/guides/dependent-queries.md
+++ b/docs/react/guides/dependent-queries.md
@@ -78,7 +78,7 @@ const usersMessages = useQueries({
           queryFn: () => getMessagesByUsers(id),
         };
       })
-  : [], // if users is underfined an empty array will be returned
+  : [], // if users is undefined, an empty array will be returned
 })
 ```
 
--- a/docs/react/guides/dependent-queries.md
+++ b/docs/react/guides/dependent-queries.md
@@ -61,13 +61,12 @@ Dynamic parallel query - `useQueries` can depend on a previous query also, here'
 [//]: # 'Example'
 
 ```tsx
-// Get the users Data
-const { data: users } = useQuery({
+// Get the users ids
+const { data: userIds } = useQuery({
   queryKey: ['users'],
   queryFn: getUsersData,
+  select: users => users.map(user => user.id),
 })
-
-const usersId = users?.filter(user => user?.id)
 
 // Then get the users messages
 const usersMessages = useQueries({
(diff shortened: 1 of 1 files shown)


docs/react/guides/dependent-queries.md
```diff
@@ -51,3 +53,33 @@ Once we have the projects, it will go to:
...
+// Then get the users messages
+const usersMessages = useQueries({
+  queries: users
+    ? usersId.map(id => {
+        return {
+          queryKey: ['messages', id],
+          queryFn: () => getMessagesByUsers(id),
+        };
+      })
+  : [], // if users is underfined an empty array will be returned
```
TkDodo, line 81:
```suggestion
  : [], // if users is undefined, an empty array will be returned
```

docs/react/guides/dependent-queries.md
```diff
@@ -51,3 +53,33 @@ Once we have the projects, it will go to:
 status: 'success'
 fetchStatus: 'idle'
 ```
+
+## useQueries dependent Query
+
+Dynamic parallel query - `useQueries` can depend on a previous query also, here's how to achieve this:
+
+[//]: # 'Example'
```
TkDodo, line 61: we need a different name than `Example` here, and it also needs to have a corresponding, closing "tag".

docs/react/guides/dependent-queries.md
```diff
@@ -51,3 +53,33 @@ Once we have the projects, it will go to:
...
+
+```tsx
+// Get the users Data
+const { data: users } = useQuery({
+  queryKey: ['users'],
+  queryFn: getUsersData,
+})
+
+const usersId = users?.filter(user => user?.id)
```
TkDodo, line 70:
```suggestion
// Get the users ids
const { data: userIds } = useQuery({
  queryKey: ['users'],
  queryFn: getUsersData,
  select: users => users.map(user => user.id),
})
```

docs/react/guides/dependent-queries.md
```diff
@@ -51,3 +53,33 @@ Once we have the projects, it will go to:
...
+
+// Then get the users messages
+const usersMessages = useQueries({
+  queries: users
+    ? usersId.map(id => {
```
TkDodo, line 75:
```suggestion
  queries: userIds
    ? usersId.map(id => {
```

docs/react/guides/dependent-queries.md
jomefavourite: I think the Example is commented out, so it doesn't show in the preview.

As for the closing tag, I'm sorry, I don't understand how you mean, could you explain more?

TkDodo: see e.g. here:

https://github.com/TanStack/query/blob/716c5d8b096c624f6643dc9f263fbdd6f0565021/docs/react/guides/paginated-queries.md#L8-L15

jomefavourite: Okay, I get it now, I hope `Example2` will be okay.

However, I don't understand why the syntax is needed or what it does` [//]: # 'Example' `

TkDodo: it's a custom renderer that specifies example blocks in react code that can be switched out with other code in other frameworks like vue or svelte

jomefavourite: Okay, I've changed it to `Example2`.
Is this resolved now?

PR #5966, opened 2023-09-06
SSR docs rewrite

Ephem:
Most of this content is applicable in v4 as well, except for naming of `useSuspenseQuery` etc, but I've written and targeted this to the v5 branch.

There are four guides that progressively builds knowledge:

- [x] Performance & Request Waterfalls (new)
  - This is just about Request Waterfalls performance today, and that's reflected by the slug. We could add other performance tips later or split that to it's own page. This is mostly a background/foundation page, but does include practical tips too.
- [x] Prefetching & Router Integration
  - Client focused, but lays the foundation for the server parts.
  - Mostly kept the old content around at the top of the page for quick reference.
- [x] Server Rendering & Hydration
- [x] Advanced Server Rendering (new)

I'm open to suggestions on naming for the pages.

There are two existing doc pages that are related to these:

* [Parallel Queries](https://tanstack.com/query/latest/docs/react/guides/parallel-queries)
* [Dependent Queries](https://tanstack.com/query/latest/docs/react/guides/dependent-queries)

I'm not sure what to do with these. While there is some duplicate content there, they are also short and practical so I think I'm leaning towards keeping them around for discoverability. It's also good that these are mentioned early/high up in the guide pages navigation, while these 4 guides are more in depth and makes sense to have further down in the navigation.

I removed some content related to setting up React Query with custom SSR solutions. I don't think that's relevant to a majority of people, and the step by step instructions are still there so people can figure it out. If we want this I think it would fit better as an example, but I wont include that in this PR. Other than that, I don't think I removed any information, just added, rewrote or restructured.


TkDodo (commented):
awesome work so far @Ephem 👏

docs/react/guides/prefetching.md
```diff
@@ -1,11 +1,32 @@
...
+
+In this guide, we'll take a look at the first three, while the fourth will be covered in depth in the [Server Rendering & Hydration guide](../guides/ssr) and the [Advanced Server Rendering guide](../guides/advanced-ssr).
+
+One specific use of prefetching is to avoid Request Waterfalls, for an in-depth background and explanation of those, see the [Performance & Request Waterfalls guide](../guides/request-waterfalls).
+
+## prefetchQuery & prefetchInfiniteQuery
+
+Before jumping into the different specific prefetch patterns, let's look at the `prefetchQuery` and `prefetchInfiniteQuery` functions. First a few basics:
+
+- If **fresh** data for this query is already in the cache, the data will not be fetched
```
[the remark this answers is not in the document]
Ephem: Excellent point! I copied this from the old docs and didn't think much about it, should be fixed now.

docs/react/guides/prefetching.md
```diff
@@ -1,11 +1,32 @@
...
+- If a `staleTime` is passed eg. `prefetchQuery({ queryKey: ['todos'], queryFn: fn, staleTime: 5000 })` and the data is older than the specified `staleTime`, the query will be fetched
+  - (If you want to ignore `staleTime` and instead always return data if it's available in the cache, you can use the `ensureQueryData` function.)
+- If no instances of `useQuery` appear for a prefetched query, it will be deleted and garbage collected after the time specified in `gcTime`
+- These functions does not return the query data. If that's something you need, use `fetchQuery`/`fetchInfiniteQuery` instead.
```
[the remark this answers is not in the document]
Ephem: Oh yeah! The error one has bit even me in the past, glad you pointed it out. Fixed.

docs/react/guides/prefetching.md
```diff
@@ -36,30 +48,326 @@ const prefetchTodos = async () => {
...
+
+  const prefetch = useCallback(() => {
+    queryClient.prefetchQuery({
+      queryKey: ['details'],
+      queryFn: getDetailsData,
+      // Prefetch only fires when data is older than the staleTime,
+      // so in a case like this you definitely want to set one
+      staleTime: 60000,
+    })
+  }, [queryClient])
```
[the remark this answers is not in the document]
Ephem: Good point, fixed!

docs/react/guides/prefetching.md
```diff
@@ -36,30 +48,326 @@ const prefetchTodos = async () => {
...
+```
+
+This results in a request waterfall looking like this:
+
+```
+1. |> getArticleById()
+2.   |> getArticleCommentsById()
+```
+
+As mentioned in that guide, one way to flatten this waterfall and improve performance is to hoist the `getArticleCommentsById` query to the parent and pass down the result as a prop, but what if this is not feasible or desirable, for example when the components are unrelated and has multiple levels between them?
```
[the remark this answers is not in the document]
Ephem: Thanks! Resolved.

docs/react/guides/prefetching.md
```diff
@@ -36,30 +48,326 @@ const prefetchTodos = async () => {
...
+  const { data: articleData, isPending } = useQuery({
+    queryKey: ['article', id],
+    queryFn: getArticleById,
+  })
+
+  // Prefetch
+  useQuery({
+    queryKey: ['article-comments', id],
+    queryFn: getArticleCommentsById,
+  })
```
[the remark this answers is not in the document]
Ephem: Nice one, added! Didn't know about that myself actually.

Yeah, I think there is a strong argument for adding that, it also encourages prefetching which is nice. If we add it, let's make sure it matches any fetch/read API for the suspense side of things. We could probably have just one common prefetch hook for both.

docs/react/guides/prefetching.md
```diff
@@ -36,30 +48,326 @@ const prefetchTodos = async () => {
...
+    // Fetch comments asap, but don't block
+    queryClient.prefetchQuery(commentsQueryOptions)
+
+    // Don't render the route at all until article has been fetched
+    await queryClient.prefetchQuery(articleQueryOptions)
+  },
+  component: ({ useRouteContext }) => {
+    const { articleQueryOptions, commentsQueryOptions } = useRouteContext()
+    const articleQuery = useQuery(articleQueryOptions)
+    const commentsQuery = useQuery(commentsQueryOptions)
```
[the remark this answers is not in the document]
Ephem: Yeah, it's neat! Can't take credit though, stole it from the Tanstack router example. 😁

I have built similar abstractions in the past. Reusing the options between prefetch/useQuery is a common usecase we should probably take a look at from a library or docs standpoint.

Ephem:
Phew, I need to re-read this during the weekend, but it should be ready for review now. It might not be perfect, but I hope it's better than the current docs at least.

I'd love high level feedback as well. Are the pages too long/verbose/tries to cover too much? Are there sections we could skip or are there sections we should add? Also, I've made a few "official recommendations" in there, do you agree with them? I am biased towards prefetching, does that shine through too much? What about the naming of the guides, specifically, should we call the "Advanced Server Rendering" guide "Server Components & Streaming" instead?

Finally, this messes up the Vue docs for prefetching since those are just copied from the React docs, and there are now React specific things in there. I wont have time to fix this properly right now, do you want me to just copy the old doc over into the Vue docs folder in this PR?

docs/react/guides/prefetching.md
```diff
@@ -36,30 +51,328 @@ const prefetchTodos = async () => {
...
+
+You can also do this with suspense queries, but instead of using two `useSuspenseQuery` you'll have to use `useSuspenseQueries` to group the queries instead, just ignore the result from the query you are prefetching:
+
+```tsx
+const [articleResult] = useSuspenseQueries({
+  queries: [
+    { queryKey: ['article', id], queryFn: getArticleById },
+    // Prefetch
+    { queryKey: ['article-comments', id], queryFn: getArticleCommentsById },
+  ],
```
TkDodo, line 185: a little gotcha here: Now, your component will only resume after suspending once both queries have finished. if `getArticleCommentsById` takes a long time or e.g. needs to go through retries, you'll see the suspense fallback for much longer than necessary. So I don't think this is the best idea after all :/

docs/react/guides/ssr.md
```diff
@@ -1,134 +1,136 @@
...
-The setup is minimal and this can be a quick solution for some cases, but there are a **few tradeoffs to consider** when compared to the full approach:
+Read on to learn how to implement these three steps with React Query.
 
-- If you are calling `useQuery` in a component deeper down in the tree you need to pass the `initialData` down to that point
-- If you are calling `useQuery` with the same query in multiple locations, you need to pass `initialData` to all of them
-- There is no way to know at what time the query was fetched on the server, so `dataUpdatedAt` and determining if the query needs refetching is based on when the page loaded instead
+## A quick note on Suspense
 
-### Using Hydration
+This guide uses the regular `useQuery` API, but as long as you are prefetching your queries you can replace this with `useSuspenseQuery` freely if you wish and things will work the same, but you get to use `<Suspense>` for loading states on the client. In the [advanced guide](../guides/advanced-ssr) we'll switch this around and use `useSuspenseQuery` as the default, as Suspenseful data fetching is necessary for the advanced patterns.
```
TkDodo, line 41: I don't think you can `useSuspenseQuery` on the server without a framework that understands suspense and streaming. For example, I don't think this works in next12 or even in next13 inside the pages directory (at least it didn't last I checked)

docs/react/guides/ssr.md
```diff
@@ -151,371 +153,376 @@ function Posts() {
 
 The setup is minimal and this can be a quick solution for some cases, but there are a **few tradeoffs to consider** when compared to the full approach:
 
-- If you are calling `useQuery` with the same query in multiple locations, you need to pass `initialData` to all of them
+- If you are calling `useQuery` in a component deeper down in the tree you need to pass the `initialData` down to that point
+- If you are calling `useQuery` with the same query in multiple locations, passing `initialData` to only one of them can be brittle and break when your app changes since. If you remove or move the component that has the `useQuery` with `initialData`, the more deeply nested `useQuery` might no longer have any data. Passing `initialData` to **all** queries that needs it can also be cumbersome.
 - There is no way to know at what time the query was fetched on the server, so `dataUpdatedAt` and determining if the query needs refetching is based on when the page loaded instead
+- If there is already data in the cache for a query, `initialData` will never overwrite this data, even if the new data is fresher than the old one
```
TkDodo, line 159: I would emphasize here that this happens on every client side transition. In the `getServerSideProps` example, if you navigate a way from the page and then back again, you would:

- fetch `posts` again on the server
- send them down to the client as `initialData`
- but there, it would be completely ignored.

I think this is the biggest downside of the `initialData` approach

docs/react/guides/ssr.md
```diff
@@ -151,371 +153,376 @@ function Posts() {
...
+## Using the Hydration APIs
 
-To support caching queries on the server and set up hydration:
+With just a little more setup, you can use a `queryClient` to prefetch queries during a preload phase, pass a serialized version of that `queryClient` to the rendering part of the app and reuse it there. This avoid the drawbacks above. Feel free to skip ahead for full Next.js pages router and Remix examples, but at a general level these are the extra steps:
 
-- Create a new `QueryClient` instance **inside of your app, and on an instance ref (or in React state). This ensures that data is not shared between different users and requests, while still only creating the QueryClient once per component lifecycle.**
-- Wrap your app component with `<QueryClientProvider>` and pass it the client instance
-- Wrap your app component with `<HydrationBoundary>` and pass it the `dehydratedState` prop from `useDehydratedState()`
+- In the framework loader function, create a `const queryClient = new QueryClient(options)`
+- In the loader function, do `queryClient.prefetchQuery(...)` for each query you want to prefetch, and await them
```
TkDodo, line 168:
```suggestion
- In the loader function, do `await queryClient.prefetchQuery(...)` for each query you want to prefetch, and await them
```

docs/react/guides/ssr.md
```diff
@@ -151,371 +153,376 @@ function Posts() {
...
+  - This can be done for each route, or at the top of the application to avoid boilerplate, see examples
 
-```bash
-npm i use-dehydrated-state
-# or
-pnpm add use-dehydrated-state
-# or
-yarn add use-dehydrated-state
-```
+> An interesting detail is that there are actually _three_ `queryClient` involved. The framework loaders are a form of "preloading" phase that happens before rendering, and this phase has it's own `queryClient` that does the prefetching. The dehydrated result of this phase gets passed to **both** the server rendering process **and** the client rendering process which each has it's own `queryClient`. This ensures they both start with the same data so they can return the same markup.
```
TkDodo, line 175:
```suggestion
> An interesting detail is that there are actually _three_ `queryClient`s involved. The framework loaders are a form of "preloading" phase that happens before rendering, and this phase has it's own `queryClient` that does the prefetching. The dehydrated result of this phase gets passed to **both** the server rendering process **and** the client rendering process which each has it's own `queryClient`. This ensures they both start with the same data so they can return the same markup.
```

docs/react/guides/ssr.md
```diff
@@ -151,371 +153,376 @@ function Posts() {
...
+// For Remix, rename this to loader instead
+export async function getServerSideProps() {
+  const queryClient = new QueryClient()
 
-export function Posts(props) {
-  const { data } = useQuery({
-    queryKey: ['posts'],
-    queryFn: getPosts,
-    initialData: props.posts,
+  const user = queryClient.fetchQuery({
```
TkDodo, line 412:
```suggestion
  const user = await queryClient.fetchQuery({
```

docs/react/guides/ssr.md
```diff
@@ -151,371 +153,376 @@ function Posts() {
...
+## Serialization
 
-### Streaming, Suspense and server-side fetching
+When doing `return { props: { dehydratedState: dehydrate(queryClient) } }` in Next.js, or `return json({ dehydratedState: dehydrate(queryClient) })` in Remix, what happens is that the `dehydratedState` representation of the `queryClient` is serialized by the framework so it can be embedded into the markup and transported to the client.
 
-Right now, you always have to `await` the data in the Server Component. In the future, the goal is to be able to _start_ prefetching in a Server Component but not block rendering, instead streaming markup and data to the client incrementally as it gets available. This is currently lacking support in both React and Query.
+By default, these frameworks only supports returning things that are safely serializable/parsable, and therefor does not support `undefined`, `Date`, `Map`, `Set`, `BigInt`, `Infinity`, `NaN`, `-0`, regular expressions etc. This also means that you can not return any of these things from your queries. If returning these values is something you want, check out [superjson](https://github.com/blitz-js/superjson) or similar packages.
```
TkDodo, line 471:
```suggestion
By default, these frameworks only supports returning things that are safely serializable/parsable, and therefor does not support `undefined`, `Date`, `Map`, `Set`, `BigInt`, `Infinity`, `NaN`, `-0`, `Error`, regular expressions etc. This also means that you can not return any of these things from your queries. If returning these values is something you want, check out [superjson](https://github.com/blitz-js/superjson) or similar packages.
```

docs/react/guides/ssr.md
```diff
@@ -151,371 +153,376 @@ function Posts() {
...
 
-Similarily, you _must_ currently prefetch the data in a Server Component if you want it to be server rendered. A `useQuery()` call even with the `suspense` option enabled will not fetch data on the server, only on the client. We hope to support this in the future, but exact details are still unknown.
+If you are using a custom SSR setup, you need to take care of this step yourself. Your first instinct might be to use `JSON.stringify(dehydratedState)`, but because this doesn't escape things like `<script>alert('Oh no..')</script>` by default, this can easily lead to **XSS-vulnerabilities** in your application. [superjson](https://github.com/blitz-js/superjson) also **does not** escape values and is unsafe to use by itself in a custom SSR setup (unless you add an extra step for escaping the output). Instead we recommend using a library like [Serialize JavaScript](https://github.com/yahoo/serialize-javascript) or [devalue](https://github.com/Rich-Harris/devalue) which are both safe against XSS injections out of the box.
```
TkDodo, line 473: TIL 😱

docs/react/guides/ssr.md
```diff
@@ -151,371 +153,376 @@ function Posts() {
...
-}
+```
+1. |> JS for <Feed>
+2.   |> getFeed()
+3.     |> JS for <GraphFeedItem>
+4.       |> getGraphDataById()
 ```
 
-### Client
+This is because with SPA's server rendering only works for the initial page load, not for any subsequent navigation.
```
TkDodo, line 513:
```suggestion
This is because with SPA's, server rendering only works for the initial page load, not for any subsequent navigation.
```

docs/react/guides/advanced-ssr.md
```diff
@@ -0,0 +1,390 @@
...
+
+You might want to read the [Server Rendering & Hydration guide](../guides/ssr) before this one as it teaches the basics for using React Query with SSR, and [Performance & Request Waterfalls](../guides/request-waterfalls) as well as [Prefetching & Router Integration](../guides/prefetching) also contains valuable background.
+
+We'll be using `useSuspenseQuery` here, so also check out the [Suspense guide](../guides/suspense).
+
+Before we start, let's note that while the `initialData` approach outlined in the SSR guide also works with Server Components, we'll focus this guide on the hydration APIs.
+
+## Server Components & Next.js app router
+
+We won't cover Server Components in depth here, but the short version is that they are components that are guaranteed to always run on the server, not only on the initial page view but **also on page transitions**. This is similar to how Next.js `getServerSideProps`/`getStaticProps` and Remix `loader` works, as these also always run on the server but while those can only return data, Server Components can do a lot more. The data part is central to React Query however, so let's focus on that.
```
TkDodo, line 16: Not sure how to word this but I think we should mention that server components **only** run on the server (and never on the client). It reads a bit weird if we do:

> that are guaranteed to always **only** run on the server, not only ...

docs/react/guides/advanced-ssr.md
```diff
@@ -0,0 +1,390 @@
...
+
+How do we take what we learned in the Server Rendering guide about [passing data prefetched in framework loaders to the app](../guides/ssr#using-the-hydration-apis) and apply that to Server Components and the Next.js app router? The best way to start thinking about this is to consider Server Components as "just" another framework loader.
+
+### A quick note on terminology
+
+So far in these guides, we've been talking about the _server_ and the _client_. It's important to note that confusingly enough this does not match 1-1 with _Server Components_ and _Client Components_. Server Components are guaranteed to always run on the server, but Client Components actually runs in both places. The reason for this being that they are also rendered during the initial _server rendering_ pass.
+
+One way to think of this is that even though Server Components also _render_, they happen during a "loader phase" (always happens on the server), while Client Components run during the "application phase", and that application can run both on the server during SSR, and in for example a browser.
```
TkDodo, line 24: Is it worth noticing that this is nextJs specific ? Other server component implementations, like [waku](https://github.com/dai-shi/waku), might choose to not render client component on the server at all.

Read that here:
 - https://x.com/dai_shi/status/1668134236589793281
 - https://x.com/dai_shi/status/1668171134691590144

docs/react/guides/advanced-ssr.md
```diff
@@ -0,0 +1,390 @@
...
+Next, we'll look at what the Client Component part looks like:
+
+```tsx
+// app/posts/posts.jsx
+'use client'
+
+export default function Posts() {
+  // This useSuspenseQuery could just as well happen in some deeper
+  // child to <Posts>, data will be available immediately either way
+  const { data } = useSuspenseQuery({ queryKey: ['posts'], queryFn: getPosts })
```
TkDodo, line 163: does this really work? I thought if we `useSuspenseQuery` in the app directory, it will enable streaming 🤔

TkDodo (commented):
wonderfully written ❤️

docs/react/guides/prefetching.md
Ephem: Good point, and nice find! I'll rewrite this to use a `useQuery` for prefetching like the other one (above the suspense one), and explain why. Long term this is another good argument for a `usePrefetchQuery` or similar.

docs/react/guides/ssr.md
Ephem: I think that as long as the data is already in the cache and fresh so that it doesn't suspend, it'll work? I should probably formulate that clearer though.

docs/react/guides/ssr.md
Ephem: Good point! I'll make this clearer.

docs/react/guides/ssr.md
Ephem: Not sure which part was the TIL, but I learned this the hard way. 😬 It was easily the most common React vulnerability back when custom SSR was more common (might still be).

Say you make a url-parameter part of the `queryKey` (like an id). Now anyone can craft a link that contains `?articleId=<script>BAD STUFF</script>` and spread it, and if that is not escaped and included in the markup for hydration, now that JS is executed in the visitors browser. Same with user generated data ofc unless it's validated/escaped in the backend.

docs/react/guides/advanced-ssr.md
Ephem: Very true, current wording is confusing. I'll reword it!

docs/react/guides/advanced-ssr.md
Ephem: Yeah, it's not really Next.js specific, and it currently says "can run", not "does run", but I definitely agree this could be clearer. I'll add a note.

docs/react/guides/advanced-ssr.md
Ephem: If we have already awaited the data so it's in the cache and it's not stale, we shouldn't suspend so it should work? But I do think we need to add `staleTime` here and point that out.

As an aside we should probably have a full codesandbox example with this pattern (prefetching+suspense), either for the pages router, the app router or both.

Ephem (commented):
Super valuable feedback @TkDodo, thanks so much! I'll tackle it right away.

docs/react/guides/ssr.md
TkDodo: > Not sure which part was the TIL

everything really 😅

docs/react/guides/advanced-ssr.md
Ephem: Hmm, yeah, it's a problem. It's like if this is a weird middle ground between `useQuery` and `useSuspenseQuery`, where neither fits perfectly. It _should_ never suspend, it should always have data, it should never have the loading state. I'll think some more but maybe it is best to just do `useQuery` here for now.

Ephem: I just went ahead and changed it out to `useQuery` and did some tweaks in other places to reflect this. Can revisit later if we want to but I agree `useQuery` probably makes the most sense right now.

Ephem:
I think I managed to fix all your feedback @TkDodo, I resolved the minor stuff directly but left the comments about clarifications/reformulations open for you to look at. 😄

docs/react/guides/advanced-ssr.md
```diff
@@ -0,0 +1,389 @@
...
+
+If you do use it, a good rule of thumb is to avoid `queryClient.fetchQuery` unless you need to catch errors. If you do use it, don't render its result on the server or pass the result to another component, even a Client Component one.
+
+From the React Query perspective, treat Server Components as a place to prefetch data, nothing more.
+
+Of course, it's fine to have Server Components own some data, and Client Components own other, just make sure those two realities don't get out of sync.
+
+## Streaming with Server Components
+
+The Next.js app router automatically streams any part of the application that is ready to be displayed to the browser as soon as possible, so finished content can be displayed immediately without waiting for still pending content. It does this along `<Suspense>` boundary lines. Note that if you create a file `loading.tsx`, this automatically creates a `<Suspense>` boundary behind the scenes.
```
TkDodo, line 329: I've seen that we get an infinite fetch loop on the client if we don't have a suspense boundary with our experimental streaming plugin.

Not sure if it works with a `loading.tsx`, but maybe we should just recommend to always set a suspense boundary? you'd need it for client transitions anyways I guess!

Ephem: Good idea mentioning that at least for the streaming plugin, I added a note there. I think it's harder for us to have recommendations for Suspense boundaries in general though because they are so framework and application specific. It really matters where they go etc. I don't you necessarily need need one for the general approach described on this page, the reason the streaming lib is different is because it happens during the SSR phase, not the RSC phase.

PR #5970, opened 2023-09-06
Feature(Devtools): make data of queries editable 

bastibuck:
It would be nice if one could edit data of queries directly inside the devtools without changing the actual return values of a potential mock or real backend running during development. 

For example if I wanted to test and develop my UI with an empty list I could just remove all entries from a already fetched list instead of changing the server code to return an empty list, restart that thing and refetch my queries.

I therefor created this very early (visual) draft of making data editable inside devtools.

image

Will potentially need some guidance on how to best get the activeQuery into the right places to set the data from there onChange/onClick. 

Open for suggestions and ideas :) 

PS: focusing on string fields first, other types like numbers can follow once the functionality is in place.


bastibuck:
That would make updating way easier. I'll take a look on Sunday. 

Was currently struggling to identify the exact position in the objects and/arrays to update with my approach.

I bet it can be done. But as a first take I'd follow your idea. This would also allow for resetting the manually edited data back to what the cache originally had if we treat the manual edits as overrides.

bastibuck:
I've updated the code to use a JSON editor as an initial draft. The previos approach with the input fields and delete buttons is still in the PR but would be removed upon a descion on what path to follow.

image


## Concept and reasoning
It uses https://jsoneditoronline.org/ to make `data` editable and sets the activeQuery's data `onChange`.  

- following @vossmalte 's idea of making the whole JSON editable as text instead of single fields and entries this PR now does exactly that. This allows for very easy updates to the cached data.
- I've hidden the editor inside a modal because it might take up a low of space and is not as really useable inside the small panel where also the **Data Explorer** lives
- I've hidden all controls of the original JSON editor package to keep the interface simple and chosen `mode: "text"` for the easiest editing experience
- I've tested and only minimally adjusted colors to match the theme as a proof of concept for theming

## Open issues
Right now the editor doesn't have the greatest UX but it works. 

The `onChange` event fires fairly quick and the editor then loses focus because it is re-initialized which makes changing content kind of hard and not a joy.
This could be avoided when using a stable editor instance and updating the content onChanges as well. 

But before grinding out the edges I'd like to get feedback if this is a path we should continue. @TkDodo @vossmalte what's you opinion on this approach? 


### Next steps

- [ ] avoid re-initialization onChange
  - [ ] if successful check if UX is still as bad due to onChange triggering too quickly. If so use another event or a submit button or something
- [ ] will `data` always be somehow editable in JSONEditor?
- [ ] error handling on query-base (only enable modal on successful queries?)
- [ ] error handling for JSON content (show validation result somehow)
- [ ] align styling with tanstack-query
- [ ] use betters icons for open/close buttons
- [ ] do we need a "reset"? We could trigger a refetch or store the original data somewhere
- [ ] clean-up ts-ignore and casting of `content`

ardeora:
Hey there! Great work! Taking a look at this. One thing to keep in mind also is that TanStack Query is not opinionated on what the query data should be. So it's not necessary that all query data will always be JSON serializable. A query function could also return a Map, Set, Iterable, etc. 

@TkDodo in such cases should the editor be disabled?

TkDodo:
we're serializing those values with `superjson`, not sure if it makes sense to allow editing them or how hard it would be

TkDodo:
> Serialize data as a (nicely formatted) string and display it in a textarea. 

I think I liked the inline editing better 🤔

bastibuck:
> I think I liked the inline editing better 🤔

Me too! 

I just pushed a new version where it's possible to update string values. How do you like this approach? 

The types for the new util function are still wonky and `@ts-ignored` for the first draft. 
Also updating `number` values will follow along with some polishing.

bastibuck:
> I think I liked the inline editing better 🤔

Me too! 

I just pushed a new version where it's possible to update string values. How do you like this approach? 

The types for the new util function are still wonky and `@ts-ignored` for the first draft. 
Also updating `number` values will follow along with some polishing. 

Additionally coming back to the original idea of this PR I made arrays clearable in the latest commit. Don't think it makes a lot of sense to add something similar to other types than array. What do you think?

packages/query-devtools/src/__tests__/utils.test.ts
```diff
@@ -0,0 +1,139 @@
...
+        title: 'Spaghetti bolonese',
+        id: 2,
+        ingredients: ['spaghetti', 'tomato sauce', 'minced meat'],
+      },
+    ]
+    const newData = deleteNestedDataByPath(oldData, ['1', 'ingredients', '0'])
+
+    expect(newData).not.toBe(oldData) // should not be the same reference
+
+    expect(newData).toMatchInlineSnapshot(`
```
bastibuck, line 116: Do you like the inline snapshot for this kind of test? IMO it makes it super clear what happens and makes additional tests obsolute (e.g. check all array items and properties manually). If you like it I'd use it in other tests too

packages/query-devtools/src/Explorer.tsx
```diff
@@ -265,13 +345,48 @@ export default function Explorer(props: ExplorerProps) {
...
+
+                const newData = updateNestedDataByPath(
+                  oldData,
+                  currentDataPath,
+                  type() === 'number'
+                    ? changeEvent.target.valueAsNumber
+                    : changeEvent.target.value,
+                )
+
+                props.activeQuery?.setData(newData)
```
TkDodo, line 373: when calling `setData` like this, we should pass `options` to it as well:

- `manual: true` to avoid messing with `fetchStatus`
- `dataUpdatedAt` should probably be set to the current `dataUpdatedAt` of the query? Otherwise, every write would re-set staleTime counters

packages/query-devtools/src/utils.tsx
```diff
@@ -111,3 +111,56 @@ export const convertRemToPixels = (rem: number) => {
...
+export const updateNestedDataByPath = (
+  oldData: unknown,
+  updatePath: Array<string>,
+  value: unknown,
+): any => {
+  if (updatePath.length === 0) {
+    return value
+  }
+
+  // @ts-ignore
```
TkDodo, line 131: if we have to keep these, please prefer `@ts-expect-error`, because it will type-error if there is nothing to ignore.

packages/query-devtools/src/Explorer.tsx
TkDodo: actually, the `dataUpdatedAt` might not be necessary. I think its fine to treat the data you write in the devtools as `fresh`. Otherwise, refetches might overwrite it too eagerly. But the `manual: true` is something we should do.

If we want to use a different api, it would be:

```
queryClient.setQueryData(props.actieveQuery.queryKey, newData)
```

because that function sets `manual: true` automatically. It's the more "public" api

bastibuck: Nice updated the code accordingly

packages/query-devtools/src/utils.tsx
bastibuck: I updated the comments to use `@ts-expect-error` for now. 

We should probably come back to the issue @ardeora raised earlier though:

> A query function could also return a Map, Set, Iterable, etc.

Currently the two util functions are only covering arrays and objects properly. How can we handle other instaces? Can't find a good solution right now. Any idea?

bastibuck:
@TkDodo @ardeora Added the missing handling for Maps and Sets. 

Now the handling (deleting and updating) of text and number values inside Arrays, Objects (plain), Sets and Maps is done explicitly and per case. Very happy with how this turned out.

Anything else to address? 

The remaining `@ts-expect-error` cases are somewhat hard to get rid off. If you have any idea let me know.

TkDodo (approved):

PR #5979, opened 2023-09-08
feat: suspense error handling

TkDodo:
only hard errors are thrown to ErrorBoundaries

### DIFF
--- a/docs/react/guides/suspense.md
+++ b/docs/react/guides/suspense.md
@@ -23,7 +23,32 @@ const { data } = useSuspenseQuery({ queryKey, queryFn })
 
 This works nicely in TypeScript, because `data` is guaranteed to be defined (as errors and loading states are handled by Suspense- and ErrorBoundaries).
 
-On the flip side, you therefore can't conditionally enable / disable the Query. `placeholderData` also doesn't exist for this Query. To prevent the UI from being replaced by a fallback during an update, wrap your updates that change the QueryKey into [startTransition](https://react.dev/reference/react/Suspense#preventing-unwanted-fallbacks).
+On the flip side, you therefore can't conditionally enable / disable the Query. This generally shouldn't be necessary for dependent Queries because with suspense, all your Queries inside one component are fetched in serial.
+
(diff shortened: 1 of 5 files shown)


docs/react/guides/suspense.md
```diff
@@ -23,7 +23,32 @@ const { data } = useSuspenseQuery({ queryKey, queryFn })
...
+Since you can't change `throwOnError` (because it would allow for `data` to become potentially `undefined`), you have to throw errors manually if you want all errors to be handled by Error Boundaries:
+
+```tsx
+import { useSuspenseQuery } from '@tanstack/react-query'
+
+const { data, error } = useSuspenseQuery({ queryKey, queryFn })
+
+if (error) {
+    throw error
+}
```
Ephem, line 47: I know we discussed this in the issue, but reading this again today this stood out to me. Something feels a bit "weird" about data and error both being defined at the same time. Maybe even feels a bit surprising that `useSuspenseQuery` can return error at all?

Maybe it's a pain to implement, but one thought that popped into my head was calling it something like `refetchError` instead? That would also make the purpose of the rethrowing clearer:

```tsx
if (error) {
    throw error
}
```

TkDodo: > Something feels a bit "weird" about data and error both being defined at the same time.

But that's just how RQ works 😅 . In `error` state, you can potentially have `data` defined, because we never remove data that was once fetched.

> Maybe it's a pain to implement, but one thought that popped into my head was calling it something like refetchError instead?

I think that would have more downsides than upsides, because it would be a different API to learn, and its also not how we have things internally.

Ephem: > But that's just how RQ works 😅

Haha, yeah, true, I buy that argument. 😁 I do think it's a tiny bit unintuitive that error is named error just like in `useQuery` when it's only ever available for refetches though. I imagine people migrating from `useQuery` to `useSuspenseQuery` and just keeping their error-handling logic in components as is because the types tells them that's fine, but now that's only applicable to refetches which could be surprising.

Probably not a huge issue though, and I buy the counterarguments too, just thinking out loud. 😄

PR #5989, opened 2023-09-11
Hydrate existing queries in a useEffect

### LINKED ISSUE #4691: hydration
Hydration implementation details (major)
Step 1: Should still happen in render for new queries (and queries with no observers?), but be scheduled for a useEffect for queries that someone is actively listening to since this results in re-renders/observable side effects
Step 2 (optional): When hydrating in useEffect also add the new data to a context for that subtree, refactor all reads to patch in data from that context. That subtree safely gets the new data during the same render instead of after, "true" cache and thus anything listening outside of that subtree gets updated via the useEffect (This might be overkill, we should evaluate if step 1 is good enough first)

@Ephem

Ephem:
Closes #4691

I think this PR is finished. I've updated one test and added a new one and I've verified it fixes the codesandbox in the issue/below. I have _not_ verified it against any larger existing applications however (since I don't currently have access to any) which we should make sure to do during the beta/RC stage.

# Hydration timing

React Query currently hydrates all queries inside of render. This is considered safe for new queries, because those have no _observable_ side effects since they have no existing observers, but for queries that already exist in the cache, this isn't safe to do. For example, when hydrating during a React transition, existing observers currently update their data immediately, even though the transition has not yet finished, or could be aborted.

For an example, see this sandbox: https://codesandbox.io/s/hydrate-in-render-demonstration-uvpe46

This PR changes this so that _new_ queries are still hydrated during the render phase, which is important so SSR still works, and so we can prerender as much as possible during a transition, but makes it so any query that already exists in the cache is not hydrated until in a useEffect that happens _after_ the transition has committed.

This means the new page will first render with existing data+new data, to immediately rerender to show only new data when the effect has run. 

While this doesn't change the API in any way it's technically a breaking change, though I suspect and hope very few will notice any difference, especially since most people currently only use `HydrationBoundary` once during the initial page load and since all queries are new at that point, that behaviour should not have changed.

# Hydrating mutations

Mutations are no longer hydrated by `HydrationBoundary` as this isn't safe to do in the regular component lifecycle. The persist plugin already calls `hydrate()` directly and we are not aware of any usecases were you'd want to pass a paused mutation from the server to the client. If there are, this would still be possible by calling `hydrate()` directly with the mutations.

I thought about splitting `hydrate()` and `hydrateMutations()` but decided to keep it as is.


TkDodo:
> I think this also means it's impossible(?) to implement a perfectly safe way to hydrate them inside of the component lifecycle.

I agree, and I don't think we need to hydrate them during render at all.

> In the long run, we should probably remove mutations from hydrate and instead add a new hydrateMutations. We would also remove hydrating mutations from HydrationBoundary and let the developer call the hydrateMutations directly.

This is fine for SSR. The biggest case for hydrating mutations is restoring them from localstorage (persistQueryClient plugin). There, we call `hydrate` at the moment:

https://github.com/TanStack/query/blob/1d3ee672d2a1536e3463243813c91902aa97d973/packages/query-persist-client-core/src/persist.ts#L93

But we could just as well call `hydrate` and `hydrateMutations` separately.

The Question is: why not do this change right here? It's technically a breaking change to remove hydrating mutations from `HydrationBoundary`, so I wouldn't want to wait until v6 with it, but it would be a good cleanup...

Ephem:
Oh, I didn't know that was already being called separately there! I can't imagine why you'd want to hydrate paused mutations in the `HydrationBoundary` in that case, I'll just go ahead and remove it and see about splitting the functions too.

Ephem:
I went ahead and removed the hydration of mutations from `HydrationBoundary`, but I decided to keep `hydrate()` as is. 

The fact that the queries part of that needs to be idempotent is an internal detail and if you are calling `dehydrate()` and `hydrate()` explicitly, it makes sense that `hydrate()` acts on the whole dehydrated state, not just part of it.

packages/react-query/src/HydrationBoundary.tsx
```diff
@@ -3,7 +3,11 @@ import * as React from 'react'
...
 import { hydrate } from '@tanstack/query-core'
 import { useQueryClient } from './QueryClientProvider'
-import type { HydrateOptions, QueryClient } from '@tanstack/query-core'
+import type {
+  DehydratedState,
+  HydrateOptions,
+  QueryClient,
+} from '@tanstack/query-core'
 
 export interface HydrationBoundaryProps {
```
TkDodo, line 12: `HydrationBoundaryProps` don't seem to have changed, which means I can still pass `options.defaultOptions.mutations`, which doesn't make much sense.

I think we could `omit` this on type level

packages/react-query/src/HydrationBoundary.tsx
```diff
@@ -19,19 +23,83 @@ export const HydrationBoundary = ({
   queryClient,
 }: HydrationBoundaryProps) => {
   const client = useQueryClient(queryClient)
+  const [hydrationQueue, setHydrationQueue] = React.useState<
+    DehydratedState['queries'] | undefined
+  >()
 
   const optionsRef = React.useRef(options)
   optionsRef.current = options
```
TkDodo, line 31: nit: I would put this in a useEffect that always runs, as writing to refs during render is technically also not pure

```suggestion
  React.useEffect(() => {
    optionsRef.current = options
  })  
```

packages/react-query/src/HydrationBoundary.tsx
TkDodo, line 1: maybe a bit late for the naming question, but isn't this now a `QueryHydrationBoundary` ?

Ephem: It is technically and I'm open to it, but I don't think most people even associate mutations as something that is possible to hydrate (in the SSR context where you're using this), so I think it's fine? I don't think we gain much clarity by calling it that? I don't envision us getting questions about "why isn't the `HydrationBoundary` hydrating my mutations?" so to speak. 😄

packages/react-query/src/HydrationBoundary.tsx
Ephem: Yeah, I've been staring at that ref trying to understand it. I'm not sure if I wrote it at the beginning or if it happened later, but if the purpose is to not retrigger hydration just because the options have changed, we can do that by just ignoring it in the dep arrays (rarely great, but better than a ref in that case).

With `useEffect` there's also this new case where another fiber could render this component with other options and that would translate over to the effect in "this" fiber (this is one unsafe part about refs I guess).

If we just use the prop as is, and add it to the dep arrays, that would require users to memoize it or it would cause rerunning hydration on every render.

Doing it in an effect like this means new options wont be applied to the hydrations that happen in render though, so not sure that's great either?

TkDodo: > we can do that by just ignoring it in the dep arrays

wouldn't do that. if options contain functions that could closure over something from a previous render, you'll have a stale closure problem.

I got the pattern to write to often changing objects or functions into a ref and update them in an effect from Kent, see: https://epicreact.dev/the-latest-ref-pattern-in-react/

we do the same e.g. here:

https://github.com/TanStack/query/blob/1d3ee672d2a1536e3463243813c91902aa97d973/packages/react-query/src/useMutationState.ts#L60-L62

Ephem: Yeah, it's a nice pattern sometimes, but that case in `useMutationState` works quite differently than this does. Here, if we used that pattern, the `optionsRef` wouldn't update until _after_ render, but we need the updated value inside of render when we are hydrating new queries.

First render:

```tsx
<HydrationBoundary state={queryAState} options={{ defaultOptions: { queries: { gcTime: 10000 } } }} />
```

On server components navigation:

```tsx
<HydrationBoundary state={queryBState} options={{ defaultOptions: { queries: { gcTime: 1000 } } }} />
```

Because `queryBState` is not already in the cache, it will get hydrated inside of render with the old 10000 value, since options and `gcTime` has not had a time to change to the new value which would now happen in an effect.

packages/react-query/src/HydrationBoundary.tsx
```diff
@@ -19,19 +23,83 @@ export const HydrationBoundary = ({
...
+  // the new data on the current page, only _after_ the transition is committed.
+  // If the transition is aborted, we will have hydrated any _new_ queries, but
+  // we throw away the fresh data for any existing ones to avoid unexpectedly
+  // updating the UI.
   React.useMemo(() => {
     if (state) {
-      hydrate(client, state, optionsRef.current)
+      if (typeof state !== 'object') {
+        return
+      }
```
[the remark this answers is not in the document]
Ephem: I deleted the previous comment because I was being tired and stupid and gave bad examples.. 😄 (A version of that example has actually happened to me, it was just a bit more involved)

I included this `if` because it's a simple check that stands between crashing the entire app hard or not. As long as it's an object, even if `.queries` isn't available on there we'll fail gracefully and the client can fetch any queries that were meant to be hydrated but weren't. This `if` just takes that graceful fallback to non-object values. Sure, that doesn't protect from `.queries` having a bad format, but that should be more unlikely to happen by accident so I don't think we need runtime validation for that.

On the other hand, it should hopefully still be uncommon that state is truthy but not an object too and crashing hard also gives a better chance to catch it in dev so I don't feel strongly about it, I can go ahead and remove it, can always be added later if it shows up in practice.

packages/react-query/src/HydrationBoundary.tsx
TkDodo: ah yes, because we hydrate in useMemo as well. As usual, you are right 👍

packages/react-query/src/HydrationBoundary.tsx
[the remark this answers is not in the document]
TkDodo: fair enough 👍

packages/react-query/src/HydrationBoundary.tsx
Ephem: Nice catch, should be fixed now.

PR #5991, opened 2023-09-11
feat: validate queryOptions

TkDodo:


packages/react-query/src/__tests__/queryOptions.types.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+import type { Equal, Expect } from './utils'
+
+describe('queryOptions', () => {
+  it('should not allow excess properties', () => {
+    doNotExecute(() => {
+      // @ts-expect-error stallTime does not exist
+      return queryOptions({
+        queryKey: ['key'],
+        queryFn: () => Promise.resolve(5),
+        stallTime: 1000,
```
TkDodo, line 12: this works lovely, thanks @Andarist ❤️

packages/react-query/src/__tests__/queryOptions.types.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+  it('should infer types for callbacks', () => {
+    doNotExecute(() => {
+      return queryOptions({
+        queryKey: ['key'],
+        queryFn: () => Promise.resolve(5),
+        staleTime: 1000,
+        select: (data) => {
+          const result: Expect<Equal<number, typeof data>> = true
+          return result
+        },
```
TkDodo, line 25: @Andarist one more question: I found out that `data` here is inferred as `undefined` rather than `number`, so this test fails. Do you know why?

TkDodo: note: this is not introduced with the change in this PR - this test would even fail on the current `beta` branch.

However, this worked before #5805

Andarist: Hm, I see `unknown` here and not `undefined`

Andarist: This is the case of intra-expression inference limitation. What was introduced in TS 4.7 - [Improved Function Inference in Objects and Methods](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#improved-function-inference-in-objects-and-methods) - didn't fix issues like this.

We can reduce this failure to this [TS playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgPIAczAPYgM4A8AKgCJxhwA0yRAyhADYQKQAmAfMgN4BQyyARwCu0AJ4AxEAC5kACgCUyALydS5OAG4+yPI2ZgZs1uplqKilTXpMWEVloC+PHq2YM4UFDCEgWOEIIiUKIYWLiE2mZUkdb6dpSRof54yBAAHpAgrClJ4cRkFNR0erYcCeyy2JjJprn48rXV4Vo8wmJ1eLK8-G3BkoYWnACMAAwjCfy6NgZyxubKnHNwAHRg2LRgUKAA5goJDvJaQA)

So inferring `TOptions` as a whole might not be best here after all.

Andarist: A workaround for this could be using an intersection like this: [TS playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgPIAczAPYgM4A8AKgCJxhwA0yRAyhADYQKQAmAfMgN4BQyyARwCu0AJ4AxEAC5kACgCUyALydS5OAG4+yPI2ZgZs1uplqKilTXpMWEVloC+PHq2YM4UFDCEgWOEIIiUKIYWLiEZlRWeras1ESh-ni0ABZw6BDsstrYmEkyieHEZBTx1vp2nABk3MgA2gDSyKDIANYQotgwNIX4qekQALoyPq0g2ADuAQ6UPPIyvPyRpiWa2nQxbKblsVpLvclpGaYH-RmOWjwI4WDNIDDQnqwAShB4QgxgeMqBYgeyi1+wUkhgsnAAjAAGSGzfi6GwGOTGczKTjIuAAOjA2FoYCgoAA5gpZg55JdQA8oE9Xu9PngeAB6BnIAB6AH4gA). Based on the inferred results it should be roughly possible to reconstruct the type of the input options.

TkDodo: > Hm, I see unknown here and not undefined

yes, sorry, I meant `unknown` 🙈 

> We can reduce this failure to this [TS playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgPIAczAPYgM4A8AKgCJxhwA0yRAyhADYQKQAmAfMgN4BQyyARwCu0AJ4AxEAC5kACgCUyALydS5OAG4+yPI2ZgZs1uplqKilTXpMWEVloC+PHq2YM4UFDCEgWOEIIiUKIYWLiE2mZUkdb6dpSRof54yBAAHpAgrClJ4cRkFNR0erYcCeyy2JjJprn48rXV4Vo8wmJ1eLK8-G3BkoYWnACMAAwjCfy6NgZyxubKnHNwAHRg2LRgUKAA5goJDvJaQA) and the tracking issue for this in the TS repo would be https://github.com/microsoft/TypeScript/issues/47599

Thanks, I've subscribed to the issue.

> A workaround for this could be using an intersection like this:

Hmm, I don't really know how to apply this workaround to our real `queryOptions`

TkDodo: as discussed, bypassing for now; can be fixed in a follow up

https://github.com/TanStack/query/pull/5991/commits/a70b43643d50e407fffad40af9ef6c84550f704a

Issue #5968, opened 2023-09-06
Race condition near query cancellation/invalidation while implementing optimistic update


KartheekJavvaji:
### Describe the bug

**Requirements**: 
Assume the application has multiple entities present such as docs, slides, etc. We should now add a bookmarking ability to the system.

**Implementation**: 
I have implemented an optimistic update approach near bookmarking an entity in the application and have also made use of the abort signal for query cancellations.

**Issue**: 
Let's say I bookmark certain docs while on the docs page and quickly move to the bookmarks page, I see that the bookmarks list API-related query is getting canceled and is stuck in isFetching: true. Because of this, we are left with the entries added by the UI optimistic update and not the ones from the server.

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/loving-forest-2t69p7

### Steps to reproduce

In the example, we have 2 routes -> `Add Items` and `Item List`. I have added a mock service that mocks an API call with a ~1-second delay.

I differentiate the UI-added entry and the mock service-added entry by having a different pattern for IDs. All the entries added through optimistic updates will have a prefix "**\_\_UI\_\_** " in their IDs. All the entries from the mock service will have the prefix "**From Service**" in their IDs.

1. Go to the **Add Items** page and add 3-4 items by pressing the _Add Item_ button again and again before even the re-fetch can finish (this is to trigger the query cancellation). We can see the optimistic update-related logic at `src/useItems.ts`.
2. After the last addition of the items, move to **Item List** before the fetch can finish
3. We can now see that the last re-fetch that is triggered due to the query invalidation written in the onSettled method is getting canceled. And now we end up with entries added by UI optimistic update and not the actual server state.

### Expected behavior

How do we ensure we can always have the server state in this scenario?  I am guessing that component unmount is causing this query cancellation. Please watch the video attached as this race condition doesn't happen every single time. And if we give it enough time near the `Add Items` page, there won't even be any issue popping up. 

Without having the actual server state [actual IDs assigned by BE],  we won't be able to act upon those entries.

### How often does this bug happen?

Often

### Screenshots or Videos

https://github.com/TanStack/query/assets/12635717/b4580697-387b-468b-aebe-4ba00c8b5809

### Platform

- macOS 13.2
- Chrome 112.0.5615.137

### Tanstack Query adapter

None

### TanStack Query version

4.32.6

TkDodo:
[...]

PR #5997, opened 2023-09-12
fix: avoid reverting state to a fetching one

Andarist:
fixes https://github.com/TanStack/query/issues/5968 , cc @TkDodo @KartheekJavvaji

### DIFF
--- a/packages/query-core/src/tests/query.test.tsx
+++ b/packages/query-core/src/tests/query.test.tsx
@@ -912,8 +912,7 @@ describe('query', () => {
     unsubscribe()
   })
 
-  // TODO: use a better title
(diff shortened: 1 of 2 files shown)


TkDodo (approved):

packages/query-core/src/tests/query.test.tsx
```diff
@@ -911,4 +911,57 @@ describe('query', () => {
 
     unsubscribe()
   })
+
+  // TODO: use a better title
+  test('#5968', async () => {
```
TkDodo, line 916:
```suggestion
  test('should always revert to idle state (#5958)', async () => {
```

PR #5998, opened 2023-09-12
Disallow `undefined` in the `queryFn` return typings

braeden:
We explicitly disallow `undefined` as a return type since the introduction of react-query v4 (https://tanstack.com/query/v4/docs/react/guides/migrating-to-react-query-4#undefined-is-an-illegal-cache-value-for-successful-queries). And we're explicitly loud on development when this happens.

https://github.com/TanStack/query/blob/1614c31f84744c62b5bee8f2ff97fc15a1652e86/packages/query-core/src/query.ts#L457-L464

This will be a "type" break, but has not functional changes -- so we'll have to figure out how this reflects in semver. 

------

Hopefully after addressing internal compile breaks + adding some typing tests, we'll be able to resolve this question: https://github.com/TanStack/query/discussions/4457#discussioncomment-4059497

> > Also, this should make TS fail by not allowing undefined as a possible useQuery type, instead of throwing console errors on dev environment.

> yep, we tried that, but it's harder to do than it looks :)


TkDodo:
nice one, thank you 🙏 . I added two type tests and it looks good 👍 

> This will be a "type" break, but has no functional changes -- so we'll have to figure out how this reflects in semver.

Yeah, we usually don't treat type changes as breaking, so we could merge it to `main`. However, I would rather have that this PR goes to the `beta` branch (v5), because we'll do an rc release soon. Could you do that please?

packages/query-core/src/types.ts
```diff
@@ -14,7 +14,9 @@ export type QueryFunction<
   T = unknown,
   TQueryKey extends QueryKey = QueryKey,
   TPageParam = any,
-> = (context: QueryFunctionContext<TQueryKey, TPageParam>) => T | Promise<T>
+> = (
+  context: QueryFunctionContext<TQueryKey, TPageParam>,
+) => unknown extends T ? unknown : undefined extends T ? never : T | Promise<T>
```
TkDodo, line 19: we also have a `NonUndefinedGuard` in the repo that looks a bit simpler:

https://github.com/TanStack/query/blob/3f057a24778c15c55665d11fc8415200a20e4840/packages/react-query/src/queryOptions.ts#L13

braeden: This actually has a different meaning, it _only_ prevents the value `undefined` by itself. `undefined extends T` passes if the generic union contains `undefined`. 


I think they'll both correctly cause compile errors, but `undefined extends T` seems like it gives a clearer error message closer to the source. 

`T extends undefined`
image
`undefined extends T`
image
 


See: https://www.typescriptlang.org/play?#code/C4TwDgpgBAKgogD2BAdgEwM4FV0QGYCWKEaAPDAHxQC8sUESqmUArrocWlAPxTEBuEAE5QAXLADcAKFCQoADRqxEydNnZESpFCwC2AI2FQAPqw2cK0gPRWodgHrcpM8NBxp8mtCqYYY5Klo2Dw4SekY1Ol4BI3EYaVloAE0ld09OHzV-HQMjU2D0kkspGzsoRyA

But curious for your thoughts! (maybe we should be switching that existing implementation too?)

braeden:
> nice one, thank you 🙏 . I added two type tests and it looks good 👍
> 
> > This will be a "type" break, but has no functional changes -- so we'll have to figure out how this reflects in semver.
> 
> Yeah, we usually don't treat type changes as breaking, so we could merge it to `main`. However, I would rather have that this PR goes to the `beta` branch (v5), because we'll do an rc release soon. Could you do that please?

Sounds good, thanks for the CR + tests! Just changed the target branch & rebased onto `beta`. Will keep an eye on it to make sure stuff continues to build correctly with the new target. 

Do you know when the V5 is intended to have its first stable/non-rc release?

TkDodo:
> Do you know when V5 is intended to have its first stable/non-rc release?

should happen this year 😅

packages/query-core/src/types.ts
TkDodo: > maybe we should be switching that existing implementation too?

yeah that sounds like the right thing to do; I'll merge your PR and maybe you can drop another one that does this ? we have good test coverage for `initialData` typings

TkDodo: hmm can you take a look at the errors now after I merged in beta here? they are not on the beta branch ...

braeden: Yep, more than happy to follow up with the other change. 

Taking a look at the build failures! Might be a bit before I can figure out exactly what's wrong here. Something with `QueryPersister` isn't playing nice, I think from https://github.com/TanStack/query/pull/5125.

braeden: Okay @TkDodo it should be hopefully fixed with 4eba53bd6ee78cc2fe50fe7223d5b2fc506aaf3d -- also added tests about `null` return values + initialState. 

Couple of notes:
- I honestly don't understand how the fix worked...

```ts
unknown extends T ? unknown : undefined extends T ? never : T | Promise<T>
// vs
unknown extends T ? T : undefined extends T ? never : T | Promise<T>
```
- I think it made the error messages slightly worse 

image
- It also means that `data` ends up being inferred with the types, including `undefined` -- instead of `unknown` as it was previously. (because of this, I removed the type `Expect<>` and just left the `@ts-expect-error` annotations

image




Regardless of all of this weirdness, I still 100% think this is better than the existing state + avoids some nasty/unexpected runtime issues. But would also appreciate another quick smoketest from your side too!

TkDodo: wow, this looks really good. Thank you 🙏

TkDodo (approved):

TkDodo:
hmm we get a type error now in solid.js:

```
src/__tests__/createQuery.types.test.tsx(92,26): error TS2769: No overload matches this call.
  Overload 1 of 2, '(options: UndefinedInitialDataOptions<{ wow: boolean; }, Error, { wow: boolean; }, string[]>, queryClient?: (() => QueryClient) | undefined): CreateQueryResult<...>', gave the following error.
    Type '() => { wow: boolean; } | undefined' is not assignable to type 'undefined'.
  Overload 2 of 2, '(options: DefinedInitialDataOptions<{ wow: boolean; } | undefined, Error, { wow: boolean; } | undefined, string[]>, queryClient?: (() => QueryClient) | undefined): DefinedCreateQueryResult<...>', gave the following error.
    Type '() => { wow: boolean; }' is not assignable to type 'QueryFunction<{ wow: boolean; } | undefined, string[], never>'.
      Type '{ wow: boolean; }' is not assignable to type 'never'.
```

braeden:
> hmm we get a type error now in solid.js:
> 
> ```
> src/__tests__/createQuery.types.test.tsx(92,26): error TS2769: No overload matches this call.
>   Overload 1 of 2, '(options: UndefinedInitialDataOptions<{ wow: boolean; }, Error, { wow: boolean; }, string[]>, queryClient?: (() => QueryClient) | undefined): CreateQueryResult<...>', gave the following error.
>     Type '() => { wow: boolean; } | undefined' is not assignable to type 'undefined'.
>   Overload 2 of 2, '(options: DefinedInitialDataOptions<{ wow: boolean; } | undefined, Error, { wow: boolean; } | undefined, string[]>, queryClient?: (() => QueryClient) | undefined): DefinedCreateQueryResult<...>', gave the following error.
>     Type '() => { wow: boolean; }' is not assignable to type 'QueryFunction<{ wow: boolean; } | undefined, string[], never>'.
>       Type '{ wow: boolean; }' is not assignable to type 'never'.
> ```

Ohfff, spent a bit of time debugging today without much luck. Will play around with this later in the week and see if it feels like it's solvable 🤞.

TkDodo:
sorry for that, please re-open again `main`

PR #6001, opened 2023-09-13
fix: avoid implicit any for the param of placeholderData

Andarist:
cc @TkDodo


TkDodo:
weird, it still shows up as implicit any for me :/

Screenshot 2023-09-13 at 08 54 10

TkDodo:
> weird, it still shows up as implicit any for me :/

nevermind, I needed to re-build the project and restart the TS server. all good 👍

packages/react-query/src/__tests__/useQuery.types.test.tsx
```diff
@@ -156,3 +156,24 @@ describe('initialData', () => {
     })
   })
 })
+
+describe('placeholderData', () => {
+  it('should receive contextual parameter type', () => {
```
Andarist, line 161: This shows how this is an improvement in the general sense, and not just for `queryOptions`. This test would fail today: [TS playground](https://www.typescriptlang.org/play?ts=5.2.2#code/JYWwDg9gTgLgBAbzgVwM4FMCKz1QJ5wC+cAZlBCHAEQACMAhgHaoMDGA1gPRTr2swBaAI458VANxxOnODDxh0qAFxwArADoADFoEAjdA3UAmVQFgAUBdYRm8CGBjAbqOAF5EFuHBG48AaXQ8FQBtKhIICCoAXTh6F2tbABpPb1E8ADFGFQAKAEo3AD44AAVyEGAMdR5UCAAbADd0bNVc5PMvFnpa9AAVUHQVAEZNOAAqOGHNTTbCcQsLNCw07IQU9XX7R2c2rzBavnQACzqAE1wAEXoGHLAeeqc0S4Z81yLV9q84W-R7iEer+hzD5eaRwAB6AH4Ul4eDBkFBGF87g9UE9ASlCDNckCgA)

packages/query-core/src/types.ts
```diff
@@ -46,7 +46,13 @@ export type QueryFunctionContext<
...
 export type InitialDataFunction<T> = () => T | undefined
 
-type NonFunctionGuard<T> = T extends Function ? never : T
+declare const queryCoreFunctionBrand: unique symbol
+type NotAFunction<T> = T & { [queryCoreFunctionBrand]?: never }
+declare global {
+  interface Function {
+    [queryCoreFunctionBrand]?: true
+  }
+}
```
Andarist, line 55: This is definitely somewhat unorthodox... and I'm battling myself if I should do it in XState. This is the best way I could find to prevent a type parameter from being inferred as a function when your property can be `T | (() => T)`. I don't have a source to back up my claim - but *similar* things were often~ suggested by TS team members in their issue tracker. I wish there would be a better way to do this but well, it is what it is (?) 🤷‍♂️

Andarist:
The CI didn't come back clean :v I'll have to investigate this

Andarist:
actually... is this even a new error? 🤔 I can repro the same problem on [TS playground](https://www.typescriptlang.org/play?ts=5.2.2#code/JYWwDg9gTgLgBAbzgVwM4FMCKz1QJ4A0cAqhtrngErqrIA28AvnAGZQQhwBEAAjAIYA7VAIDGAawD0UdP1EwAtAEcc+LgG44kyXBh4wNAFxwArADoADJYUAjdALMAmEwFgAUO5bJB84BEFwqHTo6GAAFDCg6BDIMMaCyCB2UACUxgAK7CDAGAA8AG4QwAAmAHyI7nBwMjDIUAGC6ADucJkcOehhYTKoEHT56EQA+jIAVujyKXAAvOUIlVWB9gAqUTEw3TR9A0SRINGxKeoLjEfujO7uxRN0-DKs3r7+cOIhYJno+X5oACL8ArllqUwgswDIvjFUH8BMZlnAAD4oQTXFjARrFAjuNJwOGI7wotHoYrHDxuUT+ESBAQwIxwACCUCg-DwuVIWFUVBo9BguQSSVwpXK0zgAG0ALokq43O7oODk4TwV54YwiKBogDm4pJ11Et3u8sp5O8cTgfOSkrJFPgIn+suFaHZFDC8zcVRUFAA0uhlaKlUQjYIYGLMa64O78AAxQTGfioPA+OBhKazCqhqr8Jr8YDW4KhMIARgsZzT1XsdQCAZgJKqjBDVTAt1E6AAFn1rlBofxjK9Qh8Ib9-vwQ6cLTaaagzGA0M2wmP0MWgA)

packages/query-core/src/types.ts
TkDodo: do we even need the NonFunctionGuard? We also don't have this for `initialData`, which I think should be the same on type level

TkDodo:
it seems to be a real error because in your TS playground, `state` is inferred as `UseQueryResult<unknown, DefaultError>` instead of `UseQueryResult<number, DefaultError>` :/

Andarist:
> it seems to be a real error because in your TS playground, state is inferred as UseQueryResult<unknown, DefaultError> instead of UseQueryResult<number, DefaultError> :/

Right - and the local failure in this branch is related to the same fact. So I'm currently a little bit confused as to why this test would pass if on the playground it behaves roughly the same as locally here. I didn't have time for an in-depth analysis of the situation here though.

packages/query-core/src/types.ts
Andarist: It seems to behave slightly differently there - and I don't know why. I will investigate this but it might take some time before I get to it.

TkDodo:
sorry for that, please re-open again `main`

PR #6008, opened 2023-09-13
types(query-core): ensureQueryData TError default type

klarstrup:
I couldn't figure out the reason for this omission nor find any reason in a brief search of the repo for `ensureQueryData`, so I assume this is just an oversight.


TkDodo:
You're right, that's an oversight. Thank you 🙏

Issue #5959, opened 2023-09-05
devtools: "trigger loading" button changes state when "refetch" is clicked


TkDodo:
can be seen in the simple example in the docs:

https://tanstack.com/query/v5/docs/react/examples/react/simple

- click `refetch` 
- observe that while the fetch is going on, `trigger loading` switches to `restore loading`
- the button switches back to its normal state when the fetch finishes

![screencast 2023-09-05 11-35-46](https://github.com/TanStack/query/assets/1021430/61bbcc90-39b3-4735-b51e-8dd2078dad5d)

judicaelandria:
I think this is fixed, I can't reproduce this anymore

TkDodo:
@judicaelandria no work has been done on this, the `simple` example still shows this error (api is pretty fast, in the gif, I slowed it down)

judicaelandria:
I had a closer look and I think that's expected
because firing the `refetch` make the `statusLabel` update to fetching
more
https://github.com/TanStack/query/blob/2d41d98854b919168b4ac158983a59e2176efa67/packages/query-devtools/src/Devtools.tsx#L1037

or there is a context I'm missing (if not that's expected)

TkDodo:
I'd say its not expected that this happens. We have a separate state for restore loading:

https://github.com/TanStack/query/blob/2d41d98854b919168b4ac158983a59e2176efa67/packages/query-devtools/src/Devtools.tsx#L813

So I guess we should use that one ?

judicaelandria:
yes I think so too, I'll update the condition and will send a pr

PR #6010, opened 2023-09-14
fix(devtools): fix restore/trigger loading cond

judicaelandria:
This PR fix the button trigger loading state
closes https://github.com/TanStack/query/issues/5959


TkDodo (changes requested):
if I now click `trigger loading`, it doesn't change to `restore loading` either.

not sure why, maybe that state isn't what we're actually looking for 🤷 

@ardeora might be able to help here. Also, we're still missing the tests for the devtools from v4. @ardeora I could start adding them for the react adapter, but I think most tests should actually be for the agnostic `query-devtools` package, right? So I'd need to see one or two examples to get this going because I've never written a test in solid 😬

judicaelandria:
hmm that's weird, I'll have a closer look again to see if I can find the real state

judicaelandria:
@TkDodo I think it should work as expected now

ardeora (approved):
This is looking good! I can see the expected behavior here. One thing that could also be addressed in this PR maybe that when Trigger loading is clicked. We should disable `Trigger Error` button? Ideally we should be able to trigger an Error while in loading state, but currently seems way more complex to achieve this

ardeora:
> if I now click `trigger loading`, it doesn't change to `restore loading` either.
> 
> not sure why, maybe that state isn't what we're actually looking for 🤷
> 
> @ardeora might be able to help here. Also, we're still missing the tests for the devtools from v4. @ardeora I could start adding them for the react adapter, but I think most tests should actually be for the agnostic `query-devtools` package, right? So I'd need to see one or two examples to get this going because I've never written a test in solid 😬

I have been trying to add the tests. The only issue currently I'm stuck on is we use the `solid-transition-group` package to implement the enter/exit animations. The package itself depends on the `ResizeObserver` API which currently doesn't exist in jsdom so simulating an opening devtools causes error. Trying to find a workaround to that.

judicaelandria:
> We should disable `Trigger Error` button? Ideally we should be able to trigger an Error while in loading state, but currently seems way more complex to achieve this

yes, I'll disable to the button while in loading state

TkDodo:
looks good, but please also disable the `invalidate` and `reset` button, as they will also trigger a refetch. clicking them will make the `restore loading` not work

judicaelandria:
@TkDodo  fixed

TkDodo (approved):

Issue #5609, opened 2023-06-19
devtools: adding lots of style attributes


TkDodo:
seems like emotion will create a new css class every time we resize the devtools. For changing values, inline styles are probably better than css-in-js. @ardeora FYI
![screencast 2023-06-19 13-03-10](https://github.com/TanStack/query/assets/1021430/fea7086d-b0ab-4610-918a-eb42171c27d9)

PR #6032, opened 2023-09-18
fix(query-devtools): Fix styling issues in devtool

ardeora:
Fixes https://github.com/TanStack/query/discussions/5940
Fixes https://github.com/TanStack/query/issues/5609

This change brings a few styling changes to the core devtools

1. Switches [Emotion CSS](https://emotion.sh/docs/introduction) to [goober](https://goober.js.org/). 
  - Emotion in dev mode adds multiple style tags in dev mode. Since we use the devtools in dev mode primarily, there would be multiple style tags polluting the document
  - Goober has the exact same API like Emotion but has a way lighter footprint
  - All Goober devtools styles are added to a single style tag
2. Adds classnames to most elements of the devtools so that they can be targeted and styled with a custom stylesheet in user-land (#5940)
3. More specific styles that do not get overriden by vite default template
4. Adjusting height/width of the devtools do not change the class name so more performant (#5609)

### DIFF
--- a/packages/query-devtools/package.json
+++ b/packages/query-devtools/package.json
@@ -40,19 +40,21 @@
     "test:lib": "vitest run --coverage",
     "test:lib:dev": "pnpm run test:lib --watch",
     "test:build": "publint --strict",
-    "build": "tsup"
+    "build": "tsup",
+    "build:dev": "tsup --watch"
   },
   "files": [
     "build",
     "src"
   ],
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


TkDodo (approved):

TkDodo:
should we add documentation on how to overwrite custom styles?

also, are the options listed here still correct?

https://tanstack.com/query/v5/docs/react/devtools#options

I can see that `panelProps` etc has been removed, so we need to show how to do it now

PR #6040, opened 2023-09-20
Update dependent-queries.md

alewis729:
Fix typo


docs/react/guides/dependent-queries.md
```diff
@@ -71,7 +71,7 @@ const { data: userIds } = useQuery({
 // Then get the users messages
 const usersMessages = useQueries({
   queries: users
```
Yuddomack, line 73: In the context of the code above, shouldn't this be modified too?

```suggestion
  queries: usersIds
```

alewis729: Yeah, I think so, makes sense. I missed that bc I was fixated at the `userIds`. This would work as far as I see.

```js
const usersMessages = useQueries({
  queries:
    userIds?.map(id => ({
      queryKey: ['messages', id],
      queryFn: () => getMessagesByUsers(id)
    })) ?? [] // if userIds is undefined, an empty array will be returned
});
```

TkDodo:
seems like this is already fixed on main

PR #6043, opened 2023-09-21
fix(vue-query): avoid use sync watchers that cause frequent requests

Mini-ghost:


DamianOsipiuk:
@Mini-ghost could you get this to `rc` branch as well?

Mini-ghost:
@DamianOsipiuk Sure! I will give it a try in the next few days.

PR #6073, opened 2023-09-28
use `loading` instead of `pending` for mutations

KATT:
- I don’t get the terminology “pending” for mutations 
  - loading is equally correct 
  - loading is more widely used for this sort of thing
  - it's backward compatible
- It’s a refactor for thousands of apps or a small one here

Feel free just to kill this PR. It was just a search and replace :)


TkDodo:
we decided to keep the change to be aligned with `useQuery`, and to be aligned with new hooks and terminology from react. It's also going to be easier for new starters.

PR #6077, opened 2023-09-28
feat: add stable-query-client rule

Newbie012:
This PR introduces a new rule named `stable-query-client`. It ensures that the QueryClient is stable. It checks if the QueryClient is either extracted from the component or wrapped in React.useState. If not, it suggests a fix to wrap the QueryClient instantiation in a React.useState call. This rule is recommended to be set as an error.

### DIFF
--- a/packages/eslint-plugin-query/src/rules/stable-query-client/stable-query-client.rule.ts
+++ b/packages/eslint-plugin-query/src/rules/stable-query-client/stable-query-client.rule.ts
@@ -32,6 +32,7 @@ export const rule = createRule({
         if (
           node.callee.type !== AST_NODE_TYPES.Identifier ||
           node.callee.name !== 'QueryClient' ||
+          node.parent?.type !== AST_NODE_TYPES.VariableDeclarator ||
           !helpers.isSpecificTanstackQueryImport(
             node.callee,
             '@tanstack/react-query',
@@ -46,28 +47,13 @@ export const rule = createRule({
           return
         }
 
-        const declarator = ASTUtils.getClosestVariableDeclarator(node)
-
-        if (
-          declarator?.init?.type === AST_NODE_TYPES.CallExpression &&
-          ['useState', 'React.useState'].includes(
-            context.getSourceCode().getText(declarator.init.callee),
-          )
-        ) {
-          return
-        }
-
         context.report({
-          node: node.parent ?? node,
+          node: node.parent,
           messageId: 'unstable',
           fix: (() => {
             const { parent } = node
 
-            if (
-              parent === undefined ||
-              parent.type !== AST_NODE_TYPES.VariableDeclarator ||
-              parent.id.type !== AST_NODE_TYPES.Identifier
-            ) {
+            if (parent.id.type !== AST_NODE_TYPES.Identifier) {
               return
             }
 
--- a/packages/eslint-plugin-query/src/rules/stable-query-client/stable-query-client.rule.ts
+++ b/packages/eslint-plugin-query/src/rules/stable-query-client/stable-query-client.rule.ts
@@ -42,8 +42,12 @@ export const rule = createRule({
         }
 
         const fnAncestor = ASTUtils.getFunctionAncestor(context)
+        const isReactServerComponent = fnAncestor?.async === true
 
-        if (!ASTUtils.isValidReactComponentOrHookName(fnAncestor?.id)) {
+        if (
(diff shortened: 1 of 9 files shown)


packages/eslint-plugin-query/src/rules/stable-query-client/stable-query-client.rule.ts
```diff
@@ -0,0 +1,90 @@
...
+  create(context, _, helpers) {
+    return {
+      NewExpression(node) {
+        if (
+          node.callee.type !== AST_NODE_TYPES.Identifier ||
+          node.callee.name !== 'QueryClient' ||
+          !helpers.isSpecificTanstackQueryImport(
+            node.callee,
+            '@tanstack/react-query',
+          )
```
Newbie012, line 38: I was uncertain if this rule is relevant to non-react packages, so currently it works only if QueryClient is imported from `@tanstack/react-query`.

packages/eslint-plugin-query/src/utils/ast-utils.ts
```diff
@@ -203,27 +203,37 @@ export const ASTUtils = {
...
+      /^(use|[A-Z])/.test(identifier.name)
+    )
   },
   getFunctionAncestor(
     context: Readonly<RuleContext<string, readonly unknown[]>>,
   ) {
-    return context.getAncestors().find((x) => {
-      if (x.type === AST_NODE_TYPES.FunctionDeclaration) {
-        return true
+    for (const ancestor of context.getAncestors()) {
```
Newbie012, line 218: I changed from `.find` to `for` to get a better return type.

Newbie012:
One last thing, we should probably add a reference in the docs to this rule as well.

packages/eslint-plugin-query/src/rules/stable-query-client/stable-query-client.rule.ts
TkDodo: good call, I think e.g. for solid it doesn't matter.

packages/eslint-plugin-query/src/rules/stable-query-client/stable-query-client.rule.ts
```diff
@@ -0,0 +1,88 @@
...
+
+        if (!ASTUtils.isValidReactComponentOrHookName(fnAncestor?.id)) {
+          return
+        }
+
+        const declarator = ASTUtils.getClosestVariableDeclarator(node)
+
+        if (
+          declarator?.init?.type === AST_NODE_TYPES.CallExpression &&
+          ['useState', 'React.useState'].includes(
```
TkDodo, line 53: I guess it would also work inside `useRef`, but that's hard to find I guess?

```
const ref = useRef()
if (!ref.current) {
  ref.current = new QueryClient()
}
```

Newbie012: Oh, It's doable of course, but is this a real use case?

TkDodo: interestingly, there are lots of users who like to use refs for this. There's also a popular `useConstant` package that does it like that:

https://github.com/Andarist/use-constant/blob/bdd74e6348c14b7947c278fa5a5abdc03a635ac2/src/index.ts#L5-L13

come to think about it, maybe we should only error if `new QueryClient()` is called directly in the render function of a component, without being wrapped in anything? Because:

```
const client = useConstant(() => new QueryClient())
```

should also work.

Newbie012: Sounds logical. Even this is considered valid:

```
const client = React.useMemo(() => new QueryClient(), [])
```

I'll update the PR accordingly.

TkDodo: thanks. another thing I just realised: `new QueryClient()` is actually okay, and recommended, inside server components (async components). From our docs:

```
export default async function PostsPage() {

  await queryClient.prefetchQuery({
    queryKey: ['posts'],
    queryFn: getPosts,
  })
```

this is fine because server components never re-render, and if we'd put the QueryClient outside of the component, it would be shared between users and requests. Can we account for that?

Newbie012: Sure, that would make sense. I have updated the PR accordingly.

TkDodo (approved):

PR #6079, opened 2023-09-28
fix(vue-query): avoid sync watchers to reduce frequent requests

Mini-ghost:
The reason for this PR is the same as RP #6043


Mini-ghost:
This seems to have been merged from `main` to `rc`, so close it.

PR #6092, opened 2023-10-02
fix(vue-query): solve type error about `useQuery`

### LINKED ISSUE #6091: [vue-query] Query type breaks when using generics as a parameter of useQuery
### Describe the bug

When using a generic type in `useQuery`, the return value of `data` is not correctly inferred.

I got this after updating from 4.35.3 to 4.35.6. After changelog inspection, I believe it's related to this change:

PR: https://github.com/TanStack/query/pull/6043

Type change in `useBaseQuery`: https://github.com/TanStack/query/pull/6043/files#diff-996721944b77ab3155cb5f86364575311132aeaaff8f7f715755bfcc9fc09108R34-R38

I tried to reproduce in a Code Sandbox but couldn't get the environment right, so I did a Github Repro instead, apologies for the inconvenience.

For convenience, I'm pasting the reproduction code here:

```typescript
import { useQuery } from '@tanstack/vue-query'

let f: <T>() => T

interface MyType {
  a: number
}

const useTestRepro = <K extends MyType>() => {
  const myFetch = async (): Promise<K> => {
    return Promise.resolve(f<K>())
  }

  const query = useQuery([], myFetch)

  console.log(query.data.value.a)

  return query
}

export default useTestRepro
```

I'm not an expert in Typescript generics, so I think I need to inform Typescript that my generic does not have properties that are functions, but after trying for a while, I couldn't get it right. So it's either that or the new typing is wrong.

https://github.com/joaopslins/vue-query-repro

### Steps to reproduce

1. Checkout https://github.com/joaopslins/vue-query-repro locally, install dependencies with `npm install`

2. Open https://github.com/joaopslins/vue-query-repro/blob/master/src/composables/useTestRepro.ts in your IDE

3. Assert that `.value` property access throws an error

### Expected behavior

`.value` property access shouldn't throw an error

Every time

### Screenshots or Videos

image

### Platform

- OS: Mac
- IDE: RubyMine (Jetbrains)

### Tanstack Query adapter

vue-query

### TanStack Query version

v4.35.6

### TypeScript version

v5.2.2

### Additional context

_No response_

Mini-ghost:
### Linked issue

resolve #6091

### DIFF
--- a/packages/vue-query/src/useBaseQuery.ts
+++ b/packages/vue-query/src/useBaseQuery.ts
@@ -32,9 +32,13 @@ export type UseQueryReturnType<
   TError,
   Result = QueryObserverResult<TData, TError>,
 > = {
-  [K in keyof Result]: Result[K] extends (...args: any[]) => any
+  [K in keyof Result]: K extends
+    | 'fetchNextPage'
+    | 'fetchPreviousPage'
+    | 'refetch'
+    | 'remove'
     ? Result[K]
-    : ToRef<Result[K]>
+    : ToRef<Readonly<Result>[K]>
 } & {
   suspense: () => Promise<Result>
 }


joaopslins:
Thanks for the quick response about the issue!

I tried to reproduce the issue in the provided codesandbox link, and although `.value` is not throwing an error anymore, the type is lost. Maybe that wasn't intended?

Here's the link: https://codesandbox.io/p/sandbox/repro-92g5d4. Filename is `useTestRepro.ts`.

image
image

Mini-ghost:
Thank you! This may need confirmation because I placed your code into `@tanstack/vue-query` v4.35.3, and it still produces the results you're seeing.

joaopslins:
Oh, apologies, you are right! I confirmed in my local project and indeed it's a similar result in v4.35.3.

I would expect the type inference to work properly in this case, but I agree it might be outside the scope of this fix, and that the regressed bug is fixed 🎉 

Looking forward to have this released, thanks for the change 🙇

PR #6119, opened 2023-10-09
Feature/typed query key

TkDodo:


TkDodo:
@Andarist FYI :)

packages/react-query/src/queryOptions.ts
```diff
@@ -32,26 +32,26 @@ export function queryOptions<
   TError = DefaultError,
   TData = TQueryFnData,
   TQueryKey extends QueryKey = QueryKey,
-  TOptions extends UndefinedInitialDataOptions<
-    TQueryFnData,
-    TError,
-    TData,
-    TQueryKey
-  > = UndefinedInitialDataOptions<TQueryFnData, TError, TData, TQueryKey>,
->(options: ValidateQueryOptions<TOptions>): TOptions
```
TkDodo, line 41: removing this generic here made this test fail:

https://github.com/TanStack/query/blob/cc9bab714a5b520f37ac1460d9d564c5f0fda87f/packages/react-query/src/__tests__/useQuery.types.test.tsx#L48-L59

However, it made other things a lot easier 😅 . For example, I couldn't get ` DataTag<TQueryKey, TData>` to infer correctly - I had to re-extract the types from `TOptions['queryKey']` etc. Basically the changes from this commit:

- https://github.com/TanStack/query/pull/6119/commits/37f2a8bb4633ba2193e6eb14e262e0eaad584810

So, I'm considering keeping this as a known limitation, because the workaround for that situation would be to change `options` to take select as an arg:

```
const options = <TData = number,>(select?: (data: number) => TData) =>
  queryOptions({
    queryKey: ['key'],
    queryFn: () => Promise.resolve(1),
    select,
  })
useQuery(options((data) => data > 1))
```

which is not that bad. what do you think @Andarist ?

packages/react-query/src/__tests__/queryOptions.types.test.tsx
```diff
@@ -20,12 +25,156 @@ describe('queryOptions', () => {
...
+      doNotExecute(() => {
+        const { queryKey } = queryOptions({
+          queryKey: ['key'],
+          queryFn: () => Promise.resolve(5),
+        })
+
+        const queryClient = new QueryClient()
+
+        // @ts-expect-error value should be a number
+        queryClient.setQueryData(queryKey, '5')
```
TkDodo, line 171: @Andarist ideally, we would get an error here, because `queryKey` is tagged with `number`, but we pass in a `string` as value. However, I think it just takes the second overload and infers `string` because the first overload is not a match. Is there a way to make the first overload match and thus get an error here?

Note that it works fine if we use the functional updater form (test above this one), because `prev` will always be a `number` and we thus have to return a number as well.

PR #6122, opened 2023-10-09
feat(angular-query): add angular query

arnoud-dv:


TkDodo (commented):
awesome. is it okay if I add the missing glue code to make publishing work?

TkDodo:
Ah I also noted that `createInfiniteQuery` is missing - are you planning on adding that? It's basically the same as `createQuery`, except that it uses an `InfiniteQueryObserver`

arnoud-dv:
> awesome. is it okay if I add the missing glue code to make publishing work?

Ofcourse, go ahead :-)

arnoud-dv:
> Ah I also noted that `createInfiniteQuery` is missing - are you planning on adding that? It's basically the same as `createQuery`, except that it uses an `InfiniteQueryObserver`

Yes on the project I made Angular Query for we don't need infinite queries at the moment. I do intend to add it in my own time.

PR #6123, opened 2023-10-09
update to Vue Style type

roblabat:
My updates in Vue Style Types,

I also added a small fixe for sur select function issue, I'm not really satisfied of this fix, need new ideas or proposal

### DIFF
--- a/packages/query-core/src/queryClient.ts
+++ b/packages/query-core/src/queryClient.ts
@@ -11,7 +11,6 @@ import { focusManager } from './focusManager'
 import { onlineManager } from './onlineManager'
 import { notifyManager } from './notifyManager'
 import { infiniteQueryBehavior } from './infiniteQueryBehavior'
-import type { DataTag, NoInfer } from './types'
 import type { QueryState } from './query'
 import type {
   CancelOptions,
@@ -34,6 +33,7 @@ import type {
   RefetchQueryFilters,
   ResetOptions,
   SetDataOptions,
+  TypedQueryKey,
 } from './types'
 import type { MutationFilters, QueryFilters, Updater } from './utils'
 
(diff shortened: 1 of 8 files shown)


TkDodo:
I added some more tests on the base branch, I think you need to adapt this here as well?

```
> @tanstack/query-core@5.0.0-rc.5 test:types /home/runner/work/query/query/packages/query-core
> tsc

src/tests/queryClient.types.test.tsx(67,35): error TS2304: Cannot find name 'DataTag'.
src/tests/queryClient.types.test.tsx(70,7): error TS2578: Unused '@ts-expect-error' directive.
src/tests/queryClient.types.test.tsx(73,7): error TS2578: Unused '@ts-expect-error' directive.
src/tests/queryClient.types.test.tsx(78,13): error TS2322: Type 'true' is not assignable to type 'false'.
src/tests/queryClient.types.test.tsx(78,28): error TS2344: Type 'false' does not satisfy the constraint 'true'.
src/tests/queryClient.types.test.tsx(103,13): error TS2322: Type 'true' is not assignable to type 'false'.
src/tests/queryClient.types.test.tsx(103,28): error TS2344: Type 'false' does not satisfy the constraint 'true'.
 ELIFECYCLE  Command failed with exit code 2.
```

TkDodo:
nevermind, I fixed those issues. it was just one more occurrence of the old DataTag import

roblabat:
> fix: tests

Yes I didn't saw this one ^^ thanks

TkDodo:
so I got some feedback from @Andarist about this approach, and he mentioned that this approach uses an unused type param `Data`:

```
export declare interface TypedQueryKey<Data> extends QueryKey {}
```

Unused type params can be forgotten by the compiler; for now, I'll go ahead and merge the `feature/typed-query-key` PR, but please rebase this one onto the `rc` branch and we can keep discussion if we want to switch to this solution.  For the users, there shouldn't be any differences

roblabat:
Yes there shouldn't be any difference so there is no issue switching between solutions later.

Effectively most of the time unused types or forgotten by the compiler. But that's not issue in this particular use case as this type is a "fake" type only used for inference, it has not existence after compilation so it's intentional that the type is cleaned. whereas with the symbolic approach we define a type that will never be correct regarding the variable at runtime as the property [dataTagSymbol]: Type will never be used at runtime

roblabat:
here is the PR to rc branch https://github.com/TanStack/query/pull/6127

PR #6125, opened 2023-10-10
fix(solid-query): Default `reconcile` to false

### LINKED ISSUE #5738: v5: solid-query key and data mismatch with `staleTime` set
### Describe the bug

There's data and key mismatch with `staleTime` configured using Solid Query `v5` (only `v5`, `v4` works as expected). In another setup (not the reproduction), the mismatch always happens whether `staleTime` is set or not.

https://stackblitz.com/edit/solid-query-key-and-data-mismatch

### Steps to reproduce

1. Go to the [StackBlitz reproduction](https://stackblitz.com/edit/solid-query-key-and-data-mismatch)
2. Wait for the dependencies installation and server to start
3. Initially the `select` option set to `One` and the fetching starts for the key `['posts', '1']`
4. Change the `select` option to `Two` to fetch for `['posts', '2']` and the data below it will update
5. Switch to option `One` with `['posts', '1']`, the data will change as expected
6. Switch back to option `Two` with `['posts', '2']` but it will keep showing data for key `['posts', '1']` ⚠️

### Expected behavior

Switching between the two keys should return the data corresponding to each key.

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Brave (Chromium-based)
- Version: Brave `v1.52.130` Chromium `v114.0.5735.198`

### Tanstack Query adapter

solid-query

### TanStack Query version

v5.0.0-alpha.85

### TypeScript version

v5.1.3

### Additional context

_No response_

### LINKED ISSUE #5994: [solid-query v5]: placeholderData returns wrong values when returned data is an array
### Describe the bug

In solid-query v5, if you use `placeholderData: keepPreviousData` with a query that returns arrays, the cacheing breaks and returns the wrong results when returning to previously fetched queries.

See the demonstration in the attached video and codesandbox. In the codesandbox, the returned result should always match with the `search` property. This works fine when NOT using placeholderData. However, with placeholderData on, the `query.data` values seem to get out of sync with the latest query key.

https://codesandbox.io/s/solid-query-placeholder-broken-392lsq?file=/index.js

### Steps to reproduce

1. Type in "a". Notice that the correct result gets returned
2. Type in "ac". Notice that correct result gets returned
3. Delete "c" to return to "a", which should be cached. Notice that the query returns the wrong data

### Expected behavior

When using `placeholderData`, the query cache should return the same values as if you weren't using placeholderData.

Every time

### Screenshots or Videos

https://github.com/TanStack/query/assets/5554373/764b17f6-e413-4918-a5a0-0f628ada55f2

### Platform

OSX, Chrome

### Tanstack Query adapter

None

### TanStack Query version

v5 beta 23

### TypeScript version

_No response_

### Additional context

_No response_

### LINKED ISSUE #6102: [solid-query v5] Solid Start dependent query from context throws error (missing `defaultQueryOptions`)
### Describe the bug

Hey! I'm using Solid Start with the Cloudflare Pages adapter, `ssr: false` and solid-query v5-rc.4.

I've noticed that when I provide a query via context within a layout to my routes and use that result for a different query that depends on the result somewhere inside the nested route, I get the following error when navigating from somewhere else to that route:
CleanShot 2023-10-03 at 20 25 03

Reloading that route or directly navigating to it works fine though. It only happens when the route wasn't loaded before.

https://stackblitz.com/edit/github-czovmr

### Steps to reproduce

1. Go to the home route
2. Click on the "Nested" menu item
3. Error appears

### Expected behavior

There shouldn't be an error

Every time

### Screenshots or Videos

CleanShot 2023-10-03 at 20 29 51

### Platform

- macOS Ventura 13.6
- Chrome v117.0.5938.92

### Tanstack Query adapter

solid-query

### TanStack Query version

rc-4

### TypeScript version

v5.2.2

### Additional context

_No response_

ardeora:
Fixes #5738
Fixes #5994
Fixes #6102 

There seems to be a bug in how reconcile works with previously reconciled objects. Potentially a solid-js issue. Either way, The `reconcile` option needs to be used carefully and is not suitable for most cases. So I am making this opt-in. It wasn't documented yet so it's safe to default it to false.

### DIFF
--- a/packages/solid-query/src/QueryClient.ts
+++ b/packages/solid-query/src/QueryClient.ts
@@ -30,7 +30,7 @@ export interface QueryObserverOptions<
    * Set this to a reconciliation key to enable reconciliation between query results.
    * Set this to `false` to disable reconciliation between query results.
    * Set this to a function which accepts the old and new data and returns resolved data of the same type to implement custom reconciliation logic.
-   * Defaults reconciliation key to `id`.
+   * Defaults reconciliation to false.
    */
   reconcile?:
     | string
@@ -60,7 +60,7 @@ export interface InfiniteQueryObserverOptions<
    * Set this to a reconciliation key to enable reconciliation between query results.
    * Set this to `false` to disable reconciliation between query results.
    * Set this to a function which accepts the old and new data and returns resolved data of the same type to implement custom reconciliation logic.
-   * Defaults reconciliation key to `id`.
+   * Defaults reconciliation to false.
    */
   reconcile?:
     | string
--- a/packages/solid-query/src/__tests__/createQuery.test.tsx
+++ b/packages/solid-query/src/__tests__/createQuery.test.tsx
@@ -1046,6 +1046,7 @@ describe('createQuery', () => {
           count++
           return count === 1 ? result1 : result2
         },
+        reconcile: 'id',
       }))
 
       createRenderEffect(() => {
--- a/packages/solid-query/src/createBaseQuery.ts
+++ b/packages/solid-query/src/createBaseQuery.ts
@@ -111,7 +111,7 @@ export function createBaseQuery<
   const isRestoring = useIsRestoring()
 
   const defaultedOptions = createMemo(() =>
-    mergeProps(client().defaultQueryOptions(options()), {
+    mergeProps(client()?.defaultQueryOptions(options()) || {}, {
       get _optimisticResults() {
         return isRestoring() ? 'isRestoring' : 'optimistic'
       },
@@ -167,7 +167,7 @@ export function createBaseQuery<
           return reconcileFn(
             store,
             result,
-            reconcileOptions === undefined ? 'id' : reconcileOptions,
+            reconcileOptions === undefined ? false : reconcileOptions,
           )
         })
         // If the query has data we dont suspend but instead mutate the resource


TPReal:
Yeah, this bug hit us in mblajek/Memo twice at least. We have a TanStack Table populated from TanStack Query, with all of sorting, filtering, paging etc done on the backend. We observed various terrible bugs, like page 2 of results still showing data from page 1 (actually new data arrived, but the identity of query.data did not change), then we had bugs with filters being stuck, e.g. I set a filter, go to a different page (of the single page app), then back, the filter is then clear and everything is created anew, but I see filtered data. Why? Because the data in the cache for unfiltered and for filtered lists were somehow structurally not separate. The most "interesting" debugging session I've had in years. Figured out the reconcile finally, but yes, this is very broken right now.

ardeora:
> Yeah, this bug hit us in mblajek/Memo twice at least. We have a TanStack Table populated from TanStack Query, with all of sorting, filtering, paging etc done on the backend. We observed various terrible bugs, like page 2 of results still showing data from page 1 (actually new data arrived, but the identity of query.data did not change), then we had bugs with filters being stuck, e.g. I set a filter, go to a different page (of the single page app), then back, the filter is then clear and everything is created anew, but I see filtered data. Why? Because the data in the cache for unfiltered and for filtered lists were somehow structurally not separate. The most "interesting" debugging session I've had in years. Figured out the reconcile finally, but yes, this is very broken right now.
> 
> I was trying to report a bug, but it turns out I'm not able to prepare a small repro of the problems. I'm really glad someone else figured this out without my bug report :)

Thank you so much for the detailed feedback 😄  I would also like to encourage you to open any issues/discussions when you run into any challenges with Solid Query. This package (solid-query) is less mature than React query, so there might be some rough edges that we would like to iron out. Feedback like this really helps us to improve this library!

PR #6131, opened 2023-10-10
feat(devtools): allow editing of boolean values

bastibuck:
Hey,

stumbled upon a case where I needed to flip a flag in my app-code today several times. Thought I'd add it to the editable field values here as well to make this easier in the future.

image


ardeora (approved):
Looking great! Thanks

PR #6155, opened 2023-10-14
fix(types): useQueries can not infer type of TError

liaoliao666:
```ts
type BizError = Error & { code: number }

const todoOptions = queryOptions<string, BizError>({
  queryKey: [`todo`],
  queryFn: () => 'string',
})

// this is will not work if u manualy set `BizError` to todoOptions
useQueries({
  queries: [todoOptions],
})
```

### DIFF
--- a/packages/query-devtools/package.json
+++ b/packages/query-devtools/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tanstack/query-devtools",
-  "version": "5.0.0-rc.10",
+  "version": "5.0.0-rc.13",
   "description": "Developer tools to interact with and visualize the TanStack Query cache",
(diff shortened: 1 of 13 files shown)


TkDodo:
thank you 🙏 

could you please:

- add a type test to `useQueries.test.tsx`. There are a bunch of tests there already, it would be good to see a test that fails without your fix and works with your changes
- have a quick look at the vue, solid and svelte adapters, because they also have the useQueries / createQueries types and potentially have the same problem. Of course it would be ideal to unify those types somehow, but I guess this is out of scope

liaoliao666:
Already add test case and support infer TError for all frameworks

TkDodo (approved):

PR #6169, opened 2023-10-17
Update solid examples

birkskyum:
solid-start example is broken because the dependencies were outdated.

I updated all the solid examples.

### DIFF
--- a/examples/solid/basic-graphql-request/package.json
+++ b/examples/solid/basic-graphql-request/package.json
@@ -11,13 +11,13 @@
   "dependencies": {
     "@tanstack/solid-query": "^5.0.0",
     "@tanstack/solid-query-devtools": "^5.0.0",
-    "graphql": "^16.6.0",
+    "graphql": "^16.8.1",
     "graphql-request": "^6.1.0",
-    "solid-js": "^1.7.8"
+    "solid-js": "^1.8.1"
   },
   "devDependencies": {
-    "typescript": "^5.0.4",
-    "vite": "^4.4.4",
-    "vite-plugin-solid": "^2.7.0"
+    "typescript": "^5.2.2",
+    "vite": "^4.4.11",
+    "vite-plugin-solid": "^2.7.2"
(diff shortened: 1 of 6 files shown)


DamianOsipiuk:
@ardeora @birkskyum This actually broke the pipeline.

birkskyum:
Right, it's probably because `pnpm i` wasn't run after the rebase to sync it all.

ardeora:
Okay I think it's fixed now. I think the issue was that the pnpm-lockfile wasn't resolving the svelte vite plugin types package correctly

birkskyum:
@TkDodo  Great! the initial reason I started this was that I noticed the Solid examples at https://tanstack.com/query/latest didn't load. They still don't, but now they at least seem to work alright when accessed [directly](https://codesandbox.io/p/sandbox/github/tanstack/query/tree/main/examples/solid/simple?embed=1) in codesandbox.


Screenshot 2023-10-20 at 19 37 02

DamianOsipiuk:
It does not work, cause example is named `basic-typescript` instead of `basic`. So if you rename the example folder, it will work.

birkskyum:
You're right. I can see the ci.json says:

```json
  "sandboxes": [
    "/examples/react/basic-typescript",
    "/examples/solid/basic",
    "/examples/svelte/basic",
    "/examples/vue/basic"
  ]
```

 it's probably faster to update https://tanstack.com/query/latest to match those.

DamianOsipiuk:
I would prefer for all adapters to expose `basic` example to avoid putting conditions in multiple places.

birkskyum:
> I would prefer for all adapters to expose `basic` example to avoid putting conditions in multiple places.

In that case there is a PR here:
https://github.com/TanStack/query/pull/6169#issuecomment-1773285732

But it will break this example site which actually uses `basic-typescript`
https://tanstack.com/query/latest/docs/solid/examples/solid/basic-typescript

PR #6172, opened 2023-10-18
fix: update types field to make packages cjs default in esm

manudeli:
I updated all packages' `types` field of package.json to make package using cjs default following `main` field after trying using `exports` field and failing using it

I'm not sure but isn't this right? I'm so confused even after read this doc https://www.typescriptlang.org/docs/handbook/modules/reference.html

@ahejlsberg Could you help this case and updating doc to setup package.json like this case

### DIFF
--- a/packages/eslint-plugin-query/package.json
+++ b/packages/eslint-plugin-query/package.json
@@ -15,7 +15,7 @@
     "url": "https://github.com/sponsors/tannerlinsley"
   },
   "type": "module",
-  "types": "build/legacy/index.d.ts",
+  "types": "build/legacy/index.d.cts",
   "main": "build/legacy/index.cjs",
   "module": "build/legacy/index.js",
   "exports": {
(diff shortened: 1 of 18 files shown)


TkDodo:
I'd need to ask @lachlancollins and @DamianOsipiuk about this 🙏

DamianOsipiuk:
@manudeli Did you experienced any error by using the current setup?

Tests are failing because `svelte` and `solidjs` packages do not have legacy builds, so they should be left as-is.

manudeli:
> @manudeli Did you experienced any error by using the current setup?
> 
> Tests are failing because `svelte` and `solidjs` packages do not have legacy builds, so they should be left as-is.

This Pull Request was not created because there was an error. After looking at package.json, I was wondering if these packages should look at index.d.cts, not index.d.ts, which has the same `main` field (index.cjs) and `types` field, so I made this Pull Request. That's all.

Since the type field was changed to module in v4 and v5, I wanted to make sure this part was confirmed.

lachlancollins:
@manudeli this is an interesting thought, I hadn't really considered it - https://arethetypeswrong.github.io/ reports compatibility with all moduleResolution options, so if it's not causing issues I think it's best left as-is for now.

PR #6175, opened 2023-10-18
docs(suspensive-react-query): update links

manudeli:
https://suspensive.org is migrated from docusaurus to nextra
so I updated suspesnive-react-query.md's links

### DIFF
--- a/docs/react/community/suspensive-react-query.md
+++ b/docs/react/community/suspensive-react-query.md
@@ -5,7 +5,7 @@ title: Suspensive React Query
 
 Typesafe useQuery, useQueries, useInfiniteQuery with default suspense option.
 
-Use @suspensive/react-query, delegate loading and error handling to the outside of the component with [useSuspenseQuery](https://suspensive.org/docs/react-query/src/useSuspenseQuery.i18n), [useSuspenseQueries](https://suspensive.org/docs/react-query/src/useSuspenseQueries.i18n) and [useSuspenseInfiniteQuery](https://suspensive.org/docs/react-query/src/useSuspenseInfiniteQuery.i18n), and focus on success inside the component.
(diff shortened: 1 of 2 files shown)


TkDodo:
if you don't mind - what's the advantage of suspensive-react-query now that we have `useSuspenseQuery` hooks built-in ?

TkDodo:
> There are still many v4 users

yes, and we have different docs for v4, we wouldn't be changing that:

https://tanstack.com/query/v4/docs/react/community/suspensive-react-query

but for v5, if you agree that it's better to use the built-in hooks, I'd rather remove it from the docs.

manudeli:
> > There are still many v4 users
> 
> yes, and we have different docs for v4, we wouldn't be changing that:
> 
> https://tanstack.com/query/v4/docs/react/community/suspensive-react-query
> 
> but for v5, if you agree that it's better to use the built-in hooks, I'd rather remove it from the docs.

I added now a commit(04eb45e) to improve this document after your feedback

I think @suspensive/react-query still could be a good package for people using @tanstack/react-query v4's suspense feature as they gradually migrate to v5. So I hope the documentation also remains in v5. If you okay, I would appreciate it if we could add something like this change(04eb45e) and maintain this document in the v5 official documentation. if v5 official docs have this document, Users of @suspensive/react-query will also be able to migrate gradually from @suspensive/react-query's hooks to @tanstack/react-query v5's built in hooks without confusion.

How about like this(04eb45e)? if still you not okay, I will make commit to remove this doc.

TkDodo:
I'll keep this in for now, but I'll re-do the whole community resources section soon. I don't think that we'll list suspensive-react-query then, sorry.

> still could be a good package for people using @tanstack/react-query v4's suspense feature

if users are on v4, they will read the v4 docs, where the package is listed. If they are on v4 and use our experimental suspense, it would be better if they'd upgrade to the built-in suspense hooks.

PR #6183, opened 2023-10-18
docs(suspensive-react-query): update links for v4

manudeli:
I update v4 links of @suspensive/react-query


manudeli:
@TkDodo this is just update doc for valid links, easily accept this please

PR #6187, opened 2023-10-19
docs(useQuery): remove suspense option in useQuery() v5

diggymo:
In v5, To use `suspense` developers can use `useSuspenseQuery()` and can't use `useQuery()`.

However, according to the documentation, it is different.

[In the actual type definition, `suspense` option is Omited](https://github.com/TanStack/query/blob/debece31a5aa5f4a1c800c7be84ed669fba5a092/packages/react-query/src/types.ts#L39C11-L39C11), so the documentation is incorrect. So I fixed it

### DIFF
--- a/docs/react/reference/useQuery.md
+++ b/docs/react/reference/useQuery.md
@@ -51,7 +51,6 @@ const {
   select,
   staleTime,
   structuralSharing,
-  suspense,
   throwOnError,
 })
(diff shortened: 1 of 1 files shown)


TkDodo:
thanks, it's an oversight that we still have this in the docs for useQuery 👍

PR #6193, opened 2023-10-19
fix(svelte-query): Prevent getIsRestoringContext from throwing error

### LINKED ISSUE #6188: Can't createQuery outside of a svelte component, even when providing your own queryClient
### Describe the bug

When providing my own queryClient in a `+page.ts` load function, calling `createQuery(opts, queryClient)` throws an error before the query is created.

Svelte's `getContext` throws if it's not inside a component, and this happens to be called during the createQuery call.

```log
+page.ts:18 Error: Function called outside component initialization
    at get_current_component (chunk-XXZCQZAU.js?v=e97699bd:1202:11)
    at getContext (chunk-XXZCQZAU.js?v=e97699bd:1241:10)
    at getIsRestoringContext (@tanstack_svelte-que…?v=e97699bd:2603:23)
    at useIsRestoring (@tanstack_svelte-que…?v=e97699bd:2614:10)
    at createBaseQuery (@tanstack_svelte-que…?v=e97699bd:2632:23)
    at createQuery (@tanstack_svelte-que…?v=e97699bd:2657:10)
    at load (+page.ts:10:18)
    at async load_node (client.js?v=e97699bd:530:14)
```

Specifically it's from here

 `@tanstack/svelte-query/context.ts`
```typescript
export const getIsRestoringContext = (): Readable<boolean> => {
  const isRestoring = getContext(_isRestoringContextKey) // <-- this line
  if (!isRestoring) return readable(false)
  return isRestoring as Readable<boolean>
}
```

https://stackblitz.com/edit/stackblitz-starters-6h9bak?file=src%2Froutes%2F%2Bpage.ts

### Steps to reproduce

1. open the stackblitz link
2. wait for it to run
3. observe error in the browser tools

### Expected behavior

I expect to be able to create the query if I provide the queryClient outside of a svelte component.

Every time

### Screenshots or Videos

_No response_

### Platform

- node: 18.18.1
- svelte: "^4.2.1"
- @sveltejs/kit: "^1.25.1"

### Tanstack Query adapter

svelte-query

### TanStack Query version

v5.0.0

### TypeScript version

v5.2.2

### Additional context

I've got a patch to workaround this - and my application works as expected.
[@tanstack+svelte-query+5.0.0.patch](https://github.com/TanStack/query/files/13038629/%40tanstack%2Bsvelte-query%2B5.0.0.patch)

```patch
diff --git a/node_modules/@tanstack/svelte-query/src/context.ts b/node_modules/@tanstack/svelte-query/src/context.ts
index 98ca5d6..860efcb 100644
--- a/node_modules/@tanstack/svelte-query/src/context.ts
+++ b/node_modules/@tanstack/svelte-query/src/context.ts
@@ -26,9 +26,13 @@ const _isRestoringContextKey = '$$_isRestoring'
 
 /** Retrieves a `isRestoring` from Svelte's context */
 export const getIsRestoringContext = (): Readable<boolean> => {
-  const isRestoring = getContext(_isRestoringContextKey)
-  if (!isRestoring) return readable(false)
-  return isRestoring as Readable<boolean>
+  try {
+    const isRestoring = getContext(_isRestoringContextKey)
+    if (!isRestoring) return readable(false)
+    return isRestoring as Readable<boolean> 
+  } catch (error) {
+    return readable(false);
+  }
 }
 
 /** Sets a `isRestoring` on Svelte's context */
```

I'm unsure of any downstream impacts, but it's a simple change that allows the application to work, and means there isn't a bunch of repeated boilerplate "createQuery" calls inside each component.

softgripper:
Calling getIsRestoringContext could throw an error.

Now it can't.

The types usage is improved, and the method has a testcase

Fixes #6188


softgripper:
> I'm not sure if try/catch on a `getContext` is really the idiomatic way. There is also a [hasContext](https://svelte.dev/docs/svelte#hascontext) check?

Unfortunately `hasContext` also throws `Error: Function called outside component initialization`.

TkDodo (approved):

softgripper:
> interesting api. So `hasContext` either throws or returns true?

I only know it throws because I attempted to use it on this branch after you mentioned it above.  (The doc also mentions 'Must be called during component initialisation.')

I believe it's because context is based on the component hierarchy.  So it's not asking `hasContext` but it's asking `hasContextGotKeyEntry`.  Basically a function to a map lookup that fails if the actual map itself doesn't exist (or can't exist).  I haven't delved into it's source, but that's the behavior, and what the docs suggest.

Component.svelte
```
<script>
import { hasContext } from 'svelte';

if (hasContext('mykey')) {
  // do something
}
</script>
```

softgripper:
Sorry, I performed that merge action by mistake.  I hope it doesn't require another manual approval for that. I'm not exactly sure what the full flow is for PR.

TkDodo:
we need to approve every CI run for first time contributors 🤷

lachlancollins (approved):
Sorry I missed this PR, thanks @softgripper !

PR #6194, opened 2023-10-19
docs: update the doc of react query kit

liaoliao666:


TkDodo:
I'll merge this, but I think it's generally not ideal to have detailed examples / descriptions in our docs. It's too easy to get outdated. I'll have a new approach in mind for the community resources

PR #6195, opened 2023-10-19
Feature/angular query

arnoud-dv:
## Experimental Angular adapter

Uses signals, will support Angular 17+

The Svelte adapter served as a very good base, thanks! ❤️

Feedback is very welcome! Contact me on the Tanstack Discord in the angular-query channel or in [this Github discussion](https://github.com/TanStack/query/discussions/6293).

### To do
- [x] react to attribute changes of dev tools
- [x] find a solution for dev tools needing "skipLibCheck": true setting in app using them
- [x] reduce boilerplate for options signals
- [x] redo dependency injection and providers
  - [x] improve query client mounting and unmounting
  - [x] inject* functions such as injectQuery
- [ ] injectInfiniteQuery
- [ ] injectMutationState
- [x] docs
- [x] update unit tests to use the new function based API
- [ ] more unit tests
- [x] more examples 
- [x] integration test

image


docs/angular/devtools.md
```diff
@@ -0,0 +1,120 @@
...
+---
+
+## Install and Import the Devtools
+
+The devtools are a separate package that you need to install:
+
+```bash
+$ npm i @tanstack/angular-query-devtools-experimental
+# or
+$ pnpm add @tanstack/react-query-devtools-experimental
```
eneajaho, line 15:
```suggestion
$ pnpm add @tanstack/angular-query-devtools-experimental
```

docs/angular/devtools.md
```diff
@@ -0,0 +1,120 @@
...
+# or
+$ yarn add @tanstack/react-query-devtools-experimental
```
eneajaho, line 17:
```suggestion
$ yarn add @tanstack/angular-query-devtools-experimental
```

docs/angular/devtools.md
```diff
@@ -0,0 +1,120 @@
...
+```
+
+You can import the devtools like this:
+
+```tsx
+import { AngularQueryDevtools } from '@tanstack/angular-query-devtools-experimental'
+```
+
+By default, React Query Devtools are only included in bundles when `process.env.NODE_ENV === 'development'`, so you don't need to worry about excluding them during a production build.
```
eneajaho, line 26:
```suggestion
By default, Angular Query Devtools are only included in bundles when `process.env.NODE_ENV === 'development'`, so you don't need to worry about excluding them during a production build.
```

docs/angular/devtools.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Floating Mode
+
+Floating Mode will mount the devtools as a fixed, floating element in your app and provide a toggle in the corner of the screen to show and hide the devtools. This toggle state will be stored and remembered in localStorage across reloads.
+
+Place the following code as high in your React app as you can. The closer it is to the root of the page, the better it will work!
```
eneajaho, line 32:
```suggestion
Place the following code as high in your Angular app as you can. The closer it is to the root of the page, the better it will work!
```

docs/angular/devtools.md
arnoud-dv: Docs are work in progress ;-)

packages/angular-query-experimental/src/QueryClientService.ts
```diff
@@ -0,0 +1,37 @@
...
+import type { OnDestroy } from '@angular/core'
+import type { QueryClient } from '@tanstack/query-core'
+
+/**
+ * Takes care of mounting and unmounting the query client.
+ * @internal
+ */
+@Injectable()
+export class QueryClientService implements OnDestroy {
+  readonly #queryClient: QueryClient | undefined
```
eneajaho, line 12:
```suggestion
  readonly #queryClient = inject(QUERY_CLIENT, { optional: true });
```

eneajaho: We can just optionally inject it here, and in the constructor can mount it if it exists.

packages/angular-query-experimental/src/useIsFetching.ts
```diff
@@ -0,0 +1,42 @@
...
+} from '@angular/core'
+import {
+  type QueryClient,
+  type QueryFilters,
+  notifyManager,
+} from '@tanstack/query-core'
+import { QueryClientService } from './QueryClientService'
+import type { Signal } from '@angular/core'
+
+export function useIsFetching(
```
eneajaho, line 15: Instead of the use keyword, in Angular, I'd use the inject keyword (injectIsFetching). 

This way the developer knows that this should be run in an injection context (because we are also using the inject fn inside this fn).

packages/angular-query-experimental/src/useIsFetching.ts
```diff
@@ -0,0 +1,42 @@
...
+    notifyManager.batchCalls(() => {
+      const newIsFetching = queryClient!.isFetching(filters)
+      if (isFetching !== newIsFetching) {
+        // * and update with each change
+        isFetching = newIsFetching
+        result.set(isFetching)
+      }
+    }),
+  )
+  const destroyRef = inject(DestroyRef)
```
eneajaho, line 39: I'd move the injection to the top, so it's visible.

packages/angular-query-experimental/src/providers.ts
```diff
@@ -0,0 +1,86 @@
...
+): EnvironmentProviders {
+  return makeEnvironmentProviders([
+    { provide: QUERY_CLIENT, useValue: queryClient },
+    { provide: CreateQuery, useValue: createQuery },
+    { provide: CreateMutation, useValue: createMutation },
+    { provide: CreateQueries, useValue: createQueries },
+    {
+      provide: QueryClientService,
+      useClass: QueryClientService,
+      deps: [QUERY_CLIENT],
```
eneajaho, line 69: You won't need `deps` here, because you are using inject in QueryClientService

packages/angular-query-experimental/src/providers.ts
```diff
@@ -0,0 +1,86 @@
...
+ * })
+ * ```
+ *
+ *
+ * @publicApi
+ * @param queryClient The `QueryClient` to provide.
+ * @returns A set of providers to set up Angular Query.
+ */
+export function provideAngularQuery(
+  queryClient: QueryClient,
```
eneajaho, line 59: I'd like the configuration to be an object, so later we can pass other configs. 

```ts
{
  client: QueryClient,
  // defaultOptions ?
}
```

examples/angular/simple/src/simple-example.component.ts
```diff
@@ -0,0 +1,43 @@
...
+  stargazers_count: number
+  forks_count: number
+}
+
+@Component({
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  selector: 'simple-example',
+  standalone: true,
+  template: `
+    <ng-container *ngIf="query().isPending">Loading...</ng-container>
```
eneajaho, line 20: I'd like to see the query being a normal object and `isPending` being a signal, so we can be surgical in template updates (when signal components get released)

Same thing for error and data. 

This way only the part of the template that has a signal update will get updated.

examples/angular/simple/src/simple-example.component.ts
```diff
@@ -0,0 +1,43 @@
...
+      <strong>🍴 {{ data.forks_count }}</strong>
+    </div>
+    <angular-query-devtools initialIsOpen />
+  `,
+  imports: [AngularQueryDevtoolsComponent, CommonModule],
+})
+export class SimpleExampleComponent {
+  query = inject(CreateQuery)(() => ({
+    queryKey: ['repoData'],
+    queryFn: () =>
```
eneajaho, line 38: In an Angular world, I'd like to see the queryFn accept an observable rather than a promise (or both), because most of the time this will be in a service using Angular's http client instead of axios, because of interceptors and better SSR support in general.

examples/angular/simple/src/simple-example.component.ts
```diff
@@ -0,0 +1,43 @@
...
+      <strong>👀 {{ data.subscribers_count }}</strong>
+      <strong>✨ {{ data.stargazers_count }}</strong>
+      <strong>🍴 {{ data.forks_count }}</strong>
+    </div>
+    <angular-query-devtools initialIsOpen />
+  `,
+  imports: [AngularQueryDevtoolsComponent, CommonModule],
+})
+export class SimpleExampleComponent {
+  query = inject(CreateQuery)(() => ({
```
eneajaho, line 36:
```suggestion
  query = injectQuery(() => ({
```

This kind of API is easier to write but also easier to read. 

Later, we can enhance it by passing some helper functions.

examples/angular/default-query-function/src/default-query-function-example.component.ts
```diff
@@ -0,0 +1,141 @@
...
+  }
+  postIdSignal = signal<number>(0)
+
+  // You can even leave out the queryFn and just go straight into options
+  postQuery = inject(CreateQuery)<Post>(() => ({
+    enabled: this.postIdSignal() > 0,
+    queryKey: [`/posts/${this.postIdSignal()}`],
+  }))
+
+  queryClient = inject(UseQueryClient)
```
eneajaho, line 65: Why is this needed?

packages/angular-query-experimental/src/createBaseQuery.ts
```diff
@@ -0,0 +1,79 @@
...
+  })
+
+  /** Creates the observer */
+  const observer = new Observer<
+    TQueryFnData,
+    TError,
+    TData,
+    TQueryData,
+    TQueryKey
+  >(queryClient, defaultedOptionsSignal())
```
gmfun, line 55: One issue with this approach is this will require the computed value to be evaluated immediatly. It may be possible that it may cause error.

example lets say the api call requires `id` passed to the component as input. This input is stored as signal and used to make the api call. If this computeds value is evaluated here, it would be undefined. 
Also similar issue will be there with signal component whenever we have that

I had implemented a weird approach to prevent this. I am initializing `Observer` with some empty option.

```
 const emptyQueryKey: QueryKey = [''] as const
const queryObserver = new Observer<any, any, any, any, QueryKey>(
    client,
    client.defaultQueryOptions({
      queryKey: emptyQueryKey,
      queryFn: () => {
        return Promise.resolve(false)
      },
      enabled: false,
    })
  )
```

packages/angular-query-experimental/src/useIsFetching.ts
```diff
@@ -0,0 +1,42 @@
...
+  type QueryFilters,
+  notifyManager,
+} from '@tanstack/query-core'
+import { QueryClientService } from './QueryClientService'
+import type { Signal } from '@angular/core'
+
+export function useIsFetching(
+  filters?: QueryFilters,
+  queryClient?: QueryClient,
+): Signal<number> {
```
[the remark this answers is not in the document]
arnoud-dv: No, but it [seems to confuse more people](https://github.com/TanStack/query/discussions/6291).

packages/angular-query-experimental/src/createBaseQuery.ts
```diff
@@ -0,0 +1,79 @@
...
+  effect(() => {
+    // Do not notify on updates because of changes in the options because
+    // these changes should already be reflected in the optimistic result.
+    observer.setOptions(defaultedOptionsSignal(), { listeners: false })
+  })
+
+  const result = signal(observer.getCurrentResult())
+  const unsubscribe = observer.subscribe(
+    notifyManager.batchCalls((val) => {
+      result.set(val)
```
gmfun, line 66: In my case I need to call result.set inside injector.get(NgZone).run otherwise change detected where not triggered for some case. Not sure it is required

packages/angular-query-experimental/src/useIsFetching.ts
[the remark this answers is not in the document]
arnoud-dv: Will consider renaming. It would not be consistent with other adapters but might be worth it.

examples/angular/default-query-function/src/default-query-function-example.component.ts
arnoud-dv: The queryClient? It's used by the styling on the links to check if it's cached.

packages/angular-query-experimental/src/useIsFetching.ts
arnoud-dv: Sounds like it might be a good idea to name these functions injectX, but I am not aware of such a naming convention. Do you maybe have some examples from open source software? If there are none it might still be a good idea but I would like to hear some more opinions on it.

examples/angular/simple/src/simple-example.component.ts
arnoud-dv: Agree, it's not hard to convert an observable to a promise but it's what Angular developers would commonly expect to just work out of the box.

packages/angular-query-experimental/src/useIsFetching.ts
tomalaforge: `use*` is a tsx convention.

There is no naming convention per say, but most of the Angular OSS is using `inject*` since the `inject` was introducted in v14. As @eneajaho said, it gives a warning to the developer that this function is injecting Services and need to be run in injection context.

examples/angular/simple/src/simple-example.component.ts
arnoud-dv: Honestly just getting rid of the pair of parentheses is already convincing enough to rename it 😁

packages/angular-query-experimental/src/useIsFetching.ts
eneajaho: Open source Angular lib that uses inject convention a lot: https://ngxtension.netlify.app/utilities/injectors/inject-is-intersecting/

examples/angular/default-query-function/src/default-query-function-example.component.ts
eneajaho: That would be a great usecase for an angular directive to be honest 😄

arnoud-dv: The example apps could probably be improved, or completely new ones written specific to Angular Query. On the other hand I just quickly ported them from React and it's nice if the simple ones can fit on a small amount of screen real estate inside documentation.

packages/angular-query-experimental/src/createBaseQuery.ts
arnoud-dv: This is something I came across in React Query too. A solution is for the developer to set enabled to false until all information required by the query is present. If there is a better solution I would welcome it but I cannot think of anything.

packages/angular-query-experimental/src/createBaseQuery.ts
arnoud-dv: Do you maybe have a sandbox demonstrating this? I don't think I understand the issue.

packages/angular-query-experimental/src/useIsFetching.ts
alex-okrushko: While I agree that `inject` would be better, referencing your own library feels like cheating @eneajaho 😂

packages/angular-query-experimental/src/useIsFetching.ts
```diff
@@ -0,0 +1,42 @@
...
+  type QueryClient,
+  type QueryFilters,
+  notifyManager,
+} from '@tanstack/query-core'
+import { QueryClientService } from './QueryClientService'
+import type { Signal } from '@angular/core'
+
+export function useIsFetching(
+  filters?: QueryFilters,
+  queryClient?: QueryClient,
```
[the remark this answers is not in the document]
alex-okrushko: 100%

packages/angular-query-experimental/src/useIsFetching.ts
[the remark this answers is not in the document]
arnoud-dv: How about injectFetchCount?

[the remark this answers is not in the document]
arnoud-dv: Or injectIsFetching() with both a boolean and a number. It is probably common to treat it as a boolean to just show some loading indicator.

packages/angular-query-experimental/src/providers.ts
arnoud-dv: I started with implementing this but I think I will pass on this one: looking at for example `provideRouter` in Angular itself it works with separate parameters instead of an options object.

`defaultOptions` needs to be passed to the `QueryClient` constructor (which comes straight from Tanstack query-core) and I do not foresee any other configuration will be needed. It would also mean an extra nesting level. Using an overload your suggestion could be implemented in a non-breaking way later.

packages/angular-query-experimental/src/useIsFetching.ts
tomalaforge: [Angular Three](https://github.com/nartc/angular-three) is using a lot of them as well. but @eneajaho and Chau are from of the same species 🤣

examples/angular/simple/src/simple-example.component.ts
arnoud-dv: I put this on my todo list to look at. Not sure how much work it would be.

examples/angular/simple/src/simple-example.component.ts
arnoud-dv: Put on my todo list, with high prio. I do think it's important to support both promises and observables.

examples/angular/simple/src/simple-example.component.ts
```diff
@@ -0,0 +1,43 @@
...
+      <strong>👀 {{ data.subscribers_count }}</strong>
+      <strong>✨ {{ data.stargazers_count }}</strong>
+      <strong>🍴 {{ data.forks_count }}</strong>
+    </div>
+    <angular-query-devtools initialIsOpen />
+  `,
+  imports: [AngularQueryDevtoolsComponent, CommonModule],
+})
+export class SimpleExampleComponent {
+  query = injectQuery(() => ({
```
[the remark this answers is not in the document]
arnoud-dv: Hi Netanel, thank you for your feedback. Can you be more specific?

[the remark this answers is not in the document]
arnoud-dv: Will add documentation about recommended architectures to work with this adapter.

eneajaho:
Hello @arnoud-dv 
I just created a PR with some QOL improvements in https://github.com/inshared/query/pull/1 

I refactored the DI stuff to be aligned together and also exported the client from most of the options callback fn, so the devs won't have to inject it in the component, but can directly use it if they want it. 

Also updated the doc to use angular's HTTP client with `lastValueFrom` rxjs operator which converts an observable to a promise and it should work fine with queryClient, without needing to convert anything from the inside

Please review it and let me know what you think

examples/angular/basic/src/basic-example.component.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  // Until Angular supports signal-based inputs, we have to set a signal
+  @Input({ required: true })
+  set postId(value: number) {
+    this.postIdSignal.set(value)
+  }
+  postIdSignal = signal<number>(0)
+  postQuery = injectQuery(() => ({
+    enabled: this.postIdSignal() > 0,
+    queryKey: ['post', this.postIdSignal()],
```
ApplY3D, line 60: Should we add the possibility to pass an observable into the `queryKey`? Great things are now available with signals, but we still have observables.

ApplY3D: for example:

```typescript
queryKey: ['post', this.postId$]
// or
queryKey: postId$.pipe(map(id => ['post', id]))
```

docs/angular/devtools.md
arnoud-dv: It may be worth looking into adding first class support, but [toSignal](https://angular.dev/guide/signals/rxjs-interop#tosignal) may be sufficient. What do you think?

examples/angular/basic/src/basic-example.component.ts
ApplY3D: Also other props like `enabled` might be reactive. Probably `injectQuery` itself should be able to receive `Observable`

arnoud-dv: Hi, what do you think about  using [toSignal()](https://angular.dev/guide/signals/rxjs-interop#tosignal)?

`queryKey: ['post', toSignal(this.postId$)]`

Would that be sufficient? Willing to look into adding first class support for observables if many people want it but on the other hand it might increase complexity quite a bit and Angular has good interop between observables and signals.

ApplY3D: Yes, it might suffice, but it needs to be tested in real situations for potential leaks or other issues.

TkDodo (approved):

Issue #5760, opened 2023-07-21
v5: keepPreviousData codemod


TkDodo:
- for feature: #4688

```
keepPreviousData: true
```

should become:

```
placeholderData: (previousData) => previousData
```

OR, alternatively:

```
import { keepPreviousData } from '@tanstack/react-query'

placeholderData: keepPreviousData
```

If a `placeholderData` was already specified, it's fine if the codemod doesn't do anything and prints a warning instead. I think this case must be reviewed manually anyways.

PR #6196, opened 2023-10-19
 `placeholderData` codemod

balazsmatepetro:
Closes https://github.com/TanStack/query/issues/5760


packages/codemods/src/v5/keep-previous-data/__testfixtures__/default.input.tsx
```diff
@@ -0,0 +1,42 @@
...
+    const { data } = await axios.get(
+        'https://jsonplaceholder.typicode.com/posts',
+    )
+    return data
+  }
+  const { data } = useQueries({
+    queryKey: ['posts'],
+    queryFn: queryFn,
+    keepPreviousData: true,
+    placeholderData: (previousData) => previousData
```
TkDodo, line 38: in v4, `placeholderData` didn't have a param, so I'd change this example to:

```
```suggestion
    keepPreviousData: true,
    placeholderData: () => "somePlaceholderData"
```

also, can we add an example where placeholderData isn't a function:

```
keepPreviousData: true,
placeholderData: "somePlaceholderData"
```

and then maybe some combos where we only have `placeholderData`, but not `keepPreviousData`, which shouldn't be touched at all.

packages/codemods/src/v5/keep-previous-data/__testfixtures__/default.input.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const queryFn = async (): Promise<Array<Post>> => {
+    const { data } = await axios.get(
+      'https://jsonplaceholder.typicode.com/posts',
+    )
+    return data
+  }
+  const { data } = useQuery({
+    queryKey: ['posts'],
+    queryFn: queryFn,
+    keepPreviousData: true,
```
TkDodo, line 21: I would only touch this if we have explicitly set `keepPreviousDat: true`. Because things like `keepPreviousData: someCondition` are harder to migrate. So we might need an example that proves that these are not touched and get a proper warning message.

balazsmatepetro: So it should be executed only when the `keppPreviousData` property has value `true`, right?

packages/codemods/src/v5/keep-previous-data/__testfixtures__/default.input.tsx
balazsmatepetro: - a0b452d47d9b8228b824f46b0a2196200bf943fa
- 1813e22a3c84cf4016072c6691ef05e5b35d8e46

packages/codemods/src/v5/keep-previous-data/__testfixtures__/default.input.tsx
balazsmatepetro: 9fd29542826b96b968dc0d75d2a90c475d01bf33

balazsmatepetro:
@TkDodo please review the contents of the `default.input.tsx` and `default.output.tsx` files. If these files contain the necessary amount of examples, and the output of the code mod is correct, then I think we can merge it.

Shall I mention somewhere it's designed to work with only object expressions?

packages/codemods/src/v5/keep-previous-data/__testfixtures__/default.input.tsx
```diff
@@ -0,0 +1,90 @@
+import * as React from 'react'
```
balazsmatepetro, line 1: @TkDodo please review the examples.

packages/codemods/src/v5/keep-previous-data/__testfixtures__/default.output.tsx
```diff
@@ -0,0 +1,90 @@
+import * as React from 'react'
```
balazsmatepetro, line 1: @TkDodo please review the output of the codemod.

packages/codemods/src/v5/keep-previous-data/keep-previous-data.js
```diff
@@ -0,0 +1,208 @@
...
+  const root = jscodeshift(file.source)
+  const utils = createUtilsObject({ root, jscodeshift })
+  const filePath = file.path
+
+  const dependencies = { jscodeshift, utils, root, filePath }
+
+  const { isKeepPreviousDataInUse } = transformUsages({
+    ...dependencies,
+    config: {
+      hooks: ['useInfiniteQuery', 'useQueries', 'useQuery'],
```
balazsmatepetro, line 184: @TkDodo other hooks might be affected?

packages/codemods/src/v5/keep-previous-data/__testfixtures__/default.input.tsx
TkDodo: thanks for the comments, this looks perfect

TkDodo (approved):

packages/codemods/src/v5/keep-previous-data/keep-previous-data.js
TkDodo: oops I missed this comment. We also have the global defaults like:

```
new QueryClient({
  defaultOptions: {
    queries: {
      keepPreviousData: true
    }
  }
})
```

and: 

```
queryClient.setQueryDefaults(['key'], {
  keepPreviousData: true
})
```

balazsmatepetro: Don't worry, I will open another PR.

balazsmatepetro: Covered by:
- https://github.com/TanStack/query/pull/6283

PR #6127, opened 2023-10-10
Feature/typed query key

roblabat:
New implementation for typedQueryKey for discussion

see discusions from previous PR https://github.com/TanStack/query/pull/6123

PR #6201, opened 2023-10-20
Feature/typed query key

roblabat:
Reopenned from https://github.com/TanStack/query/pull/6127

### DIFF
--- a/packages/query-core/src/queryClient.ts
+++ b/packages/query-core/src/queryClient.ts
@@ -11,7 +11,6 @@ import { focusManager } from './focusManager'
 import { onlineManager } from './onlineManager'
 import { notifyManager } from './notifyManager'
 import { infiniteQueryBehavior } from './infiniteQueryBehavior'
-import type { DataTag, NoInfer } from './types'
 import type { QueryState } from './query'
 import type {
   CancelOptions,
@@ -34,6 +33,7 @@ import type {
   RefetchQueryFilters,
   ResetOptions,
   SetDataOptions,
+  TypedQueryKey,
 } from './types'
 import type { MutationFilters, QueryFilters, Updater } from './utils'
 
@@ -108,16 +108,9 @@ export class QueryClient {
     return this.#mutationCache.findAll({ ...filters, status: 'pending' }).length
   }
 
-  getQueryData<
-    TQueryFnData = unknown,
-    TaggedQueryKey extends QueryKey = QueryKey,
-    TInferredQueryFnData = TaggedQueryKey extends DataTag<
-      unknown,
-      infer TaggedValue
-    >
-      ? TaggedValue
-      : TQueryFnData,
-  >(queryKey: TaggedQueryKey): TInferredQueryFnData | undefined
+  getQueryData<TQueryFnData = unknown>(
+    queryKey: TypedQueryKey<TQueryFnData> | QueryKey,
+  ): TQueryFnData | undefined
   getQueryData(queryKey: QueryKey) {
     return this.#queryCache.find({ queryKey })?.state.data
   }
@@ -146,24 +139,12 @@ export class QueryClient {
       })
   }
 
-  setQueryData<
-    TQueryFnData = unknown,
-    TaggedQueryKey extends QueryKey = QueryKey,
-    TInferredQueryFnData = TaggedQueryKey extends DataTag<
-      unknown,
-      infer TaggedValue
-    >
-      ? TaggedValue
-      : TQueryFnData,
-  >(
-    queryKey: TaggedQueryKey,
-    updater: Updater<
-      NoInfer<TInferredQueryFnData> | undefined,
-      NoInfer<TInferredQueryFnData> | undefined
-    >,
+  setQueryData<TQueryFnData>(
+    queryKey: TypedQueryKey<TQueryFnData> | QueryKey,
+    updater: Updater<TQueryFnData | undefined, TQueryFnData | undefined>,
     options?: SetDataOptions,
-  ): TInferredQueryFnData | undefined {
-    const query = this.#queryCache.find<TInferredQueryFnData>({ queryKey })
+  ): TQueryFnData | undefined {
+    const query = this.#queryCache.find<TQueryFnData>({ queryKey })
     const prevData = query?.state.data
     const data = functionalUpdate(updater, prevData)
 
--- a/packages/query-core/src/queryClient.ts
+++ b/packages/query-core/src/queryClient.ts
@@ -157,12 +157,10 @@ export class QueryClient {
       : TQueryFnData,
   >(
     queryKey: TaggedQueryKey,
-    updater:
-      | NoInfer<TInferredQueryFnData>
-      | undefined
-      | ((
-          prev: TInferredQueryFnData | undefined,
-        ) => NoInfer<TInferredQueryFnData> | undefined),
+    updater: Updater<
+      NoInfer<TInferredQueryFnData> | undefined,
+      NoInfer<TInferredQueryFnData> | undefined
+    >,
     options?: SetDataOptions,
   ): TInferredQueryFnData | undefined {
     const query = this.#queryCache.find<TInferredQueryFnData>({ queryKey })
--- a/packages/query-core/src/queryClient.ts
+++ b/packages/query-core/src/queryClient.ts
@@ -11,7 +11,7 @@ import { focusManager } from './focusManager'
 import { onlineManager } from './onlineManager'
 import { notifyManager } from './notifyManager'
 import { infiniteQueryBehavior } from './infiniteQueryBehavior'
-import type { DataTag, dataTagSymbol } from './types'
+import type { DataTag, NoInfer } from './types'
 import type { QueryState } from './query'
 import type {
   CancelOptions,
@@ -108,12 +108,16 @@ export class QueryClient {
     return this.#mutationCache.findAll({ ...filters, status: 'pending' }).length
   }
 
-  getQueryData<TaggedQueryKey extends DataTag<unknown, unknown>>(
-    queryKey: TaggedQueryKey,
-  ): TaggedQueryKey[typeof dataTagSymbol] | undefined
-  getQueryData<TQueryFnData = unknown>(
-    queryKey: QueryKey,
-  ): TQueryFnData | undefined
+  getQueryData<
+    TQueryFnData = unknown,
+    TaggedQueryKey extends QueryKey = QueryKey,
+    TInferredQueryFnData = TaggedQueryKey extends DataTag<
+      unknown,
+      infer TaggedValue
+    >
+      ? TaggedValue
+      : TQueryFnData,
+  >(queryKey: TaggedQueryKey): TInferredQueryFnData | undefined
   getQueryData(queryKey: QueryKey) {
     return this.#queryCache.find({ queryKey })?.state.data
   }
@@ -142,22 +146,23 @@ export class QueryClient {
       })
   }
 
-  setQueryData<TaggedQueryKey extends DataTag<unknown, unknown>>(
+  setQueryData<
+    TQueryFnData = unknown,
+    TaggedQueryKey extends QueryKey = QueryKey,
+    TInferredQueryFnData = TaggedQueryKey extends DataTag<
+      unknown,
+      infer TaggedValue
+    >
+      ? TaggedValue
+      : TQueryFnData,
+  >(
     queryKey: TaggedQueryKey,
-    updater: Updater<
-      TaggedQueryKey[typeof dataTagSymbol] | undefined,
-      TaggedQueryKey[typeof dataTagSymbol] | undefined
-    >,
-    options?: SetDataOptions,
-  ): TaggedQueryKey[typeof dataTagSymbol]
-  setQueryData<TQueryFnData>(
-    queryKey: QueryKey,
-    updater: Updater<TQueryFnData | undefined, TQueryFnData | undefined>,
-    options?: SetDataOptions,
-  ): TQueryFnData | undefined
-  setQueryData<TQueryFnData>(
-    queryKey: QueryKey,
-    updater: Updater<TQueryFnData | undefined, TQueryFnData | undefined>,
+    updater:
+      | NoInfer<TInferredQueryFnData>
+      | undefined
+      | ((
+          prev: TInferredQueryFnData | undefined,
+        ) => NoInfer<TInferredQueryFnData> | undefined),
     options?: SetDataOptions,
   ): TInferredQueryFnData | undefined {
     const query = this.#queryCache.find<TInferredQueryFnData>({ queryKey })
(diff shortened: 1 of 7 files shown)


NetanelBasal:
@TkDodo any ETA for this?

TkDodo:
I don't think there's a need to switch to this, the current implementation works fine. This one is based on an unused generic, which is a bit risky because typescript can choose to "forget" it. Is there any reason you're asking for it?

NetanelBasal:
Yes. I finished the migration to v5 but the `DataTag` approach isn't working for me. Btw, the svelte adapter doesn't use it. I'm not sure how it works for you as the signature of the `queryOptions queryKey` and the `useQuery queryKey` isn't the same and typescript complains.

NetanelBasal:
You can see a reproduction [here](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcCKBXYVEGliJwC8cASsAIYAmEAdgDaICCUUFiAPBrQNa0QB3WgD4AUKKQo4AIQoBnYJmyIA8mBgBLOnOJwA3qLhwAjlhz5EALnRm8BUQF9xAY23wI6rbTnXZCpThqmtq6BkamyhbWANoA5BSxALpw8nCu3jCO4qCQsHBUwM70FFCo6XLwVBQwFAAqFADmAMqIALYARhD01twaEXBybZ30ANzZ4NDwkqgAItV1jRy1yMAANHAAahT0WMK6y1IAZPqGcNFVNfXNQ12J1ls7qFmiAGbczsG0cA3AMEFeciWATshFAMGAtCoOmBFl0MII61qcxqum4fEEIgAFB5Pj59CZbFE4MiFg0gYSEXAkfM9g4AJT6JyiH5-TzabFs7x0oA)

TkDodo:
you probably need a conditional type like we have it here: https://github.com/TanStack/query/blob/b67107114ce6ccf0d38c2ab22b3088df19ec119b/packages/query-core/src/queryClient.ts#L111-L123

NetanelBasal:
But you don't have it in the useQuery signature, so how does it works for you?

TkDodo:
- queryOptions creates a DataTag
- useQuery takes a QueryKey
- you can pass a DataTag where a QueryKey is expected
- we don't have functions that expect a DataTag to be passed, because that would limit us to accept something created by queryOptions. Instead, the functions that could benefit from the type on the DataTag expect a generic that extends QueryKey where we then have the conditional type.

NetanelBasal:
> you can pass a DataTag where a QueryKey is expected

I understand how it works in getQueryData, but how it works with `useQuery(queryOptions)`? You can see in my basic [reproduction](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcCKBXYVEGliJwC8cASsAIYAmEAdgDaICCUUFiAPBrQNa0QB3WgD4AUKKQo4AIQoBnYJmyIA8mBgBLOnOJwA3qLhwAjlhz5EALnRm8BUQF9xAY23wI6rbTnXZCpThqmtq6BkamyhbWANoA5BSxALpw8nCu3jCO4qCQsHBUwM70FFCo6XLwVBQwFAAqFADmAMqIALYARhD01twaEXBybZ30ANzZ4NDwkqgAItV1jRy1yMAANHAAahT0WMK6y1IAZPqGcNFVNfXNQ12J1ls7qFmiAGbczsG0cA3AMEFeciWATshFAMGAtCoOmBFl0MII61qcxqum4fEEIgAFB5Pj59CZbFE4MiFg0gYSEXAkfM9g4AJT6JyiH5-TzabFs7x0oA) that typescript is complains (and it makes sense).

TkDodo:
I don't understand your basic reproduction because it doesn't do what we're doing in useQuery:

```
options: { queryKey: DataTag<TQueryKey, TData> }
```

I guess  `getOptions` should be `useQuery` ? If so, you're expecting the options that are passed in to to have a queryKey that is of type `DataTag`, and as I tried to say, that's not what we're doing.

NetanelBasal:
@TkDodo I make it more clear. Check [this](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcCKBXYVEGliJwC8cASsAIYAmEAdgDaICCUUFiAPBrQNa0QB3WgD4AUKKQo4AIQoBnYJmyIA8mBgBLOnOJwA3qLhwAjlhz5EALnRm8BUQF9xoSLDhVgAY3oUoqT9rwVBQwFAAqFADmAMqIALYARhD01twapqhy8Un0ANzO4NDwkqgAIiHhURxhyMAANHAAahT0WMK6NVIAZPqGcADawaERMdnJALrWza2ojuIAZtyemnQmtmortHLVSuYEDWHlocIAFBDqWlvWsgq7qhfaAJTWemvKFtZHlZE7thYHX3aDl6RiMfhgGCgtDgAEY4PJ4bREHNRAEtsUIDQ5BtLjoSBkcDjtCcDKC3nsrAMAOQUKnjeE6NFyGCOR4LJabOAYW62X7vAhwUAwYC0Kg6O4WXQS-ZwQ4VXTcPiCERnB5XfTkuyUm6KdZquQOR76Jyibm65QnGCYiDY-WPIA) please

TkDodo:
I see. The queryKey returned from our queryOptions functions is an intersection of QueryKey and DataTag. [Fixed playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcCKBXYVEGliJwC8cASsAIYAmEAdgDaICCUUFiAPBrQNa0QB3WgD4AUKKQo4AIQoBnYJmyIA8mBgBLOnOJwA3qLhwAjlhz5EALnRm8BUQF9xoSLDhVgAY3oUoqT9rwVBQwFAAqFADmAMqIALYARhD01twapqhy8Un0ANzO4NDwkqgAIiHhURxhyMAANHAAahT0WMK6NVIAZPqGcADawaERMdnJALrWza2ojuIAZtyemnQmtmortHLVSuYEDWHlocIAFBDqWlvWsgq7qhfaAJTWemvKFtZ3FnA9R5WRO1sFgOf3aDl6RiMfhgGCgtDgAEY4PJkbREHNRAEtsUIDQ5BtLjoSBkcATtCcDJC3nsrAMAOQUOnjZE6LFyGCOR4LJabOAYW62QHvAhwUAwYC0Kg6L4ikgyxAgiq6bh8QQiM4PK4yeSKdaanSPfROUT83XKE4wXEQfH6x5AA)

NetanelBasal:
But that's not in your code 🤔

https://github.com/TanStack/query/blob/main/packages/react-query/src/queryOptions.ts#L45

TkDodo:
it is because it's an intersection:

```
DefinedInitialDataOptions<TQueryFnData, TError, TData, TQueryKey> & {
  queryKey: DataTag<TQueryKey, TData>
}
```

and `DefinedInitialDataOptions` defines `QueryKey` as `TQueryKey`, so the two get intersected

NetanelBasal:
That's correct. I don't know what I'm missing. Anyway, thanks, I'll dig into it.

roblabat:
> Fixed playground



> I don't think there's a need to switch to this, the current implementation works fine. This one is based on an unused generic, which is a bit risky because typescript can choose to "forget" it. Is there any reason you're asking for it?

@TkDodo I don't understand this point that typescript could choose to "forget" it, what do you mean by that ? Maybe I'm missing something but I don't see any problem on this...

For me there is two reasons to migrate to this approach: 
- First the DataTag approach is defining a type which is not compatible with the execution object it refers as our key will never have the [DataTagSymbol] property. Our implementation force typescript to assume the key has this property but I don't think it's a good practice as it breaks typesafety (in our case it could be acceptable as we have no reason to use the [DataTagSympol] property but it's still not a good pratice)
- The second reason which for is the most important is that the other approach could enable an other way of using this by typing the key directly as we could define the options without the queryOptions this way: 

``` ts
const TodoKey = ['todos'] as TypedQueryKey<Todo[]>;

useQuery({ QueryKey: TodoKey, QueryFn: TodoQueryFn });
```

I personally prefer this definition as I'm sure of my query typing when I define my key but that's more a personal preference.
I still need to use this to have a better point of view on that especially on how it will works with variable dependent queries.

TkDodo:
> First the DataTag approach is defining a type which is not compatible with the execution object it refers as our key will never have the [DataTagSymbol] property. Our implementation force typescript to assume the key has this property but I don't think it's a good practice as it breaks typesafety (in our case it could be acceptable as we have no reason to use the [DataTagSympol] property but it's still not a good pratice)

I don't understand this, can you show an example? We don't expect a key to have the DataTagSymbol, ever. We have two functions that work better if the key has it: `getQueryData` and `setQueryData`. But if the key doesn't have it, that's fine, too. Keys can be just `Array<string>` as well.

> could enable an other way of using this by typing the key directly

this is not the intention, because it's not type-safe at all. If we change what the `queryFn` returns, this will diverge. What good is a typed query key if we don't infer the types from the source of origin? If you want type assertions, you can just as well do them on the `getQueryData` / `setQueryData` side.

> I don't understand this point that typescript could choose to "forget" it, what do you mean by that ? Maybe I'm missing something but I don't see any problem on this...

I got that info from Andarist (not tagging him because I don't want to drag him into this). Basically, as far as TypeScript is concerned, the generic `Data` is unused:

```
export declare interface TypedQueryKey<Data> extends QueryKey {}
```

so the compiler can choose to eliminate it. So, from TypeScript's perspective, this unused type param doesn't exist since it isn't used by the structure of this type anywhere.

TkDodo:
regarding:

```
const test = todosOptions.queryKey[dataTagSymbol];
```

the symbol is for internal use and we shouldn't expose it to consumers.

---

I wasn't talking about the compiler in terms of "stripping types" for js emit. Of course it strips types here. I was actually referring to the type checker. The Generic on the `TypedQueryKey` type is unused so the type checker can choose to forget it. It doesn't do it at the moment, and it works, and maybe it will never change, but I don't see why this is better than the current approach. It looks like you want it to do some things that weren't intended, like:

 ```
const TodoKey = ['todos'] as TypedQueryKey<Todo[]>;

useQuery({ QueryKey: TodoKey, QueryFn: TodoQueryFn });
```

no, that's not good. The only way to create a typed query key should be via `queryOptions`, and having a private symbol assures this. In that example, if the `QueryFn` changes and returns `NotTodo[]`, you'll get out of sync with this type assertion on `TodoKey`. If you have `queryOptions` and you change the `queryFn` there, the type inferred on the key will change with it, and all the occurrences of `queryClient.getQueryData(TodoKey)` will change, too.

roblabat:
> the symbol is for internal use and we shouldn't expose it to consumers.

Yes if we keep this approche we should at least remove this symbol from exports.
But I'm not sure this symbol will not limit us in more advanced use cases (mainly on keys using variables but I need more investigations on this)

> I wasn't talking about the compiler in terms of "stripping types" for js emit. Of course it strips types here. I was actually referring to the type checker. The Generic on the TypedQueryKey type is unused so the type checker can choose to forget it.

This would be a huge breaking change if typescript choosed ignore that kind of generic types, as the declare keyword is used to tell typescript that you are typing an already existing code the code checker need to keep all given information as it can represent some typing informations that are not represented directly in the object (like our typing linked to the key). Other big libs rely on this mechanism so I don't think this eventuality should be considered as this kind of breaking changes can't be anticipated...

>  It looks like you want it to do some things that weren't intended, like:`const TodoKey = ['todos'] as TypedQueryKey<Todo[]>;
useQuery({ QueryKey: TodoKey, QueryFn: TodoQueryFn });`
no, that's not good. The only way to create a typed query key should be via queryOptions, and having a private symbol assures this. In that example, if the QueryFn changes and returns NotTodo[], you'll get out of sync with this type assertion on TodoKey. If you have queryOptions and you change the queryFn there, the type inferred on the key will change with it, and all the occurrences of queryClient.getQueryData(TodoKey) will change, too.

That's exactly what I want to do but it doesn't break type safety. Yes if the queryFn returns change it will not be propagated to occurrences of queryClient.getQueryData(TodoKey) but it will be thrown directly on the queryOptions definition as I want to check this directly in the useQuery and queryOption functions parameters. I made a quick example of this [here](https://www.typescriptlang.org/play?#code/C4TwDgpgBAigrhATiA0hEUC8UBKECGAJgPYB2ANiAIKKL4gA8cpA1qcQO6kB8AUL4QgBjcvkTQAlqWBIAZviHQAKuAiF4SVOgZLuUAN4BffqEhQAQvgDOEDcgDyYYBLJWdAEXzB8e7Pt5QUACOCMhoIABcUCqQ6qFajEqe3noAPrDx4QHB8QBipFEAFACUWHoACojEALYSNh5ePrzG-LLMQs5kOZqOnaRuSnYJADTRyT6FxE4u-VGWNkO9MwPj3MVR9rXADPO28UuuDSmjAOQhmuEnegBkBt1h6FExakPhRz5QhgbZgeLAcIhSFAAIxQaxg0ggZqtdp9KBwBbxHSvdBQCAADxkpEIVgyF1R2BRIFGSUaWHhrHYXG4k2mrjm1j2PTp-XeelK+igfwBQKmfSsADpzsh8iUWgB6cVQIQA8TSeFWfAAc0kuPYwCgACMqiwIKR+EJXBrgMQSFYDv1ycKQBarIV-IFAtbwlEANonfAnAC6wx+9xA+Si1hApCEUBKmD0JwAFhByORiCdmsUDUaoLqQKDsCazbahZl0ABuXiSx1QAB6AH5eIb+saIFZgFmFUzkIUc8RzSyrKUS1LHVX+KXSBAOArlZJSLIkFBTA3pcRaMJgJRU3X0+gAEzk92a71g3HPOL4xiNxBSJXcYulgfVmtpmSN7fYBGtkD27LOx4bkCb31OvICgPEMwxKMooBjOME03JNDGKYteEfYBNz7MtB1QkcxwRCdZ2jKoxyQKpECgLpNSIf18i5CB-kBKx73XDMAGYdxOPcvQPaJVGPB5T2Ac9SEva9+0CdDa0bWcG2AZjsGtW0PwAk8ogzP9P0AoMrBA8NSkjEFkwQsT60bAAWclXyGeT-RdH8VIUkUgODUMtPAlC4OLIA) see the errors on test3 and test4 queryFn returns that doesn't match the expected return type extracted from the queryKey. I'm not saying this way of defining the queryOptions is better or worst but in some use case it could make sense. For example in my team multiple developers and screens relies on the same queries so we bundled the implementations of or useQuery function in custom hooks that are called in components. Using this new patern would make any change on the queryFn to throw directly in our custom hook, as it's return type is not any more the expected one, instead of throwing in all components using it, this behavior would be easier to troubleshoot for the developer that made a change to the queryFn but is not responsible for all its implementations in components. I think it would be a good think to make the lib compatible with both approaches if approaches are both typesafe. What to you think of this ?

roblabat:
@TkDodo any updates on this ?

TkDodo:
as it stands, the symbol is fine and this discussion is pretty stale, so closing.

PR #6205, opened 2023-10-20
Fix solid-frontpage-example

birkskyum:
This renames the solid basic-typescript example to basic, so that it loads on tanstack query frontpage.

Also, it's aligned with svelte / vue, which also have the typescript examples called 'basic'


TkDodo:
not sure why this never got merged, sorry for that. we re-structured the docs, let me know if this is still an issue. if renaming fixes it (every framework should have the `basic` example), we should do it.

PR #6230, opened 2023-10-24
feat(solid-query): Add useMutationState

brenelz:
Add useMutationState to solid-query. At lot of the tests and functionality were copied over from React Query

image

### DIFF
--- a/packages/solid-query/src/__tests__/useMutationState.test.tsx
+++ b/packages/solid-query/src/__tests__/useMutationState.test.tsx
@@ -1,0 +1,90 @@
+import { describe, expect, expectTypeOf, it } from 'vitest'
+import { fireEvent, render, waitFor } from '@solidjs/testing-library'
+import { createEffect } from 'solid-js'
+import { useMutationState } from '../useMutationState'
+import { createMutation } from '../createMutation'
+import { QueryClientProvider } from '../QueryClientProvider'
+import { createQueryClient, doNotExecute, sleep } from './utils'
+import type { MutationState, MutationStatus } from '@tanstack/query-core'
+
+describe('useMutationState', () => {
+  describe('types', () => {
+    it('should default to QueryState', () => {
+      doNotExecute(() => {
+        const result = useMutationState(() => ({
+          filters: { status: 'pending' },
+        }))
+
+        expectTypeOf<Array<MutationState>>(result())
+      })
+    })
+    it('should infer with select', () => {
+      doNotExecute(() => {
+        const result = useMutationState(() => ({
+          filters: { status: 'pending' },
+          select: (mutation) => mutation.state.status,
+        }))
+
+        expectTypeOf<Array<MutationStatus>>(result())
+      })
+    })
+  })
+  it('should return variables after calling mutate', async () => {
+    const queryClient = createQueryClient()
+    const variables: Array<Array<unknown>> = []
+    const mutationKey = ['mutation']
+
+    function Variables() {
+      const states = useMutationState(() => ({
+        filters: { mutationKey, status: 'pending' },
+        select: (mutation) => mutation.state.variables,
+      }))
+
+      createEffect(() => {
+        variables.push(states())
+      })
+
+      return null
+    }
+
+    function Mutate() {
+      const mutation = createMutation(() => ({
+        mutationKey,
+        mutationFn: async (input: number) => {
+          await sleep(150)
+          return 'data' + input
+        },
+      }))
+
+      return (
+        <div>
+          data: {mutation.data ?? 'null'}
+          <button onClick={() => mutation.mutate(1)}>mutate</button>
+        </div>
+      )
+    }
+
+    function Page() {
+      return (
+        <div>
+          <Variables />
+          <Mutate />
+        </div>
+      )
+    }
+
+    const rendered = render(() => (
+      <QueryClientProvider client={queryClient}>
+        <Page />
+      </QueryClientProvider>
+    ))
+
+    await waitFor(() => rendered.getByText('data: null'))
+
+    fireEvent.click(rendered.getByRole('button', { name: /mutate/i }))
+
+    await waitFor(() => rendered.getByText('data: data1'))
+
+    expect(variables).toEqual([[], [1], []])
+  })
+})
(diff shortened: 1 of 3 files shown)


packages/solid-query/src/createMutationState.ts
```diff
@@ -0,0 +1,59 @@
...
+            (mutation): TResult =>
+                (options.select
+                    ? options.select(
+                        mutation as Mutation<unknown, DefaultError, unknown, unknown>,
+                    )
+                    : mutation.state) as TResult,
+        )
+}
+
+export function createMutationState<TResult = MutationState>(
```
ardeora, line 36: Thanks for working on this! This is looking great! A few things. We should still call this hook `useMutationState` since we're still using data from already created state. Similar to how `useIsMutating` is named!

packages/solid-query/src/createMutationState.ts
```diff
@@ -0,0 +1,59 @@
...
+    options: Accessor<MutationStateOptions<TResult>> = () => ({}),
+    queryClient?: Accessor<QueryClient>,
+): Array<TResult> {
```
ardeora, line 39: This should return an `Accessor<TResult>`. I think this will also fix the failing test too. Since solid hooks are run once, We need to preserve it's reactivity by returning a signal instead

packages/solid-query/src/createMutationState.ts
```diff
@@ -0,0 +1,59 @@
...
+    const client = createMemo(() => useQueryClient(queryClient?.()))
+    const mutationCache = createMemo(() => client().getMutationCache())
+
+    let result = getResult(mutationCache(), options());
```
ardeora, line 43: Based on the comments above this should be 
```ts
const [result, setResult] = createSignal(getResult(mutationCache(), options()))
```

packages/solid-query/src/createMutationState.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    const unsubscribe = mutationCache().subscribe(() => {
+        const nextResult = replaceEqualDeep(
+            result,
+            getResult(mutationCache(), options()),
+        )
+        if (result !== nextResult) {
+            result = nextResult;
```
ardeora, line 51: Should be 
```ts
setResult(nextResult)
```

ardeora (changes requested):

packages/solid-query/src/createMutationState.ts
brenelz: Made the suggested changes. Still having some trouble with the pushing to variables and the following

```
expect(variables).toEqual([[], [1], []])
```

brenelz: I think I got them passing? Added a createEffect

packages/solid-query/src/__tests__/useMutationState.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+            createEffect(() => {
+                const result = useMutationState(() => ({
+                    filters: { mutationKey, status: 'pending' },
+                    select: (mutation) => mutation.state.variables,
+                }));
+
+                variables.push(
+                    result()
+                )
+            })
```
ardeora, line 45: Should be 

```ts
const states = useMutationState(() => ({
    filters: { mutationKey, status: 'pending' },
    select: (mutation) => mutation.state.variables,
  }))

  createEffect(() => {
    variables.push(states())
  })
```

packages/solid-query/src/useMutationState.ts
```diff
@@ -0,0 +1,58 @@
...
+import {
+    type DefaultError,
+    type Mutation,
+    type MutationCache,
+    type MutationFilters,
+    type MutationState,
+    type QueryClient,
+    replaceEqualDeep,
+    useQueryClient,
+} from "."
```
ardeora, line 11: Imports should be like so 
```ts
import { createMemo, createSignal, onCleanup } from 'solid-js'
import { replaceEqualDeep } from '@tanstack/query-core'
import type {
  DefaultError,
  Mutation,
  MutationCache,
  MutationFilters,
  MutationState,
} from '@tanstack/query-core'
import type { Accessor } from 'solid-js'
import type { QueryClient } from './QueryClient'
```

ardeora (changes requested):
Looking great! Last few tweaks

ardeora:
@brenelz Could you also run the formatter? Would just be a `pnpm prettier:write` from the root directory

brenelz:
> @brenelz Could you also run the formatter? Would just be a `pnpm prettier:write` from the root directory

updated

packages/solid-query/src/useMutationState.ts
```diff
@@ -0,0 +1,61 @@
...
+    const nextResult = replaceEqualDeep(
+      result(),
+      getResult(mutationCache(), options()),
+    )
+    if (result() !== nextResult) {
+      setResult(nextResult)
+    }
+  })
+
+  onCleanup(unsubscribe)
```
ardeora, line 58: Last fix I swear Can we wrap this in a `createEffect`?

```ts
createEffect(() => {
    const unsubscribe = mutationCache().subscribe(() => {
    const nextResult = replaceEqualDeep(
      result(),
      getResult(mutationCache(), options()),
    )
    if (result() !== nextResult) {
      setResult(nextResult)
    }
  })

  onCleanup(unsubscribe)
})
```

The reason to do this is that if the `options` object changes in the client code. We should run the sub function with the new `options`

For example:
```ts
// First render
const [count, setCount] = createSignal(0)

const mutationState = useMutationState(() => ({
  mutationKey: ['mutation', count()] // count() is 0
}))
```

```ts
// When count is updated to 2
const [count, setCount] = createSignal(0)

const mutationState = useMutationState(() => ({
  mutationKey: ['mutation', count()] // count() is 2
}))

// This will work now!!!
```

ardeora (changes requested):

packages/solid-query/src/useMutationState.ts
ardeora: @brenelz  last change and we're good to go on this PR

brenelz: Yeah thats a good catch. Should be good now

ardeora (approved):

PR #6245, opened 2023-10-25
fix(query-devtools): Inherit font size from parent instead of root

### LINKED ISSUE #6226: Devtools scale messed up in v5 due to 62.5% trick
### Describe the bug

In the application I am working on we have used the 62.5% trick to make 1rem = 10px. The devtools are styled with rem now and everything has become extremely tiny. Unfortunately, since it is rem, it's not something we can change just for the devtools without affecting the rest of our app.

https://codesandbox.io/p/sandbox/blue-bush-xl39zm

### Steps to reproduce

Apply the following css:
```css
html {
  font-size: 62.5%;
}

body {
  font-size: 1.6rem;
}
```

View the devtools and everything will be very small.

### Expected behavior

The devtools look normal as they did previously

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: Windows
- Browser: Chrome
- Version: 118.0.5993.89

### Tanstack Query adapter

react-query

### TanStack Query version

5.0.5

### TypeScript version

4.9.5

### Additional context

_No response_

ardeora:
Fixes #6226

We cannot use `em` units because the scale of these units change based on the styles of the parent container. So the design tokens would have different behavior for nested elements.

Instead we calculate the computed fontSize style whenever the devtools are mounted and set a CSS variable `--tsqd-font-size`. The rest of the tokens can derive their value based on this CSS variable.

I have also added a window focus listener so that this computation is run again everytime the user focuses back on the browser. So if they change any styles in their stylesheet, we can recompute the inherited font-size.


ardeora:
Opening a new PR

PR #6255, opened 2023-10-26
feat(vue-query): component based vue devtools - Vue3 only

DamianOsipiuk:


Mini-ghost:
How is this different from the built-in development tools of Vue Devtools?

DamianOsipiuk:
> How is this different from the built-in development tools of Vue Devtools?

It has more features. Some of them would be hard to add to `Vue devtools` due to how restrictive it is.  
I will try to add those to `Vue devtools` as well.

This component is a small wrapper around the `core` package built in `solid`, and will automatically receive all devtools updates without the need of porting them.

reslear:
wow, @DamianOsipiuk Fajnie ;) I mean, there's got to be some way to integrate with https://github.com/webfansplz/vite-plugin-vue-devtools or https://devtools.nuxt.com/

PR #6259, opened 2023-10-28
chore: update to nextjs v14

lithdew:
Update to latest NextJS v14.0 for examples and @tanstack/react-query-next-experimental.

### DIFF
--- a/packages/react-query-next-experimental/package.json
+++ b/packages/react-query-next-experimental/package.json
@@ -45,14 +45,14 @@
     "@tanstack/react-query": "workspace:*",
     "@types/react": "^18.2.31",
     "@types/react-dom": "^18.2.14",
-    "next": "^13.5.6",
+    "next": "^14.0.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-error-boundary": "^3.1.4"
   },
   "peerDependencies": {
     "@tanstack/react-query": "workspace:^",
-    "next": "^13.0.0",
+    "next": "^13 || ^14",
     "react": "^18.0.0",
     "react-dom": "^18.0.0"
   }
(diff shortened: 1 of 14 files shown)


packages/react-query-next-experimental/package.json
```diff
@@ -45,14 +45,14 @@
...
-    "next": "^13.5.6",
+    "next": "^14.0.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-error-boundary": "^3.1.4"
   },
   "peerDependencies": {
     "@tanstack/react-query": "workspace:^",
-    "next": "^13.0.0",
+    "next": "^14.0.0",
```
TkDodo, line 55: I think it should be 13 || 14 ?

lithdew: How about >=13.0.0?

lithdew:
Updated peerDependency for @tanstack/react-query-next-experimental for NextJS to `>=13.0.0`.

Set default runtime for nextjs-suspense-streaming example to `edge` to mitigate a bug where prerendering route `/` attempts to fetch `http://localhost:3000/api/wait` on the `nodejs` runtime causes a prerender error at a build-time.

packages/react-query-next-experimental/package.json
```diff
@@ -45,14 +45,14 @@
...
-    "next": "^13.5.6",
+    "next": "^14.0.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-error-boundary": "^3.1.4"
   },
   "peerDependencies": {
     "@tanstack/react-query": "workspace:^",
-    "next": "^13.0.0",
+    "next": ">=13.0.0",
```
TkDodo, line 55: let's be explicit here please. If v15 comes out and this plugin isn't compatible anymore, it's better to fail installs.

lithdew: Updated the peerDependency on next to `^13 || ^14`.

TkDodo:
please fix the formatting

PR #6261, opened 2023-10-28
fix(query-core): ensureQueryData call fetchQuery when the query actually does not exist

LineGu:
# `queryClient.ensureQueryData` 

Bug:
```typescript

  const { data: count } = useQuery({
    queryKey: ['cacheKey'],
    queryFn: async () => {
      // return falsy data. ex) 0 / null / '' / false ...
     return 0
    },
  });

  const handleClick = async () => {
    const data = await queryClient.ensureQueryData({ queryKey: ['cacheKey'] });
    // queryFn runs every time click
  };
```

If the cached data for the existing query is falsey data ( 0, null, "", false ... ) , `queryClient.fetchQuery` will be called when using `queryClient.ensureQueryData`. This leads to unintentional queryFn (API) calls.

So ensure a clear check for the existence of the query's cached data.

### DIFF
--- a/packages/query-core/src/tests/queryClient.test.tsx
+++ b/packages/query-core/src/tests/queryClient.test.tsx
@@ -435,13 +435,13 @@ describe('queryClient', () => {
 
     test('should return the cached query data if the query is found and cached query data is falsey', async () => {
       const key = queryKey()
-      const queryFn = () => Promise.resolve('data')
+      const queryFn = () => Promise.resolve(0)
 
-      queryClient.setQueryData([key, 'id'], '')
+      queryClient.setQueryData([key, 'id'], null)
 
       await expect(
         queryClient.ensureQueryData({ queryKey: [key, 'id'], queryFn }),
-      ).resolves.toEqual('')
+      ).resolves.toEqual(null)
     })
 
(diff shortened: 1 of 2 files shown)


packages/query-core/src/tests/queryClient.test.tsx
```diff
@@ -433,6 +433,17 @@ describe('queryClient', () => {
       ).resolves.toEqual('bar')
     })
 
+    test('should return the cached query data if the query is found and cached query data is falsey', async () => {
+      const key = queryKey()
+      const queryFn = () => Promise.resolve('data')
```
TkDodo, line 438: good fix, but the returned data here isn't falsy?

LineGu: Thanks. My intention was to test the case where the cache data is falsey and the new data is not falsey.

LineGu: 0557237b22bf09cb9b4d0e5b15d0efe1b997dce1
make the test case more intuitive, and I have fixed all the returned data covered by the test case to be falsy values.
If there are any parts that need improvement, please let me know.

TkDodo (approved):

PR #6270, opened 2023-10-31
fix(mutationCache): export MutationCacheNotifyEvent type

LucaDiba:
Export `MutationCacheNotifyEvent` type to make it consistent with `QueryCacheNotifyEvent` (which is exported) and make it easier to write listener functions.

https://github.com/TanStack/query/blob/a320b4301078ea0bbf190e3f5c0e032f1c602e1b/packages/query-core/src/queryCache.ts#L71

We could even think about exporting `QueryCacheListener` and `MutationCacheListener` as well to give the correct type to the entire listener function.

https://github.com/TanStack/query/blob/a320b4301078ea0bbf190e3f5c0e032f1c602e1b/packages/query-core/src/queryCache.ts#L80

https://github.com/TanStack/query/blob/a320b4301078ea0bbf190e3f5c0e032f1c602e1b/packages/query-core/src/mutationCache.ts#L80


packages/query-core/src/mutationCache.ts
```diff
@@ -69,7 +69,7 @@ interface NotifyEventMutationUpdated extends NotifyEvent {
   action: Action<any, any, any, any>
 }
 
-type MutationCacheNotifyEvent =
+export type MutationCacheNotifyEvent =
```
TkDodo, line 72: the index barrel on `query-core/index.ts` needs to re-export this to make it available to consumers. We don't magically re-export everything as that could leak internals.

PR #6288, opened 2023-11-02
Better error in dev to identify bad useQuery calls

GLabat:
Starting v5, useQuery only accepts the Object form argument => https://tanstack.com/query/latest/docs/react/guides/migrating-to-v5#supports-a-single-signature-one-object.

 For projects without TS transitioning from v4 to v5, it is "easy" to miss some changes in useQuery calls. When this happens, an uncaught error is raised: `client.defaultQueryOption is not a function`.

To help developers in this case, a more meaningful error is now raised to help identifying such bad calls.

This was discussed here: https://discord.com/channels/719702312431386674/1168590642482528286

### DIFF
--- a/packages/react-query/src/__tests__/useQuery.test.tsx
+++ b/packages/react-query/src/__tests__/useQuery.test.tsx
@@ -6238,7 +6238,8 @@ describe('useQuery', () => {
 
     function Page() {
       // Invalid call on purpose
-      const { data } = useQuery(key, { queryFn })
+      // @ts-expect-error
+      useQuery(key, { queryFn })
       return <div>Does not matter</div>
     }
 
--- a/packages/react-query/src/__tests__/useQuery.test.tsx
+++ b/packages/react-query/src/__tests__/useQuery.test.tsx
@@ -6225,4 +6225,24 @@ describe('useQuery', () => {
 
     await waitFor(() => rendered.getByText('Works'))
   })
+
+  // For Project without TS, when migrating from v4 to v5, make sure invalid calls due to bad parameters are tracked.
+  it('should throw in case of bad arguments to enhance DevX', async () => {
+    // Mock console error to avoid noise when test is run
+    const consoleMock = vi
+      .spyOn(console, 'error')
+      .mockImplementation(() => undefined)
+
+    const key = queryKey()
+    const queryFn = () => 'data'
+
+    function Page() {
+      // Invalid call on purpose
+      const { data } = useQuery(key, { queryFn })
+      return <div>Does not matter</div>
+    }
+
+    expect(() => render(<Page />)).toThrow('Bad argument type')
+    consoleMock.mockRestore()
+  })
 })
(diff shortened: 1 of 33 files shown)


packages/react-query/src/useBaseQuery.ts
```diff
@@ -31,6 +31,14 @@ export function useBaseQuery<
   Observer: typeof QueryObserver,
   queryClient?: QueryClient,
 ) {
+  if (process.env.NODE_ENV !== 'production') {
+    if (!isPlainObject(options)) {
```
TkDodo, line 35: I think we should inline the check here to avoid making the `isPlainObject` function exported from the core. A simple check like this should suffice:

```
!!options && !Array.isArray(options) && typeof options === 'object'
```

we could also leave out the Array check because that was never a supported syntax in the first place :)

GLabat: Yep I agree. I did this at first and then saw `isPainObject` and I gave it a chance :).
Will change this later.

Thx.

TkDodo: yep 👍

GLabat: pushed

packages/react-query/src/__tests__/useQuery.test.tsx
```diff
@@ -6225,4 +6225,24 @@ describe('useQuery', () => {
...
+    const consoleMock = vi
+      .spyOn(console, 'error')
+      .mockImplementation(() => undefined)
+
+    const key = queryKey()
+    const queryFn = () => 'data'
+
+    function Page() {
+      // Invalid call on purpose
+      const { data } = useQuery(key, { queryFn })
```
TkDodo, line 6241: doesn't this need a `@ts-expect-error` ?

TkDodo: yeah, there are two errors:

```
src/__tests__/useQuery.test.tsx(6241,13): error TS6133: 'data' is declared but its value is never read.
src/__tests__/useQuery.test.tsx(6241,33): error TS2769: No overload matches this call.
```

GLabat: … not familiar enough with TS… will fix soon.

GLabat: Ok so tests pass locally using: `pnpm test` in root dir and `pnpm test:lib` in packages/react-query dir.
But CI seems to fail…

Any idea?

GLabat: A race condition in useQueries tests?

TkDodo (approved):

packages/react-query/src/__tests__/useQuery.test.tsx
TkDodo: flaky tests :(

PR #6302, opened 2023-11-03
feat(eslint-plugin): add no-rest-destructuring rule (#6265)

alan910127:
This pull request adds an eslint rule `no-rest-destructuing` to make sure that the query results are not captured by a rest element.

Discussion: #6265 

I referenced the commit for `stable-query-client` rule for what I should do to make a new eslint rule in tanstack/eslint-plugin-query, please inform me if I still missed something. 😄

### DIFF
--- a/packages/eslint-plugin-query/src/rules/no-rest-desctructuring/no-rest-destructuring.utils.ts
+++ b/packages/eslint-plugin-query/src/rules/no-rest-desctructuring/no-rest-destructuring.utils.ts
@@ -1,0 +1,11 @@
+import { AST_NODE_TYPES } from '@typescript-eslint/utils'
+import type { TSESTree } from '@typescript-eslint/utils'
+
+export const NoRestDestructuringUtils = {
+  isObjectRestDestructuring(node: TSESTree.Node): boolean {
+    if (node.type !== AST_NODE_TYPES.ObjectPattern) {
+      return false
+    }
+    return node.properties.some((p) => p.type === AST_NODE_TYPES.RestElement)
+  },
+}
--- a/docs/config.json
+++ b/docs/config.json
@@ -310,6 +310,10 @@
             {
               "label": "Stable Query Client",
               "to": "react/eslint/stable-query-client"
+            },
+            {
+              "label": "No Rest Destructuring",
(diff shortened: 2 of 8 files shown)


TkDodo (approved):

alan910127:
@TkDodo Thanks for your approval!
Furthermore, as I am not quite familiar with creating eslint rules, I cannot figure out how to write the autofix for this rule. 
Hope someone can help make this rule better by introducing the autofix option.

TkDodo:
I don't think we can or need to auto fix this because we can't really know what it should be auto fixed towards?

alan910127:
That's right, thank you!

packages/eslint-plugin-query/src/rules/no-rest-desctructuring/no-rest-destructuring.utils.ts
```diff
@@ -0,0 +1,11 @@
+import { AST_NODE_TYPES } from '@typescript-eslint/utils'
+import type { TSESTree } from '@typescript-eslint/utils'
+
+export const NoRestDestructuringUtils = {
+  isObjectRestDestructuring(node: TSESTree.Node): boolean {
```
Newbie012, line 5: I would rename it to hasRestElement and move it to ASTUtils, but that's totally fine as well.

Newbie012 (commented):
I would add another test case that allows destructing non-query hook with the same name (such as useQuery of chakra-ui).

Other than that, LGTM

packages/eslint-plugin-query/src/rules/no-rest-desctructuring/no-rest-destructuring.utils.ts
alan910127: I think hasRestElement would be confused with the array rest destructuring `[a, ...rest]`, but I'm okay with moving it to ASTUtils.

alan910127:
@Newbie012 I have implemented the test cases for allowing rest destructuring hooks from other packages with the name `useQuery`, `useInfiniteQuery` and `useQueries`.

Newbie012 (approved):

Issue #6298, opened 2023-11-03
useSuspenseQueries observers count mismatch


maciej-baruch:
### Describe the bug

When the amount of queries in `useSuspenseQueries` increases, the new queries resolution/suspense is split into two batches.
The first batch takes care of `newQueries.slice(0, previousQueries.length)` and stops the suspense, causing the rest of the queries to be still pending (and have undefined data) before the second batch (rest of the `newQueries`) kicks in.

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/friendly-jang-gjp9p8

### Steps to reproduce

1. Click the button.
2. Observe the lifecycle of queries in the console.

### Expected behavior

As a user, I would expect `newQueries` to be resolved all at once in a single suspense.

### How often does this bug happen?

Every time

### Platform

Platform agnostic.

### Tanstack Query adapter

react-query

### TanStack Query version

5.4.3

### TypeScript version

5.2.2

maciej-baruch:
I'd like to attempt to fix that. I will keep you posted.

TkDodo:
the problem is even easier to see if we move the devtools out of the suspense boundary:

https://codesandbox.io/p/sandbox/fast-sound-53plff?file=%2Fsrc%2Findex.jsx%3A9%2C1

Feel free to look into this. I also have an idea where we might go wrong :)

maciej-baruch:
I was thinking about simply removing `useEffect` around `setObservers` but this seemed like an unnecessary performance cost.

Added some tests to ensure I did not break anything.
Let me know if you think I should cover some more cases or if you have any doubts.

PR #6303, opened 2023-11-03
fix(react-query): non continuous suspense with useSuspenseQueries (#6298)

maciej-baruch:
#6298


TkDodo (commented):
I had this fix in mind:

```diff
Subject: [PATCH] fix: useQueries suspense
---
Index: packages/react-query/src/useQueries.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/packages/react-query/src/useQueries.ts b/packages/react-query/src/useQueries.ts
--- a/packages/react-query/src/useQueries.ts	
+++ b/packages/react-query/src/useQueries.ts	
@@ -1,7 +1,11 @@
 'use client'
 import * as React from 'react'
 
-import { QueriesObserver, notifyManager } from '@tanstack/query-core'
+import {
+  QueriesObserver,
+  QueryObserver,
+  notifyManager,
+} from '@tanstack/query-core'
 import { useQueryClient } from './QueryClientProvider'
 import { useIsRestoring } from './isRestoring'
 import { useQueryErrorResetBoundary } from './QueryErrorResetBoundary'
@@ -262,9 +266,9 @@
   const suspensePromises = shouldAtLeastOneSuspend
     ? optimisticResult.flatMap((result, index) => {
         const opts = defaultedQueries[index]
-        const queryObserver = observer.getObservers()[index]
 
-        if (opts && queryObserver) {
+        if (opts) {
+          const queryObserver = new QueryObserver(client, opts)
           if (shouldSuspend(opts, result, isRestoring)) {
             return fetchOptimistic(opts, queryObserver, errorResetBoundary)
           } else if (willFetch(result, isRestoring)) {

```

I think we don't need to use the matching observer on the QueriesObserver - the fetch is fire-and-forget anyways, so I just created a `new QueryObserver` to do the fetch.

the `fetchOptimistic` method on the observer is just a small wrapper around `query.fetch()`, so we technically don't even need it ... but changing that could be considered breaking, so this is the most minimal fix I could find. Can you try if your tests also work with that solution?

maciej-baruch:
Nice and simple.
The tests run.

TkDodo (approved):

TkDodo:
There are some warnings with the way the tests are written:

```
Warning: The current testing environment is not configured to support act(...)
```

maybe relates to:
- https://github.com/testing-library/react-testing-library/issues/1061

not sure if we'd need to update testing-library or vitest or both?

Maybe you could also change the tests to be written more like the other tests that we have? We basically don't test hooks, but wrap them in components like so:

https://github.com/TanStack/query/blob/c014ea7a29db5bdd0770733e3c531c9050d7e4e9/packages/react-query/src/__tests__/suspense.test.tsx#L123-L148

maciej-baruch:
Happy to comply.

PR #6306, opened 2023-11-03
Update eslint-plugin-query.md

ProjectFlinn:
TK Mentioned that the docs would be easier to read if the recommended config was above the alternative config.


TkDodo:
Thank you 🙏

PR #6308, opened 2023-11-04
test(asyncThrottle): migrated to useFakeTimers

OrlovAlexei:
Hi @TkDodo,  I was on the twitch stream and asked about using fake timers in tests. This is my first example, on a small codebase.

The test time has noticeably decreased, 4.56 s => 468 ms.

What do you think about it ? I can also migrate other packages gradually.

### DIFF
--- a/packages/query-async-storage-persister/src/__tests__/asyncThrottle.test.ts
+++ b/packages/query-async-storage-persister/src/__tests__/asyncThrottle.test.ts
@@ -1,10 +1,17 @@
-import { describe, expect, test, vi } from 'vitest'
+import { afterEach, describe, expect, test, vi } from 'vitest'
 import { asyncThrottle } from '../asyncThrottle'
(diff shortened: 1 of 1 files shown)


TkDodo:
love it!

small suggestion:

let's put `useFakeTimers` into `beforeEach`, and move the `beforeEach` and `afterEach` block into the top `describe` block

OrlovAlexei:
> love it!
> 
> small suggestion:
> 
> let's put `useFakeTimers` into `beforeEach`, and move the `beforeEach` and `afterEach` block into the top `describe` block

Good idea, I did so 👍🏼

TkDodo:
> and move the beforeEach and afterEach block into the top describe block

I don't think this has happened? I'd want to avoid the beforeEach and afterEach running for more than just the one `describe` block. Not sure what vitest does with a top level thing? Is it for the test file only then?

PR #6309, opened 2023-11-04
refactor: improved multiple types

ssi02014:
@TkDodo 👋
I improved `type` by declaring and utilizing the `MutationActionType` and `QueryActionType` types. 

This is the same as `NotifyEvent` and `NotifyEventType`. What do you think about maintaining consistency?

Additionally, I added a type for the common property `status`.

### DIFF
--- a/packages/query-core/src/types.ts
+++ b/packages/query-core/src/types.ts
@@ -559,6 +559,7 @@ export interface InfiniteQueryObserverBaseResult<
   hasPreviousPage: boolean
   isFetchingNextPage: boolean
   isFetchingPreviousPage: boolean
+  status: QueryStatus
 }
 
 export interface InfiniteQueryObserverLoadingResult<
@@ -732,6 +733,7 @@ export interface MutationObserverBaseResult<
   isSuccess: boolean
   mutate: MutateFunction<TData, TError, TVariables, TContext>
   reset: () => void
+  status: MutationStatus
 }
 
 export interface MutationObserverIdleResult<
--- a/packages/query-core/src/types.ts
+++ b/packages/query-core/src/types.ts
@@ -841,3 +841,21 @@ export type NotifyEventType =
 export interface NotifyEvent {
   type: NotifyEventType
 }
+
+export type MutationActionType =
+  | 'failed'
+  | 'pending'
+  | 'success'
+  | 'error'
+  | 'pause'
+  | 'continue'
+
+export type QueryActionType =
+  | 'failed'
+  | 'fetch'
+  | 'success'
+  | 'error'
+  | 'invalidate'
+  | 'pause'
+  | 'continue'
+  | 'setState'
--- a/packages/query-core/src/types.ts
+++ b/packages/query-core/src/types.ts
@@ -859,3 +859,11 @@ export type QueryActionType =
   | 'pause'
   | 'continue'
   | 'setState'
+
+export interface MutationAction {
+  type: MutationActionType
+}
+
+export interface QueryAction {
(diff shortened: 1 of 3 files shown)


packages/query-core/src/types.ts
```diff
@@ -732,6 +733,7 @@ export interface MutationObserverBaseResult<
   isSuccess: boolean
   mutate: MutateFunction<TData, TError, TVariables, TContext>
   reset: () => void
+  status: MutationStatus
```
ssi02014, line 736: 스크린샷 2023-11-05 오전 3 44 05

I think this is safer and clearer because if you actually put in an incompatible state, it will throw an error.

ssi02014: The other tasks have similar intentions.
They're meant to make your code more consistent, and they can be clearer and safer.

packages/query-core/src/types.ts
```diff
@@ -841,3 +843,29 @@ export type NotifyEventType =
...
+
+export type QueryActionType =
+  | 'failed'
+  | 'fetch'
+  | 'success'
+  | 'error'
+  | 'invalidate'
+  | 'pause'
+  | 'continue'
+  | 'setState'
```
TkDodo, line 863: the idea is that `Action` defines the potential actions, and the type can be inferred by doing:

```
type QueryAction = Action["type"]
```

this makes `Action` the single source of truth. If we add one action, it also becomes a valid type. With this change, we'd need to define it in two places. I don't see the advantage of that.

ssi02014:
```ts
type QueryActionType<TData, TError> = Action<TData, TError>['type']
```
Does this mean that it can be utilized in the same way as above? I understood your comment.

https://github.com/TanStack/query/pull/6309#discussion_r1382442647 
I initially thought it was safer because it throws a type error when written incompatibly.

Crucially, I also thought it was the right thing to do for consistency, since "queryCache" was also done this way.

```ts
interface NotifyEventQueryAdded extends NotifyEvent {
  type: 'added'
  query: Query<any, any, any, any>
}
```
```ts
export type NotifyEventType =
  | 'added'
  | 'removed'
  | 'updated'
  | 'observerAdded'
  | 'observerRemoved'
  | 'observerResultsUpdated'
  | 'observerOptionsUpdated'

export interface NotifyEvent {
  type: NotifyEventType
}
```

or `QueryObserverBaseResult`

```ts
export type QueryStatus = 'pending' | 'error' | 'success'
export type FetchStatus = 'fetching' | 'paused' | 'idle'

export interface QueryObserverBaseResult<
  TData = unknown,
  TError = DefaultError,
> {
  // ...
  status: QueryStatus
  fetchStatus: FetchStatus
}
```
```ts
export interface QueryObserverLoadingResult<
  TData = unknown,
  TError = DefaultError,
> extends QueryObserverBaseResult<TData, TError> {
  // ...
  status: 'pending'
}

export interface QueryObserverLoadingErrorResult<
  TData = unknown,
  TError = DefaultError,
> extends QueryObserverBaseResult<TData, TError> {
  // ...
  status: 'error'
}
```

ssi02014: @TkDodo Doesn't it make sense to make our code consistent?

TkDodo: those are public interfaces, and we need the event types for queries and mutations to be the same

ssi02014:
@TkDodo 
In your opinion, `NotifyEventType`, `NotifyEvent` should also be modified, do you agree?

PR #6345, opened 2023-11-09
fix(examples-react-suspense): add font-awesome dependency to display …

krisko4:
…Spinner correctly

Spinner is not visible in the Suspense example due to missing font-awesome dependency.

### DIFF
--- a/examples/react/suspense/package.json
+++ b/examples/react/suspense/package.json
@@ -6,6 +6,7 @@
     "@tanstack/react-query": "^5.0.0",
     "@tanstack/react-query-devtools": "^5.0.0",
     "axios": "^1.5.1",
+    "font-awesome": "^4.7.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-error-boundary": "^3.1.4"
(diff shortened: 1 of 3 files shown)


TkDodo:
You need to update the pnpm lockfile if you add a dependency

krisko4:
I'm sorry, there must be something I'm doing wrong but I am unable to pass the CI step. I've commited a pnpm-lock.yaml file generated after running pnpm install, but now it seems there are some other issues:

 ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with packages/react-query/package.json
 
 This error appears also when I clone the main branch and try running pnpm install --frozen-lockfile in my terminal (without providing any changes).

TkDodo:
are you sure you have the correct version of pnpm ?

https://github.com/TanStack/query/blob/main/CONTRIBUTING.md#development

krisko4:
Thank you very much, wrong pnpm version was the exact reason!

TkDodo (approved):

PR #6346, opened 2023-11-09
Fix generic type passed to structural sharing function

GavinMeierSonos:
I noticed when upgrading that our code, that is using structural sharing, could no longer properly infer the type of the oldData and newData from the query. 

It looks like structural sharing was expanded before v5 to take the type of T instead of TData. I could not find a way to make it work so I thought it must be a bug and decided to open up a PR to fix it. 

If there is a way to fix it without the changes let me know but I did confirm that the type of oldData and newData in the test was not being inferred properly so I added explicit typing to the test that was added when the change went in

### DIFF
--- a/packages/query-core/src/types.ts
+++ b/packages/query-core/src/types.ts
@@ -158,7 +158,7 @@ export interface QueryOptions<
    * Set this to a function which accepts the old and new data and returns resolved data of the same type to implement custom structural sharing logic.
    * Defaults to `true`.
    */
-  structuralSharing?: boolean | (<T>(oldData: T | undefined, newData: T) => T)
+  structuralSharing?: boolean | ((oldData: TData | undefined, newData: TData) => TData)
   _defaulted?: boolean
   /**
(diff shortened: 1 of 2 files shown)


TkDodo:
this was changed in v5 on purpose because we're not only calling `structuralSharing` with data returned from the queryFn - it's also called with the result from `select`, which can be _anything_ . 

If you want a custom implementation, it's best to be prepared to get anything in, which `unknown` accomplishes.

gavinmeier25:
@TkDodo ahh ok that makes sense, so since a generic defaults to an unknown value, it's just being used for the return result. I will just parse it with a zod schema to ensure it's the correct type of data coming back. 

Isn't it possible to infer the returned type of select though from the value that is used? I get the reasoning behind the change, just curious on the situation where the incoming data wouldn't be the generic that is inferred from select being present or not.

TkDodo:
No, there isn't. select can be different per usage (observer), but structuralSharing cannot.

PR #6349, opened 2023-11-09
fix: pass TError, TQueryData and TQueryKey to PlaceholderDataFunction

bviebahn:
The `previousData` and `previousQuery` parameters of the placeholder function seem to not be properly types. I think this should fix it. I tested it like this:
```typescript
const observer = new QueryObserver(
  {} as any,
  {
    queryKey: ['A', 'B'] as const,
    queryFn: () => 'Foo',
    placeholderData: (prev, prevQuery) => {
      // typeof key is now ['A', 'B'] instead of QueryKey
      const key = prevQuery?.queryKey
      return prev
    },
  },
)
```
Please let me know if I missed something.

### DIFF
--- a/packages/query-core/src/types.ts
+++ b/packages/query-core/src/types.ts
@@ -140,7 +140,7 @@ export interface QueryOptions<
   retry?: RetryValue<TError>
   retryDelay?: RetryDelayValue<TError>
   networkMode?: NetworkMode
-   /**
+  /**
    * The time in milliseconds that unused/inactive cache data remains in memory.
    * When a query's cache becomes unused or inactive, that cache data will be garbage collected after this duration.
    * When different garbage collection times are specified, the longest one will be used.
--- a/packages/query-core/src/types.ts
+++ b/packages/query-core/src/types.ts
@@ -319,7 +319,12 @@ export interface QueryObserverOptions<
    */
   placeholderData?:
     | NonFunctionGuard<TQueryData>
-    | PlaceholderDataFunction<NonFunctionGuard<TQueryData>>
+    | PlaceholderDataFunction<
+        NonFunctionGuard<TQueryData>,
+        TError,
+        NonFunctionGuard<TQueryData>,
+        TQueryKey
+      >
 
   _optimisticResults?: 'optimistic' | 'isRestoring'
 }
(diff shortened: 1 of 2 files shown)


bviebahn:
Not sure what to do about the failing tests, running them locally they're all green. I only changed a type, so I think the failing test is not related to my change.

packages/query-core/src/types.ts
```diff
@@ -319,7 +319,12 @@ export interface QueryObserverOptions<
    */
   placeholderData?:
     | NonFunctionGuard<TQueryData>
-    | PlaceholderDataFunction<NonFunctionGuard<TQueryData>>
+    | PlaceholderDataFunction<
+        NonFunctionGuard<TQueryData>,
```
bviebahn, line 323: I'm not sure if this should be `TQueryFnData`?

TkDodo: no I think this is fine. The difference is only for InfiniteQueries, where one is wrapped in `InfiniteData`. But maybe you can add a couple of type tests?

bviebahn: Sure, I added some tests. The last one would have also passed without my changes. Let me know if it needs more work.

TkDodo (approved):

TkDodo:
format is failing

bviebahn:
I don't understand why, when I run `pnpm run test:format` I'm not getting any errors. When I run `pnpm prettier --write` it gets formatted exactly as it currently is. Any idea?

TkDodo:
just run the --write and commit it?

TkDodo:
oh, likely you have the wrong version of prettier:

```
> prettier --plugin-search-dir . "{examples,integrations,packages,scripts}/**/*.{md,js,jsx,cjs,ts,tsx,json,vue,svelte}" "--check"

Checking formatting...
[warn] packages/query-core/src/types.ts
```

bviebahn:
Oh, not sure what the problem was but after cloning the repo again I got the same prettier error. Should be fixed.

TkDodo:
yeah actually, it was me changing something in a PR suggestion on a different PR, and merging it via the GH ui . sorry, it should be fixed on main now

PR #6350, opened 2023-11-10
  Fix #6344: useSuspenseQuery() returns undefined as data when network is offline 

### LINKED ISSUE #6344: useSuspenseQuery() returns undefined as data when network is offline
### Describe the bug

`useSuspenseQuery()` returns undefined as data when network is offline. This is strange because  the type of `useSuspenseQuery()` says that `data` should never be `undefined`.

https://codesandbox.io/p/sandbox/frosty-field-7khlvr?file=%2Fsrc%2FApp.tsx%3A1%2C1

### Steps to reproduce

1. Go to example linked above
2. Open DevTool and modify the network throttling to "Offline"
3. Click "Load" button

### Expected behavior

I expected a promise to be thrown and the fallback ("Loading...") is shown, but actually `undefined` was returned as data.

None

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome 118

### Tanstack Query adapter

None

### TanStack Query version

v5.8.1

### TypeScript version

_No response_

### Additional context

_No response_

incepter:
This is a reproduction for #6344.

I suggest not merging it yet to main since it prints some errors in the console.

### DIFF
--- a/packages/react-query/src/__tests__/suspense.test.tsx
+++ b/packages/react-query/src/__tests__/suspense.test.tsx
@@ -6,6 +6,7 @@ import { onlineManager } from '@tanstack/query-core'
 import {
   QueryCache,
   QueryErrorResetBoundary,
+  useQuery,
   useQueryErrorResetBoundary,
   useSuspenseInfiniteQuery,
   useSuspenseQueries,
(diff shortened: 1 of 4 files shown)


packages/react-query/src/__tests__/suspense.test.tsx
```diff
@@ -887,4 +887,52 @@ describe('useSuspenseQueries', () => {
...
+    function Page() {
+      const [id, setId] = React.useState(0)
+
+      const { data } = useSuspenseQuery({
+        queryKey: [id],
+        queryFn: () => Promise.resolve(`Data ${id}`),
+      })
+
+      // defensive guard here
+      if (data === undefined) {
```
incepter, line 903: Commenting the resolution in `suspense.ts` will make lead to undefined and thus throwing this error

incepter:
@TkDodo Could you please take a look at the PR and the failing test cause ? I don't know actually why it fails on CI. That particular test is passing locally.

TkDodo:
@incepter are you working on fix for the issue? Because I'd like to get a fix released and if you're not looking into it, I would do it :)

incepter:
@TkDodo can you please take a look at the PR. I think this fixes it. If you have any remarks please let me know.

packages/react-query/src/suspense.ts
```diff
@@ -41,7 +41,12 @@ export const shouldSuspend = (
     | undefined,
   result: QueryObserverResult<any, any>,
   isRestoring: boolean,
-) => defaultedOptions?.suspense && willFetch(result, isRestoring)
+) => {
+  if (defaultedOptions?.suspense && result.isPending && !isRestoring) {
+    return true
+  }
+  return defaultedOptions?.suspense && willFetch(result, isRestoring)
```
TkDodo, line 48: I think we can just simplify this to:

```
return defaultedOptions?.suspense && result.isPending
```

because we really want to suspend every time when we have no data, which `isPending` guarantees.

the `!isRestoring` will have the same problem as the offline mode - while restoring from external storage, we would render the component with `undefined` ... that is desired for normal useQuery, but not for suspense

TkDodo:
oh sorry, I thought this was only the issue reproduction, not also the fix :)

- formatting needs fixing please
- ignore the error, I think it was flakiness (i know, i know ...)

packages/react-query/src/suspense.ts
incepter: OK on it.

TkDodo (approved):

TkDodo:
ah yes, the prettier issues were introduced by me on another branch and fixed with updating with main. sorry 🙈 

@lachlancollins we get build errors and sherif errors on this PR now, I'm not sure why they show up right now 🤔 . Can you have a look please?

lachlancollins:
@TkDodo I've just cloned the project and run `test:ci` without issues - where did you see the build and sherif issues?

incepter:
> @TkDodo I've just cloned the project and run `test:ci` without issues - where did you see the build and sherif issues?

These errors are actually the result of `pnpm test` on my local (windows 11) machine.

Check this link for the nx cloud report: https://cloud.nx.app/runs/IiLFqaJCYL

lachlancollins:
> These errors are actually the result of `pnpm test` on my local (windows 11) machine.
> 
> Check this link for the nx cloud report: [cloud.nx.app/runs/IiLFqaJCYL](https://cloud.nx.app/runs/IiLFqaJCYL)

Hmm I don't have a good answer - sometimes I get these errors and need to delete my `node_modules` folder and re-install dependencies. It's reassuring though that on the most recent github PR workflow (https://cloud.nx.app/runs/trXy6FQR5s) the react-query build passed (even though it failed on your local machine). And you can also see that the [CI workflows](https://github.com/TanStack/query/actions/workflows/ci.yml), which run every test, pass without issue. I don't think this PR has affected build or sherif.

jinjor:
Thank you all for the quick action! 👍

PR #6355, opened 2023-11-11
fix(query-core): fix context type error in `onSuccess`

Mini-ghost:
## Description:

When defining types in our code, it seems that the type on `onSuccess` for `mutate` is different from the one on `useMutation`:

```ts
interface Context {
  // define
}

const { mutate } = useMutation<Data, Variables, Context>({
  mutationFn: () => fetch('url') // POST API
})

mutate(data, {
  onSuccess(data, variables, context) {
    // context is `undefined` but type is `Context`
  }
})
```

Perhaps there should be consistent type definitions here:

https://github.com/TanStack/query/blob/3f82dc858817516fea2869cc2b72730954fd3003/packages/query-core/src/types.ts#L699

I've also made adjustments to the documentation.

---


TkDodo:
hm, I think the types are the way they are on purpose. `Context` is usually inferred from what `onMutate` returns. The only reason why it's `TContext | undefined` in `onError` and `onSettled` is that the `onMutate` function itself could throw, in which case there is no context. But since `onSuccess` is only called if the request (and thus also `onMutate`)  was successful, there is no way that context is undefined when `onSuccess` is reached.

That is, of course, if you have `onMutate` implemented. If you don't, then `TContext` itself is undefined, and it doesn't matter if we join it with undefined or not.

so, by setting generics manually with:

```
useMutation<Data, Variables, Context>
```

you're basically setting `Context` to a specific type, even though you don't have `onMutate` implemented, which seems wrong.

Mini-ghost:
My apologies for missing some details. If you set it up this way on `useMutation`, the result might be different:

```ts
const { mutate } = useMutation<Data, Variables, Context>({
  mutationFn: () => fetch('url'), // POST API
   onSuccess(data, variables, context) {
     // Here, the `context` type is `Context | undefined`
   }
})

mutate(data, {
  onSuccess(data, variables, context) {
    // 'context' is `undefined`, but the type is `Context`
  }
}
```

Shouldn't both `onSuccess` instances have the same type hint for `context`?

TkDodo:
you're right, but I think `onSuccess` of `useMutation` should have `context` typed as `Context` instead of `Context | undefined` for the reasons I outlined above.

Mini-ghost:
Got it! I've learned a lot from your explanation. I'll adjust the type of `onSuccess` for `useMutation` to be `Context`.

Thanks a bunch for your patient clarification!

TkDodo (approved):

PR #6356, opened 2023-11-11
fix(query-core): rename Removable to GarbageCollectable

manudeli:
I thought if Class's method name is containing `gc`(garbageCollect), class name should be `GarbageCollectable`. like [`Subscribable`](https://github.com/TanStack/query/blob/main/packages/query-core/src/subscribable.ts) have `subscribe` method. so I renamed `Removable` to `GarbageCollectable`

### DIFF
--- a/packages/query-core/src/garbageCollectable.ts
+++ b/packages/query-core/src/garbageCollectable.ts
@@ -1,6 +1,6 @@
 import { isServer, isValidTimeout } from './utils'
 
-export abstract class Removable {
+export abstract class GarbageCollectable {
   gcTime!: number
   #gcTimeout?: ReturnType<typeof setTimeout>
 
@@ -13,7 +13,7 @@ export abstract class Removable {
 
     if (isValidTimeout(this.gcTime)) {
       this.#gcTimeout = setTimeout(() => {
-        this.optionalRemove()
+        this.optionalGc()
       }, this.gcTime)
     }
   }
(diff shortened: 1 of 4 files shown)


TkDodo (commented):
this is debatable from a naming perspective. At the end of the day, we call the `remove` method. If it can't be removed, we schedule a gc. Yes, the naming is not 100% streamlined, but I think it's fine like it is nevertheless.

manudeli:
wonder if I don't understand the difference between `remove` and `gc`, but I have the courage to suggest this. In my opinion, `remove` and `gc` seem to play the same role, and from the library user's point of view, it seems quite good if they are unified.

what if `QueryClient.removeQueries` were unified into `QueryClient.gcQueries`? Or, I think it might be more intuitive for `gcTime` to become `removeTime` with `Removable`.  I know that this cannot be implemented right away because it is a BREAKING CHANGE, but How about including unifying `remove` or `gc` in the work plan for the next major version?

https://tanstack.com/query/v5/docs/react/reference/QueryClient#queryclientremovequeries

My preference is to use the name `gc` than `remove`. Because this name is pretty cool haha.
and I like the name `gc` because it has more of a meaning of clearing the cache.

TkDodo:
> what if QueryClient.removeQueries were unified into QueryClient.gcQueries

that's not what's it is doing though. you can remove a query even though its currently being used, while something can only be garbage collected if its unused.

PR #6359, opened 2023-11-11
chore(query-core): resolve prettier error

manudeli:
Every pull request meet this prettier error. so I resolve it

image


TkDodo:
thanks, and sorry, that was my mistake

PR #6354, opened 2023-11-11
chore: Fix monorepo issues reported by Sherif

lachlancollins:
[Sherif](https://github.com/QuiiBz/sherif) is a tool designed to detect issues with workspace config. This PR fixes the majority of issues (can't fix the vue version or typescript version warnings).

PR #6376, opened 2023-11-15
chore: Add Sherif to CI to lint monorepo config

lachlancollins:
Following on from #6354

We cannot currently upgrade to TS 5.2+ since typescript-eslint v5 is incompatible (see [this StackOverflow post](https://stackoverflow.com/questions/76996326/parsing-error-deprecationerror-originalkeywordkind-has-been-deprecated-since)). I've added pnpm overrides to remove the extremely old eslint/typescript-eslint versions added by CRA v4.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,9 +12,10 @@
     "preinstall": "node -e \"if(process.env.CI == 'true') {console.log('Skipping preinstall...'); process.exit(1)}\" || npx -y only-allow pnpm",
     "install:csb": "corepack enable && pnpm install --frozen-lockfile",
     "test": "pnpm run test:ci",
-    "test:ci": "nx run-many --exclude=examples/** --targets=test:format,test:eslint,test:lib,test:types,test:build,test:bundle,build",
+    "test:ci": "nx run-many --exclude=examples/** --targets=test:format,test:sherif,test:eslint,test:lib,test:types,test:build,test:bundle,build",
     "test:eslint": "nx affected --target=test:eslint",
     "test:format": "pnpm run prettier --check",
+    "test:sherif": "sherif -i react-scripts",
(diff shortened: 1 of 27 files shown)


package.json
```diff
@@ -81,14 +83,22 @@
...
     "rimraf": "^5.0.5",
     "semver": "^7.5.4",
+    "sherif": "^0.5.0",
     "solid-js": "^1.8.1",
     "stream-to-array": "^2.3.0",
     "ts-node": "^10.7.0",
     "tsup": "^7.2.0",
     "type-fest": "^4.5.0",
-    "typescript": "^5.0.4",
+    "typescript": "5.1.6",
```
TkDodo, line 92: @lachlancollins we should probably revive the good parts of this PR:

- https://github.com/TanStack/query/pull/5640/files#diff-15c806aa509538190832852f439e9921a23bec2da81f95ed0e4bf13c14e5b160R101-R105

to test against multiple typescript versions in CI. We want to support TS 4.7, and if we code with TS 5+, we might use a feature that doesn't exist for lower versions.

lachlancollins: Interesting, do you know why this PR wasn't completed?

`main` was already using a combination of TS 5.1.6 (eslint) and 5.2.2 (everything else) so this was more of a consistency fix rather than a TS version bump

TkDodo: > Interesting, do you know why this PR wasn't completed?

no, we did this on a live stream but then zack closed it. I don't remember if there were any issues.

> main was already using a combination of TS 5.1.6 (eslint) and 5.2.2 (everything else) so this was more of a consistency fix rather than a TS version bump

yes, I'm aware. It would still be nice I think  :)

PR #6388, opened 2023-11-16
`isLoading` codemod

balazsmatepetro:


packages/codemods/src/v5/is-loading/is-loading.js
```diff
@@ -0,0 +1,201 @@
...
+  const root = jscodeshift(file.source)
+  const utils = createUtilsObject({ root, jscodeshift })
+  const filePath = file.path
+
+  const dependencies = { jscodeshift, utils, root, filePath }
+
+  transformUsages({
+    ...dependencies,
+    config: {
+      hooks: ['useQuery'],
```
balazsmatepetro, line 195: @TkDodo what other hooks are affected?

packages/codemods/src/v5/is-loading/__testfixtures__/default-import.input.tsx
```diff
@@ -0,0 +1,81 @@
+import * as React from 'react'
```
balazsmatepetro, line 1: @TkDodo does it cover most of the use cases? Feel free to provide a few if you think it lacks some of possible use cases. :)

TkDodo: 👍

packages/codemods/src/v5/is-loading/is-loading.js
TkDodo: `useMutation` is also affected

balazsmatepetro: https://github.com/TanStack/query/pull/6388/commits/66035671e34e06c064e2acfdcc3fb4136988f933

PR #6413, opened 2023-11-21
Update useMutation.md

mwmcode:
cacheTime => gcTime


TkDodo:
it's `cacheTime` in v4, so this is correct.

mwmcode:
My bad, sorry.

PR #6428, opened 2023-11-24
docs(ssr): Fix `prefetchQuery()` function calls

shahzeb1:
This is the correct signature for [prefetchQuery](https://tanstack.com/query/v4/docs/react/reference/QueryClient#queryclientprefetchquery): 

```js
await queryClient.prefetchQuery({ queryKey, queryFn })
```

I noticed the docs were using a slightly stale signature.

Thank you for building an amazing library. :)


TkDodo:
thanks 🙏

PR #6431, opened 2023-11-24
fix(core): make sure queries initialize correctly

### LINKED ISSUE #6369: combine returns a new instance each time when useQueries is called without queries
### Describe the bug

When `useQueries` is used with an empty array `[]` for `queries`, the `combine` function will return a new instance each time the component carrying the hook is rendered. This can cause loop-rendering issues. It is possible to end up with an empty array for `queries` when the queries are determined dynamically. For instance:

```tsx
export const useGetDevices = ({ ids }: { ids: string[] }) => {
 return useQueries({
   // queries are dynamically determined from the given ids
   // when ids is empty, queries will be []
   queries: ids.map((id) => {
     return {
       queryKey: ["get-device", id],
       queryFn: async () => getDevice(id)
     };
   }),
   combine: (results) => {
      // here we simply reduce the results into a Record
      // when ids is not empty the reduced record will be the same instance each time
      // but when ids is empty the reduced record will be a new instance of an empty object each time
      return results.reduce<Record<string, Device[]>>((acc, result) => {
        const { data } = result;
        if (!data) return acc;
        return {
          ...acc,
          [data.id]: data.content,
        };
      }, {});
    }
 });
};
```

More details in [this discussion](https://github.com/TanStack/query/discussions/6337#discussioncomment-7555277)

https://codesandbox.io/s/react-query-combine-empty-array-7n29nr?file=/src/useGetDevices.ts:523-773

### Steps to reproduce

1. In app.tsx: set `ids` to a non-empty array (for instance `["1"]`)
2. Click on "render" with the console opened 
3. See that the useEffect `devicesRecord changed` is NOT called when the component is rendered
4. Edit app.tsx to set `ids` to an empty array `[]`
5. Click on "render" again
6. See that now the useEffect is called each time the component is rendered

### Expected behavior

The useEffect should not be triggered each time the component is rendered

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- browser: chrome, FF

### Tanstack Query adapter

react-query

### TanStack Query version

5.8.3

### TypeScript version

_No response_

### Additional context

_No response_

TkDodo:
setQueries sets them, so it needs to be set before setResult, where we already access `this.queries`

closes #6369

### DIFF
--- a/packages/query-core/src/queriesObserver.ts
+++ b/packages/query-core/src/queriesObserver.ts
@@ -33,7 +33,7 @@ export class QueriesObserver<
 > extends Subscribable<QueriesObserverListener> {
   #client: QueryClient
   #result!: Array<QueryObserverResult>
-  #queries: Array<QueryObserverOptions>
+  #queries!: Array<QueryObserverOptions>
   #observers: Array<QueryObserver>
   #options?: QueriesObserverOptions<TCombinedResult>
   #combinedResult!: TCombinedResult
@@ -46,11 +46,10 @@ export class QueriesObserver<
     super()
 
     this.#client = client
-    this.#queries = []
     this.#observers = []
 
-    this.#setResult([])
     this.setQueries(queries, options)
+    this.#setResult([])
   }
 
   #setResult(value: Array<QueryObserverResult>) {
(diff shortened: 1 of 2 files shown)


RomainLanz:
I am not sure it is related to this commit, but with the latest version (just updated from `5.8.4`), the `combine` method is now called first with an empty array instead of `results`, and those are populated after the first run.

CleanShot 2023-12-06 at 12 39 29

TkDodo:
looking at the code, the constructor has always called:

```
this.#setResult([])
```

and `setResult` calls `combine`, so this seems to have always been the case.

PR #6448, opened 2023-11-28
Update ssr.md

ShurinokOleksandr:
Add type of children


TkDodo:
the examples are in js (`layout.jsx`) and I think most users will know how to type children in react if they use TS

Issue #6441, opened 2023-11-27
RFC: change `experimental_createPersister`'s `serialize`/`deserialize` types


frederikhors:
### Describe the bug

The great @TheTimeWalker [proposed here](https://github.com/TanStack/query/discussions/6213#discussioncomment-7532381) to avoid using serialization/deserialization using IndexedDB:

> One suggestion: The advantage of IndexedDB is that you can safely put objects into the storage without having to rely on serializing/deserializing the content like e.g. with LocalStorage. I have noticed that right now it would by default still stringify the content and then save it to IndexedDB. My suggestion would be to additionally add this into your persister:

```ts
experimental_createPersister({
  // ...
  // Just pass the data
  serialize: (persistedQuery) => persistedQuery,
  deserialize: (cached) => cached,
})
```

> The only problem is that TypeScript will complain that this serialize/deserialize function is not assignable, because it expects a string as a return type and cached type.

> Specifically this:

```
Type '(persistedQuery: PersistedQuery) => PersistedQuery' is not assignable to type '(persistedQuery: PersistedQuery) => string'.
  Type 'PersistedQuery' is not assignable to type 'string'.ts(2322)

Type '(cached: string) => string' is not assignable to type '(cachedString: string) => PersistedQuery'.
  Type 'string' is not assignable to type 'PersistedQuery'.ts(2322)
```

[...]

PR #6450, opened 2023-11-28
feat(experimental_createPersister): support async serialization and storage of arbitrary values instead of strings

DamianOsipiuk:
Fixes
- https://github.com/TanStack/query/discussions/6446
- https://github.com/TanStack/query/issues/6441


TkDodo (approved):
nitpick: can we change the type parameter to be `TStorageValue` to be in line with everything else? I might add a lint rule for this later, too :)

Jack-Works (approved):
also, change the serialize type in `CreateAsyncStoragePersisterOptions`.

Issue #6442, opened 2023-11-27
`queryOptions()` helper using `select` doesn't type-check when used with `useQueries`


larrifax:
### Describe the bug

I'm trying to use a `QueryOptions` object created with the new `queryOptions()` helper in a `useQueries()` call.  
This doesn't typecheck as long as the options object contains a `select`-fn where the return type doesn't equal the return type of the `queryFn`.

It works fine when the same options object is used in a `useQuery()`. Same story with `useSuspenseQueries()` and `useSuspenseQuery()`

### Your minimal, reproducible example

https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgRwK4FMoE8DyYbAQB2AzgDRyonoCKG2FVt9w6JcAvnAGZQQhwARAAEYAQ1LiAxgGsA9FHRipMALRpMWQQChtU4iXgbsAaXRY4AXjgAKAJRWAfHADagmecEBdPQaP0sADEiK1sHS2cABT4QYGoAOkUSCAAbADd0GwBGO19JFACAZXQU9BVoUJsAEzFxAC44IlQQACNMcOca8ThnLN19fIh8QlJQ41xhgxsEbTgCzTMsBvHF+zJZ+exg5YDg9bnqUpUdzWKjmGh1jjsAbn6-Taws0KY6TRshggNb+-zxgCYXtQ3lBWCRppswQ0XJ8RiQvJwfkA

### Steps to reproduce

1. Hover over the error reported in the `useQueries` call in `query2`
2. Observe that it is expecting `select` to return the same type as `queryFn` 

### Expected behavior

I would expect that the `useQueries` call would type-check the same way as the `useQuery` call

### How often does this bug happen?

None

### Platform

Any platform

### Tanstack Query adapter

react-query

### TanStack Query version

5.8.7

### TypeScript version

5.3.2

charlotte-bone:
@TkDodo I think that I have a fix for this. It looks as though it's because select should be optional. I think it affects a few other places too. Could you please assign to me and I'll get the PR ready.

charlotte-bone:
@TkDodo I've managed to get this working - although I couldn't find type tests for most of the methods changed. Let me know if I've done anything wrong as it's my first time opening a PR here.

TkDodo:
@charlotte-bone thank you for this ❤️ 

> although I couldn't find type tests for most of the methods changed

[...]

PR #6471, opened 2023-12-01
fix(tests): add type tests for useQueries

charlotte-bone:
Creating new useQueries type tests in react-query and vue-query and fixing a couple of type bugs.

discussed in https://github.com/TanStack/query/issues/6442


TkDodo:
yeah we never got to the point where initialData removed undefined for userQueries as well ❤️

TkDodo (approved):

PR #6481, opened 2023-12-03
Check if dates are equal in structural sharing

### LINKED ISSUE #6482: Structural sharing doesn't check if dates are equal
### Describe the bug

Noticed in the past couple months that structural sharing in useQuery doesn't check if two date objects are equal to each other.

For my own project, I've had to create a custom `structuralSharing` function that does this check, since the majority of my query results include date objects (NOTE: not date strings).

For context, I use `superjson` with trpc to transform my API responses, which includes transforming date strings to Date objects. My assumption is that not _that_ many people use trpc with superjson, but I also think this is a relatively simple thing to support, a problem that can occur outside of trpc and supersjson, and seems sensible to support

[Code Sandbox](https://codesandbox.io/p/sandbox/pensive-feistel-2tcy6l?layout=%257B%2522sidebarPanel%2522%253A%2522EXPLORER%2522%252C%2522rootPanelGroup%2522%253A%257B%2522direction%2522%253A%2522horizontal%2522%252C%2522contentType%2522%253A%2522UNKNOWN%2522%252C%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522id%2522%253A%2522ROOT_LAYOUT%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522UNKNOWN%2522%252C%2522direction%2522%253A%2522vertical%2522%252C%2522id%2522%253A%2522clppq4itd00063b6fyrqtq8xe%2522%252C%2522sizes%2522%253A%255B100%252C0%255D%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522EDITOR%2522%252C%2522direction%2522%253A%2522horizontal%2522%252C%2522id%2522%253A%2522EDITOR%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522EDITOR%2522%252C%2522id%2522%253A%2522clppq4itd00023b6fhis6340b%2522%257D%255D%257D%252C%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522SHELLS%2522%252C%2522direction%2522%253A%2522horizontal%2522%252C%2522id%2522%253A%2522SHELLS%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522SHELLS%2522%252C%2522id%2522%253A%2522clppq4itd00033b6fm1cc1ft9%2522%257D%255D%252C%2522sizes%2522%253A%255B100%255D%257D%255D%257D%252C%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522DEVTOOLS%2522%252C%2522direction%2522%253A%2522vertical%2522%252C%2522id%2522%253A%2522DEVTOOLS%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522DEVTOOLS%2522%252C%2522id%2522%253A%2522clppq4itd00053b6f22abs8pq%2522%257D%255D%252C%2522sizes%2522%253A%255B100%255D%257D%255D%252C%2522sizes%2522%253A%255B50%252C50%255D%257D%252C%2522tabbedPanels%2522%253A%257B%2522clppq4itd00023b6fhis6340b%2522%253A%257B%2522tabs%2522%253A%255B%257B%2522id%2522%253A%2522clppq4itd00013b6fvxfugm1d%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522FILE%2522%252C%2522filepath%2522%253A%2522%252Fsrc%252Findex.js%2522%257D%255D%252C%2522id%2522%253A%2522clppq4itd00023b6fhis6340b%2522%252C%2522activeTabId%2522%253A%2522clppq4itd00013b6fvxfugm1d%2522%257D%252C%2522clppq4itd00053b6f22abs8pq%2522%253A%257B%2522tabs%2522%253A%255B%257B%2522id%2522%253A%2522clppq4itd00043b6f98rieq8j%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522UNASSIGNED_PORT%2522%252C%2522port%2522%253A0%252C%2522path%2522%253A%2522%2522%257D%255D%252C%2522id%2522%253A%2522clppq4itd00053b6f22abs8pq%2522%252C%2522activeTabId%2522%253A%2522clppq4itd00043b6f98rieq8j%2522%257D%252C%2522clppq4itd00033b6fm1cc1ft9%2522%253A%257B%2522tabs%2522%253A%255B%255D%252C%2522id%2522%253A%2522clppq4itd00033b6fm1cc1ft9%2522%257D%257D%252C%2522showDevtools%2522%253Atrue%252C%2522showShells%2522%253Afalse%252C%2522showSidebar%2522%253Atrue%252C%2522sidebarPanelSize%2522%253A15%257D)

### Steps to reproduce

You can reproduce this with this component that uses a useQuery hook that returns the same date object on every fetch. 

What you will see with this component is that every second the render count will increment::

```js
function DateComponent() {
  const renderCount = useRef(0);

  const { data, isLoading, error } = useQuery({
    queryKey: ["fooLastModified"],
    queryFn: async () => {
      return new Date("2023-12-03T16:58Z");
    },
    refetchInterval: 1000,
  });

  if (isLoading) return <div>Loading...</div>;
  if (error) return <div>Error: {error.message}</div>;

  renderCount.current += 1;

  return (
    <div>
      <div>Foo was last modified at: {data.toString()}</div>
      <em>Render count: {renderCount.current}</em>
    </div>
  );
}
```

https://github.com/TanStack/query/assets/17055383/22cc38b0-c31d-4390-a9e5-58a15c3b36e7

For comparison, when the useQuery hook returns a date string, the render count stays static

```js
function DateComponent() {
  // Use a renderCount ref to keep track of the number of renders
  const renderCount = useRef(0);

  const { data, isLoading, error } = useQuery({
    queryKey: ["fooLastModified"],
    queryFn: async () => {
      return "2023-12-03T16:58Z";
    },
    refetchInterval: 1000,
  });

  if (isLoading) return <div>Loading...</div>;
  if (error) return <div>Error: {error.message}</div>;

  renderCount.current += 1;

  return (
    <div>
      <div>Foo was last modified at: {data.toString()}</div>
      <em>Render count: {renderCount.current}</em>
    </div>
  );
}
```

https://github.com/TanStack/query/assets/17055383/ec85a4d3-35ad-4b7f-9917-f1d9e7228296

### Expected behavior

As a user I expected date objects with the same value to be treated as equal

Every time

### Screenshots or Videos

https://github.com/TanStack/query/assets/17055383/41c6c99c-2eef-4883-96e8-9bc5b0825662

### Platform

- OS: Any OS
- Browser: Any browser
- Version: Any browser version

### Tanstack Query adapter

None

### TanStack Query version

v4.26.1

### TypeScript version

_No response_

### Additional context

_No response_

juanrgon:
Closes: https://github.com/TanStack/query/issues/6482

### DIFF
--- a/packages/query-core/src/utils.ts
+++ b/packages/query-core/src/utils.ts
@@ -221,6 +221,10 @@ export function replaceEqualDeep(a: any, b: any): any {
     return a
   }
 
+  if (a instanceof Date && b instanceof Date) {
+    return a.getTime() === b.getTime() ? a : b;
+  }
+
   const array = isPlainArray(a) && isPlainArray(b)
 
   if (array || (isPlainObject(a) && isPlainObject(b))) {


TkDodo:
We only support json serializable values on purpose. If we start to support date, we also need Set / Map and everything else. You can always pass your own function to structuralSharing

juanrgon:
> We only support json serializable values on purpose. If we start to support date, we also need Set / Map and everything else.

@TkDodo I understand this rationale, but:

- I assume that Set / Map are less common than Date objects. I have yet to use a Set / Map in any of my API responses for my app (https://meetmotif.com)
- If this is blocker, I can update this function to support Set/Map

>  You can always pass your own function to structuralSharing

I'm currently doing this in my project. However, it took me over 6 months to understand this was happening. I had the false assumption that react-query by design **always** triggered a render when any data was loaded, because virtually all of my queries include a date object (e.g. `createdAt` or `modifiedAt`)

I don't want to push too hard because there is a work-around, but I'm assuming that I'm not the only one who has this issue (regardless of if others know this or not)

Also, thank you for the quick reply! 🙇🏽

TkDodo:
> are less common than Date objects

Not really. The main use-case for TanStack Query is still to manage async state that comes from data fetching, where only JSON is supported. You're using superjson with trpc, and that's a different story. If we would need to support everything that superjson supports, we would need to probably add superjson itself, which means another 3kb of bundle size, which is not acceptable for a somewhat niche use-case.

My suggestion here is that this is something that trpc would need to handle if you don't want to handle this in user land. They create the QueryClient for you, so if they know that you use the superjson transform, they should also provide a better `structuralSharing` function.

For "vanilla" react query, this is definitely out of scope, sorry.

tran-simon:
@TkDodo 

> You can always pass your own function to structuralSharing

Is there a better way to customize the `replaceEqualDeep` function? Currently I had to copy the whole function as well as the `isPlainArray`, `isPlainObject and hasObjectPrototype

tran-simon:
@TkDodo 

> You can always pass your own function to structuralSharing

Is there a better way to customize the `replaceEqualDeep` function? Currently I had to copy the whole function as well as the `isPlainArray`, `isPlainObject` and `hasObjectPrototype` (these are not exported by `@tanstack/query-core`) just to add the 3 lines to compare objects. It would be nice to have a factory function or something similar to create a customized `replaceEqualDeep`

TkDodo:
> Is there a better way to customize the replaceEqualDeep function?

use [`superjson`](https://github.com/blitz-js/superjson)

TkDodo:
then just copy paste our code into your own, it's fine. Or don't send dates from the backend but serialize it to a ISO-8601 string and parse it again on the frontend.

> when all we want is to add a few line to the built-in replaceEqualDeep

Really, we're not going to go the route of starting to serialize non json data. I hope I made that clear. You're free to disagree with that of course, and you have multiple options to work around this limitation.

gaearon:
This may be very tangential but I wanted to note that RSC has built-in support for [more things than JSON](https://github.com/facebook/react/issues/25687) — partially inspired by what the [`structuredClone` standard](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) supports. In particular, Dates, Maps, and Sets are supported.

Although RQ by itself is unrelated to RSC, for people using the two together, it won't be uncommon to plumb some dates down from the server — given that they _can_ serialize well and are unambiguous (and RSC supports it). So it's a bit unfortunate that RQ becomes a limiting factor here breaking a default optimization. It's easy to miss why it broke without very targeted debugging. Will become more of an issue as passing dates down becomes more common.

Another case in which we'd bump into that is custom SDKs that might include dates in their hydrated response objects. It's fair to say that these might be too custom but that's how we accidentally regressed on this (https://github.com/bluesky-social/social-app/pull/3785 is the fix). A date crept in, the code kept working, the optimization broke.

So I wanted to voice support for raising the bar from JSON-serializable to _some_ superset. Of course you'd have to choose _which_ superset, and there could be an endless debate where nobody goes away happy. But you could consider a pragmatic subset of [Structured Clone](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm), maybe a narrower one than React did. Curiously, RQ is already technically doing that because its comparison algorithm already handles `undefined`, `null`, `Infinity`, `NaN`, `-0`, and `BigInt`s — none of which are JSON-serializable (but are in Structured Clone).

I think that makes the argument for adding `Date`s as well (without a bunch of other logic) a bit stronger. IMO it would be a significant improvement, would not compromise the code size or complexity, and maybe doesn't have to be a slippery slope.

Thanks for considering!

gaearon:
Proposal.

We could start with [Structured Clone JavaScript Types](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#javascript_types) and narrow it down.

Let's exclude a few for a similar reason React excluded them pragmatically:

- Wrapper objects like `Boolean`, `Number`, `String` — too exotic, no real use case.
- `RegExp` — possible security implications for passing down.

Then let's exclude binary and low-level stuff:

- `ArrayBuffer`
- `DataView`
- `TypedArray`

We're left with `Array` (already supported), `Date` (needs support), `Map` and `Set` (need support), and `NaN` (needs support). So it's not a lot of code and the list is very limited and pragmatically bounded.

Screenshot 2024-05-01 at 05 23 23

TkDodo:
Thanks for taking the time to post these details. We currently have multiple places where we say that we need json serializable data:

- for `structuralSharing` to work (as you've noticed)
- for hashing the `queryKey`
- for our `persistence` API, that is used for writing data to / reading data from storages (like localstorage)

All of these places can be customized:

- for `structuralSharing`, you can pass your own function to it.
- for hashing the `queryKey`, you can set a `queryKeyHashFn`
- for persistence, you can pass `serialize` and `deserialize` to `createPersister` or `createSyncStoragePersister` / `createASyncStoragePersister`.

If we draw the line at `Date`, `Map`, `Set` and `NaN` it feels a bit arbitrary (though the `structuredClone` reference makes sense), and I fully expect people to ask us about "but why no Regex then?". It's much easier to argue along the lines of "only JSON is allowed, everything else, you're on your own".

I usually recommend [superjson](https://github.com/blitz-js/superjson) because they provide a thin wrapper around `stringify` / `parse` that can cover most if not all of the cases in user-land.

I can see that this is difficult for `structuralSharing` per se, because you'd need to swap out everything and do other field comparisons, too. Maybe there's something we can do better here, I'll think about it.

zeorin:
For reference, I've suggested adding a more comprehensive version of `replaceEqualDeep` to Remeda: https://github.com/remeda/remeda/issues/1369. I would welcome any input on its design.

PR #6516, opened 2023-12-10
refactor(query-core): Updates the Return Type of the isFetching and isMutating methods from Number to Boolean

Lennon57:
## Changes Made

I proceeded with updating the return type to boolean, in line with the is-prefix for improved readability.

### Changes Overview

- **query-core**
  - Modified internal logic in `queryClient.ts` to handle return values of `isFetching` and `isMutating` as boolean.
- **react-query** 
  - Change the return type of `useIsFetching.ts` to `boolean` from `number`.
- **angular-query-experimental**
  - Change the return type of `injectIsFetching.ts` and `injectIsMutating.ts` to `boolean` from `number`.
- **solid-query and svelte-query** 
  - Change the return type of `useIsFetching.ts` and `useIsMutating.ts` to `boolean` from `number`.
- **vue-query** 
  - Change the return type of `isFetching` and `isMutating` from number to boolean within the `queryClient.ts` file.
  - Change the return type of `useIsFetching.ts` to `boolean` from `number`.

### Test

- ‼️ There may be a need for tests related to the number of queries, so if you guys agree with my opinion, I'll modify the test code accordingly and commit it together.

### DIFF
--- a/packages/angular-query-experimental/src/injectIsFetching.ts
+++ b/packages/angular-query-experimental/src/injectIsFetching.ts
@@ -7,7 +7,7 @@ import type { Injector, Signal } from '@angular/core'
 export function injectIsFetching(
   filters?: QueryFilters,
   injector?: Injector,
-): Signal<number> {
+): Signal<boolean> {
   return assertInjector(injectIsFetching, injector, () => {
     const queryClient = injectQueryClient()
     const destroyRef = inject(DestroyRef)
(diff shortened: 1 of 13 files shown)


TkDodo:
number is correct - this would be a breaking change...

Lennon57:
> number is correct - this would be a breaking change...

Thank you for the information! I'll try to contribute to another area.

PR #6518, opened 2023-12-10
feat: tsr autostreaming

juliusmarminge:


packages/react-query-next-experimental/package.json
juliusmarminge, line 1: might warrant a different name now, alternatively split out the core to a new lib and keep this one that exports a "wrapped" one that passes `useServerInsertedHtml` internally for backwards compat, but it's an experimental lib so I think this small breaking change is fine?

Issue #6427, opened 2023-11-24
when gcTime is zero, it make hydration fail warning


MOONLEE:
### Describe the bug

When gcTime is zero in app router, it make hydration failed.  
If you use page router, it is not happen.  

``` 
     const [queryClient] = React.useState(() => {
              return new QueryClient({
                  defaultOptions: {
                      queries: {
                          staleTime: 5 * 1000,
                          gcTime: 0, // it make hydration failed.  
                      }
                  }
              })
          }
      );
  
      return (
          <RecoilRoot>
              <QueryClientProvider client={queryClient}>
                  {children}
                  <ReactQueryDevtools initialIsOpen={false} />
              </QueryClientProvider>
          </RecoilRoot>
      )
```

### Your minimal, reproducible example

https://github.com/MOONLEE/react-query-hydration-fail.git

### Steps to reproduce

1. download git repo (https://github.com/MOONLEE/react-query-hydration-fail.git)
2. yarn run install 
3. yarn run dev
4. refresh localhost:3000 in web browser 

### Expected behavior

![img](https://github.com/TanStack/query/assets/10557561/f6a6d0ac-2547-411a-ac15-35c2f0481f93)

### How often does this bug happen?

Every time

### Platform

- OS : Mac
- Browser :  chrome (19.0.6045.159)
- @tanstack/react-query : 5.7.2
- nextjs: 4.0.1
- node: 18.18.2
 

### Tanstack Query adapter

None

### TanStack Query version

5.7.2

### TypeScript version

5.2.2

TkDodo:
I opened the reproduction in codesandbox and it works fine:

https://codesandbox.io/p/github/MOONLEE/react-query-hydration-fail/main

MOONLEE:
I tested it using MacBook Pro at home, it almost work fine. 
but sometime, hydration failed occurred.

* work fine in codesandbox
![success_sandbox_contents](https://github.com/TanStack/query/assets/10557561/a62f4711-6940-42ee-8e8d-19cd5324d939)
![success_sandbox_log](https://github.com/TanStack/query/assets/10557561/697694a9-b638-4135-95d2-20e852486fdc)



* hydration failed occurred  in codesandbox
![failed_sandbox_contents](https://github.com/TanStack/query/assets/10557561/10ea6116-5974-47f2-8e66-de43ea2c9a1b)
[...]

PR #6544, opened 2023-12-16
docs: add documentation for gcTime error for react-query

brendenehlers:
Currently, the documentation doesn't warn developers about an error caused when `gcTime` is `0`: 
https://github.com/TanStack/query/blob/60cbd75e4010a47ceba63ef8b7b6a69f384a9817/docs/react/guides/important-defaults.md?plain=1#L23

This change updates the docs to include this sentence: **Avoid setting `gcTime` to `0` as that can cause a hydration error.**

Issue reference: #6427

### DIFF
--- a/docs/react/guides/important-defaults.md
+++ b/docs/react/guides/important-defaults.md
@@ -20,7 +20,7 @@ Out of the box, TanStack Query is configured with **aggressive but sane** defaul
 - Query results that have no more active instances of `useQuery`, `useInfiniteQuery` or query observers are labeled as "inactive" and remain in the cache in case they are used again at a later time.
 - By default, "inactive" queries are garbage collected after **5 minutes**.
 
-  > To change this, you can alter the default `gcTime` for queries to something other than `1000 * 60 * 5` milliseconds.
+  > To change this, you can alter the default `gcTime` for queries to something other than `1000 * 60 * 5` milliseconds. Avoid setting `gcTime` to `0` as that will cause a hydration error.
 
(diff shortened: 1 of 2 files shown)


docs/react/guides/important-defaults.md
```diff
@@ -20,7 +20,7 @@ Out of the box, TanStack Query is configured with **aggressive but sane** defaul
 - Query results that have no more active instances of `useQuery`, `useInfiniteQuery` or query observers are labeled as "inactive" and remain in the cache in case they are used again at a later time.
 - By default, "inactive" queries are garbage collected after **5 minutes**.
 
-  > To change this, you can alter the default `gcTime` for queries to something other than `1000 * 60 * 5` milliseconds.
+  > To change this, you can alter the default `gcTime` for queries to something other than `1000 * 60 * 5` milliseconds. Avoid setting `gcTime` to `0` as that can cause a hydration error.
```
TkDodo, line 23: I would rather put this as a separate section in `Tips, Tricks and Caveats` in the [SSR docs page](https://tanstack.com/query/latest/docs/react/guides/ssr#tips-tricks-and-caveats), because it's strictly SSR related.

We already have a paragraph there explaining why `gcTime` defaults to `Infinity` [on the server](https://tanstack.com/query/latest/docs/react/guides/ssr#high-memory-consumption-on-server), so adding another one why setting it to zero is not a good idea would be good.

TkDodo (changes requested):

docs/react/guides/important-defaults.md
brendenehlers: Updated 👍

brendenehlers (commented):
updated

PR #6558, opened 2023-12-19
refactor(angular-query): remove dependency to ngxtension

mikelgo:
To maintain a stable lib I would prefer to maintain such utils on my own, just to reduce external dependencies


arnoud-dv:
Agree, was already my intention to not depend on ngxtension later on but I wanted to get to an API as quickly as possible. Can highly recommend it to be used together with Angular Query though.

mikelgo:
> Can you take a look at the unit tests? They are failing. You can run all the tests by running `pnpm test:ci`

I'll take a look probably tomorrow

packages/angular-query-experimental/src/util/create-injection-token/create-injection-token.ts
```diff
@@ -0,0 +1,268 @@
...
+  TOptions = Pick<
+    CreateInjectionTokenOptions<() => void, []>,
+    'extraProviders'
+  > &
+    (TMulti extends true ? { multi: true } : Record<string, never>),
+>(description: string, options?: TOptions) {
+  type TReturn = TMulti extends true ? Array<TValue> : TValue;
+
+  const token =
+    (options as CreateInjectionTokenOptions<() => void, []>).token ||
```
garrettld, line 255: It looks like the test failure is related to this line, `options` is an optional parameter to `createNoopInjectionToken`, so it just needs an optional chaining operator:

```suggestion
    (options as CreateInjectionTokenOptions<() => void, []>)?.token ||
```

mikelgo:
@arnoud-dv @garrettld thank you for taking care of the failing test. I'm still on vacation and could not do it.

PR #6576, opened 2023-12-23
feat(query-devtools): Add btn position relative support

happycollision:
In addition to adding support for relative positioning of the button, I've added in opacity changes for all in/out animations.

This change makes it possible to control the location of the button, in case you have other floating panels, etc., in your app.

The gif below shows a change to the playground example (not committed), that illustrates what is now possible.

CleanShot 2023-12-22 at 22 29 51

My motivation was purely because I have an application with a bunch of devtools in a drawer that I move around the screen as necessary while I develop. The previous version of RQ's devtools allowed me to pass a style directly to the button to control its position so that I could contain it within the drawer. This newer version does not. But allowing the button to just be positioned relatively works just fine for that purpose.

You can **see below (in the lower right of each image)** the best I could do to get v5 to play ball, compared to v4. (And _forget_ moving the panel around!)

|v4 devtools|v5 devtools (without this PR)|
|-|-|
|CleanShot 2023-12-22 at 23 06 00@2x|CleanShot 2023-12-22 at 23 03 43@2x|

### DIFF
--- a/docs/framework/react/devtools.md
+++ b/docs/framework/react/devtools.md
@@ -56,9 +56,10 @@ function App() {
 
 - `initialIsOpen: Boolean`
   - Set this `true` if you want the dev tools to default to being open
-- `buttonPosition?: "top-left" | "top-right" | "bottom-left" | "bottom-right"`
+- `buttonPosition?: "top-left" | "top-right" | "bottom-left" | "bottom-right" | "relative"`
   - Defaults to `bottom-right`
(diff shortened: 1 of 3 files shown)


TkDodo:
looks good. can you please fix the conflicts?

happycollision:
@TkDodo Just letting you know it's done.

ardeora:
Thanks for working on this! Looks good

ardeora (approved):

PR #6592, opened 2023-12-26
fix(query-core): replace scheduleMicrotask with queueMicrotask

NathanHuisman:
scheduleMicrotask was implemented incorrectly with setTimeout, causing the passed callback to run in the next task instead of a microtask. This sometimes resulted in a layout shift in my solid project. Fix this by replacing scheduleMicrotask with queueMicrotask, which is supported by all browsers query supports.[^tanstack][^mdn]

[^tanstack]: https://tanstack.com/query/v5/docs/react/installation#requirements
[^mdn]: https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask#browser_compatibility


TkDodo:
There are different tradeoffs here. In react, this will likely result in more re-renders because fewer things can be batched together. This is likely shown in the failing tests. I think to not make this breaking, we could offer a batching strategy as an option somewhere, like timeout (default, current version), queueMicroTask and maybe even requestAnimationFrame that the user can choose from. I think redux toolkit has such an option: https://redux-toolkit.js.org/api/autoBatchEnhancer#autobatchenhancer-1

NathanHuisman:
Hmm, I did not consider that, requestAnimationFrame could also be interesting. Maybe a setter on the notify manager like this:
```typescript
function setBatchType(config: { type: 'raf' | 'tick' } | { type: 'timer', timeout?: number}) { /* ... */ }
```
But then I'm not really sure how to expose this to the api surface in solid-query, react-query, etc.

NathanHuisman:
I implemented a PoC here: https://github.com/NathanHuisman/query/commit/38b41a31f42ddf6c79136074c95571644e9ce0be

TkDodo:
That looks really good. I think we'd need some docs and tests, too please

NathanHuisman:
I'm closing this pull request in favor of opening  a new one with the new changes I made

PR #6600, opened 2023-12-27
feat(notifyManager): batch method configuration

NathanHuisman:
As discussed previously in #6592, adding a way to configure the batch scheduling behavior can be useful, e.g. to prevent unwanted rerenders in solid-js (although I haven't fully gotten to the bottom of why solid-query does this).

This pull request adds `notifyManager.setBatchMethod(...)`

~Sets the method for batching. There are four methods:~

- ~`{ type: 'timer', timeout?: number }`:  The default is to use setTimeout with a timeout of 0. This has the effect of
  scheduling a javascript task in the event loop some time later.~
- ~`{ type: 'raf' }`: This batches all updates in a requestAnimationFrame, so that it runs before the next frame is rendered by the browser.~
- ~`{ type: 'tick' }`: This uses queueMicrotask to schedule a microtask to batch updates, which runs immediately after the current task is done.~
- ~`{ type: 'custom', fn: (() => void) => void}`: The custom method allows you to specify your own way of scheduling batches~

**UPDATE: The API is now the following:**

```ts
import { notifyManager } from '@tanstack/react-query'
// Schedule batches in the next microtask
notifyManager.setScheduler(queueMicrotask)
// Schedule batches before the next frame is rendered
notifyManager.setScheduler(requestAnimationFrame)
// Schedule batches some time in the future
notifyManager.setScheduler((cb) => setTimeout(cb, 10))
```

This API was inspired by the [redux autoBatchEnhancer API](https://redux-toolkit.js.org/api/autoBatchEnhancer#autobatchenhancer-1)

I added some tests, and documented the notifyManager while I was at it. I would appreciate feedback, because I've not written many docs before and I'm not 100% sure if I documented all the functionality correctly.

I also added a demonstration in the solid-start-streaming example to demonstrate how setTimeout (the default batch method) can sometimes result in layout shifts solidjs. I don't know if this is due to the way how solid works or if solid-query can adapt to prevent this from happening. Either way, changing the batching method to 'raf' or 'tick' also prevents it from happening.


arnoud-dv:
Although I would need to look deeper into this to be sure I suspect this may be beneficial for the Angular adapter too. setTimeout triggers change detection and queueMicrotask may well be more efficient. At the very least it's good for adapters to have options.

docs/react/reference/notifyManager.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+## `notifyManager.setBatchMethod`
+
+Sets the method for batching. There are four methods:
+
+- `{ type: 'timer', timeout?: number }`:
+  The default is to use setTimeout with a timeout of 0. This has the effect of
+  scheduling a javascript task in the event loop some time later.
+- `{ type: 'raf' }`: This batches all updates in a requestAnimationFrame, so
```
TkDodo, line 84:
```suggestion
- `{ type: 'raf' }`: This batches all updates in a [requestAnimationFrame](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame), so
```

pnpm-lock.yaml
```diff
@@ -1055,6 +1055,21 @@ importers:
         specifier: ^2.8.0
         version: 2.8.0(solid-js@1.8.7)(vite@5.0.10)
 
+  examples/solid/bug-test:
```
TkDodo, line 1058: the changes to the lock file look weird given that we have no changes in any `package.json` file. Can we revert them please ?

TkDodo (commented):
great work, thank you 🙏

pnpm-lock.yaml
NathanHuisman: Ah yes, I was doing some experiments, didn't realise I updated the lockfile. Will revert tomorrow

packages/query-core/src/notifyManager.ts
```diff
@@ -19,6 +21,23 @@ export function createNotifyManager() {
...
+    } else if (method.type === "raf") {
+      scheduleFn = requestAnimationFrame
+    } else if (method.type === "timer") {
+      scheduleFn = (cb) => setTimeout(cb, method.timeout ?? 0)
+      // Disable because condition is necessary for typescript to detect method.fn
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    } else if (method.type === "custom") {
+      scheduleFn = method.fn
+    }
+  }
```
TkDodo, line 38: on a second thought - while the api of rtk-query is nice, wouldn't we able to achieve the exact same thing only with the custom type?

- timeout: `setBatchMethod({ type: 'custom', fn: (cb) => setTimeout(cb, 0) })`
- raf: `setBatchMethod({ type: 'custom', fn: requestAnimationFrame })`
- tick: `setBatchMethod({ type: 'custom', fn: queueMicrotask })` 

with that in mind, we could simplify our api and just offer:

```
setBatchMethod: (callback: () => void) => void
```

and show examples in the docs on how to implement it with timeout / raf / tick ?

@markerikson maybe you can help us shed some light on how you decided on that specific api for the autobatchenhancer ?

markerikson: Not sure there was much thinking put into it. My initial implementation was hardcoded to `queueMicrotask`, and then I realized it might be good to make it configurable. I wanted to build in the implementations so users didn't have to write the callback implementations, so I opted for the enum / literal option approach.

packages/query-core/src/notifyManager.ts
```diff
@@ -19,6 +21,23 @@ export function createNotifyManager() {
   let batchNotifyFn: BatchNotifyFunction = (callback: () => void) => {
     callback()
   }
+  let scheduleFn: ScheduleFunction
+
+  const setBatchMethod = (method: BatchMethod) => {
```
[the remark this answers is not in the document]
TkDodo: that's fine by me 👍

packages/query-core/src/notifyManager.ts
TkDodo: thanks. in that case, @NathanHuisman could you maybe refactor towards this syntax?

```
setBatchMethod: (callback: () => void) => void
```

because for raf / tick, it would just be:

```
notifyManager.setBatchMethod(requestAnimationFrame)
notifyManager.setBatchMethod(queueMicrotask)
```

the only syntax that would be a bit more to write would be `setTimeout` with a different value, but that's fine by me given that this gives us a smaller api surface and less code on our end.

TkDodo (approved):
Thanks. I just added a tiny refactoring to initialize `scheduleFn` where we declare it rather than calling `setScheduler`.

https://github.com/TanStack/query/pull/6600/commits/c68a985eca4a76c79e33a01371bbf8258aba72c4

I think this is ready to go 🚀 . Great work 🙌

PR #6611, opened 2023-12-29
fix(react): don't set options before suspending

### LINKED ISSUE #6392: [react-query] useSuspenseQueries does not throw on errors after queryKey change
### Describe the bug

`useSuspenseQuery` does not throw errors occurring in queryFn and returns data as undefined, when queryFn was invoked in reaction to a state change.

See example and repro steps for more information.

https://codesandbox.io/s/react-query-usesuspensequery-does-not-throw-on-querykey-change-lpgyl9

### Steps to reproduce

1. Setup a standard query client, provider, suspense and error boundary
2. Setup a query that loads something based on a prop or state (for example loading a post by id)
3. Have the query succeed on first render
4. Make a state change, for example changing the desired post to a non-existing id (and have your loading logic throw an error when the post is not found)
5. Observe that the return value of useSuspenseQuery has the property `data` set to `undefined` and `error` to the error thrown in `queryFn`

In my minimal example, clicking the "Should fail" button toggles a `shouldFail` flag to true, making `queryFn` throw an error, resulting in the described behavior. Clicking the button a second time makes `useSuspenseQuery` throw and trigger the ErrorBoundary.

### Expected behavior

I would expect `useSuspenseQuery` to throw when there was no data available *for the current queryKey*, not for the previous invocation.

Every time

### Screenshots or Videos

_No response_

### Platform

Windows, behavior observed in both Firefox and Chrome.

### Tanstack Query adapter

react-query

### TanStack Query version

5.8.4

### TypeScript version

_No response_

### Additional context

I've done some digging myself, looking at `defaultThrowOnError` and noticed that `query.queryKey` matches the previous query key, rather than the current and the data that indeed belongs to that previous queryKey. This data is not undefined, and thus no throw happens.

### LINKED ISSUE #6486: useSuspenseQuery made extra request with old value
### Describe the bug

When combine useSuspenseQuery  with  startTransition or useDeferredValue , will trigger extra request with the old value 

I made some test and think the problem is introduced in 5.8.6

https://stackblitz.com/edit/stackblitz-starters-fhjkbj?file=src%2Findex.js

### Steps to reproduce

click the incre button and look at the console 

### Expected behavior

not trigger extra request with the old value 

Every time

### Screenshots or Videos

_No response_

### Platform

all

### Tanstack Query adapter

None

### TanStack Query version

v5.8.6

### TypeScript version

_No response_

### Additional context

_No response_

### LINKED ISSUE #6498: `useSuspenseQueries` and `startTransition` will trigger infinite query requesting
### Describe the bug

If you use `startTransition` to change any value used in a query of `useSuspenseQueries` you will get infinite query requesting.

https://codesandbox.io/p/sandbox/tanstack-query-bug-forked-sz564y?file=%2Fsrc%2FApp.js

Edit: Query cancellation was important for this bug to manifest: https://codesandbox.io/p/sandbox/infinite-query-retrying-example-9k2456?file=%2Fsrc%2FApp.js

### Steps to reproduce

1. Open the code sandbox
2. Open the console
3. Click on the `set seed` button
4. Watch the infinite query requesting

### Expected behavior

I expect not to get into the infinite requesting state.

Every time

### Screenshots or Videos

https://github.com/TanStack/query/assets/71276198/eaa7ccdf-a863-4a28-8bb0-7a680ae78cf8

### Platform

- OS: MacOS
- Browser: Chrome
- Version: 119.0.6045.199

### Tanstack Query adapter

react-query

### TanStack Query version

5.12.2

### TypeScript version

_No response_

### Additional context

Might be related to the following issues: 
- https://github.com/TanStack/query/issues/6486 
- https://github.com/TanStack/query/issues/6491

TkDodo:
setting options right before suspending is a side effect, and it has repercussions if used in transitions where the component stays mounted

to fix #6392 in a different way, we can't rely on `observer.getCurrentQuery()` when reading for error boundaries / suspense, because it can lag behind; instead, we can just read from the queryCache directly to get the latest values

fixes #6486
fixes #6498


TkDodo:
we can see here that this fixes the original issue:

https://codesandbox.io/p/devbox/tanstack-query-example-react-basic-typescript-forked-mqmdyw?file=%2Findex.html%3A15%2C10

PR #6621, opened 2023-12-31
feat(angular-query): add injectMutationState

arnoud-dv:


packages/angular-query-experimental/src/inject-mutation-state.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    effect(
+      () => {
+        // Setting the signal from an effect because it's both 'computed' from options()
+        // and needs to be set imperatively in the mutationCache listener.
+        result.set(getResult(mutationCache, options()))
+      },
+      {
+        allowSignalWrites: true,
+      },
```
eneajaho, line 58:
```suggestion
      () => {
        const options = options();
        
        // Setting the signal from an effect because it's both 'computed' from options()
        // and needs to be set imperatively in the mutationCache listener.
        untracked(() => result.set(getResult(mutationCache, options)));
      },
```

arnoud-dv: Is the suggested change for this reason? https://twitter.com/GergelySzerovay/status/1754915190213144949

eneajaho: Yeah. More here:  https://github.com/angular/angular/issues/53986

adam4nj:
is injectMutationState not going to be implemented?

arnoud-dv:
> is injectMutationState not going to be implemented?

Probably today, but this PR was very old.

arnoud-dv:
@adam4nj injectMutationState is now available in [v5.22 ](https://github.com/TanStack/query/releases/tag/v5.22.0)!

PR #6622, opened 2023-12-31
feat(eslint-plugin): add `no-deprecated-options` rule

iammminzzy:
Resolves https://github.com/TanStack/query/discussions/5279#discussioncomment-5942835.

I took the existing `prefer-query-object-syntax` rule and modified it to look for deprecated callbacks.

### DIFF
--- a/docs/config.json
+++ b/docs/config.json
@@ -304,6 +304,10 @@
               "to": "react/eslint/exhaustive-deps"
             },
             {
+              "label": "No deprecated options",
+              "to": "react/eslint/no-deprecated-options"
+            },
+            {
               "label": "Prefer object syntax",
               "to": "react/eslint/prefer-query-object-syntax"
             },
--- a/docs/react/eslint/eslint-plugin-query.md
+++ b/docs/react/eslint/eslint-plugin-query.md
@@ -33,6 +33,7 @@ Then configure the rules you want to use under the rules section:
 {
   "rules": {
     "@tanstack/query/exhaustive-deps": "error",
+    "@tanstack/query/no-deprecated-options": "error",
     "@tanstack/query/prefer-query-object-syntax": "error",
     "@tanstack/query/stable-query-client": "error"
   }
--- a/docs/react/eslint/no-deprecated-options.md
+++ b/docs/react/eslint/no-deprecated-options.md
@@ -1,0 +1,65 @@
+---
+id: no-deprecated-options
+title: Disallowing deprecated options
+---
+
+This rule warns about deprecated [`useQuery`](https://tanstack.com/query/v4/docs/reference/useQuery) options which will be removed in [TanStack Query v5](https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5).
+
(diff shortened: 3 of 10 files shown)


TkDodo:
Thanks, I think we _can_ ship this, but tbh, I don't really think it's worth it. The overlap of:
- people knowing about the eslint-plugin
- updating the eslint-plugin to the latest version
- still being on v4 of react-query
- knowing about and using the callbacks
- not already knowing that they are deprecated

is probably very close to zero

@Newbie012 what do you think? At least we would also need to update the docs please, and we should widen the rule to be `no-deprecated-properties` and look for all deprecated ones. But I really don't think it's worth the effort ...

NMinhNguyen:
> Thanks, I think we _can_ ship this, but tbh, I don't really think it's worth it. The overlap of:
> 
> * people knowing about the eslint-plugin
> * updating the eslint-plugin to the latest version
> * still being on v4 of react-query
> * knowing about and using the callbacks
> * not already knowing that they are deprecated
> 
> is probably very close to zero

Just to share my perspective, the rule in its current form is still incredibly helpful. At work we have a big monorepo with more than 100 apps, and although we are aware of the deprecation, it’s not easy to upgrade to v5. With this ESLint rule, we can gradually turn on the rule. But with TypeScript, it’s kinda an all or nothing approach - you can’t really upgrade package by package as well as prevent hundreds of developers from using the callbacks in new code.

Newbie012:
I think @NMinhNguyen has a point.

At work, we also didn't upgrade to v5 (due to other reasons). With ESLint, you can enforce rules incrementally, unlike upgrading to v5 which is all or nothing.

LGTM.

TkDodo:
Understood. I would still like these 3 things please:
- docs
- rename to `no-deprecated-options`
- widen the rule to include other options that are deprecated in v4

NMinhNguyen:
> widen the rule to include other options that are deprecated in v4

Could you please help enumerate what these are? Presumably there is a v4 alternative for each one of them?

TkDodo:
> > widen the rule to include other options that are deprecated in v4
> 
> Could you please help enumerate what these are? Presumably there is a v4 alternative for each one of them?

## useQuery options

- the `isDataEqual` option is deprecated
- `onSuccess`, `onError` and `onSettled` are deprecated

## others

- `remove` returned from `useQuery` is deprecated
- the `contextSharing` option on the `QueryClientProvider` is deprecated
- the `logger` option on `QueryClient` constructor is deprecated

it's fine if we just do the `useQuery` options, or, if you want to go deeper, feel free to do the others as well.

iammminzzy:
Thank you so much for the review and suggestions, @TkDodo!

> it's fine if we just do the `useQuery` options

I decided to keep it simple for now and just handle the deprecated callbacks as well as `isDataEqual`.

Additionally, the rule would previously highlight the entire `useQuery()` call, but now it should only target individual options:

### before

image


### after

image

TkDodo (approved):

PR #6630, opened 2024-01-03
Adding Playwright To First Example

ZackDeRose:
Hey folks - adding a PR here to introduce Playwright as a way of catching issues as they creep using the examples that already exist in the repo.

I'm not sure if this is something that y'all would want, so I'm just introducing it primarily with the first example that I encountered (examples/angular/basic).

The test script adds an automated test for the behavior that the example is meant to demonstrate:
+ go to main page
+ first time clicking a link causes a loading indicator to appear
+ go back to main (without a browser refresh)
+ click same link - this time no loading indicator

I also adjusted the pr.yml file so that this will run for affected projects whenever PRs are submitted.

Let me know if this is a welcome addition, I have some ways of adding more e2e tests easily (with Nx core) that shouldn't affect CI time (with Nx Cloud).

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -38,6 +38,8 @@ jobs:
           cache-dependency-path: pnpm-lock.yaml
       - name: Install dependencies
         run: pnpm --filter "./packages/**" --filter query --prefer-offline install
+      - name: Install Playwright Browsers
+        run: pnpm exec playwright install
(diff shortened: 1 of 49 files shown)


TkDodo:
it's an interesting approach, I just think the examples don't do a lot of fancy things, so we have all of these things covered with our existing tests.

PR #6634, opened 2024-01-03
fix(angular-query): type narrowing on query and mutation results

arnoud-dv:
For now this PR only contains type tests ported from Vue Query to demonstrate the problem


arnoud-dv:
Fixed in PR #6662

Issue #6491, opened 2023-12-05
useSuspenseQueries made extra requests after fetching data


angeloaltamiranom:
### Describe the bug

`useSuspenseQueries` will request data again right after receiving a response. As a side effect, this causes the result of `useSuspenseQueries` to change, which will trigger `useEffects` unnecessarily and under the right conditions, infinite query requesting.

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/tanstack-query-bug-forked-gzjkr9?file=%2Fsrc%2FApp.js%3A30%2C56

### Steps to reproduce

1. use `useSuspenseQueries` and mock multiple requests with promises and setTimeout
2. set a suspense boundary on the component that will the do the requesting
3. console log the requests being done and the responses received
4. reload the page and watch for the re-requesting of data

### Expected behavior

I expect the data request to happen once.

### How often does this bug happen?

Every time

### Platform

OS: MacOS
Browser: Chrome
Version: 119.0.6045.199

### Tanstack Query adapter

react-query

### TanStack Query version

5.12.2

TkDodo:
There is likely a problem with how we setup `staleTime` in suspense. Usually, with a single `useSuspenseQuery`, any small staleTime is "good enough", because the query suspends, and then the component renders immediately after the request returns. We set the staleTime to 1 second:

https://github.com/TanStack/query/blob/e7f6d3ae8b2f4230ec8cbea6f4cb8d4af83b7655/packages/react-query/src/suspense.ts#L24-L30

to a avoid a refetch-on-mount here.

However, with `useSuspenseQueries`, we suspend until _all_ queries have finished. Now the first query might finish fast, but the others take a longer time, in which case there will be a refetch for the first query because when the component that uses it finally mounts, more than `staleTime` time has passed already.

Your example though is flawed. You only see two useEffect calls because of `StrictMode`. data is stable. But we can reproduce the issue I'm talking about if we set the `timeout` to a higher value than the default `staleTime`, in which case we will see `Call query N` more than once. This is what I would phrase as an issue, not the multiple effects being called. Your example actually doesn't show multiple requests being made, at least for me 😅 

Here's a fork with what I mean.
- StrictMode is turned off, so only one effect
- staleTime is 500ms
- timeout for requests is 1.5s

The log output I'm seeing is:

```
Call query 1
Call query 2
Call query 3
Fallback
Call query 1
Call query 2
result 1 changed to:  [1]
result 2 changed to:  [2]
[...]

PR #6641, opened 2024-01-04
docs(useSuspenseQueries): add stale time caveat

angeloaltamiranom:
Addresses: https://github.com/TanStack/query/issues/6491


docs/react/reference/useSuspenseQueries.md
```diff
@@ -24,3 +24,7 @@ Same structure as [useQueries](../reference/useQueries), except that for each `q
 - `isPlaceholderData` is missing
 - `status` is always `success`
   - the derived flags are set accordingly.
+
+**Caveat**
+
+Keep in mind that only after all the queries have finished will the suspended component re-mount. Hence, if a query has gone stale in the time it took for all the queries to complete, this will be fetched again at re-mount. To avoid this make sure to set a high enough `staleTime`.
```
TkDodo, line 30:
```suggestion
Keep in mind that the component will only re-mount after **all queries** have finished loading. Hence, if a query has gone stale in the time it took for all the queries to complete, it will be fetched again at re-mount. To avoid this, make sure to set a high enough `staleTime`.
```

PR #6657, opened 2024-01-06
 perf(vue-query): use `shallowRef` instead of `ref`

manniL:
When only replacing the whole `ref` and *not* changing nested values, `shallowRef` allows performance improvements. More details in the [Vue docs](https://vuejs.org/api/reactivity-advanced.html#shallowref) and [my video](https://www.youtube.com/watch?v=HdDVfiHtWHE).

This comes especially handy when using data fetching.

(PS: [This comment](https://www.youtube.com/watch?v=njsGVmcWviY&lc=UgzU0RDqZGTl-oVBVGt4AaABAg) was suggesting this PR, so here I am 😂)

### DIFF
--- a/packages/vue-query/src/useQueries.ts
+++ b/packages/vue-query/src/useQueries.ts
@@ -294,7 +294,7 @@ export function useQueries<
     defaultedQueries.value,
     (options as QueriesObserverOptions<TCombinedResult>).combine,
   )
-  const state = shallowRef(getCombinedResult()) 
+  const state = shallowRef(getCombinedResult())
 
   let unsubscribe = () => {
     // noop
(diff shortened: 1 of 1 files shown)


GabrielHangor:
@manniL Hey Alexander, so what about this feature, has it been implemented? I dont see it anywhere in the docs or how to use that

manniL:
@GabrielHangor Hey! It is an under-the-hood improvement and "just works" ☺️

GabrielHangor:
@manniL  so by default it returns shallowRef as data via the useQuery? What if i need my data deeply reactive - wrapped in a ref as it used to be....

GabrielHangor:
I can see the changes in the useQueries hook but it seems that useQuery still using plain ref. Since useQuery is the most popular tool i guess it might be a good addition to add some kind of a toggle to return data wrapped in a shallowRef instead of ref...imagine having thousands of of records in api response in some data-heavy apps...lots of memory

n0099:
How about `useInfiniteQuery()`?

n0099:
https://github.com/TanStack/query/pull/7733

PR #6680, opened 2024-01-11
fix(query-core): mutation meta typing

Nicoolandgood:
This PR fixes the `QueryMeta` type initialy requiring the generic inferred type to extend `Record<string, unknown>`. The initial behaviour leads to a bug when declaring a global `queryMeta` type in `Register` where the given type is only registered if it extends `Record<string, unknown>`. This case does not appear on `mutationMeta`, this PR matches `queryMeta` behaviour to its own.


TkDodo:
we want to restrict `meta` to being an object. I think that's what the current type does. If `mutationMeta` doesn't have this restriction, I'd rather go the other way and add that ...

Nicoolandgood:
> we want to restrict `meta` to being an object. I think that's what the current type does. If `mutationMeta` doesn't have this restriction, I'd rather go the other way and add that ...

As you suggested, I reverted the previous commit and made the change on `MutationMeta` instead. I also added precisions about global typing registration in the documentation.

TkDodo (approved):

PR #6684, opened 2024-01-11
ci(eslint): add rule to encourage generic naming convention as `^(T|T[A-Z][A-Za-z]+)$`

manudeli:
@TkDodo I saw [your post  in X](https://x.com/TkDodo/status/1726525328254812414?s=20) and @juliusmarminge's solution by @typescript-eslint/naming-convention with regex
I want to add it in this repository

I've been using this rule in various repositories for quite some time, and I really liked the fact that when defining a new generic, it encouraged me not to just name it T. It make coworker and interface user feel better DX.

@juliusmarminge Thanks for sharing your cool solution too.


.eslintrc.cjs
```diff
@@ -45,6 +45,18 @@ const config = {
       { prefer: 'type-imports' },
     ],
     '@typescript-eslint/explicit-module-boundary-types': 'off',
+    '@typescript-eslint/naming-convention': [
+      'warn',
```
manudeli, line 49: I define this rule as `warn` first to use this rule quickly together

.eslintrc.cjs
```diff
@@ -45,6 +45,18 @@ const config = {
...
+      {
+        selector: 'typeParameter',
+        format: ['PascalCase'],
+        leadingUnderscore: 'allow',
+        custom: {
+          regex: '^(T|\\$)[A-Z][a-zA-Z]+[0-9]*$',
+          match: true,
```
TkDodo, line 56: hmm why would we allow leading underscores on a type param? I've set this in other projects to:

```
{
    selector: 'typeParameter',
    format: ['PascalCase'],
    leadingUnderscore: 'forbid',
    trailingUnderscore: 'forbid',
    custom: {
        regex: '^(T|\\$)([A-Z][a-z]+)*$',
        match: true,
    },
},
```

.eslintrc.cjs
TkDodo: could we try to error it and just rename the type parameters if necessary? This should be non breaking ...

manudeli: I made this rule as error in [77af7b3](https://github.com/TanStack/query/pull/6684/commits/77af7b3dde36f11583e35566a4b59ebbae51f43a)
At first, this rule was applied as a warning so that it would be applied gradually and smoothly, but I think it would be better to force it as an error. Thanks!

.eslintrc.cjs
manudeli: > hmm why would we allow leading underscores on a type param?

I banned underscores, and simplified regex in [77af7b3](https://github.com/TanStack/query/pull/6684/commits/77af7b3dde36f11583e35566a4b59ebbae51f43a)

manudeli:
> could we try to error it and just rename the type parameters if necessary? This should be non breaking ...

@TkDodo I will resolve eslint errors soon. wait a minutes please.

manudeli:
> > could we try to error it and just rename the type parameters if necessary? This should be non breaking ...
> 
> @TkDodo I will resolve eslint errors soon. wait a minutes please.

I tried to resolve eslint errors, but there was too many cases that don't need this rule like below. 🥲
How about this rule as warning now? This rule with warning also will make us follow this rule too.

```
export type MaybeRef<T> = Ref<T> | T // isn't it T also good? I think warning is enough for this case

export type MaybeRefOrGetter<T> = MaybeRef<T> | (() => T)

...smillar cases
```

TkDodo:
T alone should be valid according to the regex, no?

manudeli:
> T alone should be valid according to the regex, no?

Ah~, I thought T should be banned before.
Then K, X, Y should be valid right?

manudeli:
@TkDodo sorry, My reply was late because I didn't know much about regex and was studying it. I update regex for this rule as `^([A-Z]|T[A-Z][A-Za-z]+)$`

This rule will make contributors use only below.

### okay
alone T, U, V, etc. is okay
TQueryKey, TQueryFnData is okay

### not okay
KQueryData, Result, Depth, State, DefinedQueryObserver, Runner, Args, TaggedQueryKey, Type, Value, etc. is banned from now. these should be started with T

manudeli:
@TkDodo check this again please

TkDodo:
@manudeli I would prefer it if `T` would be the only allowed single character type parameter. This would make sure that one function can have `T` as a generic, but as soon as you add a second, you can't make `<T,U>` - you need to give meaningful names.

Imo one-character generics are best if there is only one on a given type / function. With two or more, you always need to now what they stand for ...

manudeli:
> @manudeli I would prefer it if `T` would be the only allowed single character type parameter. This would make sure that one function can have `T` as a generic, but as soon as you add a second, you can't make `<T,U>` - you need to give meaningful names.
> 
> Imo one-character generics are best if there is only one on a given type / function. With two or more, you always need to now what they stand for ...

Ahha, I understood your intention. then I'll change it as following you want

manudeli:
@TkDodo I update rule as `^(T|T[A-Z][A-Za-z]+)$` from `^([A-Z]|T[A-Z][A-Za-z]+)$` and I resolved error from this rule in some code. We can check change in [31140ac](https://github.com/TanStack/query/pull/6684/commits/31140acc529821c625fa6e47427b99ed29764229)

TkDodo:
can you please fix the conflicts ?

manudeli:
> can you please fix the conflicts ?

@TkDodo Sure, I resolved conflict now

PR #6686, opened 2024-01-11
chore: use shorthand of sherif consistently

manudeli:
We can check [shorthands for sherif](https://github.com/QuiiBz/sherif?tab=readme-ov-file#rules)
I updated script `test:sherif`'s rules consistently


TkDodo:
i think the shorthand is not really readable. `-p` for IGNORING a package :/

`--ignore-package` was added as workaround for https://github.com/QuiiBz/sherif/issues/41, which is now closed.

So I think updating sherif and removing the workaround might be better if you want to do that?

PR #6687, opened 2024-01-11
ci(eslint): add @cspell/eslint-plugin to prevent misspelling (warning)

manudeli:
I want to add this @cspell/eslint-plugin to prevent misspelling in javascript file ex) jsdoc, variable name

We can see this: the most famous spell checker plugin of VSCode. When I first think to do this, I want to add this as VSCode plugin in .vscode diretory before. but This repository don't have .vscode directory. so I add this as eslint plugin
https://marketplace.visualstudio.com/search?term=spell&target=VSCode&category=All%20categories&sortBy=Relevance

reference: https://cspell.org/


TkDodo:
if a rule is worth having, it should error please. Not sure if this can work here without too much effort ?

manudeli:
> if a rule is worth having, it should error please. Not sure if this can work here without too much effort ?

https://github.com/TanStack/query/pull/6718 I made pull request to make this cspell as eslint rule make error. not just warning

PR #6699, opened 2024-01-14
fix: repair wrong types passed to `DataTag`

lukemorales:
This PR fixes the types that `DataTag` uses to define the type of data in the cache. It was initially using `TData`, but if `select` was used in `queryOptions`, this type would change to the transformed selected data, which is wrong.
The changes in this PR consist of using `TQueryFnData` instead, and add `select` to the type tests to ensure the correct value is being used, since the lack of this prop led to false test positives

### DIFF
--- a/packages/angular-query-experimental/src/__tests__/query-options.test-d.ts
+++ b/packages/angular-query-experimental/src/__tests__/query-options.test-d.ts
@@ -78,7 +78,7 @@ test('should tag the queryKey with the result type of the QueryFn if select is u
     select: (data) => data.toString(),
   })
 
-  assertType<string>(queryKey[dataTagSymbol])
+  assertType<number>(queryKey[dataTagSymbol])
 })
 
 test('should return the proper type when passed to getQueryData', () => {
--- a/packages/angular-query-experimental/src/__tests__/query-options.test-d.ts
+++ b/packages/angular-query-experimental/src/__tests__/query-options.test-d.ts
@@ -78,7 +78,7 @@ test('should tag the queryKey with the result type of the QueryFn if select is u
     select: (data) => data.toString(),
   })
 
-  assertType<unknown>(queryKey[dataTagSymbol])
+  assertType<string>(queryKey[dataTagSymbol])
 })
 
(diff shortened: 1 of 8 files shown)


TkDodo (commented):
Thank you 🙏 Instead of changing all the tests and adding comments, please add one new test that explicitly adds select. The test title should tell us what it tests please :)

lukemorales:
@TkDodo done

packages/angular-query-experimental/src/__tests__/query-options.test-d.ts
```diff
@@ -71,6 +71,16 @@ test('should tag the queryKey with unknown if there is no queryFn', () => {
...
 })
 
+test('should tag the queryKey with the result type of the QueryFn if select is used', () => {
+  const { queryKey } = queryOptions({
+    queryKey: ['key'],
+    queryFn: () => Promise.resolve(5),
+    select: (data) => data.toString(),
+  })
+
+  assertType<unknown>(queryKey[dataTagSymbol])
```
TkDodo, line 81: wondering why this would be `unknown`. Shouldn't it be `string` ?

```suggestion
  assertType<string>(queryKey[dataTagSymbol])
```

packages/angular-query-experimental/src/__tests__/query-options.test-d.ts
```diff
@@ -71,6 +71,16 @@ test('should tag the queryKey with unknown if there is no queryFn', () => {
...
 })
 
+test('should tag the queryKey with the result type of the QueryFn if select is used', () => {
+  const { queryKey } = queryOptions({
+    queryKey: ['key'],
+    queryFn: () => Promise.resolve(5),
+    select: (data) => data.toString(),
+  })
+
+  assertType<string>(queryKey[dataTagSymbol])
```
TkDodo, line 81: aah i messed this up, this should be number of course:

```suggestion
  assertType<number>(queryKey[dataTagSymbol])
```

TkDodo (approved):

PR #6716, opened 2024-01-16
fix(query-core): computed properties of QueryObserverResult

YanZisuka:
`isLoading` (isPending && isFetching) is a value calculated from `isPending`.

Therefore, if **isPending is false, isLoading becomes false**, and if **isLoading is true, isPending also becomes true**, so modify the related type.


TkDodo (commented):
I think your changes are correct. Could you please add some type-tests to verify ?

YanZisuka:
@TkDodo 

Thank you for your quick response!

Type testing is a bit new, but I've added some cases. Thanks for the review.

TkDodo (approved):

PR #6736, opened 2024-01-19
fix(query-next-experimental): escape </ when hydrating

### LINKED ISSUE #6735: ReactQueryStreamedHydration hydrated does not properly escape </script>
### Describe the bug

If the data that is being hydrated has a `</script>` anywhere in the text, then it breaks the web page.
Specifically, the rendered HTML does not escape the `</script>` causing the browser to prematurely end the tag.

https://codesandbox.io/p/devbox/young-cloud-pq72kc

### Steps to reproduce

Checked the linked codesandbox

1. Add ReactQueryStreamedHydration to your NextJS app.
2. Create an API endpoint that returns back text/json and has `</script>` in the response somewhere.
3. Use `useSuspenseQuery` to query that that endpoint.
4. Observe in the rendered HTML that the following is present
```
<script>
window["__RQ:R6kq:"] = window["__RQ:R6kq:"] || [];window["__RQ:R6kq:"].push({"mutations":[],"queries":[{...
...</script>...
}]})
</script>
```

The above is invalid HTML, notice that there is an unescaped `</script>` because it was present in the originally fetched data.

### Expected behavior

The rendered data should escape `</script>`, perhaps by searching for all instances of `</script>` and replacing with `<\/script>`

Every time

### Screenshots or Videos

_No response_

### Platform

- Linux
- Browser: Chrome
- NextJS 14.0.4

### Tanstack Query adapter

None

### TanStack Query version

v5.17.10

### TypeScript version

v5.3.3

### Additional context

_No response_

ammubhave:
Escape </ when hydrating so that it doesn't break rendering. The recommendation after some light reading (https://stackoverflow.com/questions/14780858/escape-in-script-tag-contents) seems to be to escape all `</` in script tags with `<\/`

Fixes #6735

See attached pull request for reproduction.

Test Plan: Tested this locally by running the example attached in the pull request with this fix in place.

### DIFF
--- a/packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
+++ b/packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
@@ -130,7 +130,7 @@ export function createHydrationStreamProvider<TShape>() {
         <script
           key={count.current++}
           dangerouslySetInnerHTML={{
-            __html: html.join(''),
+            __html: html.join('').replaceAll('</', '<\\/'),
           }}
         />
       )


TkDodo:
any chance you could write a test case for this fix?

TkDodo:
thanks for the PR, we fixed this holistically with:

https://github.com/TanStack/query/commit/f2ddaf2536e8b71d2da88a9310ac9a48c13512a1

PR #6753, opened 2024-01-23
Update HydrationStreamProvider.tsx + Advanced SSR docs to fix unnecessary suspensions

### LINKED ISSUE #6116: useSuspenseQuery infinite refetch after SSR (Next.js App Router)
### Describe the bug

After a page using `useSuspenseQuery` is SSR'd in the Next.js app router, the query will infinitely refetch. This also occurs if a `loading.tsx` file is added. 

https://github.com/mgreenw/tanstack-query-suspense-bug

### Steps to reproduce

1. Install deps `pnpm i`
2. Run the server `pnpm dev`
3. Go to `http://localhost:3000`
4. View the devtools network tab

### Expected behavior

The query should only refetch at the expected times (window focus, mount, etc).

Every time

### Screenshots or Videos

Screenshot 2023-10-07 at 6 50 24 PM

### Platform

- OS: macOS 13 Ventura
- Browser: Firefox, Chrome, Safari

### Tanstack Query adapter

react-query

### TanStack Query version

5.0.0-rc.4

### TypeScript version

5.2.2

### Additional context

Found this first when fiddling with `trpc` in the Next.js app router. Seems to be affecting both v4 and v5.

mgreenw:
This PR is based on my findings in #6116 and fixes two cases of unnecessary suspensions during streaming SSR, one on code and the other in the advanced SSR docs. See [the full writeup in this comment](https://github.com/TanStack/query/issues/6116#issuecomment-1904051005), which should have all the info needed for reviewing this PR.

fixes #6116


packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
```diff
@@ -143,31 +143,28 @@ export function createHydrationStreamProvider<TShape>() {
       onEntriesRef.current = props.onEntries
     })
 
-    React.useEffect(() => {
+    // Setup and run the onEntries handler on the client only, but do it during
+    // the initial render so children have access to the data immediately
+    // This is important to avoid the client suspending during the initial render
+    // if the data has not yet been hydrated.
+    // @ts-expect-error window isn't defined on the server
+    if (typeof window !== 'undefined' && !window[id]?.initialized) {
```
TkDodo, line 151: we have an `isServer` util function that works slightly different than just checking for window. could we re-use that one ?

docs/react/guides/advanced-ssr.md
```diff
@@ -34,19 +34,37 @@ The first step of any React Query setup is always to create a `queryClient` and
...
+let browserQueryClient: QueryClient | undefined = undefined
+
+function getQueryClient() {
+  if (typeof window === 'undefined') {
+    // Server: always make a new query client
+    return makeQueryClient()
+  } else {
+    // Browser: make a new query client if we don't already have one
+    // This is very important so we don't re-make a new client if React
+    // supsends during the initial render
```
TkDodo, line 58: I guess that's why a `Suspense` boundary is a good idea - so that you can create a `queryClient` in React State outside of it, and it won't be thrown away. I don't mind the `makeQueryClient()`function - it's good that it uses a singleton on the client and a new one every time on the server 👍 . I would still probably mention somewhere that this is only needed if you don't have an explicit suspense boundary further down the tree...

TkDodo:
@juliusmarminge would you mind taking a look here? Also I think the comment in the issue is super interesting:

- https://github.com/TanStack/query/issues/6116#issuecomment-1904051005

docs/react/guides/advanced-ssr.md
```diff
@@ -34,19 +34,37 @@ The first step of any React Query setup is always to create a `queryClient` and
...
-      new QueryClient({
-        defaultOptions: {
-          queries: {
-            // With SSR, we usually want to set some default staleTime
-            // above 0 to avoid refetching immediately on the client
-            staleTime: 60 * 1000,
-          },
-        },
-      }),
-  )
```
Ephem, line 49: I'm not sure these parts of the docs needs to change? Note that most of the examples on this page until the streaming library part uses `useQuery` and prefetching, which I don't think should be susceptible to this edge case? But I haven't thought about this as long as you have, so maybe I'm missing something?

Not something I feel strongly about though and happy to keep it as is to keep it aligned too.

If we do keep it as is, I would still love a comment here along the lines of "if you are using `useSuspenseQuery`, initiating the `QueryClient` as state can have edge cases, see the docs on streaming further down".

docs/react/guides/advanced-ssr.md
```diff
@@ -350,8 +368,37 @@ import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
...
+    // Server: always make a new query client
+    return makeQueryClient()
+  } else {
+    // Browser: make a new query client if we don't already have one
+    // This is very important so we don't re-make a new client if React
+    // supsends during the initial render
+    if (!browserQueryClient) browserQueryClient = makeQueryClient()
+    return browserQueryClient
+  }
+}
```
Ephem, line 396: I really wish they would get around to implementing `cache()` in client components, it would make things like this so much neater. 😄 Until then, this looks good though!

Ephem (commented):
Nice find and thanks for submitting a PR! Suspense is a tricky beast.. I left a few comments but I do think this look good overall.

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
```diff
@@ -143,31 +143,28 @@ export function createHydrationStreamProvider<TShape>() {
       onEntriesRef.current = props.onEntries
     })
 
-    React.useEffect(() => {
+    // Setup and run the onEntries handler on the client only, but do it during
+    // the initial render so children have access to the data immediately
+    // This is important to avoid the client suspending during the initial render
+    // if the data has not yet been hydrated.
```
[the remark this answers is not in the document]
mgreenw: Interesting callout! I _think_ it's a different situation here, and what we're doing is safe. `HydrationBoundary` could happen at any part of the tree, whereas there will only be one `ReactQueryStreamedHydration` and it will be high up in the tree / wrapping the whole thing. This code will only be called once on boot, and we went any streamed data to be available ASAP on the initial render phase.

Definitely let me know if you see any problems with that logic!

docs/react/guides/advanced-ssr.md
mgreenw: Agreed 💯

docs/react/guides/advanced-ssr.md
mgreenw: I think it's important to put this in the docs in both places. I'm betting that many folks (like myself) will copy-paste the first example that they see, and having the `useState` in there will lead to a broken example **by default** until they add a suspense boundary.

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
mgreenw: Done!

packages/react-query-next-experimental/src/HydrationStreamProvider.tsx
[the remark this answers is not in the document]
mgreenw: Thanks for taking this across the finish line @TkDodo!

PR #6758, opened 2024-01-23
[WIP] New useQueryState Hook

crutchcorn:
Closes https://github.com/TanStack/query/discussions/6756

# TODO:
- [ ] Fix any broken tests
- [x] Migrate `useQueryState` logic to `useMutationState`
- [ ] Add working tests for `useQueryState`
- [ ] Mark `useIsFetching` as deprecated
- [ ] Mark `useIsMutating` as deprecated
- [ ] Update docs to note deprecated fields
- [ ] Add reference docs


crutchcorn:
There appear to be two failing tests on this PR and I can't seem to figure out _why_ they're failing. Namely, they are the `useFetching` API that's been changed.

While debugging, I've noticed the following:

- `notifyManager.schedule(onStoreChange)` in `useQueryState` is called properly with the single item
- `onStoreChange` is not ever called

I can't seem to figure out why this is the case or why it works in the other tests, but not these.

It also appears that there's some timing issues with `sleep(10)` that I've resolved by replacing with `new Promise` changes.

packages/react-query/src/useIsFetching.ts
```diff
@@ -1,24 +1,13 @@
...
-        queryCache.subscribe(notifyManager.batchCalls(onStoreChange)),
-      [queryCache],
-    ),
-    () => client.isFetching(filters),
-    () => client.isFetching(filters),
-  )
+  return useQueryState(
+    { filters: { ...filters, fetchStatus: 'fetching' } },
+    queryClient,
+  ).length
```
TkDodo, line 12: I don't really know why the tests fail here - they really shouldn't, the implementations are pretty much the same. I tracked it down to the `queryObserverOptions` event not being fired initially with the new impl.

these are the events we're receiving from the cache subscription:

```
useQueryState observerResultsUpdated
useQueryState updated
fire observerOptionsUpdated
useQueryState observerOptionsUpdated
useQueryState observerRemoved
```

with the old impl, we get:

```
fire observerOptionsUpdated
useIsFetching observerOptionsUpdated
useIsFetching observerResultsUpdated
useIsFetching updated
fire observerOptionsUpdated
useIsFetching observerOptionsUpdated
useIsFetching observerRemoved
```

so we get one more `observerOptionsUpdated` event, at which time we are still fetching. In the new impl, the first event we get is `observerResultsUpdated`, at which point we already have data. I don't currently know why that is ...

TkDodo: pff, it seems to have something to do with the ref workaround.

I refactored it back to the original implementation, step by step. Even if we store the `number` result of `client.isFetching()` in a ref, then read from the ref in `getSnapshot`, it fails, but if we directly read from `client.isFetching()` in `getSnapshot`, it works.

This is so baffling, I guess "don't read refs during render" is coming to bite us here . I tracked the difference by logging the value of the ref and reading `isFetching()` directly, and there are indeed differences:

```
getSnapshot 0 0
getSnapshot 0 0
getSnapshot 0 1
getSnapshot 0 1
getSnapshot 0 1
```

I think moving the read + structural sharing into `getSnapshot` could solve it 🤔

TkDodo: fixed it: 4f000317

`useMutationState` likely needs the same change

packages/react-query/src/__tests__/useMutationState.test.tsx
```diff
@@ -66,12 +66,15 @@ describe('useIsMutating', () => {
...
-    function IsMutating() {
+    function IsMutatingBase() {
       const isMutating = useIsMutating({ mutationKey: ['mutation1'] })
       isMutatings.push(isMutating)
       return null
     }
 
+    // Memo to avoid other `useMutation` hook causing a re-render
+    const IsMutating = React.memo(IsMutatingBase)
+
```
TkDodo, line 77: I think it's also fine to change the assertion here to 0,1,1,0 instead.

TkDodo:
closing for housekeeping

PR #6768, opened 2024-01-25
docs: Point to latest eslint rule doc link

jprosevear:


lachlancollins:
Hi @jprosevear , I think you need to add `framework` to the path (`https://tanstack.com/query/latest/docs/framework/react/eslint/${ruleName}`)

packages/eslint-plugin-query/src/utils/create-rule.ts
```diff
@@ -3,7 +3,7 @@ import { detectTanstackQueryImports } from './detect-react-query-imports'
 import type { EnhancedCreate } from './detect-react-query-imports'
 
 const getDocsUrl = (ruleName: string): string =>
-  `https://tanstack.com/query/v4/docs/eslint/${ruleName}`
+  `https://tanstack.com/query/latest/docs/react/eslint/${ruleName}`
```
TkDodo, line 6: urls have changed and are now:

https://tanstack.com/query/latest/docs/framework/react/eslint/stable-query-client

lachlancollins: Since these should actually be shown for all frameworks, I've moved them to: https://tanstack.com/query/latest/docs/eslint/stable-query-client

However, both of the above URLs would now correctly redirect thanks to changes in https://github.com/TanStack/tanstack.com

jprosevear:
@lachlancollins @TkDodo updated

lachlancollins:
@jprosevear I think you'll need to remove the extra newline

jprosevear:
> @jprosevear I think you'll need to remove the extra newline

@lachlancollins Yup.  For some reason i had autoformat on save temporarily off.  Fixed.

PR #6769, opened 2024-01-25
fix(experimental_createPersister): setQueryData, ensureQueryData supports persister

### LINKED ISSUE #6310: `setQueryData()` should be persisted too, right?
### Describe the bug

I'm using the new `experimental_createPersister` like this:

```ts
import {
  experimental_createPersister,
  type AsyncStorage,
} from "@tanstack/query-persist-client-core";
import { get, set, del, createStore, type UseStore } from "idb-keyval";

function newIdbStorage(idbStore: UseStore): AsyncStorage {
  return {
    getItem: async (key) => await get(key, idbStore),
    setItem: async (key, value) => await set(key, value, idbStore),
    removeItem: async (key) => await del(key, idbStore),
  };
}

export const queryClient = new QueryClient({
  defaultOptions: {
    queries: {
      gcTime: 1000 * 30, // 30 seconds
      persister: experimental_createPersister({
        storage: newIdbStorage(createStore("db_name", "store_name")),
        maxAge: 1000 * 60 * 60 * 12, // 12 hours
      }),
    },
  },
});
```

But I just found out that if I call

```ts
const SetPlayerTeam = (team: Team) => {
  queryClient.setQueryData<Player>(
    ['player', team.playerId],
    (old) => old && { ...old, team }
  );
};
```

the in memory cache works (the `team` is added to `player`) but there is not `setItem()` call for this last `queryClient.setQueryData()`.

**I think this is wrong, because this change should be persisted too.**

_I'll create a reproduction if you think it might be useful._

-

### Steps to reproduce

-

### Expected behavior

-

Every time

### Screenshots or Videos

_No response_

### Platform

Chrome.

### Tanstack Query adapter

svelte-query

### TanStack Query version

5.4.3

### TypeScript version

5

### Additional context

I'm using:

```json
"@tanstack/eslint-plugin-query": "5.0.5",
"@tanstack/query-persist-client-core": "5.4.3",
"@tanstack/svelte-query": "5.4.3",
"@tanstack/svelte-query-devtools": "5.4.3",
```

DamianOsipiuk:
fixes #6310

This PR attempts to fix some gaps of experimental persister.

- `setQueryData`/`setQueriesData` - can be kept as a sync function, which kicks persister in the background if defined
- `ensureQueryData` - it's already async, so we can just hook up before kicking of `fetch`

Also small internal restructure of  `createPersister` to allow re-usability of internals.

We still need some way to restore persisted items on-demand and in bulk.

### DIFF
--- a/packages/query-core/src/infiniteQueryBehavior.ts
+++ b/packages/query-core/src/infiniteQueryBehavior.ts
@@ -116,7 +116,7 @@ export function infiniteQueryBehavior<TQueryFnData, TError, TData, TPageParam>(
       }
       if (context.options.persister) {
         context.fetchFn = () => {
-          return context.options.persister?.(
+          return context.options.persister?.persisterFn(
             fetchFn as any,
             {
               queryKey: context.queryKey,
--- a/packages/query-core/src/query.ts
+++ b/packages/query-core/src/query.ts
@@ -394,7 +394,7 @@ export class Query<
       }
       this.#abortSignalConsumed = false
       if (this.options.persister) {
-        return this.options.persister(
+        return this.options.persister.persisterFn(
           this.options.queryFn,
           queryFnContext as QueryFunctionContext<TQueryKey>,
           this as unknown as Query,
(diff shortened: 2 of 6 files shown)


TkDodo (commented):
I'm not sure if this functionality should be in the query-core. Let me think about it

TkDodo:
The discussion is continuing on the issue, I posted my thoughts there. closing this because I don't think we should go in this direction

PR #6775, opened 2024-01-26
angular-query: Update quick-start.md

mikelgo:
Some minor fixes to update the quick start example to a version which would actually compile


docs/framework/angular/quick-start.md
```diff
@@ -12,8 +12,10 @@ If you're looking for a fully functioning example, please have a look at our [ba
 ### Provide the client to your App
 
 ```ts
+import {provideHttpClient} from '@angular/common/http';
+import {provideAngularQuery} from '@tanstack/angular-query-experimental'
```
mikelgo, line 16: QUestion: where to import `QueryClient` from? it is not exported by @tanstack/angular-query-experimental.
Is `@tanstack/query-core` requiered?

lachlancollins: You can import it from the angular package, since it re-exports everything from query-core. See here: https://github.com/TanStack/query/blob/main/packages/angular-query-experimental/src/index.ts

mikelgo: I see. Then it is just a stackblitz issue i was tricked by. When using the package in stackblitz it is not able to import QueryCLient.

lachlancollins: I might be wrong, @arnoud-dv could you confirm?

arnoud-dv: Yes you're correct Lachlan. Importing QueryClient from `@tanstack/angular-query-experimental` is how it's intended to be used.

@mikelgo See `packages/angular-query-experimental/src/index.ts`, everything is re-exported from core. There is no need to install or import from `@tanstack/query-core` separately.

mikelgo: As mentioned it is just a stackblitz issue, see https://stackblitz.com/edit/stackblitz-starters-xiui1d?file=src%2Fmain.ts

In the "real world" all is fine.  will also update the pr to reflect the import

arnoud-dv: @mikelgo That Stackblitz has an extremely old TypeScript version which Angular itself is not compatible with, and also contains tslint which has been deprecated for many years. It appears this is based on a very outdated template?

PR #6776, opened 2024-01-26
refactor(angular-query): provideAngularQuery by default instantiates QueryClient by itself

mikelgo:
Question:
What is the use case for having this signature `export function provideAngularQuery(queryClient: QueryClient): EnvironmentProviders`? 
If one wants to override the QueryCLient it can be done the "angular-way" via DI `{provide:QUERY_CLIENT, useClass: MyClient}` or `provideQueryClient(() => new MyClient)`
So I believe we can remove the signature? Or am I missing something?.

### DIFF
--- a/packages/angular-query-experimental/src/providers.ts
+++ b/packages/angular-query-experimental/src/providers.ts
@@ -9,9 +9,13 @@ import { provideQueryClient } from './inject-query-client'
 import type { EnvironmentProviders } from '@angular/core'
 
 export function provideAngularQuery(): EnvironmentProviders
-export function provideAngularQuery(queryClient: QueryClient): EnvironmentProviders
-export function provideAngularQuery(queryClient?: QueryClient): EnvironmentProviders {
-  const client =  queryClient ?? new QueryClient()
+export function provideAngularQuery(
+  queryClient: QueryClient,
+): EnvironmentProviders
+export function provideAngularQuery(
+  queryClient?: QueryClient,
+): EnvironmentProviders {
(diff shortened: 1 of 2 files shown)


mikelgo:
I dont get why pr/format fails. When I run `pnpm run test:format` locally it succeeds

mikelgo:
@arnoud-dv 
If you have time to review the PR let me know what you think. 
Depending on if you agree with the change I will of course also update the affected test suites and/or add new tests.

mikelgo:
> Hi, I'd prefer to keep this explicit:
> 
> * Most important: explicitly creating the `QueryClient` aligns with how it's done in other adapters. Unless there is a very good reason I don't want to deviate. It's great that if a developer used React Query before they would be able to get up to speed quickly on Angular Query or the other way around. Alignment also has maintenance benefits.
> * Not being explicit makes it unclear what happens when `provideAngularQuery` with no parameter is called again down the DI hierarchy chain. Is a new client created implicitly? This would need to be documented and the developer would need to read it to understand what's going on. When it's explicit reading the code would be enough.
> * Having to pass the client explicitly is a DI mechanism in itself, as it is in React Query. React supports DI through context. Using a provider function in Angular Query I think it's aligned well to React Query while also being Angular idiomatic.
> * When the client is already created in application code it's easier and more discoverable to change options.
> * Not having to pass a client to a function that for most applications will be called once would be such a tiny convenience it isn't worth these tradeoffs IMO.

Thanks for your thoughts. I can understand the reasoning behind.

PR #6787, opened 2024-01-29
ci(vitest): use name field of package.json for name field of vitest config consistently

manudeli:
In vite.config.ts, we use package name like (we use test name as `react-query` in `@tanstack/react-query`)

And I checked `resolvePackageJson: true` in global tsconfig, so I update viteConfig.test.name as packageJson.name. so we can manage test name like packageJson.name as one consistently


lachlancollins (approved):
This makes sense, thanks @manudeli !

PR #6799, opened 2024-01-31
[WIP] [Experiment] Refactor Subscribable and NotifyManager to TanStack Store

crutchcorn:
This PR refactors the existing implementations of `NotifyManager` and `Subscribable` instances into TanStack Store utilities as an experiment to see if we're able to consolidate on the tooling we want to use for our projects.

> This is a WIP and experiment; this may not ultimately yield any changes to Query.

### DIFF
--- a/packages/query-core/package.json
+++ b/packages/query-core/package.json
@@ -44,5 +44,8 @@
     "test:lib:dev": "pnpm run test:lib --watch",
     "test:build": "publint --strict",
     "build": "tsup"
+  },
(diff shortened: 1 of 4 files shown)


TkDodo:
looks good, but the failing tests uses the focusManager, so I think we'd need to look into this

TkDodo:
looking at the preview build, onlineManager seems to work fine, but focusManager does not - I can't get window focus refetching to work.

PR #6836, opened 2024-02-04
docs(examples): improve react-router example

ruisaraiva19:
Changes made to the React Router example:
- Converted to TypeScript
- Added `errorComponent` to all routes
- Fixed CSS styles bleeding into the React Query Devtools
- Added a route to create new contacts and reused the edit contact form
- Use `queryOptions` utility function to type query options
- Use `queryClient.ensureQueryData` instead of `queryClient.fetchQuery` + `queryClient.getQueryData`
- Use `useSuspenseQuery` instead of `useQuery` since route components only render after the loader finishes


examples/react/react-router/package.json
```diff
@@ -20,7 +21,11 @@
     "sort-by": "^1.2.0"
   },
   "devDependencies": {
+    "@types/react": "^18.2.45",
+    "@types/react-dom": "^18.2.18",
+    "@types/sort-by": "^1.2.3",
     "@vitejs/plugin-react": "^4.2.1",
+    "typescript": "5.2.2",
```
TkDodo, line 28: @lachlancollins I think this would fail on main again because `sherif` doesn't run on PRs....

pnpm-lock.yaml
TkDodo, line 1: that's a lot of changes in the lock file for just some types updates in package.json 🤔

ruisaraiva19: Lock file updated.

TkDodo (approved):
awesome work 👏

examples/react/react-router/package.json
lachlancollins: Let me temporarily fix this by adding another task like format and knip!

PR #6855, opened 2024-02-08
Adjust undefined queryKey validation warning

karlkristopher:
A small change, but if an undefined `queryKey` is acceptable, would it be ok to not show a warning here?


TkDodo:
query keys can never be undefined

PR #6863, opened 2024-02-09
ci: Introduce Nx agents

ZackDeRose:
Playing with Nx agents here.

Here's what I'm seeing:

- We can do a bit better by bumping up the agent count from 3 => 12 (free and not as restrictive as github seems to be when asking for a bunch of machine), and reducing the parallel option from 5 => 3 to not overload any one machine. Looks like now we're getting 3.5 minutes for a full run. We might try for more agents here, it's worth exploring!

- A downside from the legacy DTE setup to Nx agents is in legacy, we could setup workers and main machines all in parallel, now we have to setup the main job and then setup all other agents. We're losing about 20s here with cached node_modules - looks like this is made up for by bumping up agent counts though. When we launch Nx Workflows, we should be able to get that 20s back.

- There's some more optimizations we could try later, adding in small/med/large agent loads could help here for getting the full swarm when we need to run everything, and not get overly excited for smaller PRs

- Potentially we can add in the format and knip tasks to the swarm. This is probably trivial for perf, but it could help in the sense that all tasks can be run the same way (instead of setting up specific machines for the format and knip tasks), and we could see it in the nx cloud task list instead of spelunking github action logs!

- Nx Agents can be used for the main job as well - looks like those workflows are growing up to ~10 minutes now. I think now it's solid enough that we can use it reliably for the releases, while using the --skip-cache option to ensure we're running all checks before publishing!

@lachlancollins / @TkDodo let me know thoughts here, I'll clean this up a bit so we can merge if you like!


.github/workflows/pr.yml
```diff
@@ -41,7 +41,7 @@ jobs:
         with:
           main-branch-name: 'main'
       - name: Start CI Orchestrator
-        run: npx nx-cloud start-ci-run
+        run: npx nx-cloud start-ci-run --distribute on="3 linux-medium-js"
```
barbados-clemens, line 44:
```suggestion
        run: npx nx-cloud start-ci-run --distribute-on="3 linux-medium-js"
```

ZackDeRose: you're too good to me, caleb

ZackDeRose:
@lachlancollins / @TkDodo we're in a merge-able state now

lachlancollins (approved):

lachlancollins:
@ZackDeRose regarding the optimisations you mentioned above, how difficult would it be to implement agents for the `ci.yml` workflow? As you mentioned, CI times are now 9+ mins as we're testing examples too.

PR #6868, opened 2024-02-10
chore: Replace plugin:react/recommended with plugin:react/jsx-runtime

lachlancollins:
Replaced `plugin:react/recommended` with `plugin:react/jsx-runtime`

Disables https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/react-in-jsx-scope.md, which causes very annoying (and incorrect) highlighting of all JSX in the React examples

This is compatible with CRA v4 (see https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)

Also needed to set `"jsx"` to `"react-jsx"` in tsconfig, but this has no impact on the build output (generated by tsup).

### DIFF
--- a/examples/react/algolia/.eslintrc
+++ b/examples/react/algolia/.eslintrc
@@ -1,6 +1,6 @@
 {
   "extends": [
-    "plugin:react/recommended",
+    "plugin:react/jsx-runtime",
     "plugin:react-hooks/recommended",
     "plugin:@tanstack/eslint-plugin-query/recommended"
   ]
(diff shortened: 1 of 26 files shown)


xiaolannuoyi:
5.28.4 This version has this issue，Has it been merged？

lachlancollins:
> 截屏2024-03-16 21 54 33

Hi @xiaolannuoyi , could you please create an issue with more details about your setup and a repro?

PR #6877, opened 2024-02-12
refactor(query-core): use enums instead of strings

Mini-ghost:
Currently, our `query` and `mutation` implementations utilize strings as action types, as illustrated below:

```ts
this.#dispatch({
  data,
  type: 'success',
  dataUpdatedAt: options?.updatedAt,
  manual: options?.manual,
})
```

In this PR, I am proposing the adoption of enums as a more efficient alternative:

```ts
const enum QUERY_ACTION_TYPE {
  FAILED,
  FETCH,
  SUCCESS,
  ERROR,
  INVALIDATE,
  PAUSE,
  CONTINUE,
  SET_STATE,
}

this.#dispatch({
  data,
  type: QUERY_ACTION_TYPE.SUCCESS,
  dataUpdatedAt: options?.updatedAt,
  manual: options?.manual,
})
```

This approach not only keeping the maintainability and readability of our code but also optimizes performance by reducing the bundle size.

```ts
this.#dispatch({
  data,
  type: 2, // Corresponds to QUERY_ACTION_TYPE.SUCCESS
  dataUpdatedAt: options?.updatedAt,
  manual: options?.manual,
})
```

### DIFF
--- a/packages/query-core/src/query.ts
+++ b/packages/query-core/src/query.ts
@@ -2,6 +2,7 @@ import { noop, replaceData, timeUntilStale } from './utils'
 import { notifyManager } from './notifyManager'
 import { canFetch, createRetryer, isCancelledError } from './retryer'
 import { Removable } from './removable'
+import { type QueryCache } from './queryCache'
 import type {
   CancelOptions,
   DefaultError,
@@ -14,7 +15,6 @@ import type {
   QueryStatus,
   SetDataOptions,
 } from './types'
-import type { QueryCache } from './queryCache'
 import type { QueryObserver } from './queryObserver'
 import type { Retryer } from './retryer'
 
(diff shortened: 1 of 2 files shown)


packages/query-core/src/query.ts
```diff
@@ -86,43 +86,55 @@ export interface FetchOptions {
   meta?: FetchMeta
 }
 
+// eslint-disable-next-line no-shadow
+const enum QUERY_ACTION_TYPE {
```
TkDodo, line 90: what is this shadowing ?

Mini-ghost: I'm not sure what's going on to cause the eslint error. I checked the file contents and didn't find the issue pointed out by eslint in the context

截圖 2024-02-12 下午3 50 51

TkDodo:
I just read [const enum pitfalls](https://www.typescriptlang.org/docs/handbook/enums.html#const-enum-pitfalls) so I don't think this is worth doing

Mini-ghost:
@TkDodo Thank you for your feedback. I have taken the time to read through the concerns you've raised, and I admit that these are aspects I had not fully considered before.

However, I'm not entirely sure if the issues mentioned in the document are a significant concern in the context of this PR. From my perspective, the enums in question are only used internally within `query` and `mutation` implementations, without involving cross-file interactions. Could there be aspects I've overlooked or not considered in enough detail?

---

Additionally, I would like to offer a bit more context that may help clarify my point further. Here is a sample link that demonstrates the potential for reducing package size through the approach I've proposed. I believe this example could make it more evident how the changes could be beneficial.

https://www.typescriptlang.org/play?ts=5.3.3#code/MYewdgzgLgBApmArgWxgQQMIBUCSB5AOQH0sBNABQFEYBvAKBhgGUBVDDSppgGgZkoBKAvALoBfOnQBmiMMCgBLcDAAmCiAAcAhlGAALGAAot8pWABctGFACeGuJcy5CJCtTEBKWnwgB3BboGxqbgAHS29l70jIzAWhBw6Nj4xGRUoazsnEzmfDEwoJAgADZwocUgAOaGAORqmjr6MBCIwMBwEBA1HnkxAE5wUIh9YL1xCUnOqW6hgsICufmx4BAlZRXVderagfB9fSB93b2MA0MjfBISdPU7+oY01nYOkymu6ZkcXDCeQA

Thank you for taking the time to review my PR. Regardless of the outcome, this experience has been incredibly beneficial for me.

TkDodo:
the way I understood [this bug report](https://github.com/microsoft/TypeScript/issues/5219#issue-110947903) is that if you compile the lib with one version of ts and consumers use a different version, they might run into this error. The fixes are to not use cost enums, or to preserve them, but preserving results in larger code usually. So, imo, not worth it.

Mini-ghost:
I think I need to spend some time studying to fully understand. But thank you for your patient answer; I really learned a lot from it.

PR #6887, opened 2024-02-13
Update installation.md

bbolek:
When we copy, it was copying both as they were in same script. So it is better to have them in seperate scripts


TkDodo:
why would you install v3 from scratch ?

PR #6907, opened 2024-02-16
fix(*): remove unnecessary omitted field of MutationObserverOptions

manudeli:


packages/angular-query-experimental/src/types.ts
```diff
@@ -137,7 +137,7 @@ export interface CreateMutationOptions<
   TContext = unknown,
 > extends Omit<
     MutationObserverOptions<TData, TError, TVariables, TContext>,
-    '_defaulted' | 'variables'
```
manudeli, line 140: There is no 'variable' field to omit in MutationObserverOptions

manudeli (commented):
I left my intention

packages/angular-query-experimental/src/types.ts
arnoud-dv: All adapters except Vue Query contain this type. So I think it should be changed for all of those?

manudeli: Sure, thanks for your eagle eyes 👍

TkDodo (approved):

PR #6913, opened 2024-02-17
refactor(query-codemods): unify directory name as name field of package.json

manudeli:
I want to unify directory name as nam field of package.json but I couldn't find why error is occurred in workflow https://github.com/TanStack/query/actions/runs/7938657398/job/21677740503?pr=6913

Could maintainers care this please instead of this please?


codecov.yml
```diff
@@ -33,7 +33,7 @@ component_management:
     - component_id: query-codemods
       name: '@tanstack/query-codemods'
       paths:
-        - packages/codemods/**
+        - packages/query-codemods/**
```
manudeli, line 36: Only this package have different directory name with name field of package.json. so I matched it

manudeli (commented):
I left my intention

PR #6918, opened 2024-02-17
perf(queryClient): use `queryCache.get` instead of `find({ queryKey })`

### LINKED ISSUE #6594: useQueries and large datasets is crashing browsers
### Describe the bug

In a large react project we are using useQueries for fetching relatively large amounts of data that is deserialized with batshit.

This gives problems in useQueries when having to request more than 20k entities (each entity is represented by a GUID), this has been up in a discussion here https://github.com/TanStack/query/discussions/6305, specially with requests around 30k and larger is troublesome

Note that the codesandbox example is running with query only, without any other plugins / helper libraries

https://codesandbox.io/p/sandbox/react-query-large-queries-set-c5244l

### Steps to reproduce

Change the number of requested ids in App.tsx on line 8 (currently 20k ids)

### Expected behavior

Would expect that the browser does not crash.

Every time

### Screenshots or Videos

_No response_

### Platform

Chrome and firefox tested on both windows 11 and ubuntu 22.04. All have issues with the example

### Tanstack Query adapter

react-query

### TanStack Query version

5.7.0

### TypeScript version

5.2.2

### Additional context

As noted, there are a discussion about this as well, but chose to rise this as a bug as well

### LINKED ISSUE #6783: `setQueryData` is slow with large datasets
### Describe the bug

`setQueryData` iterates over all entries in the cache to find matches. This leads to performance issues when priming large datasets.

https://codesandbox.io/p/sandbox/react-query-setquerydata-complexity-xl4d3t

### Steps to reproduce

```
const myData = range(1, 10000) // eg. a list of search responses received from the server
for (const i of myData) {
  queryClient.setQueryData(i, i)
}
```

Internally, `setQueryData` uses `find`, which iterates over each entry in the cache. Assuming the cache has already been populated, the above function quadratic.

The issue can be avoided by getting the Query-Object directly from the hash-table:

```ts
// this returns the cache for this key if existing, or creates a new one with default options:
// Unlike `setQueryData`, it will use the hashtable to find the entry.
const query = queryClient.getQueryCache().build(client, { queryKey: /* ... */ }) 
query.setData(data)
```

### Expected behavior

I'd suggest that `setQueryData` should either...

- be updated to only work with *exact* matches, since I believe most use-cases will only want to update a single entry
- be provided with an `exact` -filter option that uses the hash-table internally

Every time

### Screenshots or Videos

Screenshot 2024-01-27 at 22 58 45

### Platform

MacOS, Chrome 23.0.6262.5 

### Tanstack Query adapter

vanilla

### TanStack Query version

5.17.19

### TypeScript version

_No response_

### Additional context

This behaviour has also been observed by users in https://github.com/TanStack/query/issues/1633. However, it was previously believed to be due to structural sharing.

SebKranz:
Methods that access a query by its `queryKey` can bypass `find` to avoid performance issues in large caches.

Fixes #6783


packages/query-core/src/queryClient.ts
```diff
@@ -204,7 +210,11 @@ export class QueryClient {
   getQueryState<TQueryFnData = unknown, TError = DefaultError>(
     queryKey: QueryKey,
   ): QueryState<TQueryFnData, TError> | undefined {
-    return this.#queryCache.find<TQueryFnData, TError>({ queryKey })?.state
+    const queryHash = hashQueryKeyByOptions(
+      queryKey,
+      this.#defaultOptions.queries,
```
TkDodo, line 215: same here, we really shouldn't be using only the default options. Maybe it should be a function on the QueryCache itself ?

TkDodo (changes requested):

packages/query-core/src/queryClient.ts
SebKranz: Ah, thank you. I wasn't clear on the difference between `defaultOptions` and `queryDefaults`. 

The updated code is compact enough that a separate method isn't necessary IMO. 

On a side-note: I'm a bit surprised that TS lets me return `unknown` as `TInferredQueryFnData` without an explicit cast.

TkDodo: > On a side-note: I'm a bit surprised that TS lets me return unknown as TInferredQueryFnData without an explicit cast.

yeah interesting - but it was the same with the previous implementation - `.find` also returns `unknown`

TkDodo (approved):

TkDodo:
sadly, the performance tests are already flaky:
- https://cloud.nx.app/runs/puVC4nNo9c?utm_source=pull-request&utm_medium=comment

```
❯ |@tanstack/query-core| src/tests/queryClient.test.tsx  (91 tests | 2 failed) 5414ms
   ❯ src/tests/queryClient.test.tsx > queryClient > setQueryData > should set 10k data in less than 500ms
     → expected 927.7827839851379 to be less than 500
   ❯ src/tests/queryClient.test.tsx > queryClient > getQueryState > should get 10k queries in less than 500ms
     → expected 603.6593210101128 to be less than 500
```

is there anything we can do here please ?

SebKranz:
I could try increasing the threshold to 2000ms and maybe go up to 20k rows. - On my machine, 10k rows without the fix took far longer (30s IIRC) so it should be possible to find a decent threshold. 

The purpose of the test is primarily to check the complexity.

SebKranz:
On second thought, I might try going down to 5k in 500ms so as to not unnecessarily slow down the pipeline. 

I'll do that now.

SebKranz:
@TkDodo I'm attempting a fix in #6937

zaidjan-devrev:
Hi @TkDodo, is there a possibility of getting this change in react query v4 too?

SebKranz:
> Hi @TkDodo, is there a possibility of getting this change in react query v4 too?

You could bypass `setQueryData` as a workaround:

```ts
// this returns the cache for this key if existing, or creates a new one with default options:
// Unlike `setQueryData`, it will use the hashtable to find the entry.
const query = queryClient.getQueryCache().build(client, { queryKey: /* ... */ }) 
query.setData(data)
```

zaidjan-devrev:
@SebKranz the problem is that our app uses react-query extensively. And doing this everywhere would take up quite some time. That coupled with the fact that this effort would become redundant when we actually move to v5 makes not want to do that.
This PR itself should be portable to `react-query-v4` right? Give the general structure of the code in v4 is similar to the original code in this. I can raise this PR, if needed. It would save a lot of work for me and team for the time being.

PR #6928, opened 2024-02-19
fix(angular-query): support required input signal

arnoud-dv:


riccardoperra:
Did a small repro 😄

https://stackblitz.com/edit/stackblitz-starters-4favya?file=src%2Fmain.ts

PR #6930, opened 2024-02-19
fix(solid-query): Don't return promise when data is available synchronously

Brendonovich:
In our application, we have `placeholderData: keepPreviousData` set globally, and try to take full advantage of Suspense and transitions.
We have refetch on window focus enabled, and every time we refocused a suspense boundary was triggering for a tiny amount of time - small enough that the fallback wasn't actually visible, but long enough that solid was modifying DOM elements, causing a CSS fade in animation to constantly retrigger.
With this change, the query resource effectively short circuits if a result is synchronously available (as is the case with `keepPreviousData` after first load), so even if a query is refetching, it won't suspend after first load.

### DIFF
--- a/packages/solid-query/src/createBaseQuery.ts
+++ b/packages/solid-query/src/createBaseQuery.ts
@@ -208,18 +208,26 @@ export function createBaseQuery<
   >(
     () => {
       const obs = observer()
-      return new Promise((resolve, reject) => {
-        if (isServer) unsubscribe = createServerSubscriber(resolve, reject)
-        else if (!unsubscribe && !isRestoring())
-          unsubscribe = createClientSubscriber()
 
-        obs.updateResult()
+      let resolve!: (d: any) => void, reject!: (r?: any) => void
 
-        if (!state.isLoading && !isRestoring()) {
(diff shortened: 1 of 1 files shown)


ardeora:
Hello! Hope you're having a great week. Thanks for fixing this! Some of the tests seem to break. Is it possible you could create a reproduction of this issue? So that I can better understand if there are some edge cases we're not handling

Brendonovich:
Made a repro: https://github.com/Brendonovich/solid-query-suspense-repro
Run it with `pnpm dev`, open the first route, then switch to another tab and back a few times, you'll notice that the CSS transition replays as DOM elements are being modified, also removing the root `Suspense` fixes the issue too.
Not sure why that test isn't passing, I'll keep looking into it.

ardeora:
Ah thank you so much for creating the reproduction! I think I have identified the issue. The issue seems to be in the implementation of the query persistent layer. I'll report back when I have figured out the solution

ardeora:
Thanks for the help on this! @Brendonovich I will be closing this in favor of #6993 and I will give you Co-Author credits there!

PR #6931, opened 2024-02-19
feat(query-devtools): Add picture-in-picture mode

ardeora:
Add picture-in-picture view to the query devtools. 

This allows the devtools to be detached from the main page and open inside a popup. The components are mounted through a portal onto the new window and all interactions work as expected

Here is a screen recording of the feature in play:

https://github.com/TanStack/query/assets/45807386/9deafda7-c3f1-4ef4-84c6-3412ff950a15


TkDodo (approved):
amazing

fh-nabeel-ahmad (commented):
LGTM

lorand-horvath:
@ardeora This is really cool! I'd just wish for the detached devtools window to not revert back to default mode as soon as I reload the page. Is that possible?

ardeora:
> @ardeora This is really cool! I'd just wish for the detached devtools window to not revert back to default mode as soon as I reload the page. Is that possible?

Hello! This is a great suggestion. I tried implementing something similar but ran into issues. I would love to hear your thoughts or suggestions if you have any feedback on how we can approach this

- We can prevent the existing popup from closing on page refresh. But how can a popup attach itself back to the reloaded window? 
- I tried re-opening the popup automatically on page refresh but most browsers require a user-triggered event to open a popup to prevent spam webpages from opening 100s of spam popups

lorand-horvath:
The state of the popup (id, size, position, etc.) could be remembered in local storage. Then the main window could read and somehow re-bind/re-attach/grab the existing popup via the stored id and re-size, re-position it. I'm not sure how can a window get a handle on an existing popup... I'll have to check.

lorand-horvath:
https://developer.mozilla.org/en-US/docs/Web/API/Window/open#target

ardeora:
Okay amazing! I think I have figured it out. Will open a PR soon

PR #6937, opened 2024-02-19
test(queryClient): Adjust performance threshholds

SebKranz:
The pipeline was found to sometimes exceed the threshold of 500ms. Since I expect the the required time to increase linearly, increase the row count by *1.5 and quadruple the time-threshhold.

## Local Tests on my Machine

Screenshot 2024-02-19 at 14 06 54

(yellow is prior to #6918, red is after, all times in ms)

### DIFF
--- a/packages/query-core/src/tests/queryClient.test.tsx
+++ b/packages/query-core/src/tests/queryClient.test.tsx
@@ -356,16 +356,6 @@ describe('queryClient', () => {
         }),
       )
     })
-
-    test('should set 5k data in less than 2000ms', () => {
-      const key = queryKey()
-      const start = performance.now()
-      for (let i = 0; i < 5000; i++) {
-        queryClient.setQueryData([key, i], i)
-      }
-      const end = performance.now()
-      expect(end - start).toBeLessThan(2000)
-    })
   })
 
   describe('setQueriesData', () => {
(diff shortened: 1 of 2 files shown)


SebKranz:
Unfortunately, I'm a bit unsure whether the best approach would be to increase or decrease the threshold & query count here.

Increasing will potentially slow down the pipeline whereas decreasing might lead to false negatives. 500ms felt like a good compromise but apparently wasn't lenient enough. (https://github.com/TanStack/query/pull/6918#issuecomment-1952314381)

@milhamm do you have opinions on this?

TkDodo:
how about a test that tests if execution doesn't grow exponentially instead ?

SebKranz:
> how about a test that tests if execution doesn't grow exponentially instead ?

I fear we would run into similar flakiness, because the cpu-load might not be constant over the course of the test (eg see the variance in my local tests). Also, the tests would become more complex since it would need some kind of regression algorithm.

I would suggest the following plan for now:

1. Try these new defaults and see if the problem occurs again
2. If it does, enable retries in vitest like so:
    ```ts
    test(
        'should set 15k data in less than 2000ms',
        () => {
             // ...
        }
        { retry: 3 },
     )
```
3. If that still fails, I'll consider an alternative testing approach.

TkDodo:
we have retries in vitest enabled

SebKranz:
I've created a PR to remove these tests: #6946

PR #6939, opened 2024-02-19
ci(eslint): add `typecheck` for cspell words

manudeli:
This change resolve below error in vitest.config.ts

image


manudeli:
@TkDodo I think this error in pr checks is not because of this Pull Request's change
image

manudeli:
@TkDodo could you check this please? this is really simple pull request for cspell words

PR #6940, opened 2024-02-19
test(react-query-devtools): add test case for not in `process.env.NODE_ENV='development'`

manudeli:
not In process.env.NODE_ENV='development', ReactDevTools of @tanstack/react-query-devtools should be return null.
so I added test case for it.


packages/react-query-devtools/src/__tests__/production.test.tsx
```diff
@@ -0,0 +1,8 @@
+import { describe, expect, it } from 'vitest'
+import { ReactQueryDevtools } from '..'
+
+describe('ReactQueryDevtools in production mode', () => {
+  it('should return null', () => {
+    expect(ReactQueryDevtools({})).toBeNull()
```
TkDodo, line 6: how does this work? I mean, how does this test run in production mode ?

manudeli: @TkDodo, Ah, my meaning is that 'not development' almost equals `process.env.NODE_ENV='production'`. but It's not perfectly equal. so I update test case not for production mode, just for not in process.env.NODE_ENV='development' exactly.

PR #6941, opened 2024-02-19
docs: Install v4 of the react devtools in the v4 docs

Fullchee:
Updating the v4 docs so that we install the v4 devtools (instead of the latest)

https://github.com/TanStack/query/discussions/6861

### DIFF
--- a/docs/framework/react/devtools.md
+++ b/docs/framework/react/devtools.md
@@ -13,14 +13,16 @@ When you begin your React Query journey, you'll want these devtools by your side
 
 ## Install and Import the Devtools
 
-The devtools are a separate package that you need to install:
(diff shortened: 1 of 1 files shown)


docs/framework/react/devtools.md
```diff
@@ -13,14 +13,16 @@ When you begin your React Query journey, you'll want these devtools by your side
 
 ## Install and Import the Devtools
 
-The devtools are a separate package that you need to install:
+The devtools are a separate package that you need to install.
+
+The major version of all the packages must be in sync.
```
Fullchee, line 18: @TkDodo 

Should we add this line to the `main` branch too?

TkDodo: sure

PR #6964, opened 2024-02-24
refactor(*): Import noop function from utils

saul-atomrigs:
I found several `noop` function declarations in various files, even though there is already one in `query-core/src/utils`!

So this PR replaces those `noop` functions with the one in `utils`. 

** Let me know if adding them in the `utils` of each directory (`svelte-query`, `react-query`, `solid-query`, `angular-query`, etc.) is a better approach.


packages/query-async-storage-persister/src/asyncThrottle.ts
```diff
@@ -1,12 +1,10 @@
+import { noop } from '../../query-core/src/utils'
```
TkDodo, line 1: I don't think we should cross-import between packages, except from their public interface if there is a dependency on that package. One `noop` util function per package should be the way to go, as we wouldn't want to export a public `noop` function from the query core.

TkDodo (changes requested):

saul-atomrigs:
Thank you for your feedback! 
I have reverted the previous commit and made changes accordingly.

packages/angular-query-experimental/src/util/index.ts
TkDodo, line 1: why do we have `src/util/index.ts` here but `src/utils.ts` in other places? That's not very consistent

saul-atomrigs: Good point. Unlike others, `/angular-query-experimental` had a `util` directory already, so I thought it would be better to use it (instead of having a duplicate `utils` file outside)

saul-atomrigs: @TkDodo Should we make a change here? Let me know how to proceed!

TkDodo: I think it's fine 👍

TkDodo (approved):

PR #6966, opened 2024-02-24
fix(angular-query): fix NG0602 error

riccardoperra:
This pr fixes the [NG0602](https://angular.io/errors/NG0602) error while using angular-query.

Since the latest update, consuming the signals inside the templates caused to log the error. 

Screenshot 2024-02-24 alle 15 13 25

Wrapping the effect inside the `untracked` function, in combination to `runInInjectionContext` at the top level (which is already present), seems resolving the issue without preventing the effect from running without registering dependencies. The effect as I understood must be executed whenever the `defaultedOptionsSignal` changes, therefore the behavior would remain unchanged.

I'm relatively new to Angular signals, I have much more experience using solid-js ones. Personally I would have solved it using nested effects but I think angular does not support them, like tracking dependencies after the first run 🤔

Also, I changed the Promise.resolve.then() using `queueMicrotask`. I think it is more suitable in this case. Is there a particular reason why Promise.resolve was used? 

PS.
While debugging using the angular query examples, I had some troubles to get the updated code after building the library. I just had to disable the ng cache in order to let it work..I think this should  be a default option for that repo examples

```json
"cli": {
  "packageManager": "pnpm",
  "analytics": false,
  "cache": {
    "enabled": false
  }
},
```

@arnoud-dv

### DIFF
--- a/packages/angular-query-experimental/src/create-base-query.ts
+++ b/packages/angular-query-experimental/src/create-base-query.ts
@@ -67,17 +67,20 @@ export function createBaseQuery<
         observer.getOptimisticResult(defaultedOptionsSignal()),
       )
 
-      effect(() => {
-        const defaultedOptions = defaultedOptionsSignal()
-        observer.setOptions(defaultedOptions, {
-          // Do not notify on updates because of changes in the options because
-          // these changes should already be reflected in the optimistic result.
-          listeners: false,
-        })
-        untracked(() => {
-          resultSignal.set(observer.getOptimisticResult(defaultedOptions))
-        })
-      })
+      // Effects should not be called inside reactive contexts
+      untracked(() =>
+        effect(() => {
+          const defaultedOptions = defaultedOptionsSignal()
+          observer.setOptions(defaultedOptions, {
+            // Do not notify on updates because of changes in the options because
+            // these changes should already be reflected in the optimistic result.
+            listeners: false,
+          })
+          untracked(() => {
+            resultSignal.set(observer.getOptimisticResult(defaultedOptions))
+          })
+        }),
+      )
 
       // observer.trackResult is not used as this optimization is not needed for Angular
       const unsubscribe = observer.subscribe(
--- a/packages/angular-query-experimental/src/lazy-init.ts
+++ b/packages/angular-query-experimental/src/lazy-init.ts
@@ -7,7 +7,7 @@ export function lazyInit<T extends object>(initializer: () => T): T {
     }
   }
 
-  queueMicrotask(() => initializeObject());
+  queueMicrotask(() => initializeObject())
 
   return new Proxy<T>({} as T, {
     get(_, prop, receiver) {
--- a/packages/angular-query-experimental/src/lazy-init.ts
+++ b/packages/angular-query-experimental/src/lazy-init.ts
@@ -7,9 +7,7 @@ export function lazyInit<T extends object>(initializer: () => T): T {
     }
   }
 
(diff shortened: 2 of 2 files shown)


arnoud-dv:
Thank you!

> Personally I would have solved it using nested effects but I think angular does not support them, like tracking dependencies after the first run 🤔

Did not look into it more than this video but [nested effects should work in Angular](https://www.youtube.com/watch?v=HstDoVQeP9g&t=9720s). Although I'm not sure this could offer a solution as an effect is scheduled too late probably. I am interested in how you would solve it.

What is your opinion on the lazy init proxy? Last week I was working a few evenings on replacing it with more computed() signals but it would require unpure computed signals and not sure if that would be better than the proxy. I would love to brainstorm about this with someone with good insight in signals.

> Also, I changed the Promise.resolve.then() using queueMicrotask. I think it is more suitable in this case. Is there a particular reason why Promise.resolve was used?

Not really, good call to use queueMicrotask instead.

>  I just had to disable the ng cache in order to let it work..I think this should be a default option for that repo examples

Agree, ran into this too.

riccardoperra:
> Did not look into it more than this video but [nested effects should work in Angular](https://www.youtube.com/watch?v=HstDoVQeP9g&t=9720s). Although I'm not sure this could offer a solution as an effect is scheduled too late probably. I am interested in how you would solve it.

In that case I usually create an effect immediately that listen for an `init` signal, then schedule another effect only once the `init` signal is valued.

Did some test yesterday but probably I was doing something wrong. Today I managed to do it using `untracked` + `injector` property

```ts
    // could be a boolean, or maybe a function to call lazily which uses signals internally
    const ready = signal<boolean>(false);
    const injector = inject(Injector);
    const initEffect = effect(() => {
      const isReady = ready();
      if (!isReady) {
        return;
      }
      // destroying the init effect in order to be sure it's immediately
      // disposed after the first time
      initEffect.destroy();
      untracked(() =>
        effect(
          () => {
            // call something that will use signals, input signals etc.
          },
          // we need to pass injector because we are outside the "constructor" phase
          { injector }
        )
      );
    });

    queueMicrotask(() => ready.set(true));
```

> What is your opinion on the lazy init proxy? Last week I was working a few evenings on replacing it with more computed() signals but it would require unpure computed signals and not sure if that would be better than the proxy. I would love to brainstorm about this with someone with good insight in signals.

Unlike other frameworks, inputs are not immediately available. Using that lazy init could be a great option, even if there are some particular cases that don't convince me.

- Using required inputs, If I try to do something with my query result inside the constructor, the error will be thrown immediately

<table>
<tr>
<td width="50%">

```ts
  postId = input.required<number>();

  postQuery = injectQuery(() => ({
    queryKey: ['post', this.postId()],
    queryFn: () => ...
  }))


  constructor() {
      console.log(this.postQuery.data());
  }
```

</td>

<td width="50%">

Screenshot 2024-02-25 alle 09 20 08

</td>
</tr>
</table>

- Using inputs (both signal or by decorator) with default value will schedule immediately the queryFn with "wrong" values. I don't know if this is a problem but the devtools is tracking the query result

image

It therefore occurs to me that the result of the query is not directly accessible from the constructor, but must necessarily be executed in a reactive context such as an effect, or after onInit. So regardless we should prevent the access in this case...after all, those who develop in Angular know that certain things cannot be done in the constructor, especially if it uses inputs

```ts
return new Proxy<T>({} as T, {
    get(_, prop, receiver) {
      if (!initialized) {
        // throw some understable error
      }
      // do something
    },
}
```

PR #6978, opened 2024-02-27
test(react-query): use vitest typecheck correctly with *.test-d.ts

manudeli:
Reference: https://vitest.dev/guide/testing-types

I convert type testing files as vitest's default *.test-d.tsx file from *.types.test.tsx

image


packages/react-query/src/__tests__/utils.tsx
```diff
@@ -78,14 +78,6 @@ export function setActTimeout(fn: () => void, ms?: number) {
...
 }
 
-export type Equal<TTargetA, TTargetB> = (<T>() => T extends TTargetA
-  ? 1
-  : 2) extends <T>() => T extends TTargetB ? 1 : 2
-  ? true
-  : false
-
-export type Expect<T extends true> = T
-
```
manudeli, line 88: This change remove unnecessary test utils

TkDodo (approved):

PR #6979, opened 2024-02-27
ci(react-query): resolve error

manudeli:


TkDodo:
I think this fails all the same on CI and I don't really know why

https://cloud.nx.app/runs/W1CWUTaekd

PR #6983, opened 2024-02-28
docs(react-query): Fix Invalid Sample Code

dev-jelly:
There is typo or something. 
Currently, 'isPending' does not work in the latest version, and this has caused some confusion.


TkDodo:
isPending should work in the latest version

PR #6993, opened 2024-02-28
fix(solid-query): Persist client implementation

ardeora:
This change fixes some loose ends with the solid query's implementation of the persist client

1. It simplifies the implementation using the new `persistQueryClientRestore` and `persistQueryClientSubscribe` functions
2. Gets rid of extra computations in `createBaseQuery`
3. Also fixes some underlying bugs mentioned in #6930 (thanks to @Brendonovich)


ardeora:
This should be fixed in v5.24.3+ Example here https://stackblitz.com/edit/github-wa8e73?file=package.json

PR #6996, opened 2024-02-29
feat(ensureQueryData): add 'revalidateIfStale' option

andredewaard:
add a prefetchWhenStale property to the fetchQuery function.
This allows the function to work like useQuery. 

- if data is fresh, return it immediately
- if data is stale, return it immediately and call prefetchQuery() in the background for the next consumer
- if there is no data in the cache (either first hit or after gcTime), await the fetch and return that

As discussed here: https://discord.com/channels/719702312431386674/1212427380912820315

Please feel free to give feedback about how to better implement this feature.


TkDodo:
Thanks for this. I thought about it for some time, and while I think there are ways to combine `ensureQueryData` / `fetchQuery` / `prefetchQuery` for a streamlined API, this would require a major version change, so I'll "defer" that decision for now.

---

I think the better API would be to add this option to `ensureQueryData` and keep `fetchQuery` as it is. `ensureQueryData` will right now:

- always return data if it exists in the cache
- if there is no data in the cache, it will fetch and wait for it, then return it.

what's missing is the piece in between: data exists, but is stale. I think we could just add a `revalidateIfStale: boolean` option here that defaults to `false` (current behaviour). If set to `true`, it would return stale data immediately, but trigger `void queryClient.prefetchQuery()` to make a background update to the cache for the next consumer.

Since your use-case is also router related, and we recommend using `ensureQueryData` in route loaders, I think this would align nicely. What do you think ?

andredewaard:
alright that also makes sense, i added it to `fetchQuery` so it would be more aligned with how `useQuery` already works and since they are both kinda the same function but `useQuery` is for components i thought it would be better to add there.

But `ensureQueryData` would be better indeed. will revert my changes and try to implement the changes.

packages/query-core/src/types.ts
```diff
@@ -407,6 +407,23 @@ export interface FetchQueryOptions<
...
+  TQueryKey extends QueryKey = QueryKey,
+  TPageParam = never,
+  > extends FetchQueryOptions<
+    TQueryFnData,
+    TError,
+    TData,
+    TQueryKey,
+    TPageParam
+  > {
+  revalidateIfStale?: boolean
```
TkDodo, line 423: going back and forth on my own proposal: since we don't use the word `revalidate` anywhere else, maybe it should be `refetchIfStale` ? Like `refetchOnMount` etc ... I think I like `revalidate` better, because it points towards it happening in the background, but I'm not sure if we should introduce new terminology 🤔

packages/query-core/src/queryClient.ts
```diff
@@ -129,12 +130,17 @@ export class QueryClient {
...
     TQueryKey extends QueryKey = QueryKey,
   >(
-    options: FetchQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+    options: EnsureQueryDataOptions<TQueryFnData, TError, TData, TQueryKey>,
   ): Promise<TData> {
-    const cachedData = this.getQueryData<TData>(options.queryKey)
+    const defaultedOptions = this.defaultQueryOptions(options)
+    const query = this.#queryCache.build(this, defaultedOptions)
+
+    if (query.state.data && query.isStaleByTime(defaultedOptions.staleTime)) {
```
TkDodo, line 138: a truthy check for `query.state.data` is not enough because `null`, `false` or `""` can all be valid values for caching a resolved query. We need to explicitly check against `undefined` please.

packages/query-core/src/queryClient.ts
```diff
@@ -129,12 +130,17 @@ export class QueryClient {
...
+      query.fetch(defaultedOptions)
+    }
 
-    return cachedData !== undefined
-      ? Promise.resolve(cachedData)
+    return query.state.data !== undefined
+      ? Promise.resolve(query.state.data)
       : this.fetchQuery(options)
   }
```
TkDodo, line 145: I think we should generally implement this a bit differently, flow-wise:

- first, look up the `cachedData`
- if it's undefined
  - return fetch
- else 
  - check if that data is stale
  - if it is - trigger a background refetch
  - in any case, return it

```
const cachedData = this.getQueryData<TData>(options.queryKey)

if (cachedData === undefined) {
  return this.fetchQuery(options)
}

const defaultedOptions = this.defaultQueryOptions(options)
if (query.isStaleByTime(defaultedOptions.staleTime)) {
  void this.prefetchQuery(defaultedOptions)
}

return cachedData
```

calling `prefetchQuery` instead of `query.fetch` is important because:

- it catches errors that happen during fetching, which is what we want for a background refetch
- it goes through `fetchQuery`, which turns off retries, which is also what we want:

https://github.com/TanStack/query/blob/bdd2b552d3647242e45c85f4c765fd293862d6d5/packages/query-core/src/queryClient.ts#L329-L331

packages/query-core/src/queryClient.ts
TkDodo: oh also - this is missing the check for `revalidateIfStale` option 😅

TkDodo:
can you also add some tests for this please ?

packages/query-core/src/types.ts
andredewaard: Yes i agree. `revalidate` is not used anywhere but it indicates clearly that is will not just invalidate the data but also revalidate it. `fetchIfStale` indicates that it will fetch the data if stale so you need to wait on it. 

Imo `revalidateIfStale` describes best what happens but i agree about the terminology not being used elsewhere, But also this isn't a function used elsewhere expect on `useQuery` but there its default behaviour.

TkDodo: > but there its default behaviour.

yes but you can turn it off with `refetchOnMount: false` for example. So maybe we should stick to `refetchIfStale` ? and then maybe rename everything to `revalidate`, so `revalidateOnMount` etc. in the next major ?

or would it make more sense to start with `revalidateIfStale` here as a new API and then streamline it in the next major. I think I like this option more

packages/query-core/src/queryClient.ts
andredewaard: A yes of course, will add that!

andredewaard:
> can you also add some tests for this please ?

Will do

andredewaard:
Test added. At first the test needs to return cached data which is 'old', we wait for the promise to resolve which will let the new call return 'new'

packages/query-core/src/queryClient.ts
```diff
@@ -130,13 +131,22 @@ export class QueryClient {
...
 
-    return cachedData !== undefined
-      ? Promise.resolve(cachedData)
-      : this.fetchQuery(options)
+    if (cachedData === undefined) return this.fetchQuery(options)
+    else {
+      const defaultedOptions = this.defaultQueryOptions(options)
+      const query = this.#queryCache.build(this, defaultedOptions)
+
+      if (query.isStaleByTime(defaultedOptions.staleTime) && options.revalidateIfStale) {
```
TkDodo, line 143: let's switch the check for an early bailout of the if statement:

```suggestion
      if (options.revalidateIfStale && query.isStaleByTime(defaultedOptions.staleTime)) {
```

TkDodo:
please fix prettier

docs/reference/QueryClient.md
```diff
@@ -189,7 +189,8 @@ const data = await queryClient.ensureQueryData({ queryKey, queryFn })
 
 **Options**
 
-- The options for `ensureQueryData` are exactly the same as those of [`fetchQuery`](#queryclientfetchquery) except that `queryKey` is required.
+- The options for `ensureQueryData` are almost the same as those of [`fetchQuery`](#queryclientfetchquery) except that `queryKey` is required and that is has an extra optional option `revalidateIfStale`. When set `revalidateIfStale` to true the function will make sure the data will refetch in the background while still returning the data from the cache.
```
TkDodo, line 192:
```suggestion
- the same options as [`fetchQuery`](#queryclientfetchquery)
- `revalidateIfStale: boolean`
  - Optional
  - Defaults to `false`
  - If set to `true`, stale data will be refetched in the background, but cached data will be returned immediately.
```

TkDodo:
looks good on unpkg:

https://unpkg.com/browse/@tanstack/query-core@5.28.0/build/modern/queryClient-UWB6EkqM.d.ts

TkDodo:
ooh, that's because `vue` has it's own wrapper around the core because of this `unref` things, and we didn't update the types there:

https://github.com/TanStack/query/blob/61f4d747883ce67cf4edcc65f9a7693fa2bc7b6a/packages/vue-query/src/queryClient.ts#L87-L98

can you do another PR to fix this please ?

andredewaard:
A alright, yes will do.

TkDodo:
@andredewaard I wanted to reach out to you because I wrote an RFC about unifying our imperative query methods, and as it stands, the `revalidateIfStale` feature might not make it to the new API in that proposal.

Please have a look and let me know what you think:

- https://github.com/TanStack/query/discussions/9135

andredewaard:
@TkDodo Thanks for the heads up!
I read the article the day it was live and it totally makes sense.
Right now it isn't really clear what every api is doing exactly and what the edge cases are. 
Like why I asked for the change in the first place, since I really thought that was going to happen.

While I do think the fetchQuery api should behave just like useQuery, revalidate data when stale, I can understand you won't add it as an extra option.

So what could be considered is making the fetchQuery and useQuery consistent with each other and have the above change, which does require a major change.

PR #6999, opened 2024-02-29
feat(react-query): add typesafe query disabling via skipToken

Jaaneek:
This pull request is heavily inspired by RTK query skipToken.

https://redux-toolkit.js.org/rtk-query/usage/conditional-fetching#overview

We are introducing skipToken which allows typesafe disabling of queries.

```tsx
  const { data } = useQuery({
      queryKey: ['todos', filter],
      // ⬇️ disabled as long as the filter is undefined or empty
      queryFn: filter ? () => fetchTodos(filter) : skipToken,
  })
  ```

### CODE REVIEW

File: packages/query-core/src/utils.ts [hunk 1]
```diff
@@ -345,3 +346,8 @@ export function addToStart<T>(items: Array<T>, item: T, max = 0): Array<T> {
   const newItems = [item, ...items]
   return max && newItems.length > max ? newItems.slice(0, -1) : newItems
 }
+
+export const skipToken = Symbol.for(
+  'QUERY/skipToken-ea7c3132-6b5b-441e-89d8-4a98dcb5eb23',
+)
```
Jaaneek, line 352: added UUID to make extra sure we avoid conflicts.

packages/query-core/src/types.ts
```diff
@@ -32,7 +32,9 @@ export type QueryFunction<
   T = unknown,
   TQueryKey extends QueryKey = QueryKey,
   TPageParam = never,
-> = (context: QueryFunctionContext<TQueryKey, TPageParam>) => T | Promise<T>
+> =
+  | ((context: QueryFunctionContext<TQueryKey, TPageParam>) => T | Promise<T>)
+  | SkipToken
```
Jaaneek, line 37: I can either leave it in here and handle it the same way the "options.enabled" is handled, or I can move it into react-query

packages/query-core/src/queryObserver.ts
```diff
@@ -683,6 +687,7 @@ function shouldLoadOnMount(
 ): boolean {
   return (
     options.enabled !== false &&
+    options.queryFn !== skipToken &&
```
Jaaneek, line 690: Alternate solution is to assign options.queryFn !== skipToken to options.enabled.
Whenever we initiate or update the observer we would 
```
options.enabled = options.queryFn !== skipToken
```

Downside of this is that we will mutate the options.enabled and if someone is checking it (nobody should do it) it might give him a value that he didn't assign. 
So it's potentially not backwards compatible

packages/query-core/src/types.ts
Jaaneek: This breaks packages/query-core/src/infiniteQueryObserver.ts
can you take a look @TkDodo <3

packages/query-core/src/utils.ts
HugeLetters: Shouldn't this just be `Symbol('QUERY/skipToken')` - then you don't have to worry about collisions at all. Since you have to use this variable to check for equality anyway.

Jaaneek: Im using symbol.for in here which might return already existing Symbol.

I prefer to use symbol.for because some other developers might want to not import the symbol itself, just get it from the "global registry of symbols". Very rare case but might happen.

image

packages/query-core/src/types.ts
```diff
@@ -147,12 +147,13 @@ export interface QueryOptions<
    * Setting it to `Infinity` will disable garbage collection.
    */
   gcTime?: number
-  queryFn?: QueryFunction<TQueryFnData, TQueryKey, TPageParam>
+  queryFn?: QueryFunction<TQueryFnData, TQueryKey, TPageParam> | SkipToken
```
Jaaneek, line 150: We can either keep it in here or only implement the logic in react package.

I'm okay with both options.

If we implement this only in react package then other packages might implement it in very similar way and we won't add anything to query-core.

packages/query-core/src/queryObserver.ts
TkDodo: I think we should centralize the check in the `defaultQueryOptions` function. There, we already set options depending on other options. It would be fine to set `enabled: false` if the `queryFn` is a `skipToken`

packages/query-core/src/query.ts
```diff
@@ -392,6 +395,11 @@ export class Query<
           new Error(`Missing queryFn: '${this.options.queryHash}'`),
         )
       }
+
+      if (this.options.queryFn === skipToken) {
+        return Promise.resolve(undefined)
+      }
```
TkDodo, line 401: this isn't ideal because a resolved value can and should never be `undefined`. Instead, I guess we should set the `queryFn` itself back to `undefined`, or to a function that throws (maybe in dev mode with a descriptive message?), in `defaultQueryOptions` when we set `enabled: false`.

Usually, the queryFn won't run, but if you pass a `skipToken` and then call `refetch` manually - it would, but that's kind of an invariant

TkDodo (commented):
two things to keep an eye on:

- it shouldn't be possible to pass a `skipToken` to suspense versions of the hooks, like `useSuspenseQuery`, because those always set `enabled: true`. We should likely forbid this on type level, and at runtime, only set `enabled: false` if we have a skipToken and enabled isn't already true
- we should watch out how this influences places where we try to infer the type of what the `queryFn` returns, because with `skipToken`, it might not be a function. `queryOptions` is one of those places, I think we should add a type-test for that.

packages/query-core/src/query.ts
Jaaneek: Sounds good, on it

Jaaneek: @TkDodo Are you sure we should set the `queryFn` in `defaultQueryOptions`. I think it should be kept as skipToken in runtime so it will be "easier" to debug and might be also nicer to display in dev tools in future.

If we override it we gain nothing but lose the information

I would prefer to keep the check and instead of 


```suggestion
// Rejects with an error if `queryFn` is set to `skipToken`, indicating that this query should not be executed.
if (this.options.queryFn === skipToken) {
  return Promise.reject(new Error(`Attempted to invoke queryFn when set to skipToken, indicating a configuration error. Query hash: '${this.options.queryHash}'`));
}
```


But I'm open to changing it

Jaaneek:
> two things to keep an eye on:
> 
> * it shouldn't be possible to pass a `skipToken` to suspense versions of the hooks, like `useSuspenseQuery`, because those always set `enabled: true`. We should likely forbid this on type level, and at runtime, only set `enabled: false` if we have a skipToken and enabled isn't already true
> * we should watch out how this influences places where we try to infer the type of what the `queryFn` returns, because with `skipToken`, it might not be a function. `queryOptions` is one of those places, I think we should add a type-test for that.

Is it only for react-query package or should this be implemented in other implementations of query as well? Once we accept queryFn as skipToken it will work like that everywhere

packages/query-core/src/queryObserver.ts
Jaaneek: Done!

TkDodo (commented):
regarding `queryOptions` - does this still work?

```
const options = queryOptions({
  queryKey: ['key'],
  queryFn: () => Math.random() > 0.5 ? skipToken : Promise.resolve(5),
})

const data = queryClient.getQueryData(options.queryKey)
```

here, `data` should be of type `number | undefined` - inferred from the `queryFn`. I just want to make sure that the `skipToken` existence doesn't destroy that.

Jaaneek:
> regarding `queryOptions` - does this still work?
> 
> ```
> const options = queryOptions({
>   queryKey: ['key'],
>   queryFn: () => Math.random() > 0.5 ? skipToken : Promise.resolve(5),
> })
> 
> const data = queryClient.getQueryData(options.queryKey)
> ```
> 
> here, `data` should be of type `number | undefined` - inferred from the `queryFn`. I just want to make sure that the `skipToken` existence doesn't destroy that.

This works:

image

Should we also handle returning the skipToken from a function? I think it's not "needed"?

TkDodo:
Oh Great 👍 
No I also don't think this is needed

packages/query-core/src/queryClient.ts
```diff
@@ -526,6 +527,10 @@ export class QueryClient {
       defaultedOptions.networkMode = 'offlineFirst'
     }
 
+    if (defaultedOptions.queryFn === skipToken) {
+      defaultedOptions.enabled = false
+    }
```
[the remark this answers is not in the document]
Jaaneek: Nice, I agree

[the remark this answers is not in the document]
TkDodo: i think the implementation is fine as-is. If you pass `enabled: true`, it will bypass `skipToken` and then error out when the queryFn is executed. I think it doesn't make sense to set both, and this behaviour is okay.

packages/react-query/src/__tests__/suspense.test-d.tsx
```diff
@@ -121,4 +123,16 @@ describe('useSuspenseInfiniteQuery', () => {
...
+    doNotExecute(() => {
+      const query: UseSuspenseQueryOptions = {
+        // @ts-expect-error
+        queryFn: skipToken,
+        queryKey: [1],
+      }
+
+      return query
+    })
+  })
```
manudeli, line 137: *.test-d.tsx file don't be excuted by vitest type test. it just type check by default
https://vitest.dev/guide/testing-types

```suggestion
  it('should not accept skipToken type for queryFn in useSuspenseQuery', () => {
    const query: UseSuspenseQueryOptions = {
      // @ts-expect-error
      queryFn: skipToken,
      queryKey: [1],
    }

    return query
  })
```

Jaaneek: Agreed, I will remove it! thx

TkDodo:
the `should respect skipToken` test is constantly failing in CI, but it works fine locally. I don't know why :/

packages/react-query/src/__tests__/useQuery.test.tsx
```diff
@@ -6334,7 +6334,7 @@ describe('useQuery', () => {
   it('should respect skipToken and refetch when skipToken is taken away', async () => {
     function Page({ enabled }: { enabled: boolean }) {
       const { data, status } = useQuery({
-        queryKey: ['key'],
+        queryKey: ['key-skip-token'],
```
TkDodo, line 6337: ah I missed this. If this works, please use the provided helper function to create unique keys:

https://github.com/TanStack/query/blob/5e648737524e18121213be243f791721c309a4b4/packages/react-query/src/__tests__/useQueries.test.tsx#L843

TkDodo:
> Maybe we should do cleanup after every test to ensure that the queryKeys are not shared? I don't think there is a use in sharing them

yeah that's what the `queryKey()` function is for. it increments a value. no test should use hardcoded keys

hichemfantar:
I can't find the documentation for the skipToken.
I only discovered this thanks to [TRPC](https://trpc.io/docs/client/react/disabling-queries)

Jaaneek:
> I can't find the documentation for skipToken.
> 
> I only discovered this thanks to [TRPC](https://trpc.io/docs/client/react/disabling-queries)
> 
> image
> 
> 

https://tanstack.com/query/latest/docs/framework/react/guides/disabling-queries

hichemfantar:
@Jaaneek thanks! any idea as to why skipToken isn't search indexed?

tychenjiajun:
I found following code works better for me:

```ts
  const { data } = useQuery({
      queryKey: ['todos', filter],
      // ⬇️ avoid network fetch
      queryFn: async () => filter ? await fetchTodos(filter) : [],
  })
```

We don't have to worry about unintentional calling `useSuspenseQuery` and/or `refetch`

PR #7005, opened 2024-03-01
fix(devtools): add shadowDOMTarget property to react-query-devtool

michelleoj:
This feature stems from this [conversation](https://github.com/TanStack/query/discussions/6070) with @ardeora

The original issue was that if you rendered the devtools in an component that lives in the shadow dom, the styles would be appended to regular DOM's head tag. Since the shadow DOM encapsulates everything, the devtools shows up unstyled. The fix to this is to pass in the shadow root as a target so that goober library will append the styles to the ShadowRoot element. 
I've gone ahead and added an example and updated the docs. Let me know if there's anything else I'm missing


michelleoj:
@TkDodo hey there 👋🏾 
I would like some eyes on this when you get the chance 😄 🙏🏾

michelleoj:
@ardeora any chance I can get some eyes on this soon please?

ardeora:
Apologies for the delay here. I'll be taking a look at this in an hour or two!

packages/query-devtools/src/Devtools.tsx
```diff
@@ -3397,5 +3436,5 @@ const stylesFactory = (theme: 'light' | 'dark') => {
   }
 }
 
-const lightStyles = stylesFactory('light')
-const darkStyles = stylesFactory('dark')
+const lightStyles = (css) => stylesFactory('light', css)
+const darkStyles = (css) => stylesFactory('dark', css)
```
ardeora, line 3440:
```suggestion
const lightStyles = (css: typeof goober['css']) => stylesFactory('light', css)
const darkStyles = (css: (typeof goober)['css']) => stylesFactory('dark', css)
```

Types are erroring could you fix the types like so

packages/query-devtools/src/Devtools.tsx
```diff
@@ -2470,7 +2509,7 @@ const createSubscribeToMutationCacheBatcher = <T,>(
   return value
 }
 
-const stylesFactory = (theme: 'light' | 'dark') => {
+const stylesFactory = (theme: 'light' | 'dark', css) => {
```
ardeora, line 2512:
```suggestion
const stylesFactory = (theme: 'light' | 'dark', css: typeof goober['css']) => {
```

Similarly over here

michelleoj:
@ardeora added your suggestions and now all tests are passing ✅

ardeora (changes requested):
These are looking perfect! Thank you for working on this. One other small addition (last one I promise) would be if we can make this feature available in other framework devtools also. You don't need to know Solid, Svelte or Angular. We already have optional props like `styleNonce` in other framework devtools adapter. We can follow the same convention to add an optional shadowDomTarget prop to all other devtools too. Once that is done, I think we are good to merge.

michelleoj:
@ardeora just added your suggestions and it's ready for another review! 🙏🏾

ardeora (approved):

ardeora:
@michelleoj Thank you so much for working on this! Everything looks great and should be included in the next minor version!

michelleoj:
@ardeora thanks so much for taking the time to review my PR! this is my first time contributing to an open source project so thanks for helping me through it 🎉 ❤️

PR #7006, opened 2024-03-02
refactor: switch typeof undefined checks with strict comparison of the value undefined

TkDodo:


packages/query-core/src/queryClient.ts
```diff
@@ -511,18 +511,15 @@ export class QueryClient {
...
-    if (typeof defaultedOptions.throwOnError === 'undefined') {
+    if (defaultedOptions.throwOnError === undefined) {
       defaultedOptions.throwOnError = !!defaultedOptions.suspense
     }
 
-    if (
-      typeof defaultedOptions.networkMode === 'undefined' &&
-      defaultedOptions.persister
-    ) {
+    if (!defaultedOptions.networkMode && defaultedOptions.persister) {
```
TkDodo, line 522: `networkMode` is a string literal type so doing the simplified `!` check should be safe. We can't do that for the above booleans obviously

packages/query-core/src/utils.ts
```diff
@@ -122,10 +122,7 @@ export function matchQuery(
     return false
   }
 
-  if (
-    typeof fetchStatus !== 'undefined' &&
-    fetchStatus !== query.state.fetchStatus
-  ) {
+  if (fetchStatus && fetchStatus !== query.state.fetchStatus) {
```
TkDodo, line 125: same here: `fetchStatus` is not a boolean, we just need to check if it was passed

SergioMorchon:
It may be handy to use a more semantic utility with a positive naming convention.
Something like
```typescript
const isDefined = <Value>(value: Value | null | undefined): value is Value =>
  value !== null && value !== undefined;
```

TkDodo:
We can't check for null because data can be null - it just can't be undefined

PR #7016, opened 2024-03-03
feat(angular-query): support required input signal on injectMutation

riccardoperra:
This pr add required input signal support for `injectMutation` and `injectMutationState`.

Like `injectQuery` does, we should support required input signals also for `injectMutation` and `injectMutationState`. The use may be related to some specific mutations that can have as `mutationKey` the value of a required input signal. Done that, we may want to get the state of the mutations via `injectMutationState`, which could have as "filter" that value passed as input

```ts
// Component A

class ComponentA { 
  readonly id = input.required<string>();
  readonly addMutation = injectMutation(() => ({
    mutationKey: ['entity', this.id(), 'add'],
    mutationFn: () => {}
  })
} 

// Component B

class ComponentB { 
  readonly id = input.required<string>();
  readonly mutationState = injectMutationState(() => ({
    filters: {mutationKey: ['entity', this.id()]}
  })
}
```

--- 

I'm adding some new tests via TestBed in both injectMutation/injectQuery in order to be sure that nothings break if we find another solution for the lazy initialization.

I'm also introducing some utils in order to set required signal input via testBed, which seems currently not "officially" supported. I found a related [discussion](https://github.com/angular/angular/issues/54013) but vitest seems not supporting that yet, so I couldn't use testing-library/angular.

- [ ] Add support for injectIsMutating --> This may should be done in another PR. It's a breaking change that requires to update the function signature

@arnoud-dv


arnoud-dv:
Nice! Thank you for the additional tests too.

PR #6943, opened 2024-02-19
React Query Dev Tools for React Native readme update

LovesWorking:
Add a link to the newly created react query dev tools made for react native https://github.com/LovesWorking/react-native-react-query-devtools

PR #7033, opened 2024-03-06
Test Nx cache when only docs are changed

lachlancollins:


lachlancollins:
@TkDodo in this PR, only `test:format`, `test:sherif` and `test:knip` missed cache when a doc was changed. It might be because `nx affected` was comparing to an earlier main commit in #6943 ?

PR #7063, opened 2024-03-09
chore(vue-query): add vite.config.ts as tsconfig's include field

manudeli:
simply resolve this error

image


packages/vue-query/vite.config.ts
```diff
@@ -16,6 +16,7 @@ export default defineConfig({
       if (log.includes('Download the Vue Devtools extension')) {
         return false
       }
+      return undefined
```
manudeli, line 19: To resolve this error

image

PR #7076, opened 2024-03-11
feat: pass context to mutationFn

momme-rtf:
Since the onMutate function offers a place for preprocessing I'd find it helpful if not only the onError, onSuccess , ... callbacks receive the context (returned by the onMutate) as an input parameter but also the mutationFn.

[Discussion](https://github.com/TanStack/query/discussions/7045)


TkDodo:
if we think this through, when we should build something similar to the `QueryFunctionContext` - an object that we pass as second parameter. The return value from `onMutate` would just be one thing that goes there, `meta` could be another (same as useQuery) and someday maybe even an AbortSignal.

Naming will be weird - I would like this object to be the `MutationFunctionContext`, but then we'd need another word than "context" for what the `onMutate` function returns 🤔

momme-rtf:
Hmm interesting.. What if the return value of onMutate would just contribute to the meta prop instead of being a separate property in the `MutationFunctionContext`?

TkDodo:
not sure if that would be better. To be on par with queries, mets should be what the user defines meta to be

PR #7096, opened 2024-03-12
fix(query-devtools): Preserve offline status after minimize

### LINKED ISSUE #7094: minimizing devtools loses offline toggle state
### Describe the bug

When clicking the offline toggle, it correctly goes into "offline" mode. However, minimizing + maximizing it shows the toggle in "online" mode, while in fact, the onlineManager is still offline.

https://tanstack.com/query/v5/docs/framework/react/examples/basic

### Steps to reproduce

- open the devtools in the basic example
- click the offline toggle
- navigate around, you'll see some paused queries
- minimize the offline toggle
- navigate some more, everything is still offline
- open the devtools again

--> the toggle now says we're online, but we're not.

### Expected behavior

I think minimizing the devtools unmounts them, so the internal state is lost. I see two possible solutions:

1) re-set the state of the onlineManager when the devtools unmount
2) keep the state of the offline toggle while the devtools are unmounted (e.g. by lifting the state up to the component that is mounted - the tanstack logo container component). I wouldn't put it in localstorage because that would persist it over page reloads, which is unexpected.

None

### Screenshots or Videos

_No response_

### Platform

all

### Tanstack Query adapter

None

### TanStack Query version

v5

### TypeScript version

_No response_

### Additional context

_No response_

ardeora:
Fixes #7094

This moves the offline detector signal higher up in the component tree to preserve offline status if it was toggled on before minimizing or before opening the devtools in pip mode

### DIFF
--- a/packages/query-devtools/src/Devtools.tsx
+++ b/packages/query-devtools/src/Devtools.tsx
@@ -110,6 +110,7 @@ const [selectedMutationId, setSelectedMutationId] = createSignal<number | null>(
   null,
 )
 const [panelWidth, setPanelWidth] = createSignal(0)
+const [offline, setOffline] = createSignal(false)
 
 export type DevtoolsComponentType = Component<QueryDevtoolsProps> & {
   shadowDOMTarget?: ShadowRoot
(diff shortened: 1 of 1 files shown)


TkDodo:
Thank you 🙏

Issue #7098, opened 2024-03-13
react-query-devtools: restore loading infinite wait


ByungyeonKim:
### Describe the bug

When clicking the restore loading button with a combination of react-query-devtools and json-server, a bug occurs causing an infinite wait.

However, downgrading to version 4 resolves the issue and it works normally.

### Your minimal, reproducible example

n/a

### Steps to reproduce

n/a

### Expected behavior

In version ^4.36.1, the Promise object is fulfilled, but in version ^5.26.3, it remains in a pending state.

### How often does this bug happen?

Every time

### Screenshots or Videos

| v^4.36.1 | v^5.26.3 |
|----------|----------|
| <img width="1016" alt="스크린샷 2024-03-13 오후 3 21 27" src="https://github.com/TanStack/query/assets/66554164/7f47403c-94fa-4ce8-b2c0-2816e59e13c7"> |  <img width="960" alt="스크린샷 2024-03-13 오후 3 20 39" src="https://github.com/TanStack/query/assets/66554164/f37c7492-1a12-4b0d-a681-966d380dd55c">  |

### Platform

- macOS
- browser: chrome(122.0.6261.112)

### Tanstack Query adapter

None

### TanStack Query version

v5.26.3

### TypeScript version

v5.3.3

### Additional context

n/a

bastibuck:
I can confirm. Also check https://stackblitz.com/github/tanstack/query/tree/main/examples/react/basic?embed=1&theme=dark&preset=node example. 

Not related to json-server.
Also it used to work in a previous version 5. So nothing that broke in v5 IMO

ByungyeonKim:
@bastibuck Thank you for your confirmation. Thanks to you, I realized that it has nothing to do with json-server. I started by comparing the v4 version with the current one, and finding the differences turned out to be the solution.

You can check the fix in the link below. Thank you.

- https://github.com/TanStack/query/pull/7099

TkDodo:
@ByungyeonKim do you want to give fixing this another try? Otherwise I'll ping Aryan, who wrote the devtools.

ByungyeonKim:
@TkDodo Thank you for giving me another chance. If you could give me time till the first week of April, I'll try to fix it and let you know. Thanks again for your consideration.

ByungyeonKim:
@ardeora I just updated to the latest version and confirmed it is working properly. Thank you!! 👍

PR #7099, opened 2024-03-13
fix(query-devtools): Change activeQuery state

ByungyeonKim:
Fixes https://github.com/TanStack/query/issues/7098

Solving the bug where clicking the restore loading button with a combination of react-query-devtools and json-server in versions above v5 results in an infinite wait.


bastibuck:
Not sure if this fixes it. 
At least the react-demo deployed to CodeSandbox (from one of the above links) is still broken.

ByungyeonKim:
@bastibuck

I didn't know anything about this as it was my first time doing it. thank you for telling me. I'll check!

PR #7114, opened 2024-03-15
fix(core): optimistic results reflect correct state

### LINKED ISSUE #7105: Typescript incorrectly narrows `error` based on `isLoading`
### Describe the bug

If I do 
```ts
const { error, isLoading } = useQuery(...);
if (isLoading) {
  ....
```
`error` is assumed to be `null` by typescript. That's wrong: if it's the first time the query is used, then yes, that's right. But if:
1. query retries until we have `isLoading === false` and `error !== null`
2. the component using the query is unmounted
3. it gets remounted
4. the query will be automatically retried (resulting in `isLoading === true` again), but `error` will be initialized with the previous `error`, so we have `error !== null` while also `isLoading === true`.
 
So either the types or the implementation are wrong. I'm hoping for the types being wrong, because I like this behavior 😅 

In case it helps: the same implementation behavior is present with `isFetching` (so, you can have  `isFetching === true` while also having `error !== null`), but the type narrowing doesn't kick in that case. 

[Ts Playground](https://www.typescriptlang.org/play?ts=5.3.3#code/JYWwDg9gTgLgBAbzgVwM4FMCKz1QJ5wC+cAZlBCHAOQACMAhgHaoMDGA1gPRTr2swBaAI458VANwAoSQBN0rADb0epZI37AIjUuhisAFgGUKu-cEYBzABQBKAFxwACuRDAMAHgQAPB42QgAI1xCAD4pSRI1DS04ADEICFtESThUuFYtFkQ4NwAZCHoZcwsAGhzUWN0DYrLccigiOABeFAxsXDwrBBEOgGl0PAcAbSoQPAF2AaoAXTKe-FjGBxIqoxMYM0tCG3C0nJI4KzyCossbZL29uugpS7TOTjgAPQB+FLTCaT3gA6OK1eK5wQ7zu1ygtzucAezzee0+nyAA)

### Steps to reproduce

Since I'm hoping for this to be a bug in the types and not the implementation, I'm providing a Ts Playground as MWE. But if you folks tell me this is an implementation bug, I can create a codesandbox instead.

### Expected behavior

Type narrowing match implementation

Every time

### Screenshots or Videos

_No response_

### Platform

-

### Tanstack Query adapter

None

### TanStack Query version

5.25.0

### TypeScript version

5.3.3

### Additional context

_No response_

TkDodo:
when computing the optimistic result, we need to set / reset all properties that the `fetch` action in the query reducer will also set. This fix extracts the logic into a function and calls it in both places

fixes #7105

### DIFF
--- a/packages/query-core/src/query.ts
+++ b/packages/query-core/src/query.ts
@@ -552,16 +552,8 @@ export class Query<
         case 'fetch':
           return {
             ...state,
-            fetchFailureCount: 0,
-            fetchFailureReason: null,
+            ...fetchState(state.data, this.options),
             fetchMeta: action.meta ?? null,
-            fetchStatus: canFetch(this.options.networkMode)
-              ? 'fetching'
-              : 'paused',
-            ...(state.data === undefined && {
-              error: null,
-              status: 'pending',
-            }),
           }
         case 'success':
           return {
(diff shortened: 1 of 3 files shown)


TkDodo:
@lgenzelis fyi

PR #7117, opened 2024-03-15
ci(*): update codecov action version of GitHub workflows

manudeli:
Recently codecov/codecov-action v4 was released. so I updated its version
https://github.com/codecov/codecov-action/releases


TkDodo:
I'm not sure if codecov is really working for our repo anymore, but we can give it a try

PR #7119, opened 2024-03-15
ci(*): add script `attw --pack` for all packages

manudeli:
https://tsup.egoist.dev/#generate-declaration-file

image

I added [@arethetypeswrong/cli](https://www.npmjs.com/package/@arethetypeswrong/cli) to check if the library users would work well with TypeScript. This is also recommended in the tsup official document. This change will make we check in every CI. `ignoreRules` of @arethetypeswrong/cli have also been added in some packages. I think this could be gradually removed if necessary.

SWR also use this tool (https://github.com/vercel/swr/blob/main/package.json#L103)


packages/angular-query-devtools-experimental/.attw.json
```diff
@@ -0,0 +1,3 @@
+{
+  "ignoreRules": ["cjs-resolves-to-esm", "internal-resolution-error"]
```
TkDodo, line 2: ❓ question: should we not fix those errors rather than ignore them ?

TkDodo: @arnoud-dv FYI

arnoud-dv: `cjs-resolves-to-esm` doesn't need to be fixed as angular-query is ESM only, from the arethetypeswrong docs:

> ## Common causes
>
> This usually happens when a library only contains ESM after making a conscious decision not to support CommonJS. (This tool tries to be neutral on that decision, but by default shows what happens in many module resolution scenarios. The author of this tool encourages library authors to consider their constraints, understand their users, and move in the direction of ESM-only when possible.)

Also looked into the `internal-resolution-error` and I think this can be fixed by flattening the emitted `index.d.ts` file so no relative imports are needed. I don't think it causes problems but I will put this on my todo list.

manudeli: > ❓ question: should we not fix those errors rather than ignore them ?

> cjs-resolves-to-esm doesn't need to be fixed as angular-query is ESM only, from the arethetypeswrong docs:

> Also looked into the internal-resolution-error and I think this can be fixed by flattening the emitted index.d.ts file so no relative imports are needed. I don't think it causes problems but I will put this on my todo list.

@TkDodo @arnoud-dv Then could we first apply ignoreRules, and if we really need, remove this ignoreRules(internal-resolution-error) with fix this errors again at that time?

arnoud-dv: @manudeli Yes I agree.

Kamahl19: @arnoud-dv have you inspected the "internal-resolution-error" further? I am getting the error when using the tanstack config as well .

arnoud-dv: Yes, was able to fix that error by rolling up type definitions using Microsoft API extractor. See https://github.com/TanStack/query/pull/7380

Kamahl19: @arnoud-dv  thanks! Is this something that should also be implemented in the tanstack/config project? @lachlancollins

arnoud-dv: I am not sure, didn't look much into the config project yet.

In general it's good practice for libs to rollup type declarations as it prevents IDEs from suggesting deep imports and especially when using Microsoft API extractor it allows better control / oversight over the public API.

PR #7123, opened 2024-03-16
ci(*): add husky + lint-staged + prettier to remove chance to meet test:format error

manudeli:
When I make Pull Request, Pretty frequently I encounter test:format error in ci.

So I think we should remove chance to meet this error by like below automatically by using husky + lint-staged + prettier

Mar-16-2024 14-58-51

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -6,3 +6,4 @@
 **/dist
 **/query-codemods/**/__testfixtures__
 pnpm-lock.yaml
+packages/**/tsup.config.bundled*.mjs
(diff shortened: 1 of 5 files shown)


manudeli:
image

.prettierignore
```diff
@@ -6,3 +6,4 @@
 **/dist
 **/query-codemods/**/__testfixtures__
 pnpm-lock.yaml
+packages/**/tsup.config.bundled*.mjs
```
manudeli, line 9: because of this error

image

TkDodo (approved):
thanks. I've seen this in PRs lately and it drives me crazy. Why do people not run prettier in their editors locally on save. Beats me 🤷‍♂️

Issue #7118, opened 2024-03-15
useSuspenseQueries inference differ from useQueries inference


wk-ouisense:
### Describe the bug

Link to TypeScript playground - [playground](https://www.typescriptlang.org/fr/play?#code/JYWwDg9gTgLgBAbzgVwM4FMCKz1WO1AGjgEccoBPAeTBmAgDsiUMBlNMdJrc-VOAL5wAZlAgg4AIgACMAIZN5AYwDWAeijo5SmAFoyuCpIBQxpY1TxG6OAF5S5arXpMAFAmNwHhgNLoKAFxwANqS1pIAuoSe3pQAYgxBcqgUDEpwrgCUdgB8iDFemjDIUAxS4XDJcOaKANwxAtECmfVmFvAwAO4QdrFOdBbuMQaUfoEhkl0QkdFeIxQJSSlpGdm2eR5ehejFpVJTkpX8NZb1Xo3Gza0n8MHWmMRTmBG9aDy4fO4jfEHBMdazOBTYwRZpwNRqODABjCXCaAAmcF0R2q7TgjCBAAsbHIoFA5BQ4PCIAQ4AwIPAlJiFABzHEMCgwTHQmmmE4QAA26AAdByIDTXPdufC5PJHt1MMLRXJMmy0XcGOh2MwpsqXvY3srONxsB8CF9eARfv9FYDgaDshCUAwVOTOmVkVUbuiykycXiCUSSfxyZTqQw6ZUGUyWXKmJyeXyBdZlVKxUDurGRfJZcYgA)

In the playground, you can see that `useSuspenseQueries` data are not inferred and becomes `unknown`. `useQueries` is working great

### Your minimal, reproducible example

[...]

PR #7141, opened 2024-03-20
fix(react-query): fix useSuspenseQueries broken result type (#7118)

xdjoshuaaz:
The type inference inside `GetSuspenseResults<T>` is missing `SkipToken` as a possible return type for `queryFn`. 

This causes inference of the returned results of `useSuspenseQueries` to not work, defaulting to a `TData` of `unknown`.

This PR just quickly adds `SkipToken` in the remainder of the code to allow inference to work again.

Fixes issue #7118.

### DIFF
--- a/packages/react-query/src/useSuspenseQueries.ts
+++ b/packages/react-query/src/useSuspenseQueries.ts
@@ -78,7 +78,7 @@ type GetSuspenseResults<T> =
               ? UseSuspenseQueryResult<TQueryFnData>
               : // Part 3: responsible for mapping inferred type to results, if no explicit parameter was provided
                 T extends {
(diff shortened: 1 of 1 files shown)


TkDodo:
Thanks. Can you please:
- add a type test so that we make sure this doesn't break again in the future. `useQueries.test-d.tsx` should be the right place
- can you please add the fix also to other adapters? The types for `useQueries` are pretty much copy-pasted to all adapters (yes, this can be simplified ... somehow, somewhen)

andykog:
@TkDodo this PR is not active for 8 days, so I've opened another PR with tests: https://github.com/TanStack/query/pull/7194
I believe this fix is not relevant for other adapters because they don't have useSuspenseQueries and createQueries don't have this issue

Issue #7035, opened 2024-03-06
`useQueries` + `skipToken` looses `.data` type inference


vjee:
### Describe the bug

When using `useQueries` with `skipToken`, `queryResult.data` is inferred as `unknown` instead of the actual return type of the `queryFn`.

```ts
const [withEnabled] = useQueries({
  queries: [{ queryKey: ["withEnabled"], queryFn: () => 1, enabled: true }],
});

const [withSkipToken] = useQueries({
  queries: [
    { queryKey: ["withSkipToken"], queryFn: true ? () => 1 : skipToken },
  ],
});

withEnabled.data;
//           ^? number | undefined
withSkipToken.data;
//             ^? unknown
```

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/tanstack-react-query-usequeries-skiptoken-issue-wdk8m5

### Steps to reproduce

- Open repro
- Hover over `withEnabled.data` and `withSkipToken.data` to see the types

### Expected behavior

I expected `withSkipToken.data` to be inferred as `number | undefined` instead of `unknown`.

### How often does this bug happen?

None

### Platform

@tanstack/react-query v5.25.0

### Tanstack Query adapter

react-query

### TanStack Query version

v5.25.0

### TypeScript version

v5.3.3

TkDodo:
@Jaaneek can you take a look here? I'm afraid I'm no expert with the `userQueries` typings 😅

Jaaneek:
Looking into it!

Jaaneek:
@vjee quick fix is to add "as const" after queries array.

```tsx
 queries: [
    { queryKey: ["withSkipToken"], queryFn: true ? () => 1 : skipToken },
  ] as const
  ```
  
 For some reason typescript doesnt make the array "readonly" by default when passing skipToken @TkDodo 
 
 I'm looking for a solution

**Comment by vjee** [comment]
Thanks for the quick fix. Great feature btw! Well done 👏

**Comment by Jaaneek** [comment]
I found a fix, pull request incoming

**Comment by Jaaneek** [comment]
Seems like some other tests are failing once thix is fixed.

Can't find a good solution at this moment.

Feel free to contribute anyone!

temporary fix is add `as const`
```
 queries: [
    { queryKey: ["withSkipToken"], queryFn: true ? () => 1 : skipToken },
  ] as const
  ```

gwansikk:
I've worked on the issue in question. Please take a look.

PR #7150, opened 2024-03-21
fix(*): infer data type correctly when conditional skipToken is passed to useQueries

gwansikk:
Hello, I'm a huge enthusiast of TanStack-query! 👋

I've fixed the issue. closes: #7035
In my quest for a better solution, I've explored several alternatives, yet this appears to be the best I could come up with. Frankly, I don't find this fix particularly elegant or "cool", so I'm open to suggestions for enhancement.

I've run the test and everything checks out, so I'm sending over this PR.

Thank you!

### DIFF
--- a/packages/angular-query-experimental/src/inject-queries.ts
+++ b/packages/angular-query-experimental/src/inject-queries.ts
@@ -136,7 +136,7 @@ export type QueriesOptions<
             [...TResult, GetOptions<Head>],
             [...TDepth, 1]
           >
-        : Array<unknown> extends T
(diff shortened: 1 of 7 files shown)


TkDodo:
thank you. Can you please fix the conflicts and also port this fix to the other adapters that have `useQueries` or `createQueries`?

gwansikk:
Thank you for the review. I have also applied it to Angular, Solid, Svelte, Vue! 🙌

manudeli:
@TkDodo test:format error will be gone after merging https://github.com/TanStack/query/pull/7163

PR #7164, opened 2024-03-23
fix(query-core): add `number`, `symbol` as safe key for `OmitKeyof` for strictness `safely`

manudeli:
## Motivation
TypeScript original Omit's 2nd type parameter's type should be string | number | symbol.

image

But Error is occured when we use OmitKeyof with 'safely' for object having number | symbol key

image

## Solution
So I update constraint of OmitKeyof's 2nd type parameter when using 'safely' like original Omit's one

```diff
  export type OmitKeyof<
    TObject,
    TKey extends TStrictly extends 'safely'
      ?
          | keyof TObject
          | (string & Record<never, never>)
+         | (number & Record<never, never>)
+         | (symbol & Record<never, never>)
      : keyof TObject,
    TStrictly extends 'strictly' | 'safely' = 'strictly',
  > = Omit<TObject, TKey>

```

## Expectation

No error when we use OmitKeyof with 'safely' for object having number | symbol key

image


manudeli:
@TkDodo @tannerlinsley, In my opinion this is simple Pull Request, review this please

TkDodo (approved):

PR #7185, opened 2024-03-27
docs: use nuxt `import.meta.*` properties

danielroe:
This is a very early PR to make these docs compatible with [changes we expect to release in Nuxt v5](https://github.com/nuxt/nuxt/issues/25323).

In [Nuxt v3.7.0](https://github.com/nuxt/nuxt/releases/tag/v3.7.0) we added support for `import.meta.*` (see [original PR](https://github.com/nuxt/nuxt/pull/22428)) and we've been gradually updating docs and moving across from the old `process.*` patterned variables.

These variables are replaced at build-time and enable tree-shaking in bundled code.
This change affects _runtime_ code (that is, that is processed by the Nuxt bundler, like vite or webpack) rather than code running in Node. So it really doesn't matter what the string is, but it makes more sense in an ESM-world to use `import.meta` rather than `process`.

### DIFF
--- a/docs/framework/vue/guides/ssr.md
+++ b/docs/framework/vue/guides/ssr.md
@@ -86,7 +86,7 @@ export default (context) => {
 
   Vue.use(VueQueryPlugin, options)
 
-  if (import.meta.client) {
+  if (process.client) {
     if (context.nuxtState && context.nuxtState.vueQueryState) {
       hydrate(queryClient, context.nuxtState.vueQueryState)
     }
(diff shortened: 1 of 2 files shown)


docs/framework/vue/guides/ssr.md
```diff
@@ -86,7 +86,7 @@ export default (context) => {
 
   Vue.use(VueQueryPlugin, options)
 
-  if (process.client) {
+  if (import.meta.client) {
```
DamianOsipiuk, line 89: I believe this change is unwanted, as it's for `Nuxt 2`

danielroe: You are right. My apologies; I overlooked this.

danielroe, line 89:
```suggestion
  if (process.client) {
```

Issue #6548, opened 2023-12-18
codemod: `remove-overloads` does not handle `useMutation`


slax57:
### Describe the bug

According to the [update guide](https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5), the following replacement needs to be performed on `useMutation`:

```diff
- useMutation(fn, options)
+ useMutation({ mutationFn, ...options })
```

Unfortunately it turns out the `remove-overloads` codemod does not perform this replacement.

### Your minimal, reproducible example

https://stackblitz.com/github/gillkyle/query-v5-codemod?file=src%2Fhooks.ts

### Steps to reproduce

1. Open the sandbox, and wait for `npm install` to finish
2. Run the codemod via the version in `node_modules`:

```bash
npx jscodeshift ./src/ \
  --extensions=ts,tsx \
  --parser=tsx \
  --transform=./node_modules/@tanstack/react-query/build/codemods/src/v5/remove-overloads/remove-overloads.js
```

### Expected behavior

I would have expected the following lines

```tsx
  return useMutation(({ groupId }: { groupId: string }) => {
    return fetch(`/api/groups/${groupId}`, {
      method: "DELETE",
    });
  });
```

to become

```tsx
  return useMutation({
    mutationFn: ({ groupId }: { groupId: string }) => {
      return fetch(`/api/groups/${groupId}`, {
        method: 'DELETE',
      });
    },
  });
```

But the lines remain untouched.

### How often does this bug happen?

Every time

### Platform

- OS: linux
- Browser: Chrome
- Version: 5.14.0

### Tanstack Query adapter

None

### TanStack Query version

5.14.0

### Additional context

Follow up to issue https://github.com/TanStack/query/issues/6204

TkDodo:
@balazsmatepetro can you have a look here please?

balazsmatepetro:
> @balazsmatepetro can you have a look here please?

Sure!

nayan-c247:
Any update on this @balazsmatepetro @TkDodo? facing issue while migrating to v5.

TkDodo:
@balazsmatepetro are you working on this ?

nayan-c247:
@balazsmatepetro @TkDodo

TkDodo:
@nayan-c247 if this is important to you, please consider contributing rather than tagging people. Thanks.

balazsmatepetro:
[...]

PR #7187, opened 2024-03-27
fix(codemod): add support for useMutation

slax57:
Fixes https://github.com/TanStack/query/issues/6548

### DIFF
--- a/packages/query-codemods/src/v5/remove-overloads/__testfixtures__/bug-reports.input.tsx
+++ b/packages/query-codemods/src/v5/remove-overloads/__testfixtures__/bug-reports.input.tsx
@@ -53,3 +53,4 @@ export function useDeleteSomethingWithOnError(): any {
     }
   );
 }
+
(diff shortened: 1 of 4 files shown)


packages/query-codemods/src/v5/remove-overloads/__testfixtures__/bug-reports.input.tsx
```diff
@@ -25,3 +25,31 @@ export function useWhatever({ thing }: { thing: string }) {
...
+    {
+      onError: (_error, _variables, context) => {
+        // An error happened!
+        console.log(
+          `rolling back optimistic delete with id ${context.id}`
+        );
+      },
+    }
+  );
+}
```
balazsmatepetro, line 55: note: Missing empty line. :)

slax57: fixed

PR #7189, opened 2024-03-27
Fix: useSuspenseQueries broken result type

elmess-mehdi:
# Fix: useSuspenseQueries broken result type

## Issue: 
[Issue 7118](https://github.com/TanStack/query/issues/7118)

## What's in this PR ?
- In some adapters ,`queryFn` doesn't include [SkipToken](https://tanstack.com/query/latest/docs/framework/react/guides/disabling-queries) as type.
- This fix included: `angular-query-experimental`, `react-query`, `solid-query`, `svelte-query`, `vue-query`.
- Added a test for `useSuspenseQueries` in `useQueries.test-d.tsx`.


TkDodo:
huh, we've already fixed #7118 with #7194

can you please verify if that fix was correct or if your additional changes are needed? We also concluded that it wasn't necessary for other adapters because they don't have dedicated `useSuspense...` hooks

PR #7191, opened 2024-03-28
Minor example code filename changes to advanced-ssr.md.

sgrobert:
Minor changes from `jsx` to `tsx` to better reflect the correct filename for the code examples displayed.


TkDodo:
the examples aren't written in TS, e.g. `children` aren't typed.

```
export default function RootLayout({ children }) {
```

so jsx is correct

PR #7211, opened 2024-04-02
docs: migrate docs to use path-based navigation instead of route-based

tannerlinsley:
TanStack Router uses path-based navigation instead of route-based navigation, so many of the relative documentation links will be invalid after migration. This fixes that.


tannerlinsley:
I would appreciate any help to track down any other links that I may have missed.

douglasnaphas:
@tannerlinsley I'm not sure if this is related, but there are a lot docs pages linked from elsewhere in the docs that lead to a 404 page.

For example, this page

https://tanstack.com/query/latest/docs/framework/react/reference/useQuery

links to

https://tanstack.com/query/latest/docs/framework/react/reference/query-keys

which leads to a 404 page.

Screenshot 2024-04-07 at 10 35 44 PM
Screenshot 2024-04-07 at 10 36 02 PM

https://tanstack.com/query/latest/docs/framework/react/reference/useQuery also links to https://tanstack.com/query/latest/docs/framework/react/reference/default-query-function, which leads to a 404.

Informally, most links I've followed from React Query docs pages the past few days have led to 404s.

kraenhansen:
@tannerlinsley I noticed the "Walkthough Guide" and "API Reference" links on the individual framework "overview" pages are broken: https://github.com/TanStack/query/blob/main/docs/framework/react/overview.md#you-talked-me-into-it-so-what-now .. this might be related to this change?

PR #7212, opened 2024-04-02
fix(types): add pageParam and direction as optional to "normal" queryFn

### LINKED ISSUE #6715: pageParam missing from default queryFn typescript definition
### Describe the bug

pageParam used to be available in the default queryFn parameter options.
It is still passed (it is working correctly with //@ts-ignore), but typescript is reporting the following error when using  it:

`Property 'pageParam' does not exist on type '{ queryKey: QueryKey; signal: AbortSignal; meta: Record<string, unknown> | undefined; }'`

https://tanstack.com/

### Steps to reproduce

1. create a default query function in a typescript project
2. try to add a pageParam parameter

### Expected behavior

pageParam is present in the typescript definition

Every time

### Screenshots or Videos

_No response_

### Platform

all

### Tanstack Query adapter

react-query

### TanStack Query version

5.17.12

### TypeScript version

4.9.5

### Additional context

_No response_

TkDodo:
because that's what we get when using a defaultQueryFn

closes #6715

### DIFF
--- a/packages/query-core/src/__tests__/query.test.tsx
+++ b/packages/query-core/src/__tests__/query.test.tsx
@@ -208,7 +208,6 @@ describe('query', () => {
     expect(queryFn).toHaveBeenCalledTimes(1)
     const args = queryFn.mock.calls[0]![0]
     expect(args).toBeDefined()
-    // @ts-expect-error page param should be undefined
     expect(args.pageParam).toBeUndefined()
     expect(args.queryKey).toEqual(key)
     expect(args.signal).toBeInstanceOf(AbortSignal)
--- a/packages/query-core/src/__tests__/queryClient.test-d.tsx
+++ b/packages/query-core/src/__tests__/queryClient.test-d.tsx
@@ -1,6 +1,7 @@
 import { describe, expectTypeOf, it } from 'vitest'
 import { QueryClient } from '../queryClient'
-import type { DataTag, InfiniteData } from '../types'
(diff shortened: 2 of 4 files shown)


TkDodo:
@DamianOsipiuk do you know why this small type change leads to this error?

> src/createPersister.ts(81,17): error TS2742: The inferred type of 'experimental_createPersister' cannot be named without a reference to '../node_modules/@tanstack/query-core/build/modern/queryClient-lGFVn5Ps'. This is likely not portable. A type annotation is necessary.

DamianOsipiuk:
> @DamianOsipiuk do you know why this small type change leads to this error?
> 
> > src/createPersister.ts(81,17): error TS2742: The inferred type of 'experimental_createPersister' cannot be named without a reference to '../node_modules/@tanstack/query-core/build/modern/queryClient-lGFVn5Ps'. This is likely not portable. A type annotation is necessary.

This is potentially due to how `tsup` is bundling types.
I remember that @Andarist potentially fixed that in TS. Not sure which version though.

Potential workaround is to import `module` from which problematic type is referenced.

More long term fix would be to fix it in `tsup` or migrate to `rollup`.

Or depend on users upgrading their TS version 😞

Andarist:
> I remember that @Andarist potentially fixed that in TS 5.4.

that fix has been reverted though (see https://github.com/microsoft/TypeScript/pull/57849 )

Issue #6489, opened 2023-12-05
Performance issue in @tanstack/query-async-storage-persister (and possibly #updateStaleTimeout)


Jack-Works:
### Describe the bug

It looks like [delayFunc](https://github.com/TanStack/query/blob/f4822a739a1e93550b45169ab47058f1d669a265/packages/query-async-storage-persister/src/asyncThrottle.ts#L37) in the `packages/query-async-storage-persister/src/asyncThrottle.ts` has some performance problem.

```js
  const delayFunc = async () => {
    clearTimeout(timeout)
    timeout = setTimeout(() => {
      if (running) {
        delayFunc() // Will come here when 'func' execution time is greater than the interval.
      } else {
        execFunc()
      }
    }, interval)
  }
```

<img width="1374" alt="img" src="https://github.com/TanStack/query/assets/5390719/385d65a8-381e-4926-a25e-8245a2c13b51">

this is the 4x slow-down profile, each `refetch` call causes `persistQueryClientSave` to be called, and then `delayFunc` is the slowest function in the tree. All small yellow bars are `setTimeout` or `clearTimeout` calls.

My throttle timeout is the default (1e3), therefore it cannot be caused by `execFunc` (all those `delayFunc` calls happen within 1 second).

I try to patch my node_modules and replace the `asyncThrottle` with this:

```js
function asyncThrottle2(func, { interval = 1e3, onError = noop } = {}) {
    if (typeof func !== 'function') throw new Error('argument is not function.')
    let currentArgs = null
    setInterval(() => {
        if (!currentArgs) return
        Promise.resolve(func(...currentArgs)).catch(onError)
        currentArgs = null
    }, interval)
    return (...args) => {
        currentArgs = args
    }
}
export { asyncThrottle2 as asyncThrottle }
```

This implementation is incorrect but fast.

<img width="1400" alt="img" src="https://github.com/TanStack/query/assets/5390719/01d4c5f0-e413-4656-a89b-c012793d2cd5">

After the patch, `persistQueryClientSave` now only costs 1 ms instead of 57 ms.

I think it looks like calling `setTimeout` or `clearTimeout` is costly. I think a better async throttle implementation can avoid calling them so often.

btw I also observed that `#updateStaleTimeout` method has a similar behavior.

<img width="1177" alt="img" src="https://github.com/TanStack/query/assets/5390719/3728b88e-5e20-4cd5-88ec-7f7c00714652">

[...]

PR #7224, opened 2024-04-04
perf(asyncThrottle): Improve performance

webNeat:
Hello @TkDodo 

This is my attempt to fix the issue https://github.com/TanStack/query/issues/6489 by improving the performance of `asyncThrottle`

Here is the code of a simple benchmark I did https://gist.github.com/webNeat/d4444931e0d2d37502815c6e7b0b6a1e

The results on my machine (running with Node):

```
The avg time it takes in milliseconds to run 1_000_000 times
old version: 127
new version: 19
```

I didn't add any new tests because I kept the exact same behavior as before.

Let me know what you think about this implementation.


TkDodo (approved):
nice work 👏 . I trust the test coverage here, if everything is still green, I'll merge this

PR #7233, opened 2024-04-05
Feature/memoize combine

### LINKED ISSUE #6840: `useQueries` results are not stable even with memorization attempts
### Describe the bug

I'm not able to memorize the result of `useQueries`. Every time the component renders, the useQueries creates a new result object which is not stable.

I tracked the issues here: https://github.com/TanStack/query/issues/2991 but this didn't solved the issue

Even your example on https://tanstack.com/query/latest/docs/framework/react/reference/useQueries is not stable:

```
// Your example: --- FAIL
  const ids = [1, 2, 3];
  const results = useQueries({
    queries: ids.map((id) => ({
      queryKey: ["post", id],
      queryFn: () => fetchPost(id),
      staleTime: Infinity,
    })),
  });

  // Try to memorize --- FAIL
  const queries = useMemo(() => {
    return [4, 5, 6].map((id) => ({
      queryKey: ["post", id],
      queryFn: () => fetchPost(id),
      staleTime: Infinity,
    }));
  }, []);

  const results2 = useQueries({
    queries: queries,
  });

  // Another try to memorize --- FAIL
  const queriesComplete = useMemo(() => {
    return {
      queries: [7, 8, 9].map((id) => ({
        queryKey: ["post", id],
        queryFn: () => fetchPost(id),
        staleTime: Infinity,
      })),
    };
  }, []);

const results3 = useQueries(queriesComplete);
```

https://codesandbox.io/p/devbox/vp3mys?file=%2Fsrc%2FApp.jsx%3A13%2C16

Updated*
### Steps to reproduce

1. Open the sandbox
2. type text within the input
3. See, that the counters going up, thats because the `useEffect` deps are not stable

### Expected behavior

I need a way to get a stable array results and the inner result too which coming from useQuery.

All my components rerender, even if they don't need to. My workaround was to put the entire useQueries in a lower memorized component and lift all states upwards which is really bad.

You may even want to add some documentation, if you found a way to solve this. The useQueries hooks is really annoying to use currently.

Every time

### Screenshots or Videos

_No response_

### Platform

Windos

### Tanstack Query adapter

react-query

### TanStack Query version

5.18.1

### TypeScript version

4.4.4

### Additional context

_No response_

TkDodo:
feat(core): memoize result of combine

this PR makes sure to only re-run the combine function if the function itself changed referentially or any of the inputs changed

fixes #6840

### DIFF
--- a/docs/framework/react/reference/useQueries.md
+++ b/docs/framework/react/reference/useQueries.md
@@ -56,3 +56,12 @@ const combinedQueries = useQueries({
 ```
 
 In the above example, `combinedQueries` will be an object with a `data` and a `pending` property. Note that all other properties of the Query results will be lost.
+
+### Memoization
+
+The `combine` function will only re-run if:
+
+- it the `combine` function itself changed referentially
+- any of the query results changed
+
+This means that an inlined `combine` function, as shown above, will run on every render. To avoid this, you can wrap the `combine` function in `useCallback`, or extract it so a stable function reference if it doesn't have any dependencies.
--- a/docs/framework/react/reference/useQueries.md
+++ b/docs/framework/react/reference/useQueries.md
@@ -61,7 +61,7 @@ In the above example, `combinedQueries` will be an object with a `data` and a `p
 
 The `combine` function will only re-run if:
 
-- it the `combine` function itself changed referentially
+- the `combine` function itself changed referentially
 - any of the query results changed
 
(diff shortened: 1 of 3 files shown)


docs/framework/react/reference/useQueries.md
```diff
@@ -56,3 +56,12 @@ const combinedQueries = useQueries({
 ```
 
 In the above example, `combinedQueries` will be an object with a `data` and a `pending` property. Note that all other properties of the Query results will be lost.
+
+### Memoization
+
+The `combine` function will only re-run if:
+
+- it the `combine` function itself changed referentially
```
TkDodo, line 64:
```suggestion
- the `combine` function itself changed referentially
```

PR #7247, opened 2024-04-08
fix(docs): broken internal links in the query docs

fulopkovacs:
I found 100+ broken links starting with `/query/v5/docs` with [broken-link-checker](https://www.npmjs.com/package/broken-link-checker). This PR fixes them.

Most of these come from moving the `core` pages away from the `framework`-related pages.

# TODO

Fix the links for:
- [x] React
- [x] Angular
- [x] Vue
- [x] Solid
- [x] Svelte 

# How I tested the links

1. Started the docs site locally on `http://localhost:3000` (https://github.com/TanStack/tanstack.com)
2. Ran the following commands for each framework in the local clone of this repo (this example is for angular):
```bash
# Check the site for broken links: print the results in the terminal and save them to this file: `broken-links-angular.txt`
broken-link-checker http://localhost:3000/query/v5/docs/framework/angular/overview  -roe --filter-level 0 --requests 20 | tee broken-links-angular.txt
# Find all the relevant broken links in the `broken-links-angular.txt` file
grep -E '.*BROKEN.*http://localhost:3000/query/v5' broken-links-angular.txt
```
These commands can be chained if you don't need the output in the terminal. For Angular, this is how the output should look like when all the links are fixed (the `injectQuery.md` page is supposed to be missing, since it'll be added in the future):

image

<details>
  <summary>Testing the broken links with a script</summary>
   
```bash
#!/usr/bin/env bash
# Check all the broken links for the React Query docs (v5)
# NOTE: You must install the `broken-link-checker` package from npm
# NOTE: You must run the `tanstack/tanstack.com` site locally on port 3000
# NOTE: This script will run for a loooong time :'(, but in the end it should
# print all the links that are broken in the `v5` docs

# 1. React
broken-link-checker http://localhost:3000/query/v5/docs/framework/react/overview  -roe --filter-level 0 --requests 20 | grep -E '.*BROKEN.*http://localhost:3000/query/v5'

# 2. Vue
broken-link-checker http://localhost:3000/query/v5/docs/framework/vue/overview  -roe --filter-level 0 --requests 20 | grep -E '.*BROKEN.*http://localhost:3000/query/v5'

# 3. Angular
broken-link-checker http://localhost:3000/query/v5/docs/framework/angular/overview  -roe --filter-level 0 --requests 20 | grep -E '.*BROKEN.*http://localhost:3000/query/v5'

# 4. Solid
[...]

PR #7248, opened 2024-04-08
docs: broken top-level relative pathing on links

SeanCassiere:
Something I found in the docs for Router, that looks to be broken on the other TanStack projects as well.

The relative navigation for top-level docs, looks to only be broken in the React, Vue, and Svelte documentation.


SeanCassiere:
Closing as is superseded by #7247

PR #7253, opened 2024-04-09
feat!(eslint-plugin-query): Migrate to ESLint v9 with Flat Config Support

davidjbng:
This PR is a starting point to add support for the newly released [ESLint v9](https://eslint.org/blog/2024/04/eslint-v9.0.0-released/) as well as exposing a new `flat/recommended` to be used in `eslint.config.js`.

*Edit: I noticed the separate `flat/recommended` config may not be necessary given it exports the same rules as the `recommended` config anyway.*

~I confirmed this works on my project testing only with `eslint@9`,~ but haven't tested the old `recommended` config, neither did I test with `eslint@8`.

## Resources
I used the following resources to find out which changes to make
- [Plugin Migration to Flat Config](https://eslint.org/docs/latest/extend/plugin-migration-flat-config#migrating-rules-for-flat-config)
- [Implentation of eslint-plugin-jest](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/index.ts)
- [Breaking Changes for plugin developers in ESLint v9](https://eslint.org/docs/latest/use/migrate-to-9.0.0#breaking-changes-for-plugin-developers)

## To Be Discussed
 
- [ ] Support for the Flat Config could be its own PR if the changes to support ESLint v9 turn out to be breaking
- [x] Use the existing 'recommended' config for Flat Config support if possible. *Rule definitions requirements seem to have changed so having a separate `flat` config is required to support both formats.
- [ ] Find a good way to fully test the exported config. Including the plugin object itself as required by flat config adds a circular reference with a lot of noise in the snapshot
- [ ] See if config types could be improved
- [ ] I had some trouble using the build locally and had to change the `exports` in packages.json from `./dist/esm/index.js` to `./dist/esm/src/index.js` to use the package installed from a directory. Probably not an issue if it works in ci, but wanted to mention it just in case.

## TODO
- [ ] Update Docs
- [ ] Test with ESLint v8
- [ ] Update examples/react/shadow-dom and examples/react/basic-typescript

## Related Issues

- typescript-eslint/typescript-eslint/issues/8211
- facebook/react/issues/28313


TkDodo:
Thanks for this. We'd definitely need to update docs and show how to use the plugin both for flat config and the v8 syntax.

TkDodo:
lockfile seems to be outdated:

```
 ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with packages/eslint-plugin-query/package.json
```

davidjbng:
> lockfile seems to be outdated:
> 
> ```
>  ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with packages/eslint-plugin-query/package.json
> ```

The root package.json requires all packages to use the same ESLint version.
https://github.com/TanStack/query/blob/fca5e7df77c7984d0505c3e924b879dcd9a5e763/package.json#L83-L89

Besides the ESLint Plugin - which I updated to v9 - eslint v8 is currently depended on by examples/react/shadow-dom and examples/react/basic-typescript which need to be fixed. Updating the examples to Flat Config is blocked by the upstream react plugin issues though.

Newbie012 (approved):
LGTM

davidjbng:
I think this has to wait until it is offically supported by `typescript-eslint` which is tracked here
- typescript-eslint/typescript-eslint/issues/8211

The tests are still failing because of differences how rule definitions work for the flat config format, but `@typescript-eslint/rule-tester` isn't up to date to report those errors in typescript. Rather you get runtime errors directly from eslint.

Newbie012:
It seems like typescript-eslint now supports ESLint 9 (https://github.com/typescript-eslint/typescript-eslint/issues/8211#issuecomment-2106427365)

jgarplind:
You may want to consider adding these efforts to the issues over in the ESLint repo tracking overall progress among plugin developers.

That would be very useful for app developers like myself who try to keep track of how the various plug-ins we are using are progressing!

If flat config is _not_ yet supported, add [here](https://github.com/eslint/eslint/issues/18093). If flat config is supported but v9 is not, add [here](https://github.com/eslint/eslint/issues/18391).

davidjbng:
I will work a new pr for eslint v9 first and then flat config

PR #7257, opened 2024-04-10
feat(vue-query): add support for infiniteQueryOptions

DavideSegullo:
The goal of this PR is to support the `infiniteQueryOptions` utility as it already does for react:
https://tanstack.com/query/latest/docs/framework/react/reference/infiniteQueryOptions

TODO:

- [x] Add tests

### DIFF
--- a/packages/vue-query/src/__tests__/infiniteQueryOptions.types.test.ts
+++ b/packages/vue-query/src/__tests__/infiniteQueryOptions.types.test.ts
@@ -1,0 +1,150 @@
+import { describe, expectTypeOf, it } from 'vitest'
+import { QueryClient } from '@tanstack/query-core'
+import { reactive } from 'vue-demi'
+import { infiniteQueryOptions } from '../infiniteQueryOptions'
+import { useInfiniteQuery } from '../useInfiniteQuery'
+import { type Equal, type Expect, doNotExecute } from './test-utils'
+import type { InfiniteData, dataTagSymbol } from '@tanstack/query-core'
+
+describe('infiniteQueryOptions', () => {
+  it('should not allow excess properties', () => {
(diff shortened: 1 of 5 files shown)


DavideSegullo:
Hola @TkDodo When you can, could I get some feedback on this PR? Thank you so much!

XantreDev:
It also do not exist for solid-query. Maybe it makes sense to add function type to the core package for easier reuse?

DavideSegullo:
> It also do not exist for solid-query. Maybe it makes sense to add function type to the core package for easier reuse?

Idk if you can share the implementation, some type for example on vue are constrained to vue types, do you have any ideas?

XantreDev:
Probably I need to make contribution with the 95% the same copy pasted code. Because there is not actual logic - only types

XantreDev:
Btw, in this implementation and react implementation pageParams has always `unknown[]` type

DavideSegullo:
> Btw, in this implementation and react implementation pageParams has always `unknown[]` type Is it desirable behvaiour?

Uhm I just followed what was being done on react

TkDodo:
`pageParams` returned from `useInfiniteQuery` has `unknown` as type, yes. There isn't much need to use those during rendering and it was a quick implementation to not have to add more type parameters.

XantreDev:
@TkDodo We already have a parameter for that. The problem that TData do not use it. Since TPageParam is declared after TData - we need to make some trick to do not make backward compatibility
```ts
UseInfiniteQueryOptions<
  TQueryFnData,
  TError,
  TData,
  TQueryFnData,
  TQueryKey,
  TPageParam
>
```

TkDodo (approved):

PR #7267, opened 2024-04-11
Update important-defaults.md

Jack-R-Long:
Link directly to TkDodo's blog


TkDodo:
liking to the community page is on purpose here.

PR #7282, opened 2024-04-15
(docs) new react-native query devtools.md

jossydeleon:
Update query docs with a new tool to debug query in react-native https://github.com/jossydeleon/react-native-query-devtool-monorepo


TkDodo:
it's a bit weird that we have some tools linked in the react native section and some here:

https://tanstack.com/query/v5/docs/framework/react/react-native

it would be great if you could consolidate a list into a single place and then link the other page to it 🙏

TkDodo:
we should  have all react-native specific things in the react-native docs, and we can link to them from the devtools docs. please file a new PR if you want that.

PR #7303, opened 2024-04-19
Use `isInternetReachable` instead of `isConnected` in the online manager example

bitcrumb:
The `isConnected` property actually only conveys that the network connection has been established (e.g. Wi-Fi connection), which not necessarily implies that there is a working _internet_ connection.

So I as wondering if it is not better to use `isInternetReachable` instead? Although I am aware (and thus a bit hesitant as well) that the netinfo library implements this through a HEAD request to a Google URL.

### DIFF
--- a/docs/framework/react/react-native.md
+++ b/docs/framework/react/react-native.md
@@ -20,7 +20,7 @@ import { onlineManager } from 'react-query'
 
 onlineManager.setEventListener(setOnline => {
   return NetInfo.addEventListener(state => {
-    setOnline(state.isConnected)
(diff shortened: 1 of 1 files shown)


TkDodo:
I guess it depends on what you want. I don't know react native much and this guide was contributed by someone else. If you say `isInternetReachable` is better, we can change it. There are some open issues about it though:

- https://github.com/react-native-netinfo/react-native-netinfo/issues/615
- https://github.com/react-native-netinfo/react-native-netinfo/issues/572

TkDodo:
closing as stale

PR #7310, opened 2024-04-20
fix(query-core): Before use globalThis, check its existence.

kaikidd:
I'm utilizing @tanstack/react-query within a React project operating in a unique environment. However, during runtime, the global object 'globalThis' is not available, causing errors to be displayed in the console and bringing the entire application down. I believe that implementing a check for the existence of 'globalThis' before its usage could resolve this issue.
截屏2024-04-20 23 33 48

### DIFF
--- a/packages/query-core/src/utils.ts
+++ b/packages/query-core/src/utils.ts
@@ -62,7 +62,7 @@ export type QueryTypeFilter = 'all' | 'active' | 'inactive'
 
 // UTILS
 
-export const isServer = typeof window === 'undefined' || 'Deno' in globalThis
+export const isServer = typeof window === 'undefined' || (typeof globalThis !== 'undefined' && 'Deno' in globalThis)
 
 export function noop(): undefined {
   return undefined


TkDodo:
what kind of environment are you working in please? `globalThis` should be available in all envs that we support:
- https://caniuse.com/mdn-javascript_builtins_globalthis

kaikidd:
I use @tanstack/react-query in @tarojs/cli(https://github.com/NervJS/taro), it is a runtime framwork tailored for developing specialized mini-applications designed to run on Android and IOS application in China.
The customized runtime environment encompasses most browser environments, but there are exceptions, such as the absence of 'globalThis'.
I use @tanstack/react-query correctly before a pull request fix(query-core): Replace the deprecated 'window' with 'globalThis' fo...(https://github.com/TanStack/query/pull/7029).
@tanstack/react-query version 5.26.3 works fine for me.

TkDodo:
> The customized runtime environment encompasses most browser environments, but there are exceptions, such as the absence of 'globalThis'.

`globalThis` is a standard that was introduced to avoid checks for `window`, `global` or `self` in different environments. It's supposed to be available in all modern environments.

If it's not available for you, please add a polyfill that will set `globalThis` to, idk, `window` or an empty object or whatever you want. As a last resort, you can also `patch-package` react-query.

hengkx:
How to resolve? Thanks

tofrankie:
When developing [Weixin Mini Program](https://developers.weixin.qq.com/miniprogram/en/dev/framework/quickstart/framework.html) using uni-app (a framework that supports developing mini-program applications using Vue.js), I encountered a problem with the [`isServer`](https://github.com/TanStack/query/blob/ee16773c702bc8add5887ec811edd88009b44436/packages/query-core/src/utils.ts#L89) conditional logic (the mini-program runtime environment lacks a `window` object), preventing me from using [`refetchInterval`](https://github.com/TanStack/query/blob/ee16773c702bc8add5887ec811edd88009b44436/packages/query-core/src/queryObserver.ts#L392-L399) because the server doesn't support periodically re-fetching data.

> The Weixin Mini Program framework's logical layer does not run in a browser, so JavaScript is not available on the web, such as `window`, `document` etc. (from [here](https://developers.weixin.qq.com/miniprogram/en/dev/framework/app-service/))

Of course, I'm not blaming TanStack Query for incompatibility with Weixin Mini Program. After all, [MiniApps](https://github.com/w3c/miniapp) (not just Weixin Mini Program) are an application format derived from standard web applications. What I mean is that MiniApps are very popular and widely used in Chinese internet companies, and I want to make full use of Tanstack Query's capabilities in my MiniApps project; it is indeed an excellent library. 👍

Are there other solutions? For example, providing configuration options that allow users to customize whether the current project is a server-side project, which would be more adaptable to various development environments.

@TkDodo

tofrankie:
@kaikidd 请问下 Taro  可以接入 @tanstack/react-query 吗？我尝试了在 Taro 4.0 接入似乎不太行。

PR #7315, opened 2024-04-21
docs(query keys): Update query-keys.md

Silverium:
links are broken


TkDodo:
all links will be fixed here:
- #7247

PR #7316, opened 2024-04-21
docs: Update link address

devnjw:
The address of `Important Defaults` was written incorrectly. I've fixed it with the correct link.


TkDodo:
all links will be fixed here:
- #7247

PR #7317, opened 2024-04-21
Update typescript.md

Best-Sardar:
In the Further Reading section, the provided links referred to pages that no longer existed. These pages were present in version 3, but have been removed since then.


TkDodo:
they haven't been removed, it's just that some links are wrong. This will be fixed with:
- #7247 

the page still exists here:
- https://tanstack.com/query/v5/docs/framework/react/community/tkdodos-blog

Issue #7154, opened 2024-03-21
codemod: ERR_REQUIRE_ESM when migrating from v4 to v5


wangwalton:
### Describe the bug

ERR_REQUIRE_ESM when migrating from v4 to v5 with codegen

### Your minimal, reproducible example

NA

### Steps to reproduce

Running this 

```
 npx jscodeshift@latest ./src/ \                                                                                                                                                                                      main *
  --extensions=ts,tsx \
  --parser=tsx \
  --transform=./node_modules/@tanstack/react-query/build/codemods/src/v5/remove-overloads/remove-overloads.js
```

gives this error:
```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/dev/node_modules/@tanstack/react-query/build/codemods/src/v5/remove-overloads/remove-overloads.js from /Users/.npm/_npx/a3dac9416cb5dcf1/node_modules/jscodeshift/src/Worker.js not supported.
remove-overloads.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead either rename remove-overloads.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /Users/./node_modules/@tanstack/react-query/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).

    at Object.newLoader [as .js] (/Users//.npm/_npx/a3dac9416cb5dcf1/node_modules/pirates/lib/index.js:121:7)
    at setup (./_npx/a3dac9416cb5dcf1/node_modules/jscodeshift/src/Worker.js:93:18)
    at Object.<anonymous> (./_npx/a3dac9416cb5dcf1/node_modules/jscodeshift/src/Worker.js:45:3) {
  code: 'ERR_REQUIRE_ESM'
}
```

### Expected behavior

No error

### How often does this bug happen?

None

### Platform

macos

### Tanstack Query adapter

None

### TanStack Query version

5.28.6

TkDodo:
this is the same as :

- https://github.com/TanStack/query/discussions/7048

if you know how to fix it, please contribute.

lachlancollins:
Fixed in #7333

Pierre-Do:
Thanks for the fix! Would you mind releasing a version that includes it? Cheers!

TkDodo:
it should've been in 5.32.1

not sure why that hasn't been released... checking ...

[...]

PR #7333, opened 2024-04-24
fix(codemods): Use .cjs extension for CommonJS files

ontanj:
Rename CommonJS files in codemods to distinguish them from ES module files otherwise used in this project.

Fix issue #7154 which occurs since package.json in codemods is removed when publishing, causing module type data to be lost.

The issue can also be fixed by adding package.json to the published package. But since the project mainly uses ES modules it makes sense to use .cjs for the others.


lachlancollins (approved):
LGTM!

ontanj:
Another solution to #7154 would of course be to rewrite codemods as ES modules. Is there a reason it is using CommonJS?

PR #7343, opened 2024-04-28
Add context as second parameter for the mutation function

asermax:
This PR adds a context that's passed as the second parameter to the `mutationFn` in a mutation. This change allows access to the mutation's `meta` (similar to query functions), which I find useful for passing in configurations (similarly as the case mentioned in https://github.com/TanStack/query/discussions/7045#discussioncomment-9151048). I've also included the `mutationKey` in the context, since it came up in a similar discussion #3504. 

While this context could eventually include other information, I kept this initial contribution small and simple as to not get all tangled up with all the typing :sweat_smile:

### DIFF
--- a/docs/framework/react/reference/useMutation.md
+++ b/docs/framework/react/reference/useMutation.md
@@ -45,10 +45,11 @@ mutate(variables, {
 
 **Options**
 
-- `mutationFn: (variables: TVariables) => Promise<TData>`
+- `mutationFn: (variables: TVariables, context: MutationFunctionContext) => Promise<TData>`
   - **Required, but only if no default mutation function has been defined**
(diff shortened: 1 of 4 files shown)


ATypescriptEnjoyer:
It'd be really cool to get this in 🙏 I use the meta for accessing bearer tokens which works for useQuery but obviously not useMutation

DavidReinberger:
We'd also really like to get this in, accessing `ky` client in queries/mutations without the workaround mentioned [here](https://github.com/TanStack/query/discussions/7045#discussioncomment-11933375).

TkDodo:
I would want to consolidate the efforts in this PR please:

- https://github.com/TanStack/query/pull/9193

PR #7350, opened 2024-04-29
feat(angular-query-experimental): support injection in injectQuery callback

ShacharHarshuv:
Before, when creating a query with injectQuery, if you injected dependencies (Like `inject(HttpClient)`) those injections would fail on subsequence changes of dependent signals. That change makes it easier to use `injectQuery` (especially together with `queryOptions` because the query can define its own dependencies.

Possible followup: I wonder if we should also wrap queryFn and initialData and placeholderData in an injection context so that it would be possible to inject dependencies in these as well. Right now it will work on the first time but fail (silently, by default) on subsequent fetches.

No breaking changes, any code that worked before should work now as well.


ShacharHarshuv:
It seems like the styling changes were done automatically on commit by lint-staged. Let me know if you think something in my set up is wrong and it shouldn't have happened.

arnoud-dv:
> It seems like the styling changes were done automatically on commit by lint-staged. Let me know if you think something in my set up is wrong and it shouldn't have happened.

Yes that shouldn't happen, there shouldn't be semicolons at the end of lines for example. See prettier.config.js in the root.

ShacharHarshuv:
Fix eslint error. 

@arnoud-dv Who needs to approve to get this merged?

arnoud-dv:
> Fix eslint error.
> 
> @arnoud-dv Who needs to approve to get this merged?

I can approve it. Looks good at first glance, will take a more in depth look tomorrow. Thank you!

Issue #7341, opened 2024-04-26
queryOptions should to pass initialData with a callback that is potentially returning undefined


ShacharHarshuv:
### Describe the bug

In angular-query
When using `queryOptions` `'@tanstack/angular-query-experimental'` and passing an initial value function, typescript will not allow that typescript to return undefined. The following code will not compile:

```ts
interface Todo {
  id: string;
  title: string;
}

export const todoQueryOption = (id: string | null) =>
  queryOptions({
    queryKey: ['todo', id],
    queryFn: () =>
      Promise.resolve({
        id: '1',
        title: 'Do Laundry',
      }),
    initialData: (): Todo | undefined => undefined,
  });
```

This should be allowed because I might want to check if I have this particular "todo" item in the cache, and if I don't, I would want angular query to request that item. Since this is conditional, I have to put it in a function that might either have a todo or not (in the example I set it to undefined just for the example).

I believe this has to do with this type definition: 

```ts
export type DefinedInitialDataOptions<
  TQueryFnData = unknown,
  TError = DefaultError,
  TData = TQueryFnData,
  TQueryKey extends QueryKey = QueryKey,
> = CreateQueryOptions<TQueryFnData, TError, TData, TQueryKey> & {
  initialData:
    | NonUndefinedGuard<TQueryFnData>
    | (() => NonUndefinedGuard<TQueryFnData>)
}
```

Which should be changed to:

```ts
export type DefinedInitialDataOptions<
  TQueryFnData = unknown,
  TError = DefaultError,
  TData = TQueryFnData,
  TQueryKey extends QueryKey = QueryKey,
> = CreateQueryOptions<TQueryFnData, TError, TData, TQueryKey> & {
  initialData:
    | NonUndefinedGuard<TQueryFnData>
    | (InitialDataFunction<TQueryFnData>)
}
```

### Your minimal, reproducible example

https://stackblitz.com/edit/stackblitz-starters-jdj53x?file=src%2Fmain.ts

### Steps to reproduce

1. Write this code in angular query: 
```ts
interface Todo {
  id: string;
  title: string;
}

export const todoQueryOption = (id: string | null) =>
  queryOptions({
    queryKey: ['todo', id],
    queryFn: () =>
      Promise.resolve({
        id: '1',
        title: 'Do Laundry',
      }),
    initialData: (): Todo | undefined => undefined,
  });
```
2. It wouldn't compile. 

### Expected behavior

The code should compile

### How often does this bug happen?

Every time

[...]

PR #7351, opened 2024-04-29
fix(angular-query-experimental): allow returning undefined in initialData function

ShacharHarshuv:
Fix incorrect types that disallowed that. This is necessary because the initialData function can typically read cache synchronously and need to be able to signal to the query if there is a cache miss and an actual fetch needs to be deployed.

Fixes https://github.com/TanStack/query/issues/7341

No breaking changes, any code that worked before should work now as well.


ShacharHarshuv:
It seems like the styling changes were done automatically on commit by lint-staged. Let me know if you think something in my set up is wrong and it shouldn't have happened.

ShacharHarshuv:
I fixed the prettier issue

TkDodo (commented):
can you please fix it for `react-query` adapter as well, and also have the test there? Thanks 🙏

ShacharHarshuv:
@TkDodo Done!

ShacharHarshuv:
@TkDodo @arnoud-dv Can this be merged?

arnoud-dv:
Looks good to me, @TkDodo agree if we merge this?

TkDodo:
sorry, this somehow slipped through my queue. will merge now.

PR #7357, opened 2024-04-30
fix: replaceDeepEqual for objects created by Object.create

sukovanej:
Currently, objects with custom prototypes (created using `Object.create(...)`) are dealt with as plain objects. That causes them to be copied without the prototype in case of refetch / invalidation and results in buggy behaviour. This is problematic when using together with https://github.com/Effect-TS/effect (e.g `Option`, `Either`, ... data structures).

BTW, based on the test suite, it might cover other checks like `prot.hasOwnProperty('isPrototypeOf')` so the impl of `isPlainObject` could be refactored a little but I'm honestly not 100% sure.


sukovanej:
This PR should be ready. @TkDodo could you please take a look?

TkDodo (approved):

PR #7360, opened 2024-05-01
fix(angular-query-experimental): run mutation callback in injection context

ShacharHarshuv:
Previously, the `injectMutation` callback ran in injection context only if accessed in the same context as the component initialization (i.e. before the `effect` callback run). By wrapping the effect callback in injection context, it is ensured that the callback will always run in injection context, and any `inject` calls will not fail.

Note there is a separate issue here, and it's that the callback is always run twice - once synchronously when initialization the mutation (when you create the `MutationObserver`), and secondly as the first callback of the `effect`. This is something that can potentially be improved to prevent redundant code run.

Another related issue - I would argue that if any signal dependencies of the mutation changed, you don't necessarily want to immediately run the callback again. You probably want to be "lazy" and run it again only when the mutation object is accessed. With the current implementation, that callback might unnecessarily run when there are a lot of dependency changes, but no mutation is used. (For example, imagine there is an "opened todo item" state, and a delete button on it. The user might switched the opened todo item state frequently without actually using the delete mutation). 

No breaking changes, any code that worked before should still work.


packages/angular-query-experimental/src/__tests__/inject-mutation.test.ts
```diff
@@ -455,4 +455,38 @@ describe('injectMutation', () => {
...
+          mutationFn: (name: string) => service.updateData(name),
+        }
+      })
+    }
+
+    const fixture = TestBed.createComponent(FakeComponent)
+    fixture.detectChanges()
+
+    // check if injection contexts persist in a different task
+    await new Promise<void>((resolve) => queueMicrotask(() => resolve()))
```
ShacharHarshuv, line 486: Without this, the test will pass with the previous implementation. That is because the callback of `effect` which causes the exception will run in a different task.

ShacharHarshuv:
Resolved conflicts. 
@arnoud-dv can you review?

ShacharHarshuv:
The tests failures don't seem to be related to my change, as they fail in the react adapter.

packages/angular-query-experimental/src/inject-mutation.ts
```diff
@@ -70,7 +70,11 @@ export function injectMutation<
         }
 
         effect(() => {
-          observer.setOptions(optionsFn(queryClient))
+          observer.setOptions(
+            runInInjectionContext(currentInjector, () =>
+              optionsFn(queryClient),
```
kfrancois, line 75: Similar to the way `defaultQueryOptions` are applied to queries, should we apply `queryClient.defaultMutationOptions` here?

ShacharHarshuv: Is this related to this PR's goal? Might be better to keep it separate.

ShacharHarshuv:
Can this be merged?

packages/angular-query-experimental/src/inject-mutation.ts
kfrancois: Somewhat - I think both should be merged, but this one isn't blocked & can go ahead first

kfrancois:
@arnoud-dv I believe this one's good to be merged, wdyt?

packages/angular-query-experimental/src/inject-mutation.ts
arnoud-dv: > Similar to the way `defaultQueryOptions` are applied to queries, should we apply `queryClient.defaultMutationOptions` here?

The other adapters don't do this either.

arnoud-dv:
Hi I just ran the test from this PR on my laptop and it passes with the current implementation. Can you create a test or a repro that demonstrates the issue?

kfrancois:
Nice catch @arnoud-dv ! Thank you for checking this out.

@ShacharHarshuv I added a test in https://github.com/kfrancois/query/commit/eb05ac27314401d551e76e962ad58c758cd79a04#diff-aa4e4fa79cbfdc8cb15c0cfb49c3b858294ec82865c5e2cdbf0ba98e2927dfe4R459-R502 which I verified fails with the current implementation. If you think this can be useful, feel free to take that change along in this PR

ShacharHarshuv:
@arnoud-dv Apparently the error wasn't caught by the test, thanks for noticing. I took @kfrancois's approach of using an "errorSpy" and manually catching the error. Now the test will fail as expected without my change.

PR #7362, opened 2024-05-01
feat(angular-query-experimental): run query options callbacks in injection context

ShacharHarshuv:
This is to allow consumers to use `inject` inside callbacks like `queryFn`, `initialData` etc.

### DIFF
--- a/packages/angular-query-experimental/src/__tests__/inject-query.test.ts
+++ b/packages/angular-query-experimental/src/__tests__/inject-query.test.ts
@@ -5,345 +5,386 @@ import {
   inject,
   input,
   signal,
-} from '@angular/core'
-import { TestBed, fakeAsync, flush, tick } from '@angular/core/testing'
-import { QueryClient } from '@tanstack/query-core'
-import { describe, expect, vi } from 'vitest'
-import { injectQuery } from '../inject-query'
-import { provideAngularQuery } from '../providers'
+} from "@angular/core";
+import { TestBed, fakeAsync, flush, tick } from "@angular/core/testing";
+import { QueryClient } from "@tanstack/query-core";
+import { describe, expect, vi } from "vitest";
+import { injectQuery } from "../inject-query";
+import { provideAngularQuery } from "../providers";
 import {
   delayedFetcher,
   getSimpleFetcherWithReturnData,
   rejectFetcher,
   setSignalInputs,
   simpleFetcher,
-} from './test-utils'
+} from "./test-utils";
 
-describe('injectQuery', () => {
+describe("injectQuery", () => {
   beforeEach(() => {
     TestBed.configureTestingModule({
       providers: [provideAngularQuery(new QueryClient())],
-    })
-  })
+    });
+  });
 
-  test('should return pending status initially', fakeAsync(() => {
+  test("should return pending status initially", fakeAsync(() => {
     const query = TestBed.runInInjectionContext(() => {
       return injectQuery(() => ({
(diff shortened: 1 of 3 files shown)


ShacharHarshuv:
Converted to draft as we probably want to add the same thing for `injectMutation` (and maybe others?)

riccardoperra:
Just my opinion, this can lead to unexpected usage from consumers and gives too much freedom 🤔

Angular doesn’t allow to inject through callback or methods, but as a workaround you can explicitly use runInInjectionContext.

Shouldn’t properties usually be injected in a class field or inside the query options callback?

ShacharHarshuv:
I get what you are saying, but in practice, and in many use cases, not being able to do it here just makes DX more difficult. 

Here's an example:

```
cosnt todosQueryOptions = () => {
  const http = inject(HttpClient);

  return queryOptions({ queryKeys: ['todos'], ... });
}
```
Now I want to invalidate that query (or optimistically update) after creating a new todo, in a type safe way:
```ts
onSuccess: async (response) => { queryClient.invalidateQueries(todosQueryOptions()) }
```
This will fail because `onSuccess` is not in an injection context, and `todosQueryOptions` is an injection function. 

The way I see it, Angular "injection context" is set in this way because code that doesn't run as initialization of a component could be run from any context and "required" any arbitrary injector. In other words - Angular wouldn't know what injector to use.

However - in this case - since there is a place you are "injecting" the query, it's fair to set it up so that all of the queries work will be (by default) in the same injection context. You can still wrap it in a different injection context with a different injection if you want, but I think that use case will not be very common.

riccardoperra:
I understood the issue 😄 

I always used options with a service so I've didn't had this error before

```ts
@Injectable({providedIn: 'root'})
class TodoQuery { 
  readonly http = inject(HttpClient);

  todosQueryOptions() {
    const http = inject(HttpClient);

    return queryOptions({ queryKeys: ['todos'], ... });
  }
} 
```

then 

```ts
readonly todoQuery = inject(TodoQuery);

onSuccess: async (response) => { queryClient.invalidateQueries(this.todoQuery.todosQueryOptions()) }
```

ShacharHarshuv:
You first example wouldn't work because you are injecting inside the `todosQueryOptions` method. When you call it in a non-injection context it will throw. I assume you meant that you don't inject that method by use the member instead? I find that approach cumbersome as I would ideally want to encapsulate my dependencies as much as possible, and also separate things like "query options" to different files / symbols and not couple them to the same service I'm using them for.

This becomes more annoying if my query options creator function accepts an argument (like id). 

Overall, while workarounds exist, I believe my proposed change improves DX and doesn't have many negative side-effects. cc @arnoud-dv What do you think?

arnoud-dv:
Matthieu Riegler wrote a [good article on `inject`](https://riegler.fr/blog/2025-01-08-inject-not-service-locator/), why it's not a service locator but can become one when used for injection after initialisation. This PR would enable injection after initialisation and while sometimes convenient it's probably an anti-pattern. So I think we should not enable this functionality and if the developer really wants to do so they can use `runInInjectionContext` themselves.

I'll also check the existing code in the adapter to see if it only uses `inject` as intended.

What do you think @ShacharHarshuv @riccardoperra ?

arnoud-dv:
IMO dependencies should be clearly visible in the code and not have the potential to be different between callback executions. Callback functions of APIs in Angular such as `computed` or `effect` don't run in the injection context and I think it's best to align with that.

PR #7388, opened 2024-05-07
fix(angular-query): missing exports InitialDataInfiniteOptions types

Arthie:
Add missing `@public` js-doc comments in infinite-query-options.ts to fix type exports. 

Currently the InitialDataInfiniteOptions types are not exported and can't be used. I assume this is due to the change to jsdoc and missing the `@public` comment for these types. This change would make this consistent with similar InitialDataOptions types form (query-options.ts).

### DIFF
--- a/packages/angular-query-experimental/.prettierignore
+++ b/packages/angular-query-experimental/.prettierignore
@@ -1,2 +1,2 @@
 # API extractor
-*.api.md
+etc/**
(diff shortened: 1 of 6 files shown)


arnoud-dv:
They were in fact not exported by the main `index.ts` file. Added that and some other fixes to the public API types. Thanks!

PR #7399, opened 2024-05-09
docs(query-core): update tsdocs for mutation and query observer result

bryanltobing:
Content is directly taken from the website API docs.

https://tanstack.com/query/latest/docs/framework/react/reference/useQuery#usequery
https://tanstack.com/query/latest/docs/framework/react/reference/useMutation#usemutation

**WHY**
- In some code editors like VSCode, we can hover through each result to get information about each key.
- We can also go to the type definition and get all information without the need to open website docs.

https://github.com/TanStack/query/assets/46083126/a3e31c60-00ac-47fc-a14f-57abebd8299e

**Prev discussions**

https://twitter.com/TkDodo/status/1784684695147483440

### DIFF
--- a/packages/query-core/src/types.ts
+++ b/packages/query-core/src/types.ts
@@ -492,35 +492,124 @@ export interface QueryObserverBaseResult<
   TData = unknown,
   TError = DefaultError,
 > {
+  /**
+   * The last successfully resolved data for the query.
+   * - Defaults to `undefined`.
+   */
   data: TData | undefined
+  /**
+   * The timestamp for when the query most recently returned the `status` as `"success"`.
+   */
   dataUpdatedAt: number
+  /**
+   * The error object for the query, if an error was thrown.
+   * - Defaults to `null`.
+   */
   error: TError | null
+  /**
+   * The timestamp for when the query most recently returned the `status` as `"error"`.
+   */
   errorUpdatedAt: number
+  /**
+   * The failure count for the query.
+   * - Incremented every time the query fails.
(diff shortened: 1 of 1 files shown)


packages/query-core/src/types.ts
```diff
@@ -492,35 +492,124 @@ export interface QueryObserverBaseResult<
...
+   */
   isFetched: boolean
+  /**
+   * Will be `true` if the query has been fetched after the component mounted.
+   * - This property can be used to not show any previously cached data.
+   */
   isFetchedAfterMount: boolean
+  /**
+   * A derived boolean from the `fetchStatus` variable, provided for convenience.
+   * - `true` whenever the queryFn is executing, which includes initial `pending` as well as background refetch.
```
TkDodo, line 544:
```suggestion
   * - `true` whenever the `queryFn` is executing, which includes initial `pending` as well as background refetch.
```

packages/query-core/src/types.ts
```diff
@@ -492,35 +492,124 @@ export interface QueryObserverBaseResult<
...
+  /**
+   * Will be `pending` if there's no cached data and no query attempt was finished yet.
+   */
   isPending: boolean
+  /**
+   * Will be `true` if the query failed while fetching for the first time.
+   */
   isLoadingError: boolean
   /**
    * @deprecated isInitialLoading is being deprecated in favor of isLoading
```
TkDodo, line 561:
```suggestion
   * @deprecated `isInitialLoading` is being deprecated in favor of `isLoading`
```

PR #7403, opened 2024-05-09
Fix incorrect ref path

dc-shackle:
react -> vue


TkDodo:
the points at the baseline. vue takes the react docs as a base so this is correct

PR #7404, opened 2024-05-10
feat(solid-query): implemented infiniteQueryOptions

XantreDev:
The goal of this PR is to support the infiniteQueryOptions utility as it already does for react:
[tanstack.com/query/latest/docs/framework/react/reference/infiniteQueryOptions](https://tanstack.com/query/latest/docs/framework/react/reference/infiniteQueryOptions)


TkDodo:
thanks, can you please fix the conflicts and it also seems that there are some prettier related changes that shouldn't have happened (like changing `'` to `"` and adding semicolons)

XantreDev:
Not a problem, seems to be prettier didn't work correctly

XantreDev:
@TkDodo done)

TkDodo (approved):

PR #7409, opened 2024-05-10
feat(react-query): add default query options provider

robingullo:
The goal of this PR is to allow providing default query options to a React subtree. So basically like `queryClient.setDefaultOptions` / `queryClient.setQueryDefaults` but scoped to a component subtree.

Technically this can already be done in userland, but it would require reading in the context before each useQuery and/or creating a custom useQuery hook. This PR would allow to do it without any changes to useQuery calls.

### Example: automatic polling 

```
import { QueryDefaultOptionsProvider } from '@tanstack/react-query'

// All queries in this page will refetch every 5s
// The same queries in another page will fetch at most once on mount
export const Home = () => {
  return (
    <QueryDefaultOptionsProvider
      options={{ queries: { refetchInterval: () => 5000 } }}
    >
      ...
    </QueryDefaultOptionsProvider>
  )
}
```

### Benefits for React Native

With React Navigation (the most popular navigation library in React Native), screens in the stack history [stay mounted](https://reactnavigation.org/docs/navigation-lifecycle/). That does not work well with React Query's default `refetchOnWindowFocus` / `refetchOnMount` options that assume that if a component is mounted, then it's on screen. This PR would allow setting global navigation-aware options on queries to reconcile the two behaviors. Something along the line:
```
import { QueryDefaultOptionsProvider } from '@tanstack/react-query'

export const Screen = () => {
  const navigation = useNavigation()
  return (
    <QueryDefaultOptionsProvider
      options={{
        queries: {
          refetchOnWindowFocus: () => navigation.isFocused(),
        },
      }}
    >
      ...
    </QueryDefaultOptionsProvider>
  )
}
```

### ⚠️ Attention points

- I have reused the same interface as `queryClient.setDefaultOptions`, though this PR only supports `{ queries: ... }` for now, not `{ mutations: ... }` (slightly more work because options defaulting is not completely done in `useMutation` yet, unlike `useBaseQuery`).

- If the component rendering `<QueryDefaultOptionsProvider>` rerenders, all components calling `useQuery` (and alternatives) will rerender because the options object is not referentially stable. The user needs to memoize the options object to be safe. I've added it to the docs, but an alternative would be to expose the context itself and let the user use `QueryDefaultOptionsContext.Provider>` to benefit from this official React eslint rule : https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-constructed-context-values.md

### DIFF
--- a/packages/react-query/src/QueryDefaultOptionsProvider.tsx
+++ b/packages/react-query/src/QueryDefaultOptionsProvider.tsx
@@ -1,0 +1,27 @@
+'use client'
+import * as React from 'react'
+import { type DefaultOptions } from '@tanstack/query-core'
+
+export const QueryDefaultOptionsContext = React.createContext<
+  Pick<DefaultOptions, 'queries'> | undefined
+>(undefined)
+
+export const useQueryDefaultOptions = () => {
+  return React.useContext(QueryDefaultOptionsContext)
+}
+
+export type QueryDefaultOptionsProviderProps = {
+  options: Pick<DefaultOptions, 'queries'>
+  children?: React.ReactNode
+}
+
+export const QueryDefaultOptionsProvider = ({
+  options,
+  children,
+}: QueryDefaultOptionsProviderProps): JSX.Element => {
+  return (
+    <QueryDefaultOptionsContext.Provider value={options}>
+      {children}
+    </QueryDefaultOptionsContext.Provider>
+  )
+}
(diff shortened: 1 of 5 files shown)


packages/react-query/src/QueryDefaultOptionsProvider.tsx
```diff
@@ -0,0 +1,27 @@
+'use client'
+import * as React from 'react'
```
2jordan3, line 2:
```suggestion
import { createContext, useContext } from 'react';
import type { ReactNode, FC } from 'react';
```

Doing it this way makes the code clearer and helps reduce the bundle size by not importing unnecessary things. What do you think about this approach?

TkDodo:
you can already build something like this quite easily in user-land by providing a new QueryClient that has the same QueryCache:

https://tkdodo.eu/blog/testing-react-query#reactqueryconfigprovider

I don't think I want to widen the API surface of the package for this.

PR #7427, opened 2024-05-14
fix(svelte-query): fix create queries get results type

ap0nia:
# Fix
The typeof `QueryOptions.queryFn` is a union between `SkipToken` and `QueryFunction` by default. The `GetResults` type in react-query seems to handle this, but svelte-query currently omits this.


ap0nia:
Accidentally included a commit from a different feature.

PR #7438, opened 2024-05-16
chore: switch to lts node version

Jungzl:


TkDodo:
this doesn't work for me with `nvm use`. Let's please use fixed numbers, but I'm happy to bump to v22.2.0

.nvmrc
```diff
@@ -1 +1 @@
-v18.19.0
+lts/Hydrogen
```
TkDodo, line 1:
```suggestion
v22.2.0
```

Jungzl: what about lts/Iron

TkDodo: what's the advantage please, and how does it work with nvm use? When I put those lts versions in nvmrc and run `nvm use`, I get:

```
TanStackQuery % nvm use
N/A: version "N/A" is not yet installed.

You need to run "nvm install N/A" to install it before using it.
TanStackQuery % 
```

and I can't run `nvm install lts/Iron` either:

```
TanStackQuery % nvm install lts/Iron
Version 'lts/Iron' not found - try `nvm ls-remote` to browse available versions.
```

then I can list the versions and find out which one that is but that makes no sense to me. I'd rather pin a version and update it from time to time.

Jungzl: you can use lowercase
```
nvm install lts/iron
```

lts version covers more rather than fixed version

PR #7441, opened 2024-05-16
fix(query-core): allow function value of `notifyOnChangeProps` query option to return `undefined`

### LINKED ISSUE #7426: Function passed to `notifyOnChangeProps` cannot return `undefined`
#### Background

`notifyOnChangeProps` is a query option that determines if the query should notify listeners if query properties change. It has the following type: https://github.com/TanStack/query/blob/1d60d44ad851472d37aaa3eef22ada07e5245989/packages/query-core/src/types.ts#L144-L148

Depending on the value set, it has the following behaviours:

- `undefined`: The query will track which properties are accessed and only notify listeners if those properties change. This is the default option.
- `string[]`: The query will notify listeners if any of the properties in the array change.
- `'all'`: The query will notify listeners if any properties change.

It can also be set with a function with the signature `() => string[] | 'all'`. The query will run the function and follow the behaviour listed above that matches the returned value.

With the available non-function options of `undefined`, `string[]`, or `'all'`, it follows that the function should also be able to return `undefined`. However, the current type signature does not allow it.

#### Real world example

One real world example of the function passed to `notifyOnChangeProps` returning `undefined` comes from React Query's documentation itself, for [disabling re-renders on out of focus screens on React Native](https://tanstack.com/query/latest/docs/framework/react/react-native#disable-re-renders-on-out-of-focus-screens). See this [TypeScript Playground](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgVwM4FMBiEDGaCiAZoejvAL5yFQQhwDkAAlOgIZkC0AdqwG7ABzVjGAQuAeh4je6egChQkWIjgA5CCMIBPAPJcAwgAtWXAegAKNMKjiVqtBoxgnUznAGtxAR2TooWjhxoWQVwaHgkNHQARV9-WyoaOiZnLld2TxZ2GA4fPy15RXCVKP1WABtygCMMgBoUDAAldEIE+2Sssnk5QmQuMlEuBqxcNHVNXQNjUwsrVAAKOTg4Lg1gbT0jEzNLCGsAfgAuNTWNqe3ZvdRauQBKRCW4ILT4QlGMABNm1oBeYe-5jAoL5bnJHlFsHhUEQSGRFstlqUKtUMvN5vcfgA+B4IhFvKHoL4tAB0eCgLC48D+QN8YNxyxYMGQUCG6LgWJx9OW+KiRMIpOZFKpVAqGEeuPI4ts9QA2gBdW43Zagx6M5msjHYhBS9ZweYAQh5n2+AvJ6Ep921XLVLLg8qlkp1rUBWjA6AgrVWE020x2c3ZPz+9F6-REYnolqlDPQTNtXvWky2M121nRDrpCJtQ3jZyTfquj0ljuerhWp0Tvsu1nZw0hY3LPouKYWKqisXy8ytCLy-kwXGOrFQWn6es1cBp6CV3biWgA0ugtMcZfQe3OF-Q5VPljmK025jdyLcgA) for a reproduction.

#### Impact

Allowing the function passed to `notifyOnChangeProps` to return `undefined` should not have any negative impact. At runtime, the query observer sets the value of `notifyOnChangeProps` or the value of the call to `notifyOnChangeProps` if it was a function to the same variable. This means if `undefined` is a valid value for `notifyOnChangeProps` then a call to `notifyOnChangeProps` should also be allowed to be `undefined`. See: https://github.com/TanStack/query/blob/1d60d44ad851472d37aaa3eef22ada07e5245989/packages/query-core/src/queryObserver.ts#L612-L616

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgVwM4FMCKz1QJ5wC+cAZlBCHAOQACMAhgHaoMDGA1gPRTr2swBaAI458VAFDi0WUXgAUCcXGVwRuPADFGALjj1UeRqzhyAlHAC8APjgwoOADRKVa-AGl0eXQG0qrvB54VAC6TipwjBAwwCR4APKMAMIAFkwA5ugACuRgqLpmljbIjAAm6CTAjOglToSmQA

### Steps to reproduce

Set `notifyOnChangeProps` query option to a function that (could) return `undefined`.

### Expected behavior

No type error should occur.

Every time

### Screenshots or Videos

_No response_

### Platform

N/A

### Tanstack Query adapter

None

### TanStack Query version

5.29.2

### TypeScript version

5.4.5

### Additional context

_No response_

winghouchan:
`undefined` is a valid return value for the function value of `notifyOnChangeProps`. See linked issue for more details.

Fixes #7426.


TkDodo:
can you amend the signature in the docs as well please?

TkDodo:
also the jsdoc please:

https://github.com/TanStack/query/blob/3a8c2b10a1c2a3657d22b0304aab640683a4db0f/packages/query-core/src/types.ts#L319-L326

TkDodo:
@winghouchan will you get this PR over the finish line or should I do it?

winghouchan:
@TkDodo: sorry, missed your messages.

I've updated the signature in documentation in 43f0722. As far as I can tell it only affects React (and Vue, although they share the same docs). Please let me know if this is incorrect.

How would you like the JSDoc updated? The current description isn't wrong, although I suppose it could be improved. Maybe something like below:

> - If set to a function, the function will be executed to compute the list of properties.
> - If the value is set or computed to `undefined`, access to properties will be tracked, and the component will only re-render when one of the tracked properties change. This is the default behaviour.
> - If the value is set or computed to an array, the component will only re-render if any of the listed properties change. For example, if the value is set to or computed to `['data', 'error']`, the component will only re-render when the `data` or `error` properties change.
> - If the value is set or computed to `"all"`, the component will opt-out of smart tracking and re-render whenever a query is updated.

PR #7453, opened 2024-05-18
refactor(svelte-query): specify StoreOrVal options in function signature

ap0nia:
# Summary
Expose the `CreateMutationOptions` as a regular object, and wrap it with `StoreOrVal` in the `createMutation` function signature.

# Reasoning
Aside from being inconvenient if I wanted to refer to the unwrapped mutation options object, it seems to be inconsistent with the other type definitions for query options? `createQuery`, `createInfiniteQuery` both have `StoreOrVal` wrapped around their options in their function signature, and the exposed type for options is just an object.


lachlancollins (approved):
Thanks @ap0nia !

PR #7481, opened 2024-05-26
Feature/hydrate promises

TkDodo:


docs/framework/react/guides/advanced-ssr.md
```diff
@@ -354,9 +354,81 @@ The Next.js app router automatically streams any part of the application that is
...
+'use client'
+
+export default function Posts() {
+  const { data } = useSuspenseQuery({ queryKey: ['posts'], queryFn: getPosts })
+
+  // ...
+}
+```
+
+> Note that you could also `useQuery` instead of `useSuspenseQuery`, and the Promise would still be picked up correctly. However, NextJs won't suspend in that case, so we won't get our component's initial state serialized into HTML.
```
TkDodo, line 429: At least this is what I have observed, does that sound about right?

Note: I also want to get this to work:

```
const { promise } = useQuery()

// further down, potentially conditional
const data = use(promise)
```

I think it should be doable but it's a different bridge to cross

packages/query-core/src/hydration.ts
```diff
@@ -65,6 +66,7 @@ function dehydrateQuery(query: Query): DehydratedQuery {
     state: query.state,
     queryKey: query.queryKey,
     queryHash: query.queryHash,
+    ...(query.state.status === 'pending' && { promise: query.promise }),
```
phryneas, line 69: I'd recommend to "hide" the error here to prevent leaking secrets.
```suggestion
    ...(query.state.status === 'pending' && { promise: query.promise.catch(() => { throw new Error("redacted") }) }),
``` 

You might also want to consider different behaviour on rehydration of errors, depending of environment. I'd probably throw an error in SSR so the suspense boundary would just retry in the browser - and in the browser, I wouldn't throw at all, but immediately retry the query if it failed in RSC.

docs/framework/react/guides/advanced-ssr.md
```diff
@@ -354,9 +354,81 @@ The Next.js app router automatically streams any part of the application that is
 
 With the prefetching patterns described above, React Query is perfectly compatible with this form of streaming. As the data for each Suspense boundary resolves, Next.js can render and stream the finished content to the browser. This works even if you are using `useQuery` as outlined above because the suspending actually happens when you `await` the prefetch.
 
-Note that right now, you have to await all prefetches for this to work. This means all prefetches are considered critical content and will block that Suspense boundary.
+As of React Query v5.40.0, you don't have to `await` all prefetches for this to work, as `pending` Queries can also be dehydrated and sent ot the client. This lets you kick off prefetches as early as possible without letting them block an entire Suspense boundary, and streams the _data_ to the client as the query finishes. This can be useful for example if you want to prefetch some content that is only visible after some user interaction, or say if you want to `await` and render the first page of an infinite query, but start prefetching page 2 without blocking rendering.
```
jorgecolonconsulting, line 357: Minor typo:

```
and sent to the client
```

packages/query-core/src/hydration.ts
TkDodo: thanks. How does this look?

03676ddd

docs/framework/react/guides/advanced-ssr.md
TkDodo: f6beb5d3

packages/query-core/src/hydration.ts
phryneas: Looks good to me :)

shuding: I think in Next.js (or maybe React does that), the error message and stack are already hidden when crossing the RSC boundary. This is also the case for Server Actions.

However, there will be a special `digest` field attached to the error which is a random ID. So developers can log the actual error on the server and use that ID from client logs to match it, without leaking anything.

https://nextjs.org/docs/app/building-your-application/routing/error-handling#securing-sensitive-error-information

TkDodo: if react does this, it means we could remove this error handling. if it's next specific, I don't think we can rely on it because other frameworks like waku might implement this differently.

So can we figure out if this is next or react who does this ?

phryneas: I thought that was only for errors thrown during render, not for transported values, but I might be wrong - double-checking might definitely be a good idea here.

Ephem: Either way, this part is in the query core, so might make sense to redact it by default if entirely other frameworks handle it differently? Would be nice to support the digest too though. 🤔 

Easier to go from redacted -> more details in the future than the other way around so I think current version is fine to ship.

TkDodo: true, I haven't even thought about other frameworks 👍

examples/react/nextjs-app-prefetching/app/get-query-client.ts
```diff
@@ -0,0 +1,33 @@
+import { QueryClient, defaultShouldDehydrateQuery } from '@tanstack/react-query'
+
+function makeQueryClient() {
+  return new QueryClient({
+    defaultOptions: {
+      queries: {
+        staleTime: 60 * 1000,
+      },
+      dehydrate: {
```
Ephem, line 9: Not sure if it's just me, but I'm getting a TS error here:

```Object literal may only specify known properties, and 'dehydrate' does not exist in type 'DefaultOptions<Error>'.```

**Review by Ephem — COMMENTED** [review]
I think this looks great! I left mostly NITs.

The one thing I'm a bit unsure of, but don't know if it affects this PR, is exactly under which circumstances NextJs makes the shell static. In the current example, I think the shell is supposed to be static with _or_ without `await` because the `fetch` is cached by default?

When I add `{ cache: "no-store" }` to that fetch, it turns the page dynamic both with and without `await`. This seems to happen in the swr example too though, so not sure what the expected behaviour is.

File: packages/query-core/src/__tests__/hydration.test.tsx [hunk 5]
```diff
@@ -1,4 +1,5 @@
 import { describe, expect, test, vi } from 'vitest'
+import { waitFor } from '@testing-library/react'
```
sheremet-va, line 2: Vitest supports `vi.waitFor` 👀

docs/framework/react/guides/advanced-ssr.md
```diff
@@ -354,9 +354,81 @@ The Next.js app router automatically streams any part of the application that is
...
+    defaultOptions: {
+      queries: {
+        staleTime: 60 * 1000,
+      },
+      dehydrate: {
+        // include pending Queries in dehydration
+        // per default, only successful Queries are included
+        shouldDehydrateQuery: (query) =>
+          defaultShouldDehydrateQuery(query) ||
+          query.state.status === 'pending',
```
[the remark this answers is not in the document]
TkDodo: to make it globally available, I had to add it to the `defaultOptions`. But I think that's a good addition

docs/framework/react/guides/advanced-ssr.md
```diff
@@ -417,7 +489,7 @@ export function Providers(props: { children: React.ReactNode }) {
     </QueryClientProvider>
   )
 }
-```
+````
```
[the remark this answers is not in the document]
TkDodo: thanks, I was wondering why my editor looked broken 🙈

examples/react/nextjs-app-prefetching/app/get-query-client.ts
TkDodo: you need to build the `query-core` package locally to make this type correctly:

```
npx nx run @tanstack/query-core:build
```

docs/framework/react/guides/advanced-ssr.md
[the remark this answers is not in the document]
TkDodo: d0781793

packages/query-core/src/hydration.ts
```diff
@@ -65,6 +66,16 @@ function dehydrateQuery(query: Query): DehydratedQuery {
     state: query.state,
     queryKey: query.queryKey,
     queryHash: query.queryHash,
+    ...(query.state.status === 'pending' && {
+      promise: query.promise?.catch((error) => {
+        if (process.env.NODE_ENV !== 'production') {
+          console.error(
+            `An error occurred while dehydrating pending query [${query.queryHash}]: ${error}; The error will be redacted in production builds`,
```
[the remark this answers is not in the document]
TkDodo: 34bd169a

docs/framework/react/guides/advanced-ssr.md
TkDodo: 6d3389a8

what I found interesting is that the example only works without JavaScript when `useSuspenseQuery` is used. With `useQuery`, the html output is just the `loading...` text. That makes sense because we never suspend and we never get the html markup, I think that's an important distinction though, just don't know how to phrase it ...

Ephem: Yeah, my thought was that this was implied by the "rendering in the pending status" part, but now that I think about it, that's probably not clear enough. Maybe something like this?

```> Note that you could also `useQuery` instead of `useSuspenseQuery`, and the Promise would still be picked up correctly. However, NextJs won't suspend in that case and the component will render in the `pending` status, which also opts out of server rendering the content.```

examples/react/nextjs-app-prefetching/app/get-query-client.ts
Ephem: Of course I do.. 🤦 Thanks, that solved it of course.

docs/framework/react/guides/advanced-ssr.md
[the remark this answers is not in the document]
Ephem: I'm torn myself, I kind of like how clear this makes it what is happening too. 😄

docs/framework/react/guides/advanced-ssr.md
TkDodo: sold 🤝 

65474238

docs/framework/react/guides/advanced-ssr.md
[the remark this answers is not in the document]
TkDodo: I like the idea of maybe adding a new RSC default later if we need to 👍

TmLev:
This is awesome, thank you!

packages/query-core/src/hydration.ts
shuding: @TkDodo @Ephem I confirmed that it's a React behavior (which makes sense, as it's done on the RSC encoding/decoding layer that frameworks can't inject into): https://github.com/facebook/react/blob/ea6e05912aa43a0bbfbee381752caa1817a41a86/packages/react-client/src/ReactFlightClient.js#L1477-L1484

If we redact it here, the client code (e.g. error boundaries, alert boxes, trackers) can't receive that digest value and get it displayed/logged.

Issue #7446, opened 2024-05-17
Unable to pass queryOptions to QueriesObserver (TypeScript error)


yannickcr:
### Describe the bug

When passing options created with the `queryOptions` helper to `QueriesObserver` we got the following TypeScript error as it is unable to assign `UseQueryOptions` to `QueryObserverOptions`:

```ts
Type 'UseQueryOptions<string, Error, string, string[]> & { initialData?: undefined; } & { queryKey: DataTag<string[], string>; }' is not assignable to type 'QueryObserverOptions<unknown, Error, unknown, unknown, QueryKey, never>'.
  Types of property 'refetchInterval' are incompatible.
    Type 'number | false | ((query: Query<string, Error, string, string[]>) => number | false | undefined) | undefined' is not assignable to type 'number | false | ((query: Query<unknown, Error, unknown, QueryKey>) => number | false | undefined) | undefined'.
      Type '(query: Query<string, Error, string, string[]>) => number | false | undefined' is not assignable to type 'number | false | ((query: Query<unknown, Error, unknown, QueryKey>) => number | false | undefined) | undefined'.
        Type '(query: Query<string, Error, string, string[]>) => number | false | undefined' is not assignable to type '(query: Query<unknown, Error, unknown, QueryKey>) => number | false | undefined'.
          Types of parameters 'query' and 'query' are incompatible.
            Type 'Query<unknown, Error, unknown, QueryKey>' is not assignable to type 'Query<string, Error, string, string[]>'.
              Types of property 'queryKey' are incompatible.
                The type 'QueryKey' is 'readonly' and cannot be assigned to the mutable type 'string[]'.
```

### Your minimal, reproducible example

[...]

PR #7492, opened 2024-05-29
fix(query-core): widen QueriesObserver queries type (#7446)

yannickcr:
Widen the `QueriesObserver` queries type so we can assign `UseQueryOptions` to `QueryObserverOptions`.

The result from the `queryOptions` helper can now be passed to `QueriesObserver`.

This align the `QueriesObserver` behavior with `QueryObserver` (that already accepts `UseQueryOptions`).

**Caveat:** The return type from `QueriesObserver` will be `QueriesObserver<QueryObserverResult[]>`, it means that the `QueryObserverResult` data type will be `unknown`. As discussed in https://github.com/TanStack/query/issues/7446#issuecomment-2120125066 it has been decided that it was an acceptable compromise for now.


TkDodo:
can you add a type test for your situation please?

TkDodo:
prettier fails now 😭

PR #7501, opened 2024-06-01
feat(react-query): add SuspenseQuery

manudeli:
I made this component early in @suspensive/react-query to prove why we need this component to use @tanstack/react-query with suspense easily. Check this link https://suspensive.org/docs/react-query/SuspenseQuery please.

# SuspenseQuery

I want to add this SuspenseQuery component to clearly express what causes suspense at the same depth with Suspense, ErrorBoundary

## Expected Effect

1. Prop-driling resulting from removing depth such as AuthorInfo and PostList for data-fetching only is also removed.
2. Changing the range of Suspense and ErrorBoundary becomes simple. Parallel processing of queries is also easier.
3. Because it manages all data-fetching within the Page component, the internal components are presentational, so it is easy to separate the components.

```jsx /SuspenseQuery/
import { SuspenseQuery } from '@tanstack/react-query'
import { Suspense, ErrorBoundary } from '@suspensive/react'
import { PostItem, AuthorProfile } from '~/components'

const PostsPage = ({ authorId }) => (
  <ErrorBoundary fallback={({ error }) => <>{error.message}</>}>
    <Suspense fallback={'loading...'}>
      <SuspenseQuery {...userQueryOptions(authorId)}>
        {({ data: author }) => <AuthorProfile {...author} />}
      </SuspenseQuery>
      <SuspenseQuery {...postsQueryOptions(authorId)} select={(posts) => posts.filter(({ isShow }) => isShow)}>
        {({ data: posts }) => posts.map((post) => <PostListItem {...post} />)}
      </SuspenseQuery>
    </Suspense>
  </ErrorBoundary>
)
```

## Motivation: useSuspenseQuery is not obvious

Because the existing useSuspenseQuery is a hook, it creates components with names such as AuthorInfo and PostList to place Suspense and ErrorBoundary on the parent.
This makes it difficult to predict what suspense and errors will be thrown inside AuthorInfo and PostList.

```jsx
// posts/page.tsx
import { Suspense, ErrorBoundary } from '@suspensive/react'
import { AuthorInfo } from './components/AuthorInfo'
import { PostList } from './components/PostList'

const PostsPage = ({ userId }) => (
  <ErrorBoundary fallback={({ error }) => <>{error.message}</>}>
    <Suspense fallback={'loading...'}>
      <AuthorInfo userId={userId} /> {/* It is difficult to predict whether a suspension will occur internally. */}
      <PostList userId={userId} /> {/* It is difficult to predict whether a suspension will occur internally. */}
    </Suspense>
  </ErrorBoundary>
)
```

```jsx /useSuspenseQuery/
// posts/components/AuthorInfo.tsx
import { useSuspenseQuery } from '@tanstack/react-query'
import { AuthorProfile } from '~/components'

// From the perspective of using this component, it is impossible to predict whether Suspense will occur internally just by the name AuthorInfo.
const AuthorInfo = ({ userId }) => {
  // We need to create this component just for data-fetching.
  const { data: author } = useSuspenseQuery(userQueryOptions(userId))

  return <AuthorProfile {...author} />
}
```

```jsx /useSuspenseQuery/
// posts/components/PostList.tsx
import { useSuspenseQuery } from '@tanstack/react-query'
import { PostListItem } from '~/components'

// From the perspective of using this component, it is impossible to predict whether a suspense will occur internally based on the name PostList alone.
const PostList = ({ userId }) => {
   // We need to create this component just for data-fetching.
   const { data: posts } = useSuspenseQuery({
     ...postsQueryOptions(userId);
     select: (posts) => posts.filter(({ isShow }) => isShow),
   })

   return (
     <>
       {posts.map((post) => (
         <PostListItem {...post} />
       ))}
     </>
   )
}
```

## etc.

If this is accepted, I want to make docs for SuspenseQuery and SuspenseInfiniteQuery component too.


TkDodo:
I think this is a very subjective reason. For example, I find it easier to have a component called `< AuthorInfo />` that "reads" its own dependencies internally so that I don't have to think about what it does - it's an implementation detail. I can also freely move it around the app when using a router that integrates with suspense (like TanStack router does).

Also, the implementation is really a one-liner, and we've always said that for things that are a) situationally good but not necessarily obviously better and b) can be easily done in user-land, we won't provide APIs because it's also a matter of keeping the API surface lean.

PR #7502, opened 2024-06-01
fix(react-query): update skipToken error message condition

manudeli:
In my opinion, to express development mode console error message, we should use `process.env.NODE_ENV === 'development'` to express development mode. because @tanstack/react-query-devtools also use `process.env.NODE_ENV === 'development'` and additionally, even for production app, process.env.NODE_ENV !== 'production' can be occurred. so I want to minimize opportunity to make this error message like @tanstack/react-query-devtools
https://github.com/TanStack/query/blob/main/packages/react-query-devtools/src/index.ts#L6


TkDodo:
all places in query that want to report user config errors use `process.env.NODE_ENV !== 'production'` checks, for example:

https://github.com/TanStack/query/blob/e5f23fcea27171a17e903a66b7b2ca90c557815b/packages/react-query/src/useBaseQuery.ts#L39-L45

This is on purpose to also see these errors in test environments. For example, you might have your test runner configured to fail on console errors, so these messages are helpful there.

PR #7503, opened 2024-06-01
ci(codecov): update config

manudeli:
ref: https://github.com/TanStack/query/pull/7502

image

I update codecov coverage threshold. because all pr could be as error in GitHub workflow checks


arnoud-dv:
I don't know why codecov is suddenly failing, but cherry picking this commit in [my PR](https://github.com/TanStack/query/pull/7504) still results in a codecov error.

arnoud-dv:
For some reason codecov shows a commit from 3 months ago as base instead of the last commit on main.

arnoud-dv:
There is this error message for codecov uploads on main ci:

```
info - 2024-05-27 18:16:24,691 -- ci service found: github-actions
Error: Codecov token not found. Please provide Codecov token with -t flag.
Warning: Codecov: Failed to properly create commit: The process '/home/runner/work/_actions/codecov/codecov-action/v4/dist/codecov' failed with exit code 1
```

arnoud-dv:
I think this can only be fixed by using a token to upload the reports.

TkDodo:
new baseline is uploaded correctly, but why don't we see the CI check here or a comment from the bot?

TkDodo:
also, different thresholds per `package` would be really appreciated. We have vastly different coverage per package

arnoud-dv:
> also, different thresholds per `package` would be really appreciated. We have vastly different coverage per package

We could make it [informational](https://docs.codecov.com/docs/quick-start#tips-and-tricks)

> 2. Not ready to start using blocking status checks? Set them as informational while you and your team get started with code coverage.

codecov.yml
```diff
@@ -2,11 +2,17 @@ coverage:
...
       default:
-        target: 90%
+        target: auto
         threshold: 1%
+        base: auto
 
 comment:
   layout: 'header, reach, diff, flags, components'
+  behavior: default
+  require_changes: false
```
manudeli, line 12: > new baseline is uploaded correctly, but why don't we see the CI check here or a comment from the bot?

`comment.require_changes: false` should force make comment from the bot without changes

https://docs.codecov.com/docs/quick-start#tips-and-tricks
image

codecov.yml
```diff
@@ -2,11 +2,17 @@ coverage:
   status:
     project:
       default:
-        target: 90%
+        target: auto
```
manudeli, line 5: https://docs.codecov.com/docs/quick-start#tips-and-tricks

`status.project.default.target: auto` compares coverage to the previous base commit

image

manudeli (commented):
@arnoud-dv @TkDodo I updated some codes, In my opinion, we can use this config. confirm it please.

codecov.yml
manudeli: I think this(https://docs.codecov.com/docs/common-recipe-list#isolate-my-coverage) is good to isolate target coverage of each packages too.

TkDodo (approved):

PR #7499, opened 2024-05-31
Update community-projects.md - Add Rapini

rametta:
Adding a popular community project that has been around for a few years now. Didn't realize this page existed until today, so adding it

PR #7508, opened 2024-06-01
chore(.vscode): force format on save by prettier without husky install

manudeli:
Because of https://github.com/TanStack/query/pull/7499#issuecomment-2143561625

I knew you(@TkDodo) don't use VSCode but as you know really many people use VSCode. If someone contributes with a simple modification to the official document, they can make a Pull Request without having to format the prettier with the help of husky, since they do it without even installing pnpm. I hope this change will lower maintenance costs by preventing cases like this early on.


TkDodo:
I think having the precommit hook and the CI checks should be enough. I don't think we should check in editor specific settings, as it might lead to people accidentally committing their personal settings and then PRs will be even slower because we need to remind them to _not_ do that.

PR #7516, opened 2024-06-04
chore(angular-query): update to angular 18

arnoud-dv:
Also updates TypeScript to 5.4.5


arnoud-dv:
Closing as it's not needed to support Angular 18 and libraries built on newer Angular versions are not officially supported on older applications.

[Ensuring library version compatibility](https://angular.dev/tools/libraries/creating-libraries#ensuring-library-version-compatibility)

> The Angular version used to build an application should always be the same or greater than the Angular versions used to build any of its dependent libraries. For example, if you had a library using Angular version 13, the application that depends on that library should use Angular version 13 or later. Angular does not support using an earlier version for the application.

Issue #7513, opened 2024-06-04
Bug: Dependent Query doesn't work with solid-js SSR


BierDav:
### Describe the bug

In the Tanstack documentation there is a section about dependent queries: https://tanstack.com/query/latest/docs/framework/react/guides/dependent-queries

The example suggest that you should use `enable` to create dependent queries. The problem when using `enable` is that when it is disabled it doesn't suspend the current context and therefore doesn't fetch on the server.

### Your minimal, reproducible example

https://codesandbox.io/p/github/BierDav/temp-tanstack-query-ssr-bug/master?import=true

### Steps to reproduce

1. Open codesandbox
2. Run ´pnpm dev`
3. Open the preview
4. Now the server waits for 2 seconds before responding because of the `timout` query in `src/routs/index.tsx`, but the moment it finishes it doesn't care about the second query getting enabed and returning it immediately to the browser.
5. This causes that the client has to fetch the remaining `state` query. You can see that in the dev tools:
![image](https://github.com/TanStack/query/assets/46626041/cb7f79e3-5ff4-4b62-bfee-b41381a59202)

### Expected behavior

The expected behavior would be that both queries are beeing evaluated before the server sends its initial response. But I don't know if this is even possible with this syntax from a technical perspective.

### How often does this bug happen?

Every time

### Screenshots or Videos

```tsx
import { createQuery } from "@tanstack/solid-query";

export default function () {
  const timeout = createQuery(() => ({
    queryKey: ["timeout"],
    queryFn: () =>
      new Promise((resolve) => setTimeout(() => resolve("test"), 2000)),
  }));

  const state = createQuery(() => ({
    queryKey: ["realQuery"],
    queryFn: async () => {
      const resp = await fetch("https://api.sampleapis.com/switch/games");
      console.log("fetched");
      return await resp.json();
    },
    enabled: !!timeout.data,
  }));

  return <>{JSON.stringify(state.data?.[0])}</>;
}
```

### Platform

- Windows
- Node v21.7.3
- Chrome

### Tanstack Query adapter

solid-query

### TanStack Query version

v5.40.0

TkDodo:
@ardeora FYI

TkDodo:
I'm not sure about solid but in react, suspense runs in serial per component, which is why `enabled` makes no sense. So `enabled: !!timeout.data,` is not necessary because `timeout.data` would always be defined. `useSuspenseQuery` also has this built-in on type level, but again, I don't know enough how solid works in this regard.

[...]

PR #7528, opened 2024-06-06
Bugfix for #7527

### LINKED ISSUE #7527: Bug: Error when hydrating an query that threw error
### Describe the bug

There is a property called `retryOnMount` which is set by default. This property tells tanstack to reevaluate a query like this, immediately after hydration:
```ts
   const errorQuery = createQuery(() => ({
        queryKey: ["errorQuery"],
        queryFn: () => {
            throw new Error("test")
        }
    }))
```
The problem is that the underlying code uses `queueMicrotask` to refetch which results in the refresh beeing executed to early i guess. This results in following error:
```
ReferenceError: Cannot access 'refetch' before initialization
    at dev.js:145:28
```

https://codesandbox.io/p/github/BierDav/temp-tanstack-query-ssr-bug/hydration-of-error-query

### Steps to reproduce

1. Run `pnpm dev`
2. Open preview

### Expected behavior

The refetch should not throw an error when executed on hydration.

Every time

### Screenshots or Videos

image

### Platform

- OS: Windows
- Browser: Chrome
- Node: v21.7.3

### Tanstack Query adapter

solid-query

### TanStack Query version

5.40.0

### TypeScript version

_No response_

### Additional context

Actually I'm not sure why this error also

I've also tried another variant of this issue with a timeout in the query: [https://codesandbox.io/p/github/BierDav/temp-tanstack-query-ssr-bug/hydration-of-error-query-timeout](https://codesandbox.io/p/github/BierDav/temp-tanstack-query-ssr-bug/hydration-of-error-query-timeout)

```ts
 const errorQuery = createQuery(() => ({
        queryKey: ["errorQuery"],
        queryFn: () => new Promise<string>((_, reject) => {
            setTimeout(() => {
                reject(new Error("test"))
            }, 1000)
        })
    }))
```

This results in a hydration error which it also shouldn't:

image

Actually I'm not sure why this error even occurs, because when you set: 
```
new QueryClient({
    defaultOptions: {
        queries: {
            retryOnMount: false,
            refetchOnMount: false
        }
    }
}
```

The same error occurs in both two variants. Even though I assumed that this should prevent the library from even thinking about refetching the query.

BierDav:
When the createResource is rejected on ssr evaluation `v` is `undefined` which is not handled. I actually don't really understand what  `v` is.

I don't want to blame here any good open source soul, but just wanted to metion in a lovely manner that naming things in a verbose way makes the code 100% more readable for other developers :)

Actually I don't know if this is correct way to approach this issue, it could be that there is an underlying problem that causes `v` to be `undefined`

fixes #7527

### DIFF
--- a/packages/solid-query/src/createBaseQuery.ts
+++ b/packages/solid-query/src/createBaseQuery.ts
@@ -208,7 +208,7 @@ export function createBaseQuery<
         }
         // Hydration data exists on first load after SSR,
         // and should be removed from the observer result
-        if (v.hydrationData) {
+        if (v?.hydrationData) {
           const { hydrationData, ...rest } = v
           v = rest
         }


TkDodo:
can you please add a test that covers this scenario?

BierDav:
I can definetly try, but currently I have no idea how to achieve that.

Btw. would be very kind if this issue #7513 gets fixed very soon :)

BierDav:
I'm not sure how to achive hydration testing in solidjs. When I just use this simple example:
```ts
 function Page() {
      return (
        <div>
        </div>
      )
    }


    const rendered = render(() => <Page/>, {hydrate: true})
```

The vitest failes with this error message:
```
TypeError: Cannot read properties of undefined (reading 'completed')
    at hydrate$1 (file:///C:/Users/David%20Bieregger/WebstormProjects/query/node_modules/.pnpm/solid-js@1.8.17/node_modules/solid-js/web/dist/dev.js:726:44)
    at hydrate (file:///C:/Users/David%20Bieregger/WebstormProjects/query/node_modules/.pnpm/solid-js@1.8.17/node_modules/solid-js/web/dist/dev.js:1062:10)
    at Module.render (file:///C:/Users/David%20Bieregger/WebstormProjects/query/node_modules/.pnpm/@solidjs+testing-library@0.8.8_@solidjs+router@0.13.3_solid-js@1.8.17__solid-js@1.8.17/node_modules/@solidjs/testing-library/dist/index.js:55:29)
    at C:\Users\David Bieregger\WebstormProjects\query\packages\solid-query\src\__tests__\createQuery.test.tsx:5925:22
    at file:///C:/Users/David%20Bieregger/WebstormProjects/query/node_modules/.pnpm/@vitest+runner@1.6.0/node_modules/@vitest/runner/dist/index.js:135:14
    at file:///C:/Users/David%20Bieregger/WebstormProjects/query/node_modules/.pnpm/@vitest+runner@1.6.0/node_modules/@vitest/runner/dist/index.js:60:26
    at runTest (file:///C:/Users/David%20Bieregger/WebstormProjects/query/node_modules/.pnpm/@vitest+runner@1.6.0/node_modules/@vitest/runner/dist/index.js:781:17)
    at runSuite (file:///C:/Users/David%20Bieregger/WebstormProjects/query/node_modules/.pnpm/@vitest+runner@1.6.0/node_modules/@vitest/runner/dist/index.js:909:15)
    at runSuite (file:///C:/Users/David%20Bieregger/WebstormProjects/query/node_modules/.pnpm/@vitest+runner@1.6.0/node_modules/@vitest/runner/dist/index.js:909:15)
    at runFiles (file:///C:/Users/David%20Bieregger/WebstormProjects/query/node_modules/.pnpm/@vitest+runner@1.6.0/node_modules/@vitest/runner/dist/index.js:958:5)
```

ardeora (approved):
Hello yeah this change looks good here

ardeora:
Thanks @BierDav for the feedback! I'll try to rename the variables for better readability. 

Also @TkDodo Adding a test suite for Solid Query SSR is proving to be a challenging task. Solid testing library either allows running the tests in either in a ssr or a client environment. So testing the correct hydration of SSR data into the client environment is difficult to achieve unfortunately. I have multiple stale branches trying to get this to work but I end up in roadblocks every now and then. I'll revisit this again someday.

PR #7538, opened 2024-06-08
feat: hydrate.transformPromise

juliusmarminge:
Allows for deserialization of prefetched promises


packages/query-core/src/hydration.ts
```diff
@@ -178,9 +183,18 @@ export function hydrate(
...
 
     if (promise) {
+      const deserialize =
+        client.getDefaultOptions().hydrate?.deserialize ?? defaultDeserialize
+
+      const initialPromise = (async () => {
+        const result = await promise
+        const deserialized = deserialize(result)
+        return deserialized
+      })()
```
juliusmarminge, line 193: for some reason i don't understand 

```ts
const initialPromise = promise.then((result) => deserialize(result))
```

results in a `Missing queryFn` error but this IIFE approach does work

packages/query-core/src/hydration.ts
```diff
@@ -88,6 +89,10 @@ export function defaultShouldDehydrateQuery(query: Query) {
   return query.state.status === 'success'
 }
 
+export function defaultDeserialize(data: unknown) {
+  return data
+}
```
juliusmarminge, line 94: didn't find any existing `identity` functions when i did a quick search. lmk if you have one somewhere already

packages/query-core/src/hydration.ts
```diff
@@ -21,6 +21,7 @@ export interface DehydrateOptions {
 
 export interface HydrateOptions {
   defaultOptions?: {
+    deserialize?: (data: any) => any
```
juliusmarminge, line 24: maybe `deserializePromise` or something would be better since this only runs on promises?

TkDodo (commented):
I think I'm missing some context here. Is this to make promise deserialization work outside of next-js? what about the serialize counterpart?

Also, it seems like the use-case is to be able to serialize a `date` - so how come this is promise related only?

juliusmarminge:
We serialize the data in the .then of the prefetch: https://github.com/trpc/trpc/pull/5451/files#diff-bd0e8d8a58b08332a96dfcd2a4ba279133c7aca91275ec9894dcc9b1780e61eeR93

So we need to deserialize it before the data gets put in the query cache and this felt like the place to do it

The Promise can only contain valid json props when it crosses the boundary so that's why the data needs to be serialized/deserialize on each side of the border

TkDodo:
I see. how about `transformPromise: Promise<A> => Promise<B>`

so that, instead of:

```
deserialize: data => data
```

you'd do:

```
transformPromise: promise => promise.then(deserialize)
```

?

juliusmarminge:
> I see. how about `transformPromise: Promise<A> => Promise<B>`
> 
> so that, instead of:
> 
> ```
> deserialize: data => data
> ```
> 
> you'd do:
> 
> ```
> transformPromise: promise => promise.then(deserialize)
> ```
> 
> ?

as mentioned here I didn't get `.then()` to work:

https://github.com/TanStack/query/pull/7538#discussion_r1632113164

packages/query-core/src/hydration.ts
TkDodo: I think we should try to figure out why that is. There shouldn't be any difference between `promise.then` and this code.

Can you add a test case and then we can maybe look into why this is failing

packages/query-core/src/hydration.ts
TkDodo: don't think we have one

packages/query-core/src/hydration.ts
juliusmarminge: where is this tested? e2e?

juliusmarminge: not sure how accurate that is since there's no RSC transport there but can give it a go in a bit

docs/framework/react/guides/advanced-ssr.md
```diff
@@ -431,6 +431,65 @@ export default function Posts() {
...
+
+  return (
+    <HydrationBoundary state={dehydrate(queryClient)}>
+      <Posts />
+    </HydrationBoundary>
+  )
+}
+
+// app/posts/posts.tsx
+;('use client')
```
TkDodo, line 482: this is due to prettier I guess 🙈 . Maybe make a separate codeblock?

docs/framework/react/guides/advanced-ssr.md
```diff
@@ -431,6 +431,65 @@ export default function Posts() {
...
+import { serialize } from './transformer'
+import Posts from './posts'
+
+export default function PostsPage() {
+  const queryClient = new QueryClient()
+
+  // look ma, no await
+  queryClient.prefetchQuery({
+    queryKey: ['posts'],
+    queryFn: getPosts.then(serialize), // <-- serilize the data on the server
```
TkDodo, line 471: I think this needs to be:

```suggestion
    queryFn: () => getPosts().then(serialize), // <-- serilize the data on the server
```

packages/query-core/src/hydration.ts
```diff
@@ -178,9 +183,17 @@ export function hydrate(
...
 
     if (promise) {
+      const transformPromise =
+        client.getDefaultOptions().hydrate?.transformPromise ??
+        defaultTransformPromise
+
+      // Note: `Promise.resolve` required cause
+      // RSC transformed promises are not thenable
+      const initialPromise = transformPromise(Promise.resolve(promise))
+
```
TkDodo, line 193: as discussed in Discord, let's change this to:

```
const transformPromise = client.getDefaultOptions.hydrate?.transformPromise

// Note: `Promise.resolve` required cause
// RSC transformed promises are not thenable
const initialPromise = transformPromise?.(Promise.resolve(promise)) ?? promise
```

That also means we don't need `defaultTransformPromise`

packages/query-core/src/hydration.ts
```diff
@@ -21,6 +21,7 @@ export interface DehydrateOptions {
 
 export interface HydrateOptions {
   defaultOptions?: {
+    transformPromise?: (promise: Promise<any>) => Promise<any>
```
TkDodo, line 24: can we set this to unknown?

```suggestion
    transformPromise?: (promise: Promise<unknown>) => Promise<unknown>
```

juliusmarminge: could, although it messes with some serializers expected types so the user would have to `as any` it

CleanShot 2024-06-11 at 11 08 10@2x

packages/query-core/src/hydration.ts
```diff
@@ -178,9 +179,17 @@ export function hydrate(
...
 
     if (promise) {
+      const { transformPromise } = client.getDefaultOptions().hydrate ?? {}
+
+      // Note: `Promise.resolve` required cause
+      // RSC transformed promises are not thenable
+      const initialPromise = transformPromise
+        ? transformPromise(Promise.resolve(promise))
+        : promise
+
```
TkDodo, line 189: not a bit fan of using `{}` as a fallback because it might widen the type to `| {}`. We could just keep it undefined and then use optional chaining as well:

```suggestion
      const transformPromise =
        client.getDefaultOptions().hydrate?.transformPromise

      // Note: `Promise.resolve` required cause
      // RSC transformed promises are not thenable
      const initialPromise =
        transformPromise?.(Promise.resolve(promise)) ?? promise

```

packages/query-core/src/hydration.ts
TkDodo: hm okay

juliusmarminge: could be accepted to do this - atleast then you're explicit that you're not certain that this data is serialized

CleanShot 2024-06-11 at 11 29 27@2x

TkDodo (approved):

PR #7549, opened 2024-06-11
fix: examples/react/basic-graphql-request/src/index.jsx

sunil-sharma-999:
Fix Bad argument type for Post query by converting args into Object form.


lachlancollins:
Hi @sunil-sharma-999 , I think this was fixed in #7704 , please open a new PR if there are any other necessary fixes!

PR #7566, opened 2024-06-14
Add possibility to pass a callback to enabled.

lotusjohn:
We are using tanstack query in react native, and i have tried to implement a way to refetch all active queries in the focused screen. 

In react native all screens remains mounted even when navigating to a new screen, and thus its easy to accidentally re-render out of focus screens, as well as accidentally refetching out of focus queries. 

notifyOnChange props solved the problem with re-renders.
This helpes to be able to use invalidateQueries in for example pull to refresh when we dont want to have to specify exactly which queries should or should not be refetched on invalidation.

As it is today, we are forced to trigger re-renders when navigation to set enabled to true or false to be able to do this, but its costly to rerender screens on navigation and especially the ones going out of focus

Related discussion and PR
https://github.com/TanStack/query/pull/5734

@paulobmarcos @TkDodo


flodlc:
My concern is whether it triggers a refetch as enabled: false/true does. 
Perhaps a programmatic `setEnabled(state: boolean)` call could achieve this? It would follow the same pattern as focusManager but be scoped to a single query.

TkDodo (commented):
I'm not against this, I just think I tried this some time ago and it wasn't that trivial.

Usually, we'd want those functions to accept the `query` itself, and not just be a function without an argument. I recently did this for `staleTime` and put that into a shared utils:

https://github.com/TanStack/query/blob/96a743eec7a7a234705d69d885856c7392c3122e/packages/query-core/src/utils.ts#L91-L101

can you try to refactor this into the same direction?

packages/query-core/src/__tests__/queryObserver.test.tsx
```diff
@@ -52,6 +52,51 @@ describe('queryObserver', () => {
     unsubscribe()
   })
 
+  test('Should not refetch when enabled is a callback and returns false', async () => {
```
TkDodo, line 55: can we add a test please that checks if the query runs when enabled is a function that returns false?

TkDodo:
can you also add updates to the api docs please ?

packages/query-core/src/__tests__/queryObserver.test.tsx
lotusjohn: Updated and added a few more tests.

lotusjohn:
> can you also add updates to the api docs please ?
Yes, updated docs as well now.

packages/query-core/src/__tests__/queryObserver.test.tsx
```diff
@@ -52,6 +52,173 @@ describe('queryObserver', () => {
     unsubscribe()
   })
 
+  describe('enabled is a callback that initially returns false and', () => {
```
TkDodo, line 55: and 😂  ?

TkDodo (approved):
looks really good 👏

TkDodo:
ah, just noticed:
- api docs needs an update here: https://github.com/TanStack/query/blob/96a743eec7a7a234705d69d885856c7392c3122e/docs/framework/react/reference/useQuery.md#L73-L75
- and jsdocs here: https://github.com/TanStack/query/blob/5bda18f37968c57ef8c5b30c21714b0cd19763e0/packages/query-core/src/types.ts#L255-L260

packages/query-core/src/__tests__/queryObserver.test.tsx
lotusjohn: Haha, fixed ;)

lotusjohn:
> ah, just noticed:
> 
> * api docs needs an update here: https://github.com/TanStack/query/blob/96a743eec7a7a234705d69d885856c7392c3122e/docs/framework/react/reference/useQuery.md#L73-L75
> * and jsdocs here: https://github.com/TanStack/query/blob/5bda18f37968c57ef8c5b30c21714b0cd19763e0/packages/query-core/src/types.ts#L255-L260

Fixed! Does that look good?

TkDodo:
eslint is failing - please have a look

https://cloud.nx.app/runs/O9IbYCGg8Y?utm_source=pull-request&utm_medium=comment

lotusjohn:
> eslint is failing - please have a look
> 
> https://cloud.nx.app/runs/O9IbYCGg8Y?utm_source=pull-request&utm_medium=comment

Thanks for letting me know! Pushed some lint fixes now.

PR #7567, opened 2024-06-14
fix(query-core): Change to prevent `undefined` from overriding Mutation default side effects.

mitchell-up:
## Problem
`undefined` means a variable that has not been assigned. Therefore, when the `onError` property in `useMutation` receives `undefined`, the side effects of the default options are retained and are expected to be invoked, as in the below example. However, It overrides default `onError` to `undefined`.

```ts
// You can pass `onError` callback optionally.
// If you intend to retain default `onError` callback,
// you will invoke this hook without any parameters.
const useLogin = (onError?) => {
 return useMutation({
    mutationFn: login,
    onError: onError,
    meta: {
      showToast: true,
    },
  })
}
```

```ts
// But if the useLogin is called without its parameters,
// default `onError` options will not be triggered.
new QueryClient({
  defaultOptions: {
    // ...
    mutations: {
      onError(error) {
        if (isHttpError(error)) {
          defaultErrorHandlers(error)
        }
      },
    },
  },
}),
```

## Suggestion
To clarify the intention of not using side effects of default options, assign `null` to the side effects. Because `null` means absence of any value, you can predict that the side effects will not be invoked.

```ts
const useLogin = (onError?) => {
 return useMutation({
    mutationFn: login,
    onError: onError,
    meta: {
      showToast: true,
    },
  })
}

// This `mutate` call will not trigger any `onError` side effects, including the default `onError`.
const { mutate } = useLogin(null)

// This `mutate` call will not trigger the default `onError`, but `errorHandler` will be called.
const { mutate } = useLogin(errorHandler)

// These `mutate` calls will trigger the default `onError`.
const { mutate } = useLogin()
const { mutate } = useLogin(undefined)
```

### DIFF
--- a/packages/query-core/src/queryClient.ts
+++ b/packages/query-core/src/queryClient.ts
@@ -580,7 +580,7 @@ export class QueryClient {
       ...this.#defaultOptions.mutations,
       ...(options?.mutationKey &&
         this.getMutationDefaults(options.mutationKey)),
-      ...options,
+      ...extractOverrideMutationOptions(options),
       _defaulted: true,
     } as T
   }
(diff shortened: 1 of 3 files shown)


TkDodo:
we don’t do this anywhere with our options so we’re not gonna make an exception here, sory.

Issue #7330, opened 2024-04-24
queryFn with server action name not working on Next.js


tsuyuni:
### Describe the bug

I'm using useQuery hook with my Next.js 14.1.3 project. I'd like to use server action as query function, so I made my codes like this.

actions/fetchTasks.ts
```ts
"use server"

const fetchTasks = async () => {
  try {
    const { data, errors } = client.graphql({ // I'm using AWS Amplify Graphql client
      query: listTasks,
      variables: {
        filter: {},
      },
    });
    if (errors) throw new Error(errors[0].message);
    return data.listTasks.items as Task[];
  } catch (err) {
    console.log(err);
    return [];
  }
};

export default fetchTasks;
```

hooks/useTasks.ts
```ts
"use client"

import fetchTasks from "../actions/fetchTasks";

const useTasks = () => {
  const { data } = useQuery({
    queryKey: ["tasks"],
    queryFn: fetchTasks, // using function name
  });
  return { data };
};

export default useTasks;
```

However, this didn't work. Data from `useTasks` hook always returned `undefined`.
When I put `console.log` at the top of inside `fetchTasks` function, this didn't show up, so I think using function name as queryFn is completely not working.

After some trials and errors, replacing function name with anonymous arrow function solved this problem.

hooks/useTasks.ts
```ts
"use client"

import fetchTasks from "../actions/fetchTasks";

const useTasks = () => {
  const { data } = useQuery({
    queryKey: ["tasks"],
    queryFn: () => fetchTasks(), // using anonymous arrow function
  });
  return { data };
};

export default useTasks;
```

Is this expected behavior? I know many of documentation codes are using function name as queryFn. (e.g. https://tanstack.com/query/latest/docs/framework/react/quick-start)

### Your minimal, reproducible example

https://codesandbox.io/p/github/tsuyuni/ReproduceQueryFnWithFunctionNameNotWorking/main?import=true

### Steps to reproduce

Just run `npm i` and `npm run dev`.

Then you will see:
**Tasks with function name** has no tasks though expected to have 3 tasks.
**Tasks with anonymous arrow function** has 3 tasks as it was expected.

### Expected behavior

queryFn with function name should properly work.

### How often does this bug happen?

Every time

### Platform

OS: Sonoma 14.2

### Tanstack Query adapter

react-query

### TanStack Query version

5.32.0

### TypeScript version

5.4.5

kam-st:
Yes, took me 5 hours to find the workaround. Drove me crazy.

Issue #6264, opened 2023-10-30
Server Action only works when queryFn isn't a reference


jschuur:
### Describe the bug

In a Next.js 14 app directory project with React Query 5 inside a client component, if I use reference (`queryFn: getProjects`) for the query function instead of an inline function that calls a function, I get this error:

`Only plain objects, and a few built-ins, can be passed to Server Actions. Classes or null prototypes are not supported.`

Doing it this way works:

```ts
useQuery({
  queryKey: ['projects'],
  queryFn: () => getProjects(),
});
```

Is this a bug?

I understand what a plain object is, and getProjects does return a promise to a serialisable object.

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/gracious-golick-fq3ss9

### Steps to reproduce

My `useQuery` call is abstracted away inside a custom hook that starts with `'use client';`:

```tsx
'use client';

import { useQuery } from '@tanstack/react-query';

import { getProjects } from '@/db/queries';

export default function useProjects() {
  return useQuery({
    queryKey: ['projects'],
    // works if inline function not if 'queryFn: getProjects'
    queryFn: () => getProjects(),
  });
}
```

This hook is used in a client component (`'use client';`). The error also happens if I use `useQuery` directly in the client component and skip the custom hook.

`getProjects` comes from a file prefixed with `'use server';` (the query is run by Drizzle on Turso):

```ts
'use server';

import 'server-only';

import { db } from '@/db/';
import { projects } from '@/db/schema';

export const getProjects = async () => db.select().from(projects);
```

I use the [HydrationBoundary pattern](https://tanstack.com/query/v5/docs/react/guides/advanced-ssr), but even if I remove that from `page.tsx` and just create a client side query client, I get the same error too, unless I use an inline query function.

### Expected behavior

`useQuery` accepts both a function reference and an inline function for `queryFn`.

### How often does this bug happen?

Every time

### Platform

- OS: macOS
- Browser: Chrome 
- Version: 118.0.5993.117

### Tanstack Query adapter

react-query

### TanStack Query version

v5.4.3

### TypeScript version

v5.2.2

### Additional context

The error only kicks in after a few seconds, so it looks like React Query is doing it's default attempt at trying a few times and backing off in between attempts.

weisisheng:
Seeing the same and your fix is also not working for me.  Oddly, I am still getting stale data from yesterday although console.log showing correct pull from database.  

tanstack/react-query 5.0.5, next.js 13.5.6

[...]

PR #7572, opened 2024-06-16
docs(react-query): `queryFn` must be a function calling

ndeitch:
I stumbled in this error and this issue was mentioned twice on #6264 & #7330.

### DIFF
--- a/docs/framework/react/guides/advanced-ssr.md
+++ b/docs/framework/react/guides/advanced-ssr.md
@@ -192,7 +192,10 @@ Next, we'll look at what the Client Component part looks like:
 export default function Posts() {
   // This useQuery could just as well happen in some deeper
   // child to <Posts>, data will be available immediately either way
-  const { data } = useQuery({ queryKey: ['posts'], queryFn: getPosts })
+  const { data } = useQuery({
+    queryKey: ['posts'],
+    queryFn: () => getPosts()
+  })
 
   // This query was not prefetched on the server and will not start
   // fetching until on the client, both patterns are fine to mix.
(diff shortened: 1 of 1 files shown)


docs/framework/react/guides/advanced-ssr.md
```diff
@@ -211,6 +214,8 @@ In the SSR guide, we noted that you could get rid of the boilerplate of having `
 
 > NOTE: If you encounter a type error while using async Server Components with TypeScript versions lower than `5.1.3` and `@types/react` versions lower than `18.2.8`, it is recommended to update to the latest versions of both. Alternatively, you can use the temporary workaround of adding `{/* @ts-expect-error Server Component */}` when calling this component inside another. For more information, see [Async Server Component TypeScript Error](https://nextjs.org/docs/app/building-your-application/configuring/typescript#async-server-component-typescript-error) in the Next.js 13 docs.
 
+> NOTE: If you encounter an error `Only plain objects, and a few built-ins, can be passed to Server Actions. Classes or null prototypes are not supported.` make sure that you're **not** passing to queryFn a function reference, instead call the function because queryFn args has a bunch of properties and not all of if would be serializable. see [Server Action only works when queryFn isn't a reference](https://github.com/TanStack/query/issues/6264).
```
ndeitch, line 217:
```suggestion
> NOTE: If you encounter an error `Only plain objects, and a few built-ins, can be passed to Server Actions. Classes or null prototypes are not supported.` make sure that you're **not** passing to queryFn a function reference, instead call the function because queryFn args has a bunch of properties and not all of it would be serializable. see [Server Action only works when queryFn isn't a reference](https://github.com/TanStack/query/issues/6264).
```

docs/framework/react/guides/advanced-ssr.md
```diff
@@ -211,6 +214,8 @@ In the SSR guide, we noted that you could get rid of the boilerplate of having `
 
 > NOTE: If you encounter a type error while using async Server Components with TypeScript versions lower than `5.1.3` and `@types/react` versions lower than `18.2.8`, it is recommended to update to the latest versions of both. Alternatively, you can use the temporary workaround of adding `{/* @ts-expect-error Server Component */}` when calling this component inside another. For more information, see [Async Server Component TypeScript Error](https://nextjs.org/docs/app/building-your-application/configuring/typescript#async-server-component-typescript-error) in the Next.js 13 docs.
 
+> NOTE: If you encounter an error `Only plain objects, and a few built-ins, can be passed to Server Actions. Classes or null prototypes are not supported.` make sure that you're **not** passing to queryFn a function reference, instead call the function because queryFn args has a bunch of properties and not all of it would be serializable. see [Server Action only works when queryFn isn't a reference](https://github.com/TanStack/query/issues/6264).
```
AhmedBaset, line 217: That's because you passes non-serliarized value from a Server Component to a Client Component. It has nothing to do with TSQ

PR #7577, opened 2024-06-16
chore(angular-query): integration tests for angular 16 up to 18

arnoud-dv:


arnoud-dv:
Closing as different versions of Angular in the monorepo result in dependencies being resolved in weird ways such as angular cli using a mismatching version of angular core.

PR #7582, opened 2024-06-17
feat: usePrefetchQuery

TkDodo:
missing:

- [x] docs
- [x] runtime tests
- [x] type tests


docs/framework/react/guides/prefetching.md
```diff
@@ -196,45 +196,41 @@ This starts fetching `'article-comments'` immediately and flattens the waterfall
...
-  if (!queryClient.getQueryState(options.queryKey)) {
-    queryClient.ensureQueryData(options).catch(() => {
-      // Avoid uncaught error
-    })
-  }
-}
-```
-
-This approach works with both `useQuery` and `useSuspenseQuery`, so feel free to use it as an alternative to the `useQuery({ ..., notifyOnChangeProps: [] })` approach as well. The only tradeoff is that the above function will never fetch and _update_ existing data in the cache if it's stale, but this will usually happen in the later query anyway.
+If you want to prefetch together with Suspense, you will have to do things a bit differently. You can't use `useSuspenseQueries` to prefetch, since the prefetch would block the component from rendering. You also can not use `useQuery` for the prefetch, because that wouldn't start the prefetch until after suspenseful query had resolved. For this scenario, you can use the [`usePrefetchQuery`](../reference/usePrefetchQuery) or the [`usePrefetchInfiniteQuery`](../reference/usePrefetchInfiniteQuery) hooks available in the library
```
Ephem, line 199:
```suggestion
If you want to prefetch together with Suspense, you will have to do things a bit differently. You can't use `useSuspenseQueries` to prefetch, since the prefetch would block the component from rendering. You also can not use `useQuery` for the prefetch, because that wouldn't start the prefetch until after suspenseful query had resolved. For this scenario, you can use the [`usePrefetchQuery`](../reference/usePrefetchQuery) or the [`usePrefetchInfiniteQuery`](../reference/usePrefetchInfiniteQuery) hooks available in the library.
```

packages/react-query/src/__tests__/prefetch.test.tsx
```diff
@@ -0,0 +1,352 @@
...
+        <React.Suspense fallback="Loading...">
+          <Suspended queryOpts={queryOpts} />
+        </React.Suspense>
+      )
+    }
+
+    const rendered = renderWithClient(queryClient, <App />)
+
+    await waitFor(() => rendered.getByText('data: Prefetch is nice!'))
+    expect(queryOpts.queryFn).toHaveBeenCalledTimes(1)
```
Ephem, line 87: Is this testing what it says? It might also be the case that `usePrefetchQuery` didn't fetch, but `useSuspenseQuery` did. Not sure if it necessarily needs fixing, but thought I should point it out.

packages/react-query/src/__tests__/prefetch.test.tsx
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    queryFn.mockImplementationOnce(async () => {
+      await sleep(10)
+
+      throw new Error('Oops! Server error!')
+    })
+
+    function App() {
+      usePrefetchQuery(queryOpts)
```
Ephem, line 132: I _think_ to replicate the bug we saw with endless loops, the `usePrefetchQuery` would have to be inside the Suspense boundary so it would rerender->refetch when unsuspended? I haven't tested so not entirely sure though.

packages/react-query/src/__tests__/prefetch.test.tsx
```diff
@@ -0,0 +1,352 @@
...
+    function App() {
+      usePrefetchQuery(firstQueryOpts)
+      usePrefetchQuery(secondQueryOpts)
+      usePrefetchQuery(thirdQueryOpts)
+
+      return (
+        <React.Suspense fallback={<Fallback />}>
+          <Suspended queryOpts={firstQueryOpts} />
+          <Suspended queryOpts={secondQueryOpts} />
+          <Suspended queryOpts={thirdQueryOpts} />
```
Ephem, line 225: With sibling prerendering (the React 18 behaviour), this wouldn't waterfall even without the prefetch I think? If the `<Suspended>` component was remade to optionally take children, I think this would be a better test:

```suggestion
          <Suspended queryOpts={firstQueryOpts}>
            <Suspended queryOpts={secondQueryOpts}>
              <Suspended queryOpts={thirdQueryOpts} />
            </Suspended>
          </Suspended>
```

packages/react-query/src/__tests__/prefetch.test.tsx
```diff
@@ -0,0 +1,352 @@
...
+
+    const rendered = renderWithClient(queryClient, <App />)
+
+    await waitFor(() => rendered.getByText('Loading...'))
+    await waitFor(() => rendered.getByText('data: Prefetch is nice!'))
+    await waitFor(() => rendered.getByText('data: Prefetch is really nice!!'))
+    await waitFor(() =>
+      rendered.getByText('data: Prefetch does not create waterfalls!!'),
+    )
+    expect(Fallback).toHaveBeenCalledTimes(1)
```
Ephem, line 238: Is this what tests for waterfalls? Seems a bit brittle to me. Maybe we could inspect the `queryClient` right after the `renderWithClient` call to see that all three queries are fetching?

packages/react-query/src/__tests__/prefetch.test.tsx
```diff
@@ -0,0 +1,352 @@
...
+    const rendered = renderWithClient(queryClient, <App />)
+
+    expect(rendered.queryByText('fetching: true')).not.toBeInTheDocument()
+    await waitFor(() =>
+      rendered.getByText('data: The usePrefetchQuery hook is smart!'),
+    )
+    expect(queryOpts.queryFn).not.toHaveBeenCalled()
+  })
+
+  it('should display an error boundary if query cache is populated with an error', async () => {
```
Ephem, line 117: NIT: I find this slightly misleading since it's the `useSuspenseQuery` that triggers the error that shows the error boundary.

Not perfect, but maybe something along the lines of:

```suggestion
  it('should let errors fall through and not refetch failed queries', async () => {
```

Ephem (approved):
I left a few NITs on the tests, but I think the implementation is fine to ship as is. 🎉

packages/react-query/src/__tests__/prefetch.test.tsx
TkDodo: yeah we should probably use a different `queryFn` for prefetching and the component so that we can assert which one ran. @brunolopesr can you add that?

packages/react-query/src/__tests__/prefetch.test.tsx
brunolopesr-dft: Hey @Ephem 

What I know about this endless loops bug, is that before it was using the `queryClient.ensureQuery` right? But I will add a test to cover this.

brunolopesr: @Ephem , I'm not sure, but this endless loops isn't about the usage of `ensureQueryData` instead of `prefetchQuery`? But I will add a test covering the usage of the hook inside the Suspense Boundary.

packages/react-query/src/__tests__/prefetch.test.tsx
brunolopesr: @Ephem @TkDodo 

Added a better assertion for this in PR #7614

packages/react-query/src/__tests__/prefetch.test.tsx
brunolopesr: Added new tests to ensure that endless loops don't occur at #7614

packages/react-query/src/__tests__/prefetch.test.tsx
brunolopesr: Added a test to assert the query state just after rendering in #7614

PR #7586, opened 2024-06-18
chore: add tests for usePrefetchQuery and usePrefetchInfiniteQuery

brunolopesr:
This PR adds tests to #7582

### DIFF
--- a/packages/react-query/src/__tests__/prefetch.test.tsx
+++ b/packages/react-query/src/__tests__/prefetch.test.tsx
@@ -100,10 +100,9 @@ describe('usePrefetchQuery', () => {
     const rendered = renderWithClient(queryClient, <App />)
 
     expect(rendered.queryByText('fetching: true')).not.toBeInTheDocument()
-    expect(
-      rendered.getByText('data: Prefetch is really nice!'),
-    ).toBeInTheDocument()
+    await waitFor(() => rendered.getByText('data: Prefetch is really nice!'))
     expect(suspendedQueryFn).not.toHaveBeenCalled()
+    expect(queryOpts.queryFn).toHaveBeenCalledTimes(1)
   })
 
   it('should display an error boundary if query cache is populated with an error', async () => {
@@ -130,7 +129,7 @@ describe('usePrefetchQuery', () => {
     }
 
     function App() {
-      usePrefetchQuery({ ...queryOpts })
+      usePrefetchQuery(queryOpts)
 
       return (
         <ErrorBoundary fallbackRender={() => <div>Oops!</div>}>
@@ -156,7 +155,7 @@ describe('usePrefetchQuery', () => {
     expect(suspendedQueryFn).not.toHaveBeenCalled()
   })
 
-  it.only('should be able to recover from errors and try fetching again', async () => {
+  it('should be able to recover from errors and try fetching again', async () => {
     const queryFn = vi.fn()
 
     const queryOpts = {
--- a/packages/react-query/src/__tests__/prefetch.test.tsx
+++ b/packages/react-query/src/__tests__/prefetch.test.tsx
@@ -291,21 +291,18 @@ describe('usePrefetchQuery', () => {
       )
     }
 
+    const Fallback = vi.fn().mockImplementation(() => <div>Loading...</div>)
+
     function App() {
       usePrefetchQuery(firstQueryOpts)
       usePrefetchQuery(secondQueryOpts)
       usePrefetchQuery(thirdQueryOpts)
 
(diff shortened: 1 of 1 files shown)


packages/react-query/src/__tests__/prefetch.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+        </React.Suspense>
+      )
+    }
+
+    const rendered = renderWithClient(queryClient, <App />)
+
+    await waitFor(() => rendered.getByText('data: Prefetch is nice!'))
+    expect(rendered.queryByText('fetching: true')).not.toBeInTheDocument()
+    expect(queryOpts.queryFn).toHaveBeenCalledTimes(1)
+    expect(anotherSpy).not.toHaveBeenCalled()
```
TkDodo, line 55: perfect 👏

packages/react-query/src/__tests__/prefetch.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    await queryClient.fetchQuery(queryOpts)
+    const rendered = renderWithClient(queryClient, <App />)
+
+    expect(rendered.queryByText('fetching: true')).not.toBeInTheDocument()
+    expect(
+      rendered.getByText('data: Prefetch is really nice!'),
+    ).toBeInTheDocument()
+    expect(anotherSpy).not.toHaveBeenCalled()
```
TkDodo, line 98: let's also add:

```suggestion
    expect(anotherSpy).not.toHaveBeenCalled()
    expect(queryOpts.queryFn).not.toHaveBeenCalled()
```

packages/react-query/src/__tests__/prefetch.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+      )
+    }
+
+    const rendered = renderWithClient(queryClient, <App />)
+
+    await waitFor(() => rendered.getByText('Loading...'))
+    await waitFor(() => rendered.getByText('data: Prefetch is nice!'))
+    fireEvent.click(rendered.getByText('showSecond'))
+    expect(rendered.queryByText('Loading...')).not.toBeInTheDocument()
+    await waitFor(() => rendered.getByText('data: Prefetch is really nice!!'))
```
TkDodo, line 184: question: since we `await` here, how do we know that the fetch isn't triggered here? I think if you remove the prefetching, this test would still pass, no?

packages/react-query/src/__tests__/prefetch.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+      usePrefetchQuery({ ...queryOpts, queryFn: anotherSpy })
+
+      return (
+        <React.Suspense fallback="Loading...">
+          <Suspended />
+        </React.Suspense>
+      )
+    }
+
+    await queryClient.fetchQuery(queryOpts)
```
TkDodo, line 91: can we please add another test where we have a failed query in the cache? This used to bite us when the implementation was still `ensureQueryData` - the query was there, but data was not.

I think you can get that with:


```
await queryClient.prefetchQuery({
  ...queryOpts,
  queryFn: () => { 
    throw new Error("oops")
  },
  retry: false
})
```

prefetching doesn't throw so there shouldn't be any try/catch needed.

TkDodo (commented):
great work 🚀

packages/react-query/src/__tests__/prefetch.test.tsx
brunolopesr-dft: Yes, you are right.

I think I need to pass an alternative spy here too, to assert that the prefetch is the one calling the right function!

brunolopesr: Yes, you are right.

I think I need to pass an alternative spy here too, to assert that the prefetch is the one calling the right function!

packages/react-query/src/__tests__/prefetch.test.tsx
brunolopesr: Of course, doing it right now

brunolopesr: Added two error scenarios in the test file

packages/react-query/src/__tests__/prefetch.test.tsx
TkDodo: I think it's better to call `mockClear()` on the spy after you've called `fetchQuery`

brunolopesr: Yep, good idea! Updated :smile:

TkDodo (approved):

PR #7592, opened 2024-06-19
usePrefetchQuery type tests and docs

brunolopesr:
Adds type tests and docs to #7582

### DIFF
--- a/docs/framework/react/guides/prefetching.md
+++ b/docs/framework/react/guides/prefetching.md
@@ -196,7 +196,7 @@ This starts fetching `'article-comments'` immediately and flattens the waterfall
 
 [//]: # 'Suspense'
 
-If you want to prefetch together with Suspense, you will have to do things a bit differently. You can't use `useSuspenseQueries` to prefetch, since the prefetch would block the component from rendering. You also can not use `useQuery` for the prefetch, because that wouldn't start the prefetch until after suspenseful query had resolved. For this scenario, you can use the [`usePrefetchQuery`](../reference/usePrefetchQuery) or the [`usePrefetchInfiniteQuery`](../reference/usePrefetchInfiniteQuery.md) hooks available in the library
+If you want to prefetch together with Suspense, you will have to do things a bit differently. You can't use `useSuspenseQueries` to prefetch, since the prefetch would block the component from rendering. You also can not use `useQuery` for the prefetch, because that wouldn't start the prefetch until after suspenseful query had resolved. For this scenario, you can use the [`usePrefetchQuery`](../reference/usePrefetchQuery) or the [`usePrefetchInfiniteQuery`](../reference/usePrefetchInfiniteQuery) hooks available in the library
 
 You can now use `useSuspenseQuery` in the component that actually needs the data. You _might_ want to wrap this later component in its own `<Suspense>` boundary so the "secondary" query we are prefetching does not block rendering of the "primary" data.
 
--- a/docs/framework/react/guides/prefetching.md
+++ b/docs/framework/react/guides/prefetching.md
@@ -196,7 +196,7 @@ This starts fetching `'article-comments'` immediately and flattens the waterfall
 
 [//]: # 'Suspense'
 
-If you want to prefetch together with Suspense, you will have to do things a bit differently. You can't use `useSuspenseQueries` to prefetch, since the prefetch would block the component from rendering. You also can not use `useQuery` for the prefetch, because that wouldn't start the prefetch until after suspenseful query had resolved. For this scenario, you can use the `usePrefetchQuery` or the `usePrefetchInfiniteQuery` hooks available in the library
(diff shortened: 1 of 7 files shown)


docs/framework/react/guides/prefetching.md
```diff
@@ -196,45 +196,41 @@ This starts fetching `'article-comments'` immediately and flattens the waterfall
...
-  if (!queryClient.getQueryState(options.queryKey)) {
-    queryClient.ensureQueryData(options).catch(() => {
-      // Avoid uncaught error
-    })
-  }
-}
-```
-
-This approach works with both `useQuery` and `useSuspenseQuery`, so feel free to use it as an alternative to the `useQuery({ ..., notifyOnChangeProps: [] })` approach as well. The only tradeoff is that the above function will never fetch and _update_ existing data in the cache if it's stale, but this will usually happen in the later query anyway.
+If you want to prefetch together with Suspense, you will have to do things a bit differently. You can't use `useSuspenseQueries` to prefetch, since the prefetch would block the component from rendering. You also can not use `useQuery` for the prefetch, because that wouldn't start the prefetch until after suspenseful query had resolved. For this scenario, you can use the `usePrefetchQuery` or the `usePrefetchInfiniteQuery` hooks available in the library
```
TkDodo, line 199: please link to the api docs here

docs/framework/react/reference/usePrefetchInfiniteQuery.md
```diff
@@ -0,0 +1,37 @@
+---
```
TkDodo, line 1: you need to add those to the `docs/config.json` file in order for them to show up in the sidebar

docs/framework/react/reference/usePrefetchQuery.md
```diff
@@ -0,0 +1,24 @@
...
+title: usePrefetchQuery
+---
+
+```tsx
+const result = usePrefetchQuery(options)
+```
+
+**Options**
+
+You can pass everything to `usePrefetchQuery` that you can pass to [`queryOptions`](../queryOptions) or [`useQuery`](../useQuery). Some options will have no effect because, under the hood, `usePrefetchQuery` uses `queryClient.prefetchQuery`. Since the options are general, TypeScript will not warn you about any excess properties.
```
TkDodo, line 12: "some options will have no effect": This is a good point actually. I think we shouldn't allow to pass in `UseQueryOptions` but rather `FetchQueryOptions`

then we will also know what we have to type-test: for example, that the `refetchInterval` option cannot be passed.

docs/framework/react/guides/prefetching.md
brunolopesr: Done!

docs/framework/react/reference/usePrefetchInfiniteQuery.md
brunolopesr: Done!

docs/framework/react/reference/usePrefetchQuery.md
brunolopesr: Done \O/

docs/framework/react/reference/usePrefetchInfiniteQuery.md
```diff
@@ -0,0 +1,37 @@
...
+- `getNextPageParam: (lastPage, allPages, lastPageParam, allPageParams) => TPageParam | undefined | null`
+
+  - **Required**
+  - When new data is received for this query, this function receives both the last page of the infinite list of data and the full array of all pages, as well as pageParam information.
+  - It should return a **single variable** that will be passed as the last optional parameter to your query function.
+  - Return `undefined` or `null` to indicate there is no next page available.
+
+- **Returns**
+
+The `usePrefetchInfiniteQuery` does not return nothing, it should be used just to fire a prefetch during render, before a suspense boundary that wraps a component that uses [`useSuspenseInfiniteQuery`](../reference/useSuspenseInfiniteQuery)
```
TkDodo, line 37:
```suggestion
The `usePrefetchInfiniteQuery` does not return anything, it should be used just to fire a prefetch during render, before a suspense boundary that wraps a component that uses [`useSuspenseInfiniteQuery`](../reference/useSuspenseInfiniteQuery)
```

docs/framework/react/reference/usePrefetchQuery.md
```diff
@@ -0,0 +1,24 @@
...
+
+  - **Required**
+  - The query key to prefetch during render
+
+- `queryFn: (context: QueryFunctionContext) => Promise<TData>`
+  - **Required, but only if no default query function has been defined** See [Default Query Function](../../guides/default-query-function) for more information.
+
+**Returns**
+
+The `usePrefetchQuery` does not return nothing, it should be used just to fire a prefetch during render, before a suspense boundary that wraps a component that uses [`useSuspenseQuery`](../reference/useSuspenseQuery).
```
TkDodo, line 24:
```suggestion
The `usePrefetchQuery` does not return anything, it should be used just to fire a prefetch during render, before a suspense boundary that wraps a component that uses [`useSuspenseQuery`](../reference/useSuspenseQuery).
```

TkDodo (approved):

PR #7593, opened 2024-06-19
refactor(angular-query): align inject-query and inject-infinite-query implementations, refactor create-base-query

Arthie:
The company I work for is moving from our custom query wrapper to the official one. I had some time to audit the code and refactored a bit.  

Aligned the `injectQuery` and `injectInfiniteQuery` implementations. These are simpler now and refactored the queryclient to createBaseQuery closer to where it's actually used. This can be changed again once there is a new API to provide a custom queryClient to injectQuery.

Had some time left to add types tests to the injectQuery tests.
Hope this is useful and that I get some more time in the future to contribute more.

Test are passing, and I've tested these changes in the examples and our company app.

### DIFF
--- a/packages/angular-query-experimental/src/__tests__/inject-query.test.ts
+++ b/packages/angular-query-experimental/src/__tests__/inject-query.test.ts
@@ -1,29 +1,253 @@
 import {
   Component,
   Injectable,
+  Injector,
   computed,
+  effect,
   inject,
   input,
   signal,
 } from '@angular/core'
 import { TestBed, fakeAsync, flush, tick } from '@angular/core/testing'
-import { QueryClient } from '@tanstack/query-core'
 import { describe, expect, vi } from 'vitest'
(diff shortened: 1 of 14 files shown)


packages/angular-query-experimental/src/__tests__/inject-query.test.ts
```diff
@@ -1,31 +1,255 @@
...
   beforeEach(() => {
+    queryCache = new QueryCache()
+    queryClient = new QueryClient({ queryCache })
     TestBed.configureTestingModule({
-      providers: [provideAngularQuery(new QueryClient())],
+      providers: [provideAngularQuery(queryClient)],
     })
   })
 
+  test('should return the correct types', () => {
```
Arthie, line 35: These are Angular's version of the correct types test for react-query:
https://github.com/TanStack/query/blob/main/packages/react-query/src/__tests__/useQuery.test.tsx

PR #7605, opened 2024-06-21
ci: Add TS version tests

lachlancollins:
Uses the strategy implemented in TanStack Form and Store to take full advantage of Nx.

This strategy can be tested and debugged locally rather than needing to run in CI (e.g. the matrix approach). It also all runs in a single GitHub Action job, making it easier to work out which task has failed (all logs are in the Nx cloud report).

Currently tests TS 5.0 to 5.3. Adding pre-5.0 tests is a bit more involved, but TanStack Form and Store have a workaround which I'll implement in a subsequent PR.


TkDodo (commented):
I think a better way would be to just have CI install a different version of typescript, like we do in router:

https://github.com/TanStack/router/blob/main/.github/workflows/pr.yml#L77-L80

then we add the TS_VERSION as input to nx to avoid cache hits: https://github.com/TanStack/router/blob/cb6c21150f9637cef5d48596a489ffd2e5366303/nx.json#L41

also: what's all the file moves about 😅 ?

lachlancollins:
> I think a better way would be to just have CI install a different version of typescript, like we do in router:
> 
> [TanStack/router@`main`/.github/workflows/pr.yml#L77-L80](https://github.com/TanStack/router/blob/main/.github/workflows/pr.yml?rgh-link-date=2024-06-21T11%3A24%3A52Z#L77-L80)
> 
> then we add the TS_VERSION as input to nx to avoid cache hits: [TanStack/router@`cb6c211`/nx.json#L41](https://github.com/TanStack/router/blob/cb6c21150f9637cef5d48596a489ffd2e5366303/nx.json#L41)
> 
> also: what's all the file moves about 😅 ?

Yeah I've seen that option too, I prefer how its been implemented for TanStack Form, as it can be tested and debugged locally too rather than needing to run in CI. It also all runs in a single GitHub Action job, making it easier to work out which task has failed (everything is in the Nx cloud report).

As for the file moves, it makes it easier to target different files for different tsconfigs. The "main" TS version can check all files (src, tests, vite config, etc.), while the "legacy" TS versions only need to check src, which speeds up testing. The same thing can be achieved with slightly more annoying "include" and "exclude" config.

arnoud-dv:
Looks good, I like that this strategy allows us to check different versions of TS per package. For example the lowest version of Angular supported by angular-query is v16, which requires TS 4.9 or higher. So unlike react-query we wouldn't need to check either 4.7 or 4.8.

PR #7607, opened 2024-06-21
chore: typescript tests 4.7 up to 5.5

arnoud-dv:


lachlancollins (commented):
It looks like tests aren't actually running on the matrix jobs at the moment, but it's still "passing"

PR #7608, opened 2024-06-21
feat(vue-query): add support for getters in query keys

suneettipirneni:
This is a based on the discussion I found in https://github.com/TanStack/query/discussions/5990. According to the author of the PR which made `enabled` a getter, they stated they didn't find a use case for query keys so it was not implemented.

Personally, I have found many use cases for getters inside of query keys. Commonly, I'll define composables for my API and then have them accept reactive variables like so:

```ts
export function useProjectDetailsQuery(userId: MaybeRef<string>, projectId: MaybeRef<string>) {
  return useQuery({
    queryKey: ['users', userId, 'projects', projectId, 'details'],
    queryFn: () => api.fetchProjectDetails(toValue(userId), toValue(projectId))
  })
}
```

When I use these composable I often have to make intermediate `computed` variables to keep the data from props reactive:

```ts
import { useQuery } from '@tanstack/vue-query';

export interface Props {
  user: User;
  project: Project;
}

const props = defineProps<Props>();

const { data: project } = useProjectDetailsQuery(
  computed(() => props.user.id),
  computed(() => props.project.id),
);
```

With this PR the we don't need intermediate `computed` properties to do this, you can write the code using getters instead:

```ts
const { data: project } = useProjectDetailsQuery(
  () => props.user.id,
  () => props.project.id,
);
```

This also makes composables like the one I defined above fall more into line with the recommended practices for vue composables https://vuejs.org/guide/reusability/composables#input-arguments.

CC @DamianOsipiuk


DamianOsipiuk:
This could improve ergonomics, but solution is insufficient.

QueryKey can be infinitely nested structure, so checking only the top level of QueryKey would not resolve deeply nested getters. This in turn will break query key serialization.

We would need to create a variant of `cloneDeepUnref` that will unwrap getters as well, but we would need to run this only on `QueryKey` as other parameters might be functions on purpose.

suneettipirneni:
> This could improve ergonomics, but solution is insufficient.
> 
> QueryKey can be infinitely nested structure, so checking only the top level of QueryKey would not resolve deeply nested getters. This in turn will break query key serialization.
> 
> We would need to create a variant of `cloneDeepUnref` that will unwrap getters as well, but we would need to run this only on `QueryKey` as other parameters might be functions on purpose.

Ok thank you for your feedback on this. I'll look into how I can conditionally deep unwrap getters specifically for the query key.

suneettipirneni:
@DamianOsipiuk I think I've managed to get nested getter functions to work. The new and old tests are passing so I (assume?) nothing broke in the process.  Basically I added a `level` and `key` parameter to the customize callback to detect if we hit the top level queryKey. For useQueries I've refactored it to first unref the top level array and then process each element the same way useQuery process it's options.

I'm pretty confident useQuery will work as expected, I just wanted to verify with you if the changes to useQueries are ok.

PR #7615, opened 2024-06-24
feat(hydration): `dehydrate.serialize` and `hydrate.deserialize`

juliusmarminge:
> [!NOTE]
> Given that the feature is brand new and there were likely not many people using it, combined with it not really working to begin with, we have decided to make a breaking change and **replace** `transformPromise` with `serialize`/`deserialize`.

When a query gets resolved on the server before we dehydrate, the dehydrated query will contain the serialized data, but the `transformPromise` function will never run to deserialize it, leading to serialized data being put in the query cache:

CleanShot 2024-06-24 at 22 19 26@2x

You can simulate the failing behavior like this:

```ts
// page.tsx
export default async function Home() {
  const queryClient = makeQueryClient()

  void queryClient.prefetchQuery({
    queryKey: ['data'],
    queryFn: async () => {
      await sleep(1)
      return tson.serialize({
        text: 'data from server',
        date: Temporal.PlainDate.from('2024-01-01'),
      })
    },
  })

  await sleep(10)

  // by now the prefetched query will have resolved, so the dehydratedState
  // will contain the serialized data, not the promise
  const dehydratedState = dehydrate(queryClient)
  console.log('dehydratedState', JSON.stringify(dehydratedState, null, 2))

  return (
    <main>
      <HydrationBoundary state={dehydratedState}>
        <ClientComponent />
      </HydrationBoundary>
    </main>
  )
}
```

To solve this, I've (as discussed) replaced the `transformPromise` option with `transformData` and added the transformation step to the other necessary places as well


packages/query-core/src/__tests__/hydration.test.tsx
```diff
@@ -943,4 +943,41 @@ describe('dehydration and rehydration', () => {
...
+      },
+    })
+
+    const promise = queryClient.prefetchQuery({
+      queryKey: ['transformedStringToDate'],
+      queryFn: () => fetchData('2024-01-01T00:00:00.000Z', 0),
+    })
+    await sleep(20)
+    const dehydrated = dehydrate(queryClient)
+    console.log('de', dehydrated)
```
KATT, line 962:
```suggestion
```

TkDodo (approved):

docs/framework/react/reference/hydration.md
```diff
@@ -37,6 +37,7 @@ const dehydratedState = dehydrate(queryClient, {
       - Return `true` to include this query in dehydration, or `false` otherwise
     - Defaults to only including successful queries
     - If you would like to extend the function while retaining the default behavior, import and execute `defaultShouldDehydrateQuery` as part of the return statement
+    - `serializeData?: (data: any) => any` A function to transform (serialize) data during dehydration.
```
juliusmarminge, line 40: Indentation

TkDodo: thx , fixed with bb50d899784849fdf865f2423ecf3a9e5765bb14

PR #7623, opened 2024-06-26
Remove the variable assignment from example code

kenrick:
Noticed that `usePrefetchQuery`'s documentation states that it doesn't return anything but the example code shows it seemingly returning a result.


TkDodo:
Please also for `usePrefetchInfiniteQuery`

kenrick:
@TkDodo done!

PR #7625, opened 2024-06-26
fix(vue-query): remove unnecessary enabled guard

il-1337:
ensure enabled as getter receives query param, remove unnecessary guard following v5.48.0 changes in query-core


DamianOsipiuk:
This would not work in a reactive way, unless you add all dependencies used in `enabled` callback into the `queryKey`.

And since getter on enabled in Vue was introduced mostly to `reactively` enable/disable query, i'm not sure if we want to break this assumption.

PR #7642, opened 2024-06-28
fix(query-core): add error management around this.setData(data)

### LINKED ISSUE #6954: The query gets stuck at `fetching` status when re-fetched data has cycles
### Describe the bug

The query gets stuck at `fetching` status when you try to re-fetch data and it has cycles in it.

https://codesandbox.io/p/sandbox/react-query-mfsz4x

### Steps to reproduce

 - create a query that returns data with cycles in it
 - re-fetch the query
 - see that the `fetchStatus` gets stuck at `fetching`

### Expected behavior

The query should change the `fetchStatus` to `idle` when re-fetch has been finished.

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: MaxOs
- Browser: Chrome
- Version: 121.0.6167.184 (Official Build) (x86_64)

### Tanstack Query adapter

react-query

### TanStack Query version

v5.22.2

### TypeScript version

_No response_

### Additional context

As far as I understand, when you turn off `structuralSharing` for the query - the issue will go away.
After some digging, I found that default `structuralSharing` functionality is to call `replaceEqualDeep` function. And when there are cycles in the data, this function goes into infinite loop because it recursively calls itself on each data property.

robin4a4:
This aim to fix #6954. I am not sure that the test perfectly reproduces the initial issue as I am really new to this code base but it was failing before the try catch and seems to pass after the fix.


packages/query-core/src/__tests__/query.test.tsx
```diff
@@ -943,4 +943,39 @@ describe('query', () => {
...
     expect(spy).toHaveBeenCalledWith('1 - 2')
   })
+
+  it('should call onError when setData throws', async () => {
+    const key = queryKey()
+
+    const queryFn = vi.fn<Array<unknown>, unknown>()
+
+    queryFn.mockImplementation(async () => {
+      await sleep(50)
```
TkDodo, line 953: let's get those numbers down:

```suggestion
      await sleep(10)
```

packages/query-core/src/__tests__/query.test.tsx
```diff
@@ -943,4 +943,39 @@ describe('query', () => {
...
+        data[0].link = data[1]
+        data[1].link = data[0]
+      }
+
+      return data
+    })
+
+    queryClient.prefetchQuery({ queryKey: key, queryFn })
+    const query = queryCache.find({ queryKey: key })!
+    await sleep(100)
```
TkDodo, line 976:
```suggestion
    await queryClient.prefetchQuery({ queryKey: key, queryFn })
    const query = queryCache.find({ queryKey: key })!
```

packages/query-core/src/__tests__/query.test.tsx
```diff
@@ -943,4 +943,39 @@ describe('query', () => {
...
+    expect(queryFn).toHaveBeenCalledTimes(1)
+    await query.fetch()
```
TkDodo, line 978: I don't understand why we fetch twice: first with the prefetching, and then with `query.fetch`. Shouldn't `await prefetchQuery(...)` be enough to get the query into error state ?

TkDodo:
some types are not quite right with your mocks, which is why the pipeline fails

TkDodo:
superseded by:
- https://github.com/TanStack/query/pull/7966

Issue #7612, opened 2024-06-24
No information about matchQuery in the docs


swingthrough:
### Describe the bug

I saw this tweet:
https://x.com/TkDodo/status/1744604450566680864

> queryClient.invalidateQueries({
  predicate: query =>
    matchQuery(['comments']) || matchQuery(['likes'])
})
>
> matchQuery can be imported from RQ

So I tried searching `matchQuery` in the docs but nothing was found. Shouldn't this be mentioned somewhere?

Edit:
I also found usage of `matchQuery` in this article:
https://tkdodo.eu/blog/automatic-query-invalidation-after-mutations#use-the-meta-option

but the usage is a little bit different:
> matchQuery({ queryKey }, query)

### Your minimal, reproducible example

...

### Steps to reproduce

1. Go to the docs
2. Search for `matchQuery`
3. No results are found

### Expected behavior

I'd expect `matchQuery` to be documented.

### How often does this bug happen?

None

### Platform

docs

### Tanstack Query adapter

None

### TanStack Query version

latest

TkDodo:
> but the usage is a little bit different:

the syntax in the tweet is wrong.

> I'd expect matchQuery to be documented.

please make a PR and add it to this page:
https://tanstack.com/query/latest/docs/framework/react/guides/filters

swingthrough:
> please make a PR and add it to this page: https://tanstack.com/query/latest/docs/framework/react/guides/filters

That page lists properties of `QueryFilters` interface, it doesn't have an example of such utilities as `matchQuery`, `matchMutation` and others that are defined here:
https://github.com/TanStack/query/blob/63552449845792954a4ea6801d843da3fbb4e293/packages/query-core/src/utils.ts#L103

Is it really the right place for it? Anyway, I'm not sure what the phrasing should be, that's why I was looking for it in the first place.

PR #7643, opened 2024-06-28
docs(query-core): add matchQuery and matchMutation utils documentation

robin4a4:
This PR adds the mention of matchQuery and matchMutation to the filter guide to address #7612


TkDodo:
thanks, please run prettier !

PR #7725, opened 2024-07-11
Update Streaming with Server Components docs to reflect correct usage

Aerilym:
[Streaming with Server Components](https://tanstack.com/query/latest/docs/framework/react/guides/advanced-ssr#streaming-with-server-components) has a minor issue with its usage example.
- Showed and explained the move of the `getQueryClient` function from the `gloabal/providers.tsx` file to its own file.
- Added missing imports.
- Replaced `new QueryClient()` with `getQueryClient()` to make use of the client with serialize and deserialize.
	The code block prior (`app/get-query-client.ts`) is used to add `serialize` and `deserialize`, I imagine the intent was to use that query client in the server component.


TkDodo:
please fix formatting

Aerilym:
Sorry about that, I've run prettier so the formatting should be all good now

PR #7733, opened 2024-07-13
fix(vue-query): use shallowReactive and shallowReadonly to prevent performance slow down for large dataset

wns3645:
Use `shallowReactive` and `shallowReadonly` instead of `reactive` and `readonly`

[`reactive`](https://vuejs.org/api/reactivity-core.html#reactive) and [`readonly`](https://vuejs.org/api/reactivity-core.html#reactive) uses deep conversion which affects all nested properties. Therefore, performance slows when the data is a large array of deeply nested objects. 

[`shallowReactive`](https://vuejs.org/api/reactivity-advanced.html#shallowreactive) is enough because we need reactivity for only shallow properties(i.e., properties at the first depth).
Although [`shallowReadonly`](https://vuejs.org/api/reactivity-advanced.html#shallowreactive) cannot strictly prevent mutating deeply nested properties, it would be enough as ref values returned from third-party composables are expected not to change directly.

### DIFF
--- a/packages/vue-query/src/useBaseQuery.ts
+++ b/packages/vue-query/src/useBaseQuery.ts
@@ -2,6 +2,7 @@ import {
   computed,
   getCurrentScope,
   onScopeDispose,
+  readonly,
   shallowReactive,
   shallowReadonly,
   toRefs,
@@ -145,7 +146,7 @@ export function useBaseQuery<
     return new Promise<QueryObserverResult<TData, TError>>(
       (resolve, reject) => {
         let stopWatch = () => {
-          //noop
+          // noop
         }
         const run = () => {
           if (defaultedOptions.value.enabled !== false) {
@@ -201,7 +202,15 @@ export function useBaseQuery<
     },
   )
 
-  const object: any = toRefs(shallowReadonly(state))
+  const readonlyState =
+    process.env.NODE_ENV === 'production'
+      ? state
+      : // @ts-expect-error
+        defaultedOptions.value.shallow
+        ? shallowReadonly(state)
+        : readonly(state)
+
+  const object: any = toRefs(readonlyState)
   for (const key in state) {
     if (typeof state[key as keyof typeof state] === 'function') {
       object[key] = state[key as keyof typeof state]
--- a/packages/vue-query/src/useBaseQuery.ts
+++ b/packages/vue-query/src/useBaseQuery.ts
@@ -2,8 +2,8 @@ import {
   computed,
   getCurrentScope,
   onScopeDispose,
-  reactive,
-  readonly,
+  shallowReactive,
+  shallowReadonly,
   toRefs,
   watch,
 } from 'vue-demi'
@@ -105,7 +105,7 @@ export function useBaseQuery<
   })
 
   const observer = new Observer(client, defaultedOptions.value)
-  const state = reactive(observer.getCurrentResult())
+  const state = shallowReactive(observer.getCurrentResult())
 
   let unsubscribe = () => {
     // noop
@@ -201,7 +201,7 @@ export function useBaseQuery<
     },
   )
 
-  const object: any = toRefs(readonly(state))
+  const object: any = toRefs(shallowReadonly(state))
   for (const key in state) {
     if (typeof state[key as keyof typeof state] === 'function') {
       object[key] = state[key as keyof typeof state]
--- a/packages/vue-query/src/useInfiniteQuery.ts
+++ b/packages/vue-query/src/useInfiniteQuery.ts
@@ -49,6 +49,8 @@ export type UseInfiniteQueryOptions<
           TPageParam
         >[Property]
       >
+} & {
+  shallow?: boolean
 }
 
 export type UseInfiniteQueryReturnType<TData, TError> = UseBaseQueryReturnType<
--- a/packages/vue-query/src/useMutation.ts
+++ b/packages/vue-query/src/useMutation.ts
@@ -2,8 +2,8 @@ import {
   computed,
   getCurrentScope,
   onScopeDispose,
-  reactive,
-  readonly,
+  shallowReactive,
+  shallowReadonly,
   toRefs,
   watch,
 } from 'vue-demi'
@@ -78,7 +78,7 @@ export function useMutation<
     return client.defaultMutationOptions(cloneDeepUnref(mutationOptions))
   })
   const observer = new MutationObserver(client, options.value)
-  const state = reactive(observer.getCurrentResult())
+  const state = shallowReactive(observer.getCurrentResult())
 
   const unsubscribe = observer.subscribe((result) => {
     updateState(state, result)
@@ -101,7 +101,7 @@ export function useMutation<
     unsubscribe()
   })
 
-  const resultRefs = toRefs(readonly(state)) as unknown as ToRefs<
+  const resultRefs = toRefs(shallowReadonly(state)) as unknown as ToRefs<
     Readonly<MutationResult<TData, TError, TVariables, TContext>>
   >
 
--- a/packages/vue-query/src/useMutation.ts
+++ b/packages/vue-query/src/useMutation.ts
@@ -2,6 +2,7 @@ import {
   computed,
   getCurrentScope,
   onScopeDispose,
+  readonly,
   shallowReactive,
   shallowReadonly,
   toRefs,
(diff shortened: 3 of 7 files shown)


DamianOsipiuk:
I'm fine with `shallowReactive`, but `shallowReadonly` would open up a footgun which was prevented multiple times alredy.

People just tried to modify returned data and opened bug reports that they can't.

Maybe we could achieve this conditionally by introducing some kind of `removeSafeguardsIKnowWhatImDoing` prop that would modify the behaviour and return shallow read-only version.

wns3645:
@DamianOsipiuk 
I understand your concern about `readonly`.

In react-query, the `data` object returned by `useQuery` is also not immutable. Issues about mutating properties in data directly can occur in `react-query`, too. I don't know the exact reason that `react-query` doesn't provide deeply nested readonly `data` object, but I think always providing readonly object seems not feasible in general.
My opinion is that preventing only directly mutating `data.value` via `shallowReadonly` can be sufficient to avoid unexpected triggering reactivity effects while providing the return value of frozen objects similar to `react-query`.


Otherwise, providing kind of `removeSafeguardsIKnowWhatImDoing` option also make sense. For example, in `vus-use`, some async state composables provide [`shallow`](https://github.com/vueuse/vueuse/blob/main/packages/core/useAsyncState/index.ts#L57-L62) option that can control the return value will be shallow or not. (default is `shallow: true` because of [the performance for large dataset](https://vueuse.org/guidelines.html#general))

teleskop150750:
it would be nice to mark readonly only in development mode

[an example from vue props](https://github.com/vuejs/core/blob/af60e3560c84e44136f950fc3d0e39b576098c6c/packages/runtime-core/src/componentProps.ts#L656)

teleskop150750:
will there be an advantage if you use shallowRef right away?

teleskop150750:
will there be an advantage if you immediately use shallowRef, without casting toRefs?

teleskop150750:
@wns3645 

```ts
export function useState<T>(
  initialValue: T | null = null,
): [Readonly<Ref<T>>, Function] {
  let value = initialValue
  let _trigger = () => {}

  const = updateState => (v)  = {
    if (value !== v) {
      value = v
      _trigger()
    }
  }

  const state = customRef((track, trigger) => {
    _trigger = trigger
    return {
      get() {
        track()
        return value
      },
      set() {
        console.error('Readonly')
      },
    }
  })

  return [state as Readonly<Ref<T>>, updateState] as const
}
```


```ts
const [state, updateState] = useState()
```

wns3645:
> will there be an advantage if you immediately use shallowRef, without casting toRefs?

In `useBaseQuery` and `useMutation`, using shallowReactive and toRefs make it easy to convert each property(data, isLoading, isFetching, ...) of return value of `observer.getCurrentResult()` being ref. 

> It can also be used in [useMutationState](https://github.com/TanStack/query/blob/74f10b773f2b1b5f5558d9db3ade0ba1e364fb05/packages/vue-query/src/useMutationState.ts#L74)

@teleskop150750 
Applied shallowRef and shallowReadonly to useMutationState, too.-> https://github.com/TanStack/query/pull/7733/commits/dcd4e58e8de4cc269dca598c5721ab596e8baeea

teleskop150750:
> > will there be an advantage if you immediately use shallowRef, without casting toRefs?
> 
> In `useBaseQuery` and `useMutation`, using shallowReactive and toRefs make it easy to convert each property(data, isLoading, isFetching, ...) of return value of `observer.getCurrentResult()` being ref.
> 
> > It can also be used in [useMutationState](https://github.com/TanStack/query/blob/74f10b773f2b1b5f5558d9db3ade0ba1e364fb05/packages/vue-query/src/useMutationState.ts#L74)
> 
> @teleskop150750 Applied shallowRef and shallowReadonly to useMutationState, too.-> [dcd4e58](https://github.com/TanStack/query/commit/dcd4e58e8de4cc269dca598c5721ab596e8baeea)

why didn't you want to use readonly only in dev mode or customRef

wns3645:
> why didn't you want to use readonly only in dev mode?

@teleskop150750 
Using deep readonly in dev mode also causes performance issues when we want to test the large dataset in dev mode. I think that's why the vue core example you gave also uses shallow readonly.
Using shallow readonly also in production mode may not be harmful to the performance.


@DamianOsipiuk , any thoughts on this PR? Should be closed or can be complemented?

teleskop150750:
In the examples given, vue/core, I also meant shallowReadonly. Still, I'm inclined to make it only in dev mode. The vue/core example is a good argument. In any case, shallowReadonly will not eliminate some violations of the rules specified in the documentation. Therefore, it makes no more sense to send it into production. Some tanstack/query implementations in other frameworks do not worry about this and leave it to the developer's responsibility.

DamianOsipiuk:
I would still like to get `shallow` option and retain `deep readonly` by default.
To prevent inexperienced devs mutating the cache by example using `v-model` on part of the data.

Stripping `readonly` completely from `prod` bundle seems ok.

teleskop150750:
example: if you use ` shallowRef` directly without wrapping it in `Proxy` and using `toRefs`

```ts
// packages/vue-query/src/useBaseQuery.ts
// 105 line
 
const observer = new Observer(client, defaultedOptions.value)

type WrapInRef<T> = {
  [K in keyof T]: T[K] extends Function ? T[K] : Ref<T[K]>;
};

type State = WrapInRef<QueryObserverResult<TData, TError>>;

const rawState = observer.getCurrentResult()

const state = rawStateToState(rawState)

function rawStateToState(payload: QueryObserverResult<TData, TError>): State {
  const ret: any = {}

  for (const key of Object.keys(payload) as Array<keyof QueryObserverResult<TData, TError>>) {
    if (typeof payload[key] === 'function') {
      ret[key] = payload[key]
    } else {
      ret[key] = shallowRef(payload[key])
    }
  }

  return ret
}

let unsubscribe = () => {
  // noop
}

// .....

const object = state as unknown as UseBaseQueryReturnType<TData, TError>

object.suspense = suspense
object.refetch = refetch

return object as DeepReadonly<UseBaseQueryReturnType<TData, TError>>
```

```ts
function updateQueryState(payload: State, update: QueryObserverResult<TData, TError>) {
  for (const key of Object.keys(update) as Array<keyof QueryObserverResult<TData, TError>>) {
    if (typeof update[key] === 'function') {
      continue
    }
    (payload[key] as any).value = update[key]
  }
}
```

DamianOsipiuk:
@wns3645 Could you check if latest package from this PR works well with `shallow` flag in your project?
`pnpm add https://pkg.pr.new/@tanstack/vue-query@7733`

wns3645:
@DamianOsipiuk 
Tested it with `shallow: true` option in my project and can confirm it works well 👍

gu-stav:
Thank you for this 🌻 

I am wondering if there is a specific reason why the `shallow` flag hasn't been added to the global `VueQueryPluginOptions`, so that it can be set for all queries at once?

coder-xiaotian:
Is there a configuration to turn off this feature, I want the entire object to be reactive.

teleskop150750:
> Is there a configuration to turn off this feature, I want the entire object to be reactive.

Why do you need this?

coder-xiaotian:
> > Is there a configuration to turn off this feature, I want the entire object to be reactive.
> 
> Why do you need this?

I usually need to modify the data's field, but now, I always need to define a other ref data out of useQuery like this.
image

teleskop150750:
> > > Is there a configuration to turn off this feature, I want the entire object to be reactive.
> > 
> > 
> > Why do you need this?
> 
> I usually need to modify the data's field, but now, I always need to define a other ref data out of useQuery like this. image

https://github.com/posva/pinia-colada/discussions/113#discussioncomment-11311927

lozinsky:
Why does the `shallow` option only change the strategy for applying readonly (`shallowReadonly`/`readonly`)?

1. it seems that the description in https://github.com/TanStack/query/pull/8692 changes doesn't really reflect the truth either
2. without the `shallow: true` option, it's easy to catch the bug because it will only reproduce in a production build https://github.com/TanStack/query/issues/8055

mary-ext:
is there a reason why `shallow` is an optional config? from what I can tell the deep reactivity doesn't really do anything anyway unless the `data` property is being reconciled, which it isn't, it's always a top-level assignment

PR #7780, opened 2024-07-24
fix(eslint-plugin): strictly type config keys

favna:
This is a follow-up from #7766.

While the changes made in #7766 are a good step forward to having a properly typed config, it still doesn't fully work and properly represent the actual code. This PR aims to change that.

With the changes from #7766 pulled in users still run into a typing issue as shown below. This is because of the union of `ESLint.ConfigData | Linter.FlatConfig` cannot be spread, because they are objects and not arrays. By strictly specifying the types of the different keys this can be resolved. 

ishare-1721808850

With this change the config is fully compatible and no `ts-expect-error` / `ts-ignore` will be required:

ishare-1721809149

### DIFF
--- a/packages/eslint-plugin-query/src/index.ts
+++ b/packages/eslint-plugin-query/src/index.ts
@@ -6,10 +6,10 @@ type RuleKey = keyof typeof rules
 
 interface Plugin extends Omit<ESLint.Plugin, 'rules'> {
   rules: Record<RuleKey, RuleModule<any, any, any>>
-  configs: Record<
-    'recommended' | 'flat/recommended',
-    ESLint.ConfigData | Linter.FlatConfig | Array<Linter.FlatConfig>
-  >
+  configs: {
+    recommended: ESLint.ConfigData | Linter.FlatConfig
+    'flat/recommended': Array<Linter.FlatConfig>
+  }
 }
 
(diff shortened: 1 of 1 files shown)


packages/eslint-plugin-query/src/index.ts
```diff
@@ -6,10 +6,10 @@ type RuleKey = keyof typeof rules
 
 interface Plugin extends Omit<ESLint.Plugin, 'rules'> {
   rules: Record<RuleKey, RuleModule<any, any, any>>
-  configs: Record<
-    'recommended' | 'flat/recommended',
-    ESLint.ConfigData | Linter.FlatConfig | Array<Linter.FlatConfig>
-  >
+  configs: {
+    recommended: ESLint.ConfigData | Linter.FlatConfig
```
AhmedBaset, line 10: Is `recommended` of type `FlatConfig`?

I don't think so

`plugins` field in the legacy config is `string[]` but in flat config it's `Record<string, Plugin>`

favna: Good point. Fixed it now. I stupidly copied it from the old types without considering if it was actually correct.

TkDodo (approved):

PR #7799, opened 2024-07-26
fix(query-core): abort fetch loop for infinite queries if getNextPageParam returns null or undefined

TkDodo:
The `fetchPage` function has a safeguard where it only returns the current data if pageParam == null, however, this means we still stay in the loop and call `getNextPageParam` unnecessarily.

This can be troublesome if you set `pages: Infinity` on queryClient.fetchInfiniteQuery to fetch an arbitrary amount of pages until the natural end is reached by returning null/undefined from getNextPageParam, because it would never actually escape the loop


mrlubos:
Thank you for the quick fix!

PR #7806, opened 2024-07-27
refactor(solid-query): separate queryOptions from createQuery file

lachlancollins:
- The other framework adapters have the queryOptions function separate to the useQuery/createQuery function
- This is an internal refactor to separate queryOptions (and infiniteQueryOptions) into its own file
- This makes it easier to ensure that logic/types are kept consistent between adapters (e.g. it doesn't have the `DefinedInitialDataOptions` fix from #5869)
- Also makes it easier to directly compare svelte-query to solid-query (API is becoming a _lot_ more similar with svelte runes!)


ardeora (approved):
Looking good @lachlancollins! Thanks for making the changes. Feel free to merge this in.

PR #7857, opened 2024-08-05
feat(query-core): add support for symbol in cache

Ayc0:
I noticed when I was using symbols as keys in the value of objects with `queryCache.setQueryData()` that keys containing symbol were sometimes stripped out

This PR addresses this by using `Object.getOwnPropertySymbols()` to iterate over symbol keys in the cache

I added new tests & ran test before and after my changes and they indeed fail on before, and pass after

Before | After
-|-
image | image

### DIFF
--- a/packages/query-core/src/__tests__/queryCache.test.tsx
+++ b/packages/query-core/src/__tests__/queryCache.test.tsx
@@ -347,4 +347,28 @@ describe('queryCache', () => {
       expect(queryCache.getAll().length).toEqual(1)
     })
   })
+
+  describe('Symbol', () => {
+    test('should allow setting in the cache', () => {
+      const key = queryKey()
+
+      const s = Symbol('s')
+
+      queryClient.setQueryData(key, { [s]: 1 })
+      expect(queryClient.getQueryData(key)).toEqual({ [s]: 1 })
+    })
+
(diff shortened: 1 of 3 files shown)


TkDodo:
sorry, symbols aren't supported in query keys because they are't json serializable, which is what we use for hashing. You can address that on your end by providing a `queryKeyHashFn` that hashes those into something unique. This can also be provided globally.

Ayc0:
But those aren't in query types. But in the cached values

Ayc0:
You can see here in the code:

```ts
const key = queryKey()
const s = Symbol('s');

queryClient.setQueryData(key, { [s]: 1, a: 1 })
```

The symbol is not in the query key but in the data itself

TkDodo:
ah, sorry about that. So you basically changed the implementation of `replaceEqualDeep` ? This feature is for structural sharing, which again, also only supports json serializable structures per default. This too can be customized by passing your own `structuralSharing: (old, new) => merged` function.

---

The reason why I'm hesitant to add this to the library is because once we go away from "must be json serializable", we have to (and already did get) many requests about Maps, Sets, Dates etc. [superjson](https://github.com/flightcontrolhq/superjson) is generally really good at that.

Ayc0:
Makes sense, I'll check that then

Issue #7872, opened 2024-08-08
DefaultError type fallback type has too strong guarantees


bvanderdrift:
### Describe the bug

The type of `error` is by default `Error`. When using `persistQueryClient` it will serialize to JSON. `Error` is not serializable. `JSON.stringify(new Error('some message'))` will return `"{}"`.

This leads persisted errors to be of type `{}`, but `useQuery` defaults to `Error`. Therefore the type fallback is incorrect.

### Your minimal, reproducible example

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgVwM4FMCKz1QJ5wC+cAZlBCHAORToCGAxjALQCOO+VAUFwxAHap4adADEIEOAF44ACgCU0gHyIucdXD6D47XARkjse2QjUbzu-AGl0eAFxwA2lRISqAXQA0Z8+st5RfgcFZVVfcLgYAAtyAHc4fnR4gFEocihZKhh0ISp5AG4fX0Iiwnkec2ASWX8AOlx02pAc1DoAc3R5UwjNAVQIABt0WoGINpqOPHq06CaW9s7C8xKi2hhkKH44f0KVriA

### Steps to reproduce

1. use `useQuery`
2. look at the type of `error`

### Expected behavior

The type of `error` defaults to `Error`, expected `unknown`.

### How often does this bug happen?

Sometimes

### Platform

- OS: Android, iOS
- Framework: React Native

### Tanstack Query adapter

react-query

### TanStack Query version

5.51.16

### TypeScript version

5.0.4

### Additional context

I actually notice the types are as expect in my reproducable code, but looking at the code of the latest version of this package the fallback is still `Error`.

bvanderdrift:
I've submitted a PR with a potential solution: https://github.com/TanStack/query/pull/7873

TkDodo:
sorry, we're not gonna revert back to `unknown`. The `Error` type assumption is "optimistic", yes, because it holds true for 99% of cases. It also falls apart if you `trow 25` in your `queryFn`, because `25` is also not an `Error`.

We've documented how you can opt out of that typings here:

https://tanstack.com/query/v5/docs/framework/react/typescript#registering-a-global-error

Basically, you want:

```
import '@tanstack/react-query'

declare module '@tanstack/react-query' {
  interface Register {
    defaultError: unknown
  }
}
```

[...]

PR #7873, opened 2024-08-08
fix: false error type assumption

bvanderdrift:
This goes in hand with #7872.

I realise this might be an oversimplified solution, but didn't want to show up empty handed. Curious to hear feedback on how to properly implement this solution.


TkDodo:
sorry, we're not gonna merge this. Please look at the reasoning here:

- https://github.com/TanStack/query/issues/7872#issuecomment-2277443070

PR #7883, opened 2024-08-10
feat(query-core): export partialMatchKey util

raidenmiro:
#4355 

This pull request exports the utility, which is very useful for a key match.


TkDodo (approved):

TkDodo:
sorry I've been thinking about it some more and isn't it enough that we already export the `machQuery` function?

TkDodo:
closed due to staleness

Issue #7887, opened 2024-08-12
docs: update overview documentation code to display error when error occurs


Rasbin:
### Describe the bug

The example code in the overview does not display error in the UI if there is error in response. 

### Your minimal, reproducible example

https://tanstack.com/query/latest/docs/framework/react/overview

### Steps to reproduce

1. Go to https://tanstack.com/query/latest/docs/framework/react/overview
2. Update the url link to "https://api.github.com/repos/TanStack/queryX"

Or, just run this code from the overview example : 
```Typescript
import {
  QueryClient,
  QueryClientProvider,
  useQuery,
} from '@tanstack/react-query'

const queryClient = new QueryClient()

export default function App() {
  return (
    <QueryClientProvider client={queryClient}>
      <Example />
    </QueryClientProvider>
  )
}

function Example() {
  const { isPending, error, data } = useQuery({
    queryKey: ['repoData'],
    queryFn: () =>
      fetch('https://api.github.com/repos/TanStack/queryX').then((res) =>
        res.json(),
      ),
  })

  if (isPending) return 'Loading...'

  if (error) return 'An error has occurred: ' + error.message

  return (
    <div>
      <h1>{data.name}</h1>
      <p>{data.description}</p>
    </div>
  )
}

### Expected behavior

With the wrong url, this should result in error and display the error message in the UI, but it does not. The overview code can be updated with adding a check if the response is ok or not and throwing the error if it is not ok.

### How often does this bug happen?

Every time

### Screenshots or Videos

_No response_

### Platform

OS: macOS
Browser: Chrome

### Tanstack Query adapter

react-query

### TanStack Query version

v5.51.23

### TypeScript version

v5.3.3

### Additional context

_No response_
```[tasklist]
### Tasks
```

TkDodo:
I think it's fine to keep the overview minimal as this is just how `fetch` works. We have lots of other examples and notes in the docs about treating non-2xx status codes as errors.

PR #7888, opened 2024-08-12
docs(react-query): update overview code to display error when error in response

Rasbin:
Fixes : https://github.com/TanStack/query/issues/7887

Current behaviour: The example code in the [overview documentation](https://tanstack.com/query/latest/docs/framework/react/overview) does not display error in the UI if there is error in response.

Expected behaviour: When there is error for example due to wrong url, the error message should be displayed in the UI.

This PR updates the overview documentation so that the error message gets displayed in the UI if there is error in response.


Rasbin:
@TkDodo Could you please have a look?

PR #7901, opened 2024-08-14
Update mutations.md - add retry context and include retry delay info

h8ku:
I had to look up another doc  to find this, having this here would help others in the future


TkDodo:
please refer to the api docs to see all possible options

https://tanstack.com/query/v5/docs/framework/react/reference/useMutation

Issue #1858, opened 2021-02-23
useMutation returns a new object on every call


ashleymercer:
**Describe the bug**

If I'm reading the code right, I think the `useMutation` function creates a new object on every call - so if this object is used directly in the dependency array for any further React hooks, it will trigger additional renders.

**To Reproduce**

```typescript
function FooComponent() {
  const callback    = useCallback(() => Promise.resolve("done"), []);
  const mutation    = useMutation(callback);
  const mutationRef = useRef(mutation);

  useEffect(() => {
    if (mutationRef.current !== mutation) {
      console.log(`This should never get logged`);
    }
  })

  return <></>
}
```

**Expected behavior**

The log line should never fire: the arguments to `useMutation` never change so I would expect the same reference back each time (I think this is similar to how `useQuery` works - at least, I don't seem to have the same issue there).

**Additional context**
Looking at the [current code](https://github.com/tannerlinsley/react-query/blob/master/src/react/useMutation.ts#L109) would it be sufficient to memoise the returned object on `currentResult` and `mutate`? Or would that likely cause further problems?

bjarkehs:
The object returned from `useMutation` always changes, but `mutate` at least never changes. I'm not sure it's expected behaviour that the object returned from `useMutation` never changes. I think most hooks work this way.

TkDodo:
I think it’s expected. useQuery behaves the same now: the functions / objects on the returned object are referentially stable if you need to pass them to memoized components or add as dependency to an effect, but the actual object is not.

thoo1:
relatively new to using this library, is there a reason for this behaviour? that the returned object is not stable but the values inside the object are?

TkDodo:
It was introduced to make the library concurrent-mode safe. See here: https://github.com/tannerlinsley/react-query/pull/1775

> Unfortunately I was not able to keep the referential integrity of the result object between renders because we do not know which optimistic result will eventually get committed.

it’s a minimal tradeoff :)

thoo1:
makes sense, thanks!

tschanz-pcv:
@TkDodo Thanks a lot for sharing this info. Maybe this is common knowledge among React devs, I certainly didn't expect this. Could you maybe add something to the docs regarding referential stability so it's clear what can be used in deps arrays e.g. for useEffect.

TkDodo:
Sure, I’d accept a PR to the docs :)

[...]

PR #7911, opened 2024-08-18
feat(eslint-plugin-query): add rule that disallows putting the result of query hooks directly in a React hook dependency array

schiller-manuel:
I stumbled across https://github.com/TanStack/query/issues/1858 a few times, so I wrote this eslint plugin.
It will detect if the return value of `useMutation`, `useQuery` etc. is directly put in a React hook dependency array.


TkDodo (commented):
Thank you 🙏 . The same reasoning applies to the return value of `useQuery`, `useSuspenseQuery`, `useQueries`, `useSuspenseQueries`, `useInfiniteQuery` and `useSuspenseInfiniteQuery` as well. Do you think it would make sense to unify this into a single rule?

schiller-manuel:
@TkDodo  I generalized the rule, please have a look.

maybe there is a better name though ...

docs/config.json
```diff
@@ -792,6 +792,10 @@
         {
           "label": "No Rest Destructuring",
           "to": "eslint/no-rest-destructuring"
+        },
+        {
+          "label": "No Unstable Query/Mutation in Deps",
+          "to": "eslint/no-unstable-query-mutation-in-deps"
```
TkDodo, line 798: yeah let's just name it `no-unstable-deps` ?

TkDodo (approved):

PR #7913, opened 2024-08-18
fix(query-core): shorten listeners code

manudeli:


packages/query-core/src/subscribable.ts
```diff
@@ -1,10 +1,7 @@
-type Listener = () => void
-
-export class Subscribable<TListener extends Function = Listener> {
```
manudeli, line 3: There are no case using default Listener of TListener

packages/query-core/src/subscribable.ts
```diff
@@ -1,10 +1,7 @@
...
-  protected listeners: Set<TListener>
+export class Subscribable<TListener extends Function> {
+  protected listeners = new Set<TListener>()
```
manudeli, line 2: I shorten listeners code

manudeli (commented):
I left intention

TkDodo (approved):

PR #7914, opened 2024-08-18
fix(query-core): type override subscribe in InfiniteQueryObserver mor…

manudeli:
…e clearly

### DIFF
--- a/packages/query-core/src/infiniteQueryObserver.ts
+++ b/packages/query-core/src/infiniteQueryObserver.ts
@@ -4,6 +4,7 @@ import {
   hasPreviousPage,
   infiniteQueryBehavior,
 } from './infiniteQueryBehavior'
+import type { Subscribable } from './subscribable'
 import type {
   DefaultError,
   DefaultedInfiniteQueryObserverOptions,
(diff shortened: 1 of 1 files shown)


packages/query-core/src/infiniteQueryObserver.ts
```diff
@@ -38,9 +39,9 @@ export class InfiniteQueryObserver<
   TQueryKey
 > {
   // Type override
-  subscribe!: (
-    listener?: InfiniteQueryObserverListener<TData, TError>,
-  ) => () => void
+  subscribe!: Subscribable<
+    InfiniteQueryObserverListener<TData, TError>
+  >['subscribe']
```
manudeli, line 44: When doing type overriding, I want to use Subscribable and make it clearer to override while keeping the constraints of Subscribable, so I think it reduces the possibility of making constraint mistakes.

TkDodo: shouldn't we do the same for other type overrides, like `fetch` a couple of lines below as well?

manudeli: I tried this on other type overrides(getCurrentResult, fetch) in [ccb4a62](https://github.com/TanStack/query/pull/7914/commits/ccb4a62a84e23bb8cc56ce6275033911de5978cc)

packages/query-core/src/queryObserver.ts
```diff
@@ -34,7 +34,7 @@ export interface NotifyOptions {
   listeners?: boolean
 }
 
-export interface ObserverFetchOptions extends FetchOptions {
+interface ObserverFetchOptions extends FetchOptions {
```
manudeli, line 37: I removed exporting because we don't use it in other files

TkDodo (approved):

PR #7916, opened 2024-08-18
type(query-core): add Override

manudeli:


packages/query-core/src/types.ts
```diff
@@ -20,6 +20,12 @@ export type OmitKeyof<
   TStrictly extends 'strictly' | 'safely' = 'strictly',
 > = Omit<TObject, TKey>
 
+export type Override<TTargetA, TTargetB> = {
+  [AKey in keyof TTargetA]: AKey extends keyof TTargetB
+    ? TTargetB[AKey]
+    : TTargetA[AKey]
+}
```
manudeli, line 27: I extract useful Override type from each framework packages to @tanstack/query-core to share it

TkDodo (approved):

PR #7922, opened 2024-08-20
fix(query-core): update Subscribable's method hasListeners as protected

manudeli:


packages/query-core/src/subscribable.ts
```diff
@@ -16,7 +16,7 @@ export class Subscribable<TListener extends Function> {
     }
   }
 
-  hasListeners(): boolean {
+  protected hasListeners(): boolean {
```
manudeli, line 19: There is no use case using hasListeners as public method.

TkDodo: we're about to use it in a separate PR 😅

PR #7935, opened 2024-08-21
fix(query-core): convert notifyManager from factory function to class

manudeli:
It's not must, but FocusManager, OnlineManager is class. so I thought that NotifyManager should be class consistently


TkDodo:
I remember moving this to a function because it was significantly smaller in bundle size. Can you check the size differences please?

manudeli:
> I remember moving this to a function because it was significantly smaller in bundle size. Can you check the size differences please?

Okay, I'll check

manudeli:
## Size differences
modern: 773,829 bytes -> 775,039 bytes (+1210 bytes)
modern/notifyManager.js: 1,534 bytes -> 2,193 bytes (+659 bytes)

The JavaScript bundle size will increase, but I don't think it will increase significantly. We can expect library users to minify notifyManager through their bundlers for each service, and if so, we can expect the difference to be even smaller.

### AS IS: factory function
image
image

### TO BE: class
image
image

## JSDoc differences (What I want to focus)

But In my opinion, the biggest difference, other than the size, is that jsdoc in notifyManager is not included when it is a factory function. notifyManager is a public API of @tanstack/query-core and I thought it would be a more useful choice to keep the jsdoc where each method is used rather than having it disappear.

### AS IS: factory function (jsdoc of methods will be disappeared)
setBatchNotifyFunction or other methods' jsdoc will be removed
image

### TO BE: class (jsdoc of methods will be restored)
setBatchNotifyFunction or other methods' jsdoc will be restored
image

I don't think it's necessarily a matter of which way to go.
@TkDodo I'm curious about your opinion on this.

TkDodo:
I think we can rather move the jsdoc, or even inline the functions into the return statement:

```
return {
  batch: <T>(callback: () => T): T => { ... }
}
```

we also don't really have any meaningful jsdocs here.

PR #7942, opened 2024-08-22
fix(vue-query): mark the watcher as post flush

RomainLanz:
Hey there! 👋🏻 

This PR changes the watcher mode for `useQueries` from `sync` to `post`.
The change is needed to avoid having the watcher run during component unmount.

> [!NOTE]
> This change will probably be needed on `useBaseQuery`, but I haven't tested it yet.

Related VueJS issues:
- https://github.com/vuejs/core/issues/2291
- https://github.com/vuejs/core/issues/7030
- https://github.com/vuejs/router/issues/1945
- https://github.com/vuejs/vue-router/issues/3393
- https://github.com/posva/unplugin-vue-router/issues/86

---

Currently, using the route params to define some query params triggers a new fetch when leaving a route.

For example, let's say we have the following code that runs when we are on `/players/:id`.
```ts
const route = useRoute();
const paramId = computed(() => +route.params.id);

const queries = useMyViewData(paramId)
```

- When going from `/players` to `/players/1`, TanStack Query will trigger a fetch for Player `1` **(expected)**.
- When going from `/players/1` to `/players`, TanStack Query will trigger a fetch for Player `undefined` **(unexpected)**.
- When going from `/players/1` to `/transfers/12345`, TanStack Query will trigger a fetch for Player `12345` **(unexpected)**.

This is due to the watcher running when the page changes.

I haven't created an easy and sharable reproduction. Let me know if it is needed.

### DIFF
--- a/packages/vue-query/src/useQueries.ts
+++ b/packages/vue-query/src/useQueries.ts
@@ -341,7 +341,7 @@ export function useQueries<
       )
       state.value = getCombinedResultPersisted()
     },
-    { flush: 'sync' },
+    { flush: 'post' },
   )
 
   onScopeDispose(() => {


RomainLanz:
pnpm patch for the one who wants to test it:

```diff
diff --git a/build/modern/useQueries.js b/build/modern/useQueries.js
index c172ecd85df4646ab4869411725cc028ccc0e7e1..ec584988a4ecb9f24c073eb94a4b5cbc9e8eb255 100644
--- a/build/modern/useQueries.js
+++ b/build/modern/useQueries.js
@@ -81,7 +81,7 @@ function useQueries({
       );
       state.value = getCombinedResultPersisted();
     },
-    { flush: "sync" }
+    { flush: "post" }
   );
   onScopeDispose(() => {
     unsubscribe();

```

RomainLanz:
We deployed this patch to production on Monday and found no issues for the moment.

DamianOsipiuk:
We could potentially remove sync watcher entirely.
We did it for other hooks some time ago.

https://github.com/TanStack/query/pull/6043

RomainLanz:
I'm not entirely sure about the implications either. If removing this `watch` statement is the only change, I can create another patch and test how my application behaves.

DamianOsipiuk:
You would also need to override `refetch` to update observer options before actual refetch is called.

RomainLanz:
Then we could merge that one first and work towards removing all watchers in a follow-up PR/issue.

DamianOsipiuk:
I do not think merging this and immediately changing solution in a follow-up is the way to go.  
Better way would be to fix it properly and add automated test that verifies the fix.  
For the mean time, you can patch the package locally, as you already did.

RomainLanz:
I might be misunderstanding, but I think we’re talking about different things.

This PR is updating a watcher for the `useQueries` hook, and there’s no refetch concept involved at this level. The refetch system is handled within the underlying `QueriesObserver` layer.

DamianOsipiuk:
> Could someone review this again?
> 
> This issue affects all Vue applications using useQueries. It's a significant problem that shouldn't be disregarded.

As mentioned before i do not think this change should be merged as it will introduce another bug in specific cases where states would not be properly synced up when calling `refetch`.

There is a similar PR that handled that for other hooks, and this MR should be adjusted to follow similar pattern. https://github.com/TanStack/query/pull/6043

In that MR you could see additional block that patches `refetch` function returned from `observer` - https://github.com/TanStack/query/pull/6043/files#diff-996721944b77ab3155cb5f86364575311132aeaaff8f7f715755bfcc9fc09108R124-R128

Mini-ghost:
I have opened a PR(#8443) to address this issue. If there are any areas where I may have overlooked something, please kindly let me know.

Thank you!

PR #7952, opened 2024-08-25
Throw-on-server-default-fix

buddamus:
fix: stop defaulting to throwOnError=true when isServer===true so that SSR can render the page the same way the client will rather than showing the error boundary fallback


TkDodo:
right, maybe solid is different to react here. @ardeora FYI

PR #7966, opened 2024-08-27
fix(query-core): handle errors that occur inside `setData` method

Efimenko:
As [PR](https://github.com/TanStack/query/pull/7642) is a bit stuck, this PR implements the same, with respect to the comments left in the first PR. Fix https://github.com/TanStack/query/issues/6954


Efimenko:
@TkDodo Also, currently, after the second fetch, the error will be `maximum call stack size exceeded.` So, probably, we need to check if the data is serializable or not, rather than wait `call stack size exceeded`. Inside `replaceEqualDeep`, some kind of:
```js
try {
  JSON.stringify(data)
} catch {
  throw new Error('Data is not serializable')
}
```

packages/query-core/src/__tests__/query.test.tsx
```diff
@@ -965,4 +965,44 @@ describe('query', () => {
...
     expect(spy).toHaveBeenCalledWith('1 - 2')
   })
+
+  it('should have an error status when queryFn data is not serializable', async () => {
+    const key = queryKey()
+
+    const queryFn = vi.fn()
+
+    queryFn.mockImplementation(async () => {
+      await sleep(1000)
```
TkDodo, line 975: since we run with real timers:

```suggestion
      await sleep(10)
```

packages/query-core/src/__tests__/query.test.tsx
```diff
@@ -965,4 +965,44 @@ describe('query', () => {
...
+    })
+
+    await queryClient.prefetchQuery({ queryKey: key, queryFn })
+
+    const query = queryCache.find({ queryKey: key })!
+
+    expect(queryFn).toHaveBeenCalledTimes(1)
+
+    // NOTE: need second query fetch, cause for first fetch 'prevData' for structural sharing is 'undefined'
+    await query.fetch()
```
TkDodo, line 1004: this is weird. are you saying the query goes in to success state first, and only after a refetch, it will go into error state ?

packages/query-core/src/query.ts
```diff
@@ -499,7 +499,11 @@ export class Query<
           return
         }
 
-        this.setData(data)
+        try {
+          this.setData(data)
+        } catch (error) {
+          onError(error as TError)
+        }
```
TkDodo, line 506: note: we're missing a `return` here; otherwise, we'll go on with calling the `onSuccess` callback, which isn't good if we're in error state.

packages/query-core/src/__tests__/query.test.tsx
Efimenko: yes, it goes in a success state because on the first prefetch, this condition is not `true`: https://github.com/TanStack/query/blob/main/packages/query-core/src/utils.ts#L251
Because `this.state.data` will be `undefined` for the first prefetch: https://github.com/TanStack/query/blob/main/packages/query-core/src/query.ts#L215

packages/query-core/src/query.ts
Efimenko: sure, my bad

packages/query-core/src/__tests__/query.test.tsx
TkDodo: right, makes sense. I want to avoid having to do a safety check like you showed with `try JSON.stringify`, because it would add and additional serialization step for everyone who sends "normal" data.

What we can do is add this check to `replaceEqualDeep` behind an env check, like we do here: 

https://github.com/TanStack/query/blob/bd3296bfed9b4db1882e42fe378f2f35dc240381/packages/query-core/src/hydration.ts#L85-L89

That would also make sure that we don't need two fetches in the test. Yes, in production it would still work for the first fetch and only error on the second, but I think that's okay

Efimenko: kk, will implement

Efimenko: @TkDodo please have a look when you have time

packages/query-core/src/utils.ts
```diff
@@ -246,6 +246,15 @@ export function replaceEqualDeep(a: any, b: any): any {
     return a
   }
 
+  if (process.env.NODE_ENV !== 'production') {
+    try {
+      JSON.stringify(a)
+      JSON.stringify(b)
+    } catch {
+      throw new Error('Values are not serializable')
```
TkDodo, line 254: let's please add a good `console.error` log here, like we do in other places as well. Otherwise, it's up to the user to inspect this error and it might get unnoticed.

this likely means that the test will need to stub the console and assert that the log happened. We have some good examples of this already

TkDodo (changes requested):

packages/query-core/src/utils.ts
Efimenko: @TkDodo, Okay. I just thought that the idea of this catch was an early throw of error instead of waiting until the call stack is exceeded and the query transitions to an error state.
So, it's rather just a hint for the developer in this case, right?

TkDodo: I definitely want the error log. I don't really care if we throw or not 🤷‍♂️  (in dev mode).

Efimenko: @TkDodo ok, I've added the `console.error`, but also left the `throw` to initiate the early transition to `error` status. Also, I lifted this check a bit to access the query `options`.

TkDodo (approved):
thanks. I adapted the message a bit

packages/query-core/src/utils.ts
```diff
@@ -353,6 +353,19 @@ export function replaceData<
   if (typeof options.structuralSharing === 'function') {
     return options.structuralSharing(prevData, data) as TData
   } else if (options.structuralSharing !== false) {
+    if (process.env.NODE_ENV !== 'production') {
+      try {
+        JSON.stringify(prevData)
+        JSON.stringify(data)
```
ethos-vitalii, line 359: This now throws an error when data contains BigInt. But it seems like react-query supports having BigInt in the query result. Should I create an issue for this?

TkDodo: I'm think it's still good to log the error because structuralSharing creates an overhead and you should likely turn it off if you have something that isn't serializable. But we should likely remove the throw. Logging is enough. Please file a PR

ethos-vitalii: Thank you @Efimenko for raising that PR 🙌 
I have a question. From what I see in [`replaceEqualDeep`](https://github.com/TanStack/query/blob/8af72729f0457289e675816453ba9d2d2ed4f604/packages/query-core/src/utils.ts#L243-L281), there's nothing preventing BigInt comparison to break. For example, `123n === 123n` will be `true`. Is there anything else I'm missing where `BigInt` could break`this.setData()`?

TkDodo: interesting. I didn't know that BigInt was referentially equal, but not json serializable. According to [the documentation](https://tanstack.com/query/latest/docs/framework/react/guides/render-optimizations#structural-sharing), we only support json serializable values in `structuralSharing`, so I think the warning is still warranted.

jxom: Yeah, this seems to have affected a lot of upstream Wagmi users. Wonder if this should have been a breaking change.

TkDodo: It's "just" a dev mode warning. Why would it be breaking?

jxom: Apologies. Didn't see [this commit](https://github.com/TanStack/query/commit/fb9d4f6138645a4812d7590a39476f880d181b40#diff-9701bc1ef4ed4515a08de71beb3a01aac6de69115ceda88ad67995f1d5feab8cL366).

jxom: @TkDodo – would you be open to a PR that checked for cyclic references instead of checking for non-serializable properties via `JSON.stringify`? Feel like a lot of consumers would be invoking query functions that return `BigInt`s, so it might be a bit odd seeing these warning messages flying around when `replaceEqualDeep` is also compatible with `BigInt`s.

TkDodo: > Wonder if it’s worth conveying that structural sharing is moreso only compatible with referentially stable values instead of JSON-serializable values.

I simply wasn't aware that `replaceEqualDeep` works with `BigInt` - our general recommendation is to use json serializable values. If we say BigInt is supported, then why isn't Map supported, or Set, or Date, or ... it's a slippery slope. Also, in queryKeys, we actually use `JSON.stringify`, so if you use a BigInt there, it will fail.

Why is JavaScript so weird that BigInt is referentially stable across values, but not serializable lol. 

>  would you be open to a PR that checked for cyclic references instead of checking for non-serializable values via JSON.stringify?

yeah I guess that's fine. I think just calling `replaceEqualDeep` itself in a try/catch would just work, too ?

jxom: done! https://github.com/TanStack/query/pull/8030

PR #7970, opened 2024-08-28
docs(react-query): recommend against using server actions in queryFn

### LINKED ISSUE #7934: Next.js server action not getting executed
### Describe the bug

Please bear with me, because I have no idea how to best start. I've also done my best to narrow this down as much as possible, but I haven't been able to provide a minimal reproducible example.

# Setting the stage

We were noticing more and more lately that our (Cypress) tests are flakey so I started to investigate it. I've managed to track it down to a custom hook not resolving with the correct data.

The hook, `useCurrentUser`, is in essence a very simple hook

```tsx

import { getCurrentUserWithRoles } from "@/actions/users";

export const useCurrentUser = () => {
  const { data: currentUserWithRoles } = useQuery({
    queryFn: async () => {
      return getCurrentUserWithRoles();
    },
    queryKey: ["currentUserWithRoles"],
  });

  if (!currentUserWithRoles) {
    return null;
  }

  return currentUserWithRoles;
};
```

`getCurrentUserWithRoles` is also a very simple server action that does some database fetching.

```ts
"use server";

import { AuthDAL } from "@/data/auth";

export async function getCurrentUserWithRoles() {
  return await AuthDAL.getCurrentUserWithRoles();
}
```

However, when running `cypress open` and running a test that relies on the `useCurrentUser` returning correct information, the hook **sometimes** fails to return any data.

# What I've tried

### Adding logs

If I add logs to the necessary places, I can see where it goes wrong

```tsx
    queryFn: async () => {
      console.log("About to get roles");
      const res = getCurrentUserWithRoles();
      console.log("res: ", res);

      return res;
    },
```

```ts
export async function getCurrentUserWithRoles() {
  console.log("Inside server action");
  return await AuthDAL.getCurrentUserWithRoles();
}
```

The first log gets printed, but the other two don't appear. So for some reason, it seems that it never reaches the server action.

### Removing React Query

If I remove React Query and just rely on `useEffect` and `useState`, there's no flakiness and test passes 100% of the time

```tsx
export const useCurrentUser = () => {
  const wasCalled = useRef(false);
  const [currentUserWithRoles, setCurrentUserWithRoles] = useState<Awaited<
    ReturnType<typeof getCurrentUserWithRoles>
  > | null>(null);

  useEffect(() => {
    if (!wasCalled.current) {
      wasCalled.current = true;

      getCurrentUserWithRoles().then((res) => {
        setCurrentUserWithRoles(res);
      });
    }
  }, []);

  if (!currentUserWithRoles) {
    return null;
  }

  return currentUserWithRoles;
};
```

### Replacing the server action with an actual API call

I also went ahead and created an actual API call (via `route.ts`) where I returned hardcoded data. And again, it works 100% of the time.

```ts
import { NextResponse } from "next/server";

export async function POST() {
  return NextResponse.json({
    groupRoles: [
      {
        groupIdId: "77fc6325-96ff-449b-91a8-cd0a8c79e019",
        role: "ADMIN",
      },
      {
        groupId: "4b102177-be72-4910-a46d-90060ca7b178",
        role: "ADMIN",
      },
    ],
    firstName: "Fred",
    id: "33d4cde9-c366-4c77-9669-decdf0e0c6ad",
    lastName: "Flinstone",
    appRole: "ADMIN",
  });
}
```

```tsx
export const useCurrentUser = () => {
  const { data: currentUserWithRoles } = useQuery({
    queryFn: async () => {
      // return getCurrentUserWithRoles();
      const call = await fetch("/api/v1/test", {
        method: "POST",
      });

      const res = await call.json();

      return res;
    },
    queryKey: ["currentUserWithRoles"],
  });

  if (!currentUserWithRoles) {
    return null;
  }

  return currentUserWithRoles;
};
```

### Replacing Cypress with Playwright

I even rewrote the test in Playwright but to no avail. The tests are still just as flaky, feeling even worse (but that may be my mental state acting up by now).

# Some extra insights

However, the switch to Playwright has brought some extra insights that may be helpful

- If I only enable one of the tree browsers in Playwright (so `chromium`, `firefox` and `webkit`), the tests always pass for `chromium` and `firefox`, and fail way less for `webkit` (10% of the times instead of 70%)
- Same scenario for setting `workers` to 1 in Playwright
- I tried [this suggestion](https://github.com/TanStack/query/issues/1657#issuecomment-1912378686) to no avail
- All libraries are at their latest versions

I tried, but can't condense it enough to have it always fail

### Steps to reproduce

/

### Expected behavior

I expect React Query to actually fire the Next.js server action and return the data

Often

### Screenshots or Videos

_No response_

### Platform

- OS: MacOS
- Browser: Multiple

### Tanstack Query adapter

react-query

### TanStack Query version

v5.52.0

### TypeScript version

v5.3.3

### Additional context

Please let me know what other information I can provide. I felt like I've been going down the rabbit hole, with no solution in sight..

bartcheers:
Update the useQuery documentation to clarify that server actions should not be used in `queryFn`. Server actions are meant for mutations, run serially, and cannot execute in parallel, potentially leading to unexpected behavior. Recommend using API routes instead for data fetching.

Closes #7934

### DIFF
--- a/docs/framework/react/reference/useQuery.md
+++ b/docs/framework/react/reference/useQuery.md
@@ -70,6 +70,7 @@ const {
   - The function that the query will use to request data.
   - Receives a [QueryFunctionContext](../../guides/query-functions#queryfunctioncontext)
   - Must return a promise that will either resolve data or throw an error. The data cannot be `undefined`.
+  - Note: It is not recommended to use server actions in `queryFn`. Server actions are intended for mutations, run serially, and can't execute in parallel, which can lead to unexpected behavior when used in `queryFn`. For fetching data, consider using API routes instead.
 - `enabled: boolean | (query: Query) => boolean`
   - Set this to `false` to disable this query from automatically running.
   - Can be used for [Dependent Queries](../../guides/dependent-queries).


TkDodo:
This shouldn't go to the api reference, but to the advanced ssr guide. Also, please link to this paragraph from the react docs:
https://react.dev/reference/rsc/use-server#caveats

bartcheers:
> This shouldn't go to the api reference, but to the advanced ssr guide. Also, please link to this paragraph from the react docs:
> 
> https://react.dev/reference/rsc/use-server#caveats

Happy to update. However, to my understanding calling server actions from the server does not turn them into post requests, and does not make them blocking. 

When people run into this issue they are making the fetch call from the client using the queryfn. I figured this would be the place they'd look first when running into issues.

TkDodo:
> However, to my understanding calling server actions from the server does not turn them into post requests, and does not make them blocking.

That's right, and this isn't what I am proposing. The SSR guide has tons of `usQuery` usages in client components because that's what you need to do after the "handoff" from the server. Have you read it? We already have a NOTE about server actions on that page, I would probably merge the two. "Server actions only work when queryFn isn't a reference" is another reason to not recommend them for fetching data.

> I figured this would be the place they'd look first when running into issues.

It will be more confusing for all other people who don't use next-js and read the useQuery API docs. They might not even know what "server actions" are. Recommendations like that are not for api references (they tell you what's possible), but for guides (they tell you what you should do or avoid).

TkDodo:
@bartcheers it’s been a while. Do you plan on finishing this docs update?

PR #7988, opened 2024-08-30
feat: add support for `React.use()`

KATT:
Closes #7980

---

### Progress

- [x] add basic functionality for `useQuery()`
- [x] make other tests pass (👋  help wanted!)
- [x] add test for `useInfiniteQuery()`
- [x] do more tests with some example
- [x] update docs?


packages/query-core/src/queryObserver.ts
```diff
@@ -582,6 +582,11 @@ export class QueryObserver<
       isRefetchError: isError && hasData,
       isStale: isStale(query, options),
       refetch: this.refetch,
+      promise: (prevQuery.promise ??
+        query.promise ??
+        new Promise(() => {
+          // never resolves? 🤷‍♂️
+        })) as any,
```
TkDodo, line 589: so, I know we are supposed to cache the promise, but is this also true for immediately resolved promises? Like, I'm thinking: If we have `data` (`hasData`), then we should _always_ immediately resolve with it. There are too many cases where this could be set, like with `keepPreviousData` or when we have stale data, or `initialData`, ...

So the only time where we want to integrate with suspense / error boundaries is on hard load. So would this work?

```
promise: hasData
  ? Promise.resolve(data)
  : (query.promise ?? new Promise(() => {})),
```

KATT: Yeah, the promise should be a stable reference to `Promise.resolve(data)` when it's already resolved

I just wanted to see if I could prototype get _anything_ working, knowingly it was wrong. Didn't have much time for it though

KATT:
Will close this as I realistically won't work on it for a week or so at least

(I had some issues following CONTRIBUTING.md btw, some of the builds didn't work and I struggled a bit to setup a vitest watcher that wasn't slow only for this particular test)

packages/react-query/package.json
```diff
@@ -25,7 +25,7 @@
     "test:types:ts51": "node ../../node_modules/typescript51/lib/tsc.js -p tsconfig.legacy.json",
     "test:types:ts52": "node ../../node_modules/typescript52/lib/tsc.js -p tsconfig.legacy.json",
     "test:types:ts53": "tsc",
-    "test:lib": "vitest --retry=3",
+    "test:lib": "vitest",
```
KATT, line 28:
```suggestion
    "test:lib": "vitest --retry=3",
```

packages/query-core/src/types.ts
```diff
@@ -661,6 +661,10 @@ export interface QueryObserverBaseResult<
    * - See [Network Mode](https://tanstack.com/query/latest/docs/framework/react/guides/network-mode) for more information.
    */
   fetchStatus: FetchStatus
+  /**
+   * A stable promise that will be resolved with the data of the query.
+   */
```
KATT, line 666: This is kinda incorrect for what it is

KATT:
cc @TkDodo the tests that are added here works now

been testing with

```shell
pnpm run test:lib useQuery.test.ts -t ".promise" --watch --coverage.enabled=false
```
inside `packages/react-query`

it's kinda funky but works - if you wanna do a meet and chat about it this week, hmu at @ https://cal.com/alexander (we're overdue a face-to-face anyway)

packages/query-core/src/queryObserver.ts
```diff
@@ -582,6 +592,11 @@ export class QueryObserver<
       isRefetchError: isError && hasData,
       isStale: isStale(query, options),
       refetch: this.refetch,
+      promise: hasData
+        ? fulfilledThenable(data as TData)
+        : status === 'error'
+          ? rejectedThenable(error)
+          : neverResolvedPromise,
```
KATT, line 599: I do feel like this should be [a deferred](https://github.com/trpc/trpc/blob/f1efc8b29c562f4cc88add27aefaeb6a104b67ac/packages/server/src/unstable-core-do-not-import/stream/utils/createDeferred.ts#L2-L8) on the `QueryObserver` instead, but since none of the functions that transform results (`select`, etc) are stable, I don't know if that'd work well and the promise would still be recreated with every result... 

but there's probably ab better way of doing this, possibly one that doesn't need the `Thenable`-type at all...

packages/react-query/src/__tests__/useQuery.test.tsx
```diff
@@ -6581,4 +6582,180 @@ describe('useQuery', () => {
...
+          </React.Suspense>
+        )
+      }
+
+      const rendered = renderWithClient(queryClient, <Page />)
+      await waitFor(() => rendered.getByText('initial'))
+      await waitFor(() => rendered.getByText('test'))
+
+      expect(suspenseRenderCount).toBe(0)
+      // page should only be rendered twice since - promise will get swapped out when result comes in
```
KATT, line 6666:
```suggestion
      // page should only be rendered 2 times - promise will get swapped out when result comes in
```

TkDodo:
> if you wanna do a meet and chat about it this week, hmu at @ https://cal.com/alexander (we're overdue a face-to-face anyway)

done. looking forward to it 🚀

TkDodo:
regarding the other failing tests: vitest seems to now pick up on the rejected promises and reports them as Unhandled Rejection. I'm not yet sure why they would bubble up 🤔 . Good test to start at in the query-core is this one:

https://github.com/TanStack/query/blob/50315acbb39d0c18dbb8343bfd2e13c1ac588b6f/packages/query-core/src/__tests__/query.test.tsx#L838

I can confirm that commenting out this line makes it work:

https://github.com/TanStack/query/pull/7988/files#diff-144cf89d34f29ea1e30da14ec1cd6c323b645d0d446435aa1373b7d0fa00ccbcR605

TkDodo:
one more thing about tracked queries: If we use the promise, we also need to "watch" the `data` property, as that is what it unwraps to.

did that here: d1184ba

packages/query-core/src/queryObserver.ts
```diff
@@ -680,6 +680,9 @@ export class QueryObserver<
       if (this.options.throwOnError) {
         includedProps.add('error')
       }
+      if (includedProps.has('promise')) {
+        includedProps.add('data')
+      }
```
KATT, line 685: > if users use the `promise` returned from useQuery, they are actually interested in the `data` it unwraps to.

if you `use(query.promise)`, you don't necessarily care about `query.data` and shouldn't need to re-render the whole tree

I think this should be deleted

```suggestion
```

I even added a test for this exactly this:

https://github.com/TanStack/query/blob/d1184babd1ba570da09798cad0e8613b7c5b69c8/packages/react-query/src/__tests__/useQuery.test.tsx#L6653-L6654

TkDodo: hmm, I was thinking about this situation:

```
const { promise } = useQuery(...)
const data = use(promise)

return <div>data</div>
```

now, if an automatic refetch occurs (e.g. window focus, or calling `queryClient.refetchQuery` in a sibling component) and that refetch yields new data, the component wouldn't be informed about that _unless_ it also gets a new promise. Is it guaranteed to get a new promise? If so we can remove that code again.

TkDodo: okay nice, then just revert my commit 🙃

KATT: i reverted and added a test case for the above

KATT:
I've added quite extensive tests now that seem to work

CI is giving me

> Agent 'job-66d5f32c5a8b250146e0e1c3-agent-9' exited with the error: Critical Error in Agent: "Critical error while executing Nx tasks.".


Not sure what that is

packages/react-query/src/useBaseQuery.ts
```diff
@@ -67,6 +67,13 @@ export function useBaseQuery<
 
   useClearResetErrorBoundary(errorResetBoundary)
 
+  if (
+    !client.getQueryState(options.queryKey) &&
+    defaultedOptions.enabled !== false
+  ) {
+    client.prefetchQuery(defaultedOptions)
```
KATT, line 74: (I suspect this is what causes the unhandled promise rejection errors, but I am on my phone)

packages/react-query/src/useBaseQuery.ts
```diff
@@ -131,6 +132,29 @@ export function useBaseQuery<
...
+        // noop
+      })
+      .finally(() => {
+        // If there is an existing promise, we tap into it to resolve the currentThenable
+        // This is because `use()` actually unmounts `useQuery()` immediately where the observer is never subscribing
+        if (!observer.hasListeners()) {
+          observer.onQueryUpdate()
+        }
+      })
+  }
```
KATT, line 156: cc @TkDodo 

sooo.... this works but it isn't ideal 😅😅😅😅😅

packages/query-core/src/types.ts
```diff
@@ -661,6 +661,10 @@ export interface QueryObserverBaseResult<
...
+  promise: Promise<TData>
```
KATT, line 667: should this be prefixed with `unstable_` or `experimental_`?

TkDodo: yeah let's do `experimental_promise`

TkDodo: and should we omit this from `useSuspenseQuery` ?

packages/react-query/src/useBaseQuery.ts
```diff
@@ -131,6 +132,21 @@ export function useBaseQuery<
     result,
   )
 
+  if (!isServer && willFetch(result, isRestoring) && !observer.hasListeners()) {
```
TkDodo, line 135: should we hide this whole thing behind an `experimental_autoprefetch` option? We could make it a requirement to turn it on if yo want to work with the `experimental_promise`. 

The thing is I'm not sure if this fetching during render will not cause troubles for other cases that we can't forsee yet.

TkDodo: alternatively, we could ship a new experimental hook that has this turned on ?

KATT:
cc @tkodo I've updated it so it's behind a feature flag now

TkDodo:
I will talk with @Ephem about this next week

TkDodo:
I added some more tests where I wasn't sure if that will work, but it does 👏 .

I will discuss this PR with @Ephem last week but I think we can ship it then. I might want to rename the feature flag to `experimental_autoPrefetching` or something like that, because you can theoretically turn it on without ever using the promise, and it will give you auto prefetching during rendering.

TkDodo:
So I made some more tests for normal `useQuery` with auto-prefetching on and off, on a simple component:

```
it('test', async () => {
  function Page() {
    console.log('render', Date.now())
    const { data } = useQuery({
      queryKey: ['test'],
      queryFn: () => {
        console.log('queryFn', Date.now())
        return 'test'
      },
    })

    return (
      <div>
        <h1>{data}</h1>
      </div>
    )
  }

  renderWithClient(queryClient, <Page />)
})
```

with the flag on, the two logged timestamps consistently differ by **2ms**, because we start the fetch immediately.

with the flag off, we start the fetch in the uSES subscription / an effect, which can be later, depending on how intensive rendering is. In this little example, it was usually **4-5ms**. However, if we add a bit of cpu intensive work to our component, like rendering 1k divs:

```
{Array.from({ length: 1000 }, (_, i) => String(i)).map((v) => (
  <div key={v}>{v}</div>
))}
```

the delay was around **30ms**, but obviously still a constant **2ms** with the flag on.

This probably doesn't matter if your fetch takes 500ms, but I've just had a discussion with someone who said their fetches were delayed by 1-3 seconds, and it turns out it was something in their component rendering.

So the question is: is this actually a good optimization for normal `useQuery` users too? Even if they don't suspend, it should be safe to trigger the fetch during render because there are no active observers (as we trigger the fetch conditionally). It's similar to how our hydration writes to the cache during render for seeding it, or how `usePrefetchQuery` can fetch during render. Thoughts @Ephem ?

packages/react-query/src/useBaseQuery.ts
```diff
@@ -131,6 +136,26 @@ export function useBaseQuery<
...
   )
 
+  if (
+    defaultedOptions.experimental_promise &&
+    !isServer &&
+    willFetch(result, isRestoring) &&
+    !observer.hasListeners()
+  ) {
+    const promise = isNewCacheEntry
+      ? // Fetch immediately on mount in order to ensure `.promise` is resolved even if the component is unmounted
```
Ephem, line 146: NIT: This comment has me a bit confused, is it "in render" instead of "on mount"?

packages/react-query/src/useBaseQuery.ts
```diff
@@ -131,6 +136,26 @@ export function useBaseQuery<
...
+        fetchOptimistic(defaultedOptions, observer, errorResetBoundary)
+      : // subscribe to the "cache promise" so that we can finalize the currentThenable once data comes in
+        client.getQueryCache().get(defaultedOptions.queryHash)?.promise
+
+    promise?.catch(noop).finally(() => {
+      if (!observer.hasListeners()) {
```
Ephem, line 152: I'm trying to wrap my head around this all. Why are we checking for `!observer.hasListeners()` here?

packages/query-core/src/retryer.ts
Ephem, line 1: Just making sure I understand, is this file now using `pendingThenable` for convenience, or is returning `thenable` from here serving some specific purpose related to the core functionality in this PR I'm missing?

(Nice change either way!)

packages/query-core/src/queryObserver.ts
```diff
@@ -82,6 +85,13 @@ export class QueryObserver<
 
     this.#client = client
     this.#selectError = null
+    this.#currentThenable = pendingThenable()
```
Ephem, line 88: I just want to double check that I'm thinking about this correctly. This is a "proxy"-thenable that is kept in sync with the promise that lives in the query cache, and the mechanism for that "syncing" is the `updateResult`-function?

So any means of resolving, rejecting or replacing the original promise needs to call `updateResult` on all observers to "sync" to the thenables?

Just thinking out loud, this makes me wonder what happens if you:

* useQuery -> use(promise)
* fetchQuery / refetch before above query resolves, and either:
  * The first query resolves first
  * The fetchQuery / refetch resolves first

Does this have a defined behaviour? I'm not even sure what _should_ happen. 😅

Ephem (commented):
Fantastic! 👏👏 This is pretty complex, so I'll need to some more time to really wrap my head around things.

So many great tests here! Do you think it would make sense to also add some tests with `<HydrationBoundary>`, both when we de/rehydrate already awaited data, and when we de/rehydrate a promise?

Ephem:
> Thoughts @Ephem ?

I think I need to marinate this a bit. I mean, on one hand it should be "safe" to do. On the other, is it always desirable? I'm thinking about (future) cases like pre-rendering, the `<Offscreen>` API etc. Probably? But maybe not if a router prerenders all possible routes based on Links on a page? Will devs need some more control (prefetch, but not if this is a prerender).

If there are any issues there, I fully expect `usePrefetchQuery` to have the same ones, but at least that's more explicit. I think that's mostly an argument for not aiming to make this the "default" behaviour though and a big part of me wants to say this is a nice little optimisation and we can tackle any challenges when they come up. 😄

TkDodo:
> If there are any issues there, I fully expect usePrefetchQuery to have the same ones

yes, this auto_prefetching was pretty much taken from the `usePrefetchQuery` implementation. I think we are aligned that we could keep it as an experimental flag though, and that you need to turn it on to be able to `use(promise)`, which makes the whole promise feature experimental, too :)

packages/query-core/src/retryer.ts
TkDodo: convenience really. Thenable is just the abstraction for creating a lazy promise here. In the future, we can hopefully replace it with [Promise.withResolvers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers)

packages/query-core/src/queryObserver.ts
TkDodo: we can add tests for this, but the way I see it, `fetchQuery` would just pick up the promise that's already running and not start a new fetch. Of course, if you do `queryClient.refetchQueries` and target this key, it would _cancel_ the ongoing fetch per default. Not sure what would happen then, cancel should silently rollback to the previous (pending) state, so maybe it would hang forever.

Definitely interesting case that's worth a couple of tests 👍

packages/query-core/src/retryer.ts
Ephem: I thought so but wanted to be sure I wasn't missing anything, thanks!

packages/query-core/src/queryObserver.ts
```diff
@@ -82,6 +85,13 @@ export class QueryObserver<
...
+    if (!this.options.experimental_promise) {
```
TkDodo, line 89: let's rename to: `experimental_prefetchInRender`

KATT:
Ping me know if you need anything on my end here or have questions that I can help answer, I don't see much tangible feedback as of yet ☺

packages/react-query/src/useBaseQuery.ts
KATT:
```suggestion
      ? // Fetch immediately on render in order to ensure `.promise` is resolved even if the component is unmounted
```

packages/react-query/src/useBaseQuery.ts
KATT: Just to prevent `updateResult()` to be called twice - if it has listeners, `updateResult` will be called

TkDodo:
This user-land implementation got me thinking: why don't we also do the promise handling in the react layer rather than the observer 🤔. It looks way simplified. What am I missing 😂?

https://github.com/nucleartux/react-query-read

TkDodo:
@KATT I added more tests, and one scenario is failing ([this one](https://github.com/TanStack/query/pull/7988/files#diff-3cd8cae55f836a38ea976d6cbf9f1b59e8fdfa4d3218b123cb1f9dccfc6824d2R7216)): if we cancel the query while we are suspending, we stay in a forever pending state.

I _think_ the problem is that we just revert to the previous state if we get a query cancelation, and state happens to be `status: 'pending'` because we didn't have any previous data:

https://github.com/TanStack/query/blob/55b6c19dc0c2e5c56a3a1b037160572bedc3247c/packages/query-core/src/query.ts#L587-L589

I'm not even sure rejecting the promise the right expectation here, as the query isn't in error state. Honestly not sure what should happen here - maybe staying in `pending` is the right call, idk ...

KATT:
cc @TkDodo I added what I think is correct code for cancellation

https://github.com/TanStack/query/blob/d3bd66c44aca673afc5cbd1ee3c837d2f7d4ea59/packages/query-core/src/queryObserver.ts#L611-L621

I admit it isn't pretty

KATT:
Now there's a bad test due to the fact that we recreate the thenable when it "cancels" twice which is the case of below since the query is disabled:

https://github.com/KATT/react-query/blob/d3bd66c44aca673afc5cbd1ee3c837d2f7d4ea59/packages/query-core/src/__tests__/queryObserver.test.tsx#L1085 

First run it'll create a thenable that is rejected, which is recreated with new results and immediately rejected :|

packages/query-core/src/queryObserver.ts
```diff
@@ -593,6 +608,42 @@ export class QueryObserver<
...
+            prevThenable.resolve(nextThenable.value)
+          } else if (nextThenable.status === 'rejected') {
+            prevThenable.reject(nextThenable.reason)
+          }
+          break
+
+        case 'fulfilled':
+        case 'rejected':
+          if (!isThenableEqual(prevThenable, nextThenable)) {
+            // Replace the thenable and result promise when the results have changed
```
KATT, line 640:
```suggestion
            // Replace the thenable when the results have changed
```

packages/query-core/src/queryObserver.ts
```diff
@@ -593,6 +608,42 @@ export class QueryObserver<
...
+
+        case 'fulfilled':
+        case 'rejected':
+          if (!isThenableEqual(prevThenable, nextThenable)) {
+            // Replace the thenable when the results have changed
+            this.#currentThenable = nextResult.promise = nextThenable
+          }
+          break
+      }
+    }
```
KATT, line 645: cc @TkDodo 

This code has updated a bit to support for cancellation - since the query client itself doesn't have the concept of "abort errors" (AFAICT?) we instead:

1. Eagerly create a new thenable of what the "next" thenable should represent
  1. If the previous thenable is pending, we finalize it if the next reason or value
  2. If the previous is not pending, we do an equal-check to see if we should replace it (since the cancellation error is a constant, the equal-check will work)

packages/react-query/src/useBaseQuery.ts
TkDodo: yeah I just checked, if we remove both checks for `!observer.hasListeners()`, all the tests still pass. if we want to keep these checks, we should add tests where this fails @KATT. I know that calling `updateResult` multiple times might e.g. re-run inline `select` functions, but I'm not sure if that's worth optimizing for ...

packages/query-core/src/queryObserver.ts
```diff
@@ -593,6 +608,42 @@ export class QueryObserver<
       | undefined
 
     const nextResult = this.createResult(this.#currentQuery, this.options)
+
+    if (this.options.experimental_prefetchInRender) {
```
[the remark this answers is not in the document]
KATT: Sounds good to me, I don't think the positioning was deliberate; probably just put it first to avoid premature debugging

packages/query-core/src/queryObserver.ts
TkDodo: @KATT I think I changed my mind on this one. Rejecting the promise doesn't seem right - the user shouldn't see such a cancelation error, as it's not actionable. I think we can just let the promise stay pending, and then it needs to resolve when data comes in.

I did that here: [849b1b6](https://github.com/TanStack/query/pull/7988/commits/849b1b6d22c679fe9f522ee523a5bd76ca00c0f1)

let me know if that's okay, implementation wise?

TkDodo: it's pretty much the same behaviour if a query is `enabled: false` and you call `use()` on the promise. Basically every situation where you are `status: 'pending'` and `fetchStatus` isn't `'fetching'`. Added a test for that here: [c75afaa](https://github.com/TanStack/query/pull/7988/commits/c75afaaae9a266447a7a805326fdae01a7b0cde2)

KATT: Sounds fine to me, does that align with the behavior of

```ts
const resPromise = client.fetchQuery(/* ... */)
client.cancelQueries(/* ... */)

const res = await resPromise; // <-- never resolves currently, i guess?
```

it does seem to be that this should result in an [`AbortError`](https://developer.mozilla.org/en-US/docs/Web/API/DOMException#aborterror) but if it aligns with the current behavior of a fetch/cancel dance, it sgtm

packages/react-query/src/useBaseQuery.ts
KATT: I tried adding a test for it by

```ts
vi.spyOn(QueryObserver.prototype, 'updateResult')
```

but it's called a bunch of times within a render so adding a test the way I thought will be quite hard

feel free to delete the optimization

TkDodo: I've deleted the first check, because it's technically impossible that `willFetch` is true _and_ we have an observer. I've kept the second check because maybe an observer shows up in between: [a9148ba](https://github.com/TanStack/query/pull/7988/commits/a9148baaa678fcb5186b985c87e797d7a43f04dd)

KATT:
cc @TkDodo anything I can do to get this over the line?

TkDodo:
@KATT updating docs would be great ❤️

KATT:
I added some docs

docs/framework/react/reference/queryOptions.md
```diff
@@ -17,3 +17,8 @@ You can generally pass everything to `queryOptions` that you can also pass to [`
 - `queryKey: QueryKey`
   - **Required**
   - The query key to generate options for.
+- `experimental_prefetchInRender?: boolean`
+  - Optional
+  - Defaults to `false`
+  - When set to `true`, enables the experimental `useQuery().promise` functionality
+  - This allows prefetching queries during render, which can be useful for certain optimization scenarios
```
TkDodo, line 24:
```suggestion
  - When set to `true`, queries will be prefetched during render, which can be useful for certain optimization scenarios
  - Needs to be turned on for the experimental `useQuery().promise` functionality
```

TkDodo (approved):
@Ephem are you good with shipping this?

timvandam:
Hi @KATT, I came across the experimental_prefetchInRender property and think it would be a nice addition to allow prefetching during server-side rendering. Currently the prefetch branch is guarded with `!isServer`: https://github.com/TanStack/query/blob/d07a3db968f9d836a1b658c043bc723ac424b304/packages/react-query/src/useBaseQuery.ts#L144-L159

Changing the type from `boolean` to e.g. `boolean | 'server' | 'client' | 'all'` would make it very easy to set up server-side rendering in Vite based setups that don't use suspense by simply waiting for fetches to resolve (eg like this https://github.com/timvandam/react-vite-trpc-ssr/blob/main/src/entry.server.tsx#L50-L68). What do you think?

PR #8007, opened 2024-09-02
feat(angular-query): injectQueries

Arthie:
Noticed injectQueries had yet to be fully implemented, this is my attempt to get it working.
Did some limited testing, and added simple test cases that are all passing.

Looking for feedback.


arnoud-dv:
That's great, finishing `injectQueries` is one of the major pieces of work remaining. I'll review it but may be a few days.

Timebutt:
I just pulled this branch into my project since I needed the `injectQueries` functionality, and found it non-functional in the current implementation in `main`. Is this something we can get reviewed so that it could make its way into the main branch? Thanks!

arnoud-dv:
Thanks for your PR @Arthie.

API wise this PR differs from the other adapters as these have a `queries` property in their `useQueries` equivalent options. I'd like to stick to the other adapters as close as possible so users can transfer their knowledge between framework implementations and it makes maintenance easier.

The other functions in the Angular adapter reactive to options such as `injectQuery` have a single callback function returning the options.

In the result returned from these functions each property is a distinct signal, which helps with fine grained reactivity.

In [this PR](https://github.com/TanStack/query/pull/8690) I think we're almost there in implementing an API consistent with these principles. Feel free to contribute either in code or testing, we maintain a list of what needs to be done in the PR.

PR #8029, opened 2024-09-06
fix(react-query): QueryErrorResetBoundaryValue as class

manudeli:


TkDodo:
not sure what the advantage of this PR would be, but the disadvantage is that you can't destruct classes, yet this is what the docs suggest and probably everyone is doing:

```
const { reset } = useQueryErrorResetBoundary()
```

⬆️ this will fail to work because of `this` bindings.

PR #8030, opened 2024-09-08
feat: throw when `replaceDeepEqual` contains circular references

jxom:
This PR is a follow up on #7966 to:
- not surface console errors for values that are referentially stable (e.g. `BigInt`)
- detect circular references in `replaceDeepEqual`
- catch errors from `replaceDeepEqual` in `replaceData`, and surface them in the console + propagated to the query.

This is not a breaking change as `replaceDeepEqual` would already throw a "stack too deep" error for circular references anyway – this change eagerly catches it.

### DIFF
--- a/packages/query-core/src/utils.ts
+++ b/packages/query-core/src/utils.ts
@@ -242,42 +242,54 @@ export function partialMatchKey(a: any, b: any): boolean {
  */
 export function replaceEqualDeep<T>(a: unknown, b: T): T
 export function replaceEqualDeep(a: any, b: any): any {
-  if (a === b) {
-    return a
-  }
+  const seen = new WeakSet()
 
-  const array = isPlainArray(a) && isPlainArray(b)
+  function _replaceEqualDeep(a: any, b: any) {
+    if (a === b) {
+      return a
+    }
+  
+    const array = isPlainArray(a) && isPlainArray(b)
+  
+    if (array || (isPlainObject(a) && isPlainObject(b))) {
+      if (seen.has(a)) {
+        throw new Error('circular reference detected.')
+      }
+      seen.add(a)
 
-  if (array || (isPlainObject(a) && isPlainObject(b))) {
-    const aItems = array ? a : Object.keys(a)
-    const aSize = aItems.length
-    const bItems = array ? b : Object.keys(b)
-    const bSize = bItems.length
-    const copy: any = array ? [] : {}
+      const aItems = array ? a : Object.keys(a)
+      const aSize = aItems.length
+      const bItems = array ? b : Object.keys(b)
+      const bSize = bItems.length
+      const copy: any = array ? [] : {}
+  
+      let equalItems = 0
+  
+      for (let i = 0; i < bSize; i++) {
+        const key = array ? i : bItems[i]
 
-    let equalItems = 0
-
-    for (let i = 0; i < bSize; i++) {
-      const key = array ? i : bItems[i]
-      if (
-        ((!array && aItems.includes(key)) || array) &&
-        a[key] === undefined &&
-        b[key] === undefined
-      ) {
-        copy[key] = undefined
-        equalItems++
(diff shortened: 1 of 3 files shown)


packages/query-core/src/utils.ts
```diff
@@ -242,42 +242,54 @@ export function partialMatchKey(a: any, b: any): boolean {
...
+      return a
+    }
+
+    const array = isPlainArray(a) && isPlainArray(b)
+
+    if (array || (isPlainObject(a) && isPlainObject(b))) {
+      if (seen.has(a)) {
+        throw new Error('circular reference detected.')
+      }
+      seen.add(a)
```
jxom, line 258: these lines were added

packages/query-core/src/utils.ts
```diff
@@ -242,42 +242,54 @@ export function partialMatchKey(a: any, b: any): boolean {
  */
 export function replaceEqualDeep<T>(a: unknown, b: T): T
 export function replaceEqualDeep(a: any, b: any): any {
-  if (a === b) {
-    return a
-  }
+  const seen = new WeakSet()
```
jxom, line 245: this line was added

TkDodo (changes requested):
I think we had a bit of a miscommunication. Ideally, we don't want any runtime changes for production builds - that's why the initial code was behind an env check.

So my idea was something like this:

```
  try {
    return replaceEqualDeep(prevData, data)
  }
  catch(error) {
    console.error(...)
  }
}
return replaceEqualDeep(prevData, data)
```

This will remove the dev mode overhead of json stringify. If everything goes right, `replaceEqualDeep` will only be called once. If there is a circularity, the first call will log the error and the second call will then throw, so it will be caught in `setData` and the query will go to error state.

I think this would be the best of both worlds.

packages/query-core/src/utils.ts
```diff
@@ -354,19 +366,17 @@ export function replaceData<
...
-      } catch (error) {
-        console.error(
-          `StructuralSharing requires data to be JSON serializable. To fix this, turn off structuralSharing or return JSON-serializable data from your queryFn. [${options.queryHash}]: ${error}`,
-        )
-      }
+    try {
+      // Structurally share data between prev and new data if needed
+      return replaceEqualDeep(prevData, data)
+    } catch (error) {
+      console.error(
```
jxom, line 373: Realised this should be under a dev env guard. Will fix when I’m back at computer.

jxom:
> I think we had a bit of a miscommunication. Ideally, we don't want any runtime changes for production builds - that's why the initial code was behind an env check.
> 
> So my idea was something like this:
> 
> ```
> if (process.env.NODE_ENV !== 'production') {
>   try {
>     return replaceEqualDeep(prevData, data)
>   }
>   catch(error) {
>     console.error(...)
>   }
> }
> return replaceEqualDeep(prevData, data)
> ```
> 
> This will remove the dev mode overhead of json stringify. If everything goes right, `replaceEqualDeep` will only be called once. If there is a circularity, the first call will log the error and the second call will then throw, so it will be caught in `setData` and the query will go to error state.
> 
> I think this would be the best of both worlds.

Got it. Can do that!

jxom:
@TkDodo - lmk what you think

packages/query-core/src/utils.ts
```diff
@@ -356,15 +356,13 @@ export function replaceData<
   } else if (options.structuralSharing !== false) {
     if (process.env.NODE_ENV !== 'production') {
       try {
-        JSON.stringify(prevData)
-        JSON.stringify(data)
+        replaceEqualDeep(prevData, data)
```
akshatmittal, line 359: I believe you can return here to avoid the double `replaceEqualDeep` calls.

jxom: good catch!

TkDodo (approved):

PR #8062, opened 2024-09-16
fix(experimental_createQueryPersister): return more utilities, rename persister

DamianOsipiuk:
Restructure plugin to expose more utilities useful in user-land.

Returned utilities:
- persisterFn - main persister function that can be passed to `useQuery`
- persistQueryByKey - persist query to persister storage. Ex call it just after `setQueryData`
- retrieveQuery - retrieve one query from persister storage
- persisterGc - gc pass for the persister storage to clean up unusable entries
- persisterRestoreAll - restore are persisted queries from storage

TODO:
- [x] tests

### DIFF
--- a/docs/framework/react/plugins/createPersister.md
+++ b/docs/framework/react/plugins/createPersister.md
@@ -1,6 +1,6 @@
 ---
 id: createPersister
-title: experimental_createPersister
+title: experimental_createQueryPersister
 ---
 
 ## Installation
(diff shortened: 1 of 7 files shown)


frederikhors:
@DamianOsipiuk thank you very much for this!

TkDodo:
does this close any of the open issues around `createPersister` ?

DamianOsipiuk:
> does this close any of the open issues around `createPersister` ?

This should close https://github.com/TanStack/query/issues/6310.  
It exposes utilities to do persistence on demand in user-land.  
And if i recall correctly we decided that we do not want to put more code to make that work out-of-the-box into the core.

TkDodo:
yeah if we want more utilities, they should come from the persister package

PR #8072, opened 2024-09-19
feat(eslint-plugin-query): add rule to ensure property order of infinite query functions

schiller-manuel:


docs/eslint/infinite-query-property-order.md
```diff
@@ -0,0 +1,62 @@
+---
+id: infinite-query-property-order
+title: Ensure correct order of inference sensitive properties for `useInfiniteQuery` and `infiniteQueryOptions`
+---
+
+For the following functions, the property order of the passed in object matters due to type inference:
+
+- `useInfiniteQuery`
+- `infiniteQueryOptions`
```
TkDodo, line 9: we also need to add `useSuspenseInfiniteQuery` please :)

packages/eslint-plugin-query/src/index.ts
```diff
@@ -29,6 +29,7 @@ Object.assign(plugin.configs, {
       '@tanstack/query/no-rest-destructuring': 'warn',
       '@tanstack/query/stable-query-client': 'error',
       '@tanstack/query/no-unstable-deps': 'error',
+      '@tanstack/query/finite-query-property-order': 'error',
```
TkDodo, line 32:
```suggestion
      '@tanstack/query/infinite-query-property-order': 'error',
```

packages/eslint-plugin-query/src/index.ts
```diff
@@ -41,6 +42,7 @@ Object.assign(plugin.configs, {
         '@tanstack/query/no-rest-destructuring': 'warn',
         '@tanstack/query/stable-query-client': 'error',
         '@tanstack/query/no-unstable-deps': 'error',
+        '@tanstack/query/finite-query-property-order': 'error',
```
TkDodo, line 45:
```suggestion
        '@tanstack/query/infinite-query-property-order': 'error',
```

packages/eslint-plugin-query/src/rules/infinite-query-property-order/infinite-query-property-order.rule.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export const name = 'infinite-query-property-order'
+
+export const rule = createRule({
+  name,
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'Ensure correct order of inference sensitive properties for `useInfiniteQuery` and `infiniteQueryOptions`',
```
TkDodo, line 25:
```suggestion
        'Ensure correct order of inference sensitive properties for infinite queries',
```

TkDodo (changes requested):

docs/eslint/infinite-query-property-order.md
schiller-manuel: done

packages/eslint-plugin-query/src/index.ts
schiller-manuel: done

packages/eslint-plugin-query/src/index.ts
schiller-manuel: done

packages/eslint-plugin-query/src/rules/infinite-query-property-order/infinite-query-property-order.rule.ts
schiller-manuel: done

packages/eslint-plugin-query/src/rules/infinite-query-property-order/constants.ts
```diff
@@ -0,0 +1,13 @@
...
+  'useSuspenseInfiniteQuery',
+] as const
+
+export type InfiniteQueryFunctions = (typeof infiniteQueryFunctions)[number]
+
+export const checkedProperties = [
+  'queryFn',
+  'getPreviousPageParam',
+  'getNextPageParam',
+] as const
```
TkDodo, line 13: is this the full order that we must have in the object? Because it shouldn't matter if `getNextPageParam` comes before or after `getPreviousPageParam` - it only matters that they are after `queryFn`. So maybe it needs like a dependency tree:

```
const propertyDependencies = {
  queryFn: ['getPreviousPageParam', 'getNextPageParam']
}
```

this would define that `'getPreviousPageParam'` and `'getNextPageParam'` depend on `queryFn` and thus must be after it. What do you think?

schiller-manuel: yes ideally a tree like structure would be best. I decided against this in router (similar issue with validateSearch and params.parse) to keep this plugin simple.

TkDodo: ok fine by me for now

TkDodo (approved):

timvandam:
Can’t NoInfer<> make type inference work regardless of the order of properties, or what exactly can cause type inference issues here?

schiller-manuel:
this is explained here

https://www.totaltypescript.com/property-order-matters

timvandam:
I see, thanks for sharing

PR #8088, opened 2024-09-23
document: add guide for using props as a key of query

S-sohofi:


suneettipirneni:
As of https://github.com/TanStack/query/pull/7608. You no longer need an intermediate `computed` statement for props. You can just use the getter syntax `() => props.foo` directly. It would probably be worth documenting this as using a computed ref for a simple property access is less efficient than using a getter.

S-sohofi:
> As of #7608. You no longer need an intermediate `computed` statement for props. You can just use the getter syntax `() => props.foo` directly. It would probably be worth documenting this as using a computed ref for a simple property access is less efficient than using a getter.

you mean I should do like this?

```
<script setup>
  const props = defineProps([todoId])
  const queryKey = ['todos', ()=>props.todoId]
  const result = useQuery(queryKey, () => fetchTodoById(todoIdValue.value))
</script>
```
but it didn't work for me.
are these changes merged with the main branch?

TkDodo (changes requested):
not sure what happened here but prettier seems to not handle those setups scripts very well ?

S-sohofi:
@TkDodo i've fixed the problem with prettier.

TkDodo:
> I’ve fixed the problem with prettier.

you’ve worked around prettier. Now prettier runs again and it automatically re-formats again.

see: https://github.com/TanStack/query/pull/8088/commits/9b9c095147eeae6dc2d4f7da1bfcc3b17bc32cf4

S-sohofi:
I've done it in another way, recheck it @TkDodo .

TkDodo:
please use prettier locally; it re-formatted again; gonna close this as it kinda wastes my time, sorry.

PR #8092, opened 2024-09-23
[perf] improve perf, as discussed in #8091

joeljeske:


Aghassi:
> Thanks - I think proxies should work well. Just for context, when we started out, we had to support browsers that didn’t have proxy support and we never re-visited it.
> 
> I think there are two other places that use `Object.defineProperty`. Can you take a stab at them as well please?
> 
> also, docs would need an update here please:
> 
> https://tanstack.com/query/v5/docs/framework/react/guides/render-optimizations#tracked-properties

@TkDodo Thanks for this! @joeljeske and I work together and from our profiling this is the only major hot spot. We didn't see improvement in updating the other call sites. Does your test suite cover any performance tests to assert no regressions if other sites are updated? I think it may be useful to incrementally land this as this one is a known issue that we discovered. 

Also happy to help update the docs, what would you like updated? I can help take this over from @joeljeske as we'd love to give this fix back to the community and not maintain it internally

TkDodo:
> We didn't see improvement in updating the other call sites

Understood, but for having a unified approach, I think we should use either proxies or `Object.defineProperty` everywhere.

> Also happy to help update the docs, what would you like updated?

This page (https://tanstack.com/query/v5/docs/framework/react/guides/render-optimizations#tracked-properties) shouldn’t refer to `custom getters` anymore, but state that Proxy is used instead.

TkDodo:
@joeljeske @Aghassi do you plan to continue working on this contribution ?

Aghassi:
@TkDodo yes sorry, been very busy with work and the holidays. I will be taking this up, sorry for the delay

TkDodo:
2 more months have passed, @Aghassi can we please wrap this up, otherwise I’ll close the PR. Thanks 🙏

TkDodo:
superseded by:
- https://github.com/TanStack/query/pull/9079

PR #8096, opened 2024-09-25
**perf** better timeout handling

joeljeske:
This minor tweak reduces the number of setInterval/clearInterval/clearTimeout calls that are made, which improves performance by not needing to interact with native timers, which are slower than JS calls.


packages/query-core/src/queryObserver.ts
```diff
@@ -399,9 +400,21 @@ export class QueryObserver<
...
+    // we already have a properly set interval running
+    if (
+      prevInterval == this.#currentRefetchInterval &&
+      this.#refetchIntervalId
+    ) {
+      return
+    }
+
+    this.#clearRefetchInterval()
+
```
TkDodo, line 417: I don’t think this is correct. The previous implementation re-sets an interval every time new data comes in, which this implementation does not. Suppose the following scenario:

- we have a refetchInterval set to every 5 minutes (timestamp = 0)
- after 3 minutes, we get another update into our cache (e.g. with setQueryData or with an invalidation after a mutation) (timestamp = 3 min)

--> that is supposed to reset the interval, so that we don’t get another refetch after 2 minutes (timestamp = 5min), but 5 minutes _after the last update_ (so timestamp = 8min)

I hope we have a test case for this that now fails, otherwise, we should probably add one :)

TkDodo:
I think the failing test around intervals shows that this approach is not correct.

PR #8100, opened 2024-09-25
refactor(react-query): apply early return/continue

kickbelldev:
I've made improvements to the react-query package by applying the early return/continue pattern to reduce nested if statements and if-else constructs.


TkDodo:
it’s subjective, but I don’t think negated conditions and continue statements make the code any more readable.

PR #8107, opened 2024-09-28
fix: improve long running task performance in query core

Aghassi:
This addresses two hot spots we have noticed
on a large scale enterprise app when profiling
with chrome. These changes help to cut down
on long running tasks when there are many query
calls on the page


Aghassi:
This uses parts of https://github.com/TanStack/query/pull/8092 as well

Aghassi:
@tannerlinsley Would you be open to taking this fix? Anything I can provide additionally to help with making it easier to take in? Would love to give this one to the community and not maintain an internal patch if I can.

TkDodo:
> This uses parts of https://github.com/TanStack/query/pull/8092 as well

yeah let’s not mix PRs / improvements please. I have reviewed the Proxy one already and left a comment. I’m also fine with this change in isolation.

Aghassi:
> > This uses parts of #8092 as well
> 
> yeah let’s not mix PRs / improvements please. I have reviewed the Proxy one already and left a comment. I’m also fine with this change in isolation.

Sounds good, I will drop the other diff from this and focus just on the one file

Aghassi:
@TkDodo Updated to just address the one performance issue

TkDodo (approved):

PR #8117, opened 2024-10-01
fix(eslint-plugin-query): update deprecated `getSourceCode()` method

saul-atomrigs:
Captura de pantalla 2024-10-02 a las 12 03 04 a  m
https://eslint.org/docs/latest/extend/custom-rules#accessing-the-source-code

`getSourceCode()` has been deprecated in favor of `sourceCode`. This PR introduces the latter, while also keeping the deprecated one as fallback.


TkDodo:
you need to disable the linter for that fallback:

```
/home/workflows/workspace/packages/eslint-plugin-query/src/rules/stable-query-client/stable-query-client.rule.ts
  72:32  error  Unnecessary conditional, expected left-hand side of `??` operator to be possibly null or undefined  @typescript-eslint/no-unnecessary-condition
```

saul-atomrigs:
Thanks for the reminder, I've just added the eslint-disable

Newbie012 (approved):
LGTM

PR #8134, opened 2024-10-05
fix(useInfiniteQuery): prevent error when pages is undefined

sinasun:
This address a very rare bug that happened few times to my users with useInfiniteQuery, where the pages was undefined in getNextPageParam function
should not break anything
image

### DIFF
--- a/packages/query-core/src/infiniteQueryBehavior.ts
+++ b/packages/query-core/src/infiniteQueryBehavior.ts
@@ -132,14 +132,13 @@ function getNextPageParam(
   options: InfiniteQueryPageParamsOptions<any>,
   { pages, pageParams }: InfiniteData<unknown>,
 ): unknown | undefined {
-  if (!pages || pages.length === 0 || !pageParams || pageParams.length === 0) {
-    return undefined;
+  if (!pages?.length || !pageParams?.length) {
+    return undefined
   }
-
   const lastIndex = pages.length - 1
 
   if (lastIndex >= pageParams.length) {
-    return undefined;
+    return undefined
   }
 
   return pages.length > 0
--- a/packages/query-core/src/infiniteQueryBehavior.ts
+++ b/packages/query-core/src/infiniteQueryBehavior.ts
@@ -132,7 +132,16 @@ function getNextPageParam(
   options: InfiniteQueryPageParamsOptions<any>,
   { pages, pageParams }: InfiniteData<unknown>,
 ): unknown | undefined {
+  if (!pages || pages.length === 0 || !pageParams || pageParams.length === 0) {
+    return undefined;
+  }
+
   const lastIndex = pages.length - 1
+
+  if (lastIndex >= pageParams.length) {
+    return undefined;
+  }
+
   return pages.length > 0
     ? options.getNextPageParam(
         pages[lastIndex],


sinasun:
I have found when it happens, will push some code on the weekend. just some examples of when it can happen:

1. when it's trying to fetch the query and it gets cancelled (happened to me when I refereshed the page during the fetching), it put data: {} with no pages/pageParams attribute for InfinteQuery data object
2. using setQueries, you can set the InfinteQuery data object to {}, which doesn't force you to add pages/pageParams attributes

will put tests + fix typing for InfinteQuery (right now it assumes pages/pageParams always exist, probably we can assume pages can be undefined as well)

TkDodo:
> using setQueries, you can set the InfinteQuery data object to {}, which doesn't force you to add pages/pageParams attributes

this is a programming error; if you use `infiniteQueryOptions` to make `setQueryData` type-safe, it _will_ error with `property pages is missing` ([playground](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbwLACg5wIoFcCmUCeAwgDbA4B2MANKusOQGb3Aw7Z74DyYMwE5AZxooAvnAZQIIOAHIAAjACGgpQGMA1gHooORapgBaAI64CM1BZSr+A+BB59BcALxx6TcizamuDmwApkNDgTDgBpHHwALjgAbRlWWxkAXWF0UIIAMXIY-wBKFwA+OAAFSRBgARwAOh0BCGIANxx-AFY8tLdPXkViEsUAcxx+qEUQGIBGTqGYADkcAA8YfqGRsdyC52KscgATHA8cXdQRPMtrFRCfEjJKFzhyHAB3TGvSChh8ywyid8pqqowdgEAAiiiU-nsvBs1R+EXwVEQcDAg2GilGIAEMViyTgpyAA))

> when it's trying to fetch the query and it gets cancelled (happened to me when I refereshed the page during the fetching), it put data: {} with no pages/pageParams attribute for InfinteQuery data object

query cancellation will re-set to the previous state, and that might be a `pending` state but it should never be an invalid state. So I also don’t see it happening with cancellation, but I’m gladly proven wrong here :)

ugurcanatas:
hey there, i recently opened a similar [Q&A discussion](https://github.com/TanStack/query/discussions/8214) where we see the same exact error logs on infinite queries. Trying it out to reproduce but I am still not sure how.

But i can tell that we are not using infinite query data by using setQueryData

ugurcanatas:
By following the logs, I created a list where the trace logs points to in the minified bundle. Maybe its helpful to create some sort of steps to reproduce. @sinasun @TkDodo 

- useSomeQuery.useInfiniteQuery
- useBaseQuery
```js
// specifically starts from here
function p(t,n){return(0,r(d[0]).useBaseQuery)(t,r(d[1]).InfiniteQueryObserver,n)}
``` 
- Then getOptimisticResults in which is most likely useBaseQuery.ts
```js
C=_.getOptimisticResult(b);if(y.useSyncExternalStore(y.useCallback((function(t)...
```

- Then getOptimisticResult
```js
function(e){return e.behavior=(0,r(d[6]).infiniteQueryBehavior)(),r(d[5])(r(d[0])(s.prototype),"getOptimisticResult",this).call(this,e)}}
```

- Then createResult in query observer
```js
o=this.createResult(e,t)
```

- Then hasNextPage is called
```js
hasNextPage:(0,r(d[6]).hasNextPage)(t,c.data)
```

That calls getNextPageParam is called
```js
function p(n,t){var o=t.pages,u=t.pageParams,l=o.length-1;return o.length>0?n.getNextPageParam(o[l],o,u[l],u):void 0}
```

here is why i am lost because I am not sure how pages can be undefined or null based on checking the code in `InfinitieQueryBehaviour` file

PR #8151, opened 2024-10-08
fix: new cache entry check

jxom:
Fixes https://github.com/wevm/wagmi/issues/4322

Seems that https://github.com/TanStack/query/pull/7988 introduced a subtle breaking change whereby the check for `isNewCacheEntry` would not account for a custom `queryKeyHashFn`. This is because the implementation of `getQueryState` does not pass custom options from the query to `defaultQueryOptions`. As a result, this broke consumer code that relyed on `queryKeyHashFn` to serialize bigints. This PR aims to access the cache directly with `getQueryCache` instead.


TkDodo (approved):
Nice find, thanks 🙏

Issue #8132, opened 2024-10-04
Cannot pass an optional infiniteData parameter to infiniteQueryOptions function


jimmycallin:
### Describe the bug

It doesn't seem possible to pass an optional initialData parameter to infiniteQueryOptions, while it does work for useInfiniteQuery. 

Example query not working:

[See playground example here](https://www.typescriptlang.org/play/?ssl=18&ssc=1&pln=19&pc=1#code/JYWwDg9gTgLgBAbzgVwM4FMCSA7AZsbYGdARWXSgE84BfOXKCEOAIgAEYBDbVLgYwDWAeijpOfGAFoAjuSosA3ACgluZNgnAI2FBgDqUTmDAUAFEgJFgnADYARTl1oAuRHEsxr9x5wD8rpHQAD05wG3RXGChyWloASkQlODg+bV44WQpKAHkwTzS4AF53PA9SORy8rR5zJOSMioBpdEpXAG0WYNCwcJYAXQAaOuTMqgAxbFdOVEoNOFMEwoA+EqtbBy4h+tXPdZ9XDy8NzmH633nFlfM4ME4Ac3RUdsO9rkGb+-QABU5DECe4G0+vFTslXOoACbofDYdAQrb1B4wABywRgPwePz+rgWRRWAEYEckXjYMd9fqFXIS6jQ4soaCo1Bp8jo0OgDEYTFAyFlriTji43PyfP43F0whE4FEYnRaYlkqIYMgoKyMDgYURyryEKdRpRmq1AZ0QhL+kSGlkJlMZnNccsdkcfObhVwDoRdt4uKC4Oc7VckLcHgC2i7OO9A+S-sHgbTveDsFCYXDzUjUUF0Z8sZSLni4NTtiSyVmQFSEbT6UA).

### Your minimal, reproducible example

[...]

PR #8154, opened 2024-10-09
fix(react-query): Allow optional initialData in infiniteQueryOptions

sungpaks:
Issue fix : #8132  
There was a typescript error(but could be run with no problem) when using optional `initialData` in `infiniteQueryOptions`


packages/react-query/src/infiniteQueryOptions.ts
```diff
@@ -42,6 +42,7 @@ export type DefinedInitialDataInfiniteOptions<
   initialData:
     | NonUndefinedGuard<InfiniteData<TQueryFnData, TPageParam>>
     | (() => NonUndefinedGuard<InfiniteData<TQueryFnData, TPageParam>>)
+    | undefined
```
TkDodo, line 45: pretty sure this should go to `UndefinedInitialDataInfiniteOptions`, like we have for `queryOptions`. compare:

https://github.com/TanStack/query/blob/fdb8ce1c38bc30fec9e984218b2fdc7802aef695/packages/react-query/src/queryOptions.ts#L15

TkDodo (changes requested):
can you please add a type test to `packages/react-query/src/__tests__/infiniteQueryOptions.test-d.tsx` that fails before and succeeds with your fix ?

packages/react-query/src/infiniteQueryOptions.ts
sungpaks: Thankyou,
I just saw that you mentioned it in the issue. I'll reflect this.

But one thing what I wonder is, what is difference between two of them : 
- `(() => NonUndefinedGuard<InfiniteData<TQueryFnData, TPageParam>>)`
- `InitialDataFunction<NonUndefinedGuard<InfiniteData<TQueryFnData, TPageParam>>>`

TkDodo: `InitialDataFunction` has `undefined` baked into the return value:

```
type InitialDataFunction<T> = () => T | undefined
```

sungpaks: Thank you for your kindness!

sungpaks:
> can you please add a type test to `packages/react-query/src/__tests__/infiniteQueryOptions.test-d.tsx` that fails before and succeeds with your fix ?

I added a type test that failed before and succeeds with my fix.

TkDodo (approved):

packages/react-query/src/infiniteQueryOptions.ts
```diff
@@ -20,7 +21,11 @@ export type UndefinedInitialDataInfiniteOptions<
   TQueryKey,
   TPageParam
 > & {
-  initialData?: undefined
+  initialData?:
+    | undefined
+    | InitialDataFunction<
+        NonUndefinedGuard<InfiniteData<TQueryFnData, TPageParam>>
```
jimmycallin, line 27: should this have been something like

```ts
  initialData?:
    | undefined
    | InfiniteData<TQueryFnData, TPageParam>
    | InitialDataFunction<
        NonUndefinedGuard<InfiniteData<TQueryFnData, TPageParam>>
```

to allow passing initialData as object and not within a callback function?

jimmycallin:
thanks for addressing my issue so quickly! i added a post-merge comment with a potential overlook. :)

packages/react-query/src/infiniteQueryOptions.ts
TkDodo: we can pass initialData as an object via `DefinedInitialDataInfiniteOptions`:

https://github.com/TanStack/query/blob/febbde676d0888d8dd4868177868a3316686cb58/packages/react-query/src/infiniteQueryOptions.ts#L41-L45

jimmycallin: sure, but it still won't work conditionally, see this modified playground based on the original example: 

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgVwM4FMCSA7AZsbYGdARWXSgE8AaOA-Q4siygeTBmAm1TgF84uKBBBwARAAEYAQx4yAxgGsA9FHTT5MALQBHclTEBuAFDHcybJq7YUGAOpRpYMBQAUSAkWDSANgBFpGX4ALkQ6Rm9-QOkAflCkdAAPaXAfdFCYKHJ+fgBKRGM4OHluVHg9FnZOUrgAXnCGIlJ9Ng5rVHdCorgKqgBpdEpQgG0xJJSwNLEAXWouot7KADFsUOlUSks4V3zagD5wr18AmTnuw85j6NDPS6iZee6YsLBpAHN0VBHbyJPpWbgrw+AAVpI4QF84MNpvxHkVQhYACboBjoRFnbofGAAOSSMFBILBKVCOzqBwAjBiij9fAT0KDwaFKV0+LkTHxTOZLNUbGh0A4nC4oMwqO4Lr9oiEwjT7rF4nBxql0nBMtkBKyCkU1DBkFBeRgcI0mC1OudFgMhlCxsklTMqT0Wis1hstqT9uKrqdHjK-jcIp7pHC4M83QcxUDPt9-bKARGGSlIdC8kGEdhkaj0Y8sbjEvj3vSiSASbsKfaZXT40W4Myiqz2UA

TkDodo: okay then what’s the fix please 😅

jimmycallin: I made a PR here :) https://github.com/TanStack/query/pull/8157

Hope that clarifies things, or perhaps I'm misunderstanding something!

PR #8227, opened 2024-10-27
docs (react-query): replace 'refetchType' with 'type'

PieterDePauw:
The key 'refetchType' seems to be used incorrectly in the documentation for "queryClient.invalidateQueries". The correct term seems to be 'type', which aligns with the actual implementation and usage within the codebase.


TkDodo:
no, `refetchType` is correct. `type` is a query filter (described in the query filter docs) that will narrow what will match for the invalidation. Then, you can further decide which queries to refetch with `refetchType`.

PR #8265, opened 2024-11-09
test(react-query): useInfinitePrefetchQuery -> usePrefetchInfiniteQuery

manudeli:


packages/react-query/src/__tests__/usePrefetchInfiniteQuery.test-d.tsx
```diff
@@ -1,7 +1,7 @@
 import { describe, expectTypeOf, it } from 'vitest'
 import { usePrefetchInfiniteQuery } from '..'
 
-describe('useInfinitePrefetchQuery', () => {
```
manudeli, line 4: usePrefetchInfiniteQuery is right

PR #8294, opened 2024-11-15
refactor(*): unify & use noop

manudeli:


packages/query-core/src/utils.ts
```diff
@@ -68,9 +68,7 @@ export type QueryTypeFilter = 'all' | 'active' | 'inactive'
 
 export const isServer = typeof window === 'undefined' || 'Deno' in globalThis
 
-export function noop(): undefined {
-  return undefined
-}
```
manudeli, line 73: unified this noop return type from undefined to void

packages/query-core/src/queryObserver.ts
```diff
@@ -349,7 +348,7 @@ export class QueryObserver<
     )
 
     if (!fetchOptions?.throwOnError) {
-      promise = promise.catch(noop)
+      promise = promise.catch(() => undefined)
```
TkDodo, line 351: why can’t we continue to use `noop` here ?

manudeli: If we use noop(function noop(): void {}, not function noop(): undefined {}), TypeScript error will be occured

TkDodo: at runtime, there’s no difference between a function that returns `undefined` and a function that returns `void`. would be good if we could keep `noop` here

manudeli: I agree, I reflect your review now in [92b4171](https://github.com/TanStack/query/pull/8294/commits/92b4171b1a7fdf5f4c179798201bb6b608bbd572)

TkDodo: nice 👍

TkDodo (approved):

packages/query-core/src/utils.ts
manudeli: Now @tanstack/query-core's noop's return type is undefined

packages/query-core/src/queryObserver.ts
TkDodo: hmm TypeScript doesn’t like that :/

manudeli:
I resolved it now😀

Scutua (approved):

Scutua (approved):

Scutua (approved):

PR #8295, opened 2024-11-15
fix(useQueries): make sure we don't lose property tracking

TkDodo:
if we call `combine` with `#this.result`, we don't have a tracked version of it, as tracking happens in useQueries (to avoid tracking internal access)


eriklynd:
After upgrading to this version from v5.59.20, `useQueries` got significantly slower. Maybe it is caused by this commit.

TkDodo:
potentially. what do you mean by “significantly slower” ? Can you show a reproduction and open a new issue please?

Scutua (approved):

PR #8299, opened 2024-11-16
fix(*): change noop function return type from undefined to void

kangju2000:
I modify the return type of noop function from `() => {}` to `noop():void` for type consistency.

### [AS-IS]
- noop function returns `undefined` while methods in query-core return `Promise<void>`
- Methods using `promise.catch(noop)` have inconsistent return types with their declarations
- `#executeFetch` method has the same issue where `Promise<TQueryData | undefined>` meets `promise.catch(noop)`

### [TO-BE]
- Explicitly type noop function to return `void`
- Replace `promise.catch(noop)` with `promise.catch(() => undefined)` in `#executeFetch` to match its return type
- Apply `noop():void` to other promise chains in query-core where return type is `Promise<void>`

### DIFF
--- a/packages/query-core/src/utils.ts
+++ b/packages/query-core/src/utils.ts
@@ -68,9 +68,7 @@ export type QueryTypeFilter = 'all' | 'active' | 'inactive'
 
 export const isServer = typeof window === 'undefined' || 'Deno' in globalThis
 
-export function noop(): undefined {
-  return undefined
-}
+export function noop(): void {}
 
 export function functionalUpdate<TInput, TOutput>(
   updater: Updater<TInput, TOutput>,
--- a/packages/query-core/src/utils.ts
+++ b/packages/query-core/src/utils.ts
@@ -68,9 +68,9 @@ export type QueryTypeFilter = 'all' | 'active' | 'inactive'
 
 export const isServer = typeof window === 'undefined' || 'Deno' in globalThis
 
-export function noop(): void {}
-
-export const returnUndefined = noop as () => undefined
+export function noop(): void
+export function noop(): undefined
+export function noop() {}
 
 export function functionalUpdate<TInput, TOutput>(
   updater: Updater<TInput, TOutput>,
--- a/packages/query-core/src/utils.ts
+++ b/packages/query-core/src/utils.ts
@@ -70,9 +70,7 @@ export const isServer = typeof window === 'undefined' || 'Deno' in globalThis
 
 export function noop(): void {}
 
-export function returnUndefined(): undefined {
-  return undefined
-}
+export const returnUndefined = noop as () => undefined
 
 export function functionalUpdate<TInput, TOutput>(
   updater: Updater<TInput, TOutput>,
--- a/packages/query-core/src/utils.ts
+++ b/packages/query-core/src/utils.ts
@@ -70,6 +70,10 @@ export const isServer = typeof window === 'undefined' || 'Deno' in globalThis
 
 export function noop(): void {}
 
+export function returnUndefined(): undefined {
+  return undefined
+}
+
 export function functionalUpdate<TInput, TOutput>(
   updater: Updater<TInput, TOutput>,
   input: TInput,
--- a/docs/framework/react/guides/infinite-queries.md
+++ b/docs/framework/react/guides/infinite-queries.md
@@ -110,7 +110,7 @@ To ensure a seamless querying process without conflicts, it's highly recommended
 [//]: # 'Example1'
 
 ```jsx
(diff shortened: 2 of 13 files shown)


TkDodo:
didn’t you just literally change the same thing @manudeli ?

- https://github.com/TanStack/query/pull/8294

I’m confused

manudeli:
> didn’t you just literally change the same thing @manudeli ?
> 
> * [refactor(*): unify & use noop #8294](https://github.com/TanStack/query/pull/8294)
> 
> I’m confused

@kangju2000 check this comment  https://github.com/TanStack/query/pull/8294#discussion_r1843688199 please

kangju2000:
Sorry for not checking PR https://github.com/TanStack/query/pull/8294#discussion_r1843688199. 
I tried to change it to `void` for semantic clarity of "no operation", but as discussed there, we need `undefined` for practical type safety in Promise chains. I agree with keeping `noop(): undefined`. 

What do you think about adding this JSDoc comment to help future contributors understand the design decision? @TkDodo @manudeli

```typescript
/**
 * No-operation function that returns undefined for type safety in Promise chains
 */
function noop(): undefined {
  return undefined
}
```

manudeli:
@TkDodo Since noop returning void and functions returning undefined are widely used in all packages,

How about providing both `function noop(): void {}` and `function returnUndefined(): undefined {}` in @tanstack/query-core and using them in all packages such as @tanstack/react-query?

TkDodo:
That's fine if TypeScript likes it everywhere 👍

saul-atomrigs:
If that makes typescript unhappy, we may just let it be inferred .. 

```ts
export function noop(): {
  return undefined
}
```

up to you guys 😅

kangju2000:
@TkDodo Separate `noop` and `returnUndefined` functions as reflected in f2820ac3774472949a606f6b07e314a96eeb7649

manudeli:
> @TkDodo Separate `noop` and `returnUndefined` functions as reflected in [f2820ac](https://github.com/TanStack/query/commit/f2820ac3774472949a606f6b07e314a96eeb7649)

Looking at this comment(https://github.com/TanStack/query/pull/8294#discussion_r1843688199), it seems that TkDodo did not want to implement noop and returnUndefined separately and include them in the package bundle since they are the same implementation in JavaScript.

So, how about implementing it as follows?

```tsx
// @tanstack/query-core
export function noop(): void {}
export const returnUndefined = noop as () => undefined

// @tanstack/react-query
import { noop, returnUndefined } from '@tanstack/query-core'

// @tanstack/vue-query
import { noop, returnUndefined } from '@tanstack/query-core'

// @tanstack/svelte-query
import { noop, returnUndefined } from '@tanstack/query-core'
```

kangju2000:
Thank you for the good suggestion! From a bundle size optimization perspective, type re-assertion seems better. I'll try changing it to export const returnUndefined = noop as () => undefined as suggested, and if there are any readability issues, we can consider rolling back to the previous approach.

Additionally, these utility functions seem to fall outside the core responsibilities of @tanstack/query-core and would be better handled through dedicated utility libraries.

packages/query-core/src/utils.ts
```diff
@@ -68,9 +68,9 @@ export type QueryTypeFilter = 'all' | 'active' | 'inactive'
 
 export const isServer = typeof window === 'undefined' || 'Deno' in globalThis
 
-export function noop(): undefined {
-  return undefined
-}
+export function noop(): void {}
+
+export const returnUndefined = noop as () => undefined
```
TkDodo, line 73: would overloads also work?

```
export function noop(): undefined
export function noop(): void
export function noop() {}
```

kangju2000: Thanks for the suggestion!🙌 reflected in 454956e23a54defc8ef208df55540d73ddebee0

TkDodo (approved):

TkDodo:
it doesn’t work. please make sure locally that everything compiles properly.

manudeli:
@kangju2000 You have to run script `test:types` https://github.com/TanStack/query/blob/main/package.json#L22 in root

kangju2000:
@TkDodo 

Sorry for the hassle. I've added type overrides for noop only where necessary. Using the overridden noop in the following code causes errors, so I kept it unchanged in solid-query.

```ts
// packages/solid-query/src/createQueries.ts:309
let unsubscribe = noop
createComputed<() => void>((cleanup) => {
  cleanup?.()
  unsubscribe = isRestoring() ? noop : subscribeToObserver()
  return () => queueMicrotask(unsubscribe)
})
```

PR #8305, opened 2024-11-19
fix(vue-query): fix queryOptions return type

jiwlee97:
I updated the return type of the queryOptions function with Vue's UnwrapRef.

[AS-IS]
image

[TO-BE]
image


DamianOsipiuk:
I have tried that before, but it would not work.

1. You can actually pass `options` as a `ref`, so this change would actually produce wrong types.
2. It breaks type inference in other cases.

I have yet to find a proper fix for this.

senaria:
Thanks for the great work @DamianOsipiuk. 

May I ask if you came further on this issue ? Would be really helpful to get  this solved.

PR #8309, opened 2024-11-19
add ability to warn when suspense queries aren't prefetched

juliusmarminge:
When using the promise-rsc-hydration thingy we aim to prefetch all queries in the RSCs and just consume the streamed promises in our components. 

This can be a bit fragile since forgetting to prefetch may not be obvious to find. This PR adds a new option that, when enabled, will log at warning level when `useSuspenseQuery` is called on a new query that's not already observed.

as with all options, this can be set globally when constructing the QueryClient and overridden locally cause you may have queries 

### Questions

- naming? idk, just went with the first thing that came to mind. as a matter of fact it's not even bound to being on the server as you may wanna use this with a `usePrefetchQuery -> useSuspenseQuery` pattern in a SPA as well
- given @TkDodo's [recent talk](https://youtu.be/QB-S6UU5YEU?si=RVi8kdV4jablIH4K&t=936), this should probably also accept a callback that gets the query(?) and not just a static boolean flag
- warning log might be too specific, would it be better for this to be a function and allow the developer to control what should happen? Perhaps they want to log more/less info about the query to help track down the query that's missing a prefetch

### DIFF
--- a/packages/query-core/src/types.ts
+++ b/packages/query-core/src/types.ts
@@ -223,6 +223,12 @@ export interface QueryOptions<
    * Maximum number of pages to store in the data of an infinite query.
    */
   maxPages?: number
+  /**
+   * If set to `true`, the query will warn if it is fetched on the server.
+   * Use this if you intend to only consume prefetched queries.
+   * Defaults to `false`.
+   */
+  warnOnServerFetches?: boolean
 }
 
 export interface InitialPageParam<TPageParam = unknown> {
--- a/integrations/react-next-15/app/client-component.tsx
+++ b/integrations/react-next-15/app/client-component.tsx
@@ -1,11 +1,11 @@
 'use client'
 
 import React from 'react'
-import { useQuery } from '@tanstack/react-query'
+import { useQuery, useSuspenseQuery } from '@tanstack/react-query'
 import { Temporal } from '@js-temporal/polyfill'
 
 export function ClientComponent() {
-  const query = useQuery({
+  const query = useSuspenseQuery({
     queryKey: ['data'],
     queryFn: async () => {
       await new Promise((r) => setTimeout(r, 1000))
@@ -16,13 +16,13 @@ export function ClientComponent() {
     },
   })
 
-  if (query.isPending) {
-    return <div>Loading...</div>
-  }
+  // if (query.isPending) {
+  //   return <div>Loading...</div>
+  // }
 
-  if (query.isError) {
-    return <div>An error has occurred!</div>
-  }
+  // if (query.isError) {
+  //   return <div>An error has occurred!</div>
+  // }
 
   return (
     <div>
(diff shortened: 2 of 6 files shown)


packages/query-core/src/types.ts
```diff
@@ -223,6 +223,12 @@ export interface QueryOptions<
    * Maximum number of pages to store in the data of an infinite query.
    */
   maxPages?: number
+  /**
+   * If set to `true`, the query will warn if it is fetched on the server.
+   * Use this if you intend to only consume prefetched queries.
+   * Defaults to `false`.
+   */
+  warnOnServerFetches?: boolean
```
stramel, line 231: bike shedding: discussed with co-workers, something like `strictPrefetch`/`strictPrefetches`? Happy with whatever the consensus is

rmolinamir: We could also consider different levels, so for example `strict` might lead to a runtime exception when compared to passing a `warn` value to simply log it 🤔 

> - given @TkDodo's [recent talk](https://youtu.be/QB-S6UU5YEU?si=RVi8kdV4jablIH4K&t=936), this should probably also accept a callback that gets the query(?) and not just a static boolean flag
> - warning log might be too specific, would it be better for this to be a function and allow the developer to control what should happen? Perhaps they want to log more/less info about the query to help track down the query that's missing a prefetch

I do like the callback approach more though, we could add an option called `onPrefetchMiss` or something along those lines.
This would give us more control over the outcome, e.g. could add observability, warning, throw an exception, etc.

juliusmarminge: I do like the callback approach more though, we could add an option called onPrefetchMiss or something along those lines.
This would give us more control over the outcome, e.g. could add observability, warning, throw an exception, etc.

yea definetely see this useful for much more than just warning for un-prefetched queries. would love @TkDodo's thoughts on what API he prefers

TkDodo:
I had as similar idea recently, but wanted to actually create some kind of `StrictMode` component that would only do this in dev-mode, but there, trigger a hard failure. Please have a look at the RFC:

- https://github.com/TanStack/query/discussions/8064

juliusmarminge:
Yea that's a much more thought out idea compared to this PR which I put together in 5 minutes after a meeting :)

TkDodo:
> Yea that's a much more thought out idea compared to this PR which I put together in 5 minutes after a meeting :)

😂 . Would you like to give implementing the first part of it a try? Specifically, the under-prefetching part, as it overlaps with what you were trying to do here:

> Whenever useSuspenseQuery would suspend the component with a new promise, it would log an error. This will alert developers that they are using a suspense query, but haven’t prefetched it. When the component renders, it should merely pick up a promise that is already in the cache. This would address the under-prefetching issue.

Bonus points if it would also work with the `use(promise)` experimental feature 😁

PR #8313, opened 2024-11-20
Feature/stale time on query

TkDodo:
DO NOT MERGE, contains potentially breaking changes

---

This PR moves the `staleTime` handling from `QueryObserver` to the `Query` itself, which mostly disallows different `staleTimes` on the same Query while being mounted simultaneously. Note that we can still have different `staleTimes` for different screens.

## How it worked before

Every observer had a `staleTime`, and every observer _also_ triggered a timer. When the timer was done, we triggered `createResult()` for the observer, which would update `isStale` on it. This means multiple timers were running - one for each observer - which could cause performance problems once you hit a certain threshold, at about 1k observers for the same query. Largely, those observers all have the same `staleTime`, so using one timer would be a lot better.

Additionally, I _think_ this implementation lead to cases where one observer had `isStale: true` while another observer could have `isStale: false`. But conceptually, `data` is either stale for a screen or it isn’t. It can’t be both, because that would mean we would need to show inconsistent data, which we can’t. As soon as one observer is marked as “stale”, a smart refetch would re-fetch data, thus showing it for all observers. So this isn’t really a case we should be supporting.

## How it works now

`staleTime` moved from `QueryObserverOptions` to `QueryOptions`. You can still pass it to `useQuery` or define a global default - nothing changed here from user perspective. But this means we’ll have `staleTime` on the `Query` itself, and since there can be only one Query per key, we’ll also only have one timer.

The timer will update itself when new options come in, or, when new data comes in for the query.

When the timer elapses, it will just set the `isInvalidated` flag on the Query itself to `true`. This is the same flag that we use for `queryClient.invalidateQueries`. Conceptually, I think this is neat because we don’t really need to distinguish between a Query that has been marked as “invalid” because it was invalidated by the user programmatically, or because the timer has elapsed.

Setting this flag will then inform all observers, so all screens are always up-to-date. `staleTime: Infinity` will just not set a timer.

What’s also neat about this is that we can now persist that setting, because it’s part of the query state. When restoring from localStorage, we already know if that data is stale or not. This wasn’t the case before because we would only know as soon as the first observer mounted.

### Alternative design: separate state

Note that I did consider not re-using the `isInvalidated` flag to keep the distinction between manual invalidations and timer-based ones, and use an `isStale` flag instead. I don’t think it is necessary, and it would actually lead to invalid states (yay, booleans), so we would need an enum, something like: `isStale: 'invalidated' | 'timer' | false` and I don't necessarily like that here / want to do a breaking change for that.

## staleTime: 0, a special case

Additionally, I’ve changed how `staleTime: 0` works, because it’s a special case. Previously, 0 was considered a valid timeout, and we set a `setTimeout(1)` on the observer, which would then trigger the query to be marked as stale (because we have to add +1 to the timeout for edge cases). This was pretty unnecessary - when staleTime is zero, we want the query to be stale immediately, without a timer.

So I added some special handling for this: When the query is created with staleTime zero, we set the `initialState` to `isInvalidated: true`. Also, whenever the query updates, we set it to `isInvalidated: true` for that `staleTime` immediately, without setting a timer.

Lastly, when the timeout changes and it would result in a stale query, we also update `isInvalidated`. Having a query with data from 10 seconds ago with a staleTime of 1 minute will therefore give us `isStale: false`, but if we update the `staleTime` (e.g. because it’s a function depending on `data`) to be 1 second, we will immediately transition to `isStale: true`)

---

This refactoring got rid of a lot complexity: We used to have `isStale` and `isStaleByTime` functions on the `query` and and additional `isStale` function on the observer. Now, it’s just:

```
isStale(): boolean {
  return this.state.isInvalidated || this.state.data === undefined
}
```

which is beautiful. Note that queries without data are always stale, this was the same before too (this is largely for the `error` case).

## disabled observers

Disabled observers were previously exempt from being stale; I made that because of some issue of what shows up in the devtools, but I think that was a mistake. Observers just show `data`, and the stale-ness refers to the data. Disabled observers also show data and update if that data changes, so it’s important to reflect that in the `isStale` property of those. This is where the majority of test updates comes from.

## why this could be breaking

- I had to delete a test where we used different `staleTimes` on the same query with two observers mounted at the same time, because it doesn’t make sense anymore with this implementation. If users do this, they might see changes in behaviour.
- queries are now always `isStale: true` immediately if they start out with `staleTime: 0`, while previously, they might have started with `isStale: false` followed by an immediate update.
- disabled observers now reflect the stale-ness of data.

All of these could be considered fixes as well 😅


TkDodo:
Note: I didn’t fix the other framework adapters yet - only the core and react, so that’s why the tests are failing.

packages/query-core/src/queryClient.ts
```diff
@@ -346,10 +343,9 @@ export class QueryClient {
     }
 
     const query = this.#queryCache.build(this, defaultedOptions)
+    query.setOptions(defaultedOptions)
 
-    return query.isStaleByTime(
-      resolveStaleTime(defaultedOptions.staleTime, query),
-    )
+    return query.isStale()
```
TkDodo, line 348: one of the more tricky things is that we now need to call `setOptions` for `fetchQuery` and `ensureQueryData` because they accept `staleTime` as an option, but they didn’t update the options, so calling `query.isStale()` wouldn’t reflect that.

I think this is also correct because if you pass different settings here, like a `gcTime`, you would want that to be reflected.
Note that `query.fetch` _will_ set the options for us under the hood, but it’s not guaranteed that it is invoked, because we might just return cached data from `fetchQuery. This is likely also an edge case bug in the current version 😅 .

packages/query-core/src/queryObserver.ts
```diff
@@ -395,8 +355,7 @@ export class QueryObserver<
     if (
       isServer ||
       resolveEnabled(this.options.enabled, this.#currentQuery) === false ||
-      !isValidTimeout(this.#currentRefetchInterval) ||
-      this.#currentRefetchInterval === 0
+      !isValidTimeout(this.#currentRefetchInterval)
```
TkDodo, line 358: note: timeout 0 is now not valid per default, so we can remove the extra check here for `refetchInterval`, too.

packages/query-core/src/removable.ts
```diff
@@ -11,7 +11,7 @@ export abstract class Removable {
   protected scheduleGc(): void {
     this.clearGcTimeout()
 
-    if (isValidTimeout(this.gcTime)) {
+    if (isValidTimeout(this.gcTime) || this.gcTime === 0) {
```
TkDodo, line 14: note: this was the _only_ place where we wanted 0 to be a valid timeout, so I re-added the extra check here. Otherwise, gcTime 0 would not garbage collect.

snewell92:
@TkDodo this is the most important thing here:

> ...we don’t really need to distinguish between a Query that has been marked as “invalid” because it was invalidated by the user programmatically, or because the timer has elapsed.

You have simplified a core concept while retaining essentially the same API. This will make maintaining and using TSQ easier.

The more you can remove or rip out, the better imo. Stay focused, stay lean.

packages/query-core/src/queryClient.ts
TkDodo: Reflected on this a bit more, and I don’t think the change here is correct. When we say:

```
queryClient.prefetchQuery({ queryKey, queryFn, staleTime: 10 * 60 * 1000 })
```

we basically want to express: prefetch this query _if_ data is older than 10 minutes.

it doesn’t mean we want  to set the staleTime of the query to 10 minutes. For example, if there is an active observer that has a staleTime of 2 minutes - that shouldn’t change.

So I need to revert that change and instead, somehow revive the `isStaleByTime` check for the queryClient methods.

Ephem:
First off, let me start by staying I agree with the goal of this PR, having a single timer for a query would be a great thing! I love the clear description and comments you've left. Let me also be clear that right now I don't have an opinion yet whether the approach in this PR is correct or not. Also I have yet only skimmed the technical details, to stay unbiased for this next part.

## `staleTime`, `stale` & `invalidated` - My mental model

I'm not quite sure how to start approaching this PR, but since it touches the very foundations of what `staleTime`, `staleness` and `invalidated` IS, how it should work and how we model it, I'll start by taking a step back to try to write down my high level mental model.

> [!NOTE]
> I don't write this as a guide or set of truths, but to expose it so you can poke holes in my reasoning.

I think I see `staleTime` and `staleness` as less tightly connected than others, let's start with `staleTime`.

### `staleTime`

In my head, `staleTime` is an option which gets its meaning from the context it is being used. If I `useQuery`, `staleTime` means roughly "this is how up to date my data needs to be in this place in the app". (This is complicated by the fact that data doesn't _refetch_ after this time unless you configure it to, it only applies to _mount_, _reconnect_ and _window focus_ depending on options, but let's ignore that.)

Having the need for different "freshness" at different positions in the app is a natural thing.

> [!NOTE]
> Feedback: I don't think having two observers with different `staleTime` mounted at the same time is abnormal or worth warning about. It's something we should support without fuss.
>
> With this warning, just inserting a component which happens to have data fetching someplace new might result in a warning that's hard to take action on.

As you've noted, `staleTime` in the context of `(pre)fetchQuery`/`ensureQueryData` is tricky. On the one hand, it might mean "if this data already exists in the cache and is fresher than this, just return it" - that is, the same meaning as `useQuery`. On the other hand it might additionally mean "update the query `staleTime` to this"?

> Question: Can a query have a `staleTime`?

I think this is the crux of the problem. Yes, but also no. Obviously, you can set a global `staleTime`, so in some practical sense, it can. I _think_ I only see this as a default option though, the `staleTime` should still _live_ on the observer. Can a query with no observers ever be considered stale? It's complex, but I think I would argue no. A query is only stale in the eyes of the observers (pun intended), BUT, see next section.

### `staleness`

While I see `staleTime` as something that is the eyes of the observer, a query's current `staleness` is definitely something that lives at the query level. It's globally boolean. Though not implemented that way, in my head, I actually see this is a derived state! "If data is older than all current observers _needs_ (expressed as `staleTime`), the query is stale, otherwise not". This fact is what would let us move the timer to the query. It makes no sense that a query is only _stale_ for one observer, but not for another. While different observers might have different needs, the most needy one wins and when it's stale it should be stale for all.

If it _was_ a derived state, we might not need a timer at all? I'm sure there are good reasons it might not be feasible to implement this way, but it's an interesting thought experiment.

### `invalidated`

In my head this is different from `staleness`. Staleness is about expressing how important fresh data is in different parts of the application, invalidation is when we _know_ some piece of data is no longer up to date, or at least when strongly suspect it and want to treat it that way. This means the invalidated state belongs to the query, not the observers. Also, while to me it makes no sense for a query without observers to be `stale`, it does makes sense for a query with no observers to be `invalidated`.

## Reflections

All of the above is high level and doesn't necessarily have to reflect what we call things in the code. When I talk about staleness and invalidation, I'm not talking about the existing variables in the code, It's not a comment on whether we should use `isInvalidated` to keep track of staleness or not etc, etc.

I wanted to write this out first to see how much common ground we have in our mental models, but I will come back with more practical feedback on the PR. I'll review the code more closely and I have a few test cases in mind I will check if they already exist or if I can write them. :smile:

DamianOsipiuk:
This makes sense and should simplify things. 🚀 

One general note though is about current usage, and potential misuse by users.

Considering that `same query` can be reused in `multiple components` and have different `staleTime` set for each usage place.  
It might lead to a cascade of updates, where `different observers` might try to update `staleTime` one after the other.

- Observer 1 sets `staleTime` to `1` marking query as `stale`
- Observer 2 sets `staleTime` to `10` marking query as `not stale`
- Observer 3 sets `staleTime` to `3` marking query as `stale`
- etc..

I imagine this can be a problem even now.  
Should we actually keep track of quickly changing options like `staleTime` and warn users in `dev` that they might have a configuration problem? Should be easier to do that now, when we keep track of it in one place.

TkDodo:
> It might lead to a cascade of updates, where different observers might try to update staleTime one after the other.

This is a problem in the new implementation, yes, and the test I deleted was actually failing because of that 😂 . Well spotted.

It’s not an issue in the current implementation because each observer has its own timer, and when the timer elapses, it only updates the result of this observer. Nothing in the query is updated, so there are no chances to influence other observers.

---

After reading @Ephem’s comment, and talking about this with him some more, I think I want to try a different approach:

- keep the `staleTime` option on the observer
- move only the timer to the query
- the timer would update itself when an observer is added, or removed, or when the options of an observer change (different staleTime).
- The timer would then be set to the smallest staleTime value of all current observers. I still need to think about if `disabled` observer should be taken into account here or not. What do you think?
- we would _not_ set `invalidated:true` on the query when the timer elapses - we would just inform observers about that, so that they can re-calculate the `isStale` prop. There are edge cases around `prefetchQuery` where it makes sense to keep those two decoupled - e.g. a hard call to `invalidateQueries` should make a `prefetchQuery` fetch, even if it has `staleTime: Infinity`.
- Another open question is: if one observer has staleTime 5 seconds, and the other 10 minutes, would we update _both_ observer results to show `isStale: true` after 5 seconds? Right now, I think one would show `true` while the other would still show `false`, because it only checks its own `options.staleTime`. But if we implement the query’s `staleTime` as: the derived state of all currently subscribed observer’s staleTime, it would be 5 seconds for both of them, in this context.

TkDodo:
> The timer would then be set to the smallest staleTime value of all current observers. I still need to think about if disabled observer should be taken into account here or not. What do you think?

My thought process would be:

since a disabled observer can’t trigger fetches on its own, it also shouldn’t influence when that data is considered “outdated”, so it should be excluded from deriving the staleTime value.

> would we update both observer results to show isStale: true after 5 seconds?

for this, I would say YES. also, disabled observer should get that update. If we treat `isStale` as basically derived state of the query (derived from the currently mounted, enabled observers), then we can always know if a “query” is stale. This is what should be displayed in the devtools, and this is what all observers should return, even if it means that you get `isStale: true` after 5 seconds even though you have set it to 10 minutes, as the 10 minutes might not count in this context when there’s a second observer with a smaller time.

“disabled observers not getting the correct `isStale` update” was posted as a question recently:
- https://github.com/TanStack/query/discussions/8306#discussion-7514038

and would be fixed implicitly with this proposal, too.

DamianOsipiuk:
> The timer would then be set to the smallest staleTime value of all current observers. I still need to think about if disabled observer should be taken into account here or not. What do you think?

When i think about `disabled` observers, those usually hang in that state due to two cases:
- you disable the observer cause you are missing one of the parameters that is needed to `fetch` properly
  - but ultimately it will result in a different `queryKey`, so it should not be a problem here.
- you want this query to be `lazy` and fetch only on user interaction
  - again if you are waiting for user input to construct better `queryKey`, it should not matter
  - you want to control exactly when fetch happens
    - why would you set `staleTime` in this case? You would potentially use `refetch` or `skipToken`
    - since you can potentially not set `staleTime` it will result in `0` being the default, so IMO it should not be considered for `staleTime` calculation of other observers
    
> Another open question is: if one observer has staleTime 5 seconds, and the other 10 minutes, would we update both observer results to show isStale: true after 5 seconds? Right now, I think one would show true while the other would still show false, because it only checks its own options.staleTime. But if we implement the query’s staleTime as: the derived state of all currently subscribed observer’s staleTime, it would be 5 seconds for both of them, in this context.

I would say, it should say `stale` everywhere, and we should potentially show a `dev warning` about mismatching configuration.  
Why? Cause if any of those observers happen to trigger `refetch` it will update data in all of them, despite them showing as `not stale`.  
On the other hand, `it shows as stale, but if i calculate time from fetch it should not be. a bug!` -> thus dev warning.

TkDodo:
> so IMO it should not be considered for staleTime calculation of other observers

seems we are aligned on that 👍 

> I would say, it should say stale everywhere

I also agree with that

> and we should potentially show a dev warning about mismatching configuration.

I also added a dev warning, but @Ephem is skeptical about it, and I kinda agree. Imagine a scenario where you have two observers on the same query: one selects the `count` (like the number of issues), and you want that to be always up-to-date, so you set a small `staleTime`. Then, you have another screen where you show the full list, which has a higher `staleTime`. 

This is fine, but now you refactor a bit and put the counter also in the header of the list page. Yes, this would make the whole list get the smaller staleTime, but is it worth warning about? The only way out would be to make that component customizable to take in a different staleTime, so you have them be the same on the same screen.

It gets worse if you have a component that opens in a dialog that sets a different staleTime, because that’s only for the time when that dialog is opened.

Finally, in React Native, screens might still be mounted even though they are not visible, so you might really have multiple observers with different staleTimes mounted. That _could_ be seen as a config error, but we actually want to ship an improvement to properly disconnect those observers (see https://github.com/TanStack/query/pull/8348).

All in all I don’t think a warning in dev mode is warranted, but it’s not hard to implement and we can think about it. The things we log in dev mode can’t be turned off (because they should be actionable), and I don’t think these warnings would be something that the user can or should act upon.

TkDodo:
Another interesting question that @Ephem had:

If we have two observers, one with `staleTime: 2seconds` and one with `staleTime: 10minutes`, and we get, a window focus event after 5 minutes, should we `refetchOnWindowFocus` ? YES. I think this case is clear - the window focus event will look at all observers to determine query stale-ness.

BUT, what if another observer with `staleTime: 1hour` mounts after 5 minutes - should we get a `refetchOnMount` for that one? We could argue both ways:

- Yes, because the query is, in this context, seen as stale after 2 seconds, so it’s stale right now, so we should refetch.
- No, because this specific observer that gets mounted says `staleTime` is 1h, so why should it?

I think the answer should still be YES, just given by the fact that we will return `isStale: true` (as discussed above), and it would be weird to not trigger a refetch even though we get that value returned.

Just wanted to note down my thoughts on it while they are still fresh 😂

snewell92:
@TkDodo your last comment confused me slightly, ~~I'm going to attempt to argue for the "NO" case in the 1hour mounted after 5 minutes case~~ I will argue for both because live is full of surprises, **final edit actually I want it configurable** and restate some of your scenario to see if I can align my brain properly.

Pls correct me where/when I'm wrong or don't have the concept clear.

Restate concepts: I think of staleTime as declarative, invalidation as imperative, and staleness as derived from staleTime. I will simplistically think of observers as instances of useQuery calls (with or without a selection), and queries as the keyed query function we all know and love.

Key Question: **Should staleness be exclusively a property of queries, observers or actually both?** To retain the utility of expressing selections of the same query having different priority & use in an app, I think **staleness should be exclusively a property of observers**. IE a query cannot be stale, only observers of queries can be stale.

<details><summary>On Staleness</summary>

I find this approach of observers **only** having staleness simple and useful. It does not
result in any loss of functionality, because you merge/inherit global staleTime into each observer's staleTime, and the timer is still on the query, TSQ has to iterate over all subscribed observers to determine what is stale (optimization: take minimum of active observers? what is active? is there eagerness?).

This may be false or wrong in ways I currently don't understand.

</details> 

Restated scenario:
1. There is one query.
2. There are two observers on mount, `staleTime: 2sec`, and `staleTime: 10 minutes`.
3. There is a third observer mounted after 5 minutes with `staleTime: 1hour` after some navigation in an app.
  a. This mount happens on navigation, and the first two observers are not on this new page.
  b. This mount _adds_ to the page, so **all observers are 'active'**.

I will rephrase your question, instead of refetch on mount, I would simply ask - what is stale when our third observer is mounted? If staleness is only a property of observers, then it is easy imo.

In my sub-scenario a, the active observer, the third one, is not stale so we don't refetch, we can serve from cache and respect what the author has declared via staleTime.

In my sub-scenario b, we have this new 'mount' event, and we have an active observer that has declared staleness, so we could use this as an opportunity to refetch so we respect the first observer's staleness. So maybe both? Buuut.. consider that the new component is a descendant, do we want some deep descendant who is mounting cause a parent to get new data? Actually, probably not. If we navigated away and came back, then yes I would expect fresh data. But if I'm revealing more information on a sub tab or table or expander, I wouldn't expect my top level nav to parent to get fresh information unless I configured it that way, and perhaps there isn't a way to configure that scenario?

This makes me think that in the case of a "far away" observer being mounted, we may want a new option in QueryOptions to mark an observer as "eager" (is this already a thing?) so if this scenario happens, TSQ always iterates and takes the minimum of all active observers to see if anyone wants fresh data, and if so it refetches.

This may cause some weirdness though, we can technically serve stale data to the third component while refetching, and then it could update. But I guess that's what isRefetching is for anyways so nvm (hence my "eager" or opt in suggestion).

<details><summary>Sorry for the essay, here's a picture of one of our cats</summary>

This is Pippin and he gets fed because I and all our teams use TSR and TSQ everyday <3.

image

</details>

snewell92:
@TkDodo I'm now realizing your scenario of a window focus event is like going back to the tab after getting distracted on youtube (not me ofc), not navigating in an app. In which case the logic above would say "YES" you have a stale observer of the shared query and need to refetch.

TkDodo:
@snewell92 thank you for this, really, there are good insights here. I think we are mostly aligned on what we want, namely:

> I think of staleTime as declarative, invalidation as imperative, and staleness as derived from staleTime. I will simplistically think of observers as instances of useQuery calls (with or without a selection), and queries as the keyed query function we all know and love.

This is spot on 👍 

> I think staleness should be exclusively a property of observers. IE a query cannot be stale, only observers of queries can be stale.

Yep 👍 

> you merge/inherit global staleTime into each observer's staleTime, and the timer is still on the query, TSQ has to iterate over all subscribed observers to determine what is stale

Also 👍 . What’s currently on `main` is that each observer gets a timer, so with 100 observers which all have the same `staleTime`, you get 100 timers instead of 1. This is mainly what we want to change by moving the timer only to the query. This PR moves the `staleTime` itself there. This is likely a mistake. I will close this PR to avoid further confusion. I’m actually working on a new PR that implements what I’ve described [here](https://github.com/TanStack/query/pull/8313#issuecomment-2498055322).

---

Regarding the scenario:

> a: this mount happens on navigation, and the first two observers are not on this new page.

Since the first two observers aren’t on the new page, they are irrelevant, because they get unmounted. So on this new page, there is only one observer with `staleTime: 1hour`. Since the mount happens after 5 minutes, you would _not_ get a `refetchOnMount`, and the timer would be set to 55 minutes.

> b. This mount adds to the page, so all observers are 'active'.

Now we have 2 observers when we add a 3rd one, so we take the minimum `staleTime` of all 3 observers and potentially adjust the timer. In this case, the lowest `staleTime` is  `2seconds`, so the mount of the 3rd observer with a higher `staleTime` doesn’t change that. The current timer of `2seconds` (lowest staleTime) has already elapsed, so when the 3rd observer mounts, we are already stale, so we will see a background refetch, and after that, a new `2second` timer will be started.

It might seem a bit confusing that we get a background refetch here even though the `staleTime` of the observer who mounted is longer. But it makes more sense for automatic refetches (mount, window focus, reconnect) to consider all observers. window focus would also re-fetch because there is no single observer that triggers it. Here, in this context, the stale-ness is defined by the lowest observer (`2seconds`), so a new observer mount will have to adhere to the context and refetch.

Does that makes sense?

Issue #8277, opened 2024-11-11
Next js 15 with dynamicIO: true returned server error


yudistiraashadi:
### Describe the bug

Running the "React Example: Nextjs App Prefetching"  [from the docs](https://tanstack.com/query/latest/docs/framework/react/examples/nextjs-app-prefetching) with `dynamicIO: true` setting on next.config.ts return this error shown below. The code is available [here](https://github.com/yudistiraashadi/next15-tanstack-query-canary).

![Image](https://github.com/user-attachments/assets/17c427ac-23d7-4529-9e45-a2fe3c9964ab)

### Your minimal, reproducible example

https://github.com/yudistiraashadi/next15-tanstack-query-canary

### Steps to reproduce

1. Install next.js 15 canary `npx create-next-app@canary`
2. Install tanstack query and devtools `pnpm add @tanstack/react-query` and `pnpm add -D @tanstack/react-query-devtools`
3. Copy all the code from the "React Example: Nextjs App Prefetching" [here](https://tanstack.com/query/latest/docs/framework/react/examples/nextjs-app-prefetching), to replace all the code inside src folders on the nextjs app folder (or /src folder, I'm using src folder btw)
4. Run dev server `pnpm run dev` and open localhost:3000

### Expected behavior

No error

### How often does this bug happen?

Every time

### Screenshots or Videos

https://github.com/user-attachments/assets/54bbb464-f4a2-4030-b307-04df16852898

### Platform

- OS: Ubuntu 22.04
- Browser: Chrome
- Tanstack Query version: 5.59.20
- Next.js 15 canary: 15.0.4-canary.5

### Tanstack Query adapter

None

### TanStack Query version

5.59.20

### TypeScript version

5.6.3

TkDodo:
Not sure if I discussed this with you, but it came up on Discord as well. The reason is likely that we initiate our mutationIds with `Date.now`:

https://github.com/TanStack/query/blob/24f1d45cbd3a252e2c8bb3471591502a2418b6ad/packages/query-core/src/mutationCache.ts#L91

It doesn’t have to be `Date.now`, but it needs to be a unique id that we can increment and that doesn’t overlap with ids when they become restored them from localStorage.

If you have an idea what that could be, please file a PR.

TkDodo:
Actually, we only serialize the `scope`, which defaults to `String(mutation.mutationId)`:

[...]

PR #8315, opened 2024-11-20
fix(query-core): use performance instead of Date.now to initialize mutationId

ali-idrizi:
Previously the `MutationCache.#mutationId` was initialized with `Date.now()`. When using `dynamicIo` with Next it was throwing:

```
Error: Route "/" used `Date.now()` instead of using `performance` or without explicitly calling `await connection()` beforehand. See more info here: https://nextjs.org/docs/messages/next-prerender-current-time
```

The error is fixed by using `performance.timeOrigin + performance.now()` instead.

I have updated the `next` and `react` dependencies on the Next 15 integration and example from RC to stable. However, the `dynamicIo` is still an experimental feature and requires the canary dependency, so this not tested. I can confirm that the error was thrown locally by `react-next-15` integration when enabling `dynamicIo`, and this change fixes it.

Closes https://github.com/TanStack/query/issues/8277

### DIFF
--- a/integrations/react-next-15/package.json
+++ b/integrations/react-next-15/package.json
@@ -9,9 +9,9 @@
     "@js-temporal/polyfill": "^0.4.4",
     "@tanstack/react-query": "workspace:*",
     "@tanstack/react-query-devtools": "workspace:*",
-    "next": "^15.0.0-rc.0",
-    "react": "19.0.0-rc-4c2e457c7c-20240522",
-    "react-dom": "19.0.0-rc-4c2e457c7c-20240522",
+    "next": "^15.0.3",
+    "react": "19.0.0-rc-7ac8e612-20241113",
+    "react-dom": "19.0.0-rc-7ac8e612-20241113",
     "tupleson": "0.23.1"
   },
   "devDependencies": {
--- a/integrations/react-next-15/package.json
+++ b/integrations/react-next-15/package.json
@@ -9,9 +9,9 @@
     "@js-temporal/polyfill": "^0.4.4",
     "@tanstack/react-query": "workspace:*",
     "@tanstack/react-query-devtools": "workspace:*",
-    "next": "^15.0.3",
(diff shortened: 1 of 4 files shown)


integrations/react-next-15/package.json
```diff
@@ -9,9 +9,9 @@
     "@js-temporal/polyfill": "^0.4.4",
     "@tanstack/react-query": "workspace:*",
     "@tanstack/react-query-devtools": "workspace:*",
-    "next": "^15.0.0-rc.0",
-    "react": "19.0.0-rc-4c2e457c7c-20240522",
-    "react-dom": "19.0.0-rc-4c2e457c7c-20240522",
+    "next": "^15.0.3",
+    "react": "19.0.0-rc-7ac8e612-20241113",
+    "react-dom": "19.0.0-rc-7ac8e612-20241113",
```
TkDodo, line 14: can we please use `19.0.0-rc.1`, and use it _everywhere_. I’d really like to have the versions streamlined, thanks 🙏

packages/query-core/src/mutationCache.ts
```diff
@@ -88,7 +88,7 @@ export class MutationCache extends Subscribable<MutationCacheListener> {
   constructor(public config: MutationCacheConfig = {}) {
     super()
     this.#mutations = new Map()
-    this.#mutationId = Date.now()
+    this.#mutationId = performance.timeOrigin + performance.now()
```
TkDodo, line 91: should we `Math.round` this to get a full number ?

samyung0:
Any update on this?

TkDodo:
we have a second PR open that solves the issue in a different way, so we have to decide which one is going to make it.

TkDodo:
we’ll handle this with #8451 without the performance API, as this might also be excluded from prerenders in the future. see:

https://github.com/TanStack/query/pull/8450#issuecomment-2563076253

PR #8332, opened 2024-11-23
Add types for QueryFilters which flow down to Query<> and QueryKey/DataTag types

Nick-Lucas:
At tRPC we're working on a new react query client which only provides factories to React Query interfaces. Many of these factories already support full type-safety, for instance a QueryKey can have a DataTag which hints at the query data type.

QueryKey (with DataTag) and Query<> both support typesafety, but QueryFilters hides this and only exposes the defaults. This means our queryFilter factory method cannot be type-safe without replicating the tanstack query types into the tRPC codebase. 

I propose upstreaming this type-safety so that everyone can benefit and tRPC can utilise it natively

Here is the work I'm doing on the tRPC side: https://github.com/trpc/trpc/pull/6265


TkDodo (commented):
looks lovely. Can you maybe add a type test please (in a `.test-d.ts` file) so that we are future proof and can see per example what this achieves? Thank you 🙏

Nick-Lucas:
> test-d.ts

Done

Good I'm doing this. I think getQueryState could be next on my agenda, DataTag has no way to communicate an error type but getQueryState can get you back the error. I've left a test with a comment for now to document the current behaviour

TkDodo (approved):

PR #8348, opened 2024-11-25
feat(react-query): allow useQuery and useQueries to unsubscribe from the query cache with an option

TkDodo:


TkDodo:
@hirbod could you help me re-vamp the React Native docs page if we ship this?

hirbod:
@TkDodo Sure. How should we proceed? Should we keep what we have, replace it, or additionally add the new solution?
While we're at it, we should also add a section for [react native devtools](https://github.com/expo/dev-plugins/tree/main/packages/react-query)

I love that plugin

TkDodo:
Yeah there are some mentions about devtools on the top of the page, as well as on the devtools page: https://tanstack.com/query/v5/docs/framework/react/devtools

I think what I would want is:

- the devtools page mentions that devtools exist as 3rd party integrations for RN, and it should link to the react native guide: https://tanstack.com/query/v5/docs/framework/react/react-native
- the react native guide should have a `Devtools` section, listing all the options

Regarding the sections, I think we should do:

- Devtools
- Online status management
- Refetch on app focus
- The new thing :)

The new thing should cover what you can do with subscribe, maybe show the standard example with  react-navigation, maybe show a more advanced example with the react-native stack (e.g. keep the top two screen subscribed)

I don’t think we need these anymore:

- Refresh on Screen focus
- Disable re-renders on our of focus Screens
- Disable queries on out of focus screens

if we can cover everything with the `subscribed` flag

TkDodo:
There’s a new prop: `subscribed: boolean`. You can set that to false to detach the observer from the component.

TkDodo:
so:

```
const subscribed = useShouldBeSubscribed()

useQuery({
  queryKey,
  queryFn,
  subscribed
})
```

where `useShouldBeSubscribed` needs to figure out if it’s the top screen or not, in a reactive way (which is something that I was hoping you could fill in for me because I don’t know how that’s done)

hirbod:
Well, basically, `subscribed` should use  (https://reactnavigation.org/docs/use-is-focused

If we want to keep both the second and the topmost `subscribed`, we’d need to use a mix of [useNavigationState](https://reactnavigation.org/docs/use-navigation-state/) `useIsFocused`.

Untested, something like this:

```tsx
const isTopScreen = useNavigationState((state) => {
    const currentRouteKey = route.key;
    const topRouteKey = state.routes[state.routes.length - 1].key;
    return currentRouteKey === topRouteKey;
});

const isSecondScreen = useNavigationState((state) => {
    if (state.routes.length > 1) {
        const secondRouteKey = state.routes[state.routes.length - 2].key;
        return route.key === secondRouteKey;
    }
    return false;
});
```

IMO, overkill. Since focus should ideally trigger a `refetchOnWindowFocus`, a disabled subscription would be enabled the moment `useIsFocused` changes again, which I believe is sufficient.

flodlc:
Great to see that the experience with focused screens in React Native is improving.
Just one thing: while the subscribe flag seems like a good solution, it will need to be set to true or false during navigation. This will trigger a re-render on navigation, which is not recommended for smooth, high-quality transitions.

Would an imperative function do the job instead?

Currently, I’m doing something like this, but I’m encountering issues with the isStale() function retuning false while it's actualy stale.

```typescript
export const useEnabledOnFocus = (enabled = true): ((query: Query) => boolean) => {
  const isFocused = useRef(false);
  const query = useRef<Query<unknown, Error, unknown, QueryKey>>();
  useFocusEffect(
    React.useCallback(() => {
      isFocused.current = true;
      const task = InteractionManager.runAfterInteractions(() => {
        if (
          enabled &&
          ((query.current?.state.status === "pending" &&
            query.current?.state.fetchStatus === "idle") ||
            query.current?.isStale())
        ) {
          console.log("fetch");
          query.current?.fetch();
        }
      });

      return () => {
        isFocused.current = false;
        task.cancel();
      };
    }, [])
  );

  return (_query: Query<unknown, Error, unknown, QueryKey>) => {
    query.current = _query;
    const computedEnabled = enabled && isFocused.current;
    return computedEnabled;
  };
};
```

Usage:
```typescript
useQuery({
    ...,
    enabled: useFocusQuery()
})
```

hirbod:
FYI, runAfterInteractions does not fire in native-stack, only in the JS stack. And since the native-stack is the recommended way, re-renders are not a concern, as the navigation runs on the UI thread.

flodlc:
Ok, thanks for the explanation ! Can't wait to try the new api

TkDodo:
@hirbod did you have some time to try out the preview build?

hirbod:
@TkDodo will do this weekend. The week was very tough with our release and hotfixes :D

hirbod:
@flodlc There’s one with nearly 100% feature parity and React Navigation adapters (also works with Expo Router): https://github.com/okwasniewski/react-native-bottom-tabs.  

Either way, bottom tabs simply switch screens without animation, so there’s no such thing as jank with them.

@TkDodo finally have so time, pulling now and testing

hirbod:
@TkDodo Also, check Discord, but FYI: I gave it a thorough test—it works perfectly fine. LGTM!

flodlc:
Actually I was thinking about the react navigation implementation that allows "shift" animation.
Not sure if it really impact the animation to rerender though

flodlc:
Hi there! Wishing you the best for 2025 🍾
Are there any updates on this PR? I can't wait to try it on production.

TkDodo:
@flodlc docs are still missing. I was hoping that @hirbod was working on them, but if you want, you can maybe join the effort?

flodlc:
Sure! I'll work on it today :)

hirbod:
@TkDodo Sorry! The holidays and a few other OSS tasks I've been working on lately completely blocked me, and I forgot about the docs. The feature is definitely ready and already tested. If @flodlc wants to finish the docs, that'd be great. Otherwise, I'll try to free up a bit of time tomorrow.

We basically just need to explain how to use this new prop correctly in React Native, using available hooks like `useIsFocused` from `react-navigation` to enable or disable `useQuery` hooks that might still be mounted but invisible.

flodlc:
Here it is: https://github.com/TanStack/query/pull/8506

Scutua (approved):

Scutua (approved):

PR #8351, opened 2024-11-25
fix(QueriesObserver): fix improper sorting in QueriesObserver's #find…

jmtoung:
…MatchingObservers method

This commit fixes a bug in the QueriesObserver's #findMatchingObservers. The bug is present when there are duplicate unsorted queries, for example, ['A', 'B', 'A']. The bug results in #findMatchingObservers to return the queries in sorted order, or ['A', 'A', 'B'], which results in a mismatch between Array<QueryObserverOptions> and Array<QueryObserverMatch>. This bug was introduced in https://github.com/TanStack/query/commit/54995777af37d219f34f5a03d18fbc1abb06b8a5


packages/query-core/src/queriesObserver.ts
```diff
@@ -240,28 +240,14 @@ export class QueriesObserver<
...
       } else {
-        const existingObserver = this.#observers.find(
-          (o) => o.options.queryHash === defaultedOptions.queryHash,
-        )
         observers.push({
           defaultedQueryOptions: defaultedOptions,
-          observer:
-            existingObserver ??
-            new QueryObserver(this.#client, defaultedOptions),
+          observer: new QueryObserver(this.#client, defaultedOptions),
```
TkDodo, line 245: I doubt that this is a correct fix, and I’d expect some tests to fail. This means that a re-render with the same queries would throw away observers and re-create them (see the `setQueries` function)

TkDodo (changes requested):
Having the same key inside `useQueries` multiple times is not something that we support. I’m sure we have a discussion / issue about this where we mentioned this. Maybe documenting this is the better way to go.

jmtoung:
> Having the same key inside `useQueries` multiple times is not something that we support. I’m sure we have a discussion / issue about this where we mentioned this. Maybe documenting this is the better way to go.

Not supporting same key is fine, but this bug was introduced by this recent change https://github.com/TanStack/query/commit/54995777af37d219f34f5a03d18fbc1abb06b8a5 

Prior to this change, duplicate keys wouldn't result in a mismatch between input and output.

TkDodo:
> Prior to this change, duplicate keys wouldn't result in a mismatch between input and output.

but “duplicate keys” and “same keys” are the same thing? Our warning even says that the behaviour is potentially unexpected.

packages/query-core/src/queriesObserver.ts
TkDodo: yeah I think you’re right - it seems like this code doesn’t do anything. Good find 👍

jmtoung:
> oh, there should even be a warning in development mode. are you not seeing that ?
> 
> * [fix(useQueries): warn on duplicate keys #8246](https://github.com/TanStack/query/pull/8246)

I am seeing the warning, but I think we should still fix it to return the correct order, even in spite of duplicate queries. 

In other words, after the bug was introduced, `useQueries(['A', 'B', 'A'])` returns `['A', 'A', B']`, whereas it should return `['A', 'B', 'A']`. This bug was recently introduced here https://github.com/TanStack/query/commit/54995777af37d219f34f5a03d18fbc1abb06b8a5. 

In that commit, we refactored things to efficiently loop over `Array<QueryObserverOptions>` once, but we introduced a bad sort that sorts by `options.queryHash`

Aghassi (approved):

PR #8361, opened 2024-11-27
Include optional TError on DataTag so that getQueryState can infer the error type correctly

Nick-Lucas:
Current `getQueryState` can infer the data type correctly via the DataTag type.

DataTag does not include any type information for the Error type however.

At tRPC we are working on a package to generate query keys, query filters, and query options automatically based on your tRPC type. This will provide the ability to call `getQueryState` but currently no ability to create inference for the error because of the current limitations of DataTag. https://github.com/trpc/trpc/pull/6265 - similar to https://github.com/TanStack/query/pull/8332 it seems right to improve this upstream so everybody can benefit, though this time it also can't be easily improved downstream

This PR updates DataTag with a 3rd optional type parameter for TError, defaulting to `unknown` to avoid breaking changes for RQ users. getQueryState now infers the error type from this information or automatically falls back on its own TError value if the DataTag is unset

Alternative implementation could be to do an `ErrorTag` alongside `DataTag` but this one seems more in-line with how type params are set up elsewhere in RQ


Nick-Lucas:
The tRPC PR now has this implemented and looks good, though you'd need to locally link to this PR to make it work: https://github.com/trpc/trpc/pull/6265

`pnpm up @tanstack/react-query@path/to/tanstack-query/packages/react-query -r`

packages/query-core/src/queryClient.ts
```diff
@@ -229,12 +229,22 @@ export class QueryClient {
...
       : TQueryFnData,
+    TInferredError = TTaggedQueryKey extends DataTag<
+      unknown,
+      unknown,
+      infer TaggedError
+    >
+      ? unknown extends TaggedError
+        ? TError
+        : TaggedError
+      : TError,
```
TkDodo, line 240: why is this extra `unknown extends TaggedError` necessary? We don’t do that for `TInferredQueryFnData` 🤔

packages/query-core/src/utils.ts
TkDodo, line 1: slight mistake made it in here, can you please address that:

```diff
export interface QueryFilters<
  TQueryFnData = unknown,
- TError = Error,
+ TError = DefaultError,
  TData = TQueryFnData,
  TQueryKey extends QueryKey = QueryKey

File: packages/query-core/src/types.ts [hunk 2]
```diff
@@ -43,8 +43,10 @@ export type DefaultError = Register extends {
 export type QueryKey = ReadonlyArray<unknown>
 
 export declare const dataTagSymbol: unique symbol
-export type DataTag<TType, TValue> = TType & {
+export declare const dataTagErrorSymbol: unique symbol
+export type DataTag<TType, TValue, TError = unknown> = TType & {
```
TkDodo, line 47:
```suggestion
export type DataTag<TType, TValue, TError = DefaultError> = TType & {
```

packages/query-core/src/queryClient.ts
Nick-Lucas: It's not needed for query data because it's a mandatory argument, so if DataTag is there then we take the value inside it.

TaggedError though is an optional argument with default value of `unknown` to avoid breaking changes, so we also have to check it's not the default value and fall-back correctly on TError

Could achieve this with an UnsetMarker symbol instead which would be more obvious?

packages/query-core/src/types.ts
Nick-Lucas: So as per the discussion elsewhere - if DataTag is set without the 3rd argument, and a user sets TError via the QueryClient method, should this default win or should the TError win? I believe the second as the first could be a breaking change

Maybe we use a default of UnsetMarker here?

packages/query-core/src/utils.ts
Nick-Lucas: Done. Weird I could have sworn I'd done that in the last PR, but glad we caught it now

packages/query-core/src/types.ts
TkDodo: aah I get it now. Yeah a separate marker might be more correct because you can set `DefaultError` globally to `unknown` as well and then want to use `TError` on the querfClient method and then it wouldn’t work I think

packages/query-core/src/queryClient.ts
Nick-Lucas: Just pushed the proposed change. It did throw up a few new problems with inferring the data value, the third arg has to be set to `unknown` now during inference.

Also caught a few DataTag usages which I'd missed when passing down TError so those are fixed

packages/query-core/src/types.ts
Nick-Lucas: Cool, already pushed the change so see what you make of it

packages/query-core/src/types.ts
```diff
@@ -43,8 +43,12 @@ export type DefaultError = Register extends {
 export type QueryKey = ReadonlyArray<unknown>
 
 export declare const dataTagSymbol: unique symbol
-export type DataTag<TType, TValue> = TType & {
+export declare const dataTagErrorSymbol: unique symbol
+export declare const unsetMarker: unique symbol
+export type UnsetMarket = typeof unsetMarker
```
TkDodo, line 48:
```suggestion
export type UnsetMarker = typeof unsetMarker
```

😬

Nick-Lucas: 😬

TkDodo (approved):

PR #8375, opened 2024-11-30
Fix type errors when using fully-typed inputs like QueryFilters<Data, Error>, and test all QueryClient methods to detect similar issues

Nick-Lucas:
After adding type parameters to QueryFilters, some methods are proving to be incompatible with fully-typed usage

This PR adds a complete suite of fully-typed calls to all QueryClient methods, and method overloads which accept these inputs. 

The core issue is that while `TFoo & { bar: string }` can be assigned to `TFoo`, if there are any callbacks (ie. predicate) that pass `TFoo` back, typescript will attempts to assign `TFoo` to `TFoo & { bar: string }`, and the default of unknown has the same problem, this is a limitation of subtyping in typescript. To get around this we make the QueryFilter type a generic parameter which can adopt the type of the method argument

image

This is impacting the new tRPC client which the QueryFilter typings were originally added for, in two ways:

image

Firstly the error type needs declaring because tRPC doesn't expect an Error to serialise across the wire, second once declaring this the data type shows up as incompatible because the QueryFilters type is ignored. This PR adds generic type parameters for QueryFilters to improve both situations


packages/query-core/src/__tests__/queryClient.test-d.tsx
```diff
@@ -184,3 +197,151 @@ describe('defaultOptions', () => {
...
+    expectTypeOf(defaultQueryOptions).toEqualTypeOf<
+      DefaultedQueryObserverOptions<TData, TError, TData, TData, QueryKey>
+    >()
+
+    const mutationOptions2 = queryClient.defaultMutationOptions(mutationOptions)
+    expectTypeOf(mutationOptions2).toEqualTypeOf<
+      MutationOptions<TData, TError, void, unknown>
+    >()
+
+    // TODO: should we DataTag MutationKey?
```
Nick-Lucas, line 319: Open question, I don't mind tackling this now if it's agreed to be a useful idea

Nick-Lucas:
Hey @TkDodo I don't want to rush this one in and this changeset is wide-reaching. I experienced quite a few potential breaking changes I had to work around while developing this, so want to be really sure I'm not going to cause problems.

The described issues on 5.62.0 should not impact current usage of RQ, but if anybody wants to use the `QueryFilters<...>` form then they'll quickly find they can't call many of the QueryClient methods with it. This PR addresses that

packages/query-core/src/utils.ts
```diff
@@ -36,9 +35,7 @@ export interface QueryFilters<
   /**
    * Include queries matching this query key
    */
-  queryKey?: unknown extends TQueryFnData
-    ? QueryKey
-    : QueryKey & DataTag<unknown, TQueryFnData, TError>
```
Nick-Lucas, line 41: This was the wrong idea, any fully-typed usage will define the TQueryKey and pass it in, and this approach was proving unreliable when I added richer tests.

packages/query-core/src/__tests__/queryClient.test-d.tsx
```diff
@@ -184,3 +197,307 @@ describe('defaultOptions', () => {
...
+    expectTypeOf(queriesData).toEqualTypeOf<
+      Array<[QueryKey, TData | undefined]>
+    >()
+
+    const queryData3 = queryClient.setQueryData(queryKey, { foo: '' })
+    type SetQueryDataUpdaterArg = Parameters<
+      typeof queryClient.setQueryData<unknown, typeof queryKey>
+    >[1]
+
+    expectTypeOf(null as unknown as SetQueryDataUpdaterArg).toEqualTypeOf<
```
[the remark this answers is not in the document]
Nick-Lucas: Nice <3

Nick-Lucas:
Okay simplified the changeset quite a lot now, instead of adding overloads it's looking a lot more like it looks on other methods which took generics and the new tests & existing builds seem happy

TkDodo (approved):

PR #8383, opened 2024-12-02
fix: no hydration when new promise comes in

juliusmarminge:


juliusmarminge:
really weird, queryFn is getting called and returns the updated value, but the serializer is called only with teh initial count

CleanShot 2024-12-02 at 17 16 59

packages/query-core/src/__tests__/hydration.test.tsx
```diff
@@ -1066,4 +1066,82 @@ describe('dehydration and rehydration', () => {
...
+    expect(serializeDataMock).toHaveBeenCalledWith(0)
+
+    expect(deserializeDataMock).toHaveBeenCalledTimes(1)
+    expect(deserializeDataMock).toHaveBeenCalledWith(0)
+
+    // --- server ---
+    countRef.current++
+    const promise2 = serverQueryClient.prefetchQuery(query)
+
+    const dehydrated2 = dehydrate(serverQueryClient)
```
juliusmarminge, line 1125: assume this is a server fn calling `revalidatePath()` so the page re-renders

TkDodo:
you’re right: we only send the promise when we are in `pending` state. The fix is likely to send the promise when we are in `fetchStatus: 'fetching'` instead 💡

juliusmarminge:
i believe another issue is we're getting `dataUpdatedAt: 0` so it does't rehydrate

CleanShot 2024-12-02 at 19 54 15@2x

juliusmarminge:
feels like we need some other state that would let us know if the promise is fresher then we need to set the queries' promise or something.. this goes deeper than I thought

juliusmarminge:
struggling to figuring out how to properly filter out the queries to put in the hydration queue 
https://github.com/user-attachments/assets/2b6bf540-e735-4eca-a549-e15ab223c1de

TkDodo:
@juliusmarminge have you widened this check?

https://github.com/TanStack/query/blob/8fe90104c6fce9654539e006bd739074e15ae70f/packages/react-query/src/HydrationBoundary.tsx#L74-L76

to something like:

```
const hydrationIsNewer =
  dehydratedQuery.state.dataUpdatedAt >
    existingQuery.state.dataUpdatedAt || dehydratedQuery.promise
```

so that whenever a promise is included, we put it the queue.

juliusmarminge:
yea that causes infinite rerenders

TkDodo:
interesting; let’s debug that together today if you have some time?

juliusmarminge:
> interesting; let’s debug that together today if you have some time?

sure! what time? I can do anytime after 10am gmt+1.

TkDodo:
I pinged you on discord ;)

packages/react-query/src/HydrationBoundary.tsx
```diff
@@ -68,12 +73,33 @@ export const HydrationBoundary = ({
...
+        // )
         if (!existingQuery) {
           newQueries.push(dehydratedQuery)
         } else {
           const hydrationIsNewer =
             dehydratedQuery.state.dataUpdatedAt >
-            existingQuery.state.dataUpdatedAt
+              existingQuery.state.dataUpdatedAt ||
+            // @ts-expect-error
+            dehydratedQuery.promise?.status !== existingQuery.promise?.status
```
juliusmarminge, line 91: this might actually be working

TkDodo:
@juliusmarminge what are we gonna do with this one?

juliusmarminge:
Sorry completely forgot about this 😅

I think I had it sort of working but happy to jump on a call and talk about it still! Hit me up on discord when you have some time to go over it!

packages/react-query/src/HydrationBoundary.tsx
Ephem: What if we prefetch a query without `await` in a page that is static or cached, either fully or partially, and then client navigate to it with newer data in the cache? My guess is that there are situations where we get a promise that will resolve to data that is actually older than the one we have on the client already?

Maybe the entirely correct way to do this would be to inspect the data the query resolves to before determining whether to update the cache with that data? Implementation-wise that's a lot tricker though unfortunately. :(

TkDodo: > Maybe the entirely correct way to do this would be to inspect the data the query resolves to

this is something only users could verify though. Without query meta-data like `dataUpdatedAt`, react-query doesn’t know the age of the data. Are you suggesting we provide a way for users to extract a timestamp (age) from the promise data in some sort of callback option?

Ephem: Is it? If we are using the fetch timestamp from the server to determine this when hydrating _data_, we can surely do that for promises too, just after they resolved? I'm sure it might be a big painful thing to implement, but is there some inherent thing about user/library land that blocks us from doing this in the library using the same logic?

To be clear, if we already have this query in the cache, this might require us to wait for the promise _outside of the cache_ and only put the data in. Or even worse, to support fetching states properly, it might requires us to have some sort of "possibleUpdatePromise" or something that would not commit it's result to the cache if it's older.

TkDodo:
@juliusmarminge we’re now getting a build error for the react15 example:

```
Error occurred prerendering page "/". Read more: https://nextjs.org/docs/messages/prerender-error
Error: redacted
    at /home/workflows/workspace/integrations/react-next-15/.next/server/chunks/750.js:1:41244
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

do you know why that is?

TkDodo:
> what's the unredacted error?

I wish I knew 😂. How can we find out?

stramel:
> > what's the unredacted error?
> 
> I wish I knew 😂. How can we find out?

I was able to find the error by passing `NODE_ENV=development` to the `build` command.

image

```log
❯ NODE_ENV=development pnpm build

> react-next-15@ build /Users/mstramel/GitHub/query/integrations/react-next-15
> next build

 ⚠ You are using a non-standard "NODE_ENV" value in your environment. This creates inconsistencies in the project and is strongly advised against. Read more: https://nextjs.org/docs/messages/non-standard-node-env
   ▲ Next.js 15.1.2

   Creating an optimized production build ...
 ✓ Compiled successfully
   Skipping linting
 ✓ Checking validity of types
 ✓ Collecting page data
Error occurred prerendering page "/_not-found". Read more: https://nextjs.org/docs/messages/prerender-error
TypeError: Cannot read properties of null (reading 'useEffect')
    at t.useEffect (/Users/mstramel/GitHub/query/node_modules/.pnpm/next@15.1.2_@babel+core@7.26.0_babel-plugin-react-compiler@0.0.0_react-dom@19.0.0_react@19.0.0__react@19.0.0_sass@1.81.0/node_modules/next/dist/compiled/next-server/app-page.runtime.prod.js:68:6422)
    at o (/Users/mstramel/GitHub/query/integrations/react-next-15/.next/server/chunks/379.js:1:18168)
    at react-stack-bottom-frame (/Users/mstramel/GitHub/query/node_modules/.pnpm/next@15.1.2_@babel+core@7.26.0_babel-plugin-react-compiler@0.0.0_react-dom@19.0.0_react@19.0.0__react@19.0.0_sass@1.81.0/node_modules/next/dist/compiled/next-server/app-page.runtime.dev.js:25:229305)
    at renderWithHooks (/Users/mstramel/GitHub/query/node_modules/.pnpm/next@15.1.2_@babel+core@7.26.0_babel-plugin-react-compiler@0.0.0_react-dom@19.0.0_react@19.0.0__react@19.0.0_sass@1.81.0/node_modules/next/dist/compiled/next-server/app-page.runtime.dev.js:25:66965)
    at renderElement (/Users/mstramel/GitHub/query/node_modules/.pnpm/next@15.1.2_@babel+core@7.26.0_babel-plugin-react-compiler@0.0.0_react-dom@19.0.0_react@19.0.0__react@19.0.0_sass@1.81.0/node_modules/next/dist/compiled/next-server/app-page.runtime.dev.js:25:81693)
    at retryNode (/Users/mstramel/GitHub/query/node_modules/.pnpm/next@15.1.2_@babel+core@7.26.0_babel-plugin-react-compiler@0.0.0_react-dom@19.0.0_react@19.0.0__react@19.0.0_sass@1.81.0/node_modules/next/dist/compiled/next-server/app-page.runtime.dev.js:25:140140)
    at renderNodeDestructive (/Users/mstramel/GitHub/query/node_modules/.pnpm/next@15.1.2_@babel+core@7.26.0_babel-plugin-react-compiler@0.0.0_react-dom@19.0.0_react@19.0.0__react@19.0.0_sass@1.81.0/node_modules/next/dist/compiled/next-server/app-page.runtime.dev.js:25:136545)
    at finishFunctionComponent (/Users/mstramel/GitHub/query/node_modules/.pnpm/next@15.1.2_@babel+core@7.26.0_babel-plugin-react-compiler@0.0.0_react-dom@19.0.0_react@19.0.0__react@19.0.0_sass@1.81.0/node_modules/next/dist/compiled/next-server/app-page.runtime.dev.js:25:67975)
    at renderElement (/Users/mstramel/GitHub/query/node_modules/.pnpm/next@15.1.2_@babel+core@7.26.0_babel-plugin-react-compiler@0.0.0_react-dom@19.0.0_react@19.0.0__react@19.0.0_sass@1.81.0/node_modules/next/dist/compiled/next-server/app-page.runtime.dev.js:25:83174)
    at renderElement (/Users/mstramel/GitHub/query/node_modules/.pnpm/next@15.1.2_@babel+core@7.26.0_babel-plugin-react-compiler@0.0.0_react-dom@19.0.0_react@19.0.0__react@19.0.0_sass@1.81.0/node_modules/next/dist/compiled/next-server/app-page.runtime.dev.js:25:135151)
Export encountered an error on /_not-found/page: /_not-found, exiting the build.
 ⨯ Static worker exited with code: 1 and signal: null
 ```

**Comment by juliusmarminge** [comment]
`export const dynamic = "force-dynamic"` on app/page.tsx also makes it go away, so it seems like it doesn't identify that the page is doing a dynamic fetch call 🤔

**Comment by juliusmarminge** [comment]
OK I managed to fix this, but it means that Next.js internals are leaking into `query-core`:

CleanShot 2025-01-26 at 11 41 21@2x

File: packages/query-core/src/hydration.ts [hunk 3]
```diff
@@ -82,6 +82,9 @@ function dehydrateQuery(
     queryHash: query.queryHash,
     ...(query.state.status === 'pending' && {
       promise: query.promise?.then(serializeData).catch((error) => {
+        if ('digest' in error && error.digest === 'DYNAMIC_SERVER_USAGE') {
+          return Promise.reject(error)
+        }
```
juliusmarminge, line 87: i'm assuming we don't want these codes to leak into query-core, so should the `defaultQueryOptions` take some `onPrefetchError`/`onPromise...` or similar prop that let's you specify these errors you want to be re-thrown. I'm guessing you might want `notFound()`, `redirect()` errors to have a similar behavior

juliusmarminge:
Hmm seems like the CI test run is running out of memory 👀  https://cloud.nx.app/runs/Qjh3Jk3Wxc/task/%40tanstack%2Freact-query%3Atest%3Alib

packages/query-core/src/hydration.ts
juliusmarminge: Added a `shouldRedactError` option that let's you control whether or not the error should be redacted on dehydration: https://github.com/TanStack/query/pull/8383/commits/75849289ba1bf90bf3a24cf31c0c90d3f083e54b

TkDodo: we’d need to document this too please

juliusmarminge: done

juliusmarminge:
Can you help me get the PR checks passing @TkDodo and/or @Ephem ?

TkDodo:
seems like we’re running into out of memory:

```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

let’s try setting more memory for the test run: [71a4453](https://github.com/TanStack/query/pull/8383/commits/71a4453cc764e1df453796c47ac548e33fdfaf08)

packages/react-query/src/HydrationBoundary.tsx
TkDodo: @juliusmarminge the tests failures are happening because the tests hang indefinitely. Even locally, `vitest` never terminates. If I remove this line of code, everything is back to normal :/

juliusmarminge: How do you run the tests locally? When I run `vitest` from the `query-core` directory all tests pass?

TkDodo: it’s the `react-query` tests that are failing. So go to `packages/react-query` and do `pnpm run test:lib` there.

stramel:
Thank you @juliusmarminge and @TkDodo for all the work to get his fixed! We appreciate the hard work!

PR #8394, opened 2024-12-03
Fix issues with queryOptions factory method types 

Nick-Lucas:
This tackles a few things we've spotted in tRPC:

* Fix issues with queryOptions factory method types not passing through the query key despite taking it
* Fix DataTag creating intersections upon intersections when a data-tagged querykey is passed as the first argument, now if its TType arg is already a datatag it just returns the TType

tRPC build here: https://github.com/trpc/trpc/pull/6290

fixes https://github.com/TanStack/query/pull/8394


packages/query-core/src/types.ts
```diff
@@ -46,10 +46,20 @@ export declare const dataTagSymbol: unique symbol
...
+export type DataTag<
+  TType,
+  TValue,
+  TError = UnsetMarker,
+> = TType extends AnyDataTag
+  ? TType
+  : TType & {
+      [dataTagSymbol]: TValue
+      [dataTagErrorSymbol]: TError
+    }
```
Nick-Lucas, line 62: Fixes this: 

image

becomes this:

image

TkDodo:
> There might be more like this as I'm having trouble with linked builds of react-query 

can you try the PR preview builds?

```
npm i https://pkg.pr.new/@tanstack/react-query@8394
```

packages/react-query/src/__tests__/queryOptions.test-d.tsx
```diff
@@ -241,10 +243,40 @@ describe('queryOptions', () => {
     }
 
     const options = queryOptions({
-      queryKey: ['key'],
+      queryKey: ['key'] as QueryKey,
```
TkDodo, line 246: why is this type assertion necessary? I don’t think this should be needed, because it means we can’t pass what `queryOptions` creates to a function that accepts `UseQueryOptions` 🤔

TkDodo (changes requested):

packages/react-query/src/__tests__/queryOptions.test-d.tsx
Nick-Lucas: Okay got to the bottom of this one

So now that queryOptions is correctly typed (it was wrong before) it knows the query key is a `string[]` - this is broadly a good thing

`UseQueryOptions<number>` though has the query key down as `QueryKey` by default, also broadly a good thing.

But `UseQueryOptions` also contains a callback `QueryFunction` which includes the query key

What happens when typescript structurally tries to treat a supertype as the subtype like it does when comparing types passed in callbacks? It says no "you can't assign supertype to subtype". The solution is:

1. type userland code correctly (see the updated test where I use `UseQueryOptions<number, any, any, Array<string>>` or could use `UseQueryOptions<number, any, any, any>`
2. Change some default query key types in the codebase to `any` and ensure that wherever react-query uses the types internally they get the parameters passed in (should be doing this anyway but this PR fixes a few cases of that being missed)

Nick-Lucas:
> > There might be more like this as I'm having trouble with linked builds of react-query
> 
> can you try the PR preview builds?
> 
> ```
> npm i https://pkg.pr.new/@tanstack/react-query@8394
> ```

Thanks, that's a life-safer, the content of this PR has tRPC's codebase passing now (https://github.com/trpc/trpc/pull/6290) albeit with one fix for correctness related to the DataTag change: `packages/react-query/src/shared/types.ts`

packages/react-query/src/__tests__/queryOptions.test-d.tsx
```diff
@@ -236,7 +237,9 @@ describe('queryOptions', () => {
   })
 
   it('should be passable to UseQueryOptions', () => {
-    function somethingWithQueryOptions(options: UseQueryOptions<number>) {
+    function somethingWithQueryOptions(
+      options: UseQueryOptions<number, any, any, Array<string>>,
```
Nick-Lucas, line 241: Related discussion: https://github.com/TanStack/query/pull/8394#discussion_r1870260884

packages/react-query/src/__tests__/queryOptions.test-d.tsx
TkDodo: yeah I think 2 is better. I think I’ve stumbled upon this before, too. can you do that?

Nick-Lucas: I can, though I am conscious of how this could inadvertently de-type folk's codebases because many/most types are able to be used both internally and in userland

Will have a look into it it

Nick-Lucas: Okay so I've made this switch on all interfaces and types, but not on any functions or classes and it's not looking like the right way

Inference in DataTags breaks, because of course `any extends DataTag<any, infer Data, any> ? true : false` is always `true` and there are other similar classes of issues. Test suite here is up to date and number of errors not that huge but I don't see a way around them right now

Nick-Lucas: Just to make a case, we're in a position already where some subtle breaking changes went out because of my previous PRs as I missed them. 

This PR with option 1 above, moves those breakages to somewhere that userland can trivially fix by typing interfaces correctly, and should not affect the vast majority of codebases anyway, whereas tRPC can't actually use 5.62.2 right now (for reasons I'd need to refresh my brain on but they're the reason I came back with this PR)

Nick-Lucas: I can't say it's better, took quite a few changes to fix query-core and now every package has problems - pushed at this point so you can at least review the classes of errors. Seems a much bigger impact for the public surface area than my original idea on this PR

packages/angular-query-experimental/src/infinite-query-options.ts
```diff
@@ -13,7 +13,7 @@ export type UndefinedInitialDataInfiniteOptions<
   TQueryFnData,
   TError = DefaultError,
   TData = InfiniteData<TQueryFnData>,
-  TQueryKey extends QueryKey = QueryKey,
+  TQueryKey extends QueryKey = Array<any>,
```
TkDodo, line 16: my idea was to keep this to:

```
```

and then just change what `QueryKey` is:

```diff
- export type QueryKey = ReadonlyArray<unknown>
+ export type QueryKey = Array<any>
```

Nick-Lucas: Right! Sorry my bad, you did write that. Have tried it just to double check but no dice. The initial error about it being readonly is part of the problem, but the full problem is the flipped order of assigning sub-type to super-type caused by the type being referenced in a callback. 

image

The fix for this is always to provide your types correctly at the top layer so both directions of structural comparison can pass, and Tanstack Query has been masking this problem through query keys not always being respected on Client methods. This masking became really obvious with the new tRPC Client though because we can't type-anything without the DataTag flowing through

Nick-Lucas: Wait I made another mistake there keeping unknown, will come back

Nick-Lucas: Okay yep same answer as above:

`Array<unknown>` -> we get assignment error in callbacks

`Array<any>` -> assignment errors fixed, but we instead of get errors when assigning const/inline keys to query keys (can't assign readonly to mutable)

`ReadonlyArray<any>` -> we get assignment error in callbacks because they're assigning a readonly to a mutable (but this time in reverse)

Nick-Lucas: Only way to win here is to correctly type interfaces, which anybody using the standard methods in Tanstack Query will get for free, but anybody passing around query options would need to update their argument types as per the one failing test in this PR

packages/react-query/src/__tests__/queryOptions.test-d.tsx
```diff
@@ -236,7 +237,9 @@ describe('queryOptions', () => {
   })
 
   it('should be passable to UseQueryOptions', () => {
-    function somethingWithQueryOptions(options: UseQueryOptions<number>) {
+    function somethingWithQueryOptions<TQueryOpts extends AnyUseQueryOptions>(
+      options: TQueryOpts,
+    ) {
```
Nick-Lucas, line 242: Hey @TkDodo so I think the other two approaches proved to be dead ends, and I'm also unsure about making major internal changes like those as they risk creating a new wave of breaking issues.

What I propose is we export types like `AnyUseQueryOptions`, and that way consumers wanting to create flexible re-usable methods can use a more correct form of generics like above (don't necessarily need the generics but for anything non-contrived it's probably useful). This is how tRPC and a few other projects solve this issue and it's rock solid

packages/angular-query-experimental/src/infinite-query-options.ts
Nick-Lucas: Pushed an alternate approach, it's back to where I started but with a clearer user story inspired by tRPC: https://github.com/TanStack/query/pull/8394#discussion_r1881067882

TkDodo (approved):

PR #8398, opened 2024-12-04
fix(react-query): ensureSuspenseTimers should ALWAYS set staleTime to 1000 when it is below 1000.

minseong0324:
[issue link](https://github.com/TanStack/query/discussions/8366)

Changes
- Modified ensureSuspenseTimers to enforce a minimum staleTime of 1000ms in Suspense mode
- Added support for both direct values and function-based staleTime configurations
- Maintained the existing behavior for values greater than 1000ms


TkDodo (commented):
can you please add a test for this scenario?

minseong0324:
changes:

1. Fixed logic in ensureSuspenseTimers to properly handle all cases where staleTime needs to be enforced (values below 1000ms, undefined values, and functions returning values below 1000ms)
2. Added comprehensive test coverage for these scenarios to ensure the behavior works as intended
3. Removed the "should refetch when re-mounting" test as it no longer aligns with the new implementation

packages/react-query/src/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -327,61 +327,6 @@ describe('useSuspenseQuery', () => {
     consoleMock.mockRestore()
   })
 
-  it('should refetch when re-mounting', async () => {
```
TkDodo, line 330: why was this test removed?

packages/react-query/src/__tests__/suspense.test.tsx
```diff
@@ -0,0 +1,231 @@
...
+    rendered.rerender(
+      <Suspense fallback="loading">
+        <Component />
+      </Suspense>,
+    )
+
+    // Wait longer than 1000ms but less than returned staleTime
+    await act(async () => {
+      await sleep(2000)
+    })
```
TkDodo, line 226: this will make for a very long test. `fakeTimers` from vitest are probably the way to go here. We have some examples already, e.g. here:

https://github.com/TanStack/query/blob/1eebfc9d3355c411b16146cc8df76c18e63d59ad/packages/react-query/src/__tests__/useSuspenseQueries.test.tsx#L37

minseong0324: image

I improved the test code through a lot of thought to improve the test time, and the useFakeTimers you gave me were especially helpful. Thanks.

46057ca7076a40c7019346f7d0e4720d00f078f3

packages/react-query/src/__tests__/useSuspenseQuery.test.tsx
minseong0324: Since we now enforce a minimum staleTime of 1000ms in Suspense mode for better performance and consistency, the "should refetch when re-mounting" test is no longer valid. This test expected immediate refetching on remount with staleTime: 0, but this behavior would conflict with our implementation that ensures queries remain fresh for at least 1000ms when using Suspense.

The test was designed to verify a behavior that we've intentionally changed to prevent unnecessary refetches and provide a more stable user experience in Suspense mode. Removing this test ensures our test suite accurately reflects our current design decision of enforcing minimum staleTime.

TkDodo (approved):

PR #8408, opened 2024-12-06
chore: examples to react19

TkDodo:


TkDodo:
@KATT I’ve just seen that all our `useQuery().promise` tests fail with the stable version of React19. Not sure yet why that is 😔

PR #8414, opened 2024-12-07
example usage of `@testing-library/react-render-stream`

phryneas:
I was just talking tests with @tkdodo and wanted to do a quick example on what a test with RRSTL would look like

### DIFF
--- a/packages/react-query/src/__tests__/useSuspenseQuery.test.tsx
+++ b/packages/react-query/src/__tests__/useSuspenseQuery.test.tsx
@@ -17,7 +17,11 @@ import type {
   UseSuspenseInfiniteQueryResult,
   UseSuspenseQueryResult,
 } from '..'
-import { createRenderStream, disableActEnvironment, cleanup } from '@testing-library/react-render-stream'
+import {
+  createRenderStream,
+  disableActEnvironment,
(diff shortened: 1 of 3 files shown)


packages/react-query/src/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -52,22 +54,42 @@ describe('useSuspenseQuery', () => {
...
+    {
+      const { snapshot, withinDOM } = await renderStream.takeRender()
+      expect(snapshot).toMatchObject({ data: 1, status: 'success' })
+      withinDOM().getByText("data: 1") // I'd probably skip this assertion
+    }
     fireEvent.click(rendered.getByLabelText('toggle'))
-
-    await waitFor(() => rendered.getByText('data: 2'))
-
-    expect(renders).toBe(4)
```
phryneas, line 67: no need to explicitly assert on four renders anymore - we called `takeRender` four times, and we assert that no more renders will happen at the end of the test

phryneas:
the test failures here seem to be unrelated and from another test? 😅

TkDodo:
thanks for the pointers - I incorporated this into some other tests 👍

PR #8419, opened 2024-12-09
fix(angular-query): fix error NG0600 on Angular v16-v18

arnoud-dv:
Effect `onCleanup` sets a signal which throws an error on Angular < v19 unless `allowSignalWrites` is set.

### DIFF
--- a/packages/angular-query-experimental/src/create-base-query.ts
+++ b/packages/angular-query-experimental/src/create-base-query.ts
@@ -92,10 +92,12 @@ export function createBaseQuery<
         })
       })
       onCleanup(() => {
-        resultFromSubscriberSignal.set(null)
+        ngZone.run(() => resultFromSubscriberSignal.set(null))
       })
     },
     {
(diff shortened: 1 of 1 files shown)


packages/angular-query-experimental/src/create-base-query.ts
```diff
@@ -92,10 +92,12 @@ export function createBaseQuery<
         })
       })
       onCleanup(() => {
-        resultFromSubscriberSignal.set(null)
+        ngZone.run(() => resultFromSubscriberSignal.set(null))
       })
     },
     {
+      // Set allowSignalWrites to support Angular < v19
+      allowSignalWrites: true,
```
hakimio, line 100: @arnoud-dv 
This will log a warning on Angular `19`. You should check Angular version before setting this config.

arnoud-dv: Didn't expect a warning message for a noop 🤦‍♂️

Thanks for letting me know.

Issue #8431, opened 2024-12-12
"Duplicate Queries found" warning, incorrectly logging when creating a QueriesObserver instance


nick-avi-say:
### Describe the bug

The duplicate keys check occurs prior to default options being set on the queries which generates the queryHash.

The below contains 2 unique query key's. However the duplicate keys warning logs since these queries are checked as-is and they do not yet have a queryHash.

`const observer = new QueriesObserver(queryClient, [
      {
        queryKey: [
          'Hippo',
          'Layout',
          'detail',
          'id1',
        ],
      },
      {
        queryKey: [
          'Hippo',
          'Layout',
          'detail',
          'id2',
        ],
      },
    ])`

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/55fr36

### Steps to reproduce

1. Create a QueryObserver instance with 2 queries with 2 unique keys. Do not provide a custom queryHash.
2. View the warning in your console

(CTA set up to do this for you in the codesandbox)

### Expected behavior

We expect unique keys to not throw a duplicate key console warning.

### How often does this bug happen?

None

### Platform

All

### Tanstack Query adapter

None

### TanStack Query version

v5.62.7

nick-avi-say:
Added a PR here
https://github.com/TanStack/query/pull/8432

PR #8432, opened 2024-12-12
fix(QueriesObserver): include default options on duplicate query, queryHash check

nick-avi-say:
This commit fixes the issue logged here 
https://github.com/TanStack/query/issues/8431

The duplicate keys check occurs prior to default options being set on the queries which generates the queryHash.

The below contains 2 unique query key's. However the duplicate keys warning logs since these queries are checked as-is and they do not yet have a queryHash.

`const observer = new QueriesObserver(queryClient, [ { queryKey: [ 'Hippo', 'Layout', 'detail', 'id1', ], }, { queryKey: [ 'Hippo', 'Layout', 'detail', 'id2', ], }, ])`

### DIFF
--- a/packages/query-core/src/queriesObserver.ts
+++ b/packages/query-core/src/queriesObserver.ts
@@ -92,7 +92,9 @@ export class QueriesObserver<
     this.#options = options
 
     if (process.env.NODE_ENV !== 'production') {
-      const queryHashes = queries.map((query) => query.queryHash)
+      const queryHashes = queries.map(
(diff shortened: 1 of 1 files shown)


TkDodo (changes requested):
thanks. can you please add a test that fails on `main` but works with your fix so that we don’t introduce a regression later.

TkDodo (approved):
ah, I missed that this was behind the env check. It’s fine without a test 👍

Issue #8384, opened 2024-12-02
experimental_promise: A component was suspended by an uncached promise


TkDodo:
### Describe the bug

using the new experimental promise with `use` in React 19 leads to an error when multiple queries are used in the same transition. The error is:

```
A component was suspended by an uncached promise. Creating promises inside a Client Component or hook is not yet supported, except via a Suspense-compatible library or framework. 
    at Example2 (https://7g9lfw.csb.app/src/App.js:51:5)
    at Suspense
    at div
    at SuspenseBoundary (https://7g9lfw.csb.app/src/App.js:73:49)
    at QueryClientProvider (https://7g9lfw.csb.app/node_modules/@tanstack/react-query/build/modern/QueryClientProvider.cjs:55:3)
    at div
    at Example
```

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/suspense-transitions-7g9lfw

### Steps to reproduce

- click `decrease`
- look at the error in the devtools
- you can also see that 2.1 shows the right value, while 2.2 shows the wrong (outdated) value

everything works fine if we remove `startTransition`

### Expected behavior

Both components (2.1 and 2.2) should show the correct state values

### How often does this bug happen?

Every time

### Platform

React 19RC, ReactQuery 5.62.0

### Tanstack Query adapter

react-query

### TanStack Query version

5.62.0

### TypeScript version

-

### Additional context

Initially, I wanted to test if `placeholderData: keepPreviousData` works inside transitions. And it does - but I found those weird errors, and noticed they have nothing to do with `placeholderData`, but rather always happen if use the same transition value in multiple queries. It doesn’t matter if they are in the same suspense boundary or not.

KATT:
I wish it was documented what a _"a Suspense-compatible library or framework."_ means, it feels like we're doing something wrong but kinda fumbling in the dark.

I will try to mess around with this this week 🥲

tom-sherman:
I can't reproduce the error consistently (proof in the video 😅)

https://github.com/user-attachments/assets/8d643cea-ae33-4a59-9a24-e6ffdadf255f

TkDodo:
@tom-sherman yeah I just got it myself only after like 4 fetches :/

![Image](https://github.com/user-attachments/assets/7127a420-2361-4f3e-b8b5-98987eab0fd7)

KATT:
Life and other commitments got in the way of me wrestling with this today, but I will try to have another stab next Friday

[...]

Issue #7012, opened 2024-03-03
React component "tearing"/incorrectly stuck loading with `staleTime`/`enabled`/`queryKey` changes


braeden:
## Some components using the same `useQuery` hook get *consistently* stuck in (`fetching`/`loading`) state when they should be rendering a result (`idle`/`success`).

The example here is a bit contrived to get a minimal repro, but this is a legitimate usecase/ bug. 

### Your minimal, reproducible example

https://codesandbox.io/p/devbox/weathered-framework-ktk4kh?file=%2Fsrc%2Findex.jsx%3A40%2C21

### Apparent pre-requisites from the narrowing I was able to do:
- Non-zero `staleTime`
- Changes to `enabled` + `queryKey` values coinciding with a state change that happens _after_ the initial render.
   -  Either via a microtask:  `Promise.resolve()`/`queueMicrotask`
   -  Or macrotask: `setTimeout`.
- Several react components must render (consuming the same data)

--------

Also interesting to note: the "last" component to render gets stuck when queueing the state change a `microtask` and the "second" one always get's stuck when queuing the change as a new macrotask. (Obviously which element gets stuck is likely some strange cross interaction with React's internal queuing of component updates/commits). 

### Expected behavior

I expect all rendered components to enter the "loaded" state. 

### How often does this bug happen?

Every time. (can be forced into a race condition by lowering `staleTime`)

### Screenshots or Videos

https://github.com/TanStack/query/assets/5288805/b2a57fa2-0cef-4fa8-a6c4-f25b6aa62c5b

### Platform

`react-query: ^4.33.0`
`react: ^16.14.0"`

Chrome: `122.0.6261.94 (Official Build) (arm64)` (MacOS)

### Tanstack Query adapter

react-query

### TanStack Query version

v4.33.0

TkDodo:
your sandbox works fine after moving to react18 and TanStack Query v5:

https://codesandbox.io/p/devbox/spring-star-fgmywp

I'm sorry but I don't think I'm going to look into this combination (v4 + React16). You can try if upgrading either solves the problem, but updating both definitely does.

PR #8434, opened 2024-12-13
fix(react-query): move away of uSES

KATT:


packages/react-query/src/useIsFetching.ts
```diff
@@ -12,13 +12,27 @@ export function useIsFetching(
...
-    () => client.isFetching(filters),
-    () => client.isFetching(filters),
+  const [number, setNumber] = React.useState<number>(() =>
+    client.isFetching(filters),
   )
+
+  const filtersRef = React.useRef(filters)
+  React.useEffect(() => {
+    filtersRef.current = filters
+  }, [filters])
```
KATT, line 22: `useMutationState()` was memoizing filters, so I'm guessing we should be doing it here too

packages/query-core/src/queryObserver.ts
```diff
@@ -609,6 +610,11 @@ export class QueryObserver<
        * Create a new thenable and result promise when the results have changed
        */
       const recreateThenable = () => {
+        console.log(
+          '-----------recreateThenable()',
+          'queryHash:',
+          query.queryHash,
+        )
```
KATT, line 617:
```suggestion
```

packages/query-core/src/queryObserver.ts
```diff
@@ -698,11 +704,14 @@ export class QueryObserver<
...
-      return Object.keys(this.#currentResult).some((key) => {
+      return [...includedProps].some((key) => {
         const typedKey = key as keyof QueryObserverResult
         const changed = this.#currentResult[typedKey] !== prevResult[typedKey]
-
-        return changed && includedProps.has(typedKey)
+        console.log(`changed key "${typedKey}"?`, changed, {
+          prev: this.#currentResult[typedKey],
+          next: prevResult[typedKey],
+        })
```
KATT, line 713:
```suggestion
```

packages/query-core/src/queryObserver.ts
```diff
@@ -698,11 +704,14 @@ export class QueryObserver<
         includedProps.add('error')
       }
 
-      return Object.keys(this.#currentResult).some((key) => {
+      return [...includedProps].some((key) => {
```
KATT, line 707: Tiny optimization that also made debugging a bit easier

packages/react-query/src/__tests__/useQuery.test.tsx
```diff
@@ -2485,38 +2486,6 @@ describe('useQuery', () => {
     expect(queryCache.find({ queryKey: key })!.options.retryDelay).toBe(20)
   })
 
-  it('should batch re-renders', async () => {
```
KATT, line 2488: This test doesn't work anymore when we're force rendering

I'm guess we just have to trust the concurrency in React to do this in an optimal way?

packages/react-query/src/useBaseQuery.ts
```diff
@@ -82,26 +82,27 @@ export function useBaseQuery<
       ),
   )
 
+  const [_, setForceUpdate] = React.useState(0)
+
   const result = observer.getOptimisticResult(defaultedOptions)
```
KATT, line 87: I wanted to use a `useState()` here but it made the blast radius really big and broke `useSuspense()`'s test etc

packages/query-core/src/queryObserver.ts
TkDodo: this is a great improvement 🎉

packages/react-query/src/useBaseQuery.ts
```diff
@@ -82,26 +82,27 @@ export function useBaseQuery<
...
-    () => observer.getCurrentResult(),
-  )
+  React.useEffect(() => {
+    if (isRestoring) {
+      return
+    }
+
+    const unsubscribe = observer.subscribe(
+      notifyManager.batchCalls(() => {
+        setForceUpdate((n) => n + 1)
```
TkDodo, line 96: should we go for a new object here?

```
setForceUpdate({})
```

pros:
- it's a bit shorter
- it won't create huge numbers

cons:
- creates a new object every time ??

packages/react-query/src/useIsFetching.ts
TkDodo: this could be done without dependency array - updating the ref after every render is fine.

packages/react-query/src/useMutationState.ts
```diff
@@ -47,31 +47,26 @@ export function useMutationState<TResult = MutationState>(
...
-    ),
-    () => result.current,
-    () => result.current,
-  )!
+  const [result, setResult] = React.useState<Array<TResult>>(() =>
+    getResult(mutationCache, options),
+  )
+
+  React.useEffect(() => {
+    const sub = mutationCache.subscribe(() => {
```
TkDodo, line 59: one thing that's missing here is calling the setter directly in useEffect, like we do in  `useBaseQuery` and `useIsFetching` to make sure we do not miss any query updates between creating the observer and subscribing to it.

I know it's not really possible because the result isn't necessarily a primitive.

Do you think it would be possible to keep `useSyncExternalStore` for `useMutationState` and `useIsFetching` ?

TkDodo (commented):
Let's please make sure to test this with the reproduction from #8384 to make sure it fixes our issue, and also with the reproduction from #7012 to make sure it doesn't introduce tearing ❤️

packages/react-query/src/__tests__/transition.test.tsx
KATT, line 1: cc @TkDodo I pulled in your test from #8444 and it seemingly works?

I had to fix it in 38b88645e85bde8e2ccc42eb3d57997e9eb8deaf but I think that's fine or?

packages/react-query/src/useMutationState.ts
KATT: I'm not sure if we should keep it - it seems to me like something that should be deprecated in React if it by design causes issues as soon as you use `startTransition`?

packages/react-query/src/useBaseQuery.ts
KATT: I don't see a downside with `setForceUpdate({})` - updated the code!

packages/react-query/src/__tests__/transition.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+import * as React from 'react'
+import {
+  createRenderStream,
+  useTrackRenders,
+} from '@testing-library/react-render-stream'
+import { QueryClientProvider, useQuery } from '..'
+import { QueryCache } from '../index'
+import { createQueryClient, queryKey, sleep } from './utils'
+
+describe('react transitions', () => {
```
KATT, line 11: cc @TkDodo I pulled in your test from #8444 and it seemingly works?

I had to fix it in 38b88645e85bde8e2ccc42eb3d57997e9eb8deaf but I think that's fine or?

packages/react-query/src/useBaseQuery.ts
KATT: If we get more errors, I'd bet this could be a culprit as it's reading results synchronously from the observer

We should maybe have a `const [result, setResult]` rather than a force update, but I'd kinda want to do this as a follow-up if we do it as it changes all `useSuspense()`-test because it'll take an extra render before the component is suspended

packages/react-query/src/__tests__/transition.test.tsx
```diff
@@ -0,0 +1,288 @@
...
+
+const queryCache = new QueryCache()
+const queryClient = createQueryClient({
+  queryCache,
+})
+
+beforeAll(() => {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  global.IS_REACT_ACT_ENVIRONMENT = true
```
KATT, line 17: did what Rick suggested here https://discord.com/channels/514829729862516747/1251229763838677052/1318328930658418718

packages/react-query/src/useMutationState.ts
KATT: Or maybe it does make sense to use uSES here, idk....

KATT: Reveted them too, 76bfacf for future reference

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "editor.formatOnSave": true
+}
```
KATT, line 4: Feel free to remove this, but it helps me as a contributor - I have a bunch of settings in trpc https://github.com/trpc/trpc/tree/next/.vscode

packages/react-query/src/useBaseQuery.ts
```diff
@@ -82,33 +82,43 @@ export function useBaseQuery<
...
+    }
     // Do not notify on updates because of changes in the options because
     // these changes should already be reflected in the optimistic result.
     observer.setOptions(defaultedOptions, { listeners: false })
   }, [defaultedOptions, observer])
 
+  if (defaultedOptions.experimental_prefetchInRender) {
+    // Do not notify on updates because of changes in the options because
+    // these changes should already be reflected in the optimistic result.
+    observer.setOptions(defaultedOptions, { listeners: false })
```
KATT, line 119: Setting the options straight away for `experimental_prefetchInRender` seems to fix it.

I think the issue I had was some race condition...

I still have no reliable reproduction though

MrFlashAccount:
> make sure it doesn't introduce tearing

Unfortunately there's no way to avoid tearing without using uSES (except moving everything onto the react state). But this is fine, it's an _acceptable_ drawback in some situations.

Side note: maybe it worth introducing a separate hook for that? or a config option for those who doesn't want to have _unavoidable_ tearing?

KATT:
The only "blocker" right now is that we haven't yet been able to do a reliable reproduction to the original issue, I wasn't able to reproduce the issue within vite myself

nstadigs:
Dropping some links here in case they might help
- [Concurrent stores focus announcement in react blog](https://react.dev/blog/2025/04/23/react-labs-view-transitions-activity-and-more#concurrent-stores) 
- [Ponyfill approach by thejustinwalsh that might work today for react-query](https://github.com/thejustinwalsh/react-concurrent-store)
- [Also see alternative approach mentioned in issues of said repo](https://github.com/thejustinwalsh/react-concurrent-store/issues/3)

TkDodo:
I think with all I know as of today, I don’t think it’s likely that we will be going back to useEffect and useState. Instead, exploring concurrent stores sounds more intriguing. If there’s a working ponyfill, I would be very interested in seeing a PR that tries to integrate this in react-query

nstadigs:
https://github.com/facebook/react/pull/33215#issuecomment-3291286262

nstadigs:
> I think with all I know as of today, I don’t think it’s likely that we will be going back to useEffect and useState. Instead, exploring concurrent stores sounds more intriguing. If there’s a working ponyfill, I would be very interested in seeing a PR that tries to integrate this in react-query

The ponyfill in itself is actually quite small so maybe it can been seen as more of a reference to do something similar, but more integrated, within react-query instead of importing it? (it still involves useState and useEffect ;) )

nstadigs:
Sorry for spamming, but I'm following the discussions around this closely for my own library. I figure I should share my findings with you, since you need to solve this as well.
See bsky thread up until https://bsky.app/profile/ricky.fm/post/3lyyoh3ngfc2g

TkDodo:
we’re gonna wait for concurrent stores.

Issue #5996, opened 2023-09-12
Sometimes vue-query doesn't respect `enabled`


Mini-ghost:
### Describe the bug

Sometimes vue-query doesn't respect `enabled`, It happens when `enabled` changes from `true` to `false`

### Your minimal, reproducible example

https://stackblitz.com/edit/vitejs-vite-q1vdvf?file=src%2Fpages%2FPost.vue

### Steps to reproduce

1. Go to reproducible
2. Click the link with content `1`, and `queryFn` will be executed.
3. Click the link with content `20`, and `queryFn` will be executed.

According to the `enabled` condition, when `postId` is greater than `10`, it will be `false`

```ts
const enabled = computed(() => {
  return !!postId.value.id && postId.value.id < 10;
});
```

### Expected behavior

`queryFn` should not be called when `enabled` is `false`

### How often does this bug happen?

Sometimes

### Platform

```
System:
    OS: macOS 11.6
    CPU: (8) arm64 Apple M1
    Memory: 90.47 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 18.16.0 - ~/.nvm/versions/node/v18.16.0/bin/node
    npm: 9.5.1 - ~/.nvm/versions/node/v18.16.0/bin/npm
  Browsers:
    Firefox: 116.0.2
    Safari: 14.1.2
    Safari Technology Preview: 15.4
  npmPackages:
    vue: ^3.3.4 => 3.3.4 
```

### Tanstack Query adapter

None

### TanStack Query version

v4.34.3

### TypeScript version

v5.2.2

### Additional context

I guess it is because `{ flush:'sync' }` caused the current issue. The following are relevant references.

- https://github.com/TanStack/query/pull/5929
- https://github.com/vuejs/core/issues/9149
- https://github.com/vuejs/core/pull/5912

PR #8443, opened 2024-12-16
fix(vue-query): avoid use `sync` for `watch` that cause unnecessary requests

Mini-ghost:
### Linked issue

- #7942
- #5910
-  #5996

### Description

In PR #7942, attempting to change `watch`'s `{ flush: 'sync' }` to `{ flush: 'post' }` or removing it might lead to the same issues as #5910 and #5996.  

This PR removes `{ flush: 'sync' }` and ensures proper handling of `refresh` to prevent the issue in #5910 where `queryKey` retrieves stale data.  

If there are any considerations I may have missed, please kindly let me know. Thank you!


RomainLanz:
It seems this change might have broken `useQueries`.

I'll look into it further, but for now I've noticed that when a query is marked as stale and used in `useQueries`, it doesn’t get refetched.

RomainLanz:
Never mind, the issue was on my side. Here's a "reproduction" that can be used later to ensure everything keeps working. 👍🏻

https://codesandbox.io/p/sandbox/issue-tanstack-query-queries-y6rlg6

PR #8451, opened 2024-12-17
 perf(query-core): Improve mutationCache implementation performance

gnoff:
Closes https://github.com/TanStack/query/issues/8277

This work is stacked on #8450 

In the prior change I modeled unscoped mutations as a separate data store in a way that would speed up methods operating on the internal state of the cache when using unscoped mutations. In this change I further refatored the internal implementation to make removals faster (O(1) generally) and to make the aggregate APIs like `getAll` and `find` faster by avoiding multiple concatenations and flattening.

## Implementation Details

Scopes are more like scheduling metadata than an intrinsic structure to the set of mutations. In this change I updated the storage format of mutations so that they can be cleared more efficiently. Using a Set allows us to have O(1) removals. We can also construct the Array for getAll() more efficiently by not having to flatten the inner scope arrays. Additionally the scope array is now mutated both on add and remove to avoid allocating extra arrays and doing additional copies.

I rewrote clear to not use this.remove which avoids many intermediate operations against the underlyign data structures.

I considered using an Array again for the primary storage format and this may actually be better if the expected number of mutations isn't excessively large but since I don't know enough about mutation usage I stuck with a Set for the better deletion performance at larger mutation size.


packages/query-core/src/mutationCache.ts
```diff
@@ -109,59 +111,82 @@ export class MutationCache extends Subscribable<MutationCacheListener> {
...
-    const scope = scopeFor(mutation)
-    const mutations = this.#mutations.get(scope) ?? []
-    mutations.push(mutation)
-    this.#mutations.set(scope, mutations)
+    this.#mutations.add(mutation);
+    const scope = scopeFor(mutation);
+    if (typeof scope === 'string') {
+      const scopedMutations = this.#scopes.get(scope);
+      if (scopedMutations) {
+        scopedMutations.push(mutation);;
```
bai, line 119:
```suggestion
        scopedMutations.push(mutation);
```

TkDodo:
I don’t know why our CI doesn’t run on this PR :/ I’ll try to open a new one

TkDodo:
> I don’t know why our CI doesn’t run on this PR :/ I’ll try to open a new one

found out why that happened. That’s what you get for trying out beta features I guess 😂 

- https://github.com/orgs/community/discussions/143787#discussioncomment-11727734

PR #8478, opened 2024-12-27
handle react-native resolver

hosseinmd:
Specially in react-native repack, we need to resolve from source code

### DIFF
--- a/packages/react-query/package.json
+++ b/packages/react-query/package.json
@@ -37,6 +37,7 @@
   "types": "build/legacy/index.d.ts",
   "main": "build/legacy/index.cjs",
   "module": "build/legacy/index.js",
+  "react-native": "src/index.ts",
   "exports": {
     ".": {
       "import": {


TkDodo:
I have never seen a top-level react-native field in package.json, and a quick google search also didn't yield any good results. Can you please elaborate what this does, which bundlers interpret this field and why we need it? Thanks.

hosseinmd:
Look at this libraries:
React-navigation
React-freeze
react-native-video
react-native-gesture-handler
cross-fetch
lottie-react-native

And so many libraries which wanna work with platform specific files ('.native.ts' or '.android.ts') should be considered this.
Without specifying source for react-native loaders for handling tree shaking will load module and then import react-dom in react-native

hosseinmd:
This is repack mainFields priority
https://github.com/callstack/repack/blob/6a210ed2559bbef2468c27f517d2eaaba2510a87/packages/repack/src/utils/getResolveOptions.ts#L111

Scutua (approved):

Issue #8042, opened 2024-09-11
Angular injectMutation: `onSuccess` loses type when `queryClient` is passed to `injectMutation`


7schmiede:
### Describe the bug

When the `queryClient` is passed in the `injectMutation` method, the return type of `mutationFn` is lost in the `onSuccess` method.

### Your minimal, reproducible example

https://stackblitz.com/edit/stackblitz-starters-3zsk2e?file=test.ts

### Steps to reproduce

```ts
import { injectMutation } from "@tanstack/angular-query-experimental";

// without queryClient 
export const foo = injectMutation(() => ({
    mutationFn: () => Promise.resolve({ foo: true }),
    onSuccess: (result) => {
        result.foo; // result: { foo: boolean; }
    }
}));

// with queryClient
export const bar = injectMutation((queryClient) => ({
    mutationFn: () => Promise.resolve({ bar: true }),
    onSuccess: (result) => {
        result.bar; // result: unknown
    }
}));
```

### Expected behavior

Even if the `queryClient` is passed, the return type of the `mutationFn` should remain in the `onSuccess` method.

### How often does this bug happen?

Every time

### Platform

- OS: Windows 11
- Browser: Edge Chromium
- Version: 5.55.4

### Tanstack Query adapter

angular-query

### TanStack Query version

v5.55.4

mevans:
Yeah this is a really frustrating issue. I either have to choose between just injecting the query client at the top level, or manually typing all parameters in the callback functions

rklfss:
Same here. Figured out, specifying the queryClient type does the trick for the moment. At least better than every method arguments.
```ts
injectMutation((client: QueryClient) => ({
}));
```

ThiloAschebrock:
One could fix this by wrapping `TData` with `NoInfer<TData>` in the `MutationOptions` type of the core package at https://github.com/TanStack/query/blob/main/packages/query-core/src/types.ts#L1005, but that would be a breaking change for all adapters.

arnoud-dv:
This happens with the `select` function in `injectQuery too`

Reproduction as unit tests:

### packages/angular-query-experimental/src/__tests__/inject-query.test-d.ts

```ts
    test('should be possible to define a different TData than TQueryFnData using select with queryOptions spread into useQuery and passing queryClient', () => {
      const options = queryOptions({
        queryKey: ['key'],
        queryFn: () => Promise.resolve(1),
      })

      const query = injectQuery((client) => ({
        ...options,
[...]

PR #8486, opened 2024-12-30
Update invalidations-from-mutations.md

ayush-seth:
Updated the guide to show both ways of accessing `queryClient` inside a mutation, the comment earlier was wrong but now makes sense with the updated example.

There was also an irrelevant component code which I removed

### DIFF
--- a/docs/framework/angular/guides/invalidations-from-mutations.md
+++ b/docs/framework/angular/guides/invalidations-from-mutations.md
@@ -5,19 +5,6 @@ ref: docs/framework/react/guides/invalidations-from-mutations.md
 replace: { 'useMutation': 'injectMutation', 'hook': 'function' }
 ---
 
-[//]: # 'Example'
-
-```ts
-class TodoItemComponent {
-  mutation = injectMutation(() => ({
-    mutationFn: postTodo,
-  }))
-}
-```
-
-[//]: # 'Example'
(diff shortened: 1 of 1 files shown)


arnoud-dv:
> Updated the guide to show both ways of accessing queryClient inside a mutation

I [removed](https://github.com/TanStack/query/pull/8307) the convenience queryClient parameter as it was [causing typing issues](https://github.com/TanStack/query/issues/8042). I also removed mentions of this parameter from the documentation, but forgot to remove the comment. So that's now indeed wrong.

> There was also an irrelevant component code which I removed

In general I like the Angular examples to be in components. In almost all cases developers will use `injectQuery` and `injectMutation` in either a component or a service which are both classes and very similar so examples using components are helpful.

If you could change this PR to just remove the wrong comment it can be merged. Thanks!

ayush-seth:
@arnoud-dv I have made the requested changes, however the "component code" I removed was `TodoItemComponent` which seemed unreleated to the example completely, the invalidation example is inside `TodosComponent` which I did not remove, so I have kept it that way.

Thanks for the explanation and taking the time to add relevant links for context, you are doing great work in bringing tanstack query to the angular world and all your efforts are really appreciated.

arnoud-dv (approved):

PR #8501, opened 2025-01-04
fix(react): allow using `enabled` when using `useQuery().promise`

### LINKED ISSUE #8499: Enabling a disabled query does not unsuspend
### Describe the bug

I am using the new mechanism to suspend explained here: https://tanstack.com/query/latest/docs/framework/react/guides/suspense#using-usequerypromise-and-reactuse-experimental

Disabled queries return a pending promise (which is ok), but the promise does not resolve when the query is enabled afterwards.
This happens because the query is not refetched when enabled changes from false to true.

https://codesandbox.io/p/devbox/mystifying-mopsa-qwhjwx?workspaceId=ws_5WeinpZuX6zrSSrhdSgo5C

### Steps to reproduce

Click the button to enable the query

### Expected behavior

When the query is enabled it should fetch and the promise should finally resolve

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome
- Version: 131.0.6778.205

### Tanstack Query adapter

react-query

### TanStack Query version

@tanstack/react-query@5.62.11

### TypeScript version

typescript@5.4.5

### Additional context

_No response_

KATT:
Closes #8499


Scutua (approved):

packages/query-core/src/__tests__/queryObserver.test.tsx
```diff
@@ -1233,4 +1234,38 @@ describe('queryObserver', () => {
...
+      queryKey: key,
+      enabled: false,
+      queryFn: () => 'data',
+    })
+    const results: Array<QueryObserverResult> = []
+
+    const success = pendingThenable<void>()
+
+    const unsubscribe = observer.subscribe((result) => {
+      results.push(result)
```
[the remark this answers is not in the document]
TkDodo: agree that’s how it should be 👍

packages/react-query/src/useBaseQuery.ts
```diff
@@ -152,7 +150,16 @@ export function useBaseQuery<
...
   ) {
-    const promise = isNewCacheEntry
+    // This fetching in the render should likely be done as part of the getOptimisticResult() considering https://github.com/TanStack/query/issues/8507
+    const state = cacheEntry?.state
+
+    const shouldFetch =
+      !state ||
+      (state.data === undefined &&
+        state.status === 'pending' &&
+        state.fetchStatus === 'idle')
```
TkDodo, line 160: I would really want this to only trigger if there is no cache entry. As soon as you have a cache entry, no matter what state, the “thing” becomes observable by other components, making doing something “in render” not allowed by the rules of react.

I guess this was done to catch transitions from `enabled:false` to `enabled:true` ?

Hardanish-Singh: +1

PR #8506, opened 2025-01-05
docs(react): update the react-native.md section

flodlc:


TkDodo (commented):
thanks. do you think this section is still needed ?

https://tanstack.com/query/v5/docs/framework/react/react-native#disable-re-renders-on-out-of-focus-screens

flodlc:
> thanks. do you think this section is still needed ?
I don't think so, it becomes too complexe for a very marginal usecase. Let's remove it!

hirbod:
@TkDodo I agree, the section can be removed. The new prop is enough

ferrannp:
@flodlc @hirbod why would you bring the `useIsFocused` hook back in the docs? That is a monster https://reactnavigation.org/docs/use-is-focused/ that is not recommended:

> Note that using this hook triggers a re-render for the component when the screen it's in changes focus. This might cause lags during the animation if your component is heavy.

It hurts pretty badly when you blur a screen.

We had a long discussion here: https://github.com/TanStack/query/discussions/5588 and merged changes here https://github.com/TanStack/query/pull/5734.

If you work in big react native apps, I would not call this a marginal usecase tbh.

Ideally you could do `subscribed: () => navigation.isFocused()`. But I see that `subscribed` does not accept a function... It's really a re-render / hook problem here. Mostly discussed in the previous discussion. Would be nice to re-open it and talk alternatives.

satya164:
Yea, `useIsFocused` is meant for small components that render something based on focus. It's not designed for something like React Query hooks, which are used in many components, and we don't recommend using it like that. React Compiler will likely optimize it now, but it's still extra work.

This specific section was a bit overcomplicated and could've been simplified to this (if I understood the shape for `notifyOnChangeProps`):

```js
const { dataUpdatedAt } = useQuery({
  queryKey: ['myKey'],
  queryFn: async () => {
    const response = await fetch(
      'https://api.github.com/repos/tannerlinsley/react-query',
    )
    return response.json()
  },
  notifyOnChangeProps: () => navigation.isFocused() : undefined : [],
})
```

PR #8560, opened 2025-01-21
Feature/composite monorepo

TkDodo:
This PR changes the structure of our monorepo towards a `composite` typescript setup.

In a composite monorepo, each package can be built in isolation with TSC to provide types other packages can consume. This has the following advantages and fixes the following issues:

### Code Navigation

Clicking on an import in your IDE that imports something from a dependent package, like:

```
import { QueryObserver } from '@tanstack/query-core'
```

in a framework adapter used to take you to the `.d.ts` file in the `build` directory. Now, it takes you directly to the source.

### Changes are picked up immediately

Because packages are linked together by sources via typescript path aliases, a change in `query-core` will automatically highlight all errors in your editor in adapter files. Previously, a build was needed, and oftentimes also a restart of the ts language server in your editor.

### Tests

Tests also know about typescript path aliases thanks to the `vite-tsconfig-paths` package (and dynamic path aliases, which are necessary to override what we have in `package.json` as main/exports entries). That means we can add a log statement to a `query-core` file and execute a test in a framework adapter without an in-between build step.

### No build dependency

All tasks like linting and type checking no longer need a `build` of dependent packages - they do need a `compile` task though that instructs typescript to build the output.

### New `compile` task

(Mainly) to make CI work, we have a new `compile` task that will instruct tsc to build all packages. For this, I had to switch svelte packages towards `"moduleResolution": "Bundler"`. I hope this doesn’t break anything @lachlancollins

### Isolated packages

Packages need to be isolated, so the `rootDir` was always set to `”.”`. However, we have imported things from `../../`, like the global eslint config. To work around this, I’ve created symlinks within the `rootDir` of all packages, because those imports were no longer allowed.

### References

Each package must define its own dependencies as `references` in `tsconfig.json`. Those reference point towards other packages, so that typescript can know about the build graph. This has to be kept in-sync with `workspace` dependencies in `package.json`

### Path mappings

TypeScript path mappings are defined in the root `tsconfig` json. All packages must be listed there - this is also what we use to build `aliases` for the `vite` configs.


jakebailey:
Drive by review (I saw this on bluesky):

> Packages need to be isolated, so the rootDir was always set to ”.”. However, we have imported things from ../../, like the global eslint config. To work around this, I’ve created symlinks within the rootDir of all packages, because those imports were no longer allowed.

In a composite build `rootDir` is defaulted to the config dir, so you probably don't need that. For the other files, have you considered using a separate config for those files instead and not including them in the build for any package? (Or honestly, not even including them at all?)

Speaking of referencing from the root, I would suggest _not_ putting anything in the config at the very root of the repo, instead making that a "solution" tsconfig whose only role it is to help tsserver find other configs (or to do a big `tsc -b`). An example of this is https://github.com/microsoft/TypeScript/blob/main/src/tsconfig.json or https://github.com/microsoft/DefinitelyTyped-tools/blob/main/tsconfig.json. Then, a separate base for each thing to inherit from.

https://github.com/typescript-eslint/typescript-eslint/pull/10361 is also something you might want to look at, since it's also a project moving to this layout and also using nx (which I believe is currently undergoing a refactor to be able to generate the exact thing you want).

> Tests also know about typescript path aliases thanks to the vite-tsconfig-paths package (and dynamic path aliases, which are necessary to override what we have in package.json as main/exports entries).

I'm surprised you need path aliases at all; in a monorepo those are usually very well covered by the package manager doing symlinking and then imports "just work". Path aliases are (IMO) best avoided beacuse they entirely bypass any and all module resolution logic (which may hide issues).

TkDodo:
> In a composite build rootDir is defaulted to the config dir, so you probably don't need that.

good to know 👍 , I don’t mind the explicitness though.

> I'm surprised you need path aliases at all; in a monorepo those are usually very well covered by the package manager doing symlinking and then imports "just work".

I tried to remove the path aliases, and tests then only work _after_ I’ve made a build. It also doesn’t immediately respect changes in a dependent config, and code navigation in the IDE will take me to the `build` directory. All of these things are solved once I enable path aliases. I’m not sure if it’s possible to get that to work without those?

> Speaking of referencing from the root, I would suggest not putting anything in the config at the very root of the repo

Yeah I’ll rethink this as a follow-up. The eslint config for example shouldn’t be needed at all because we have the flat config.

jakebailey:
> I tried to remove the path aliases, and tests then only work _after_ I’ve made a build.

How are you testing without building? I would have expected whatever's enabling that no-build setup (vitest? Some other thing) to be able to find the source instead, but maybe that requires doing something like https://colinhacks.com/essays/live-types-typescript-monorepo

TkDodo:
yeah, like the article says:

> Configure compilerOptions.paths in tsconfig.json to override resolution for local package names.

I just let it point to `src` and then tests run against src.

jakebailey:
Right, I was just trying to understand who's in charge of running TS source directly.

Issue #8569, opened 2025-01-22
`throwOnError` is still on `MutationObserverOptions`


AlexErrant:
https://github.com/TanStack/query/blob/4a0a129803d444f99830d68169331d01580fc14e/packages/query-core/src/types.ts#L1079

I believe this should be removed since it doesn't do anything (I think) and is confusing to new users like me.

> Then `mutate` would return nothing, `mutateAsync` would return a promise and `throwOnError` can be removed. [Src.](https://github.com/TanStack/query/issues/1111#issuecomment-703056774)

Would you accept a PR removing it, or is there a reason it's still there?

PR #8581, opened 2025-01-24
better document `throwOnError`'s default

AlexErrant:
As a new user, I had no idea what "global query config" meant, and searching for "config" wasn't helpful. This PR updates the wording.


docs/framework/react/reference/useMutation.md
```diff
@@ -94,7 +94,7 @@ mutate(variables, {
   - Defaults to a unique id (so that all mutations run in parallel)
   - Mutations with the same scope id will run in serial
 - `throwOnError: undefined | boolean | (error: TError) => boolean`
-  - Defaults to the global query config's `throwOnError` value, which is `undefined`
+  - Defaults to the global `QueryClient`'s `defaultOptions.queries.throwOnError`.
```
AlexErrant, line 97: Is this true for _mutations_? Ref: https://github.com/TanStack/query/issues/8569

emilwall:
The throwOnError part might be a residue from before #1130? I am really new to this library so please don't take my word for it.

TkDodo (commented):
> Defaults to the global `QueryClient`'s `defaultOptions.queries.throwOnError`.

I would rather just remove that sentence. All config options can be defined as default on the `queryClient`, or they can be set with `queryClient.setDefaultOptions` for a specific key, too.

This mechanism should not be documented in the reference docs, because we would need to state that for all options.

TkDodo (approved):

PR #8584, opened 2025-01-25
fix(eslint-plugin-query): add name attribute for eslint config inspector

NishargShah:
I am working on the eslint config inspector for my big flat config and I saw many anonymous configs where one of them is related to `tanstack/query` so I added names in the flat configs


TkDodo (approved):

magneticmeganlovesjhonny (commented):
Approved

PR #8590, opened 2025-01-27
feat(react-query): remove `.Provider` for new React 19 syntax

saul-atomrigs:
Hi @TkDodo , 

In React 19, we can use `<Context>` instead of `<Context.Provider>`.

In fact, I found the following warning in `packages/react-query/` :

스크린샷 2025-01-27 오후 7 46 46

React documentation here: https://react.dev/blog/2024/12/05/react-19#context-as-a-provider

So I suggest we remove the `.Provider` that are causing the warning

### DIFF
--- a/packages/react-query/src/QueryClientProvider.tsx
+++ b/packages/react-query/src/QueryClientProvider.tsx
@@ -37,9 +37,5 @@ export const QueryClientProvider = ({
     }
   }, [client])
 
-  return (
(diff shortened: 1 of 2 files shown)


TkDodo:
That’s nice, but it would break backwards compatibility with React 18.

saul-atomrigs:
Thanks @TkDodo
Would you accept turning off the eslint warning instead?

Issue #8591, opened 2025-01-27
`useSuspenseQuery` types: `data` can be `undefined`


OliverJAsh:
### Describe the bug

`useSuspenseQuery` can return `undefined` data e.g. if `select` throws, however the types do not reflect this.

### Your minimal, reproducible example

https://stackblitz.com/edit/tanstack-query-2bfa3dy1?file=src%2Findex.tsx&preset=node

### Steps to reproduce

```ts
const query = queryOptions({
  queryKey: ['foo'],
  queryFn: () => {
    const data = { a: { b: 'c' } };
    return Promise.resolve(data);
  },
  select: (x) => {
    throw new Error('foo');
    return x.a;
  },
});

function A() {
  const queryClient = useQueryClient();
  const { data, error } = useSuspenseQuery(query);
  console.log('Query result', { data, error });
  // ❌ `data` is `undefined` at runtime, but
  // types make it look like it will always exist
  return data.b;
}
```

### Expected behavior

The types should match the runtime behaviour.

I also wonder whether in this case the error should be thrown?

[From the docs](https://tanstack.com/query/latest/docs/framework/react/guides/suspense#throwonerror-default):

> Not all errors are thrown to the nearest Error Boundary per default - we're only throwing errors if there is no other data to show.

In this case, we don't have data to show, because `select` always fails.

### How often does this bug happen?

None

### Platform

N/A

### Tanstack Query adapter

None

### TanStack Query version

5.63.0

TkDodo:
I vaguely remember that this has been reported before. We should throw to the error boundary here, but the implementation:

```
throwOnError: query => query.state.data !== undefined
```

relies on data in the cache; if the transformation fails (select), data in the cache exists, but data for the observer does not.

we do this check to avoid throwing errors that happen on refetches to the error boundary.

I think we additionally have to return true here if `result.data` is undefined:

https://github.com/TanStack/query/blob/74f0d6aea96e6ce9557f59bf724ae90a26647851/packages/react-query/src/errorBoundaryUtils.ts#L66-L72

I think it should be:

```
query &&
(result.data === undefined ||
  shouldThrowError(throwOnError, [result.error, query]))
```

would you like to contribute this (+ a test case)?

OliverJAsh:
Just to make sure I understand, with that change, it would throw in this case?

I'd love to contribute. I tried making the change locally and running the tests, but there's tens/hundreds of test failures. 👀

OliverJAsh:
[...]

PR #8593, opened 2025-01-27
`useSuspenseQuery`: throw select errors

OliverJAsh:
Fixes https://github.com/TanStack/query/issues/8591. WIP.


TkDodo:
right, this destroys all scenarios without suspense, as error boundaries can also be used without suspense. We can only do this when suspense is set, so this needs an extra check.

OliverJAsh:
Thanks, I've pushed another change. The tests seem to pass now.

TkDodo (approved):

PR #8599, opened 2025-01-30
Feature/query client on context

TkDodo:


packages/eslint-plugin-query/src/__tests__/exhaustive-deps.test.ts
```diff
@@ -354,24 +354,6 @@ ruleTester.run('exhaustive-deps', rule, {
         }
       `,
     },
-    {
-      name: 'should ignore references of the queryClient',
```
TkDodo, line 358: note: we don’t need an exception for closing over the `queryClient` anymore if we inject it into the `queryFn`. @Newbie012 would you agree?

packages/query-core/src/queryCache.ts
```diff
@@ -117,7 +117,7 @@ export class QueryCache extends Subscribable<QueryCacheListener> {
 
     if (!query) {
       query = new Query({
-        cache: this,
+        client,
```
TkDodo, line 120: @juliusmarminge will this break `trpc` ? Are you using `new Query` directly somewhere? If you’re doing `queryCache.build()`, nothing is changing ...

packages/query-core/src/types.ts
```diff
@@ -116,6 +117,7 @@ export type QueryFunctionContext<
   TPageParam = never,
 > = [TPageParam] extends [never]
   ? {
+      queryClient: QueryClient
```
TkDodo, line 120: question: should we name this `client` instead of `queryClient` ?

I think `client` is the name we also have on the `Provider`. We usually show these examples

```

<QueryClientProvider client={queryClient}>...</QueryClientProvider>
```

so I’m thinking the new prop to destruct on the `QueryFunctionContext` should also be `client`, no?

```
useQuery({
  queryKey,
  queryFn: ({ client, signal }) => ...
})
```

packages/query-core/src/queryCache.ts
juliusmarminge: Dont think so

packages/query-core/src/types.ts
tannerlinsley: 👍

packages/query-core/src/query.ts
```diff
@@ -167,6 +169,7 @@ export class Query<
   #initialState: QueryState<TData, TError>
   #revertState?: QueryState<TData, TError>
   #cache: QueryCache
+  #client: QueryClient
```
DamianOsipiuk, line 172: Would be nice to get a getter for QueryClient from the instance of Query. 
I missed that before for experimental persister.

packages/eslint-plugin-query/src/__tests__/exhaustive-deps.test.ts
Newbie012: I guess so. I never had to deal with more than one query client in a single app. Is it a valid use case to reference a foreign query client inside queryFn?

packages/query-core/src/query.ts
TkDodo: you probably don’t need that now that it’s passed to the `context`, which the persister also has access to.

holic:
perfect timing, just went looking for this 🙏

jamesarosen:
I might argue that this is a breaking API change.

We use [@hey-api/openapi-ts](https://github.com/hey-api/openapi-ts), which generates API client functions out of an OpenAPI schema. The generated function accept a `client` argument that is an API client, e.g. `fetch`.

Before this change, we could do this:
```ts
const { data, error } = useQuery({
  queryKey: ['documents'],
  queryFn: documentsControllerGetAll,
})
```

But now the `client` types are incompatible, forcing us to change this to

```ts
const { data, error } = useQuery({
  queryKey: ['documents'],
  queryFn: async () => {
    return documentsControllerGetAll();
  },
})
```

TkDodo:
@jamesarosen in that sense, adding any parameter to any function would be a breaking change?

jamesarosen:
> adding any parameter to any function would be a breaking change?

I don't think there's a single easy answer.

The orthodox rule is "removing or changing a feature is a breaking change; adding a feature is a non-breaking change."

There are two ways to look at this change:

1. It introduced a new feature: the ability to work operate on the `client`. It's not a breaking change.
2. It modified the behavior of an existing feature. It changed the contract of `queryFn`.

Postel's Law -- "Be conservative in what you send; be liberal in what you accept." -- might be relevant here. TanStack-Query is now sending (passing) something new. Then again, if the change in question were a new optional parameter to `useQuery`, TanStack-Query wouldn't be sending something new; it would be accepting something new. And that could break in exactly the same way.

PR #8648, opened 2025-02-14
feat: imperative infinite queries

TkDodo:
let's bring back imperative infinite queries, but only allow them in an all-or-nothing mode, dependent on if `getNextPageParam` has been passed


riwu:
Thanks for this! Can we get it merged? It's critical to our use case of using it with AgGrid's serverside row model.

sina-ss:
any update for this?

sgtsquiggs:
This has failing tests - someone would need to fix those and mark this as ready to review.

PR #8654, opened 2025-02-16
Add `NoInfer` to `useQuery` return types

### LINKED ISSUE #8639: `TData` generic can be inferred or specified incorrectly, causing unexpected runtime errors.
### Describe the bug

Related to https://github.com/TanStack/query/issues/4770, revitalizing since `NoInfer`/overloads can at _least_ partially help. 

 
- `TQueryFnData` is directly inferred from `queryFn` specified in the hook options `options`
- `TData` _defaults_ to `TQueryFnData`, but can be specified via a generic or annotated return type
- When `TData` generic is populated (by inference, or explicitly) , nothing enforces that there's type compatibility until you put a `select` function, leading to uncaught runtime breaks. 

https://codesandbox.io/p/devbox/vigilant-forest-7j45zd?file=%2Fsrc%2Findex.tsx%3A58%2C1

### Steps to reproduce

See that `usePosts` and `usePosts2` both have a `data` type claiming to be `number`, but are instead a `Post` at runtime.

Examples:

```ts
function usePosts() {
  return useQuery<Post[], Error, number, string[]>({
    queryKey: ['posts'],
    queryFn: async (): Promise<Array<Post>> => {
      const response = await fetch(
        'https://jsonplaceholder.typicode.com/posts'
      );
      return await response.json();
    },
  });
}

function usePosts2(): UseQueryResult<number, Error> {
  return useQuery({
    queryKey: ['posts'],
    queryFn: async (): Promise<Array<Post>> => {
      const response = await fetch(
        'https://jsonplaceholder.typicode.com/posts'
      );
      return await response.json();
    },
  });
}
```

### Expected behavior

I expect to get a type-error, because I _haven't_ specified a `select` function and `TQueryFnData` does not equal `TData`, therefore we have easy, uncaught runtime breaks. 

## Potential solutions:
- Using `NoInfer` on the `UseQueryResult<NoInfer<TData>, TError>`, will at least prevent `usePost2` issue
- We should be able to write a conditional type by using additional overloads  (untested/hypothetically) to narrow on the condition where we have a `select` specified. 

```ts
type UseQueryOptionsWithSelect = ...
type UseQueryOptionsWithoutSelect = ...

declare function useQuery<TQueryFnData = unknown, TError = DefaultError, TData = TQueryFnData, TQueryKey extends QueryKey = QueryKey>(options: UseQueryOptionsWithoutSelect<TQueryFnData, TError, TData, TQueryKey>, queryClient?: QueryClient): UseQueryResult<TQueryFnData extends TData ? NoInfer<TData> : never, TError>;
declare function useQuery<TQueryFnData = unknown, TError = DefaultError, TData = TQueryFnData, TQueryKey extends QueryKey = QueryKey>(options: UseQueryOptionsWithSelect<TQueryFnData, TError, TData, TQueryKey>, queryClient?: QueryClient): UseQueryResult<NoInfer<TData>, TError>;
```

None

### Screenshots or Videos

_No response_

### Platform

N/A

### Tanstack Query adapter

None

### TanStack Query version

5.66.0

### TypeScript version

5.7.2

### Additional context

_No response_

braeden:
## TODOs:
- We should try to push `NoInfer` upstream into the `query-core` types (eg. `QueryObserverResult`)
  - I didn't have extensive time, but didn't have any luck w/o pushing this as _close_ to the usage as possible
  - If ^ doesn't work, I'd probably need some assistance to enumerate other (non-react) codepaths to change similarly.
- We should try to use native `NoInfer` from TS (https://devblogs.microsoft.com/typescript/announcing-typescript-5-4-beta/#the-noinfer-utility-type) instead of our own utility -- not 100% what our TS compatibility story is (eg what if a project doesn't have a "native" feature like `NoInfer` when we publish `d.ts` files)

### DIFF
--- a/packages/react-query/src/__tests__/useQuery.test-d.tsx
+++ b/packages/react-query/src/__tests__/useQuery.test-d.tsx
@@ -128,10 +128,8 @@ describe('initialData', () => {
   })
 
   describe('TData type inference', () => {
-
     it('no inference of TData from return annotations', () => {
-
-      const _testFn = (): UseQueryResult<string, unknown> =>  {
+      const _testFn = (): UseQueryResult<string, unknown> => {
         // @ts-expect-error expect number to be un-assignable to string
         return useQuery({
           queryKey: [],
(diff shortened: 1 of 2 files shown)


TkDodo:
> We should try to use native NoInfer from TS

we support TS 5.0 so this is currently not possible

PR #8665, opened 2025-02-18
fix(react-query): make error to be shown when using queryOptions with initialData and skipToken in queryFn inside useSuspenseQueries

### LINKED ISSUE #8664: Error not shown when using queryOptions with initialData and skipToken in queryFn inside useSuspenseQueries
### Describe the bug

The error should be shown when we use queryOptions with initialData and skipToken in queryFn inside useSuspenseQueries.
But the error does not occur.

```typescript
const query1 = queryOptions({
  queryKey: ['key1'],
  queryFn: skipToken,
  initialData: 'initial data',
})

useSuspenseQueries({ queries: [query1] })
```

-

### Steps to reproduce

Write the code on your text editor.
You will not see the error.

### Expected behavior

The error should be shown.

Every time

### Screenshots or Videos

_No response_

### Platform

macOS
VS Code

### Tanstack Query adapter

None

### TanStack Query version

5.66.3

### TypeScript version

_No response_

### Additional context

_No response_

gs18004:
closes #8664 

Previously, the error was not shown when we used queryOptions with initialData and skipToken in queryFn inside useSuspenseQueries.

```typescript
const query1 = queryOptions({
  queryKey: ['key1'],
  queryFn: skipToken,
  initialData: 'initial data',
})

useSuspenseQueries({ queries: [query1] })
```

Now the error occurs.

### DIFF
--- a/packages/react-query/src/__tests__/useSuspenseQueries.test-d.tsx
+++ b/packages/react-query/src/__tests__/useSuspenseQueries.test-d.tsx
@@ -167,4 +167,15 @@ describe('UseSuspenseQueries config object overload', () => {
 
     expectTypeOf(query1Data).toEqualTypeOf<string>()
   })
+
+  it('queryOptions with initialData and skipToken in queryFn should not work on useSuspenseQueries', () => {
+    const query1 = queryOptions({
+      queryKey: ['key1'],
+      queryFn: skipToken,
+      initialData: 'initial data',
+    })
+
+    // @ts-expect-error
+    useSuspenseQueries({ queries: [query1] })
+  })
 })
--- a/packages/react-query/src/queryOptions.ts
+++ b/packages/react-query/src/queryOptions.ts
@@ -3,6 +3,7 @@ import type {
   DefaultError,
   InitialDataFunction,
   OmitKeyof,
+  QueryFunction,
   QueryKey,
   SkipToken,
 } from '@tanstack/query-core'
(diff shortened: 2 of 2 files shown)


packages/react-query/src/__tests__/useSuspenseQueries.test-d.tsx
```diff
@@ -167,4 +167,15 @@ describe('UseSuspenseQueries config object overload', () => {
 
     expectTypeOf(query1Data).toEqualTypeOf<string>()
   })
+
+  it('queryOptions with initialData and skipToken in queryFn should not work on useSuspenseQueries', () => {
+    const query1 = queryOptions({
+      queryKey: ['key1'],
+      queryFn: skipToken,
```
TkDodo, line 174: would be good to have a test that has a conditional skipToken:

```
queryFn: Math.random() > 0.5 ? skipToken : () => Promise.resolve(5)
```

because that’s a lot closer to a real world use-case

TkDodo (approved):

lgenzelis:
This PR introduced a TS change, and I'm not sure whether it was intended or not. Before this, you could do
```ts
const { data } = useQuery({ queryKey: ['query-key'], initialData: 42 })
```
and `data` would be inferred as `number`. Starting with this PR, it's now inferred as `number | undefined`. If it was unintended, and you folks think it's relevant, I can create an issue for it.

TkDodo:
That's very relevant, and we should also have type-tests for this

gs18004:
> This PR introduced a TS change, and I'm not sure whether it was intended or not. Before this, you could do
> 
> ```ts
> const { data } = useQuery({ queryKey: ['query-key'], initialData: 42 })
> ```
> 
> and `data` would be inferred as `number`. Starting with this PR, it's now inferred as `number | undefined`. If it was unintended, and you folks think it's relevant, I can create an issue for it.

Okay, can you make an issue? I will try to fix it.

PR #8669, opened 2025-02-19
refactor(query-core): improve type safety and performance in `getMutationDefaults`

samuel871211:
# Description

Improve type safety and performance in `getMutationDefaults` by:
1. Using explicit return type `OmitKeyof<MutationObserverOptions<any, any, any, any>, 'mutationKey'>` to ensure the return type matches the type defined in `setMutationDefaults`
2. Aligning implementation with `getQueryDefaults` by using `Object.assign`


packages/query-core/src/queryClient.ts
```diff
@@ -585,14 +585,17 @@ export class QueryClient {
...
-    let result: MutationObserverOptions<any, any, any, any> = {}
+    const result: OmitKeyof<
+      MutationObserverOptions<any, any, any, any>,
+      'mutationKey'
+    > = {}
 
     defaults.forEach((queryDefault) => {
       if (partialMatchKey(mutationKey, queryDefault.mutationKey)) {
-        result = { ...result, ...queryDefault.defaultOptions }
+        Object.assign(result, queryDefault.defaultOptions)
```
TkDodo, line 598: > Aligning implementation with getQueryDefaults by using Object.assign

@Nick-Lucas do you remember why you changed this to `Object.assign` in `getQueryDefaults` ?

Nick-Lucas: Yes, it's potentially a faster operation and in some cases looks cleaner. That's it. But the difference in speed is probably virtually irrelevant as both approaches are very fast

https://www.measurethat.net/Benchmarks/Show/2485/0/spread-vs-objectassign#:~:text=Executions%20per%20second-,Object.assign%20two%20objects,226555.7%20Ops/sec,-Object.assign%20three

PR #8673, opened 2025-02-19
feat(query-core): add cancelPausedMutations to QueryClient and MutationCache

cmdecker95:
Add new method `cancelPausedMutations` to both the `QueryClient` and
`MutationCache` classes, which cancels (by way of removing from the
mutation cache) all paused mutations in a given scope, which in turn
will delete the scope if the last mutation has been removed.

Suggested in #8667


packages/query-core/src/mutationCache.ts
```diff
@@ -227,6 +227,22 @@ export class MutationCache extends Subscribable<MutationCacheListener> {
...
+      (x) => x.state.isPaused && scopeFor(x) === scope?.id,
+    )
+
+    notifyManager.batch(() => {
+      pausedMutationsInScope.forEach((mutation) => {
+        // Notify observers that the mutation was removed
+        this.notify({ type: 'removed', mutation })
+        // Remove the mutation from the cache, which will
+        // delete the scope when the last mutation is removed
+        this.remove(mutation)
```
TkDodo, line 242: `remove` calls `notify` under the hood. I think what this essentially does is:

- find some mutations in the cache (that are paused and match the scope)
- call `remove` on them.

the “paused and matches the scope” filter is probably what you need for your use-case, but it’s totally possible that this is a bit too narrow.

with that in mind, the abstraction is likely:

```
removeMutations(criteria) {
  this.getAll().filter(criteria).forEach(mutation => mutation.remove)
}
```

Given that all of this is publicly available API, I suggest you implement this for your use-case in user-land, and we can circle back to adding this to react-query when there’s more demand or we find something that isn’t doable with public APIs.

cmdecker95: Ah I see, this makes sense to me. I’ll work to implement it on my end then.

Thank you for reviewing!

PR #8683, opened 2025-02-21
fix(react-query): fix to show type error on wrong type annotation in `useQuery`

### LINKED ISSUE #7530: useQuery() hook vulnerable to type errors
### Describe the bug

There is a problem with hidden typescript type errors in the `useQuery()` hook when `select` is not defined.

Simple 1-line example provided inline

### Steps to reproduce

Try compiling the following:

```
  const queryResult: UseQueryResult<number> = useQuery({
    queryKey: ['key'],
    queryFn: async () => 'not a number',
  });
```

### Expected behavior

This should produce a type error, since `queryResult.data` will contain a `string`, contrary to the declared type of `number`.

But it compiles cleanly with no type errors detected.

A type error will, however, be detected if a `no-op` selectFn is provided:

```
  const queryResult: UseQueryResult<number> = useQuery({
    queryKey: ['key'],
    queryFn: async () => 'not a number',
    select: (s) => s,
  });
```

Every time

### Screenshots or Videos

_No response_

### Platform

N/A

### Tanstack Query adapter

react-query

### TanStack Query version

v5.40.1

### TypeScript version

v5.2.2

### Additional context

One possible avenue of attack might be to switch the template arguments to take the type of the select function instead, and try to infer TData from that. For example (where I've simplified to focus just on the types of interest):

```
interface MyQueryOptions<TQueryFnData = unknown, TSelect = unknown> {
  queryFn: () => Promise<TQueryFnData>;
  selectFn?: TSelect;
}

interface MyUseQueryResult<TData = unknown> {
  data: TData;
}

declare function useMyQuery<TQueryFnData = unknown, TSelect = unknown>(
  options: MyQueryOptions<TQueryFnData, TSelect>,
): MyUseQueryResult<TSelect extends (data: TQueryFnData) => infer TData ? TData : TQueryFnData>;

function thisHasATypeError(): MyUseQueryResult<number> {
  return useMyQuery({ queryFn: async () => 'not a number' });
}

function thisIsTypeCorrect(): MyUseQueryResult<number> {
  return useMyQuery({ queryFn: async () => 'not a number', selectFn: (s) => Number.parseInt(s) });
}
```

skiende74:
closes #7530 

```ts
 const queryResult: UseQueryResult<number> = useQuery({
    queryKey: ['key'],
    queryFn: async () => 'not a number',
  });
```

This should produce a type error because the correct type for queryResult should be `UseQueryResult<string>`. 
Previously, due to reverse type inference from the return type context, no error was detected when using the wrong type annotation.

This PR prevents reverse type inference from the result type annotation, ensuring that incorrect type annotations are caught as type errors.


skiende74:
Thanks for your review, @TkDodo!  
I’ve updated the import to:  
```ts
import { NoInfer } from 'query-core/util';
```
Could you please take another look?

manudeli:
image

Resolve eslint error in /packages/react-query/src/__tests__/useQuery.test-d.tsx please

skiende74:
@manudeli  
I’ve resolved the ESLint error. Could you please take another look? Thanks!

TkDodo (approved):

Issue #8604, opened 2025-02-04
useQueries have quadratic performance in relation to the number of queries


GabbeV:
### Describe the bug

This PR https://github.com/TanStack/query/pull/8295 introduced a significant performance regression for us. A commenter on that PR seem to have had a similar issue https://github.com/TanStack/query/pull/8295#issuecomment-2483173186.

We use the data loader pattern together with useQueries to give individual results from a batch fetch individual cache entries in react-query. Something like this:

```TS
const fooLoader = new DataLoader(ids => fetchFoos(ids));
```
```TS
const queries = useQueries({
  queries: ids.map(id => ({
    queryKey: ["foo", id],
    queryFn: () => fooLoader.load(id),
  })),
});
```

Looking at a performance profile i noticed this will result in one call to #findMatchingObservers in QueriesObserver for every query when the fetch resolve. #findMatchingObservers will then loop over all observers for for all queries internally making the performance quadratic O(n^2).

In our case we were loading way too many ids at once for this to be reasonable so the performance was already bad before the change but became browser freezing bad after the change.

Seems like the problem however should be avoidable. #onUpdate know what observer was updated so should not have to find it. This might however require some reorganization of the code and how observers are stored in QueriesObserver.

The indexOf in #onUpdate is also a source of quadratic performance that should be avoidable however indexOf is a way faster operation than what happens in #findMatchingObservers so it might not be an issue until even more queries are fetched at the same time.

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/boring-https-dxlhnv?file=%2Fsrc%2FApp.js%3A17%2C62

### Steps to reproduce

Click on the numeric buttons to try loading different amount of queries and click on the "useQueries" and "multiple useQuery" buttons to switch between using a single useQueries or multiple useQuery and observer the performance difference.

### Expected behavior

I expect useQueries to have similar or faster performance than multiple useQuery.

### How often does this bug happen?

Every time

### Platform

- OS: Windows 11 24H2 26100.2894
- Browser: Firefox 135.0b9, Chrome 132.0.6834.160

### Tanstack Query adapter

react-query

### TanStack Query version

5.64.2

joseph0926:
> First of all, I apologize for my non-English speaking background.

> I haven't actually tested with large queries, but from looking at the code, I suspect there may be performance issues due to the following reason.

- In the original implementation, `findMatchingObservers` was called only once inside `getOptimisticResult`, and its result (`matches`) was reused for tracking.
[...]

PR #8686, opened 2025-02-22
perf(query-core): useQueries have quadratic performance in relation to the number of queries(#8604)

zktm9903:
## Task
[issue](https://github.com/TanStack/query/issues/8604)
[comment](https://github.com/TanStack/query/issues/8604#issuecomment-2676128755)

## Modification

This change optimizes the time complexity of the #onUpdate function in queriesObserver from O(n) to O(1).
```ts
  #onUpdate(observer: QueryObserver, result: QueryObserverResult): void {
    const index = this.#observerMap.get(observer)
    if (index !== undefined) {
      this.#result = replaceAt(this.#result, index, result)
      this.#notify()
    }
  }
```


TkDodo (commented):
I’m not a fan of storing the observers in both a Map and an Array, and the whole thing becomes more complex.

The only reason we have an Array is because users pass in `Array<QueryObserverOptions>`, and we need to keep that order. Maybe storing this as a `Map<index, QueryObserver>` only would work?

zktm9903:
> I’m not a fan of storing the observers in both a Map and an Array, and the whole thing becomes more complex.
> 
> The only reason we have an Array is because users pass in `Array<QueryObserverOptions>`, and we need to keep that order. Maybe storing this as a `Map<index, QueryObserver>` only would work?

You're absolutely right. Fundamentally, we need to preserve the order, so using an array is necessary. Alternatively, it’s possible to use a Map and include the order in the values (rather than the keys), but based on my attempt, it significantly increased code complexity.
I believe this PR improves performance while keeping the code complexity to a minimum, though I completely agree with your concerns regarding complexity as well.

TkDodo:
as discussed in https://github.com/TanStack/query/pull/9467, there is not much to gain from this.

PR #8694, opened 2025-02-25
docs(react-query): fix links in advanced-ssr.md

CedericPrivat:
It seems like this document was moved w/o adjusting the links accordingly.


CedericPrivat:
Links will be fixed in https://github.com/TanStack/query/pull/8697

PR #8699, opened 2025-02-26
docs(react-query): fix incorrect link in documentation

haesoo-y:
The documentation contained incorrect file paths at reference section.
Fixed the incorrect paths to match the actual directory structure.


TkDodo:
`./` links won’t work; if you are in `/foo/bar.md` and want to link to `/foo/baz.md`, it needs to be `../foo/baz.md`, not `./baz.md`.

haesoo-y:
Thanks! I fixed it.

PR #8700, opened 2025-02-26
chore(react-query): add warning when useSuspenseQuery is used without queryFn

haesoo-y:
`useSuspenseQuery` is designed to accept queryFn as an optional parameter. (related https://github.com/TanStack/query/pull/8176)

However, when I built my Next.js app using `useSuspenseQuery` without `queryFn`, I encountered the following error.

`Error: Missing queryFn:`

image

(This Picture is `@tanstack/query-example-react-nextjs`)

The reason my Next.js app was **able to run without queryFn or initialData** was that I had already received the data on the **Server-Side using HydrationBoundary**.

To prevent this situation, I have added a warning message.

Alternatively, this could be addressed by modifying the implementation like the example below, but I don’t think this would be the best approach:

```tsx
initialData: options.initialData ?? queryClient?.getQueryData(options.queryKey),
```

### DIFF
--- a/packages/react-query/src/useSuspenseInfiniteQuery.ts
+++ b/packages/react-query/src/useSuspenseInfiniteQuery.ts
@@ -36,6 +36,9 @@ export function useSuspenseInfiniteQuery<
     if ((options.queryFn as any) === skipToken) {
       console.error('skipToken is not allowed for useSuspenseInfiniteQuery')
     }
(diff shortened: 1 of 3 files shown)


haesoo-y:
Sorry, I misunderstood the root cause. I’ll submit the PR again.

PR #8701, opened 2025-02-26
chore(react-query): add warning when useBaseQuery is called without a queryFn

haesoo-y:
Added a console error in development mode when `useBaseQuery` is called without a `queryFn` or `default queryFn`. 
This helps developers avoid potential issues by ensuring at least one of them is provided.

If you prefer, I can change it to `throw Error` instead of a console error. Let me know if needed.  

> Required, but only if no default query function has been defined See Default Query Function for more information.
> -[TanStack Query React Docs](https://tanstack.com/query/latest/docs/framework/react/reference/useQuery)-

### Commit Result
image

(This Picture is `@tanstack/query-example-react-nextjs`)

### Issue Situation
I was able to run my Next.js app without `queryFn` without any issues in Dev. However,  when I built my Next.js app using `useSuspenseQuery` without queryFn, I encountered the following error.

`Error: Missing queryFn:`
image

(This Picture is `@tanstack/query-example-react-nextjs`)

**The reason** my Next.js app was **able to run without `queryFn` or default queryFn** was that I had already received the data on the **Server-Side using `HydrationBoundary`**( => initial data).

To prevent this situation, I have added a warning message.


packages/react-query/src/useBaseQuery.ts
```diff
@@ -48,6 +48,12 @@ export function useBaseQuery<
         'Bad argument type. Starting with v5, only the "Object" form is allowed when calling query related functions. Please use the error stack to find the culprit call. More info here: https://tanstack.com/query/latest/docs/react/guides/migrating-to-v5#supports-a-single-signature-one-object',
       )
     }
+    const defaultQueryFn = queryClient?.getDefaultOptions().queries?.queryFn
```
TkDodo, line 51: this isn’t enough because defaults can be defined on different “levels”.

the correct way is to check if there is a `queryFn` in `defaultedOptions` after line 62, where we do:

```
const defaultedOptions = client.defaultQueryOptions(options)
```

haesoo-y: Thanks for pointing that out! I’ve moved the check after line 62 as you suggested.
Do you think an environment check `(process.env.NODE_ENV !== 'production')` is necessary here?

Also, would it be okay not to throw an error in this case?
If someone provides `initialData` and sets a long `staleTime`, it could potentially cause issues.

TkDodo: please do it like everywhere else: with an env check + a console error.

haesoo-y: thank you. I’ve updated the code.

TkDodo:
eh, the check is still wrong as you’re not checking against `defaultedOptions` ...

packages/react-query/src/useBaseQuery.ts
```diff
@@ -59,6 +59,15 @@ export function useBaseQuery<
     defaultedOptions,
   )
 
+  if (process.env.NODE_ENV !== 'production') {
+    // Make sure the queryFn is defined
+    if (!defaultedOptions.queryFn) {
```
TkDodo, line 64:
```suggestion
    if (!defaultedOptions.queryFn) {
```

packages/react-query/src/useBaseQuery.ts
```diff
@@ -59,6 +59,14 @@ export function useBaseQuery<
     defaultedOptions,
   )
 
+  if (process.env.NODE_ENV !== 'production') {
+    if (!defaultedOptions.queryFn) {
+      console.error(
+        `[${defaultedOptions.queryHash}]: No QueryFunction was passed as an option, and no default query function was found. The query function is only optional when using a default query function. More info here: https://tanstack.com/query/latest/docs/framework/react/guides/default-query-function`,
```
TkDodo, line 65:
```suggestion
        `[${defaultedOptions.queryHash}]: No queryFn was passed as an option, and no default queryFn was found. The queryFn parameter is only optional when using a default queryFn. More info here: https://tanstack.com/query/latest/docs/framework/react/guides/default-query-function`,
```

packages/react-query/src/useBaseQuery.ts
```diff
@@ -59,6 +59,14 @@ export function useBaseQuery<
     defaultedOptions,
   )
 
+  if (process.env.NODE_ENV !== 'production') {
+    if (!defaultedOptions.queryFn) {
```
frolic, line 63: Thoughts on also checking `enabled` in this conditional?

I often use a pattern like:
```ts
const queryKey = [ ... ];
const queryOptions = someOptional && anotherOptional
  ? { queryKey, queryFn() { ... } }
  : { queryKey, enabled: false };
```

so that I don't have to duplicate conditional checks inside the `queryFn` and `enabled`

TkDodo: you want to use `skipToken` for that:

```
import { skipToken } from '@tanstack/react-query'

queryOptions = {
  queryKey,
  queryFn: someOptional && anotherOptional ? () => queryFn() : skipToken
}
```

frolic: ohh TIL, thank you!

ibrahim-ytl:
Just to clarify. We use tanstack to also store a tiny global state. So in some place we use `useQuery` without the `queryFn`. Looking at the changes, seems like we need to always pass `skipToken ` even if the `enabled` is `false` and no `queryFn` is provided.

PR #8720, opened 2025-02-27
test(query-core): add test case for queryClient.isFetching

manudeli:


packages/query-core/src/__tests__/queryClient.test.tsx
```diff
@@ -400,6 +400,27 @@ describe('queryClient', () => {
...
+
+      expect(queryClient.isFetching()).toBe(0)
+      queryClient.prefetchQuery({
+        queryKey: key1,
+        queryFn: () => sleep(300).then(() => 'data'),
+      })
+      expect(queryClient.isFetching()).toBe(1)
+      queryClient.prefetchQuery({
+        queryKey: key2,
+        queryFn: () => sleep(200).then(() => 'data'),
```
TkDodo, line 416: since we use real timers per default, please keep those times short (5ms and 10ms at most), or use fake timers and advance them how you want.

packages/query-core/src/__tests__/queryClient.test.tsx
```diff
@@ -400,6 +400,27 @@ describe('queryClient', () => {
...
 
+  describe('isFetching', () => {
+    test('should return length of fetching queries', async () => {
+      const key1 = queryKey()
+      const key2 = queryKey()
+
+      expect(queryClient.isFetching()).toBe(0)
+      queryClient.prefetchQuery({
+        queryKey: key1,
+        queryFn: () => sleep(10).then(() => 'data'),
```
manudeli, line 411:
```suggestion
        queryFn: () => sleep(30).then(() => 'data'),
```

packages/query-core/src/__tests__/queryClient.test.tsx
```diff
@@ -400,6 +400,27 @@ describe('queryClient', () => {
...
+
+      expect(queryClient.isFetching()).toBe(0)
+      queryClient.prefetchQuery({
+        queryKey: key1,
+        queryFn: () => sleep(30).then(() => 'data'),
+      })
+      expect(queryClient.isFetching()).toBe(1)
+      queryClient.prefetchQuery({
+        queryKey: key2,
+        queryFn: () => sleep(5).then(() => 'data'),
```
manudeli, line 416:
```suggestion
        queryFn: () => sleep(25).then(() => 'data'),
```

TkDodo (changes requested):
now that I think about it, we’re actually having problems with these kinds of tests, as they tend to be pretty flaky (see useIsFetching). The reason is probably batching, and how much power the machine has / how much work it has to do, so it appears often in CI that updates are batched together.

using fake timers is better (we have an example for that in the code-base somewhere, too)

manudeli:
> now that I think about it, we’re actually having problems with these kinds of tests, as they tend to be pretty flaky (see useIsFetching). The reason is probably batching, and how much power the machine has / how much work it has to do, so it appears often in CI that updates are batched together.
> 
> using fake timers is better (we have an example for that in the code-base somewhere, too)

I agree, thanks for feedback!👍

manudeli:
> now that I think about it, we’re actually having problems with these kinds of tests, as they tend to be pretty flaky (see useIsFetching). The reason is probably batching, and how much power the machine has / how much work it has to do, so it appears often in CI that updates are batched together.
> 
> using fake timers is better (we have an example for that in the code-base somewhere, too)

I reflected your review

manudeli:
> please move this to beforeAll / afterAll in the describe block:
> 
> ```
> beforeAll(() => {
>   vi.useFakeTimers()
> })
> 
> afterAll(() => {
>   vi.useRealTimers()
> })
> ```
> 
> otherwise, if the test fails, fake timers will be on and it will likely fail all other tests, which can be difficult to debug.

I reflect your review now in [0156811](https://github.com/TanStack/query/pull/8720/commits/01568112856abedcf0c9785b21eebda230a79cc7)
I didn't know about fake timer. I think it's a really good feature. I'll be using it a lot in the future. Thanks for letting me know.

TkDodo (approved):
Oh, I thought you would limit it to the `describe('isFetching', () => {` block, not the whole file 😅 . But this is great - we want to move everything to fake timers anyways, as they are faster and more reliable

PR #8723, opened 2025-02-28
Fix broken links in persistQueryClient.md

jshaver-emsi:
Fixes broken links for createSyncStoragePersistor and createAsyncStoragePersistor on the persistQueryClient page.  I'm honestly not sure how to test that this will work, but the links that end up on the docs pages just need to keep the "/plugins/" directory in the path.  I'm pretty sure this will work.


TkDodo:
./ links don’t work unfortunately. They need to be `../plugins/`

TkDodo:
i fixed them all

PR #8725, opened 2025-03-01
test(query-core): add test case for ensureQueryFn with skipToken

manudeli:


TkDodo:
I don’t think we should allow running an imperative function with a `skipToken`. There’s no `enabled:boolean` for those, either.

A better fix would be to get a type error if this is attempted.

manudeli:
> I don’t think we should allow running an imperative function with a `skipToken`. There’s no `enabled:boolean` for those, either.
> 
> A better fix would be to get a type error if this is attempted.

@TkDodo I reflect your review in [274378a](https://github.com/TanStack/query/pull/8725/commits/274378abf9ec633676fdf403c76cd909b32f84fe)

TkDodo:
sorry, I don’t see how this addresses the issue. My point was that users shouldn’t be able to call `queryClient.ensureQueryData` with a `queryFn` that contains a `skipToken`. For that, we need to change the typings on the imperative functions on the queryClient. Then, we would need type tests to make sure that we can’t pass `skipToken` to them. So, no test that asserts the runtime log should be made because users shouldn’t be able to call it this way.

PR #8734, opened 2025-03-02
feat(react-query): Add `usePrefetchQueries` hook

DogPawHat:


packages/react-query/src/usePrefetchQueries.tsx
```diff
@@ -0,0 +1,17 @@
+import { useQueryClient } from './QueryClientProvider'
+import type { FetchQueryOptions, QueryClient } from '@tanstack/query-core'
+
+export function usePrefetchQueries(
+  options: {
+    queries: ReadonlyArray<FetchQueryOptions<any, any, any, any>>
+  },
+  queryClient?: QueryClient,
+) {
```
TkDodo, line 9: is `FetchQueryOptions<any, any, any, any>` really good enough here?

DogPawHat: I extracted out the rather through recursive `SuspenseQueriesOptions` recursive type reducer from `useSuspenseQueries` and modified it for using `FetchQueryOptions` in this case. Although were not gonna be using all that type information in this case at this point.

ryanagillie:
Any update on getting this merged? This would be very useful

TkDodo:
not a big fan of adding a new API for this. The types also shouldn’t need to be that complex because we don’t need to infer return types (there is no return type).

what stops from having `usePrefetchQuery` accept variadic arguments to prefetch multiple queries ?

DogPawHat:
Yeah I wasn't sure if we really needed that complex type inference machine myself, wasn't sure what the standard was.

> what stops from having `usePrefetchQuery` accept variadic arguments to prefetch multiple queries ?

not a lot I think apart from the fact the library seems to have a convention that methods that take in multiple queries are called `*Queries` rather then `*Query` (`useSuspenseQuery` vs `useSuspenseQueries`)

DogPawHat:
You'd go with something like this I think:

```tsx
// usePrefetchQueries.tsx
export function usePrefetchQueries<
  TQueryFnData = unknown,
  TError = DefaultError,
  TData = TQueryFnData,
>(
  options:
    | FetchQueryOptions<TQueryFnData, TError, TData, TQueryKey>
    | {
        queries: ReadonlyArray<
          FetchQueryOptions<TQueryFnData, TError, TData, TQueryKey>
        >
      },
  queryClient?: QueryClient,
) {

  const queryArray = 'queries' in options ? options.queries : [options]

  for (const query of queryArray) {
    if (!client.getQueryState(query.queryKey)) {
      client.prefetchQuery(query)
    }
  }
}

// usePrefetchQuery.ts
/**
 * @deprecated Use `usePrefetchQueries` instead.
 */
export function usePrefetchQuery<
  TQueryFnData = unknown,
  TError = DefaultError,
  TData = TQueryFnData,
>(
  options: FetchQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
  queryClient?: QueryClient,
) {
  return usePrefetchQueries(options, queryClient)
}
```

with `usePrefetchQuery` removed in v6

TkDodo:
okay, I think it’s more complicated than that, my bad. It obviously needs the complex types because otherwise you can’t do this:

```
usePrefetchQueries({
  queries: [
    {
      queryKey: ['key'],
      queryFn: () => Promise.resolve(5),
    },
    {
      queryKey: ['key'],
      queryFn: () => Promise.resolve('string'),
    },
  ],
})
```

DogPawHat:
Not sure that's the case I think. Pushing new tests now, will go over this in a bit, but I don't think the default `FetchQueriesOptions` generic is blocking this.

ryanagillie:
So uhhh, a year later?

DogPawHat:
I think I'm gonna close this one out if we're not gonna adopt this as an API

TkDodo:
yeah right now I don’t think we need this

PR #8736, opened 2025-03-03
test(query-core): add hashQueryKeyByOptions test code

dngur9801:


TkDodo (commented):
this doesn’t test the `byOptions` part, because options are never passed to `hashQueryKeyByOptions` in these tests.

so this is essentially the same as #8735 right now

dngur9801:
I modified the test code to verify that when the `queryKeyHashFn` option is provided, the specified hash function is called, and when it is not, the default `hashKey` function is used.

TkDodo (approved):

dngur9801:
I fixed it! Thank you

PR #8737, opened 2025-03-03
feat(query-core): make `MutateFunction` optional undefinable-variables

dinwwwh:
**Old Behavior:**  
`.mutate`, `.mutateAsync`, ... allowed omitting the `variables` parameter only when its type was exactly `void`. 

However, in TypeScript, when using union types like `unknown | void`, the `void` portion is ignored and treated as `unknown`, which still requires that `variables` be provided.

**New Behavior:**  
`.mutate`, `.mutateAsync`, ... now allow omitting the `variables` parameter when its type can be `undefined`. 

This change significantly broadens the optional types, supporting cases such as `unknown`, `any`, `undefined | ...`, `void | ...`, ...

Another benefit of the new approach is that, for popular schema libraries like Zod and Valibot, it is easier to define an undefinable schema than a void-able one.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved mutation function typing so variables are optional when supported and required when needed.
  - Preserved support for mutation options, error callbacks, omitted arguments, and spread-argument usage.
  - Improved consistency when invoking mutations across supported frameworks.

- **Tests**
  - Added compile-time coverage for mutation calls with unknown, any, void, optional, undefined-inclusive, and required variable types.
  - Added validation for callback parameters, required arguments, and positional argument forwarding.


lonelyplanetdev:
bump

packages/query-core/src/__tests__/types.test-d.tsx
```diff
@@ -0,0 +1,59 @@
...
+      unknown,
+      number | undefined,
+      unknown
+    >
+
+    expectTypeOf<Parameters<typeof mutate>[0]>().toEqualTypeOf<
+      number | undefined
+    >()
+
+    mutate() // can be called with no arguments
```
TkDodo, line 17: that’s a good improvement 👍 . I’m missing a test for the use-case where we have a mutation function that doesn’t take anything in, so the variables are `void`.

Also, I’d prefer if we create the mutation functions not via type assertions, but by whatever a `MutationObserver` returns. Something like:

```
const { mutate } = new MutationObserver(new QueryClient(), {
  mutationFn: async (_vars: number | undefined) => {
    return null
  },
})
```

packages/query-core/src/__tests__/types.test-d.tsx
TkDodo, line 1: can we please do this in a file called `mutations.test-d.tsx` ? All `-d.tsx` files are “types” tests, so `types.test-d.tsx` is not very meaningful

packages/query-core/src/types.ts
```diff
@@ -1149,14 +1149,28 @@ export interface MutateOptions<
...
+
 export type MutateFunction<
   TData = unknown,
   TError = DefaultError,
   TVariables = void,
   TContext = unknown,
 > = (
-  variables: TVariables,
-  options?: MutateOptions<TData, TError, TVariables, TContext>,
+  ...rest: MutateFunctionRest<TData, TError, TVariables, TContext>
```
TkDodo, line 1173: please also add some tests that make sure that `options` can still be passed.

packages/query-core/src/__tests__/types.test-d.tsx
unnoq: I've added the `void` case as requested. However, I couldn’t create a test using `MutationObserver` because the `mutate` function is implemented as a standalone function rather than using `MutateFunction`. As a result, the changes don’t impact its behavior. Let me know if you have any suggestions on how to approach testing this scenario.

TkDodo: oh interesting. question is if it would also work with the `mutate` function returned from `useMutation`, as that is what most users will use. The type is defined here:

https://github.com/TanStack/query/blob/3e3fba997bfd356f9a0e0aaa8fed453fcdf2f6f4/packages/react-query/src/types.ts#L197-L204

would be great to have a test in `useMutation.test-d.tsx` in the react-query adapter for this then 🙏

unnoq: I've already covered this case. Please check my tests.

TkDodo (approved):

TkDodo (approved):

PR #8765, opened 2025-03-06
docs: fix relative links in Advanced Server Rendering guide

iamursky:


iamursky:
I've jsut checked the docs, it didn’t fix the problem.

PR #8771, opened 2025-03-08
fix(core): make sure that calling setOptions will always notifyListeners

### LINKED ISSUE #8741: React Query - Missing fetchStatus changes for disabled queries with dynamic keys
### Describe the bug

If you use `useQuery` with `enabled: false` and dynamic keys you can miss re-renders because of missing `fetchStatus` updates. 

I.e. in this example (also posted on codesandbox) provided re-render will be triggered only when fetching the data is complete.

https://codesandbox.io/p/devbox/solitary-rgb-7szrj2

### Steps to reproduce

Run the example (it's the copy of codesandbox) and press the button "Set ID and trigger user load":
```jsx
import { createRoot } from "react-dom/client";
import {
  QueryClient,
  QueryClientProvider,
  useQuery,
} from "@tanstack/react-query";
import { useState } from "react";

const useUserInfoQuery = ({
  id,
  enabled,
}: {
  id: number | null;
  enabled: boolean;
}) => {
  return useQuery({
    queryKey: ["user", id],
    queryFn: () => {
        setTimeout(() => {
          resolve({ id, name: "John" });
        }, 2000);
      });
    },
    enabled: !!id && enabled,
  });
};

const App = () => {
  const [id, setId] = useState(null);

  const searchQuery = useUserInfoQuery({ id, enabled: false });

  return (
    <>
      <div>User fetching status is {searchQuery.fetchStatus}</div>
      <UserInfo id={id} />
      <button onClick={() => setId(42)}>Set ID and trigger user load</button>
    </>
  );
};

function UserInfo({ id }: { id: number | null }) {
  const searchQuery = useUserInfoQuery({ id, enabled: true });

  return <div>UserInfo data is {JSON.stringify(searchQuery.data)} </div>;
}

createRoot(document.getElementById("root")!).render(
    <App />
  </QueryClientProvider>
);
```

### Expected behavior

Expected behavior:
- I see `User fetching status is fetching` until the user info is updated

Actual behavior:
- I always see `User fetching status is idle`

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: macOS 14.5
- Browser: Google Chrome 133.0.6943.142 (Official Build) (arm64)

### Tanstack Query adapter

None

### TanStack Query version

5.66.11

### TypeScript version

5.7.3

### Additional context

Here are some of the factors which matter for bug reproducibility:
- setting constant `queryKey` instead of dynamic makes the bug disappear
- changing composition of components affects the bug (e.g. moving `<div>User fetching status is {searchQuery.fetchStatus}</div>` to a separate component may make the bug disappear)
- if you trigger re-render by something else (e.g. by adding `useIsFetching`) the bug is not reproduced

TkDodo:
we have an optimized `shouldNotifyListeners` logic internally in the QueryObserver that will make sure we only notify listeners that are interested in a change because they are using a specific key (tracked query properties)

the optimization to force a skip of the notification when updating options because the optimistic result should already contain all the necessary output falls apart in the situation described in the test and in the linked issue #8741.

while the issue itself is a niche edge-case, the fix is to actually remove a bunch of code, so I'm all for it

fixes #8741


packages/svelte-query-persist-client/tests/PersistQueryClientProvider.test.ts
TkDodo, line 1: @lachlancollins for the svelte-query persist plugin, there seems to now be one additional render / update. Interestingly, this doesn’t happen for the react adapter. we have the same tests and had the same assertions until now.

PR #8773, opened 2025-03-08
test(query-core): use fake timer for notifyManager.test.tsx

manudeli:


TkDodo:
thank you for all these PRs - they make our test pipeline faster and a lot more stable ❤️

manudeli:
> thank you for all these PRs - they make our test pipeline faster and a lot more stable ❤️

I'll do this until the end! Thanks for teaching me this way to test timer 🙇

PR #8774, opened 2025-03-08
test(react-query): use fake timer for HydrationBoundary.test.tsx

manudeli:


packages/react-query/src/__tests__/HydrationBoundary.test.tsx
```diff
@@ -356,7 +359,12 @@ describe('React hydration', () => {
       </QueryClientProvider>,
     )
 
-    expect(hydrateSpy).toHaveBeenCalledTimes(0)
+    await Promise.all(
+      Array.from({ length: 1000 }).map(async (_, index) => {
+        await vi.advanceTimersByTimeAsync(index)
+        expect(hydrateSpy).toHaveBeenCalledTimes(0)
+      }),
+    )
```
TkDodo, line 367: what does this do?

manudeli: I think this is better

```ts
await vi.runAllTimersAsync()
expect(hydrateSpy).toHaveBeenCalledTimes(0)
```

manudeli: I think this is better

```ts
await vi.runAllTimersAsync()
expect(hydrateSpy).toHaveBeenCalledTimes(0)
```

PR #8775, opened 2025-03-08
test(query-core): use fake timers for infiniteQueryObserver.test.tsx

manudeli:


packages/query-core/src/__tests__/infiniteQueryObserver.test.tsx
```diff
@@ -7,19 +7,21 @@ describe('InfiniteQueryObserver', () => {
...
     queryClient.clear()
+    vi.useRealTimers()
   })
 
   test('InfiniteQueryObserver should be able to fetch an infinite query with selector', async () => {
     const key = queryKey()
     const observer = new InfiniteQueryObserver(queryClient, {
       queryKey: key,
-      queryFn: () => 1,
+      queryFn: () => sleep(10).then(() => 1),
```
manudeli, line 24: I added sleep in queryFn like in real situation

Issue #8712, opened 2025-02-26
Angular: remove deprecated APIs


arnoud-dv:
`provideAngularQuery` was replaced by `provideTanStackQuery`
`injectQueryClient` was replaced by `inject(QueryClient)`

Remove these APIs only in the stable release package.

arnoud-dv:
Done as part of https://github.com/TanStack/query/pull/8780

PR #8777, opened 2025-03-08
refactor(angular-query): remove deprecated APIs

dominicbachmann:
remove deprecated APIs so that the first stable release only contains stable APIs.

BREAKING CHANGE: provideAngularQuery was removed. Use provideTanstackQuery instead.
BREAKING CHANGE: injectQueryClient was removed. Use inject(QueryClient) instead.
CLOSE: #8712


arnoud-dv:
Sorry maybe I should have described it more clearly but the idea is to remove these APIs at the last moment as part of the stable release package.

It's of course rather trivial to remove these functions, but that's why there's a task for that. I added it to the task description.

PR #8778, opened 2025-03-08
fix(react-query-persist-client): remove unnecessary props spreading

manudeli:
image


packages/react-query-persist-client/src/PersistQueryClientProvider.tsx
```diff
@@ -48,7 +47,7 @@ export const PersistQueryClientProvider = ({
   }, [client, isRestoring])
 
   return (
-    <QueryClientProvider client={client} {...props}>
+    <QueryClientProvider client={client}>
```
TkDodo, line 50: The better fix would be to not destruct `client`, use `props.client` everywhere, then do `<QueryClientProvider {...props}>` here.

The idea is that if we add a new prop to `QueryClientProvider`, it will also work here. That’s why the `PersistQueryClientProviderProps` build on top of `QueryClientProviderProps`.

manudeli: Ah I agree, I reflected your review in [bef7878](https://github.com/TanStack/query/pull/8778/commits/bef7878aca5da5a68d38d9bfc59d997d2e4aea0e)

PR #8814, opened 2025-03-16
feat(core): streamedQuery

TkDodo:


fz6m:
Can we consider using the name `useStreamQuery` ?

TkDodo:
> Can we consider using the name useStreamQuery ?

It’s not a hook, so no 😅 . But generally, yes, we consider different names. It also doesn’t have to have `stream` in the name. It’s based on `AsyncIterable` but that’s such a mouthful 😂

PR #8842, opened 2025-03-21
fix(angular-query): forward queryFn errors to the NgZone

### LINKED ISSUE #8839: Angular: caught errors in the queryFn do not run in the zone
### Describe the bug

I'm experiencing weird behaviour when using the `throwOnError: true` setting as the default in the QueryClient options. When an error occurs that is not handled via the template because of `throwOnError: true`, then the error should be caught by the Angular `ErrorHandler`. This somehow does not happen on first pageload.

As discussed [in this discord question](https://discord.com/channels/719702312431386674/1352225829966577684/1352225829966577684), a working fix could be to catch synchronous errors that happen inside of the `runOutsideAngular` callback in [`create-base-query.ts`](https://github.com/TanStack/query/blob/a9b073a85edc45331d4b346be8e5bf2660507e53/packages/angular-query-experimental/src/create-base-query.ts#L107) and trigger the `NgZone.onError.emit` function manually [as proposed here](https://discord.com/channels/719702312431386674/1352225829966577684/1352262073681186816).

https://github.com/JoepKockelkorn/tanstack-angular-throw-on-error-bug

### Steps to reproduce

See the README of the repro github repo.

### Expected behavior

The `ErrorHandler.handleError` is called on initial pageload.

Every time

### Screenshots or Videos

_No response_

### Platform

MacOS 15.3 (24D60)
Browser: Brave Version 1.76.74 Chromium: 134.0.6998.89 (Official Build) (arm64)

### Tanstack Query adapter

angular-query

### TanStack Query version

5.69.0

### TypeScript version

5.7.2

### Additional context

_No response_

JoepKockelkorn:
Any errors that occur in the queryFn or mutationnFn that should be thrown in case of `throwOnError: true` should run in the `NgZone`, or else the Angular [`ErrorHandler`](https://angular.dev/api/core/ErrorHandler) will not pick them up. A manual try...catch is needed because any errors that occur inside of [`ngZone.run`](https://angular.dev/api/core/NgZone#run) are not automatically reported in the zone:

> If a synchronous error happens it will be rethrown and not reported via onError.

This fixes #8839.

### DIFF
--- a/packages/angular-query-experimental/src/create-base-query.ts
+++ b/packages/angular-query-experimental/src/create-base-query.ts
@@ -108,23 +108,19 @@ export function createBaseQuery<
         observer.subscribe(
           notifyManager.batchCalls((state) => {
             ngZone.run(() => {
-              try {
-                if (
-                  state.isError &&
-                  !state.isFetching &&
-                  // !isRestoring() && // todo: enable when client persistence is implemented
-                  shouldThrowError(observer.options.throwOnError, [
-                    state.error,
-                    observer.getCurrentQuery(),
-                  ])
-                ) {
-                  throw state.error
-                }
-                resultFromSubscriberSignal.set(state)
-              } catch (error) {
-                ngZone.onError.next(error)
-                throw error
+              if (
+                state.isError &&
+                !state.isFetching &&
+                // !isRestoring() && // todo: enable when client persistence is implemented
+                shouldThrowError(observer.options.throwOnError, [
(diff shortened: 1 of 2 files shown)


JoepKockelkorn:
Not sure if the extra wrapping try...catch is needed 🤔

JoepKockelkorn:
Removed the try...catch wrappers as they were indeed redundant.

PR #8847, opened 2025-03-23
fix(examples, docs): ignored promise from invalidate queries

keyurgovrani:
This PR fixes the issue of Promise returned from invalidateQueries being ignored in documentation and examples

### DIFF
--- a/docs/framework/react/guides/optimistic-updates.md
+++ b/docs/framework/react/guides/optimistic-updates.md
@@ -70,7 +70,7 @@ This approach works very well if the mutation and the query live in the same com
 // somewhere in your app
 const { mutate } = useMutation({
   mutationFn: (newTodo: string) => axios.post('/api/data', { text: newTodo }),
-  onSettled: async () => await queryClient.invalidateQueries({ queryKey: ['todos'] }),
+  onSettled: () => queryClient.invalidateQueries({ queryKey: ['todos'] }),
   mutationKey: ['addTodo'],
 })
 
--- a/docs/framework/react/guides/optimistic-updates.md
+++ b/docs/framework/react/guides/optimistic-updates.md
@@ -16,7 +16,7 @@ const addTodoMutation = useMutation({
   mutationFn: (newTodo: string) => axios.post('/api/data', { text: newTodo }),
   // make sure to _return_ the Promise from the query invalidation
   // so that the mutation stays in `pending` state until the refetch is finished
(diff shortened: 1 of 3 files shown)


TkDodo:
the promise in the current code is not ignored:

```
onSettled: () => queryClient.invalidateQueries({ queryKey: ['todos'] }),
```

this code returns the promise from `invalidateQueries`, because arrow functions return the statement (if they have no brackets), and `invalidateQueries` returns a promise.

Wrapping this in `return await` is unnecessary.

keyurgovrani:
Thank you @TkDodo  for your feedback on my pull request. While it's true that arrow functions without braces return the result of their expression, I encountered issues with the code's behaviour, especially when arrow functions have curly braces. In those cases, an explicit `return` is required, and ESLint flagged warnings about the promise being ignored. This would ensure the promise is properly handled while maintaining both functionality and code quality. Could you please reconsider this approach and revisit the matter?

docs/framework/react/guides/optimistic-updates.md
```diff
@@ -70,7 +70,7 @@ This approach works very well if the mutation and the query live in the same com
 // somewhere in your app
 const { mutate } = useMutation({
   mutationFn: (newTodo: string) => axios.post('/api/data', { text: newTodo }),
-  onSettled: () => queryClient.invalidateQueries({ queryKey: ['todos'] }),
+  onSettled: async () => await queryClient.invalidateQueries({ queryKey: ['todos'] }),
```
TkDodo, line 73: async await is unnecessary here

```suggestion
  onSettled: () => queryClient.invalidateQueries({ queryKey: ['todos'] }),
```

docs/framework/react/guides/optimistic-updates.md
```diff
@@ -121,8 +121,8 @@ useMutation({
     queryClient.setQueryData(['todos'], context.previousTodos)
   },
   // Always refetch after error or success:
-  onSettled: () => {
-    queryClient.invalidateQueries({ queryKey: ['todos'] })
+  onSettled: async () => {
+    return await queryClient.invalidateQueries({ queryKey: ['todos'] })
   },
```
TkDodo, line 126: I see we weren’t returning the promise before, so yes, we can do that, but we should likely use the same syntax as above:

```
onSettled: () => queryClient.invalidateQueries({ queryKey: ['todos'] })
```

Issue #7487, opened 2024-05-28
broadcastClient: removing query from one tab removes it from all tabs, even if others still have observers


c-ehrlich:
### Describe the bug

When using `broadcastQueryClient`:
- Unmounting a query (with the last observer being removed) in one tab also removes that query in the other tab after gcTime hits
- After re-mounting the query, it is still not observed by the other tab, and invalidation works in neither direction

### Your minimal, reproducible example

(no stackblitz as their containerization seems to break broadcast channels, but see "Steps to reproduce" for a repo, also I've recorded a video demo)

### Steps to reproduce

Sample repo: https://github.com/c-ehrlich/broadcast-invalidation-demo

To see the bug
- `cd before`, `pnpm i`, `pnpm dev`
- open the application in two windows, show devtools
- invalidating across windows should work
- click "hide" in one window, wait for gcTime - the query is now no longer observed in both windows, and invalidation is broken
- click "show" to get the query back - invalidation will still be broken in both directions

To see a demo of my rough fix
- `cd after`, `pnpm i`, `pnpm dev`
- repeat the same steps as above. invalidation will keep working
- changes: https://github.com/c-ehrlich/broadcast-invalidation-demo/commit/e2133b9bb01876d31d9b259933b212002becd705

### Expected behavior

- Unmounting a query (with the last observer being removed) in one tab should not remove that query in the other tab
- Invalidation should work across tabs even when a query is not being observed in the tab that invalidates it

(not sure what the intended behaviour is, but these would be my expectations for the broadcastClient)

### How often does this bug happen?

Every time

### Screenshots or Videos

https://www.loom.com/share/e7da8b5d3cdb4b36a3fbf076d7ccb6b8

(the data updates twice shortly after each other at 1:16 because i had a third tab open, see "Additional context")

### Platform

Chrome 124.0.6367.208 on MacOS 14.3 (also confirmed in Firefox on Windows 11)

### Tanstack Query adapter

react-query

### TanStack Query version

5.39.0

### TypeScript version

5.2.2

### Additional context

My implementation still has one significant inefficiency:
- open three tabs instead of two
- hide the query in one of them, and invalidate there. now _both_ of the other windows will refetch
- ideally, only one refetch would be necessary. i could imagine achieving this through either "borrowing" the query from an instance that has it, or by keeping track of invalidation across instances and making only one of them refetch

Other considerations:
[...]

PR #8849, opened 2025-03-24
Issue #7487 fix: (query-broadcast-client-experimental) - removing query from one tab doesn't remove it from all tabs

gopnik5:
Issue #7487 fix: (query-broadcast-client-experimental) - removing query from one tab doesn't remove it from all tabs

### DIFF
--- a/examples/react/basic/package.json
+++ b/examples/react/basic/package.json
@@ -9,10 +9,10 @@
     "test:eslint": "eslint ./src"
   },
   "dependencies": {
-    "@tanstack/query-sync-storage-persister": "^5.72.2",
+    "@tanstack/query-sync-storage-persister": "^5.73.1",
     "@tanstack/react-query": "^5.72.2",
     "@tanstack/react-query-devtools": "^5.72.2",
-    "@tanstack/react-query-persist-client": "^5.72.2",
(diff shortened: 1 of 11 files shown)


packages/query-broadcast-client-experimental/src/index.test.ts
```diff
@@ -0,0 +1,21 @@
...
+  let queryClient: QueryClient
+  let queryCache: QueryCache
+
+  beforeEach(() => {
+    queryClient = new QueryClient()
+    queryCache = queryClient.getQueryCache()
+  })
+
+  it('should subscribe to the query cache', async () => {
+    broadcastQueryClient({
```
TkDodo, line 15: I think we are generally missing a cleanup function here that should be returned from `broadcastQueryClient`. It should:

- call `unsubscribe` returned from `queryClient.getQuryCache().subscribe`, otherwise, we are leaking a subscription
- call [channel.close()](https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel/close) on the created `BroadcastChannel`, otherwise, it can’t be garbage collected

then of course the test should call that, too.

TkDodo (commented):
looks good, I just noticed a general missing cleanup feature that we should add too please 🙏

PR #8862, opened 2025-03-27
docs(queryClient): JSDoc hint for queryClient.getQueryData

tylersayshi:
follow up from: https://bsky.app/profile/tylur.dev/post/3llennpm4kk2z

image

### DIFF
--- a/packages/query-core/src/queryClient.ts
+++ b/packages/query-core/src/queryClient.ts
@@ -120,7 +120,8 @@ export class QueryClient {
   }
 
   /**
-   * Imperative (non-reactive) way to retrieve data for a QueryKey. Should only be used in callbacks or functions where reading the latest data is necessary, e.g. for optimistic updates. 
+   * Imperative (non-reactive) way to retrieve data for a QueryKey.
+   * Should only be used in callbacks or functions where reading the latest data is necessary, e.g. for optimistic updates. 
    *
    * Hint: Do not use this function inside a component, because it won't receive updates.
    * Use `useQuery` to create a `QueryObserver` that subscribes to changes.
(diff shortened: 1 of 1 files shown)


packages/query-core/src/queryClient.ts
```diff
@@ -119,6 +119,12 @@ export class QueryClient {
     return this.#mutationCache.findAll({ ...filters, status: 'pending' }).length
   }
 
+  /**
+   * Get the data for a query key. This is helpful when query data is needed to be gotten once (non-reactive).
+   *
+   * Hint: If you're using this inside a component and wondering why it does not update, you
+   * should use `useQuery` to get reactive data.
+   */
```
TkDodo, line 127:
```suggestion
  /**
   * Imperative (non-reactive) way to retrieve data for a QueryKey. Should only be used in callbacks or functions where reading the latest data is necessary, e.g. for optimistic updates. 
   *
   * Hint: Do not use this function inside a component, because it won't receive updates.
   * Use `useQuery` to create a `QueryObserver` that subscribes to changes.
   */
```

packages/query-core/src/queryClient.ts
```diff
@@ -119,6 +119,12 @@ export class QueryClient {
     return this.#mutationCache.findAll({ ...filters, status: 'pending' }).length
   }
 
+  /**
+   * Imperative (non-reactive) way to retrieve data for a QueryKey. Should only be used in callbacks or functions where reading the latest data is necessary, e.g. for optimistic updates.
```
TkDodo, line 123:
```suggestion
   * Imperative (non-reactive) way to retrieve data for a QueryKey.
   * Should only be used in callbacks or functions where reading the latest data is necessary, e.g. for optimistic updates. 
```

PR #8878, opened 2025-03-30
feat(eslint-plugin): no void `queryFn`

Newbie012:
According to the [docs](https://tanstack.com/query/latest/docs/framework/react/guides/migrating-to-react-query-4#undefined-is-an-illegal-cache-value-for-successful-queries), we shouldn't allow returning `undefined` from a `queryFn`. This lint rule helps avoid it.


TkDodo (approved):
awesome 🙌

TkDodo:
yeah me neither. I think I merged this prematurely, I didn’t see the CI failure, and now it fails on `main`. Should I revert the merge ?

Newbie012:
I think it would be wise. I'll try to troubleshoot it, but I'm not sure I'm familiar with the current vitest setup

TkDodo:
Ok done ✅

PR #8887, opened 2025-03-30
test(react-query): fix async to promise and remove unnecessary await to match TypeScript requirements in useQueries.test.tsx

minseong0324:
Fix ESLint warnings for async functions without await

Resolved ESLint "require-await" warnings throughout the test files by:
- Removing unnecessary async keywords where no await is used
- Using Promise.resolve() to explicitly return promises instead

These changes don't affect functionality but make the linter happy and improve code clarity.


TkDodo:
can we make the lint rule an `error` after all these are fixed please?

PR #8894, opened 2025-03-31
feat(eslint-plugin-query): add support for trpc hooks in no-unstable-deps rule

magrinj:
Follow up of #7911 

### Summary

This PR enhances the `no-unstable-deps` ESLint rule by adding support for trpc hooks. The rule now correctly identifies and handles trpc hooks in addition to the existing React hooks.

### Why These Changes?

The original rule only supported React hooks, which limited its usefulness in projects using trpc. By adding support for trpc hooks, we ensure that the rule is more comprehensive and can be used in a wider range of projects.

Below an example of how trpc called are made:

```typescript
import { trpc } from '../utils/trpc';
export function MyComponent() {
  const mutation = trpc.login.useMutation();
  const handleLogin = () => {
    const name = 'John Doe';
    mutation.mutate({ name });
  };
  return (
    <div>
      <h1>Login Form</h1>
      <button onClick={handleLogin} disabled={mutation.isPending}>
        Login
      </button>
      {mutation.error && <p>Something went wrong! {mutation.error.message}</p>}
    </div>
  );
}
```

### DIFF
--- a/packages/eslint-plugin-query/src/rules/no-unstable-deps/no-unstable-deps.rule.ts
+++ b/packages/eslint-plugin-query/src/rules/no-unstable-deps/no-unstable-deps.rule.ts
@@ -58,6 +58,29 @@ export const rule = createRule({
       return undefined
     }
 
+    function getHookName(node: TSESTree.CallExpression): string | undefined {
+      if (node.callee.type === AST_NODE_TYPES.Identifier) {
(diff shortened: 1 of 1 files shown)


TkDodo:
trpc moved away from those hooks in v11, and it’s not our place to lint for integrations that are build on top of react-query, sorry.

Issue #8799, opened 2025-03-13
Client-only API called on the server side. Run client-only code in onMount, or conditionally run client-only component with <Show>.


huseeiin:
### Describe the bug

error in Solid Example: Solid Start Streaming stackblitz example

### Your minimal, reproducible example

https://stackblitz.com/github/tanstack/query/tree/main/examples/solid/solid-start-streaming?embed=1&theme=light&preset=node&file=src/index.tsx

### Steps to reproduce

https://stackblitz.com/github/tanstack/query/tree/main/examples/solid/solid-start-streaming?embed=1&theme=light&preset=node&file=src/index.tsx

### Expected behavior

it runs

### How often does this bug happen?

Every time

### Screenshots or Videos

![Image](https://github.com/user-attachments/assets/0d36930d-eb7d-47af-a01d-0ed2a61170b1)

### Platform

linux, chrome
^5.67.3

### Tanstack Query adapter

solid-query

### TanStack Query version

^5.67.3

birkskyum:
@huseeiin , the `Client-only API` issue can be resolved by updating start/vinxi to:

```tsx
    "@solidjs/start": "^1.1.3",
    "solid-js": "^1.9.5",
    "vinxi": "^0.5.3"
```

TkDodo:
@birkskyum does the linked PR address the issue? If so, we should merge it

PR #8816, opened 2025-03-16
chore: Resolve Client-only API error by updating vinxi version

dngur9801:
issue: #8799 

## Changes  
- Upgraded `vinxi` to `0.5.1` to fix the `"Client-only API called on the server side"` error.  

## Reason for Change  
- The issue was caused by a version mismatch between `vinxi` and `Vite 6`, leading to incorrect package resolution.  
- The `vinxi@0.5.1` update resolves the compatibility issue.  

## Reference  
- Related GitHub Issue: [solidjs/solid-start#1679 (comment)](https://github.com/solidjs/solid-start/issues/1679#issuecomment-2515393943)

PR #8896, opened 2025-03-31
update to vite 6

birkskyum:
Closes https://github.com/TanStack/query/issues/8799
Supersedes https://github.com/TanStack/query/pull/8816

**Updates in tests/examples/integrations/devDeps**
@vitejs/plugin-react -> patch update
vite 5 -> 6
svelte and svelte vite plugin  -> v5
angular 17/19 prerelease -> 19 stable release
vitest 2 -> 3
typescript -> 5.8.2


TkDodo (commented):
could you update vite everywhere so that we only have one version of vite in the repo ?

birkskyum:
The angular-v17 integration appear to be using vite 5.x underneath

birkskyum:
@TkDodo , I've updated vite everywhere, but it took a few more things with it to do so.

TkDodo:
We don't use vite for bundling so why would it be breaking for consumers?

birkskyum:
That's a good point - it's just examples, tests and integrations that's updated

packages/react-query/src/__tests__/useQuery.promise.test.tsx
```diff
@@ -59,6 +59,11 @@ describe('useQuery().promise', () => {
         },
       })
 
+      // Force Page to re-render when query state changes
+      React.useEffect(() => {
+        // noop
+      }, [query.status])
+
```
TkDodo, line 66: why is this necessary ?

birkskyum: The below test with [Page, MyComponent] only had one element returned otherwise

TkDodo: would ist work if we’d just render the status in a div instead:

```
<div>status: {query.status}</div>
```

birkskyum: this is in a `describe('useQuery().promise', () => {` suite, so it probably becomes a redundant test if we don't use query.promise.

TkDodo: yeah I meant additionally, just instead of the effect

birkskyum: That works.

PR #8913, opened 2025-03-31
chore(react-query-devtools): add @typescript-eslint/require-await for notice error for react-query-devtools package 

minseong0324:
ref: https://github.com/TanStack/query/pull/8887#issuecomment-2765491556


TkDodo:
this should be set in the root config. please make 1 PR that sets it in the root config and fixes all vioaltions in all packages at the same time. those small PRs are really too much here, thanks

PR #8921, opened 2025-04-01
align on ts 5.7.3

birkskyum:
We use a mix of ts versions:
5.4.5
5.6.3
5.7.2

This align all to 5.7.3

In order to support 5.7.3 for vue-query-devtools, it was necessary to bump the minor version of vue-tsc


TkDodo:
I think angular doesn’t yet support this version?

birkskyum:
19.2.x should be supporting it 
https://angular.dev/reference/versions#actively-supported-versions

PR #8933, opened 2025-04-01
fix(query-core): Rename defaultshouldRedactErrors to defaultShouldRedactErrors

minseong0324:
Renamed `defaultshouldRedactErrors` to follow proper camelCase naming (`defaultShouldRedactErrors`). Hope this small change makes things clearer! 😊


packages/query-core/src/hydration.ts
```diff
@@ -109,7 +109,7 @@ export function defaultShouldDehydrateQuery(query: Query) {
   return query.state.status === 'success'
 }
 
-export function defaultshouldRedactErrors(_: unknown) {
+export function defaultShouldRedactErrors(_: unknown) {
```
TkDodo, line 112:
```suggestion
function defaultShouldRedactErrors(_: unknown) {
```

TkDodo (approved):

PR #8935, opened 2025-04-02
fix(paths): use customConditions to define internal imports

### LINKED ISSUE #8865: build output not written to the correct folder
### Describe the bug

when running `pnpm build` to double check the output for #8862 I noticed that many files were generated from either tsc or tsup.

n/a

### Steps to reproduce

1. Make any change in packages/query-core to invalidate the nx cache
2. `pnpm build`

### Expected behavior

From the configs it looks like they should be written to `dist-ts`, `build/modern`, or `build/legacy`. It'd be helpful to get the intended output clarified, but I'm happy to help with another PR for this.

Every time

### Screenshots or Videos

image

### Platform

  System:
    OS: macOS 15.3.2
    CPU: (8) arm64 Apple M3
    Memory: 215.98 MB / 16.00 GB
    Shell: 3.7.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 22.12.0 - ~/.local/state/fnm_multishells/35918_1743092657466/bin/node
    pnpm: 9.15.0 - ~/Library/pnpm/pnpm

### Tanstack Query adapter

None

### TanStack Query version

latest

### TypeScript version

_No response_

### Additional context

_No response_

tylersayshi:
I found a technique in the [arktype](https://github.com/arktypeio/arktype/) monorepo that uses [customConditions](https://www.typescriptlang.org/tsconfig/#customConditions) to define imports to be used in the monorepo. It seems like a really clean solution for this to me, though it does mean updating each package.json

a good description of the different approaches at this problem and why `customConditions` is ultimately a good solution for it: https://colinhacks.com/essays/live-types-typescript-monorepo

fixes #8865


TkDodo:
oh, that’s neat. However, we also use the paths for `vitest` via `vite-tsconfig-paths`. Is there a way to have vitest pick up those customConditions, or would this work per default and all we’d need to do is remove `vite-tsconfig-paths` ?

It also seems that publint doesn’t like the exports order:

```
pkg.exports["."].import.types should be the first in the object as required by TypeScript.
```

TkDodo:
yeah vitest doesn’t seem to pick it up now:

```
Error: Failed to resolve entry for package "@tanstack/query-core". The package may have incorrect main/module/exports specified in its package.json.
```

I think we need to set:

```
  resolve: {
    conditions: ['tanstack-internal']
  }
```

https://github.com/vitejs/vite-ts-monorepo-rfc/blob/bc265396753de4326054dcbe4e538ad35a9b4554/with-exports-conditions/packages/app/vite.config.ts#L4-L6

and remove the `vite-tsconfig-paths` plugin

tylersayshi:
Thanks for the assist with those! Should be looking good now 🙂

tylersayshi:
remarking this to draft. It seemed like the cmd+click was working when I tried the change earlier, but is not working now

tylersayshi:
aha! order matters!

here's the reference blog for the approach and it even compares it to `paths`
https://colinhacks.com/essays/live-types-typescript-monorepo

I'll move the condition to be `@tanstack/custom-condition` then if publint does not like this being first, I will need to find a way to override that lint error, then finally contribute back to publint since this seems like it can be a really common/nice setup in typescript monorepos.

tylersayshi:
pushed with the failing publint errors for now. created an issue here: https://github.com/publint/publint/issues/176 to hopefully have this as supported behavior.

tylersayshi:
bluwy added the publint support for this super quickly 🎉 

do I need to remove the `dynamicAliases` next? https://github.com/tanstack/query/blob/main/packages/query-codemods/vite.config.ts#L10

TkDodo:
> do I need to remove the dynamicAliases next?

yeah I guess they are no longer needed

tylersayshi:
hm, I'm pretty stumped with the current CI errors. 

What's weird is that some packages pass without issue to resolve the internal references, while others fail. 

Also, each of the failing things pass locally, so my new theory is that the ts files from each package needs to be copied during the setup step of nx maybe? Or something along those lines?

tylersayshi:
marking this ready - seems like CI does indeed pass now 🙂

TkDodo:
Love it 😍

TkDodo (approved):

TkDodo:
hmm, it fail on main after the merge, too:

https://github.com/TanStack/query/actions/runs/14272681841

could it be order dependent, or maybe caching related ? Maybe there’s an additional output by typescript that needs to be cached? nx distributes the runs over multiple machines, so maybe something is missing on some of them ...

tylersayshi:
> could it be order dependent, or maybe caching related ? Maybe there’s an additional output by typescript that needs to be cached? nx distributes the runs over multiple machines, so maybe something is missing on some of them ...

huh, ok. This should probably be reverted for now then.

Is there a way to disable nx cache to run the tests locally?

tylersayshi:
> could it be order dependent

I think this is the case. When run locally, query-core compiles first, which would make it available, where in the CI it shows as not found

image

nx is supposed to understand dependency order, so this seems strange that it could run out of order though

https://github.com/nrwl/nx/issues/6577#issuecomment-891912104

TkDodo:
looking at the build output it definitely runs before:

https://github.com/TanStack/query/actions/runs/14272681841/job/40009248769

Screenshot 2025-04-05 at 17 41 24

so, I really don’t know. We also had everything green on the PR.

TkDodo:
let’s try with `--skip-nx-cache` on the task that runs in CI:

3319d35d

TkDodo:
yeah that worked; I hope I find a way to purge the remote cache

TkDodo:
I changed `nx.json` to get complete cache invalidation, and even with all cache misses, it still fails. I don’t think there’s something wrongly cached, I think it’s more related with distributed execution. If compile compile something and creates an output, but test runs on a different machine that doesn’t get access to what compile has produced, that would imo explain the behaviour.

but why was it working before and not now? are outputs still written to `dist-ts` like they were before? Because that’s what `compile` caches:

https://github.com/TanStack/query/blob/57d159c425e9579d61b91989182451b1e7c07f20/nx.json#L31

tylersayshi:
So the fails are specific to `vitest`... This is a shot in the dark, but what if we exclude `resolve.conditions` when `process.env.CI` is true? The build output from compile should be usable at that point, so it should be safe to do.

TkDodo:
it also fails on test:types, I don’t think its `vitest` specific, because why would it work locally then?

They purged the cache for us, and then it almost went through with just 3 pending tasks, but something else failed. I re-started, and now 167 tasks came from cache, and we have a failure again:

https://cloud.nx.app/runs/Wu8RdNSBRi

So my thinking is still that we are not persisting everything in the cache that is needed, but I don’t know what’s missing

TkDodo:
only if I run:

```
npx nx run @tanstack/query-core:build
```

the tests start to work!

but that is not a dependency we have defined in nx, so it’s clear why it works sometimes and mostly doesn’t.

now the real question is: why does vite needs a build before it can resolve those? As a quick workaround, we could make the test task depend on the build task (instead of the compile task), but since we build with tsup I’d rather avoid this...

PR #8943, opened 2025-04-02
fix(solid-query): use QueryClient from query-core

birkskyum:


TkDodo:
I think now you can't set `reconcile` anymore as a global defaultOption as this was solid specific

birkskyum:
Why is that option exclusively available to solid?

TkDodo:
I don’t know what it does, it seems to be specific to solid, which is why it was living in the adapter. I think we might need to revert this PR to avoid breaking the public API.

PR #8949, opened 2025-04-03
test(query-core): add test case for hydration

minseong0324:
improve test coverage of `hydration.ts` from 85% to 100%.

## Before
image

## After
image


packages/query-core/src/__tests__/hydration.test.tsx
```diff
@@ -1142,4 +1142,130 @@ describe('dehydration and rehydration', () => {
...
+    const dehydrated = dehydrate(queryClient)
+
+    expect(dehydrated.queries[0]?.promise).toBeInstanceOf(Promise)
+
+    try {
+      await dehydrated.queries[0]?.promise
+      expect(true).toBe(false)
+    } catch (error) {
+      expect(error).toBe(testError)
+    }
```
TkDodo, line 1177: I think this should be:

```
await expect(dehydrated.queries[0]?.promise).rejects.toBe(testError)
```

packages/query-core/src/__tests__/hydration.test.tsx
```diff
@@ -1142,4 +1142,130 @@ describe('dehydration and rehydration', () => {
...
+    await promise
+    queryClient.clear()
+  })
+
+  test('should handle errors in promises for pending queries', async () => {
+    const consoleMock = vi.spyOn(console, 'error')
+    consoleMock.mockImplementation(() => undefined)
+
+    const originalNodeEnv = process.env.NODE_ENV
+    process.env.NODE_ENV = 'production'
```
TkDodo, line 1188: why is that needed?

packages/query-core/src/__tests__/hydration.test.tsx
```diff
@@ -1142,4 +1142,130 @@ describe('dehydration and rehydration', () => {
...
+      }
+      expect(consoleMock).toHaveBeenCalledWith(
+        expect.stringContaining('test error'),
+      )
+    }
+
+    process.env.NODE_ENV = originalNodeEnv
+    consoleMock.mockRestore()
+    await promise
+    queryClient.clear()
```
TkDodo, line 1269: instead of clearing the queryClient at the end, let’s just not share a client between tests but have each test instantiate their own please.

packages/query-core/src/__tests__/hydration.test.tsx
minseong0324: Thanks for your feedback! I reflected in 3ab87c9dc9aa1ce7dfb4d98d59c366a319a4de1a

packages/query-core/src/__tests__/hydration.test.tsx
minseong0324: Initially added NODE_ENV manipulation to test both production and development error handling paths separately, but realized it's unnecessary since we're testing the error redaction logic itself, not the environment-specific logging behavior. The core error transformation remains the same in both environments.

I reflected in d43131fefe546d901d3fa5d6607212dd6107e09e

packages/query-core/src/__tests__/hydration.test.tsx
minseong0324: Ok! I reflected at afed97894df04b363e622f2153a8ed20b13b3817

packages/query-core/src/__tests__/hydration.test.tsx
TkDodo: yeah but there are a lot more places where you did this. I though you would maybe change them all to this pattern?

minseong0324: I reflected at b86dd29c901e47fbd54f444228b2f73ff9d7232e

minseong0324: I reflected at b86dd29c901e47fbd54f444228b2f73ff9d7232e .

minseong0324: I reflected at https://github.com/TanStack/query/commit/1f5e12b209eab679973ea21eb60a6263d76c52a0 @TkDodo

PR #8956, opened 2025-04-04
feat(vue-query): useInfiniteQuery options support maybe ref type

aa900031:
The useInfiniteQuery can directly use Ref in the runtime, but it's not supported at the Type level.

### DIFF
--- a/packages/vue-query/src/useBaseQuery.ts
+++ b/packages/vue-query/src/useBaseQuery.ts
@@ -49,7 +49,7 @@ type UseQueryOptionsGeneric<
       TQueryFnData,
       TError,
       TData,
-      TQueryFnData,
+      TQueryData,
       TQueryKey,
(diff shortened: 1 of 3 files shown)


aa900031:
@DamianOsipiuk Please review this, thank you

packages/vue-query/src/useBaseQuery.ts
```diff
@@ -49,7 +49,7 @@ type UseQueryOptionsGeneric<
       TQueryFnData,
       TError,
       TData,
-      TQueryData,
+      TQueryFnData,
```
DamianOsipiuk, line 52: I'm not sure if this is a proper change. Why this is needed?

aa900031: Oops! that's my fault, I'm already fix it

PR #8958, opened 2025-04-05
Remove predicate typing for query filters

### LINKED ISSUE #8768: invalidateQueries predicate option: queryKey property is typed too strictly
### Describe the bug

When using the `invalidateQueries` method of the QueryClient, the queryKey is typed a bit too strictly in the `predicate` method that we pass to it through the QueryFilters parameter. 

When "exact" is false, the `queryKey` filter is not an exact match, so the queries passed to the `predicate` option can contain query keys that have a different type from the type of the top-level `queryKey`. The latest version of the library however types it as the exact same type as the provided queryKey.

In the minimal reproducible example, this is highlighted using a const array of string as query key and the type of the queryKey in the predicate function being the exact same type, even though it could be any other array starting with the same string and then containing any unknown values.

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgRQK4FMoE8DCAbYdAO3gF84AzKCEOAcgAEYBDIgZxYGMBrAeinTNOMALQBHDNjoBuAFCzOEdvG7oscALxwA2nQAmEAOZs6AXThxmbOIuVzevOADkIcTNSiyi6AO4pJuATEMAAUAJQAdMBEAG7MBHrMMOhomIRsIQiyFhKYWADSagBccKpYADTZbgAeQjAlFPFs6JUWYAJ6wJxJ6CUhudhhmgB8iFUWDnAAssyqcGw06KXRenAQFDYAFug8a0RwMNtwAhxrG4dLA+plNgI91qxuUB7jcMAb-QERV4VY2gCM5g0wPoEB83igdCGWQssOO6BgqCg+0aeGacjhpHkcIEiORBygGAxcCxpDCciAA

### Steps to reproduce

The Typescript error is visible in the playground and in the errors panel on the right.

### Expected behavior

I would expect the type of the queryKey value in the `predicate` method to only be so strict when the "exact" option is true, or to always be less strict by only stating that it starts with the same values as the provided `queryKey` in the QueryFilter.

Every time

### Screenshots or Videos

_No response_

### Platform

macOS, Typescript 5.8.2

### Tanstack Query adapter

None

### TanStack Query version

5.67.1

### TypeScript version

5.8.2

### Additional context

_No response_

Nick-Lucas:
Fixes #8768


Nick-Lucas:
So @TkDodo I'm in two minds if we should go as far as stripping out the generics entirely. It's more of a breaking change to remove them than to leave them unused, especially since the 4th type param (query key) does get used currently.

What do you think?

TkDodo:
TypeScript errors on unused type parameters:

```
src/utils.ts(19,3): error TS6133: 'TError' is declared but its value is never read.
src/utils.ts(20,3): error TS6133: 'TData' is declared but its value is never read.
```

so what option do we have other than deleting them?

Nick-Lucas:
Should be possible to ignore the warnings I think? So it's more a question of where your mind is at as a maintainer

I didn't do anything today beyond start this draft, there will be tests failing as well that I can sort out either way

TkDodo:
> Should be possible to ignore the warnings I think?

this isn’t just an eslint warning - it’s a TS error that you would need to silence with `ts-ignore`, which is not something I’d recommend because if the comments get stripped, the error might resurface

Nick-Lucas:
Hey @TkDodo I've stripped this out everywhere, might need synchronising with tRPC as I think we made some changes to support these generics but that should be quick

Nick-Lucas:
failing test seems to be flake rather than anything I've done

TkDodo (approved):

PR #8959, opened 2025-04-06
fix: relocate custom condition

tylersayshi:
fixes the CI failures added in: https://github.com/TanStack/query/pull/8935

thanks! @beaussan for the help: https://bsky.app/profile/beaussan.io/post/3lm3oodmras2e

apologies for my oversight on adding this originally

### DIFF
--- a/packages/eslint-plugin-query/package.json
+++ b/packages/eslint-plugin-query/package.json
@@ -39,10 +39,11 @@
   "module": "dist/index.js",
   "exports": {
     ".": {
+      "@tanstack/custom-condition": "./src/index.ts",
       "import": {
         "@tanstack/custom-condition": "./src/index.ts",
-        "types": "./dist/index.d.ts",
-        "default": "./dist/index.js"
+        "types": "./dist/esm/index.d.ts",
+        "default": "./dist/esm/index.js"
       },
       "require": {
         "types": "./dist/index.d.cts",
--- a/packages/query-async-storage-persister/package.json
+++ b/packages/query-async-storage-persister/package.json
@@ -40,8 +40,8 @@
   "react-native": "src/index.ts",
   "exports": {
     ".": {
+      "@tanstack/custom-condition": "./src/index.ts",
       "import": {
-        "@tanstack/custom-condition": "./src/index.ts",
         "types": "./build/modern/index.d.ts",
         "default": "./build/modern/index.js"
       },
--- a/packages/query-broadcast-client-experimental/package.json
+++ b/packages/query-broadcast-client-experimental/package.json
@@ -37,8 +37,8 @@
   "module": "build/legacy/index.js",
   "exports": {
     ".": {
+      "@tanstack/custom-condition": "./src/index.ts",
       "import": {
-        "@tanstack/custom-condition": "./src/index.ts",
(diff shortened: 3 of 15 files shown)


tylersayshi:
hm, maybe there's another thing wrong. Not sure why it passed before merging main and failed after 👀

TkDodo:
I did the same steps-to-reproduce here and it still fails for me locally:

- `pnpm run clean`
- `npx nx run @tanstack/query-core:compile`
- `npx nx run @tanstack/react-query:test:lib`

beaussan:
Here is my findings so far:

Based on those steps:
- pnpm run clean
- npx nx run @tanstack/query-core:compile
- npx nx run @tanstack/react-query:test:lib

Setting `typecheck: { enabled: false },` to the vitest config for react-query, we have _more_ tests running, but still have one error:

image

TkDodo:
FYI, I changed the dependencies in nx to have `test` depend on `build` until this is sorted, as other PRs can’t complete otherwise:

1771dd25

If this is figured out, please revert this change here 🙏

TkDodo:
and now we get errors like this:

```
TypeCheckError: Output file '/home/workflows/workspace/packages/query-persist-client-core/dist-ts/src/index.d.ts' has not been built from source file '/home/workflows/workspace/packages/query-persist-client-core/src/index.ts'.
```

It’s a pretty messy spot we’re in right now, @tylersayshi I think we should revert the `customConditions` until this is figured out completely

tylersayshi:
> I think we should revert the customConditions until this is figured out completely

I agree. Sorry for the trouble it's caused so far 🫠. I assumed it would be more straightforward

tylersayshi:
added the revert PR: https://github.com/TanStack/query/pull/8973

tylersayshi:
> I did the same steps-to-reproduce here and it still fails for me locally:
> 
> * `pnpm run clean`
> * `npx nx run @tanstack/query-core:compile`
> * `npx nx run @tanstack/react-query:test:lib`

When I run this it succeeds:

* `pnpm run clean`
* `npx nx run @tanstack/query-core:compile`
* `npx nx run @tanstack/query-persist-client-core:compile`
* `npx nx run @tanstack/react-query:test:lib`

Is "@tanstack/query-persist-client-core" supposed to automatically compile when compiling query-core?

TkDodo:
@tylersayshi I’ve changed `test:lib` to depend on `^build` now and that’s why it works.

so, if we change `test:lib` in `nx.json` back to `"dependsOn": ["^compile"],` as it should be, we still get the error. Note that if I do:

- `pnpm run clean`
- `npx nx run @tanstack/react-query:test:lib`

I get the error, and I can see that `dist-ts` directories exist in the following packages:

- `query-core`
- `query-persist-client-core`

That means the `compile` dependencies are set up correctly. The question is why it’s not picked up.

Also note that the `build` task writes to the `build` directory, so it seems like vitest (incorrectly) expects output to be there?

TkDodo:
okay, I merged the revert so this is obviously outdated now. Besides the struggles I still like the `customCondition` approach and would love it if you found out how that works can help us get this over the finish line ❤️

PR #8960, opened 2025-04-06
feat(react-query): add mutationOptions

Ubinquitous:
mutationOptions helps extracting mutation options into separate functions.


packages/react-query/src/__tests__/mutationOptions.test-d.tsx
```diff
@@ -0,0 +1,23 @@
...
+  it('should infer types for callbacks', () => {
+    return mutationOptions({
+      mutationFn: () => Promise.resolve(5),
+      mutationKey: ['key'],
+      onSuccess: (data) => {
+        expectTypeOf(data).toEqualTypeOf<number>()
+      },
+    })
+  })
+})
```
manudeli, line 23: In my opinion, Adding type testing about DataTag would be better

Ubinquitous: Thank you, I just added a test case for DataTag, but I can't test it locally because of an error in @tanstack/query-core. Could you please tell me how to run test in local?

Here is the error message:
스크린샷 2025-04-07 오후 5 13 51

packages/react-query/src/mutationOptions.ts
```diff
@@ -0,0 +1,105 @@
...
+  'mutationFn'
+> & {
+  mutationFn?: Exclude<
+    UseMutationOptions<
+      TMutationFnData,
+      TError,
+      TData,
+      TMutationKey
+    >['mutationFn'],
+    SkipToken | undefined
```
TkDodo, line 39: skipToken is specific to queries. This isn’t needed.

packages/react-query/src/mutationOptions.ts
```diff
@@ -0,0 +1,105 @@
...
+export type DefinedInitialDataOptions<
+  TMutationFnData = unknown,
+  TError = DefaultError,
+  TData = void,
+  TMutationKey = unknown,
+> = Omit<
+  UseMutationOptions<TMutationFnData, TError, TData, TMutationKey>,
+  'mutationFn'
+> & {
+  initialData:
```
TkDodo, line 54: there is no initialData for mutations

packages/react-query/src/mutationOptions.ts
```diff
@@ -0,0 +1,105 @@
...
+  TMutationKey = unknown,
+>(
+  options: DefinedInitialDataOptions<
+    TMutationFnData,
+    TError,
+    TData,
+    TMutationKey
+  >,
+): DefinedInitialDataOptions<TMutationFnData, TError, TData, TMutationKey> & {
+  mutationKey: DataTag<TMutationKey, TMutationFnData, TError>
```
TkDodo, line 73: we also don’t need a `DataTag`, this is query specific.

TkDodo (commented):
sorry but it seems like you’ve copy-pasted queryOptions and replaced the word `query` with the word `mutation` :/

your starting point should’ve been `useMutation`, not `queryOptions`

Ubinquitous:
Thank you for reviewing my PR. I thought queryOptions and mutationOptions could be structured similarly since it was an options-related function. I re-created useMutation as start. I changed it to only have UseMutationOptions, excluding unnecessary data tags and initialData.

packages/react-query/src/mutationOptions.ts
```diff
@@ -0,0 +1,15 @@
...
+export function mutationOptions<
+  TData = unknown,
+  TError = DefaultError,
+  TVariables = void,
+  TContext = unknown,
+>(
+  options: UseMutationOptions<TData, TError, TVariables, TContext>,
+): UseMutationOptions<TData, TError, TVariables, TContext>
+
+export function mutationOptions(options: unknown) {
```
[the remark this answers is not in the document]
Ubinquitous: Thank you, I changed it to a non-overloaded function.

docs/framework/react/typescript.md
```diff
@@ -239,6 +239,24 @@ const data = queryClient.getQueryData<Group[]>(['groups'])
...
+  const queryClient = useQueryClient()
+
+  return mutationOptions({
+    mutationKey: ['groups'],
+    mutationFn: executeGroups,
+    onSuccess: () => {
+      queryClient.invalidateQueries({ queryKey: ['posts'] })
+    },
+  })
+}
```
manudeli, line 257: It might be worth highlighting in the documentation that mutationOptions can be reused across different interfaces—such as useMutation, useIsMutating, and queryClient.isMutating.

```suggestion
function groupMutationOptions() {
  return mutationOptions({
    mutationKey: ['groups'],
    mutationFn: addGroup,
  })
}

useMutation({
  ...groupMutationOptions()
  onSuccess: () => queryClient.invalidateQueries({ queryKey: ['groups'] })
})
useIsMutating(groupMutationOptions())
queryClient.isMutating(groupMutationOptions())
```

manudeli (commented):
@TkDodo, @Ubinquitous I'd like to hear your thoughts on my review

manudeli (commented):
`mutationOptions` is needed to be exported from src/index.ts

packages/react-query/src/mutationOptions.ts
```diff
@@ -0,0 +1,13 @@
...
+export function mutationOptions<
+  TData = unknown,
+  TError = DefaultError,
+  TVariables = void,
+  TContext = unknown,
+>(
+  options: UseMutationOptions<TData, TError, TVariables, TContext>,
+): UseMutationOptions<TData, TError, TVariables, TContext> {
+  return options
+}
```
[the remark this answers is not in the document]
Ubinquitous: If mutationOptions is used not only in useMutation but also in other interfaces such as useIsMutating, I think it would be better to make it a required value.

manudeli:
Resolve eslint error please

Ubinquitous:
Sorry, I fix error that 'should not allow excess properties' test don't have assertions

andredewaard:
Really looking forward to this PR!

Its really helpful to add reusable optimistic updates, rollbacks etc callbacks for mutations.
But what if we want to use these reusable callbacks but also want a specific side effect in one component on onSucces, like on login I want to redirect.
Is there a way to merge these callbacks somehow?

Ubinquitous:
Thank you, @andredewaard.

I'm not sure if I understood it correctly, but it seems like this can basically be implemented using the spread operator. I’ve used a similar pattern when working with queryOptions as well. Is this what you were referring to?

```tsx
const options = mutationOptions({
  ...
})

useMutation({
  ...options,
  onSuccess: () => {},
})
```

If you're only changing the funnel in a specific component, I recommend wrapping it in a function for better management. I used to do it that way as well when working with `queryOptions`.

```jsx
export const postQueries = {
  all: () => ['post-list'],
  list: (): UseQueryOptions<Post[], Error, string> =>
    queryOptions({
      queryKey: [...postListQueries.all()],
      queryFn: () => getPostList(),
      select: (data) => getPostListString(data),
    }),
  detail: (type: PostType): UseSuspenseQueryOptions<PostInfo, Error> =>
    queryOptions({
      queryKey: [...postListQueries.all(), 'detail', type],
      queryFn: () => getPostDetail({ type }),
      select: (data) => data,
    }),
  ...
};
```

packages/react-query/src/mutationOptions.ts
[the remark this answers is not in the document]
TkDodo: yeah we can’t make it required; yes filters won’t work then; it’s one of the reasons why I’m against this helper in the first place 😅

andredewaard:
@Ubinquitous Thanks for your reply.
I meant to not override the logic happening in onSuccess if I add an onSuccess callback in the component.

So more like this.
```ts
useMutation({
  ...options,
  onSuccess: () => {
    options.onSuccess()
    // added logic
  },
})
```

TkDodo:
yes, with optional chaining on the options callback:

```
useMutation({
  ...options,
  onSuccess: () => {
    options?.onSuccess()
    // added logic
  },
})
```

Ubinquitous:
const compose = (...functions) => (...args) =>
	functions.forEach((fn) => fn?.(...args))

packages/react-query/src/__tests__/mutationOptions.test-d.tsx
```diff
@@ -0,0 +1,26 @@
...
```
manudeli, line 26: In my opinion, we should add more test cases before merging this Pull Request—assuming TkDodo agrees with the proposed interface. I'm not sure whether TkDodo will agree with this interface.

```tsx
const Example = () => {
  const mutation = useMutation({
    ...mutationOptions({
      mutationKey: ['key'],
      mutationFn: () => Promise.resolve({ field: 'test' })
    }),
    onSuccess: (data) => expectTypeOf(data).toEqualTypeOf<{ field: string }>()
  })
  expectTypeOf(mutation).toEqualTypeOf ...
}
```

packages/react-query/src/mutationOptions.ts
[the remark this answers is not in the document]
andredewaard: Can't we make mutationKey required? I know its not really used in useMutation often, but it forces us to write code that would work with any api.
I can think of developers wondering why `useIsMutating` isn't working when they forgot the key.

Or if we can't do this add a default key the api can fall back to if no key is provided?

packages/react-query/src/__tests__/mutationOptions.test-d.tsx
Ubinquitous: Thanks, I added more test cases along with the ones that work with useMutation.

Ubinquitous:
I find mutationOptions useful. Currently, I am using a style that manages queryKeys and mutationKeys by bundling them into javascript objects, and mutationOptions can reduce developers' type mistakes and boilerplate code when using it this way. In addition, it can be used with useMutation, useIsMutating, and queryClient.isMutating.

Just like queryOptions allows call invalidateQueries and use that variable, wouldn't mutationOptions do the same (at useMutation, useIsMutating), keeping the code clean?

```jsx
export const QUERY_KEY = (uniqueKey) => {
  return {
    get: [...queries.all(), 'get-key', uniqueKey],
    list: [...queries.all(), 'list-key', uniqueKey],
  };
};

export const queries = {
  all: () => ['query-key'],
  get: ({ some }) =>
    queryOptions({
      queryKey: QUERY_KEY(some).get,
      queryFn: () => ...,
    }),
  list: ({ some }) =>
    queryOptions({
      queryKey: QUERY_KEY(some).list,
      queryFn: () => ...,
    }),
  // this
  update: (): UseMutationOptions<
    null,
    Error,
    Request
  > => ({
    mutationFn: (req) => ...,
  }),
};
```

packages/react-query/src/mutationOptions.ts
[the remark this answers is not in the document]
TkDodo: that’s an interesting approach. I think we should try to make it `required` right now and see how it’s received. if there’s pushback we can either make it optional (non-breaking) or add the module augmentation override.

But maybe `required` is fine for most people 👍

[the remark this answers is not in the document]
manudeli: I agree 👍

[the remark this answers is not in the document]
Ubinquitous: I changed mutationKey to required along with updated test code and documentation.

packages/react-query/src/__tests__/mutationOptions.test-d.tsx
```diff
@@ -0,0 +1,121 @@
...
+        mutationKey: ['key'],
+        mutationFn: () => Promise.resolve<{ field: string }>({ field: 'test' }),
+      }),
+      onSuccess: (data) =>
+        expectTypeOf(data).toEqualTypeOf<{ field: string }>(),
+    })
+    expectTypeOf(mutation).toMatchTypeOf<
+      UseMutationResult<{ field: string }, DefaultError, void, unknown>
+    >()
+  })
```
manudeli: @Ubinquitous It seems better to add type test cases for `mutationOptions` not only with `useMutation`, but also with `useIsMutating`, `queryClient.isMutating`, etc. Being compatible with these interfaces is also an important feature of `mutationOptions` and should be tested IMO

Ubinquitous: Thanks! I added test code related to useIsMutating, queryClient.isMutating, and useMutationState to mutationOptions.test.tsx.

packages/react-query/src/mutationOptions.ts
```diff
@@ -0,0 +1,19 @@
...
+  TContext = unknown,
+>(
+  options: WithRequired<
+    UseMutationOptions<TData, TError, TVariables, TContext>,
+    'mutationKey'
+  >,
+): WithRequired<
+  UseMutationOptions<TData, TError, TVariables, TContext>,
+  'mutationKey'
+> {
```
Nick-Lucas: Just had a quick thread with @TkDodo [on bsky](https://bsky.app/profile/tkdodo.eu/post/3lr3jww3t422t) and thought I'd leave this as a possible tweak. How about these overloads to infer whether mutationKey is actually there at compile time? (might need tweaking a bit)

```suggestion
export function mutationOptions<
  TData = unknown,
  TError = DefaultError,
  TVariables = void,
  TContext = unknown,
>(
  options: WithRequired<
    UseMutationOptions<TData, TError, TVariables, TContext>,
    'mutationKey'
  >,
): WithRequired<
  UseMutationOptions<TData, TError, TVariables, TContext>,
  'mutationKey'
> {
export function mutationOptions<
  TData = unknown,
  TError = DefaultError,
  TVariables = void,
  TContext = unknown,
>(
  options: Omit<
    UseMutationOptions<TData, TError, TVariables, TContext>,
    'mutationKey'
  >,
): Omit<
  UseMutationOptions<TData, TError, TVariables, TContext>,
  'mutationKey'
> & { mutationKey?: undefined }
export function mutationOptions<
  TData = unknown,
  TError = DefaultError,
  TVariables = void,
  TContext = unknown,
>(
  options: UseMutationOptions<TData, TError, TVariables, TContext>,
): UseMutationOptions<TData, TError, TVariables, TContext> {
```

Ubinquitous: Thanks! I added type overloading!

andredewaard:
any ETA on this? would be really helpful in my current app.

docs/framework/react/typescript.md
```diff
@@ -239,6 +239,26 @@ const data = queryClient.getQueryData<Group[]>(['groups'])
...
+function groupMutationOptions() {
+  return mutationOptions({
+    mutationKey: ['groups'], // mutationKey is required
+    mutationFn: addGroup,
+  })
+}
+
+useMutation({
+  ...groupMutationOptions()
+  onSuccess: () => queryClient.invalidateQueries({ queryKey: ['groups'] })
```
TkDodo: I think this might be a bit misleading, as the `mutationKey` and the `queryKey` are the same here, so it could seem like they are related / have to be the same, but that’s not true.

docs/framework/react/typescript.md
```diff
@@ -239,6 +239,26 @@ const data = queryClient.getQueryData<Group[]>(['groups'])
...
 [//]: # 'Materials'
 
+## Typing Mutation Options
+
+Similarly to `queryOptions`, you can use `mutationOptions` to extract mutation options into a separate function:
+
+```ts
+function groupMutationOptions() {
+  return mutationOptions({
+    mutationKey: ['groups'], // mutationKey is required
```
TkDodo:
```suggestion
    mutationKey: ['addGroup'], // mutationKey is required
```

TkDodo (approved):

docs/framework/react/reference/mutationOptions.md
```diff
@@ -0,0 +1,15 @@
...
+```tsx
+mutationOptions({
+  mutationFn,
+  ...options,
+})
+```
+
+**Options**
+
+You can generally pass everything to `mutationOptions` that you can also pass to [`useMutation`](./useMutation.md).
```
TkDodo:
```suggestion
You can generally pass everything to `mutationOptions` that you can also pass to [`useMutation`](../useMutation.md).
```

docs/framework/react/typescript.md
```diff
@@ -239,6 +239,26 @@ const data = queryClient.getQueryData<Group[]>(['groups'])
...
 [//]: # 'Materials'
 
+## Typing Mutation Options
+
+Similarly to `queryOptions`, you can use `mutationOptions` to extract mutation options into a separate function:
+
+```ts
+function groupMutationOptions() {
+  return mutationOptions({
+    mutationKey: ['addGroup'], // mutationKey is required
```
manudeli: mutationKey is not required now

```suggestion
    mutationKey: ['addGroup'],
```

packages/react-query/src/__tests__/mutationOptions.test-d.tsx
```diff
@@ -0,0 +1,153 @@
...
+    const mutationOpts = mutationOptions({
+      mutationKey: ['key'],
+      mutationFn: () => Promise.resolve(5),
+    })
+
+    const mutationState = useMutationState({ filters: mutationOpts })
+    expectTypeOf(mutationState).toEqualTypeOf<
+      Array<MutationState<unknown, Error, unknown, unknown>>
+    >()
+  })
```
manudeli: image

@TkDodo  When I read [this bsky comment](https://bsky.app/profile/tkdodo.eu/post/3lr3jww3t422t), 

image

I thought below should make type error. but current implementation doesn't make type error

```suggestion
  it('should infer types when used with useMutationState', () => {
    const mutationState = useMutationState({
      filters: mutationOptions({
        mutationKey: ['key'],
        mutationFn: () => Promise.resolve(5),
      }),
    })
    expectTypeOf(mutationState).toEqualTypeOf<
      Array<MutationState<unknown, Error, unknown, unknown>>
    >()

    // @ts-expect-error this is a good error, because mutationOptions have no mutationKey
    useMutationState({
      filters: mutationOptions({ mutationFn: () => Promise.resolve(5) }),
    })
  })
```

image

manudeli: @TkDodo @Ubinquitous I changed mutationOptions's type in [f32a7e0](https://github.com/TanStack/query/pull/8960/commits/f32a7e04a5dbb91a4a2f6ecfd8de2726f8d0e178)

Then, we can avoid using mutationOptions without mutationKey with filters with TypeError
Double-check them out please

@Nick-Lucas Could you also review this please?

manudeli (approved):

packages/react-query/src/__tests__/mutationOptions.test-d.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  it('should allow mutationKey to be omitted', () => {
+    return mutationOptions({
+      mutationFn: () => Promise.resolve(123),
+      mutationKey: ['key'],
+      onSuccess: (data) => {
+        expectTypeOf(data).toEqualTypeOf<number>()
+      },
+    })
+  })
```
TkDodo: the test says `'should allow mutationKey to be omitted'`, but the implementation of the test passes a `mutationKey`, so it does the opposite ...

packages/react-query/src/__tests__/mutationOptions.test-d.tsx
```diff
@@ -0,0 +1,185 @@
...
+        mutationFn: () => Promise.resolve('data'),
+        onSuccess: (data) => {
+          expectTypeOf(data).toEqualTypeOf<string>()
+        },
+      }),
+    )
+    expectTypeOf(mutation).toEqualTypeOf<
+      UseMutationResult<string, DefaultError, void, unknown>
+    >()
+  })
```
TkDodo: we should have a test _without_ `mutationKey` here too, please (like we have below with `useIsMutating` and `useMutationState`)

manudeli:
@Nick-Lucas Thanks for idea suggestion to solve mutationKey optional problem! I added you as co-author in 2f7eb30

image

Ubinquitous:
I fixed the part where the test code was incorrect and added test code to check that useMutation works properly when there is no mutationKey.

packages/react-query/src/__tests__/mutationOptions.test-d.tsx
```diff
@@ -0,0 +1,194 @@
...
+          expectTypeOf(data).toEqualTypeOf<string>()
+        },
+      }),
+    )
+    expectTypeOf(mutation).toEqualTypeOf<
+      UseMutationResult<string, DefaultError, void, unknown>
+    >()
+
+    useMutation(
+      // @ts-check should allow when used with useMutation without mutationKey
```
TkDodo:
```suggestion
      // should allow when used with useMutation without mutationKey
```

TkDodo:
Thanks everyone who worked on this ❤️

binajmen:
is there a plan to port this to `solid-query`?

Issue #8046, opened 2024-09-12
Infinite query page param resets back to the first page during refetch if a page request is retried


parkyoung555:
### Describe the bug

I am using `useInfiniteQuery` to additively load more data into a table from a `nextToken` based list API (e.g. [example AWS SDK list API](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/sso/command/ListAccountsCommand)). Requesting next pages and refetching works great for the most part, but I'm having some trouble figuring out a way to preserve the next page param for a **retried** request triggered by a manual or automated `refetch`.

When fetching next pages with `fetchNextPage` and the `queryFn` rejects with an error response, the request is retried as expected using the same page params. However, I noticed retried requests from a manual or automated `refetch` will start over back to the first page after the first error response is encountered. I understand the need to fetch sequentially from the first page for the refetch, however, I would expect the failed request to retry using the same page params and continue onto the next page request once a retry succeeds.

The backend service I'm working with has a reasonable rate limit set (50 TPM in a 60 minute sliding window), but I'm trying to ensure a manual or automatic `refetch` will be able to recover from worst-case rate limiting scenarios where the user can continuously rate-limit themselves by refetching a table view loaded with 50+ pages.

Original discussion: https://github.com/TanStack/query/discussions/8044

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/mqvdz4

### Steps to reproduce

1. Open [code sandbox](https://codesandbox.io/p/sandbox/mqvdz4) and open browser console to view logs
2. Click the "Get next page" button 3 times to paginate through all the mock data. The second requested page is hard coded to simulate a rate limiting error and will be retried 3 times before succeeding.
3. After loading all the data, click the "Refetch" button to invoke query `refetch`.
4. You should be able to observe the page param being reset back to `0` during the refetch when the rate limiting error is encountered with page `2`. This will continue to loop until the number of retry attempts are exhausted.

Example console logs:

```
// fetch next page

getting page 0...
got page 0
getting page 1...
got page 1
getting page 2... // simulating rate limit for page index 2
rate limited!
getting page 2...
rate limited!
getting page 2...
rate limited!
getting page 2...
got page 2 // rate limit quota restored
getting page 3...
got page 3

// refetch

getting page 0...
got page 0
getting page 1...
got page 1
getting page 2...
rate limited!
getting page 0... // page param is reset to 0, expected page 2 to be retried here
got page 0
getting page 1...
got page 1
[...]

PR #8961, opened 2025-04-06
test(query-core): use fakeTimers for infiniteQueryBehavior.test.tsx

manudeli:


TkDodo:
this one seems to timeout:

```
   × InfiniteQueryBehavior > InfiniteQueryBehavior should not enter an infinite loop when a page errors while retry is on #8046 5007ms
     → Test timed out in 5000ms.
```

manudeli:
> this one seems to timeout:
> 
> ```
>    × InfiniteQueryBehavior > InfiniteQueryBehavior should not enter an infinite loop when a page errors while retry is on #8046 5007ms
>      → Test timed out in 5000ms.
> ```

I fixed it

PR #8966, opened 2025-04-07
feat(eslint): add type-checked configuration for eslint plugin

Newbie012:
Created a new configuration for type-checked rules in the ESLint plugin. Added a `flat/recommendedTypeChecked` config that includes the `no-void-query-fn` rule which requires TypeScript type information. Added an example project demonstrating the type-checked ESLint setup and updated the documentation to explain how to use it.


TkDodo:
> It seems like test:lib actually depends on build rather than compile

yes, we are aware and working on it.

TkDodo:
@Newbie012 I’ve temporarily set `build` as a dependency for `test:lib` to work around this problem.

TkDodo:
@Newbie012 seems like we forgot about this but I still think it’s a good idea. can we revive it?

Newbie012:
Sure
Will look into it and report back

docs/eslint/eslint-plugin-query.md
```diff
@@ -46,6 +46,21 @@ export default [
...
+### Recommended type-checked setup
+
+If you're using TypeScript and want to enable rules that require type information, you can use the `flat/recommendedTypeChecked` config:
+
+```js
+import pluginQuery from '@tanstack/eslint-plugin-query'
+
+export default [
+  ...pluginQuery.configs['flat/recommendedTypeChecked'],
+  // Any other config...
```
Hardanish-Singh, line 58: [nit]: Any other `configs`

PR #8971, opened 2025-04-07
fix(eslint): classic build

Newbie012:
This PR ensures compatibility with classic (legacy) eslint plugin cjs build.

### DIFF
--- a/packages/eslint-plugin-query/tsup.config.ts
+++ b/packages/eslint-plugin-query/tsup.config.ts
@@ -18,4 +18,11 @@ export default defineConfig({
   sourcemap: true,
   clean: true,
   external: ['typescript'],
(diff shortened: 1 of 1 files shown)


Newbie012:
@TkDodo we should probably add a legacy config example which the CI should lint.

ptts:
🤔 I am still getting 
`Module '"/Users/ptts/dev/project/node_modules/@tanstack/eslint-plugin-query/dist/index"' has no default export. ts(1192)` on `"@tanstack/eslint-plugin-query": "5.72.1"`.

I tried reinstalling multiple times but the error persists and disappears when I downgrade to `5.71.5`.

Could it be on my end or does anyone else still have the same problem?

PR #8972, opened 2025-04-07
fix(eslint-plugin) separate legacy and modern dts files

Newbie012:


Newbie012:
@TkDodo It seems like there's something off with NX cache

TkDodo:
the issue appears locally on this PR as well (try running `pnpm run build:all`) but it doesn’t appear on main, so I think there’s something wrong on this PR

examples/react/eslint-legacy/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@tanstack/query-example-react-basic",
```
TkDodo, line 2: I think this is the problem: multiple examples with the same name!

TkDodo: fixed that here: c836a9533464791f9f6f6204a9ff891d6314eb9f

TkDodo:
there’s a legit error now:

```
root.tsup.config.js(3,1): error TS6133: 'esbuildPluginFilePathExtensions' is declared but its value is never read.
```

examples/react/eslint-legacy/package.json
Newbie012: Oh thanks 🤦

examples/react/eslint-legacy/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@tanstack/query-example-eslint-legacy",
+  "private": true,
+  "type": "module",
+  "scripts": {
```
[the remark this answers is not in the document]
Newbie012: done, but it seems like something is off with packages that I didn't touch 😅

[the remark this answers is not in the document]
TkDodo: yeah sadly, some tests are flaky. I’ll ship it

thgreasi:
Hi,
I think that this changed the result of `require('@tanstack/eslint-plugin-query')` from
```
{
  meta: { name: '@tanstack/eslint-plugin-query' },
  configs: {
    recommended: { plugins: [Array], rules: [Object] },
    'flat/recommended': [ [Object] ]
  },
  rules: {
    'exhaustive-deps': { create: [Function: create], defaultOptions: [], meta: [Object] },
    'stable-query-client': { create: [Function: create], defaultOptions: [], meta: [Object] },
    'no-rest-destructuring': { create: [Function: create], defaultOptions: [], meta: [Object] },
    'no-unstable-deps': { create: [Function: create], defaultOptions: [], meta: [Object] },
    'infinite-query-property-order': { create: [Function: create], defaultOptions: [], meta: [Object] },
    'no-void-query-fn': { create: [Function: create], defaultOptions: [], meta: [Object] }
  }
}
```
to
```
{
  default: {
    meta: { name: '@tanstack/eslint-plugin-query' },
    configs: {
      recommended: { plugins: [Array], rules: [Object] },
      'flat/recommended': [ [Object] ]
    },
    rules: {
      'exhaustive-deps': { create: [Function: create], defaultOptions: [], meta: [Object] },
      'stable-query-client': { create: [Function: create], defaultOptions: [], meta: [Object] },
      'no-rest-destructuring': { create: [Function: create], defaultOptions: [], meta: [Object] },
      'no-unstable-deps': { create: [Function: create], defaultOptions: [], meta: [Object] },
      'infinite-query-property-order': { create: [Function: create], defaultOptions: [], meta: [Object] },
      'no-void-query-fn': { create: [Function: create], defaultOptions: [], meta: [Object] }
    }
  },
  plugin: {
    meta: { name: '@tanstack/eslint-plugin-query' },
    configs: {
      recommended: { plugins: [Array], rules: [Object] },
      'flat/recommended': [ [Object] ]
    },
    rules: {
      'exhaustive-deps': { create: [Function: create], defaultOptions: [], meta: [Object] },
      'stable-query-client': { create: [Function: create], defaultOptions: [], meta: [Object] },
      'no-rest-destructuring': { create: [Function: create], defaultOptions: [], meta: [Object] },
      'no-unstable-deps': { create: [Function: create], defaultOptions: [], meta: [Object] },
      'infinite-query-property-order': { create: [Function: create], defaultOptions: [], meta: [Object] },
      'no-void-query-fn': { create: [Function: create], defaultOptions: [], meta: [Object] }
    }
  }
}
```

I realize that the suggested way to import the package is
```
import pluginQuery from '@tanstack/eslint-plugin-query'
```
which if someone is using the TypeScript import helper, it will probably not be an issue at all, but I wanted to raise anyway that the result of the plain commonjs `require()` calls have changed, just in case it was missed.

PR #8989, opened 2025-04-10
refactor(query-core): remove unused common type 

novice0840:
I change a type a little. 

# Optional

I change `OmitKeyof` to `Omit`. Since the third parameter TStrictly in OmitKeyof is always fixed, there's no need to use OmitKeyof in this case.


TkDodo:
The whole `Optional` type is unused and can be removed.

novice0840:
Yes, you are right. type `Optional` is not used. So I removed it

PR #8990, opened 2025-04-10
fix(query-devtools): move solid to deps

birkskyum:
move solid (which the agnostic devtools are built with) to deps, like it is in the router devtools:

- https://github.com/TanStack/router/pull/3732


TkDodo:
does that mean if you use the `react-query-devtools`, you have a runtime dependency on solid ?

PR #9007, opened 2025-04-12
fix(query-core): make sure we don't invoke select too often when using placeholderData

TkDodo:


packages/query-core/src/__tests__/queryObserver.test.tsx
```diff
@@ -925,42 +925,6 @@ describe('queryObserver', () => {
...
-      placeholderData: placeholderData1,
-    })
-
-    expect(observer.getCurrentResult().isPlaceholderData).toBe(true)
-
-    observer.setOptions({
-      queryKey: key,
-      queryFn: () => data,
-      // @ts-expect-error
-      select: () => undefined,
```
TkDodo, line 957: note: this test was quite weird to begin with because it’s not allowed to return undefined from select 🤷

PR #9011, opened 2025-04-13
chore: add @typescript-eslint/require-await for notice error and remove unnecessary async

minseong0324:


TkDodo:
since we want this rule everywhere, please enable it in `root.eslint.config.js`

minseong0324:
> since we want this rule everywhere, please enable it in `root.eslint.config.js`

@TkDodo If you're talking about the entire `query` monorepo, can we proceed with one PR including removing unnecessary async keyword and adding lint rule? Because the context is distributed to upload to a separate PR, creating a bottleneck in the work.

minseong0324:
@TkDodo If you're referring to the whole query monorepo, would it be okay to proceed with a single PR that removes the unnecessary async keyword and adds the corresponding lint rule? I was thinking that combining them might help reduce the overhead from having to split and manage multiple PRs for related changes.

TkDodo:
@minseong0324 yes, you can add multiple PRs that will work towards that eslint rule, but please only one PR at the very end that enables the eslint rule in the root config. I don’t want this setting set to true in multiple eslint configs.

The eslint configs in each package are only meant for really package specific overrides. For example, the react lint rules only make sense in react lated packages. But something that we want everywhere should be enabled in the root config.

PR #9014, opened 2025-04-14
fix(react-query): allow null type for result ref in useMutationState

umsungjun:
The result ref is initialized with null, but its type does not reflect null, causing issues in TypeScript projects with --strictNullChecks enabled. This commit updates the type to resolve the problem.

### DIFF
--- a/packages/react-query/src/useMutationState.ts
+++ b/packages/react-query/src/useMutationState.ts
@@ -47,7 +47,7 @@ export function useMutationState<TResult = MutationState>(
 ): Array<TResult> {
   const mutationCache = useQueryClient(queryClient).getMutationCache()
   const optionsRef = React.useRef(options)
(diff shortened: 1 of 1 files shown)


TkDodo:
which version of react are you on?

umsungjun:
Probably React 18

umsungjun:
I did not test it directly, but I believe that adding null to the type is the correct solution, so I included it.

TkDodo:
there are type errors now so I don’t think this is corect:

```
src/useMutationState.ts(59,3): error TS2322: Type '(TResult | null)[]' is not assignable to type 'TResult[]'.
  Type 'TResult | null' is not assignable to type 'TResult'.
    'TResult' could be instantiated with an arbitrary type which could be unrelated to 'TResult | null'.
```

if you have a problem, please file an issue with a reproduction rather than a PR. Thank you

PR #9029, opened 2025-04-15
chore: add typescript-eslint/require-await rule

minseong0324:
ref: https://github.com/TanStack/query/pull/9011#issuecomment-2804846344


eslint.config.js
```diff
@@ -48,6 +48,7 @@ export default [
     rules: {
       ...vitest.configs.recommended.rules,
       'vitest/expect-expect': 'warn',
+      '@typescript-eslint/require-await': 'error',
```
TkDodo, line 51: can we enable this for all files, not just tests ?

minseong0324: Ok I'll fix it😊

minseong0324: @TkDodo I would like to ask your opinion on what you think is the best way to solve this problem. 🙏🏻

image

minseong0324: And I think we can decide whether to merge this PR after https://github.com/TanStack/query/pull/9039
 becomes a review first. There was something I missed in the task of removing unnecessary async.

eslint.config.js
```diff
@@ -40,6 +40,7 @@ export default [
       '@typescript-eslint/no-empty-function': 'off',
       '@typescript-eslint/no-unsafe-function-type': 'off',
       'no-case-declarations': 'off',
+      '@typescript-eslint/require-await': 'error',
```
TkDodo, line 43:
```suggestion
      '@typescript-eslint/require-await': 'error',
      'no-case-declarations': 'off',
```

TkDodo:
seems like there are still some violations left (or were added in the meantime). Can you fix them in this PR ?

minseong0324:
> seems like there are still some violations left (or were added in the meantime). Can you fix them in this PR ?

Sure, I’ll take care of it!

manudeli:
@minseong0324 Resolve ci error please

image

minseong0324:
> @minseong0324 Resolve ci error please


From [this point](https://github.com/TanStack/query/pull/9029#discussion_r2050841556) on, I've been figuring out the CI problem in `query-codemods`, but I haven't found a solution. Can you give me a little more time? Or, your opinion is also welcome :)

manudeli:
> > @minseong0324 Resolve ci error please
> 
> From [this point](https://github.com/TanStack/query/pull/9029#discussion_r2050841556) on, I've been figuring out the CI problem in `query-codemods`, but I haven't found a solution. Can you give me a little more time? Or, your opinion is also welcome :)

Hi @lachlancollins, I noticed that this ESLint error only appears in the @tanstack/query-codemods package, and not elsewhere. Since you seem to be the most knowledgeable about the ESLint setup with @tanstack/config, would you mind helping us figure out?

I've double-checked the configuration and it seems consistent with other packages, so I'm not sure why the issue is isolated to this one. 🥲

manudeli (approved):
It's time to add this rule finally

PR #9030, opened 2025-04-16
fix(query-core, vue-query): fix type inference in `setQueryData` when used in functions with explicit return types

### LINKED ISSUE #9024: `setQueryData` updater function's parameter incorrectly typed as `unknown` in expression body arrow functions
### Describe the bug

When using `queryClient.setQueryData` inside an expression body arrow function (without curly braces), the updater function's data parameter is incorrectly typed as `unknown`.

However, when used in a block body arrow function (with curly braces), the type is correctly inferred as expected.

https://stackblitz.com/edit/tanstack-query-wcdibka8?file=src%2Findex.tsx

### Steps to reproduce

1.  [Visit the StackBlitz reproduction](https://stackblitz.com/edit/tanstack-query-wcdibka8?file=src%2Findex.tsx)
2. Check the type inference on both mutation1 and mutation2 examples
3. Observe that in mutation1 (expression body), the data parameter is typed as unknown
4. Observe that in mutation2 (block body), the data parameter is correctly typed as { data: string } | undefined

### Expected behavior

The updater function's `data` parameter should be correctly typed in both cases, regardless of whether the arrow function uses an expression body or block body syntax.

Every time

### Screenshots or Videos

_No response_

### Platform

- MacOS
- I don't think this issue is related to what platform I use.

### Tanstack Query adapter

None

### TanStack Query version

v5.74.3

### TypeScript version

v5.8.3

### Additional context

The `setQueryData` updater function's [`NoInfer` type used](https://github.com/TanStack/query/blob/main/packages/query-core/src/queryClient.ts#L184-L185) for passing [this test](https://github.com/TanStack/query/blob/main/packages/query-core/src/__tests__/queryClient.test-d.tsx#L90-L96) appears to be the cause of the problem.
I am looking into it and once I find a solution, I'll submit a PR.

custardcream98:
Applied the `NoInfer` utility type to the return type of `setQueryData` to prevent the outer function's return type from affecting the updater function's type inference.

closes #9024

### DIFF
--- a/packages/query-core/src/__tests__/queryClient.test-d.tsx
+++ b/packages/query-core/src/__tests__/queryClient.test-d.tsx
@@ -113,6 +113,19 @@ describe('setQueryData', () => {
 
     expectTypeOf(data).toEqualTypeOf<string | undefined>()
   })
+
+  it('should preserve updater parameter type inference when used in functions with explicit return types', () => {
+    const queryKey = ['key'] as DataTag<Array<string>, number>
+    const queryClient = new QueryClient()
+
+    // Simulate usage inside a function with explicit return type
+    // The outer function returns 'unknown' but this shouldn't affect the updater's type inference
+    ;(() =>
+      queryClient.setQueryData(queryKey, (data) => {
+        expectTypeOf(data).toEqualTypeOf<number | undefined>()
+        return data
+      })) satisfies () => unknown
+  })
 })
 
 describe('getQueryState', () => {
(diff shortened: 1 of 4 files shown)


custardcream98:
The CI test failure seems unrelated to the changes in this PR. My changes only affect TypeScript type definitions and a type test, with no runtime behavior modifications. The failure appears to be in an unrelated test.

TkDodo (approved):

PR #9033, opened 2025-04-16
Fix broken links on QueryCache reference page

emma-sg:
It looks like the `docs` part of several urls was being removed, so this PR moves the relative link one level up by removing one `../` in those docs. As requested, this also adds `.md` extensions.


TkDodo (commented):
there are more wrong links on this page, could you fix them too, please?

emma-sg:
> there are more wrong links on this page, could you fix them too, please?

Yeah! Found a couple more here.

PR #9040, opened 2025-04-18
refactor(query-async-storage-persister): remove unnecessary export "AsyncThrottleOptions"

minseong0324:
AsyncThrottleOptions are only used within `asyncThrottle.ts`, so I decided that it would be better not to export them.

### DIFF
--- a/packages/query-async-storage-persister/src/asyncThrottle.ts
+++ b/packages/query-async-storage-persister/src/asyncThrottle.ts
@@ -1,6 +1,6 @@
 import { noop } from './utils'
 
-export interface AsyncThrottleOptions {
+interface AsyncThrottleOptions {
   interval?: number
   onError?: (error: unknown) => void
 }


TkDodo:
`knip` should find this. The fact that it doesn’t means something is setup wrongly with knip :/

minseong0324:
> `knip` should find this. The fact that it doesn’t means something is setup wrongly with knip :/

@TkDodo Got it! I’ll take a look at the current `knip` setup to see if it’s configured properly. If I find any misconfigurations or areas for improvement, I’ll follow up with a PR to fix it.

minseong0324:
> `knip` should find this. The fact that it doesn’t means something is setup wrongly with knip :/

@TkDodo Since knip didn’t catch these unused exports, it seems like there might be a config issue.
Would it make sense to open an issue to track this?

TkDodo:
yeah, if you want to take a look at it, please do 🙏

PR #9075, opened 2025-04-27
test(react-query): use test-d file for useMutationState

manudeli:


packages/react-query/src/__tests__/utils.tsx
```diff
@@ -92,5 +92,3 @@ export function setIsServer(isServer: boolean) {
     })
   }
 }
-
-export const doNotExecute = (_func: () => void) => true
```
manudeli, line 96: We can remove this

PR #9077, opened 2025-04-28
fix(vue-query): fix initialData Type Inference Bug in useQuery

### LINKED ISSUE #9069: [vue-query] Type error returning undefined from `initialData` in `queryOptions`
### Describe the bug

In vue-query if `initialData` function returns undefined typescript error is displayed. 

According to the docs this should be allowed https://tanstack.com/query/latest/docs/framework/react/guides/initial-query-data#conditional-initial-data-from-cache and query should fetch from hard loading state instead.

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgRwK4FMoE8DyYbAQB2AznAL5wBmUEIcARAAIwCGpbAxgNYD0AbhgC0aTFgYBuAFBTgRGJiqtO6OAGFiVYAHNEUuHH6sANhgBccEjChzt08jKqoinAsTjb0MDUS3aAFACUegZQXqhQRHAACrQgwCToAHRhJBDG-Oj+SEam6BYA5AAW6MbGEAUUcKxkPn6B9jIAJuicxqxhHuUARiYhcHIKUEoqcACSRMAEJgAirGz9BpyaOgD8FnU60gYO+oYdcAD6h3JTwLPzrBYTZxdsjVLLHHAQ+ISkcAC8KBjYeG6kbJ7UTYADS6CwFgA2gwnn4GABdapkJ5WAA0wN+WAAYkQLJ5vCttBiDKdpsY5mwLEEvgA+RZwMIwCJRbQ9EwAFSKCSSxzJ5wplyScK2e3IGPIDSAA

### Steps to reproduce

1. Go to minimal repro typescript playground
2. See type error for `initialData`

```
  Overload 1 of 2, '(options: UndefinedInitialQueryOptions<Config, Error, Config, readonly ["config"]>): UndefinedInitialQueryOptions<Config, Error, Config, readonly [...]> & { ...; }', gave the following error.
    Type '() => Config | undefined' is not assignable to type 'undefined'.
  Overload 2 of 2, '(options: DefinedInitialQueryOptions<Config, Error, Config, readonly ["config"]>): DefinedInitialQueryOptions<Config, Error, Config, readonly [...]> & { ...; }', gave the following error.
    Type '() => Config | undefined' is not assignable to type 'Config | (() => Config) | (MaybeRefDeep<Config | InitialDataFunction<Config> | undefined> & (Config | (() => Config)))'.
      Type '() => Config | undefined' is not assignable to type '() => Config'.
        Type 'Config | undefined' is not assignable to type 'Config'.
          Type 'undefined' is not assignable to type 'Config'.(2769)
hydration-BlEK5ylC.d.ts(606, 5): The expected type comes from property 'initialData' which is declared here on type 'UndefinedInitialQueryOptions<Config, Error, Config, readonly ["config"]>'
hydration-BlEK5ylC.d.ts(606, 5): The expected type comes from property 'initialData' which is declared here on type 'DefinedInitialQueryOptions<Config, Error, Config, readonly ["config"]>'
```

### Expected behavior

`initialData` allows to return undefined. No type errors produced.

Every time

### Screenshots or Videos

_No response_

### Platform

N/A

### Tanstack Query adapter

vue-query

### TanStack Query version

v5.74.6

### TypeScript version

v5.8.3

### Additional context

Query still fetches just fine from hard loading state so the logic is correct. Also no type errors in `react-query` for the exact same code.

ss0526100:
Overview

In the previous PR(#9073), there was a bug where data was incorrectly inferred as undefined even when initialData was defined. This PR ensures that when initialData is TData, data is correctly inferred as TData, and adds tests to verify this behavior.

Changes

- Bug Fix: Improved type inference for non-undefined initialData
  - Modified to ensure data.value is accurately inferred as TData when initialData is provided with a non-undefined value.
  - Commit: 6e73779

- Test Addition: Type validation for defined initialData
  - Added tests to verify that data is returned as TData type only when initialData is defined.
  - Commit: 2ff92c5ebbc1331ea62a6222e70219c591078d0e, 7684df1

Related Files

- packages/vue-query/src/useQuery.ts
- packages/vue-query/src/__tests__/useQuery.test-d.ts
- packages/vue-query/src/__tests__/useQueries.test-d.ts

closes #9069

### DIFF
--- a/packages/vue-query/src/__tests__/useQueries.test-d.ts
+++ b/packages/vue-query/src/__tests__/useQueries.test-d.ts
@@ -35,7 +35,7 @@ describe('UseQueries config object overload', () => {
     })
 
     expectTypeOf(queriesState[0].data).toEqualTypeOf<{ wow: boolean }>()
-    expectTypeOf(queriesState[1].data).toEqualTypeOf<string | undefined>()
+    expectTypeOf(queriesState[1].data).toEqualTypeOf<string>()
     expectTypeOf(queriesState[2].data).toEqualTypeOf<string | undefined>()
   })
 
@@ -54,9 +54,7 @@ describe('UseQueries config object overload', () => {
 
     const { value: queriesState } = useQueries({ queries: [options] })
 
-    expectTypeOf(queriesState[0].data).toEqualTypeOf<
-      { wow: boolean } | undefined
-    >()
+    expectTypeOf(queriesState[0].data).toEqualTypeOf<{ wow: boolean }>()
   })
 
   it('should be possible to define a different TData than TQueryFnData using select with queryOptions spread into useQueries', () => {
--- a/packages/vue-query/src/__tests__/useQuery.test-d.ts
+++ b/packages/vue-query/src/__tests__/useQuery.test-d.ts
@@ -23,7 +23,7 @@ describe('useQuery', () => {
         }),
       )
 
-      expectTypeOf(data).toEqualTypeOf<{ wow: boolean } | undefined>()
+      expectTypeOf(data).toEqualTypeOf<{ wow: boolean }>()
     })
 
     it('TData should be defined when passed through queryOptions', () => {
@@ -40,7 +40,7 @@ describe('useQuery', () => {
       })
       const { data } = reactive(useQuery(options))
 
-      expectTypeOf(data).toEqualTypeOf<{ wow: boolean } | undefined>()
+      expectTypeOf(data).toEqualTypeOf<{ wow: boolean }>()
     })
 
     it('should be possible to define a different TData than TQueryFnData using select with queryOptions spread into useQuery', () => {
--- a/packages/vue-query/src/__tests__/useQuery.test-d.ts
+++ b/packages/vue-query/src/__tests__/useQuery.test-d.ts
@@ -74,7 +74,7 @@ describe('useQuery', () => {
         }),
       )
 
-      expectTypeOf(data).toEqualTypeOf<{ wow: boolean } | undefined>()
+      expectTypeOf(data).toEqualTypeOf<{ wow: boolean }>()
     })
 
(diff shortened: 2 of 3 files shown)


saul-atomrigs:
I think we also need to rollback this test case:
https://github.com/TanStack/query/blob/b397fba936a38d55e7609ac0e8562eeba16998ec/packages/vue-query/src/__tests__/useQuery.test-d.ts#L62-L78

ss0526100:
@saul-atomrigs 

As I understand it, when initialData is defined, the type of data should be determined.

In the rollback case you suggested, it seems that even though initialData is defined, it returns an undefined type.

Please explain why this test case should have an undefined value. I don't understand well :(

---

A test code with the same description as the test for which you suggested a rollback also exists in the React Query tests, and the expected value is defined.

https://github.com/TanStack/query/blob/b397fba936a38d55e7609ac0e8562eeba16998ec/packages/react-query/src/__tests__/useQuery.test-d.tsx#L50-L64

saul-atomrigs:
It should NOT have `undefined`. We should remove it, like in react-query

ss0526100:
@saul-atomrigs 

Sorry, I misunderstood that my PR was incorrect.

I've recognized the issue and completed the fix. I'll push the commit.

romansp:
Thank you for working on the fix. 

Sorry if I'm missing something but wouldn't it be more correct to add a new test case that will cover minimal repro I provided in the original issue #9069?

Like literally add a case for `useQuery` or `queryOptions` usage where `initialData` is set to a function which returns `TData | undefined`. And this should not produce any typescript errors. I just don't see this usage being covered by existing test cases.

ss0526100:
@romansp

I understand.

I also realized that the issue from #9069 causes an error in the current version.

I'll create and submit a PR for this and tag you.

PR #9088, opened 2025-04-29
fix(vue-query): move `queryOptions` and `useQuery` function overloads 

### LINKED ISSUE #9069: [vue-query] Type error returning undefined from `initialData` in `queryOptions`

romansp:
Fixes #9069. Unfortunately solutions from #9073 and #9077 didn't fully resolve the original issue.

I looked at what `react-query` does to ensure correct type safety for 4 different cases of `initialData`. These are: `undefined`, `T`, `() => T`, `() => T | undefined` and how come we're seeing troubles with type inference in `vue-query`.

So what I noticed is that react version declares overloads of `useQuery` and `queryOptions` where `initialData` is defined **_before_** the ones where `initialData` is undefined. [But overload ordering matters](https://github.com/microsoft/TypeScript/issues/1860#issuecomment-72154737). In a nutshell, more specific overload should come before a more general one. Vue was doing the opposite where defined `initialData` overloads are declared later which caused TypeScript to pick the wrong overload instead. That was causing both forbidding of `undefined` in function returns and leaving `undefined` in type union for `data`.

So I changed type declarations to align with `react-query` types. And it looks to be working as expected now.

I also copied `initialData` type test cases from `useQuery` into `queryOptions` to verify that when query is using `queryOptions` API then `data` types are still inferred correctly.

### DIFF
--- a/packages/vue-query/src/__tests__/queryOptions.test-d.ts
+++ b/packages/vue-query/src/__tests__/queryOptions.test-d.ts
@@ -125,17 +125,19 @@ describe('queryOptions', () => {
 
   it('TData should always be defined when initialData is provided as a function which ALWAYS returns the data', () => {
     const { data } = reactive(
-      useQuery(queryOptions({
-        queryKey: ['key'],
-        queryFn: () => {
-          return {
+      useQuery(
+        queryOptions({
+          queryKey: ['key'],
+          queryFn: () => {
+            return {
+              wow: true,
+            }
+          },
+          initialData: () => ({
             wow: true,
-          }
-        },
-        initialData: () => ({
-          wow: true,
+          }),
         }),
-      })),
+      ),
     )
 
     expectTypeOf(data).toEqualTypeOf<{ wow: boolean }>()
@@ -143,14 +145,16 @@ describe('queryOptions', () => {
 
   it('TData should have undefined in the union when initialData is NOT provided', () => {
     const { data } = reactive(
-      useQuery(queryOptions({
-        queryKey: ['key'],
-        queryFn: () => {
-          return {
-            wow: true,
-          }
-        },
-      })),
+      useQuery(
+        queryOptions({
+          queryKey: ['key'],
+          queryFn: () => {
+            return {
+              wow: true,
+            }
+          },
+        }),
+      ),
     )
 
     expectTypeOf(data).toEqualTypeOf<{ wow: boolean } | undefined>()
@@ -158,23 +162,8 @@ describe('queryOptions', () => {
 
   it('TData should have undefined in the union when initialData is provided as a function which can return undefined', () => {
     const { data } = reactive(
-      useQuery(queryOptions({
-        queryKey: ['key'],
-        queryFn: () => {
-          return {
-            wow: true,
-          }
-        },
-        initialData: () => undefined as { wow: boolean } | undefined,
-      })),
-    )
-
-    expectTypeOf(data).toEqualTypeOf<{ wow: boolean } | undefined>()
-  })
-
-  it('TData should be narrowed after an isSuccess check when initialData is provided as a function which can return undefined', () => {
-    const { data, isSuccess } = reactive(
-      useQuery(queryOptions({
+      useQuery(
+        queryOptions({
           queryKey: ['key'],
           queryFn: () => {
             return {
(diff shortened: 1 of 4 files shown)


ss0526100:
Could you check out [this branch](https://github.com/ss0526100/query/tree/test/vue-query-initialData) and take a look at `useQuery.test-d.ts` in `vue-query`?

In this branch, I tested all possible combinations of `initialData`.
However, no issues were found.

From what I can tell, the problem seems to be in `queryOptions`.

romansp:
@ss0526100 Yes, if I understand correctly the issue comes specifically from combining `queryOptions` with `useQuery`. 

Because there are 2 layers of function type overloads these should be declared in correct order otherwise `DefinedInitialQueryOptions` by `queryOptions` maps to `UndefinedInitialQueryOptions` in `useQuery` and vice-versa.

ss0526100:
I implemented the test for queryOptions as well and identified issues.

I rebased to the branch suggested in the PR and applied it to queryOptions and useQuery.
Refer to [queryOptions.test-d.ts](https://github.com/ss0526100/query/blob/fix/vue-query-initial-type-error/packages/vue-query/src/__tests__/queryOptions.test-d.ts) in the [this branch](https://github.com/ss0526100/query/tree/fix/vue-query-initial-type-error).
(Note: Do not include the tests from this branch, as they may significantly slow down CI/CD.)

The following issues were found:

queryOptions:
1. When initialData is an optional property, it cannot be assigned to the argument.
2. When initialData is a required property and is undefined, () => undefined, TData | undefined, or () => TData | undefined, it cannot be assigned to the argument.

PR #9092, opened 2025-05-01
feat: refetchMode replace for streamedQuery

TkDodo:


packages/query-core/src/streamedQuery.ts
```diff
@@ -6,46 +6,63 @@ import type { QueryFunction, QueryFunctionContext, QueryKey } from './types'
...
-          return prev.concat(chunk)
-        },
-      )
+
+      // don't append to the cache directly when replace-refetching
+      if (!isRefetch || refetchMode !== 'replace') {
+        context.client.setQueryData<Array<TQueryFnData>>(
+          context.queryKey,
+          (prev = []) => {
+            return prev.concat([chunk])
```
TkDodo, line 54:
```diff
- return prev.concat(chunk)
+ return prev.concat([chunk])
```

this tiny change is hard to see, but it fixes an issue where `TQueryFnData` is in itself an array, which got flattened implicitly by `concat`

TkDodo: added a test for this, too: 0de02229

Issue #9020, opened 2025-04-15
useMutation skips pending state when triggered in constructor or ngOnInit


ThiloAschebrock:
### Describe the bug

`useMutation` does indicate a pending state and remains idle until the mutation concludes if triggered in a Component constructor or `ngOnInit` lifecycle method, whereas triggering the mutation from `ngAfterContentInit` works as expected.

### Your minimal, reproducible example

https://stackblitz.com/edit/angular-query-aqv6yuzm?file=src%2Fmain.ts

### Steps to reproduce

1. Refresh the page.
2. Observe the state changes for the three mutations triggered in an Angular `constructor`, `ngOnInit`, or `ngAfterContentInit`.
3. Notice that the two mutations triggered in the `constructor` and `ngOnInit` remain idle and do not indicate the pending state while the mutations are running.

### Expected behavior

As a user, I expect the mutation to always indicate the correct idle and pending state, i.e., switch from idle to pending when a mutation has been triggered.

### How often does this bug happen?

Every time

### Platform

- OS: Windows, maxOS,
- Browser: Chrome, Edge

### Tanstack Query adapter

angular-query

### TanStack Query version

v5.74.3

### TypeScript version

v5.6.3

benjavicente:
We started to release a RC version of the stable package that solves this issue. It requires Angular 20.1 and up, upgrading Angular and the adapter fixes the issue shown in the reproduction.

```sh
npm i @tanstack/angular-query@rc
```

It has some breaking changes. See the [migration guide](https://github.com/TanStack/query/blob/angular-adapter-pre/docs/framework/angular/guides/migrating-from-experimental.md). If you have any question or issue, please let us knwon in here or the [Angular Channel in Discord](https://discord.com/channels/719702312431386674/1187361780998078515).

PR #9098, opened 2025-05-03
fix(angular-query): ensure initial mutation pending state is emitted

ThiloAschebrock:
Fixes https://github.com/TanStack/query/issues/9020

The issue was that the effect that subscribes to the observer runs after the observer emits the first state change if the mutation is triggered in the `constructor` of a component or `ngOnInit`. This is fixed by immediately subscribing when the initial result is computed.

One could also consider combining `resultFromSubscriberSignal` and `resultFromInitialOptionsSignal` into a single `linkedSignal`, but this would require dropping support for Angular 18.

### DIFF
--- a/packages/angular-query-experimental/src/__tests__/inject-mutation.test.ts
+++ b/packages/angular-query-experimental/src/__tests__/inject-mutation.test.ts
@@ -62,8 +62,6 @@ describe('injectMutation', () => {
       }))
     })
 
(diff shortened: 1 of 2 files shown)


ThiloAschebrock:
Hi @arnoud-dv,
resolved conflicts and aligned my changes with your approach of mimicking linked signal of #9808.

I would appreciate if you find some time to review my proposed changes.

PR #9105, opened 2025-05-05
fix(types): useSuspenseQuery should type-narrow the Error field

TkDodo:


packages/react-query/src/types.ts
```diff
@@ -155,7 +156,7 @@ export type UseQueryResult<
 export type UseSuspenseQueryResult<
   TData = unknown,
   TError = DefaultError,
-> = OmitKeyof<
+> = DistributiveOmit<
```
TkDodo, line 159: @manudeli I think your `OmitKeyof` helper doesn’t work over union types (just like the build-in `Omit` also doesn’t work with it).

Because `DefinedQueryObserverResult` is a union of:

```
| QueryObserverRefetchErrorResult<TData, TError>
| QueryObserverSuccessResult<TData, TError>
```

it “doesn’t work” in a sense that type narrowing to `error` doesn’t happen anymore because it’s not distributive.

I’ve added the standard `DistributiveOmit` helper and things work fine, but if you can improve your helper type to not fail the added test case, we can switch back.

packages/query-core/src/types.ts
```diff
@@ -9,6 +9,10 @@ import type { QueryFilters, QueryTypeFilter, SkipToken } from './utils'
 import type { QueryCache } from './queryCache'
 import type { MutationCache } from './mutationCache'
 
+export type DistributiveOmit<TKey, TValue> = TKey extends any
+  ? Pick<TKey, Exclude<keyof TKey, TValue>>
+  : never
```
manudeli, line 14: it's not a big deal, but how about below naming like below?

```suggestion
export type DistributiveOmit<TObject, TKey> = TObject extends any
  ? Pick<TObject, Exclude<keyof TObject, TKey>>
  : never
```

TkDodo: Yeah my bad, that's better 😂

manudeli:
https://github.com/TanStack/query/blob/89846b71eb286cdb05858ad06e1344a9206564ce/packages/vue-query/src/types.ts#L58-L60

I think the DistributiveOmit from this line in Vue Query can be removed as well

packages/query-core/src/types.ts
```diff
@@ -9,6 +9,10 @@ import type { QueryFilters, QueryTypeFilter, SkipToken } from './utils'
 import type { QueryCache } from './queryCache'
 import type { MutationCache } from './mutationCache'
 
+export type DistributiveOmit<TObject, TKey> = TObject extends any
+  ? Pick<TObject, Exclude<keyof TObject, TKey>>
+  : never
```
manudeli, line 14:
```suggestion
export type DistributiveOmit<
  TObject,
  TKey extends keyof TObject,
> = TObject extends any ? Omit<TObject, TKey> : never
```

By doing this, we can change it to the following:

### AS-IS

image

### TO-BE

image

TkDodo:
> https://github.com/TanStack/query/blob/89846b71eb286cdb05858ad06e1344a9206564ce/packages/vue-query/src/types.ts#L58-L60
> 
> I think the DistributiveOmit from this line in Vue Query can be removed as well

done: https://github.com/TanStack/query/pull/9105/commits/5cd8471c771df7e19d33fa5c8dfdff35f9d64caf

PR #9116, opened 2025-05-07
chore: add typescript-eslint/require-await rule 

manudeli:


packages/query-codemods/eslint.config.js
```diff
@@ -9,6 +8,7 @@ export default [
     rules: {
       'cspell/spellchecker': 'off',
       '@typescript-eslint/no-unnecessary-condition': 'off',
+      '@typescript-eslint/require-await': 'off',
```
manudeli, line 11: @minseong0324  I'm going to treat it like this now. 

We can apply @typescript-eslint/require-await rule as error to the rest by turning off the this rule only for @tanstack/query-codemods

PR #9125, opened 2025-05-08
refactor: integrate util "noop" function in each packages

novice0840:
I integrated util function "noop". The function "noop" is already made in "query-core" package. So, we don't need to make again in "react-query" package

## Summary by CodeRabbit

- Refactor
  - Consolidated the internal no-op utility to the shared core across Angular, Solid, Svelte, and persister packages for consistency. No behavior or public API changes.
- Chores
  - Bumped @tanstack/query-async-storage-persister to 5.85.5.
- Notes
  - No user-facing changes; apps should behave exactly the same.
  - Improves internal consistency and maintainability without altering existing functionality.


pnpm-lock.yaml
TkDodo, line 1: there shouldn’t be any changes in package-lock if there are no changes in package.json

novice0840: Hello, I deleted all changes in "package.json" file. Please check and tell me if there is some problem

novice0840: My code is already applied by other person. So this PR is not needed anymore. I'm gonna close my PR

PR #9130, opened 2025-05-09
refactor: integration should throw error

novice0840:
In each packages, there was a same util function "shouldThrowError". So, I integrated those in common query-core package.


TkDodo (approved):
nice, thank you

PR #9132, opened 2025-05-09
docs: for QueryCache, list subscribe events

Laurin-W:


TkDodo:
I fear this will get outdated pretty quickly, and linking to a line-of-code in `main` is never a good idea as that is subject to change as well.

We’re working on a way of getting data from `jsdoc` into our api documentation to streamline that.

PR #9139, opened 2025-05-11
feat(core): staleTime: 'static'

TkDodo:


minseong0324:
In my opinion, if "static" is added, it would be better if explicit changes were added to devtools in response. What do you think? If you think it's necessary, I'd like to pull this branch and work on it as a new PR.

minseong0324:
> yeah we could make a separate label like we have for `disabled` 👍 

OK, I'll work on it as soon as possible.

PR #9145, opened 2025-05-13
test(query-core): add test case for query

minseong0324:
I improved test coverage of `query` 

## Before
image

## After
image


packages/query-core/src/__tests__/query.test.tsx
```diff
@@ -1019,4 +1026,71 @@ describe('query', () => {
...
+      queryKey: key,
+      queryFn: () => 'data',
+      persister: () => Promise.resolve('persisted data'),
+    })
+
+    const query = queryCache.find({ queryKey: key })!
+    expect(query.state.data).toBe('persisted data')
+  })
+
+  test('should use queryFn from observer if not provided in options', async () => {
```
TkDodo, line 1043: we could also assert here that `query.options.queryFn` is set to the one of the observer at the end.

minseong0324: I reflected at 41eabff90acfd028397ab2d2544cae05a4abf040

PR #9157, opened 2025-05-16
Fix misc hydration bugs

### LINKED ISSUE #8677: Maximum update depth exceeded
### Describe the bug

I am getting the error "Maximum update depth exceeded"

https://stackblitz.com/~/github.com/yWilliamLima/test-error

### Steps to reproduce

It usually occurs after a while of running whenever changing pages/routes

### Expected behavior

Do not return this error

Sometimes

### Screenshots or Videos

Image
Image

### Platform

OS: Windows 11
Browser: Chrome
Node: 22.12.0

### Tanstack Query adapter

None

### TanStack Query version

5.66.5

### TypeScript version

5.7.3

### Additional context

The error in the console does not occur locally, only in stackblitz
Maybe because of this error in the stackblitz console it doesn't happen, because it gives a refresh
Because it is a minimal example, it takes a while to occur, but in my project it occurs frequently

Ephem:
This PR:

* Fixes #8677 - A bug where hydrating failed queries (and possibly other conditions) could result in an infinite loop
* Fixes an undocumented bug where an old promise could overwrite newer data
* Fixes an undocumented bug where if a promise finished fast and was delivered as a synchronous thenable before React markup hydration, React Query would not hydrate it synchronously.
  * The effect would be that a page would show its full content, but revert to the Suspense state on hydration because the data was not available in the React Query cache, only to immediately show the full content again.
  * This behaviour also seem to have slowed down the initial load of pages with very fast APIs somewhat, so should be snappier now.
* Implements tests for these bugs

At the point of hydrating a promise, we don't know when the promise will _finish_, that is, we don't know the `dataUpdatedAt` of the hydrated query, so we can't compare that to the existing query. The PR sidesteps this by introducing `dehydratedAt` as an approximation. This is not perfect, but is an improvement.

To fix the infinite loop bug I also opted to never hydrate promises if the existing query is already pending, as there is no way to know which of these pending queries will finish first or have the newest data. We could do fancier things like a `Promise.race` and use whatever finishes first, but that seemed like overengineering at this point.

### DIFF
--- a/packages/query-core/src/thenable.ts
+++ b/packages/query-core/src/thenable.ts
@@ -7,6 +7,8 @@
  * @see https://github.com/facebook/react/blob/4f604941569d2e8947ce1460a0b2997e835f37b9/packages/react-debug-tools/src/ReactDebugHooks.js#L224-L227
  */
 
+import { noop } from './utils'
+
 interface Fulfilled<T> {
   status: 'fulfilled'
   value: T
@@ -97,9 +99,9 @@ export function tryResolveSync(promise: Promise<unknown> | Thenable<unknown>) {
       data = result
       return result
     })
-    // This can be unavailable on certain kinds of thenable's
+    // .catch can be unavailable on certain kinds of thenable's
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-    ?.catch(() => {})
+    ?.catch(noop)
 
   if (data !== undefined) {
     return { data }
--- a/packages/query-core/src/thenable.ts
+++ b/packages/query-core/src/thenable.ts
@@ -80,3 +80,30 @@ export function pendingThenable<T>(): PendingThenable<T> {
 
   return thenable
 }
+
+/**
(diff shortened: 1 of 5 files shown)


packages/query-core/src/thenable.ts
```diff
@@ -80,3 +80,30 @@ export function pendingThenable<T>(): PendingThenable<T> {
...
+  let data: unknown
+
+  promise
+    .then((result) => {
+      data = result
+      return result
+    })
+    // This can be unavailable on certain kinds of thenable's
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    ?.catch(() => {})
```
TkDodo, line 102: using `noop` from `query-core` should get rid of the eslint-disable:

```suggestion
    ?.catch(noop)
```

tlazarov-whiskypartners:
Is this going to be part of next release?

TkDodo:
It's already released

PR #9163, opened 2025-05-19
separate concerns of clientOnly and SolidQueryDevtools

wlemahieu:


schiller-manuel:
@birkskyum  could you please have a look?

packages/solid-query-devtools/src/clientOnly.tsx
```diff
@@ -0,0 +1,43 @@
+import {
```
wlemahieu, line 1: No changes were made here. Just moving `clientOnly` to it's own file.

Issue #8988, opened 2025-04-10
useMutation callbacks context type depends on options object sorting


alexw-at:
### Describe the bug

The type of `context` in the `useMutation` callbacks depends on the sorting of the options object.

### Your minimal, reproducible example

https://codesandbox.io/p/devbox/usemutationcontext-qsnk89

### Steps to reproduce

1. Open minimal reproduction
2. Open file `src/index.tsx` or run `tsc --noemit`
3. Type of `context` in callbacks depends options object sorting
4. first `useMutation` works as intended, second `useMutation` has type errors

### Expected behavior

The type of `context` in `useMutation` callbacks should always be the same and not depend on the options object sorting.

### How often does this bug happen?

Every time

### Platform

Seems to be a types only bug. Has no effect on runtime values.

### Tanstack Query adapter

react-query

### TanStack Query version

5.72.2

### TypeScript version

5.8.2

HanhNguyenHong1802:
Should you define type for it first? Like this

```
  const { mutate: mutate2 } = useMutation<
    void,
    unknown,
    string,
    { foo: string }
  >({
    mutationFn: async (value: string) => {
      console.log(value);
    },
    onError: (_error, _variables, context) => {
      console.log("onError", context?.foo);
    },
    onSettled: (_data, _error, _variables, context) => {
      console.log("onSettled", context?.foo);
    },
    onMutate: (data) => {
      return { foo: data };
    },
  });
```
Then the type error will disappear

alexw-at:
Yes this does work and the error disappears. But the inference of `TContext` still depends on the sorting of the options object.

TkDodo:
In an object, type consumers must come after type producers. `onMutate` produces the type, `onError` and `onSettled` consume them. This is a known typescript limitation:

. https://github.com/microsoft/TypeScript/issues/53018#issuecomment-1518558545

We do have an eslint rule for `infinite-query-property-order`. @schiller-manuel do you think it would be possible to extend this to a more general rule that also covers this use-case?

jochongs:
I've created PR #9186 

Could you take a look? :)

TkDodo:
we now have a `mutation-property-order` rule

PR #9186, opened 2025-05-25
feat(eslint-plugin): add mutation-property-order rule and tests

jochongs:
## What

- Added `mutation-property-order` eslint rule

## Why 

#8988

image

**Note: No documentation included in this PR**

### DIFF
--- a/packages/eslint-plugin-query/src/__tests__/mutation-property-order.rule.test.ts
+++ b/packages/eslint-plugin-query/src/__tests__/mutation-property-order.rule.test.ts
@@ -1,0 +1,218 @@
+import { RuleTester } from '@typescript-eslint/rule-tester'
+import combinate from 'combinate'
+
+import {
+  checkedProperties,
+  mutationFunctions,
+} from '../rules/mutation-property-order/constants'
+import {
+  name,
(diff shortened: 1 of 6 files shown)


schiller-manuel:
this seems to have duplicated the `infinite-query-property-order` rule. I would rather reuse as much as possible since the logic of those rules should be pretty much identical.

so I would approach it the following way:

1. make the property-rule generic so you can pass in a function/list of functions and their checked properties
2. instantiate the rule for infinite query
3. instantiate the rule for mutation

jochongs:
Thank you for your feedback!

I initially thought it would be clearer to maintain separate rules for `infinite-query-property-order` and `mutation-property-order`, but I realize now that they share so much logic that it would make sense to extract this into a generic rule.

Would it be okay if I tried to refactor this so that the core logic can be reused, as you suggested? Thank you again for your guidance!

schiller-manuel:
maybe this can be split up into 2 separate PRs then. one to refactor and then one to add the new rule

jochongs:
@schiller-manuel 

I've created a new PR based on your approach.
Could you take a look? :)

#9191

PR #9193, opened 2025-05-27
feat(query-core): add MutationFunctionContext argument to mutateFn

travikk:
This PR adds new argument to `mutateFn` allowing for access to `queryClient` as well as `meta` object.

This feature should make it easier to attach authentication headers from within `mutateFn`, that would typically live in `QueryCache`. Currently that's made very difficult and requires either having singleton context in the codebase or requiring consumers of mutations to pass auth token each time.

This PR also makes it more similar to how `queryFn` behaves, in that it accepts `QueryFunctionContext`.

**Note**: I wasn't entirely sure about the name of `MutateFunctionContext`, but ultimately I decided to go ahead with that name to mirror the way `queryFn` behaves. I realise this might be suboptimal, due to there already existing a `TContext` for mutations. Happy to discuss and take direction here.

### DIFF
--- a/docs/framework/react/reference/useMutation.md
+++ b/docs/framework/react/reference/useMutation.md
@@ -48,10 +48,11 @@ mutate(variables, {
 
 **Parameter1 (Options)**
 
-- `mutationFn: (variables: TVariables) => Promise<TData>`
+- `mutationFn: (variables: TVariables, context: MutationFunctionContext) => Promise<TData>`
   - **Required, but only if no default mutation function has been defined**
   - A function that performs an asynchronous task and returns a promise.
   - `variables` is an object that `mutate` will pass to your `mutationFn`
+  - `context` is an object that `mutate` will pass to your `mutationFn`. Contains reference to `QueryClient` and optional `meta` object.
 - `gcTime: number | Infinity`
   - The time in milliseconds that unused/inactive cache data remains in memory. When a mutation's cache becomes unused or inactive, that cache data will be garbage collected after this duration. When different cache times are specified, the longest one will be used.
   - If set to `Infinity`, will disable garbage collection
--- a/packages/query-core/src/__tests__/mutations.test.tsx
+++ b/packages/query-core/src/__tests__/mutations.test.tsx
@@ -50,8 +50,36 @@ describe('mutations', () => {
       'vars',
     )
 
+    const args = fn.mock.calls[0]!
+
     expect(fn).toHaveBeenCalledTimes(1)
-    expect(fn).toHaveBeenCalledWith('vars')
+    expect(args[0]).toEqual('vars')
+  })
+
+  test('should provide MutationFunctionContext to mutateFn', async () => {
+    const key = queryKey()
+    const fn = vi.fn()
+    const meta = { hello: 'world' }
+
+    await executeMutation(
+      queryClient,
+      {
+        mutationKey: key,
+        mutationFn: fn,
+        meta: meta,
+      },
+      'vars',
+    )
+
+    const args = fn.mock.calls[0]!
+    expect(fn).toHaveBeenCalledTimes(1)
+    expect(args).toBeDefined()
+
+    const vars = args[0]
+    const mutationFnContext = args[1]
+    expect(vars).toEqual('vars')
+    expect(mutationFnContext.client).toEqual(queryClient)
+    expect(mutationFnContext.meta).toEqual(meta)
   })
 
   test('mutation should set correct success states', async () => {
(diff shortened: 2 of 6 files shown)


TkDodo:
we already have a PR for this:

- https://github.com/TanStack/query/pull/7343

can we consolidate the work in one PR?

I think we’d want 3 properties passed:

```
- mutationKey
- meta
- client
```

travikk:
@TkDodo updated as requested

docs/framework/react/reference/useMutation.md
```diff
@@ -48,10 +48,11 @@ mutate(variables, {
 
 **Parameter1 (Options)**
 
-- `mutationFn: (variables: TVariables) => Promise<TData>`
+- `mutationFn: (variables: TVariables, context: MutationFunctionContext) => Promise<TData>`
   - **Required, but only if no default mutation function has been defined**
   - A function that performs an asynchronous task and returns a promise.
   - `variables` is an object that `mutate` will pass to your `mutationFn`
+  - `context` is an object that `mutate` will pass to your `mutationFn`. Contains reference to `QueryClient`, `mutationKey` and optional `meta` object.
```
TkDodo, line 55: there’s one more thing that’s a bit troubling, and that’s the fact that we already refer to something as “context” for mutations: The value that gets returned from `onMutate`, which can be seen in this example here:

https://tanstack.com/query/v5/docs/framework/react/guides/optimistic-updates#updating-a-list-of-todos-when-adding-a-new-todo

and in code here:

https://github.com/TanStack/query/blob/02c3d5130e063c5f623adbfaf9bd986273b615cc/packages/query-core/src/mutation.ts#L29

since it’s part of the public `MutationState`, which is available via `useMutationState`, we can’t just rename this, so it’s a bit unfortunate.

Any ideas ?

solastley: Given that the equivalent object in `queryFn` is already called `context`, it seems the trade off is **consistency with `QueryFunctionContext` context** vs **a conflict with `MutationState` context**.

One idea is to rename the `QueryFunctionContext` to something else and name this new mutation function context accordingly. Examples:

- `QueryFunctionParams` and `MutationFunctionParams`
- `QueryFunctionInfo` and `MutationFunctionInfo`
- `QueryFunctionExecutionContext` and `MutationFunctionExecutionContext`
- `QueryFunctionDetails` and `MutationFunctionDetails`
- `QueryFunctionMetaContext` and `MutationFunctionMetaContext`

This way we get consistency between `queryFn` and `mutationFn`, but eliminate the conflict with `MutationState`.

Honestly I don't _love_ any of these, but I think the root of that issue is that these objects and the context in `MutationState` are both most accurately described as a "context".

solastley: @travikk light bump on this!

travikk: Hey apologies for missing the updates.

Yeah, I totally agree with @solastley here. There doesn't seem to be a preferred way forward here, but there's clearly a benefit in having consistency in API between `queryFn` and `mutateFn` in that we can access QueryCache or QueryClient from both.

Can we perhaps bet on a slightly tuned/better documentation to distinguish between `mutateFn` context proposed here and return value of `onMutate`?

travikk: @TkDodo do you mind weighing in?

TkDodo: would it be a breaking change to rename the return value of `onMutate` to something else? After all, it’s just as positional argument. I never liked `context` for this...

Rendez: How about calling it `scope`?

```ts
mutationFn(vars, context) {
   if (context.meta.something) {
     // ...
   }
   return fetch('...');
},
onMutate(_vars, context) {
  if (context.client) {
     // ...
  }
  const scope = {foo: 'bar'}
  return scope;
},
onSuccess(_data, _vars, scope) {
  console.log(scope)
}
```

TkDodo: @Rendez let’s do it 👍

joshuaellis:
We're quite interested in utilising this new feature at Sanity as part of our own abstraction for a `SanityQueryClient` – is there anything to do to help push this PR forward? Is it just the naming of context vs scope or more...?

TkDodo:
@joshuaellis I’m fine with the `scope` renaming of what gets returned from `onMutate`. Then, we can have `context` for what gets passed to the `mutationFn` as second argument. I think it just needs to be done. If you want to open another PR to get it over the finish line, please do.

TkDodo:
superseded by:

- https://github.com/TanStack/query/pull/9615

PR #9199, opened 2025-05-27
Support `Map` and `Set` as query keys

Zentrik:
I'm not sure if this is intended to be supported but I figured it wouldn't hurt to open this pr. Feel free to close if we don't want to support `Map` and `Set` instances as query keys. But I found this useful for me.

JSON.stringify doesn't properly support instances of these types and will just return "{}". xref https://github.com/TanStack/query/discussions/6444


TkDodo:
we want to keep being json compatible because it’s hard to justify drawing a line otherwise. If we add maps and sets, why not symbols? Why not Date? Why not `undefined` ? Why not `Regexp`? Why not ...

You can always add a global `queryKeyHashFn` and use something like [superjson](https://github.com/flightcontrolhq/superjson) to do custom serialization.

PR #9201, opened 2025-05-28
perf: improve performance of util function "difference" in quriesObserver.ts in query-core package

novice0840:
I changed the internal behavior of the "difference" function in **quriesObserver.ts** from **includes function** to using **has function** in sets

And this is faster when the arrays are longer. I included a link to the code sandbox below so you can see the specific performance differences.

## Codesandbox Link 
https://codesandbox.io/p/devbox/crazy-cherry-627tsr

### DIFF
--- a/packages/query-core/src/queriesObserver.ts
+++ b/packages/query-core/src/queriesObserver.ts
@@ -10,7 +10,7 @@ import type {
 import type { QueryClient } from './queryClient'
 
 function difference<T>(array1: Array<T>, array2: Array<T>): Array<T> {
-  return array1.filter((x) => !array2.includes(x))
+  return array1.filter((x) => !new Set(array2).has(x))
 }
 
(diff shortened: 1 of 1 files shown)


packages/query-core/src/queriesObserver.ts
```diff
@@ -10,7 +10,7 @@ import type {
 import type { QueryClient } from './queryClient'
 
 function difference<T>(array1: Array<T>, array2: Array<T>): Array<T> {
-  return array1.filter((x) => !array2.includes(x))
+  return array1.filter((x) => !new Set(array2).has(x))
```
TkDodo: can we move the `new Set` creation out of the `filter` function to only create the Set once ?

novice0840: Of course, I did it. Please check

PR #9202, opened 2025-05-28
refactor: narrow onSuccess/onError/onMutate/onSettled callback types to Promise<void> | void

braden-w:
Previously, the `onSuccess`, `onError`, `onMutate`, and `onSettled` callbacks were typed as `() => Promise<unknown> | unknown`. While `unknown` is technically valid, it implies that the return value might be used, assigned, or further processed. However, throughout the codebase, these callbacks are invoked solely for their side effects, and their return values are always ignored. Narrowing the type to `Promise<void> | void` makes this intent explicit, clarifies that any return value will be discarded, and prevents misleading type signatures that suggest otherwise.

This commit narrows their types to `() => Promise<void> | void`, which more accurately reflects their intended use.

### DIFF
--- a/docs/framework/react/plugins/persistQueryClient.md
+++ b/docs/framework/react/plugins/persistQueryClient.md
@@ -214,12 +214,12 @@ ReactDOM.createRoot(rootElement).render(
 
 - `persistOptions: PersistQueryClientOptions`
   - all [options](#options) you can pass to [persistQueryClient](#persistqueryclient) minus the QueryClient itself
-- `onSuccess?: () => Promise<unknown> | unknown`
+- `onSuccess?: () => Promise<void> | void`
   - optional
   - will be called when the initial restore is finished
   - can be used to [resumePausedMutations](../../../../reference/QueryClient.md#queryclientresumepausedmutations)
   - if a Promise is returned, it will be awaited; restoring is seen as ongoing until then
-- `onError?: () => Promise<unknown> | unknown`
+- `onError?: () => Promise<void> | void`
   - optional
   - will be called when an error is thrown during restoration
   - if a Promise is returned, it will be awaited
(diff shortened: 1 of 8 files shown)


TkDodo (approved):

Sagie501:
@braden-w @TkDodo 
I have the following code:
```
export const useCreateProduct = () => {
  const api = useApi();

  return useMutation({
    mutationFn: (createProductDto: CreateProductDto) =>
      api.products.productControllerCreateProduct(createProductDto).then((res) => res.data),
    onSuccess: (data) =>
      Promise.all([queryClient.invalidateQueries({ queryKey: /* some query key */ }), queryClient.invalidateQueries({ queryKey: /* some other query key */ })]),
  });
};
```
And with the following change I am getting the following error:
image

What would you suggest me to do?

jgarplind:
Same issue as @Sagie501 for us. Not trying to start a flood of +1s, but I think it makes sense to clarify that this is a somewhat commonly used pattern, probably impacting many users.

Sagie501:
Changing the callbacks type to `void | Promise<void> | Promise<Array<void>>`   will solve my case, does that making sense?

braden-w:
@TkDodo just made a quick PR #9245 to address the points from @Sagie501 @jgarplind —let me know what you think!

braden-w:
@Sagie501 @jgarplind For now, you can fix this by using the `void` prefix:

```typescript
mutationFn: (createProductDto: CreateProductDto) =>
  void api.products.productControllerCreateProduct(createProductDto).then((res) => res.data),
```

or using braces:

```typescript
mutationFn: (createProductDto: CreateProductDto) => {
  api.products.productControllerCreateProduct(createProductDto).then((res) => res.data)
},
```

In the first method with `void`, we're telling Typescript to evaluate `api.products.productControllerCreateProduct(createProductDto).then((res) => res.data)` but then return `undefined` (in other words, discard the returned result).

In the second method, we are using braces without any explicit return, so the function implicitly returns `undefined`.

Both methods would satisfy the current typing.

In the meantime, we can see if ~#9241~ #9245  is a good candidate for merging.

braden-w:
@Sagie501 ~I actually ended up changing the return types to just `void` in #9245! This would end up matching the widest variety of return types from a function, while still implying that we're discarding the return value. `void | Promise<void>` was being problematic.~ We're discussing changing the return type in #9245 !

PR #9212, opened 2025-05-29
refactor(core): add generic utilities for resolving value-or-function patterns, replace specialized `resolveStaleTime` and `resolveEnabled`

braden-w:
This PR introduces `isFunctionVariant()` and `resolveValueOrFunction()` utilities to replace repetitive `typeof === 'function'` checks throughout the codebase, consolidating the common "value or function that computes value" pattern.

## Problem and Solution

The codebase had scattered implementations of the same pattern across multiple files:

```typescript
// Repetitive pattern everywhere
const data = typeof options.initialData === 'function' 
  ? options.initialData() 
  : options.initialData

const delay = typeof retryDelay === 'function'
  ? retryDelay(failureCount, error) 
  : retryDelay
```

This led to code duplication, inconsistency, and maintenance overhead. We also had specialized versions like `resolveStaleTime()` and `resolveEnabled()` that could be generalized.

The new utilities provide a clean, generic solution:

```typescript
// Clean and intention-revealing
const data = resolveValueOrFunction(options.initialData)
const delay = resolveValueOrFunction(retryDelay, failureCount, error)
```

While we could inline `typeof value === 'function'` checks, TypeScript's type narrowing doesn't work properly with generic types in complex expressions. The `isFunctionVariant()` type guard provides proper type narrowing that allows `resolveValueOrFunction()` to safely call the function variant. Without it, TypeScript throws errors because it can't guarantee the type safety across the generic boundary.

Both utilities support zero-argument functions (`() => T`) and functions with parameters (`(...args) => T`), making them applicable to all value-or-function patterns in the codebase.

Updated implementations in `query.ts`, `queryObserver.ts`, and `retryer.ts` for handling `initialData`, `retryDelay`, `refetchInterval`, and `notifyOnChangeProps`. These utilities can replace existing `resolveStaleTime()` and `resolveEnabled()` functions in future iterations.

## Summary by CodeRabbit

* **New Features**
  * Placeholder data can now be provided as either a direct value or a function for more flexible initial UI states.

* **Refactor**
  * Consolidated option handling into a single, consistent resolution approach used for enabled checks, stale timing, initial/placeholder data, retry delays, and observer notifications.

* **Chores**
  * Removed legacy per-option helpers and some exported helper types; introduced a unified option-resolver utility and consolidated related logic.


packages/angular-query-devtools-experimental/package.json
```diff
@@ -58,7 +58,8 @@
     "@angular/platform-browser-dynamic": "^20.0.0",
     "@tanstack/angular-query-experimental": "workspace:*",
     "eslint-plugin-jsdoc": "^50.5.0",
-    "npm-run-all2": "^5.0.0"
+    "npm-run-all2": "^5.0.0",
+    "tsup": "^8.4.0"
```
TkDodo: why was adding `tsup` necessary here? I don’t think angular uses `tsup` for bundling 🤔

packages/query-core/src/utils.ts
```diff
@@ -80,6 +77,94 @@ export function noop(): void
...
+ * ```
+ */
+export function resolveValueOrFunction<T, TArgs extends Array<any>>(
+  value: T | ((...args: TArgs) => T),
+  ...args: TArgs
+): T {
+  return isFunctionVariant(value) ? value(...args) : value
+}
+
 export function functionalUpdate<TInput, TOutput>(
```
TkDodo: could we use `resolveValueOrFunction` here as well, as this also uses a `typeof updater === 'function'` check ?

packages/query-core/src/utils.ts
```diff
@@ -80,6 +77,94 @@ export function noop(): void
...
+ * // Replaces verbose patterns like:
+ * // const delay = typeof retryDelay === 'function'
+ * //   ? retryDelay(failureCount, error)
+ * //   : retryDelay
+ *
+ * // With:
+ * const delay = resolveValueOrFunction(retryDelay, failureCount, error)
+ * ```
+ */
+export function resolveValueOrFunction<T, TArgs extends Array<any>>(
```
TkDodo: I think a name like `resolveOption` would be descriptive enough

packages/query-core/src/utils.ts
```diff
@@ -80,6 +77,94 @@ export function noop(): void
...
+  value: T | ((...args: TArgs) => T),
```
TkDodo: wondering if we should limit `T` with `NonFunctionGuard` like we do here:

https://github.com/TanStack/query/blob/b2af5689f3da63de5c134b7bab1d5e50a36d8162/packages/query-core/src/types.ts#L422-L429

otherwise, `T` could be a function itself and then everything becomes ambiguous

TkDodo (commented):
I generally agree with the direction, please have a look at the comments though

TkDodo:
hm, the failing pipeline shows a lot of type-errors

TkDodo:
it’s been two months, there’s still a failing pipeline and conflicts. please let me know if you intend to finish this PR, because it doesn’t look like it’s close to working. I’ll close it otherwise.

braden-w:
Thanks for the ping! Will update it and get back to you by EOD.

packages/angular-query-devtools-experimental/package.json
braden-w: Removed! In earlier commits, the lack of tsup was blocking builds.

packages/query-core/src/utils.ts
braden-w: Seems like NonFunctionGuard is no longer in the codebase, so I'll refrain from adding it for now!

braden-w:
Hey @TkDodo! Updated with a rewrite that fixes the type errors.

Renamed to `resolveOption` as you suggested. Regarding `NonFunctionGuard`: I dug into this and found it doesn't work when `T` is a generic type parameter (like `TData`, `TQueryData`) because TypeScript can't evaluate `TData extends Function` at compile time—the conditional type stays unresolved and breaks inference.

The fix uses input-driven type inference instead (same pattern as React's `useState` and Zustand):

```typescript
type ResolvedValue<TValueOrFn> = TValueOrFn extends (...args: Array<any>) => infer R
  ? R
  : TValueOrFn

export function resolveOption<TValueOrFn>(
  valueOrFn: TValueOrFn,
  ...args: Array<unknown>
): ResolvedValue<TValueOrFn>
```

This infers the return type from the actual input type at each call site, sidestepping the generic limitation. For cases where the input involves a generic like `TData`, we use either a type assertion (`as TData | undefined`) or a functional wrapper (`() => result`).

Also unified this with `functionalUpdate`—`setQueryData` now uses `resolveOption` internally. Converted `initialData`, `initialDataUpdatedAt`, and `placeholderData` to use it as well.

Let me know if you'd like any tweaks!

packages/query-core/src/query.ts
```diff
@@ -726,17 +724,12 @@ function getDefaultState<
 >(
   options: QueryOptions<TQueryFnData, TError, TData, TQueryKey>,
 ): QueryState<TData, TError> {
-  const data =
-    typeof options.initialData === 'function'
-      ? (options.initialData as InitialDataFunction<TData>)()
-      : options.initialData
+  const data = resolveOption(options.initialData) as TData | undefined
```
TkDodo, line 727: curious why this type assertion is necessary for `optoins.initialData` when it works fine below for `options.initialDataUpdatedAt` 🤔

packages/query-core/src/streamedQuery.ts
```diff
@@ -114,7 +114,7 @@ export function streamedQuery<
 
     // finalize result: replace-refetching needs to write to the cache
     if (isReplaceRefetch && !cancelled) {
-      context.client.setQueryData<TData>(context.queryKey, result)
+      context.client.setQueryData<TData>(context.queryKey, () => result)
```
TkDodo, line 117: just wondering if this indicates we have a breaking type change somewhere? I guess the change is because technically, `TData` can contain a function? I’d still rather avoid the runtime change.

I guess it’s fine that we disallow having functions in that generic

TkDodo:
4 months after my last review we now have lots of conflicts. If you want to land this I think it’s easier to file a new PR. Thanks.

PR #9215, opened 2025-05-29
feat(query-devtools): add indicator for staleTime: 'static' in devtools

minseong0324:
This PR adds a visual indicator in the DevTools for queries that use the `staleTime: 'static'` option. When a query is in a `static` state, it now displays a distinct teal label in the UI, similar to how `disabled` queries are currently labeled.

This PR addresses the suggestion in [this comment](https://github.com/TanStack/query/pull/9139#issuecomment-2879752645).

## Testing

I tested in `examples/react/devtools-panel/src/index.tsx` with the following example code.

```jsx
function Example() {
  const { isPending, error, data, isFetching } = useQuery({
    staleTime: 'static', // add this line
    queryKey: ['repoData'],
    queryFn: async () => {
      const response = await fetch(
        'https://api.github.com/repos/TanStack/query',
      )
      return await response.json()
    },
  })

  // ... rest of the code ...
}
```

## Screenshots

image


TkDodo:
will we get `static` and `disabled` labels when we have both ?

minseong0324:
> will we get `static` and `disabled` labels when we have both ?

Yes. with our current implementation, a query with both `enabled: false` and `staleTime: 'static'` will display both the `disabled` and `static` labels.

image

TkDodo:
Looks good 👍

PR #9216, opened 2025-05-29
refactor: rename cache methods for better clarity and consistency

braden-w:
Renamed cache methods in QueryCache and MutationCache to better reflect their behavior and improve developer understanding:

- `QueryCache.build()` → `QueryCache.ensure()`
- `MutationCache.build()` → `MutationCache.create()`
- Enhanced documentation for `QueryCache.get()` to clarify retrieval-only behavior

The previous naming was confusing because both `QueryCache.build()` and `MutationCache.build()` had the same name but fundamentally different behaviors:

- **QueryCache.build()**: Used a "get or create" pattern - retrieved existing queries or created new ones
- **MutationCache.build()**: Always created new mutation instances (since each mutation execution should be unique)

This inconsistency made the API harder to understand and could lead to incorrect assumptions about behavior.

### QueryCache
- **`build()` → `ensure()`**: Better conveys "ensure this query exists" semantics, borrowing the `ensure` pattern used in `ensureQueryData` and `ensureQueryFn` 
- **Enhanced `get()` documentation**: Clarifies it only retrieves existing queries, returns `undefined` if not found
- **Improved `ensure()` documentation**: Explains the "get or create" behavior and when to use it

### MutationCache
- **`build()` → `create()`**: Clearly indicates it always creates new mutation instances
- **Enhanced documentation**: Explains why mutations are always created new (each execution should be unique)

### Updated References
Updated all usages across the codebase:
- `packages/query-core/src/queryClient.ts` - Updated QueryCache method calls
- `packages/query-core/src/queryObserver.ts` - Updated QueryCache method calls
- `packages/query-core/src/mutationObserver.ts` - Updated MutationCache method calls
- `packages/query-core/src/hydration.ts` - Updated both cache method calls
- `packages/query-core/src/__tests__/utils.ts` - Updated test utilities
- `packages/query-persist-client-core/src/__tests__/persist.test.ts` - Updated test
- `packages/query-broadcast-client-experimental/src/index.ts` - Updated broadcast client

### DIFF
--- a/packages/query-broadcast-client-experimental/src/index.ts
+++ b/packages/query-broadcast-client-experimental/src/index.ts
@@ -78,7 +78,7 @@ export function broadcastQueryClient({
           return
         }
 
-        queryCache.build(
+        queryCache.ensure(
           queryClient,
           {
             queryKey,
@@ -95,7 +95,7 @@ export function broadcastQueryClient({
           query.setState(state)
           return
         }
-        queryCache.build(
+        queryCache.ensure(
           queryClient,
           {
             queryKey,
--- a/packages/query-core/src/__tests__/mutations.test.tsx
+++ b/packages/query-core/src/__tests__/mutations.test.tsx
@@ -261,7 +261,7 @@ describe('mutations', () => {
 
     const mutation = queryClient
       .getMutationCache()
-      .build<string, unknown, string, string>(
+      .create<string, unknown, string, string>(
         queryClient,
         {
           mutationKey: key,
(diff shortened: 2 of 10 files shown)


TkDodo:
we can’t really do that because `QueryCache` and `MutationCache` are part of the public API of the `@tanstack/query-core` package, and the `build` method is publicly available.

So this change would be breaking from that perspective, and everyone who has built an adapter on top of the query-core would see that breaking change.

PR #9217, opened 2025-05-29
refactor: simplify `setQueryData` by using `ensure` directly instead of `get` then `ensure`

braden-w:
This commit builds on the changes from #9216.

Refactored the `setQueryData` method in `QueryClient` to simplify its logic and improve readability.

The previous implementation of `setQueryData` was redundant:

1. It first called `this.#queryCache.get()` to retrieve an existing query.
2. Then, it called `this.#queryCache.ensure()` which internally also performs a "get or create" operation.

This meant the query was potentially being retrieved or checked for existence twice. The `ensure()` method already handles the logic of getting an existing query or creating a new one if it doesn't exist.

The `setQueryData` method has been updated to:

1. Directly call `this.#queryCache.ensure()` to get or create the query.
2. Use the returned query instance to access `state.data` for the `functionalUpdate`.
3. Call `setData()` on the same query instance.

**Before:**

```typescript
    const query = this.#queryCache.get<TInferredQueryFnData>(
      defaultedOptions.queryHash,
    )
    const prevData = query?.state.data
    const data = functionalUpdate(updater, prevData)

    if (data === undefined) {
      return undefined
    }

    return this.#queryCache
      .ensure(this, defaultedOptions)
      .setData(data, { ...options, manual: true })
```

**After:**

```typescript
    const query = this.#queryCache.ensure<TInferredQueryFnData>(
      this,
      defaultedOptions,
    )
    const prevData = query.state.data
    const data = functionalUpdate(updater, prevData)

    if (data === undefined) {
      return undefined
    }

    return query.setData(data, { ...options, manual: true })
```

### DIFF
--- a/packages/query-broadcast-client-experimental/src/index.ts
+++ b/packages/query-broadcast-client-experimental/src/index.ts
@@ -78,7 +78,7 @@ export function broadcastQueryClient({
           return
         }
 
-        queryCache.build(
+        queryCache.ensure(
           queryClient,
           {
             queryKey,
@@ -95,7 +95,7 @@ export function broadcastQueryClient({
           query.setState(state)
           return
         }
-        queryCache.build(
+        queryCache.ensure(
           queryClient,
           {
             queryKey,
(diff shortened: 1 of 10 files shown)


braden-w:
Please disregard the pull request; I realized that this logic incorrectly creates a new query even if the `query` was not found. The original implementation was correct!

PR #9222, opened 2025-05-30
fix(svelte-query): fix error when creating query outside of component

arnolicious:
## Description
This fixes an `effect_orphan` error that occurs, when `createQuery` gets called in an event listener, or simply outside of component initialization.

I simply added `$effect.root` around the `$effects` in  `createBaseQuery.svelte.ts` which fixed the error on my end while seemingly not breaking anything.


arnolicious:
I wasn't sure how best to do this, because in the tests this change basically only results in not needing the `withEffectRoot` setup.
In practice, this comes in handy when you do something like this, which would throw an error otherwise:

```svelte
<script>
  const onCreateQuery = () => {
    createQuery({...})
  }
</script>

<button onclick={onCreateQuery} />
```

Let me know if you have other ideas for the tests

lachlancollins:
> I wasn't sure how best to do this, because in the tests this change basically only results in not needing the `withEffectRoot` setup. In practice, this comes in handy when you do something like this, which would throw an error otherwise:
> 
> ```svelte
> <script>
>   const onCreateQuery = () => {
>     createQuery({...})
>   }
> </script>
> 
> <button onclick={onCreateQuery} />
> ```
> 
> Let me know if you have other ideas for the tests

What's the purpose of calling createQuery in an event like that? It seems like there might be a better way to achieve what you're after, but I might be missing something!

arnolicious:
I encountered this when working on a decently complex use case, but very roughly:

- The underlying data structure is a list of items, that can be nested twice, each level being paginated.
- At first, a query is created for the root element, which returns the root and it's children, and in those each of those children, the first few grandchildren
- The user then can decide to want to load more of those grandchildren for a given child X
- In that moment, I create a new query specifically for X 

The end result is basically:
- root query
  - potential child query 1
  - potential child query 2
  - etc

And the child queries being created on demand, imperatively. Considering the small fix, this seemed like the easiest solution

lachlancollins:
> I encountered this when working on a decently complex use case, but very roughly:
> 
> * The underlying data structure is a list of items, that can be nested twice, each level being paginated.
> * At first, a query is created for the root element, which returns the root and it's children, and in those each of those children, the first few grandchildren
> * The user then can decide to want to load more of those grandchildren for a given child X
> * In that moment, I create a new query specifically for X
> 
> The end result is basically:
> 
> * root query
>   
>   * potential child query 1
>   * potential child query 2
>   * etc
> 
> And the child queries being created on demand, imperatively. Considering the small fix, this seemed like the easiest solution

It sounds like this might be more easily and more "correctly" solved with child components - are you able to spin up a repro?

arnolicious:
I won't really be able to create a reproduction for this use case, since the corresponding api is the central reason why this is needed.

Child components cannot be used, because to work with the data and display it, I actually flatten all items into a simple array.

If this doesn't get merged, I would be fine with it, as I have been running this patch since discovering the issue. 
However I also have not seen any disadvantage when using effect.root...

PR #9224, opened 2025-05-30
refactor: remove unnecessary TQueryData generic in InfiniteQueryObserverOptions

braden-w:
This PR simplifies `InfiniteQueryObserverOptions` by removing the unnecessary `TQueryData` generic parameter.

For infinite queries, the actual data stored in the cache is the per-page data (defined by `TQueryFnData`) wrapped within an `InfiniteData` structure: `InfiniteData<TQueryFnData, TPageParam>`.

The previous definition of `InfiniteQueryObserverOptions` introduced an intermediate generic, `TQueryData`, which was defaulted to `TQueryFnData`. This `TQueryData` was then used to define the cached structure as `InfiniteData<TQueryData, TPageParam>`.

```ts
// Before:
export interface InfiniteQueryObserverOptions<
  TQueryFnData = unknown,
  TError = DefaultError,
  TData = TQueryFnData,
  TQueryData = TQueryFnData, // <-- Initialized TQueryData, always the same as TQueryFnData
  TPageParam = unknown,
> extends QueryObserverOptions<
      TQueryFnData,
      TError,
      TData,
      InfiniteData<TQueryData, TPageParam>, // <-- Used the intermediate TQueryData
      TQueryKey,
      TPageParam
    >,
    InfiniteQueryPageParamsOptions<TQueryFnData, TPageParam> {}
```

This intermediate `TQueryData` step was redundant because the type of the data for each page (`TQueryFnData`) is already known. The cached data structure can therefore directly use `TQueryFnData`.

This change removes the `TQueryData` generic, simplifying the type signature:

```ts
// After:
export interface InfiniteQueryObserverOptions<
  TQueryFnData = unknown,
  TError = DefaultError,
  TData = TQueryFnData,
  TPageParam = unknown,
> extends QueryObserverOptions<
      TQueryFnData,
      TError,
      TData,
      InfiniteData<TQueryFnData, TPageParam>, // <-- Directly uses TQueryFnData
      TQueryKey,
      TPageParam
    >,
    InfiniteQueryPageParamsOptions<TQueryFnData, TPageParam> {}
```

This simplification reduces complexity and makes the type definition more direct.

The update has been consistently applied across:
*   Core type definitions in `packages/query-core/src/types.ts` (notably `InfiniteQueryObserverOptions` and `DefaultedInfiniteQueryObserverOptions`).
*   The `InfiniteQueryObserver` class in `packages/query-core/src/infiniteQueryObserver.ts`.
*   Type definitions and helper functions (e.g., `infiniteQueryOptions`, `UseInfiniteQueryOptions`) in framework packages (`react-query`, `vue-query`, `solid-query`, `angular-query-experimental`).
*   Associated test files to ensure type correctness with the new signatures.


TkDodo:
I wanted to get this done for so long but couldn’t fully wrap my head around it. Thank you 🙏

TkDodo (approved):

TkDodo:
btw @braden-w if you’re on Discord, I’d love to connect ❤️ . Let me know if you plan to stick around, your contributions are awesome 🙌

braden-w:
Amazing, thank you @TkDodo for your work and comments on my previous PRs!! This week is busy but I'll do my best to update them sometime next week :) and would love to connect on Discord! I just sent you a message and friend request.

disintegrator:
Hey @braden-w @TkDodo, this change turns out to be a breaking change. At least for folks like us that work with the `UseInfiniteQueryOptions` type that's now changed from 6 to 5 type parameters. We're currently looking at alternatives that don't require us to depend on that type but I think it's worth flagging.

TkDodo:
Thanks for flagging. I’m very cautious about accidental breaking changes, but it’s not really feasible doing that for types, too. We’ve noted that here:

https://tanstack.com/query/v5/docs/framework/react/typescript

> Changes to types in this repository are considered non-breaking and are usually released as patch semver changes (otherwise every type enhancement would be a major version!).

disintegrator:
Thanks @TkDodo that's totally fair. We'll figure something out. This one has meaningful impact for folks building libraries of hooks that wrap react query primitives. Especially painful if it means going from peer dependency of `"@tanstack/react-query": "^5"` to `"@tanstack/react-query": "^5.80"` as it will likely mean that some existing consumer may end up with two or more installs of the react query.

My current plan is to inline the type in my library. It should break the dependency on `UseInfiniteQueryOptions` and allow me to maintain `"@tanstack/react-query": "^5"` as peer dep requirement.

SuperOleg39:
> Type inference is broken since tanstack query 5.80 jotaijs/jotai-tanstack-query#102

Don't know is this is still actual, but in our react-query wrapper (for both v4/v5 RQ versions support) this snippet helped:

```ts
export type SafeUseInfiniteQueryOptions<
  TQueryFnData = unknown,
  TError = unknown,
  TData = TQueryFnData,
  TPageParam = unknown,
> = InfiniteQueryObserverOptions extends { experimental_prefetchInRender?: boolean }
  ? UseInfiniteQueryOptions<TQueryFnData, TError, TData, TQueryKey, TPageParam>
  : // @ts-expect-error backward compatibility with RQ v4
    UseInfiniteQueryOptions<TQueryFnData, TError, TData, TData, TQueryKey>;
```

`InfiniteQueryObserverOptions extends { experimental_prefetchInRender?: boolean }` - will match only if `v5` react-query is installed in our users projects.

So, only `SafeUseInfiniteQueryOptions` is used in our wrapper.

One problem here - work only for v4 and >=5.80 :(

PR #9230, opened 2025-06-01
fix(experimental_createQueryPersister): Add persisterRestoreByKey

rklomp:


rklomp:
I would also suggest to rename the fn and gc functions. Can be done in a different pull request.

`persisterFn` => `fn`
`persisterGc` =>  `garbageCollect`

This will prevent repetitive 'persister'.
For example `persister.persisterFn` would become the more logical `persister.fn`

DamianOsipiuk:
> I would also suggest to rename the fn and gc functions. Can be done in a different pull request.
> 
> `persisterFn` => `fn` `persisterGc` => `garbageCollect`
> 
> This will prevent repetitive 'persister'. For example `persister.persisterFn` would become the more logical `persister.fn`

I'm fine with `persisterGc => garbageCollect` and `persisterRestoreAll => restoreQueries`

But i would prefer a better name than just `fn`.

TkDodo:
`persisterFn` is pretty close to `queryFn` and `mutationFn`, and `persisterGc` aligns well with `persisterFn` and `gcTime` that we also have. Not sure I’d prefer the rename here.

PR #9236, opened 2025-06-03
fix(angular-query): call withDevtools function within injection conte…

michaelbe812:
…xt to not throw

# Bug description
Following [this example](https://tanstack.com/query/latest/docs/framework/angular/devtools#configuring-if-devtools-are-loaded): 

```ts
@Injectable({ providedIn: 'root' })
class DevtoolsOptionsManager {
  loadDevtools = toSignal(
    fromEvent<KeyboardEvent>(document, 'keydown').pipe(
      map(
        (event): boolean =>
          event.metaKey && event.ctrlKey && event.shiftKey && event.key === 'D',
      ),
      scan((acc, curr) => acc || curr, false),
    ),
    {
      initialValue: false,
    },
  )
}

export const appConfig: ApplicationConfig = {
  providers: [
    provideHttpClient(),
    provideTanStackQuery(
      new QueryClient(),
      withDevtools(() => ({
        initialIsOpen: true,
        loadDevtools: inject(DevtoolsOptionsManager).loadDevtools(),
      })),
    ),
  ],
}
````

will throw a runtime error that `inject` must be called from within an injection context. 

CleanShot 2025 06 03 - Google Chrome-TanStack Query devtools panel example-000777@2x

CleanShot 2025 06 03 - Google Chrome-TanStack Query devtools panel example-000779@2x

### DIFF
--- a/packages/angular-query-experimental/src/providers.ts
+++ b/packages/angular-query-experimental/src/providers.ts
@@ -2,10 +2,12 @@ import {
   DestroyRef,
   ENVIRONMENT_INITIALIZER,
   InjectionToken,
+  Injector,
   PLATFORM_ID,
   computed,
   effect,
   inject,
+  runInInjectionContext,
 } from '@angular/core'
 import { QueryClient, noop, onlineManager } from '@tanstack/query-core'
 import { isPlatformBrowser } from '@angular/common'
@@ -270,14 +272,19 @@ export function withDevtools(
         // Do not use provideEnvironmentInitializer while Angular < v19 is supported
         provide: ENVIRONMENT_INITIALIZER,
         multi: true,
-        useFactory: () => {
+        deps: [Injector],
+        useFactory: (injector: Injector) => {
           if (!isPlatformBrowser(inject(PLATFORM_ID))) return noop
           const injectedClient = inject(QueryClient, {
             optional: true,
           })
           const destroyRef = inject(DestroyRef)
 
-          const options = computed(() => withDevtoolsFn?.() ?? {})
+          const options = computed(() => {
+            return runInInjectionContext(injector, () => {
+              return withDevtoolsFn?.() ?? {}
+            })
+          })
 
           let devtools: TanstackQueryDevtools | null = null
           let el: HTMLElement | null = null

### CODE REVIEW

**Comment by michaelbe812** [comment]
Just have to think about how to write a test case for it - help welcome :)

**Comment by arnoud-dv** [comment]
I don't think a computation function should use `inject`, which is why I removed `runInInjectionContext`, forgetting that it breaks the use case and my own example code.

There's good reasons Angular doesn't run `computed` and such in the injection context. `inject` should run once something (a component, service, provider etc.) is initialized and not getting the injectable continuously during the "something" its lifetime. But I do want to support using a value from an injectable. I got a devtools branch locally which allows injectables to be used as such:

```ts
export const appConfig: ApplicationConfig = {
  providers: [
    provideTanStackQuery(
      new QueryClient(),
      withDevtools(
        (devToolsOptionsManager: DevtoolsOptionsManager) => ({
          loadDevtools: devToolsOptionsManager.loadDevtools(),
        }),
        {
          // `deps` can be used to pass one or more injectables as parameters to the `withDevtools` callback.
          deps: [DevtoolsOptionsManager],
        },
      ),
    ),
  ],
}
```

It's similar to the Angular `deps` mechanism in providers. Like with Angular `deps` the parameter does need explicit typing. What do you think about this API?

It is possible to add type inference to the Angular adapter's types btw, but because of TypeScript limitations it would need a separate overload per number of `deps` array items. Which is kind of clunky and why I choose not to add that for now.

The branch also adds conditional exports which greatly improves tree shaking and other devtools improvements. Haven't touched it in a month, I'll review it a bit and see if I can open a PR.

michaelbe812:
> I don't think a computation function should use `inject`, which is why I removed `runInInjectionContext`, forgetting that it breaks the use case and my own example code.
> 
> There's good reasons Angular doesn't run `computed` and such in the injection context. `inject` should run once something (a component, service, provider etc.) is initialized and not getting the injectable continuously during the "something" its lifetime. But I do want to support using a value from an injectable. I got a devtools branch locally which allows injectables to be used as such:
> 
> ```ts
> export const appConfig: ApplicationConfig = {
>   providers: [
>     provideTanStackQuery(
>       new QueryClient(),
>       withDevtools(
>         (devToolsOptionsManager: DevtoolsOptionsManager) => ({
>           loadDevtools: devToolsOptionsManager.loadDevtools(),
>         }),
>         {
>           // `deps` can be used to pass one or more injectables as parameters to the `withDevtools` callback.
>           deps: [DevtoolsOptionsManager],
>         },
>       ),
>     ),
>   ],
> }
> ```
> 
> It's similar to the Angular `deps` mechanism in providers. Like with Angular `deps` the parameter does need explicit typing. What do you think about this API?
> 
> It is possible to add type inference to the Angular adapter's types btw, but because of TypeScript limitations it would need a separate overload per number of `deps` array items. Which is kind of clunky and why I choose not to add that for now.
> 
> The branch also adds conditional exports which greatly improves tree shaking and other devtools improvements. Haven't touched it in a month, I'll review it a bit and see if I can open a PR.

Hey I generally agree with everything especially the part regarding `computed`. 

For me personally this API looks fine. I do not see any major advantages or disadvantages of having another syntax. It does the job and it should be quite understandable.

arnoud-dv:
Cool thanks, gives me confidence to move forward with this. API design, thinking of many use cases and avoiding future breaking changes is hard 😅

michaelbe812:
> Cool thanks, gives me confidence to move forward with this. API design, thinking of many use cases and avoiding future breaking changes is hard 😅

Agree - This is a complete different discipline :)

PR #9241, opened 2025-06-03
fix: expand callback return types to support Promise<Array<void>> for Promise.all() usage

braden-w:
This PR expands the callback return types from `Promise<void> | void` to `void | Promise<void> | Promise<Array<void>>` to address user feedback from #9202.

PR #9202 narrowed callback types from `Promise<unknown> | unknown` to `Promise<void> | void` to better reflect the intent that callback return values are ignored. However, this change broke existing user code that was using a common pattern of returning `Promise.all()` from callbacks.

## Problem

The change broke code like this:
```ts
const useCreateProduct = () => {
  return useMutation({
    mutationFn: (createProductDto) => api.products.create(createProductDto),
    onSuccess: (data) => 
      Promise.all([
        queryClient.invalidateQueries({ queryKey: ['products'] }),
        queryClient.invalidateQueries({ queryKey: ['categories'] })
      ]),
  });
};
```

Since `Promise.all()` returns `Promise<Array<void>>`, this pattern was no longer type-compatible.

## Solution

This PR adds `Promise<Array<void>>` to the union type, allowing the common `Promise.all()` pattern while maintaining the intent that individual callback return values are ignored (they're all variations of `void`).

**Updated type signature:**
```ts
onSuccess?: (...) => void | Promise<void> | Promise<Array<void>>
onError?: (...) => void | Promise<void> | Promise<Array<void>>
onSettled?: (...) => void | Promise<void> | Promise<Array<void>>
```

## Changes Made

- ✅ Updated callback return types in `MutationOptions` interface
- ✅ Updated callback return types in `MutationCacheConfig` interface  
- ✅ Updated callback return types in persist client packages
- ✅ Updated documentation to reflect new types
- ✅ Fixed Promise chain handling in persist clients to support new return types

## Addresses

- Feedback from @Sagie501 in #9202
- Feedback from @jgarplind in #9202  
- Common `Promise.all()` usage pattern in mutation callbacks

Fixes #9202 (user feedback)


braden-w:
Actually, scratch this PR: we should instead change the return types to just be `void`. Consider #9245 instead.

PR #9247, opened 2025-06-04
Make hydrationQueue a derived value

Ephem:
After the fixes in:

https://github.com/TanStack/query/pull/9157
https://github.com/TanStack/query/pull/9188

I noticed a version of this bug was still happening: https://github.com/TanStack/query/issues/8677

@TkDodo Pointed out a potential solution, which is to move `hydrationQueue` from being a state, to being a derived value instead.

I think this makes a lot of sense. It should always be safe to recalculate this and there should be no situation where the component rerenders without the effect running where we want to preserve the values.

Tests are passing without changes.

### DIFF
--- a/packages/react-query/src/HydrationBoundary.tsx
+++ b/packages/react-query/src/HydrationBoundary.tsx
@@ -56,6 +56,10 @@ export const HydrationBoundary = ({
         }
 
         const queryCache = client.getQueryCache()
+        // State is supplied from the outside and we might as well fail
+        // gracefully if it has the wrong shape, so while we type `queries`
+        // as required, we still provide a fallback.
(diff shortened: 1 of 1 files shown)


TkDodo (approved):
awesome, exactly how I imagined it would look ♥️. Gets a lot simpler too 🙌

TkDodo:
ah, the eslint disable is still needed it seems:

```
/home/workflows/workspace/packages/react-query/src/HydrationBoundary.tsx
  59:25  error  Unnecessary conditional, value is always truthy  @typescript-eslint/no-unnecessary-condition
```

Ephem:
Yeah, my editor linting was acting up and told me it was safe to remove. 😞

PR #9198, opened 2025-05-27
docs: restore react/mutations.md ducumentation

ssu2030:
This PR updates the `mutations.md` on react-guides. While working on [this](https://github.com/TanStack/query/commit/afc6302dbfbf902de470a98c44ec2452f4f6a94b) commit, I noticed that part of the original explanation had been omitted, so I added the missing details and pushed the update.

PR #9254, opened 2025-06-06
docs: Replace accidentally-deleted content in mutations.md

christina-de-martinez:
I was reading about mutations and was surprised when the guide ended mid-sentence. It looks like part of the content was deleted during a fairly large, maybe partially automated commit: https://github.com/TanStack/query/commit/afc6302dbfbf902de470a98c44ec2452f4f6a94b#diff-45f03e5c0424761a02a9cd83196c99693f26087f2e9b5d265c0eea53fa3bcaad 

This PR assumes that you want this part of the guide put back. Feel free to close this if that's not the case :)

### DIFF
--- a/docs/framework/react/guides/mutations.md
+++ b/docs/framework/react/guides/mutations.md
@@ -182,4 +182,233 @@ useMutation({
 
 [//]: # 'Example5'
 
-You might find that you want to **trigger additional callbacks** beyond the ones defined on `useMutation` when calling `mutate`. This can be used to trigger component-specific side effects. To do that, you can provide any of the same callback options to the `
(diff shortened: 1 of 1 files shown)


TkDodo:
this is the same PR as:
- https://github.com/TanStack/query/pull/9198

but I’m not getting any response there.

My feedback is that you please have to make sure that the links work. The accidental removal happened when we changed how all the links are built due to changes in TanStack Start. So if we just restore what we had before, the links are probably not working.

I’ll merge this regardless but I’d appreciate if you could file another PR that fixes the links on this page.

christina-de-martinez:
Right, sorry, here's a follow-up PR that should fix those: https://github.com/TanStack/query/pull/9260 @TkDodo

PR #9261, opened 2025-06-08
Updates examples for invalidation from mutations

srph:
If my memory serves me right, returning a Promise makes sure the data is updated before the mutation is entirely complete (i.e., `isPending` is true until onSuccess is fulfilled)

This updates the code examples to reflect that. Please correct me if I'm mistaken.


TkDodo:
“it depends” on if you want that or not 🤷

PR #9267, opened 2025-06-11
ci: add link verification step to PR workflow

manudeli:
image


package.json
```diff
@@ -29,7 +29,7 @@
     "prettier:write": "pnpm run prettier --write",
     "docs:generate": "node scripts/generateDocs.js",
     "cipublish": "node scripts/publish.js",
-    "verify-links": "node scripts/verify-links.ts"
+    "verify-links": "pnpm tsx scripts/verify-links.ts"
```
manudeli, line 32: We should use `tsx` because the [node version(22.12.0) defined in .nvmrc](https://github.com/TanStack/query/blob/main/.nvmrc) is lower than the one that supports TypeScript natively.

.github/workflows/pr.yml
```diff
@@ -35,6 +35,8 @@ jobs:
           main-branch-name: main
       - name: Run Checks
         run: pnpm run test:pr --parallel=3
+      - name: Verify Links
```
KevinVandy, line 38: I wonder if this should be added to `autofix.yml` either instead or in addition to here. Autofix will commit the fixes to the branch.

manudeli: Since this `verify-links.ts` only detects broken links but does not automatically fix them, I think it's more appropriate to keep it in `ci.yml` rather than `autofix.yml`. But please feel free to share if you had something else in mind

package.json
TkDodo: We can also update node :)

manudeli: Cool, I'll update it with removing tsx

PR #9269, opened 2025-06-11
docs: Fix syntax of batshit in batching-requests article

u9g:
See quickstart heading here: https://www.npmjs.com/package/@yornaath/batshit for correct syntax


TkDodo:
this is exactly the reason why in v5, we’ve moved away from showing how something works and instead just link to community projects: https://tanstack.com/query/v5/docs/framework/react/community/community-projects

PR #9270, opened 2025-06-11
feat(angular-query): move devtools to conditional sub-paths

### LINKED ISSUE #8705: Angular: support providing TanStack Query and `withDevtools` on lazy loaded routes
There's a wish for large enterprise applications to not add TanStack Query to the main application bundle but only on lazy loaded routes.

`provideTanStackQuery` and feature providers such as `withDevtools` should be usable on lazy loaded routes providers array.

As `withDevtools` was first intended to be used only in `app.config.ts` it currently adds new devtools every time it's provided.

### LINKED ISSUE #8824: Angular: `withDevtools` option `loadDevtools` no longer supports dependency injection
### Describe the bug

Based on the documentation, you should be able to use inject within the `loadDevtools` option in the `withDevtools` function. 
See https://github.com/TanStack/query/blob/main/docs/framework/angular/devtools.md

Now if you do this, you will get an `NG0203` error.

Breaking PR: https://github.com/TanStack/query/pull/8817

https://noop.com

### Steps to reproduce

1. Initialize angular with tanstack query
2. Write a simple service with a boolean signal that will determine if the devtool is active
3. Try injecting it in the "loadDevtools" option.

### Expected behavior

As a developer I expected the devtools not to throw an NG0203 error

Every time

### Screenshots or Videos

_No response_

### Platform

Any browser

### Tanstack Query adapter

angular-query

### TanStack Query version

v5.68.1

### TypeScript version

v5.8.2

### Additional context

_No response_

### LINKED ISSUE #9078: Angular: Esbuild always generates chunks for devtools
### Describe the bug

The Angular Query devtools are lazy loaded to optimize startup performance of the app while developing and to minimize bundle sizes. 

However, even if the code that performs the lazy load is tree shaken, esbuild wil generate unnessary chunks adding to application deployment size.

https://tanstack.com/query/latest/docs/framework/angular/examples/basic

### Steps to reproduce

- Go to the linked Stackblitz
- Remove `withDevtools` from `app.config.ts`
- Open the terminal and run `npm run build`
- The Angular CLI will now perform an optimized production build
- Note several lazy chunk files are generated even though there is no reference to `withDevtools` in the application

### Expected behavior

No chunks should be generated for devtools if they are not used

Every time

### Screenshots or Videos

_No response_

### Platform

N/a

### Tanstack Query adapter

angular-query

### TanStack Query version

v5.74.7

### TypeScript version

_No response_

### Additional context

PR in progress

arnoud-dv:
#### withDevtools moved to subpath

Most applications, to completely tree shake devtools from production builds
```ts
import { withDevtools } from '@tanstack/angular-query-experimental/devtools'
```

If you want to also use devtools in production builds
```ts
import { withDevtools } from '@tanstack/angular-query-experimental/devtools/production'
```

#### injectDevtoolsPanel moved to main package

Most applications, to completely tree shake devtools panel from production builds
```ts
import { injectDevtoolsPanel } from '@tanstack/angular-query-experimental/devtools-panel'
```

If you want to also use devtools panel in production builds
```ts
import { injectDevtoolsPanel } from '@tanstack/angular-query-experimental/devtools-panel/production'
```

### angular-query-devtools-experimental package

After moving injectDevtoolsPanel to the main package, the angular devtools package is obsolete.

### Subpath exports
Experience with subpath exports paves the way to add functionality such as RxJs integration without needing to have the whole Angular TanStack Query package depend on RxJs.

Isolating devtools to a subpath reliably fixes #9078, where esbuild was generating chunks even when devtools were not used at all. While these chunks were not adding to the main bundle size it unnecessarily increased build time and deployment size.

### Conditional exports
The devtools [subpath exports](https://nodejs.org/api/packages.html#subpath-exports) further utilize [conditional exports](https://nodejs.org/api/packages.html#conditional-exports). For development builds, the function is exported normally. For production builds a tiny stub function is exported, resulting in excellent tree shaking without having to configure environments.

### Dependency injection in the devtools callback
Fixes #8824 After no longer running the options computation function in the injection context it was no longer possible to use dependency injection to inject values. The withDevtools function now has a `deps` option to automatically inject dependencies and pass it as a parameter to the computation function. This solves this issue while not injecting a value each time the computation function is run which would be the case if the function would run in the injection context. This is similar to and internally uses `deps` on Angular's [`useFactory`](https://angular.dev/guide/di/dependency-injection-providers#factory-providers-usefactory).

```ts
withDevtools(
  (devToolsOptionsManager: DevtoolsOptionsManager) => ({
    loadDevtools: devToolsOptionsManager.loadDevtools(),
  }),
  {
    // `deps` is used to inject and pass `DevtoolsOptionsManager` to the `withDevtools` callback.
    deps: [DevtoolsOptionsManager],
  },
)
```

### Adds a check if devtools were already provided
Fixes #8705

## Summary by CodeRabbit

- **New Features**
  - Devtools & Devtools Panel: production-aware, DI-friendly integrations with lazy/on-demand loading, production stubs, and a new hideDisabledQueries option; withDevtools supports DI-enabled factory callbacks.

- **Documentation**
  - Large docs refresh with autogenerated notices, updated devtools guidance/examples, inline source pointers, and cleanup of obsolete entries.

- **Examples**
  - Angular examples updated to import devtools from the new devtools module paths.

- **Chores / Breaking Changes**
  - Devtools made optional; legacy devtools package removed; public exports and import paths relocated; several public type signatures/overloads adjusted.


ThoSap:
I see this is in active development, thank you for your work @arnoud-dv 💪🏼

PR #9275, opened 2025-06-13
docs: add Chrome extension information for TanStack Query devtools across multiple frameworks

dborysov:


TkDodo:
anything you can do about the permissions? This reads quite scary:

Screenshot 2025-06-19 at 11 25 36

TkDodo:
also, is the extension itself open-source?

TkDodo:
also, FYI, we now dispatch custom events from the devtools that you could potentially pick up instead of relying on a global queryClient?

https://github.com/TanStack/query/blob/3e7d715718adf5ff17df68b1c0b3b9c2c0ff6f2d/packages/query-devtools/src/Devtools.tsx#L2524-L2529

dborysov:
Hi @TkDodo,

> anything you can do about the permissions?

This warning is triggered because of `"matches": ["<all_urls>"]` in the content script configuration. Unfortunately, I didn't find any other way to access the client's JS context without these broad permissions. Also, I noticed that other similar extensions like `React DevTools`, `Vue DevTools`, `Redux DevTools`, `Apollo DevTools` require similar permissions

> also, is the extension itself open-source?

Not yet, but I'm planning to open source it somewhere today or tomorrow

> also, FYI, we now dispatch custom events from the devtools that you could potentially pick up instead of relying on a global queryClient?

I assume that it would mean that in order to use this extension, the user would first need to install the devtools to their app? If so, I would prefer not to follow this path for 2 reasons:
1. It sounds weird to me that the user needs the npm-based devtools in order to make chrome-based devtools work
2. I would like to support not only the most popular frameworks like Angular, React, etc., but also things like Lit or even vanilla js. But installing the devtools in such cases is not really straight forward

TkDodo:
> I assume that it would mean that in order to use this extension, the user would first need to install the devtools to their app?

yeah that’s  right, they would

> Not yet, but I'm planning to open source it somewhere today or tomorrow

let me know once that’s done please 👍

dborysov:
@TkDodo, Done ✅ - https://github.com/DeeCode-inc/tanstack-query-chrome-devtools

PR #9278, opened 2025-06-14
fix(react-query): update Incorrect Inference of GetDefinedOrUndefinedQueryResult when TData is assignable to CallableFunction

ViewableGravy:
# Overview
This PR resolves an issue that was brought up in discord where using `queryOptions` where TQueryFnData was `{ name: string }`, would cause the type of TData on `useQueries` to be `{ name: string }` instead of `{ name: string } | undefined`.

Upon further investigation into what was causing this, it looks like because `InitialData` is assignabled to `{ name: string }` when it is a function (since name is a key on CallableFunction in typescript), the type would assume the wrong pathway in the `GetDefinedOrUndefinedQueryResult` type, resulting in the incorrect return type.

Changing the initial condition from `T extends { initialData?: ... }` to `T extends { initialData: ... }` allows us to avoid hitting that pathway altogether and gives the correct type.

## Tests
I have updated the tests to ensure this case is correctly tested against, and ensured that existing tests are not affected. I have also slightly restructured the tests specifically testing `queryOptions` under a describe for better code organization.

## Original Issue
Discord Thread: https://discord.com/channels/719702312431386674/1382540005624512532


ViewableGravy:
I can see that my `pnpm.lock` file shows a large number of changes - however I've ensured that I am using the correct version of pnpm and made sure to run `pnpm install` directly from the root of the project, so I am unsure why this is the case against master.

Please let me know if I should revert the lock file change, as this should not affect this PR with or without it.

Noting that running pnpm build:all also fails against vue - this change only affects react-query

pnpm-lock.yaml
TkDodo, line 1: there should be no changes to the lock file!

TkDodo:
> however I've ensured that I am using the correct version of pnpm 

you were definitely using the wrong version, I can see that `lockfileVersion` changed:

```diff
- lockfileVersion: '9.0'
+ lockfileVersion: '6.0'
```

so yeah, definitely the wrong version of `pnpm`. Please revert, not sure why you’d even commit lockfile changes if you didn’t touch package.json 🤷

pnpm-lock.yaml
ViewableGravy: Sorry about this one, I've reverted the change now

ViewableGravy:
> > however I've ensured that I am using the correct version of pnpm
> 
> you were definitely using the wrong version, I can see that `lockfileVersion` changed:
> 
> ```diff
> - lockfileVersion: '9.0'
> + lockfileVersion: '6.0'
> ```
> 
> so yeah, definitely the wrong version of `pnpm`. Please revert, not sure why you’d even commit lockfile changes if you didn’t touch package.json 🤷

Sorry about the time waste here. I just shouldn't have committed it in the first place. I've reverted that change now

TkDodo (approved):

packages/react-query/src/__tests__/useQueries.test-d.tsx
```diff
@@ -41,23 +41,93 @@ describe('UseQueries config object overload', () => {
...
+      type Data = {
+        [K in keyof CallableFunction]: {
+          [key in K]: CallableFunction[key]
+        }
+      }[keyof CallableFunction]
+
+      // Object with individual key from CallableFunction
+      const queryOptions1 = queryOptions({
+        queryKey: ['key1'],
+        queryFn: async (): Promise<Data> => {
```
TkDodo:
```suggestion
        queryFn: (): Promise<Data> => {
```

packages/react-query/src/__tests__/useQueries.test-d.tsx
```diff
@@ -41,23 +41,93 @@ describe('UseQueries config object overload', () => {
...
+          return {
+            name: 'example',
+          }
+        },
+      })
+
+      // Object with an actual CallableFunction
+      const queryOptions2 = queryOptions({
+        queryKey: ['key2'],
+        queryFn: async (): Promise<CallableFunction> => {
```
TkDodo:
```suggestion
        queryFn: (): Promise<CallableFunction> => {
```

packages/react-query/src/__tests__/useQueries.test-d.tsx
```diff
@@ -41,23 +41,93 @@ describe('UseQueries config object overload', () => {
...
+          return () => void 0
+        },
+      })
+
+      // Object with matching key from CallableFunction, but different value type
+      const queryOptions3 = queryOptions({
+        queryKey: ['key3'],
+        queryFn: async (): Promise<{ name: number }> => {
```
TkDodo:
```suggestion
        queryFn: (): Promise<{ name: number }> => {
```

TkDodo (commented):
addressing some lint errors

TkDodo:
we now have a bunch of type errors:

```
../react-query/src/__tests__/useQueries.test-d.tsx(96,13): error TS2322: Type '{ name: string; }' is not assignable to type 'Promise<Data>'.
  Object literal may only specify known properties, and 'name' does not exist in type 'Promise<Data>'.
```

ViewableGravy:
> we now have a bunch of type errors:
> 
> ```
> ../react-query/src/__tests__/useQueries.test-d.tsx(96,13): error TS2322: Type '{ name: string; }' is not assignable to type 'Promise<Data>'.
>   Object literal may only specify known properties, and 'name' does not exist in type 'Promise<Data>'.
> ```

Didn't get around to this one on the weekend, I'll look into this during the week.

ViewableGravy:
This pull request was automatically closed by a bot because I don't currently have the time or expertise to resolve this. It can be reopened in the future if someone works out how to fix it, assuming it hasn't already been fixed.

PR #9281, opened 2025-06-17
refactor(angular-query): migrate from tsup to vite

arnoud-dv:
Allows for more flexibility in build process.

Adds prepack and postpack scripts to publish type declaration files in package root, a prerequisite to support subpath exports while also supporting applications using `"modeResolution": "node"`.

Fixes moduleResolution: node16


hakimio:
@arnoud-dv the production bundle seems to have incorrect `types` config.

image

arnoud-dv:
@hakimio do you mean it should be

```json
"types": "./index.d.ts"
```

hakimio:
@arnoud-dv no, I mean that `index.d.ts` no longer exists in the root dir - it was moved to `dist/types`, but the production bundle `package.json` was not updated correctly (like shown in the screenshot above).

arnoud-dv:
There's a prepack lifecycle script that makes sure the type declarations are linked from the package root at the moment the package is published. I want to add sub path exports and to support the still widely used `"moduleResolution": "node"`, type declarations files should follow the same path structure.

hakimio:
Ok, let's do it step-by-step:
1. download the latest `@tanstack/angular-query-experimental` package from npm:
```
npm pack @tanstack/angular-query-experimental
```
2. Open the archive and inspect its contents.
3. Does it have `index.d.ts` in the root dir? **No.**
4. Now open `package.json` and check what it says:
```json
"exports": {
    ".": {
      "types": "./index.d.ts",
      "default": "./dist/index.mjs"
    }
  }
```

It points to non-existant `index.d.ts` file.

arnoud-dv:
I see - I'll revert it for now.

Tested this very well, preview package from the PR build was working perfectly on all supported Angular versions and when looking at its files contents it was also good. But probably the actual publish script behaviour is different. I'll look at this later.

Thanks for reporting the issue!

PR #9293, opened 2025-06-20
fix(query-core): we don't want errors that get reverted internally to throw on imperative methods

TkDodo:
to achieve that, we let the retryer's `onError` "transform" the result back to potentially TData, and if so, we'll resolve the promise to that instead; that means calls to `fetchQuery` will be "successful" when a cancellation happens in the meantime, but they will then resolve to the reverted data; also, if the initial fetch is cancelled, we would still throw, as there is no data to revert to.

### DIFF
--- a/packages/query-core/src/__tests__/query.test.tsx
+++ b/packages/query-core/src/__tests__/query.test.tsx
@@ -198,6 +198,45 @@ describe('query', () => {
     }
   })
 
+  test('should not throw a CancelledError when fetchQuery is in progress and the last observer unsubscribes when AbortSignal is consumed', async () => {
+    const key = queryKey()
+
+    const observer = new QueryObserver(queryClient, {
+      queryKey: key,
+      queryFn: async () => {
+        await sleep(100)
+        return 'data'
+      },
+    })
+
+    const unsubscribe = observer.subscribe(() => undefined)
+    await vi.advanceTimersByTimeAsync(100)
+
+    expect(queryCache.find({ queryKey: key })?.state.data).toBe('data')
+
+    const promise = queryClient.fetchQuery({
+      queryKey: key,
+      queryFn: async ({ signal }) => {
+        await sleep(100)
+        return 'data2' + String(signal)
+      },
+    })
+
+    // Ensure the fetch is in progress
+    await vi.advanceTimersByTimeAsync(10)
+
+    // Unsubscribe while fetch is in progress
+    unsubscribe()
+    // await queryClient.cancelQueries()
+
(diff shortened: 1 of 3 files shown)


packages/query-core/src/__tests__/hydration.test.tsx
```diff
@@ -1398,8 +1398,5 @@ describe('dehydration and rehydration', () => {
     // Need to await the original promise or else it will get a cancellation
     // error and test will fail
     await originalPromise
-
-    clientQueryClient.clear()
-    serverQueryClient.clear()
```
TkDodo, line 1403: @Ephem we now get another `CancelledError` here if we clear the caches. They are local to this test, so clearing them shouldn’t be necessary, but I’m wondering where this is coming from in the first place 🤔

packages/query-core/src/__tests__/queryClient.test.tsx
```diff
@@ -1050,7 +1050,7 @@ describe('queryClient', () => {
         queryKey: key1,
         queryFn: () => 'data',
       })
-      queryClient.fetchQuery({
+      queryClient.prefetchQuery({
```
TkDodo, line 1052: this now throws a `CancelledError` because we call `cancelQueries` with `revert:false`, and since we don’t revert, the error makes it to the cache. It did so before too, but since the refactor to async/await, it shows up as unhandled in node.

packages/react-query/src/__tests__/useQuery.test.tsx
```diff
@@ -3036,7 +3036,8 @@ describe('useQuery', () => {
       expect(rendered.getByText('error: some error')).toBeInTheDocument(),
     )
 
-    expect(count).toBe(4)
+    // 1 original fetch + 2 retries
+    expect(count).toBe(3)
```
TkDodo: I thought hard about this but it was wrong before 😂

DamianOsipiuk (approved):

packages/query-core/src/index.ts
```diff
@@ -26,7 +26,6 @@ export {
   shouldThrowError,
 } from './utils'
 export type { MutationFilters, QueryFilters, Updater, SkipToken } from './utils'
-export { isCancelledError } from './retryer'
```
DamianOsipiuk: Isn't this a breaking change? Should we just deprecate for now?

TkDodo: huh, I think it was exported from the query-core only so that the framework adapters can use it. But you’re right, maybe this was used by someone to ignore cancelled errors, so let’s bring it back and deprecate it:

06e2f180

PR #9318, opened 2025-06-27
refactor(query-core/queryObserver): use async/await in 'fetchOptimistic' and 'fetch'

sukvvon:


manudeli:
I don't think there's a strict need to use the async/await syntax.

PR #9321, opened 2025-06-27
fix(react-query): throw type error when skipToken is present in prefetch query

tmkx:
discussion here: https://github.com/TanStack/query/pull/9301#issuecomment-3013376004


TkDodo (approved):
awesome, thanks

PR #9328, opened 2025-06-29
test(query-core/mutations): fix typo - remove duplicate "should run" in mutation test name

ienrum:
Fixed a typo in the test description by removing duplicate "should run" text.


manudeli (approved):
👀🦅 👍

PR #9334, opened 2025-06-29
feat(react-query): backport v5 apis about infinite query

manudeli:
### New Features and Utilities:
* Added `infiniteQueryOptions`
* Added `useSuspenseInfiniteQuery`
* Added type `DefinedInitialDataInfiniteOptions`, type `UndefinedInitialDataInfiniteOptions`
* Added function overload for `useInfiniteQuery` to return `DefinedUseInfiniteQueryResult` when using `infiniteData`

### Deprecations:
* Marked several overloads of `useInfiniteQuery` as deprecated, with a note that they will be removed in the next major version. (`packages/react-query/src/useInfiniteQuery.ts`)


packages/react-query/src/useInfiniteQuery.ts
```diff
@@ -2,14 +2,39 @@
...
+    TData,
+    TQueryFnData,
+    TQueryKey
+  > & {
+    initialData:
+      | NonUndefinedGuard<InfiniteData<TQueryFnData>>
+      | (() => NonUndefinedGuard<InfiniteData<TQueryFnData>>)
+      | undefined
+  },
+): DefinedUseInfiniteQueryResult<TData, TError>
```
manudeli, line 37: Another function overload that returns DefinedUseInfiniteQueryResult is added when using initialData.

@tanstack/react-query v5 also support this

packages/react-query/src/types.ts
```diff
@@ -120,6 +121,19 @@ export type UseInfiniteQueryResult<
...
+  TError = unknown,
+> = DefinedInfiniteQueryObserverResult<TData, TError>
+
+export type UseSuspenseInfiniteQueryResult<
+  TData = unknown,
+  TError = unknown,
+> = OmitKeyof<
+  DefinedInfiniteQueryObserverResult<TData, TError>,
+  'isPlaceholderData'
+>
```
manudeli, line 135: new types

packages/react-query/src/useInfiniteQuery.ts
```diff
@@ -24,6 +49,7 @@ export function useInfiniteQuery<
     TQueryKey
   >,
 ): UseInfiniteQueryResult<TData, TError>
+/** @deprecated This function overload will be removed in the next major version. */
```
manudeli, line 52: Mark some useInfiniteQuery's function overloads as deprecated

Issue #9336, opened 2025-06-29
useQuery does not perform a retry on mount when the function provided to throwOnError returns false, even if retryOnMount: true is set


andromansky:
### Describe the bug

If the query options include the following:

```
throwOnError: () => false,  
retryOnMount: true,  
staleTime: Infinity,
```

then on remount of the host component, `queryFn` is not called, even though the documentation suggests [otherwise](https://tanstack.com/query/latest/docs/framework/react/reference/useQuery#:~:text=If%20set%20to%20false%2C%20the%20query%20will%20not%20be%20retried%20on%20mount%20if%20it%20contains%20an%20error.%20Defaults%20to%20true.).

### Your minimal, reproducible example

https://stackblitz.com/edit/react-93scwtvo

### Steps to reproduce

1. Wait for the app to load and open the console.
2. Click the "**Open Modal**" button — you will see "_Fetching..._" messages in the console indicating data fetch attempts.
3. Click the "**Close Modal**" button.
4. Click "**Open Modal**" again — no more "_Fetching..._" messages appear.

### Expected behavior

I expect that if a Query is in the `error` status and I set `retryOnMount: true `in the options, then when the function passed to `throwOnError` returns `false`, the Query should still call `queryFn` on remount of the host component.

### How often does this bug happen?

Every time

### Platform

- OS: Windows, macOS;
- Browser: Chrome (137.0.7151.120);

### Tanstack Query adapter

react-query

### TanStack Query version

v5.81.4

### TypeScript version

v5.8.3

PR #9338, opened 2025-06-30
fix(react-query): allow retryOnMount when throwOnError is function (#9336)

Han5991:
## Summary

  - Fixed throwOnError retry logic to properly handle function vs boolean values
  - When throwOnError is a function, allow retryOnMount to proceed even when error boundary hasn't been reset
  - Maintains existing prevention behavior for boolean throwOnError values
  - Added comprehensive test coverage for the retry behavior

### DIFF
--- a/packages/react-query/src/errorBoundaryUtils.ts
+++ b/packages/react-query/src/errorBoundaryUtils.ts
@@ -28,11 +28,14 @@ export const ensurePreventErrorBoundaryRetry = <
 ) => {
   if (
     options.suspense ||
-    options.throwOnError ||
     options.experimental_prefetchInRender
   ) {
     // Prevent retrying failed query if the error boundary has not been reset yet
     if (!errorResetBoundary.isReset()) {
+      options.retryOnMount = false
+    }
+  } else if (options.throwOnError) {
+    if (!errorResetBoundary.isReset() && typeof options.throwOnError !== 'function') {
       options.retryOnMount = false
     }
   }
--- a/packages/react-query/src/__tests__/useQuery.test.tsx
+++ b/packages/react-query/src/__tests__/useQuery.test.tsx
@@ -6027,6 +6027,7 @@ describe('useQuery', () => {
     it('should be able to toggle subscribed', async () => {
       const key = queryKey()
       const queryFn = vi.fn(() => Promise.resolve('data'))
+
       function Page() {
         const [subscribed, setSubscribed] = React.useState(true)
         const { data } = useQuery({
@@ -6069,6 +6070,7 @@ describe('useQuery', () => {
     it('should not be attached to the query when subscribed is false', async () => {
       const key = queryKey()
       const queryFn = vi.fn(() => Promise.resolve('data'))
+
       function Page() {
         const { data } = useQuery({
           queryKey: key,
@@ -6095,6 +6097,7 @@ describe('useQuery', () => {
     it('should not re-render when data is added to the cache when subscribed is false', async () => {
       const key = queryKey()
       let renders = 0
+
       function Page() {
         const { data } = useQuery({
           queryKey: key,
@@ -6297,6 +6300,7 @@ describe('useQuery', () => {
       await sleep(5)
       return { numbers: { current: { id } } }
     }
+
     function Test() {
       const [id, setId] = React.useState(1)
 
@@ -6357,6 +6361,7 @@ describe('useQuery', () => {
       await sleep(5)
       return { numbers: { current: { id } } }
     }
+
     function Test() {
       const [id, setId] = React.useState(1)
 
@@ -6854,10 +6859,12 @@ describe('useQuery', () => {
   it('should console.error when there is no queryFn', () => {
     const consoleErrorMock = vi.spyOn(console, 'error')
     const key = queryKey()
+
     function Example() {
       useQuery({ queryKey: key })
       return <></>
     }
+
     renderWithClient(queryClient, <Example />)
 
     expect(consoleErrorMock).toHaveBeenCalledTimes(1)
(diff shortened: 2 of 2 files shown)


packages/react-query/src/errorBoundaryUtils.ts
```diff
@@ -28,13 +28,16 @@ export const ensurePreventErrorBoundaryRetry = <
...
     options.suspense ||
-    options.throwOnError ||
     options.experimental_prefetchInRender
   ) {
     // Prevent retrying failed query if the error boundary has not been reset yet
     if (!errorResetBoundary.isReset()) {
       options.retryOnMount = false
     }
+  } else if (options.throwOnError) {
+    if (!errorResetBoundary.isReset() && typeof options.throwOnError !== 'function') {
```
TkDodo: The function can return true or false so I think we'd need to evaluate it to get the result

Han5991: Updated to evaluate the throwOnError function result instead of just checking if it's a function. Thanks for the feedback!

Han5991:
@TkDodo Hi! Simple fix for retryOnMount when throwOnError is function.
Would appreciate a review when you have time! 🙏

packages/react-query/src/useBaseQuery.ts
```diff
@@ -53,6 +53,14 @@ export function useBaseQuery<
...
   const client = useQueryClient(queryClient)
   const defaultedOptions = client.defaultQueryOptions(options)
+  const query = client
+    .getQueryCache()
+    .get<
+      TQueryFnData,
+      TError,
+      TQueryData,
+      TQueryKey
+    >(defaultedOptions.queryHash)
```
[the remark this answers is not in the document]
TkDodo: yes please move this after `beforeQuery` instead of inlining another call into `ensurePreventErrorBoundaryRetry`

packages/react-query/src/errorBoundaryUtils.ts
TkDodo, line 1: I think we can simplify this quite a bit:

```
const throwOnError =
  query.state.error && typeof options.throwOnError === 'function'
    ? shouldThrowError(options.throwOnError, [query.state.error, query])
    : options.throwOnError
if (
  options.suspense ||
  options.experimental_prefetchInRender ||
  throwOnError
) {
  // Prevent retrying failed query if the error boundary has not been reset yet
  if (!errorResetBoundary.isReset()) {
    options.retryOnMount = false
  }
}
```

packages/react-query/src/errorBoundaryUtils.ts
```diff
@@ -25,16 +25,24 @@ export const ensurePreventErrorBoundaryRetry = <
     TQueryKey
   >,
   errorResetBoundary: QueryErrorResetBoundaryValue,
+  query?: Query<TQueryFnData, TError, TQueryData, TQueryKey>,
```
TkDodo: can we make this required?

Han5991:
Hi @TkDodo, thank you for the review! I've applied your suggestions:
- Moved the query retrieval after _experimental_beforeQuery in useBaseQuery.
- Simplified ensurePreventErrorBoundaryRetry logic and made the query parameter required as requested.
- Updated useQueries to safely handle the required query parameter.
All tests are passing. Ready for another look!

packages/react-query/src/errorBoundaryUtils.ts
```diff
@@ -25,11 +25,17 @@ export const ensurePreventErrorBoundaryRetry = <
     TQueryKey
   >,
   errorResetBoundary: QueryErrorResetBoundaryValue,
+  query: Query<TQueryFnData, TError, TQueryData, TQueryKey>,
```
TkDodo: just seeing this now: to be on par with `getHasError`, we should make this param required, but potentially undefined, and then do a single check in here. sorry about changing my mind here 😅 


```suggestion
  query: Query<TQueryFnData, TError, TQueryData, TQueryKey> | undefined,
```

Han5991: Updated the code as requested based on the review comment.

packages/react-query/src/useBaseQuery.ts
```diff
@@ -72,8 +80,9 @@ export function useBaseQuery<
     : 'optimistic'
 
   ensureSuspenseTimers(defaultedOptions)
-  ensurePreventErrorBoundaryRetry(defaultedOptions, errorResetBoundary)
-
+  if (query) {
+    ensurePreventErrorBoundaryRetry(defaultedOptions, errorResetBoundary, query)
+  }
```
TkDodo: the guard should be unnecessary now

```suggestion
  ensurePreventErrorBoundaryRetry(defaultedOptions, errorResetBoundary, query)
```

packages/react-query/src/useQueries.ts
```diff
@@ -242,9 +242,12 @@ export function useQueries<
...
 
-  defaultedQueries.forEach((query) => {
-    ensureSuspenseTimers(query)
-    ensurePreventErrorBoundaryRetry(query, errorResetBoundary)
+  defaultedQueries.forEach((queryOptions) => {
+    ensureSuspenseTimers(queryOptions)
+    const query = client.getQueryCache().get(queryOptions.queryHash)
+    if (query) {
+      ensurePreventErrorBoundaryRetry(queryOptions, errorResetBoundary, query)
+    }
```
TkDodo: same here:

```suggestion
    const query = client.getQueryCache().get(queryOptions.queryHash)
    ensurePreventErrorBoundaryRetry(queryOptions, errorResetBoundary, query)
```

Issue #9272, opened 2025-06-13
[Windows] symlink doesn't work yet


ss0526100:
### Describe the bug

Although #9053 claims this issue was resolved, I believe it remains unresolved.

The mentioned  PR #9042 in this issue is only a solution for TS resolve in a monorepo, but the problem of accessing symlinks on Windows still persists.

For example, react-query/root.eslint.config.js is a symlink, which Windows cannot resolve, making ESLint application difficult. (In the image below, the `.` is flagged as an unexpected token because Windows interprets symlinks like `../../eslint.config.js`, so the leading `.` becomes an unexpected token.)

![Image](https://github.com/user-attachments/assets/46b3a26b-f99a-45e7-9bdf-28cca91bacab)

I think there are a few possible solutions:

1. Officially state in CONTRIBUTING.md that tanstack-query should be developed in a Linux-based environment.
2. Separate code used by sub-repositories, like eslint.config.js or files in the scripts folder, into a multi-repo structure.
3. Copy the parent folder's files into each package individually (though this requires significant changes when updating parent files).

There might be other creative solutions I’m unaware of.

Since this issue remains unresolved, I’m opening a related issue.

### Your minimal, reproducible example

every files on windows

### Steps to reproduce

.

### Expected behavior

Windows developers can work on tanstack-query without issues without running a virtual Linux environment.

Alternatively, ensure new contributors to tanstack-query are aware they need to set up a virtual Linux environment.

### How often does this bug happen?

Every time

### Platform

Windows

### Tanstack Query adapter

None

### TanStack Query version

latest

PR #9346, opened 2025-07-02
chore:Remove symlink in react query

ss0526100:
As mentioned in #9272, there were issues with configuration imports through symlinks.

While thinking about this problem, I came across the implementation of the react-router library. I found that this library solves the issue by simply importing with relative paths.
[example](https://github.com/remix-run/react-router/blob/main/packages/react-router-dom/tsup.config.ts)

When I applied this approach to react-query and tested it on Windows, I confirmed that both ESLint and tsup work well.

If this PR gets merged, we should be able to remove other symlinks using the same approach as this PR. And it will enable development of tanstack query on Windows as well!


TkDodo (changes requested):
we have the symlink in all other frameworks too so if we want to go that route, it would need to be addressed the same everywhere please, not just in the react adapter.

ss0526100:
Thanks for your comment, @TkDodo! 
I opened #9410, and I'll close this PR!

PR #9355, opened 2025-07-04
test(react-query/usePrefetchQuery): remove 'vi.waitFor' and add 'advanceTimersByTimeAsync'

sukvvon:


manudeli (commented):
`generateQueryFn` should use `sleep` imported from `@tanstack/query-test-utils` instead of `vi.advanceTimersByTimeAsync`

image

manudeli (approved):

PR #9357, opened 2025-07-04
test(react-query/useSuspenseQuery): switch to fake timers and replace 'waitFor' with 'vi.waitFor'

sukvvon:


manudeli (approved):
Let's replace vi.waitFor with vi.advanceTimersByTimeAsync after this pull request merged

Issue #1633, opened 2021-01-12
setQueryData slow with large data sets


andrewnaeve:
**Describe the bug**
This may be a performance question, or maybe I need a paradigm shift. 

I have a large query that comes back with several mb worth of data, for a table that gets frequently sorted. When I sort the data, and call `queryClient.setQueryData()` with the new sorted rows, it takes around half a second to complete. I've tried to break the data out into React state, and when I do that, it sorts in 10ms- but getting React state to sync nicely with React Query is challenging.

In my profiler, I see quite a bit of time being taken by "replaceEqualDeep".

I'm wondering if I'm using react-query incorrectly. Should I be replacing the cached query data like this? Are there any better strategies for large datasets?

**Screenshots**
<img width="646" alt="calltree" src="https://user-images.githubusercontent.com/16847584/104270894-6db9af80-544e-11eb-903f-a4d6ec073867.png">

**Desktop (please complete the following information):**
- MacOS, latest FIrefox and Chrome
- "react-query": "^3.5.5"

TkDodo:
This might be due to react-query's `structural sharing` feature. You can turn that off by setting `structuralSharing: false` on your query.

however, I would advise _against_ mixing client state with server state and manipulating the query in the cache directly. The idea is that the data you see is just a borrowed "view" of the actual server state, so the client does not own it. If a background update happens, or another component mounts that also uses this query, the data will be re-fetched and you will "loose" your client side sorting.

I think a good alternative would be to keep server and client state separate. In your case, that would mean:
- keep the data that you get from the server in the queryClient and do _not_ change it.
- only store the sorting "choice" that the user made in react state on the frontend.
- compute the new data in the render function, possibly with `useMemo`, combining those two states.

This could be nicely abstracted in a custom hook, something like:

```
const useBigData = () => {
    const queryInfo = useQuery('bigData', () => fetchBigData())
    const [sorting, setSorting] = React.useState(undefined) // accepts 'foo' | 'bar' as sorting keys

  
   return {
        query: {
            ...queryInfo,
            data: React.useMemo(() =>{
                return sortedDataDependingOnStoredSorting(queryInfo.data, sorting)
            }, [queryInfo.data, sorting])
        },
        setSorting,
        sorting,
   }
}
```

You can use this like so in your component:

```
const { query, sorting, setSorting } = useBigData()

<Button onClick={() => setSorting('foo)}>Sort by Foo</Button>
[...]

PR #9392, opened 2025-07-08
Perf: optimize query cache and structural sharing

### LINKED ISSUE #6783: `setQueryData` is slow with large datasets

jiji-hoon96:
Optimized setQueryData performance by eliminating O(n) cache lookups and improving replaceEqualDeep algorithm for large datasets. (Fixes #6783, #1633)

### QueryCache Optimization

```ts
// Before: O(n) linear search
const query = queryCache.find(options.queryKey)

// After: O(1) hash map access  
const query = queryCache.get(queryHash)
```

### replaceEqualDeep Algorithm Improvements

- Added early reference equality checks
- Implemented chunked processing for large arrays
- Added WeakMap memoization for circular references
- Optimized object key iteration

### Memory Management

- Reused hash computations
- Reduced object allocations in hot paths
- Added cleanup for WeakMap entries

### DIFF
--- a/packages/query-core/src/utils.ts
+++ b/packages/query-core/src/utils.ts
@@ -250,10 +250,27 @@ export function partialMatchKey(a: any, b: any): boolean {
  * If not, it will replace any deeply equal children of `b` with those of `a`.
  * This can be used for structural sharing between JSON values for example.
  */
+// WeakMap for memoization to avoid processing the same objects multiple times
+const replaceCache = new WeakMap<object, WeakMap<object, any>>();
+
 export function replaceEqualDeep<T>(a: unknown, b: T): T
 export function replaceEqualDeep(a: any, b: any): any {
+  // Early return for primitive equality (optimization for most common case)
   if (a === b) {
     return a
+  }
+  
+  // Early return for non-objects to avoid unnecessary processing
+  if (a === null || b === null || typeof a !== 'object' || typeof b !== 'object') {
+    return b
+  }
+
+  // Use memoization for already processed object pairs
+  if (replaceCache.has(a)) {
+    const cachedResult = replaceCache.get(a)?.get(b);
+    if (cachedResult !== undefined) {
+      return cachedResult;
+    }
   }
 
   const array = isPlainArray(a) && isPlainArray(b)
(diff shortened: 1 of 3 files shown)


sukvvon:
@jiji-hoon96 

https://github.com/TanStack/query/blob/main/CONTRIBUTING.md#commit-message-format

IMO, above docs can help you naming PR title!

packages/query-core/src/utils.ts
```diff
@@ -250,11 +250,28 @@ export function partialMatchKey(a: any, b: any): boolean {
  * If not, it will replace any deeply equal children of `b` with those of `a`.
  * This can be used for structural sharing between JSON values for example.
  */
+// WeakMap for memoization to avoid processing the same objects multiple times
+const replaceCache = new WeakMap<object, WeakMap<object, any>>();
```
ss0526100, line 254: This might be a misguided point due to my limited understanding, but I have some concerns about using replaceCache as you suggested:

0. I understand that replaceEqualDeep is a utility function that returns a copy of b, but returns a if a and b are deeply equal.

1. From what I know, replaceEqualDeep is used when replacing existing data in a query with new data. So, I think it would be rare for existing data to change to different values multiple times.
For example, if existing data changes from A to B as a result of queryFn, in most cases the reference to the previous A would be lost, making caching meaningless.

2. As mentioned in point 1, replaceEqualDeep is used to replace existing data in a query with new data. Even if we receive values with the same structure through fetching in most cases, the references of the two values being compared would be different. Therefore, it seems rare to call get(b) again on the weakMap returned from replaceCache.get(a).

3. I believe utility functions should maintain purity. If multiple queryClients are running, replaceCache operations from one queryClient might affect the behavior of another queryClient. (Of course, based on the current implementation, affecting other clients doesn't seem to cause major issues.) So, for replaceCaching, I think it would be better to create separate replaceCache objects for each queryClient or queryObserver rather than using a utility function.

If my understanding is incorrect or if you have different thoughts, please let me know your feedback :)

jiji-hoon96: Thanks for the thoughtful feedback! You’re absolutely right about the purity concerns and global state issues.

However, I believe there’s still value here based on real usage patterns - optimistic updates, form synchronization, and incremental data changes are extremely common and would benefit from this optimization. The performance impact in issue #1633 (500ms → 50ms) demonstrates real user pain points.

Let me revise the approach: remove the global WeakMap and implement QueryClient-scoped memoization as an opt-in feature. This addresses your architectural concerns while still providing performance benefits for users who need them.

What do you think about this compromise?​​​​​​​​​​​​​​​​

ss0526100: I think the issue you mentioned has already been resolved. The setQueryData is already using get instead of find.

([PR that removed find from setQueryData](https://github.com/TanStack/query/pull/6918/files))

If test cases were included, we could verify that performance improvements can be achieved through the code in the PR.

Also, I don't think the WeakMap-based cache would work even with optimistic updates, form synchronization, and incremental data changes.

As I mentioned in my previous comment, it's hard to imagine fetching an object with the same reference value as a discarded object that has already been updated.

(Though I suppose it might be meaningful when applying the same reference-valued object created on the client to replaceEqualDeep multiple times)

jiji-hoon96:
@ss0526100  You're absolutely right - the core performance issues were already resolved in PR #6918, and my proposed WeakMap optimizations would add complexity without meaningful real-world benefits given typical object reference patterns.
Closing this PR as the optimization is unnecessary. Thank you for the thorough technical review! 🙏

PR #9400, opened 2025-07-10
fix: improve type safety for structural sharing with selected data

jiji-hoon96:
Currently, there is a TypeScript type error when using the structuralSharing and select options together in useQuery. This issue occurs because the structuralSharing function expects the original query function's return type (TQueryFnData), but at runtime, it actually receives the transformed data type (TData) returned by the select function.

### DIFF
--- a/packages/query-core/src/queryObserver.ts
+++ b/packages/query-core/src/queryObserver.ts
@@ -525,7 +525,11 @@ export class QueryObserver<
         try {
           this.#selectFn = options.select
           data = options.select(data as any)
+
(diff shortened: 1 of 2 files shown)


TkDodo (approved):
yeah I think unknown is the correct type. It’ll actually run with both `TQueryFnData` and `TQueryData` when you use `select`, and we use it in a bunch of other places too I think

TkDodo:
oh, sorry, I misread what the PR does. `unknown` is correct as I’ve stated above.

PR #9410, opened 2025-07-11
chore:remove symlink for windows

ss0526100:
This PR continues the context of #9346.

Resolved Windows symbolic link issue (#9272) across all packages.

- Changed all symbolic links to relative paths.
- Unified eslint-plugin-query's separate configuration to relative path imports (see comments).
- Source map files were generated during Windows build in scripts/tsup.config.js and the root folder's eslint.config.js. Added generated source map file information to .gitignore to resolve this.

### DIFF
--- a/packages/eslint-plugin-query/tsup.config.ts
+++ b/packages/eslint-plugin-query/tsup.config.ts
@@ -16,6 +16,11 @@ export default defineConfig([
 
       return
     },
+    esbuildPlugins: undefined,
   },
-  { ...legacyConfig({ entry: ['src/*.ts'] }), external: ['typescript'] },
+  {
+    ...legacyConfig({ entry: ['src/*.ts'] }),
+    external: ['typescript'],
+    esbuildPlugins: undefined,
+  },
 ])


packages/eslint-plugin-query/tsup.config.ts
```diff
@@ -1,7 +1,26 @@
...
+
+// @ts-ignore out of scope
+import { legacyConfig, modernConfig } from '../../scripts/getTsupConfig.js'
 
 export default defineConfig([
-  modernConfig({ entry: ['src/*.ts'] }),
-  legacyConfig({ entry: ['src/*.ts'] }),
+  {
+    ...modernConfig({ entry: ['src/*.ts'] }),
+    external: ['typescript'],
```
ss0526100, line 9: In eslint-plugin-query, applying the parent-level esbuildPlugin settings directly caused build failures.
To address this, previous versions used a modified copy of scripts/getTsupConfig.js instead of the original.
Considering potential future changes to getTsupConfig, this issue needed to be handled.
Thus, scripts/getTsupConfig.js was imported using a relative path, and settings specific to eslint-plugin-query were added separately.

TkDodo (commented):
thinking about this some more - we already have a package called `query-test-utils` that exports some utils for testing so that we don’t have to duplicate them for various adapters.

couldn’t we put the shared configs there too and just import them through the workspace dependency ?

ss0526100:
Got it. Similar to method 2 in #9272.

I'll implement it that way.

arnoud-dv:
Using symlinks makes it less likely for dependencies such as Vite to resolve to the wrong version as opposed to importing stuff outside the workspace. I often encountered this:

- Update Angular which uses Vite internally
- Suddenly vue-query gets type errors because of conflicting Vite and Vitest versions

This is very time consuming and distracting.

Perhaps using a dedicated package with its own dependencies will be a good solution but let's keep this in mind. I think the current symlinks are an elegant method to resolve dependencies such as Vite to the proper workspace version.

It's been a while since I developed on Windows, but I think it supports symlinks very well on:
- Windows Subsystem for Linux  / WSL
- devdrive
- Less convenient, but NTFS when configuring this Git setting: `git config --global core.symlinks true`. For some reason Windows requires an account with elevated permissions to create symlinks on NTFS though.

ss0526100:
You're right. I'm also developing in a Windows environment, but I'm able to write code without issues by using WSL.

However, if we keep the existing symlinks, we'll need to update the CONTRIBUTING.md document. A few people, including myself, have had problems when first trying to modify Tanstack Query's code on Windows.

This seems to be a matter of choice:

- Stick with the current code.
  - In this case, the CONTRIBUTING.md document should recommend against developing on enviroments that don't support symlinks.
- Use this Pull Request (PR) as is.
- Create a separate package for shared configurations.

TkDodo:
superseded by:

- #9451

PR #9412, opened 2025-07-12
fix(query-core): set options before checking experimental_prefetchInR…

birkskyum:
In this PR

- https://github.com/TanStack/router/pull/3893

I'm facing the error, in the /e2e/solid-start/basic-solid-query project in the Posts and Users pages:

`Cannot read properties of undefined (reading 'experimental_prefetchInRender')`

I believe it's because it's checking the property on the options object before setting the options object on the next line.

This swaps that around


birkskyum:
The issue appear to be that the client isn't passed to the QueryObserver constructor for solid-query

PR #9438, opened 2025-07-15
test(angular-query-experimental/inject-mutation-state): replace 'runAllTimersAsync', 'advanceTimersByTime' with 'advanceTimersByTimeAsync'

sukvvon:


packages/angular-query-experimental/src/__tests__/inject-mutation-state.test.ts
```diff
@@ -126,11 +126,12 @@ describe('injectMutationState', () => {
         return [
           injectMutation(() => ({
             mutationKey: mutationKey1,
-            mutationFn: () => Promise.resolve('myValue'),
+            mutationFn: () => sleep(10).then(() => Promise.resolve('myValue')),
```
manudeli:
```suggestion
            mutationFn: () => sleep(10).then(() => 'myValue'),
```

PR #9446, opened 2025-07-17
test(react-query/useInfiniteQuery): use precise time in 'advanceTimersByTimeAsync', and add 'expect' using 'toBeInTheDocument'

sukvvon:


packages/react-query/src/__tests__/useInfiniteQuery.test.tsx
```diff
@@ -81,7 +81,7 @@ describe('useInfiniteQuery', () => {
 
     renderWithClient(queryClient, <Page />)
 
-    await vi.advanceTimersByTimeAsync(100)
+    await vi.advanceTimersByTimeAsync(0)
```
manudeli:
```suggestion
    await vi.advanceTimersByTimeAsync(10)
```

Could you add queryFn like below

```tsx
    function Page() {
      const state = useInfiniteQuery({
        queryKey: key,
        queryFn: ({ pageParam }) => sleep(10).then(() => Number(pageParam)),
        getNextPageParam: (lastPage) => lastPage + 1,
        initialPageParam: 0,
      })
      states.push(state)
      return null
    }
```

manudeli (changes requested):

packages/react-query/src/__tests__/useInfiniteQuery.test.tsx
sukvvon: @manudeli 

I reflected your opinion. 👍

packages/react-query/src/__tests__/useInfiniteQuery.test.tsx
```diff
@@ -1422,7 +1422,7 @@ describe('useInfiniteQuery', () => {
     function Page() {
       const state = useInfiniteQuery({
         queryKey: key,
-        queryFn: ({ pageParam }): number => pageParam,
+        queryFn: ({ pageParam }) => sleep(10).then(() => pageParam),
```
manudeli, line 1425: 👍

packages/react-query/src/__tests__/useInfiniteQuery.test.tsx
```diff
@@ -1383,7 +1383,7 @@ describe('useInfiniteQuery', () => {
...
```
manudeli, line 1386: 👍

packages/react-query/src/__tests__/useInfiniteQuery.test.tsx
```diff
@@ -1808,20 +1808,18 @@ describe('useInfiniteQuery', () => {
...
 
     expect(cancelFn).toHaveBeenCalled()
   })
 
   it('should use provided custom queryClient', async () => {
     const key = queryKey()
-    const queryFn = async () => {
-      return Promise.resolve('custom client')
-    }
+    const queryFn = () => sleep(10).then(() => 'custom client')
```
manudeli, line 1822: 👍

packages/react-query/src/__tests__/useInfiniteQuery.test.tsx
```diff
@@ -335,7 +336,7 @@ describe('useInfiniteQuery', () => {
     function Page() {
       const state = useInfiniteQuery({
         queryKey: key,
-        queryFn: () => ({ count: 1 }),
+        queryFn: () => sleep(10).then(() => ({ count: 1 })),
```
manudeli, line 339: 👍

packages/react-query/src/__tests__/useInfiniteQuery.test.tsx
```diff
@@ -71,7 +71,7 @@ describe('useInfiniteQuery', () => {
     function Page() {
       const state = useInfiniteQuery({
         queryKey: key,
-        queryFn: ({ pageParam }) => Number(pageParam),
+        queryFn: ({ pageParam }) => sleep(10).then(() => Number(pageParam)),
```
manudeli, line 74:
```suggestion
        queryFn: ({ pageParam }) => sleep(10).then(() => pageParam),
```

https://tanstack.com/query/latest/docs/framework/react/guides/migrating-to-v5#infinite-queries-now-need-a-initialpageparam

This seems to be a leftover from TanStack Query v4! Now, TanStack Query v5 has improved this type of thing and I think this part should be improved incrementally throughout the entire code.
We can remove unnecessary code like this. I would appreciate it if you could take care of this gradually too!

manudeli (approved):
Much tighter and more realistic now 👍

packages/react-query/src/__tests__/useInfiniteQuery.test.tsx
sukvvon: > This seems to be a leftover from TanStack Query v4! Now, TanStack Query v5 has improved this type of thing and I think this part should be improved incrementally throughout the entire code. We can remove unnecessary code like this. I would appreciate it if you could take care of this gradually too!

@manudeli 

I totally agree with you. I’ll take care of this part as well in future tasks.

PR #9455, opened 2025-07-18
test(react-query/useSuspenseQuery): use precise time in 'advanceTimersByTimeAsync', and add 'expect' using 'toBeInTheDocument'

sukvvon:


packages/react-query/src/__tests__/useSuspenseQuery.test.tsx
```diff
@@ -69,12 +69,12 @@ describe('useSuspenseQuery', () => {
       </React.Suspense>,
     )
 
-    await act(() => vi.advanceTimersByTimeAsync(11))
-    rendered.getByText('data: 1')
+    await act(() => vi.advanceTimersByTimeAsync(10))
```
manudeli, line 72: tight 👍

PR #9456, opened 2025-07-18
test(react-query/QueryResetErrorBoundary): use precise time in 'advanceTimersByTimeAsync', and add 'expect' using 'toBeInTheDocument'

sukvvon:


packages/react-query/src/__tests__/QueryResetErrorBoundary.test.tsx
```diff
@@ -590,16 +590,16 @@ describe('QueryErrorResetBoundary', () => {
         </QueryErrorResetBoundary>,
       )
 
-      await act(() => vi.advanceTimersByTimeAsync(11))
-      rendered.getByText('error boundary')
-      rendered.getByText('retry')
+      await act(() => vi.advanceTimersByTimeAsync(10))
```
manudeli, line 593: Tight 👍

PR #9466, opened 2025-07-19
test(react-query/useQueries): replace inline 'Promise' delays with 'sleep', and add 'expect' using 'toBeInTheDocument'

sukvvon:


packages/react-query/src/__tests__/useQueries.test.tsx
```diff
@@ -1233,20 +1233,13 @@ describe('useQueries', () => {
...
-              }, 10)
-            })
+          queryFn: async () => {
+            if (id === 2) {
+              await sleep(10)
+              return Promise.reject(new Error('FAILURE'))
+            }
+            await sleep(10)
+            return Promise.resolve({ id, title: `Post ${id}` })
           },
```
manudeli: This is just a small thing I wanna make like this.

```suggestion
          queryFn: () => sleep(10).then(() => {
            if(id === 2) throw new Error('FAILURE')
            return { id, title: `Post ${id}` }
          }),
```

manudeli (commented):
Simple is best for me, if possible 😁

packages/react-query/src/__tests__/useQueries.test.tsx
sukvvon: @manudeli 

I agree with you. I gonna reflect your suggestion. 👍

manudeli (approved):

PR #9493, opened 2025-07-24
fix(svelte-query): `state_unsafe_mutation` error with `useIs...`

hmnd:
Fixes the bug reported by @frederikhors [here](https://github.com/TanStack/query/discussions/7413#discussioncomment-13308487).

I thoroughly dug into this issue before I found calls to `Observer` and `MutataionOBserver` were unnecessarily wrapped in a `$derived(...)`. There is already an `$effect.pre(...)` used for updating options when they changed, so this shouldn't be necessary.

I've also added proper support for reactively updating queries when the passed in `queryClient` changes. The `queryClient` was already reactive previously, but was broken by my move away from `$derived`. Now, the observer is a `$state` and is updated whenever the client changes.


hmnd:
@lachlancollins should be fixed now :)

hmnd:
> Not sure if it worked, it's still showing all the conflicts

Oh sorry, `git fetch` only pulled from my fork, not upstream :facepalm: 

One sec

pnpm-lock.yaml
lachlancollins, line 1: Can you reset the lockfile to the target branch? There shouldn't be any changes in this PR given it doesn't change any dependencies (and it's also killing CI).

lachlancollins (changes requested):

pnpm-lock.yaml
hmnd: Oops, just noticed myself too. Also fixed!

hmnd: and just fixed all the unused import errors showing up in tests

packages/svelte-query/vite.config.ts
```diff
@@ -25,5 +25,6 @@ export default defineConfig({
     setupFiles: ['./tests/test-setup.ts'],
     coverage: { enabled: false, provider: 'istanbul', include: ['src/**/*'] },
     typecheck: { enabled: true },
+    globals: true,
```
lachlancollins: Can you please remove this? I've confirmed vitest still runs without it. Also, there are a few eslint errors (one was actually in the target branch which I've just fixed).

hmnd: Any idea what those Comment errors are? 
image

I'm seeing a space after every `//`, so I don't know what it wants

lachlancollins: Yeah it's weird, but if you run `pnpm test:eslint --fix`, they all go away!

hmnd: Do they on your end? Keep getting the same errors even after fix...

hmnd: No clue what it wants :thinking: 

Eg. `containers.svelte.ts` `:65` shouldn't have a `;`??
image
image

packages/svelte-query/src/createMutation.svelte.ts
```diff
@@ -24,48 +23,62 @@ export function createMutation<
...
+  const mutate = <CreateMutateFunction<TData, TError, TVariables, TContext>>((
+    variables,
+    mutateOptions,
+  ) => {
+    observer.mutate(variables, mutateOptions).catch(noop)
   })
 
-  onDestroy(() => {
-    unsubscribe()
+  let result = $derived(observer.getCurrentResult())
```
hmnd: @lachlancollins do you think we need to keep the Svelte peer dep pinned to as low as `5.7.0`, or could we bump that to at least `5.25.0` [so that assigning to `$derived` like this is supported](https://svelte.dev/docs/svelte/$derived#Overriding-derived-values)? This was part of `eslint --fix`, but if we need to stay on the lower version, I'll have to revert this back to a manually managed `$state`.

lachlancollins: I'm happy to bump the peer dependency version!

lachlancollins: @hmnd I've updated this on the target branch.

packages/svelte-query/vite.config.ts
lachlancollins: I'm happy to ignore eslint for now, it seems to be parsing the files incorrectly. Interestingly it works if you run it with the `--fix` option.

packages/svelte-query/src/createMutation.svelte.ts
lachlancollins: It seems like this change(?) might have caused tests to start failing. It's difficult to know though, as the force pushes have erased the history.

lachlancollins: Nevermind, I found the diff in a branch I cloned earlier. Tests are all passing again, so I think I'll merge this! Thanks heaps for your time @hmnd !

hmnd: Thank you for getting this merged @lachlancollins! Sorry about the force pushes :sweat_smile:

PR #9494, opened 2025-07-25
test(svelte-query/createQueries): simplify 'queryFn' and add test for success when queries reslove at different times

sukvvon:


packages/svelte-query/tests/createQueries/createQueries.test.ts
```diff
@@ -21,29 +21,22 @@ describe('createQueries', () => {
...
             },
           ],
         },
         queryClient: new QueryClient(),
       },
     })
 
-    await vi.advanceTimersByTimeAsync(0)
     expect(rendered.getByText('Status 1: pending')).toBeInTheDocument()
     expect(rendered.getByText('Status 2: pending')).toBeInTheDocument()
```
manudeli, line 38: If queryFn's sleeping time is different, it would be better to add test code for the case after just 10ms, not only this.

packages/svelte-query/tests/createQueries/createQueries.test.ts
```diff
@@ -21,29 +21,22 @@ describe('createQueries', () => {
...
-              },
+              queryFn: () => sleep(10).then(() => 'Success 1'),
             },
             {
               queryKey: ['key-2'],
-              queryFn: async () => {
-                await sleep(5)
-                return 'Success 2'
-              },
+              queryFn: () => sleep(20).then(() => 'Success 2'),
```
manudeli: If queryFn's sleeping time is different 10ms, 20ms, it would be better to add test code for the case after just 10ms, not only for case after 20ms.

packages/svelte-query/tests/createQueries/createQueries.test.ts
sukvvon: > If queryFn's sleeping time is different, it would be better to add test code for the case after just 10ms, not only this.

@manudeli 

I understand what you said. After this PR is merged, I’ll try some tests as you suggested.

packages/svelte-query/tests/createQueries/createQueries.test.ts
sukvvon: > If queryFn's sleeping time is different 10ms, 20ms, it would be better to add test code for the case after just 10ms, not only for case after 20ms.

@manudeli 

That makes sense. I’ll add the test as you suggested.

packages/svelte-query/tests/createQueries/createQueries.test.ts
```diff
@@ -21,44 +21,64 @@ describe('createQueries', () => {
...
+
+    expect(rendered.getByText('Status 1: pending')).toBeInTheDocument()
+    expect(rendered.getByText('Status 2: pending')).toBeInTheDocument()
+
+    await vi.advanceTimersByTimeAsync(11)
+    expect(rendered.getByText('Status 1: success')).toBeInTheDocument()
+    await vi.advanceTimersByTimeAsync(10)
+    expect(rendered.getByText('Status 2: success')).toBeInTheDocument()
+  })
+
```
sukvvon, line 71: > If queryFn's sleeping time is different 10ms, 20ms, it would be better to add test code for the case after just 10ms, not only for case after 20ms.

@manudeli 

I added the test as you suggested.

manudeli (approved):

PR #9496, opened 2025-07-25
test(svelte-query/{queryOptions,infiniteQueryOptions}): add 'eslint-disable' for 'vitest/expect-expect'

sukvvon:


packages/svelte-query/tests/infiniteQueryOptions/infiniteQueryOptions.test-d.ts
```diff
@@ -5,6 +5,7 @@ import { createInfiniteQuery, infiniteQueryOptions } from '../../src/index.js'
 import type { InfiniteData } from '@tanstack/query-core'
 
 describe('queryOptions', () => {
+  // eslint-disable-next-line vitest/expect-expect
```
manudeli, line 8: [assertType of vitest](https://vitest.dev/api/assert-type) could be better for this case
Could you check that?

sukvvon: @manudeli 

```ts
  test('Should not allow excess properties', () => {
    const options = infiniteQueryOptions({
      queryKey: ['key'],
      queryFn: () => Promise.resolve('data'),
      getNextPageParam: () => 1,
      initialPageParam: 1,
      // @ts-expect-error this is a good error, because stallTime does not exist!
      stallTime: 1000,
    })

    expectTypeOf(options).toEqualTypeOf<
      CreateInfiniteQueryOptions<
        string,
        Error,
        InfiniteData<string, unknown>,
        Array<string>,
        number
      >
    >()
  })
```

How about this?

manudeli: @mmkal (expect-type member) @sheremet-va @antfu (vitest member) 

I used to think [`assertType`](https://vitest.dev/api/assert-type.html#asserttype) in vitest was the best way to type test for cases like this, but I'm not entirely sure now. What is the best way to handle vitest/expect-expect when testing for these ts-expect-errors with vitest eslint? Could you share your thoughts?

mmkal: You could do something like

```ts
expectTypeOf(infiniteQueryOptions).toBeCallableWith({
  queryKey: ['foo'],
  ...
})
```

Which I believe vitest/expect-expect will allow, and it's pretty clear what the purpose of the test is that way too.

If that's still a linter error, I believe the expect-expect rule can be configured to allow expectTypeOf - although this would be worth allowing by default in vitest since it's built-in.

manudeli: Thanks for sharing your thought! I will try it❤️

manudeli: @mmkal I tried toBeCallable but it seems like `expectTypeOf().toBeCallableWith` isn't always doing what I'd expect compared to just calling the function directly 🥲
Could this be a bug in expect-type, or is it just a TypeScript limitation?

https://github.com/user-attachments/assets/ca866f75-c713-4979-91ea-b19de1a17f71

mmkal: Ah, yes, it's using `extends` so won't do the excess property check - and yes that's a typescript limitation that probably can't be worked around in expect-type.

You can still have the test for the excess property if you go more explicit:

```ts
expectTypeOf(infiniteQueryOptions).toBeCallableWith({ 
  queryKey: ['key'],
  queryFn: () => Promise.resolve('data'),
  getNextPageParam: () => 1,
  initialPageParam: 1,
})
expectTypeOf(infiniteQueryOptions).parameter(0).not.toHaveProperty('stallTime')
```

That second assertion does what you're looking for - if the function parameter allowed `[key: string]: unknown` it would fail. But understandable if you don't want to go so all-in on expect-type in this change.

benpsnyder:
Hey! Heads up — #10355 (`ci: add oxlint as fast pre-lint pass before ESLint`) covers this PR's intent.

With the oxlint integration, `vitest/expect-expect` is now handled by oxlint at `warn` level, and `eslint-plugin-oxlint`'s `buildFromOxlintConfig()` automatically disables the overlapping ESLint rule. So ESLint will no longer fire `vitest/expect-expect` on these type-level test files — the `eslint-disable` comments this PR adds would be unnecessary once #10355 lands.

I'd recommend closing this in favor of #10355 once it's merged.

TkDodo:
thanks, we’ll likely go ahead with #10355 so going to close this.

PR #9500, opened 2025-07-26
feat(devtools): add setting to hide disabled queries

jonluca:
This PR adds a new setting to the Query Devtools package that allows for filtering out disabled queries. 

CleanShot 2025-07-25 at 17 06 47@2x

### DIFF
--- a/packages/react-query-devtools/src/ReactQueryDevtools.tsx
+++ b/packages/react-query-devtools/src/ReactQueryDevtools.tsx
@@ -42,6 +42,10 @@ export interface DevtoolsOptions {
    * Use this so you can attach the devtool's styles to specific element in the DOM.
    */
   shadowDOMTarget?: ShadowRoot
+  /**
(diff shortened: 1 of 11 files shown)


packages/react-query-devtools/src/ReactQueryDevtools.tsx
```diff
@@ -42,6 +42,10 @@ export interface DevtoolsOptions {
    * Use this so you can attach the devtool's styles to specific element in the DOM.
    */
   shadowDOMTarget?: ShadowRoot
+  /**
+   * Set this to true to hide disabled queries from the devtools panel.
+   */
+  hideDisabled?: boolean
```
TkDodo: can we rename this to `hideDisabledQueries` ? It’s kinda ambiguous with the current name.

jonluca: done!

TkDodo (approved):

antoshek (approved):

PR #9502, opened 2025-07-26
ci: introduce changesets

lachlancollins:
## 🎯 Changes

As discussed on Discord (private channel): https://discord.com/channels/719702312431386674/1345158793675276319

See also: https://github.com/TanStack/config/pull/218 and https://github.com/TanStack/virtual/pull/946

## ✅ Checklist

- [x] I have followed the steps in the [Contributing guide](https://github.com/TanStack/config/blob/main/CONTRIBUTING.md).
- [x] I have tested this code locally with `pnpm test:pr`.

## 🚀 Release Impact

## Summary by CodeRabbit

- Chores
  - Adopted Changesets for versioning, changelogs, and publishing; added Changesets config, scripts, and dev tooling.
  - Removed a custom publish script and simplified package metadata for an internal integration package.

- Documentation
  - Updated CONTRIBUTING to a Changesets-based release workflow and clarified test/setup guidance.
  - Added a PR template with change summary, checklist, and release-impact guidance.

- CI
  - Updated release workflow to run on push with staged versioning, automatic commit of version files, and publish steps via Changesets.


.github/workflows/release.yml
```diff
@@ -1,11 +1,6 @@
-name: ci
+name: Release
 
 on:
-  workflow_dispatch:
-    inputs:
-      tag:
-        description: override release tag
-        required: false
```
TkDodo, line 8: I think we need this to trigger manual releases from the github actions workflow

TkDodo: or rather, we “needed” this. how can I force publish a release with changesets only?

CONTRIBUTING.md
```diff
@@ -237,7 +167,7 @@ Use an appropriate commit type. Be especially careful with breaking changes.
 
 ## Releases
 
-For each new commit added to `main` with `git push` or by merging a pull request or merging from another branch, a GitHub action is triggered and runs the `semantic-release` command to make a release if there are codebase changes since the last release that affect the package functionalities.
+For each new commit added to `main`, a GitHub Workflow is triggered which runs the [Changesets Action](https://github.com/changesets/action). This generates a preview PR showing the impact of all changesets. When this PR is merged, the package will be published to NPM.
```
TkDodo, line 170: this sounds like more work for maintainers. Previously, I just merged a PR with `fix:` or `feat:` prefix in the commit message (which I could alter while squash merging if someone didn’t do that correctly) and it would auto-trigger a release. Now I have to:

- generate a changeset or remind people to do that
- wait for the preview PR to show up
- merge that, too

.github/workflows/release.yml
lachlancollins: Yes you can do this with changesets alone!

CONTRIBUTING.md
lachlancollins: Hey @TkDodo , yes it's definitely an extra step, maybe this wasn't explicitly addressed in that Discord thread. I'll still advocate for this change as I believe the downsides are outweighed by the benefits, but happy to discuss further on Discord!

CONTRIBUTING.md
```diff
@@ -155,79 +155,9 @@ You can use Gitpod (An Online Open Source VS Code like IDE which is free for Ope
...
-| ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------- |
-| fix(pencil): stop graphite breaking when too much pressure applied                                                                                                                                | Patch Release              |
-| feat(pencil): add `graphiteWidth` option                                                                                                                                                          | ~~Minor~~ Feature Release  |
-| perf(pencil): remove `graphiteWidth` option<br/><br/>BREAKING CHANGE: The `graphiteWidth` option has been removed.<br/>The default graphite width of 10mm is always used for performance reasons. | ~~Major~~ Breaking Release |
-
-### Revert
-
-If the commit reverts a previous commit, it should begin with `revert:`, followed by the header of the reverted commit. In the body it should say: `This reverts commit <hash>.`, where the hash is the SHA of the commit being reverted.
```
sukvvon, line 230: @lachlancollins 

Hello,
I noticed that the `commit message conventions` section in the `CONTRIBUTING.md` appears to be removed in this update, so I wanted to check.
Could you please share if there is any particular reason or background for removing this section?
Since it seems to be an important part of maintaining the team’s commit message conventions, I’d like to better understand.
Thank you!

lachlancollins: Hey @sukvvon , the commit messages are currently more important because they directly influence versioning logic. With changesets, the commit messages no longer affect versioning. I'm happy to keep in a section about writing good commit messages, but I don't think it needs to be as extensive!

sukvvon: > Hey @sukvvon , the commit messages are currently more important because they directly influence versioning logic. With changesets, the commit messages no longer affect versioning. I'm happy to keep in a section about writing good commit messages, but I don't think it needs to be as extensive!

@lachlancollins 

https://github.com/pmndrs/zustand/blob/main/CONTRIBUTING.md#committing

How about referring to the link above? It’s a brief version I wrote before, and I think it can meet our needs!

AhmedBaset:
There is one disadvantage of using Changesets over the current setup, though I'm not sure if it's a valid point for you to consider here. With the current setup, people subscribing to GitHub releases receive a single notification/a single release on the releases page. With Changesets, however, there would be tons. A "RELEASE ALL" in the query repo would make 23 releases and send 23 notifications to subscribers.

Screenshot_20250922_122955_Chrome

lachlancollins:
> There is one disadvantage of using Changesets over the current setup, though I'm not sure if it's a valid point for you to consider here. With the current setup, people subscribing to GitHub releases receive a single notification/a single release on the releases page. With Changesets, however, there would be tons. A "RELEASE ALL" in the query repo would make 23 releases and send 23 notifications to subscribers.

Hey @AhmedBaset , it's definitely quite noisy. One "solution" is to disable release notifications by choosing custom watch events. If you want to keep these notifications though, I don't think there is a way to merge them together, sorry!

image

CONTRIBUTING.md
lachlancollins: Hey @sukvvon , I think it's probably a good idea to revisit the contributing guide in a separate issue/PR, as I'm sure it could do with a more extensive update. I'm going to resolve this here, but I would like it to be continued!

PR #9532, opened 2025-08-05
feat(query-core): add custom reducer support to streamedQuery

marcog83:
Replace maxChunks parameter with flexible reducer function that delegates data aggregation to consumer code. This provides full control over how streamed chunks are combined into the final data structure.

Add support for custom placeholderData that works seamlessly with the reducer function, allowing initialization of complex data types beyond simple arrays.

https://github.com/TanStack/query/discussions/9065

BREAKING CHANGE: The maxChunks parameter has been removed from streamedQuery. Use a custom reducer function to control data aggregation behavior instead.

## Summary by CodeRabbit

- New Features
  - Streamed queries now use a reducer-based API to accumulate streamed chunks into final results.
  - Optional initialValue lets you seed accumulated data before the first chunk; required when providing a custom reducer.
  - Default reducer preserves prior append behavior for array results when no custom reducer is supplied.

- Documentation
  - Docs updated to describe reducer and initialValue usage and removed references to the deprecated maxChunks option.


packages/query-core/src/__tests__/streamedQuery.test.tsx
```diff
@@ -350,14 +350,18 @@ describe('streamedQuery', () => {
...
     const observer = new QueryObserver(queryClient, {
       queryKey: key,
-      queryFn: streamedQuery({
-        queryFn: () => createAsyncNumberGenerator(3),
-        maxChunks: 2,
-      }),
+      queryFn: streamedQuery<number, Record<number, boolean>>({
+        queryFn: () => createAsyncNumberGenerator(2),
+        reducer: (acc, chunk) => ({
+            ...acc,
```
marcog83: the first time `reducer` is invoked, `acc` is an empty array, because `placeholderData` is not specified.
it seems a bit strange to me. but i'm not sure how to solve it. 
i can apply the same logic as in native `reduce`... 
 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#initialvalue.

packages/query-core/src/streamedQuery.ts
```diff
@@ -1,6 +1,15 @@
 import { addToEnd } from './utils'
 import type { QueryFunction, QueryFunctionContext, QueryKey } from './types'
 
+type StreamedQueryParams<TQueryFnData, TData, TQueryKey extends QueryKey> = {
```
marcog83: I created this type just for clarity. should i keep it inline/move to a dedicated file or in types?
About type safety i tried to split the type in 2 like below, but i feel it is over-engineering.

```ts
type BaseStreamedQueryParams<TQueryFnData, TData> = {
   queryFn: (
    context: QueryFunctionContext<TQueryKey>,
  ) => AsyncIterable<TQueryFnData> | Promise<AsyncIterable<TQueryFnData>>
  refetchMode?: 'append' | 'reset' | 'replace'
}
type SimpleStreamedQueryParams<TQueryFnData, TData, TQueryKey extends QueryKey> = BaseStreamedQueryParams<TQueryFnData, TData> & {

  reducer?: never;
  placeholderData?: never;
}

type ReducibleStreamedQueryParams<TQueryFnData, TData, TQueryKey extends QueryKey> = BaseStreamedQueryParams<TQueryFnData, TData> & {
  
  reducer?: (acc: TData, chunk: TQueryFnData) => TData
  placeholderData?: TData
}
type StreamedQueryParams<TQueryFnData, TData, TQueryKey extends QueryKey> =
  | SimpleStreamedQueryParams<TQueryFnData, TData, TQueryKey>
  | ReducibleStreamedQueryParams<TQueryFnData, TData, TQueryKey>

export function streamedQuery<
  TQueryFnData = unknown,
  TData = Array<TQueryFnData>,
>(params: StreamedQueryParams<TQueryFnData, TData, TQueryKey>): QueryFunction<TData, TQueryKey> {

  const reducer =
    'reducer' in params && typeof params.reducer === 'function'
      ? params.reducer
      : (items: TData, chunk: TQueryFnData) =>
          addToEnd((items ?? []) as Array<TQueryFnData>, chunk) as TData;

  const placeholderData =
    'placeholderData' in params ? params.placeholderData : ([] as TData);
// ...

```

packages/query-core/src/streamedQuery.ts
```diff
@@ -11,31 +20,26 @@ import type { QueryFunction, QueryFunctionContext, QueryKey } from './types'
...
-  maxChunks,
-}: {
-  queryFn: (
-    context: QueryFunctionContext<TQueryKey>,
-  ) => AsyncIterable<TQueryFnData> | Promise<AsyncIterable<TQueryFnData>>
-  refetchMode?: 'append' | 'reset' | 'replace'
-  maxChunks?: number
-}): QueryFunction<Array<TQueryFnData>, TQueryKey> {
+  reducer = (items, chunk) => addToEnd((items ?? []) as Array<TQueryFnData>, chunk) as TData,
+  placeholderData = [] as TData,
```
marcog83: i introduced this `placeholderData` prop to mimic native `reduce` with an [initialValue](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#initialvalue).
My first approach was to use the existing `placeholderData` parameter from the Observer, but even though I can access it through the `query.options` object, TypeScript throws an error.

marcog83: Below you can find an implementation about how to mimic `Array.prototype.reduce` behavior where the initial value is optional, and if not provided, the first element becomes the accumulator. I didn't applied it to the PR because i think it better to have your point of of view about it. 

```ts
// ...existing code...
return async (context) => {
  const query = context.client
    .getQueryCache()
    .find({ queryKey: context.queryKey, exact: true })
  const isRefetch = !!query && query.state.data !== undefined
  if (isRefetch && params.refetchMode === 'reset') {
    query.setState({
      status: 'pending',
      data: undefined,
      error: null,
      fetchStatus: 'fetching',
    })
  }

  const hasInitialValue = 'placeholderData' in params;
  let result: TData;
  let isFirstChunk = true;

  const stream = await params.queryFn(context)

  for await (const chunk of stream) {
    if (context.signal.aborted) {
      break
    }
    
    if (isFirstChunk) {
      if (hasInitialValue) {
        // If we have placeholderData, use it as initial accumulator
        result = reducer(placeholderData, chunk);
      } else {
        // If no placeholderData, first chunk becomes the accumulator
        result = chunk as unknown as TData;
      }
      isFirstChunk = false;
    } else {
      result = reducer(result, chunk);
    }
    
    // don't append to the cache directly when replace-refetching
    if (!isRefetch || params.refetchMode !== 'replace') {
      context.client.setQueryData<TData>(
        context.queryKey,
        (prev) => {
          if (prev === undefined) {
            return result;
          }
          return hasInitialValue ? reducer(prev, chunk) : result;
        }
      )
    }
  }

  // Handle empty stream case
  if (isFirstChunk) {
    if (hasInitialValue) {
      result = placeholderData;
    } else {
      throw new Error('Reduce of empty stream with no initial value');
    }
  }

  if (isRefetch && params.refetchMode === 'replace' && !context.signal.aborted) {
    context.client.setQueryData<TData>(context.queryKey, result)
  }

  return context.client.getQueryData(context.queryKey)!
}
```
The key changes:

- **Initial value logic**: Only use `placeholderData` if it's explicitly provided
- **First chunk handling**: If no `placeholderData` the first chunk becomes the initial accumulator
- **Empty stream error**: Throw an error if the stream is empty and no `placeholderData` is provided

TkDodo: I think what we’d want here is a mandatory `initialValue` to the `reducer` if you pass a custom reducer. The term `placeholderData` is a bit overloaded because here, it does something different than on `useQuery` itself so I wouldn’t name it like that.

It could be two separate params, like `reducer` and `initialValue`, but we’d have to make the types so that `initialValue` is required when you pass a reducer. That’s possible but usually needs overloads or conditional types. Or, we could also just use a tuple maybe:

```
reducer: [initialValue: TData, (accumulator: TData, chunk: TQueryFnData) => TData]
```

marcog83: Great.thanks for the feedback. I'm on vacation right now,but next week I'll back. I'm going to provide a prop initial value, and I'll play with TS in order to make it mandatory when reducer is specified.

TkDodo: @marcog83 did you get around to it? I would like to ship it together with:

- https://github.com/TanStack/query/pull/9606

since they are both breaking changes

marcog83: Hi @TkDodo , sorry for the late... i resolved the conflicts. not sure about the suggestions by AI, to be honest.
Let me know if i need to apply some changes based on the AI suggestions.

packages/query-core/src/__tests__/streamedQuery.test.tsx
```diff
@@ -391,13 +391,18 @@ describe('streamedQuery', () => {
...
-  test('should support maxChunks', async () => {
+  test('should support custom reducer', async () => {
     const key = queryKey()
+
     const observer = new QueryObserver(queryClient, {
       queryKey: key,
-      queryFn: streamedQuery({
-        queryFn: () => createAsyncNumberGenerator(3),
-        maxChunks: 2,
+      queryFn: streamedQuery<number, Record<number, boolean>>({
```
TkDodo: let’s not pass types this way - as it won’t test if type inference works. Instead, let’s type the initialValue:

```
initialValue: {} as Record<number, boolean>
```

packages/query-core/src/__tests__/streamedQuery.test.tsx
```diff
@@ -409,41 +414,34 @@ describe('streamedQuery', () => {
...
-  test('maxChunks with append refetch', async () => {
+  test('should support custom reducer with initialValue', async () => {
     const key = queryKey()
     const observer = new QueryObserver(queryClient, {
       queryKey: key,
-      queryFn: streamedQuery({
-        queryFn: () => createAsyncNumberGenerator(3),
-        maxChunks: 2,
-        refetchMode: 'append',
+      queryFn: streamedQuery<number, Record<number, boolean>>({
```
TkDodo: same here please:

```suggestion
      queryFn: streamedQuery({
```

packages/query-core/src/streamedQuery.ts
```diff
@@ -55,19 +82,16 @@ export function streamedQuery<
 
       // don't append to the cache directly when replace-refetching
       if (!isRefetch || refetchMode !== 'replace') {
-        context.client.setQueryData<Array<TQueryFnData>>(
-          context.queryKey,
-          (prev = []) => {
-            return addToEnd(prev, chunk, maxChunks)
-          },
+        context.client.setQueryData<TData>(context.queryKey, (prev) =>
+          reducer(prev ?? initialValue, chunk),
```
TkDodo: we need to explicitly check for `undefined` here because it’s the only indicator for “empty” we have. `null` is technically allowed:

```suggestion
          reducer(prev === undefined ? initialValue : prev, chunk),
```

marcog83:
Thanks a lot for letting me collaborate on react-query! It’s been an awesome experience.

PR #9533, opened 2025-08-05
test(react-query/queryOptions): add 'useQuery' basic tests with 'queryOptions'

sukvvon:


manudeli:
I don't think we need to expand tests for queryOptions like this.

`queryOptions` is essentially a type-level helper—its main purpose is to guide type inference and help with consistency across various hooks (useQuery, useQueries, useSuspenseQuery, queryClient methods, etc).

At runtime, it just returns the same object, so unit testing it with various hooks and queryClient's methods doesn't add meaningful value.

Instead, what matters most is ensuring that TypeScript correctly infers and validates types in different usage contexts. I'd prefer to keep the focus there (useQuery, useQueries, useSuspenseQuery, queryClient methods, etc)

If you have a different perspective, feel free to let me know

PR #9541, opened 2025-08-07
docs: clarify default behavior of invalidateQueries refetching

### LINKED ISSUE #9531: Documentation mismatch: invalidateQueries default behavior claims "all" but only refetches "active" queries
### Describe the bug

The documentation for `queryClient.invalidateQueries()` states that the `type` parameter defaults to `'all'`, which should invalidate and refetch all matching queries (both active and inactive). However, the actual behavior only refetches active queries by default, leaving inactive cached queries stale.

https://codesandbox.io/p/sandbox/wpgfzw

### Steps to reproduce

1. Create queries with the same prefix (e.g., `['todos']`, `['todos', 1]`, `['todos', 2]`)
2. Prefetch an additional query to create an inactive cached query (e.g., `['todos', 3]`)
3. Call `queryClient.invalidateQueries({ queryKey: ['todos'] })`
4. Observe that only active queries refetch, while inactive cached queries remain stale

### Expected behavior
According to the documentation, `queryClient.invalidateQueries({ queryKey: ['todos'] })` should:
- Invalidate ALL queries with keys starting with `['todos']`
- Refetch ALL matching queries (both active and inactive)

### Actual behavior
`queryClient.invalidateQueries({ queryKey: ['todos'] })` only:
- Refetches queries that have active observers (components using them)
- Leaves inactive cached queries (like prefetched queries) stale

None

### Screenshots or Videos

_No response_

### Platform

- **OS**: macOS/Windows/Linux
- **Browser**: Chrome/Firefox/Safari (any)
- **Version of TanStack Query**: 5.84.1
- **React version**: 18.0.0+

### Tanstack Query adapter

None

### TanStack Query version

5.84.1

### TypeScript version

_No response_

### Additional context

**Root cause analysis:**
The issue appears to be in the invalidation logic where the default refetch behavior uses `'active'` instead of `'all'` despite documentation claiming otherwise.

**Workaround:**
Explicitly specify `type: 'all'` to get the documented behavior:
```javascript
queryClient.invalidateQueries({ queryKey: ['todos'], type: 'all' })
```

**Impact:**
This discrepancy can cause subtle bugs where developers expect all cached data to be refreshed after mutations, but inactive cached queries remain stale, potentially leading to inconsistent application state.

**Documentation references:**
- The documentation states that `type` defaults to `'all'`
- However, the actual implementation appears to default to `'active'` behavior

## Steps to Reproduce (Detailed)
```markdown
### Detailed Reproduction Steps

1. **Setup queries with staleTime: Infinity**
   ```javascript
     defaultOptions: {
       queries: { staleTime: Infinity }
     }
   });
   ```

2. **Create multiple queries with same prefix**
   ```javascript
   // Active queries
   useQuery({ queryKey: ['todos'], queryFn: fetchTodos });
   useQuery({ queryKey: ['todos', 1], queryFn: () => fetchTodosPage(1) });
   
   // Inactive cached query (prefetched)
   queryClient.prefetchQuery({
     queryKey: ['todos', 3],
     queryFn: () => fetchTodosPage(3)
   });
   ```

3. **Test invalidation behavior**
   ```javascript
   // This should refetch ALL queries according to docs
   queryClient.invalidateQueries({ queryKey: ['todos'] });
   ```

4. **Observe the results**
   - ✅ Active queries: Get new fetch timestamps
   - ❌ Inactive cached queries: Keep old fetch timestamps
   
5. **Test workaround**
   ```javascript
   // This works as expected
   queryClient.invalidateQueries({ queryKey: ['todos'], type: 'all' });
   ```

6. **Verify fix**
   - ✅ Active queries: Get new fetch timestamps  
   - ✅ Inactive cached queries: Also get new fetch timestamps
```


## Suggested Fix
### Suggested Solution

**Option 1: Fix Documentation**
Update documentation to clarify that the default behavior only refetches active queries, and `type: 'all'` is required for all queries.

**Option 2: Fix Implementation**
Change the default behavior to match documentation by making `type: 'all'` the actual default.

**Option 3: Breaking Change Warning**
If changing the default behavior, consider it a breaking change and document the migration path.

**Preference:** Option 1 (documentation fix) would be safest for existing applications, but Option 2 (implementation fix) would match user expectations better.

Yashkanekar:
This PR updates the documentation for the invalidateQueries method in React Query to clarify the default behavior of query invalidation and refetching.

-It explains that by default, only active queries are immediately refetched.
-Inactive queries are marked stale but not automatically refetched until they become active again.
-It adds instructions on how to force invalidation and refetching of all queries (both active and inactive) by using the type: 'all' or refetchType: 'all' option.

closes #9531

### DIFF
--- a/docs/framework/react/guides/query-invalidation.md
+++ b/docs/framework/react/guides/query-invalidation.md
@@ -23,8 +23,7 @@ When a query is invalidated with `invalidateQueries`, two things happen:
 - It is marked as stale. This stale state overrides any `staleTime` configurations being used in `useQuery` or related hooks
 - If the query is currently being rendered via `useQuery` or related hooks, it will also be refetched in the background
 
-> **Important:** By default, `invalidateQueries` will immediately refetch *only active queries*—those currently used by mounted components. Inactive queries (cached but not in use) are marked as stale but will not be refetched until they become active again (e.g., when a component utilizing the query mounts). This means that calling `invalidateQueries` without additional options does *not* guarantee all matching queries are refetched immediately.
-
+> **Important:** By default, `invalidateQueries` will immediately refetch _only active queries_—those currently used by mounted components. Inactive queries (cached but not in use) are marked as stale but will not be refetched until they become active again (e.g., when a component utilizing the query mounts). This means that calling `invalidateQueries` without additional options does _not_ guarantee all matching queries are refetched immediately.
 
 ## Query Matching with `invalidateQueries`
 
@@ -44,12 +43,12 @@ queryClient.invalidateQueries({ queryKey: ['todos'] })
 
 // Both queries below will be invalidated
 const todoListQuery = useQuery({
-queryKey: ['todos'],
-queryFn: fetchTodoList,
+  queryKey: ['todos'],
+  queryFn: fetchTodoList,
 })
 const todoListQuery = useQuery({
-queryKey: ['todos', { page: 1 }],
-queryFn: fetchTodoList,
+  queryKey: ['todos', { page: 1 }],
+  queryFn: fetchTodoList,
 })
 ```
 
@@ -122,8 +121,8 @@ const todoListQuery = useQuery({
 
 // The query below will be invalidated
 const todoListQuery = useQuery({
-queryKey: ['todos', { version: 10 }],
-queryFn: fetchTodoList,
+  queryKey: ['todos', { version: 10 }],
+  queryFn: fetchTodoList,
 })
 
 // However, the following query below will NOT be invalidated
@@ -142,12 +141,12 @@ By default, `invalidateQueries` only refetches active queries. If you want to re
 ```tsx
 queryClient.invalidateQueries({
   queryKey: ['todos'],
-  refetchType: 'all' // or type: 'all'
-});
+  refetchType: 'all', // or type: 'all'
+})
 ```
 
 [//]: # 'Example6'
 
 This forces every matching query—active or inactive—to be refetched immediately.
 
-> 🧠 **Note:** The default behavior (`refetchType: 'active'`) helps avoid unnecessary network requests. But if you're triggering an invalidation after a mutation that affects *all clients or views*, make sure to use `refetchType: 'all'` to keep all query caches up to date.
+> 🧠 **Note:** The default behavior (`refetchType: 'active'`) helps avoid unnecessary network requests. But if you're triggering an invalidation after a mutation that affects _all clients or views_, make sure to use `refetchType: 'all'` to keep all query caches up to date.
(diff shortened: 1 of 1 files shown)


docs/framework/react/guides/query-invalidation.md
```diff
@@ -23,6 +23,9 @@ When a query is invalidated with `invalidateQueries`, two things happen:
 - It is marked as stale. This stale state overrides any `staleTime` configurations being used in `useQuery` or related hooks
 - If the query is currently being rendered via `useQuery` or related hooks, it will also be refetched in the background
 
+> **Important:** By default, `invalidateQueries` will immediately refetch *only active queries*—those currently used by mounted components. Inactive queries (cached but not in use) are marked as stale but will not be refetched until they become active again (e.g., when a component utilizing the query mounts). This means that calling `invalidateQueries` without additional options does *not* guarantee all matching queries are refetched immediately.
+
```
TkDodo: this is pretty much the same as the two points above are also saying. The second one only says “is currently being rendered via `useQuery` or related hooks”, maybe this could be re-worded to “active queries”, and then explained what an active query is.

docs/framework/react/guides/query-invalidation.md
```diff
@@ -131,3 +134,20 @@ const todoListQuery = useQuery({
 ```
 
 [//]: # 'Example5'
+
+By default, `invalidateQueries` only refetches active queries. If you want to refetch **all matching queries**, including those currently inactive in the cache, you need to explicitly set the `refetchType` (or `type`) option to `'all'`:
```
TkDodo, line 137: this is partially wrong. setting `type: 'all'` wouldn't achieve that. It's also the default behaviour.

docs/framework/react/guides/query-invalidation.md
```diff
@@ -131,3 +134,20 @@ const todoListQuery = useQuery({
...
+  queryKey: ['todos'],
+  refetchType: 'all' // or type: 'all'
+});
+```
+
+[//]: # 'Example6'
+
+This forces every matching query—active or inactive—to be refetched immediately.
+
+> 🧠 **Note:** The default behavior (`refetchType: 'active'`) helps avoid unnecessary network requests. But if you're triggering an invalidation after a mutation that affects *all clients or views*, make sure to use `refetchType: 'all'` to keep all query caches up to date.
```
TkDodo: I don’t understand that sentence - what is “all clients or views” referring to? Imo there’s really no need to set `refetchType: 'all'`, as it will likely lead to overfetching. It's something we have documented in the [API docs](https://tanstack.com/query/v5/docs/reference/QueryClient#queryclientinvalidatequeries), but I don’t think we need to emphasize on this with a guide.

PR #9552, opened 2025-08-11
fix(types): improve cache-access function return types for better type safety

braden-w:
Cache-access functions like `fetchQuery` and `ensureQueryData` had incorrect return types that didn't match their runtime behavior. They returned `Promise<TData>` but actually return raw cache data (`TQueryData`), not transformed data.

This created confusion between three distinct data types:
- `TQueryFnData`: Raw data from queryFn  
- `TQueryData`: Data stored in cache
- `TData`: Data after select transformation (observers only)

The type system inconsistently used `TData` to represent both cache data and transformed data, leading to runtime mismatches and poor IDE support.

## Changes

**Core function fixes:**
```ts
// Before
fetchQuery(...): Promise<TData>              // Wrong - returns cache data  
ensureQueryData(...): Promise<TData>         // Wrong - returns cache data

// After  
fetchQuery(...): Promise<TQueryData>         // Correct - returns cache data
ensureQueryData(...): Promise<TQueryData>    // Correct - returns cache data
```

**Infinite query simplification:**
```ts
// Before: Complex generic ordering, unclear return types
fetchInfiniteQuery<TQueryFnData, TError, TQueryKey, TPageParam, TQueryData>

// After: Clear semantics, direct return type
fetchInfiniteQuery(...): Promise<InfiniteData<TQueryFnData, TPageParam>>
```

**Base interface consistency:**
- `QueryOptions<..., TData>` → `QueryOptions<..., TQueryData>` 
- `Query<..., TData>` → `Query<..., TQueryData>`
- `QueryBehavior<..., TData>` → `QueryBehavior<..., TQueryData>`
- `Retryer<TData>` → `Retryer<TQueryData>`

Updated 80+ type references across the entire cache system to establish semantic consistency.

**Type assertion cleanup:**
- Fixed `getQueriesData` incorrect cast from cache data to query function data
- Improved `fetchQuery` to use non-null assertion instead of type cast
- Verified remaining assertions are necessary

This establishes clear boundaries between queryFn data, cache data, and transformed data throughout the type system.

### DIFF
--- a/packages/query-core/src/__tests__/queryClient.test.tsx
+++ b/packages/query-core/src/__tests__/queryClient.test.tsx
@@ -11,7 +11,7 @@ import {
   skipToken,
 } from '..'
 import { mockOnlineManagerIsOnline } from './utils'
-import type { QueryCache, QueryFunction, QueryObserverOptions } from '..'
+import type { QueryCache, QueryFunction, QueryObserverOptions, InfiniteData } from '..'
 
 describe('queryClient', () => {
   let queryClient: QueryClient
@@ -806,7 +806,7 @@ describe('queryClient', () => {
         queryClient.fetchInfiniteQuery<
           StrictData,
           any,
-          StrictData,
+          InfiniteData<StrictData, number>,
           StrictQueryKey,
           number
         >({ queryKey: key, queryFn: fetchFn, initialPageParam: 0 }),
@@ -844,7 +844,7 @@ describe('queryClient', () => {
       await queryClient.prefetchInfiniteQuery<
         StrictData,
         any,
-        StrictData,
+        InfiniteData<StrictData, number>,
         StrictQueryKey,
         number
       >({ queryKey: key, queryFn: fetchFn, initialPageParam: 0 })
--- a/packages/query-core/src/query.ts
+++ b/packages/query-core/src/query.ts
@@ -34,19 +34,19 @@ import type { Retryer } from './retryer'
 interface QueryConfig<
   TQueryFnData,
   TError,
-  TData,
+  TQueryData,
   TQueryKey extends QueryKey = QueryKey,
 > {
   client: QueryClient
   queryKey: TQueryKey
   queryHash: string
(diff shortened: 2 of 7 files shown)


LeDevNovice:
If this implementation is merged, I think it is important to keep in mind that there is some modification to make in the reference part of the documentation to be correct with the new typing. We need to update the return types for fetchQuery, ensureQueryData, fetchInfiniteQuery, ensureInfiniteQueryData in reference part of the documentation and anywhere else it si necessary.

TkDodo:
> Cache-access functions like fetchQuery and ensureQueryData had incorrect return types that didn't match their runtime behavior. They returned Promise<TData> but actually return raw cache data (TQueryData), not transformed data.

I couldn’t reproduce this “issue” with the current types. For one, `select` is not invoked on imperative methods. So yes, at runtime, you will always get what’s stored in the query cache, and that’s on purpose. On type level, it should behave the same, but from what I can tell, it does.

You cannot pass `select` directly to `fetchQuery`, because it type-errors. What can happen is that a `queryOptions` object that contains `select` is passed, but even then, the return type of `fetchQuery` is correct, as you can see in this [simple TS playground](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgRwK4FMoE8DyYbAQB2AzgDRwCKG2AwgDbDpHwC+cAZlBCHAOQABGAENSIgMYBrAPRR0w8TAC0aTFj4AoDeOIl4EfIVJwAvChq5DugBQINcc2oDS6LAC44AbT6TXfALpk9o7YAGJEHtYAlKYAfHAACtwgwCToAHRyJBD0AG7o1gCsUUEOafToipEAJsIiMSbxtSLpMBAAyjBQwEQA5tFBrFFaOmJwzcKmcMIA7sLA8EToM1QWDEws0ekc6DDiABbUatYGBLrD0tIODgB6APxAA).

In any case, I’m not willing to merge a PR that changes names of types across so many files. It would be better if you’d file an issue or discussion first so that we could take isolated steps at fixing the problem if it turns out there is one.

PR #9560, opened 2025-08-12
docs: clarify queryClient refetch behavior for stale queries

imadselka:
This pull request updates the documentation for the `useRefreshOnFocus` custom hook in the React Native guide to clarify its behavior and improve its implementation. The hook now refetches all active stale queries on screen focus, rather than requiring a specific `refetch` function to be passed in.

**Documentation and code improvements:**

* Updated the description and implementation of `useRefreshOnFocus` to indicate that it refetches all active stale queries when the screen is focused, using `queryClient.refetchQueries` instead of a provided `refetch` function. [[1]](diffhunk://#diff-cb1278fded1b8132f2d7c929ec937d794c995db4c385ac1b0425d9b2a873e3b5L77-R85) [[2]](diffhunk://#diff-cb1278fded1b8132f2d7c929ec937d794c995db4c385ac1b0425d9b2a873e3b5L93-R102)
* Clarified the explanation about skipping the first focus to reflect the new implementation and behavior.


TkDodo:
i was more thinking to have this directly in the react native example instead of calling `refetch()` like we recommend now?

imadselka:
yes you are right , i changed it .
but another thing do you think we should add it in the main [refetchQueries ](https://tanstack.com/query/latest/docs/reference/QueryClient#queryclientrefetchqueries) docs like this 

```tsx
// refetch all stale queries: 
await queryClient.refetchQueries({ stale: true })

// refetch only stale queries that are currently active (recommended for RN screen-focus)
await queryClient.refetchQueries({ stale: true, type: 'active' })
```

heisenbugged:
my recommendation here is to still pass in a query key so it's refresh on focus tied to a specific query rather than a blanket refresh everything

docs/framework/react/react-native.md
```diff
@@ -90,13 +92,14 @@ export function useRefreshOnFocus<T>(refetch: () => Promise<T>) {
...
-    }, [refetch]),
+      // refetch all stale active queries
+      queryClient.refetchQueries({ queryKey:['posts'], stale: true, type: 'active' })
+    }, [queryClient]),
   )
 }
 ```
 
-In the above code, `refetch` is skipped the first time because `useFocusEffect` calls our callback on mount in addition to screen focus.
+In the above code, The first focus (when the screen is initially mounted) is skipped because `useFocusEffect` calls our callback on mount in addition to screen focus.
```
TkDodo:
```suggestion
In the above code, the first focus (when the screen is initially mounted) is skipped because `useFocusEffect` calls our callback on mount in addition to screen focus.
```

PR #9563, opened 2025-08-13
fix(query-core): focus check issue from retryer for refetchIntervallnBackground

### LINKED ISSUE #8353: Polling Stops with refetchIntervalInBackground and Retry When Tab Is Inactive
### Describe the bug

I am creating this issue based on the discussion [here](https://github.com/TanStack/query/discussions/7385#discussioncomment-11372314).

When using `refetchIntervalInBackground` with `retry`, the retry mechanism fails to execute if the browser tab is inactive. I have the following configuration for my React Query setup:

```javascript
  defaultOptions: {
    queries: {
      staleTime: 30_000, // 30 sec
      networkMode: "always",
    },
    mutations: {
      networkMode: "always",
    },
  },
});
```

```javascript
export function useGetTodoList() {
  return useQuery(
    ["todos"],
    async () => {
      // fetch logic here
    },
    {
      refetchInterval: 60_000, // 1 min
      refetchIntervalInBackground: true,
      retry: 3,
    }
  );
}
```

### **Problem**
With this configuration:
- The query is set to poll every 1 minute (`refetchInterval: 60_000`), even when the tab is inactive (`refetchIntervalInBackground: true`).
- The retry mechanism is configured to retry 3 times on failure (`retry: 3`).

However, if the tab is inactive and a request fails, the retry logic pauses and waits until the tab becomes active again. This behavior eliminates the purpose of `refetchIntervalInBackground`, as the retries do not proceed while the tab remains inactive.

https://codesandbox.io/p/sandbox/suspicious-butterfly-gk6c6t

### Steps to reproduce

1. Open codesandbox link and use the preview option in new tab(https://gk6c6t.csb.app/)
2. open chrome network tool to set the network as offline
3. switch to next tab and wait for 3 minutes, so we can expect atleast (3 requests for refetch + retry calls (3 * 3 = 9). so total of 12 requests atleast)
4. switch back to the preview tab
5. there will be only 2 failed requests instead of ~12

### Expected behavior

When `refetchIntervalInBackground: true` is used:
- Retries should execute in the background, independent of tab visibility.
- The retry mechanism should adhere to the configured retry settings (e.g., retry count, delay) without relying on the tab being active.

Every time

### Screenshots or Videos

_No response_

### Platform

Browser: Chrome

### Tanstack Query adapter

react-query

### TanStack Query version

4.36.1

### TypeScript version

_No response_

### Additional context

_No response_

Hellol77:
fixed #8353 

- Instead of removing focusManager.isFocused(), update retryer.ts to check (config.refetchIntervalInBackground === true || focusManager.isFocused())
- Passed the refetchIntervalInBackground option through query.ts.
- Added the refetchIntervalInBackground property to the QueryOptions interface in types.ts.

## Summary by CodeRabbit

* **New Features**
  * Added a Query option refetchIntervalInBackground (default: false) to allow scheduled refetches/retries to continue while the app/tab is in the background.

* **Tests**
  * Added tests verifying background refetch/retry behavior for enabled, disabled, and default cases to ensure pause/continue semantics based on focus and network state.

### DIFF
--- a/packages/query-core/src/types.ts
+++ b/packages/query-core/src/types.ts
@@ -275,6 +275,11 @@ export interface QueryOptions<
    * Maximum number of pages to store in the data of an infinite query.
    */
   maxPages?: number
+  /**
+   * If set to `true`, the query will continue to refetch while their tab/window is in the background.
+   * Defaults to `false`.
+   */
+  refetchIntervalInBackground?: boolean
 }
 
 export interface InitialPageParam<TPageParam = unknown> {
--- a/packages/query-core/src/__tests__/query.test.tsx
+++ b/packages/query-core/src/__tests__/query.test.tsx
@@ -1304,4 +1304,134 @@ describe('query', () => {
       data: 'data1',
     })
   })
+
+  test('should continue retry in background when refetchIntervalInBackground is true', async () => {
+    const key = queryKey()
+
+    // make page unfocused
+    const visibilityMock = mockVisibilityState('hidden')
+
+    let count = 0
+    let result
+
+    const promise = queryClient.fetchQuery({
+      queryKey: key,
+      queryFn: () => {
+        count++
+
+        if (count === 3) {
+          return `data${count}`
+        }
+
+        throw new Error(`error${count}`)
+      },
+      retry: 3,
+      retryDelay: 1,
+      refetchIntervalInBackground: true,
+    })
+
+    promise.then((data) => {
+      result = data
+    })
+
+    // Check if we do not have a result yet
+    expect(result).toBeUndefined()
+
+    // Query should continue retrying in background
+    await vi.advanceTimersByTimeAsync(50)
+    expect(result).toBe('data3')
+
+    // Reset visibilityState to original value
+    visibilityMock.mockRestore()
+  })
+
+  test('should pause retry when unfocused if refetchIntervalInBackground is false', async () => {
+    const key = queryKey()
+
+    // make page unfocused
+    const visibilityMock = mockVisibilityState('hidden')
+
+    let count = 0
+    let result
+
+    const promise = queryClient.fetchQuery({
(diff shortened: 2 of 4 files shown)


TkDodo (approved):

TkDodo:
plase have a look at the conflicts and failing tests

Hellol77:
I fixed the conflicts and test issues! thanks for review. @TkDodo

TkDodo:
I discussed this with @DamianOsipiuk and we think the better fix would be to check the `refetchIntervalInBackground` prop additionally in the retryer instead of removing the check. That should also fix the issue.

We might remove the check in the next major version as it’s technically a breaking change.

Hellol77:
Got it, I updated this PR to add a check for the refetchIntervalInBackground prop in the retryer instead of removing the focus check. @TkDodo

packages/query-core/src/types.ts
```diff
@@ -275,6 +275,11 @@ export interface QueryOptions<
    * Maximum number of pages to store in the data of an infinite query.
    */
   maxPages?: number
+  /**
+   * If set to `true`, the query will continue to refetch while their tab/window is in the background.
+   * Defaults to `false`.
+   */
+  refetchIntervalInBackground?: boolean
```
TkDodo, line 282: hmm we shouldn’t really add this to `QueryOptions` as it’s an observer level property.

Now that I’m seeing this, I don’t think it’s “fixable” with an additional check because you can have different observers with different values for `refetchIntervalInBackground`.

I think the best step forward is to just remove the check in v6 and document the current behaviour for v5.

Hellol77: I understand your point about the issue. 

I'll close this pr and submit docs explaining the current v5 behavior with some alternative approaches. 

Thank you for the feedback!

PR #9564, opened 2025-08-14
Allow including devtools panel in production builds

oscartbeaumont:
This PR removes the `isDev` condition from the `ReactQueryDevtoolsPanel` components.

It does not remove the same condition from `ReactQueryDevtools`.

I think it should be possible for developers to opt-in if they want these devtools panels in production.

This one might need some discussion about whether we would want to remove these checks from `ReactQueryDevtoolsPanel` too. The checks are nice to protect developers from accidentally bundling devtools in production but if that is desired there is currently no way for the developer to override it.

### DIFF
--- a/packages/react-query-devtools/src/index.ts
+++ b/packages/react-query-devtools/src/index.ts
@@ -10,11 +10,6 @@ export const ReactQueryDevtools: (typeof Devtools)['ReactQueryDevtools'] =
       }
     : Devtools.ReactQueryDevtools
 
-export const ReactQueryDevtoolsPanel: (typeof DevtoolsPanel)['ReactQueryDevtoolsPanel'] =
-  process.env.NODE_ENV !== 'development'
-    ? function () {
-        return null
-      }
-    : DevtoolsPanel.ReactQueryDevtoolsPanel
+export const ReactQueryDevtoolsPanel: (typeof DevtoolsPanel)['ReactQueryDevtoolsPanel'] = DevtoolsPanel.ReactQueryDevtoolsPanel
 
 export type DevtoolsPanelOptions = DevtoolsPanel.DevtoolsPanelOptions
--- a/packages/solid-query-devtools/src/index.tsx
+++ b/packages/solid-query-devtools/src/index.tsx
@@ -9,10 +9,6 @@ export const SolidQueryDevtools: typeof SolidQueryDevtoolsComp = isDev
       return null
     }
 
-export const SolidQueryDevtoolsPanel: typeof SolidQueryDevtoolsCompPanel = isDev
-  ? clientOnly(() => import('./devtoolsPanel'))
-  : function () {
-      return null
-    }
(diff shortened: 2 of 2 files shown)


TkDodo:
> but if that is desired there is currently no way for the developer to override it.

that’s not true, we have documentation on how to lazy load devtools in production:

https://tanstack.com/query/v5/docs/framework/react/devtools#devtools-in-production

oscartbeaumont:
Ahh, I came from the SolidJS side which doesn't have a similar feature and assumed it was unimplemented. Sorry!

Will update this PR to implement the `/production` export similar to what React does instead.

oscartbeaumont:
I will fix up the CodeRabbit comments and test failure in the morning!

packages/solid-query-devtools/src/production.tsx
```diff
@@ -0,0 +1,12 @@
+import clientOnly from './clientOnly'
+import type SolidQueryDevtoolsComp from './devtools'
+import type SolidQueryDevtoolsCompPanel from './devtoolsPanel'
```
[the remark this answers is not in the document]
oscartbeaumont: This is currently the case for the `index.ts` route so going to ignore.

oscartbeaumont:
This should be updated and ready to go but could be worth getting a review from anyone with better experience on `package.json` `exports` as i'm far from an expert on them.

packages/solid-query-devtools/.attw.json
```diff
@@ -0,0 +1,3 @@
+{
```
oscartbeaumont, line 1: This was copied from `react-query-devtools`

packages/solid-query-devtools/package.json
```diff
@@ -32,28 +32,60 @@
     "build:dev": "tsup --watch"
   },
   "type": "module",
-  "main": "./build/index.cjs",
-  "module": "./build/index.js",
-  "types": "./build/index.d.ts",
+  "main": "./build/index/index.cjs",
+  "module": "./build/index/index.js",
+  "types": "./build/index/index.d.ts",
```
TkDodo, line 39: is and `index` directory with an `index.ext` file really correct?

oscartbeaumont: It does seem to be correct. I'm not sure if we should adjust the `tsup.config.ts` so it generates a more flat structure but with the Solid preset i'm not totally sure how we can do that.

Screenshot 2025-09-06 at 1 35 35 pm

packages/solid-query-devtools/src/production.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+export const SolidQueryDevtools: typeof SolidQueryDevtoolsComp = clientOnly(
```
afurm, line 5: `clientOnly` defers the import until the component mounts on the client, which is the right pattern for SSR-safe production devtools. However, if `('./devtools')` itself has a transitive import that runs on the server (e.g., accessing `window`, `document`, or a Node-only module at module evaluation time), the deferred import will still trigger that code when the dynamic `import()` runs. Has this been verified with a server-component tree that imports `SolidQueryDevtools` at the module level?

Issue #9399, opened 2025-07-10
Hydration error using loading state of useQuery when prefetching


Icestonks:
### Describe the bug

If you prefetch data on the server and then use ``useQuery`` with ``isLoading`` on the client, it can lead to a hydration error.

### Your minimal, reproducible example

https://stackblitz.com/~/github.com/Icestonks/react-query-trpc-test

### Steps to reproduce

1. Prefetch something on the server
2. Hydrate it to the client
3. Use ``useQuery`` in the client, and have a loading fallback with ``isLoading``.

Then there should come a hydration error.

### Expected behavior

When first building and starting the project, we should see an hydration error. The error can also come at other times, but that's the most reliable way, to get the error to occur.

### How often does this bug happen?

Sometimes

### Screenshots or Videos

<img width="567" height="732" alt="Image" src="https://github.com/user-attachments/assets/3685d367-c7ce-4758-a875-c4ae88401dcf" />

### Platform

- OS: Windows
- Browser: Google Chrome
- Browser version: 138.0.7204.100

### Tanstack Query adapter

react-query

### TanStack Query version

5.82.0

### TypeScript version

5.8.3

### Additional context

Initially, the issue was thought to be related to tRPC, so there has been some discussion about the error over there: https://github.com/trpc/trpc/issues/6740

Would highly recommend reading some of the comments from that issue. Could maybe also give some more context.

TkDodo:
@Ephem this seems to be related to `tryResolveSync` that was recently added. See the linked trpc issue for more info please.

Ephem:
Yeah, I’m pretty positive that’s the cause. 😞 Suspense kind of acts like a synchronising factor so the SSR pass and hydration has the same status for the promise, but when Suspense is not used things clearly fail.

I need to think some more on it, but my first naive idea for a fix is treating all useQuery as pending during SSR+hydration if that query was hydrated via a promise (which I think was the case for _all_ queries like this before tryResolveSync, which instead was a problem with Suspense). 

This might be the time when we’ll want to implement getServerSnapshot in uSES. 🤔 That would have pending but the getSnapshot would have success and React would patch things up correctly.

TkDodo:
> This might be the time when we’ll want to implement getServerSnapshot in uSES

we already have an issue for this for quite some time 😅 

- #4690

Ephem:
[...]

PR #9572, opened 2025-08-17
fix(react-query): resolve hydration mismatch in SSR with prefetched queries

joseph0926:
Issue: #9399 
Comment: https://github.com/TanStack/query/issues/9399#issuecomment-3194133497

## Description

This PR fixes a hydration mismatch issue that occurs when using `prefetchQuery` in Next.js App Router with React Query v5.

## The Problem

I briefly looked into the issue.
It's really interesting…

I added a few debugging logs to the reporter's code and checked the results.

```tsx
"use client";

import { useQuery, useQueryClient } from "@tanstack/react-query";
import { getAmountOfUsers } from "./fetcher";
import { useEffect, useState } from "react";

const PageClient = ({}) => {
  console.log("=== PageClient Render START ===");

  const cache = queryClient.getQueryCache().find({
    queryKey: ["users", "amount"],
  });

  console.log("!!! Cache:", {
    status: cache?.state.status,
    data: cache?.state.data,
    promise: cache?.promise,
    promiseType: typeof cache?.promise,
  });

  console.log("@@@ Before useQuery:", {
    cacheExists: !!cache,
    cacheState: cache?.state.status,
    cacheData: cache?.state.data,
  });

  const query = useQuery(getAmountOfUsers());

  console.log("### After useQuery:", {
    isLoading: query.isLoading,
    data: query.data,
    fetchStatus: query.fetchStatus,
  });

  console.log("=== PageClient Render END ===");

  return (
    <div>
      Amount of users:{" "}
      {query.isLoading ? <em>Loading...</em> : query.data?.amount}
    </div>
  );
};

export default PageClient;
```

I naturally expected that since the server prefetches and hydrates before sending to the client, if the server's `data` is `undefined`, the client would also have `undefined`, and if the server has data, the client would have it as well.

But the results were as follows

```
// client
=== PageClient Render START ===
page-client.tsx:15 !!! Cache: {status: 'success', data: {…}, promise: Promise, promiseType: 'object'}data: {amount: 12}promise: Promise {<fulfilled>: {…}, status: 'fulfilled', value: {…}}promiseType: "object"status: "success"[[Prototype]]: Object
page-client.tsx:22 @@@ Before useQuery: {cacheExists: true, cacheState: 'success', cacheData: {…}}cacheData: {amount: 12}cacheExists: truecacheState: "success"[[Prototype]]: Object
page-client.tsx:30 ### After useQuery: {isLoading: false, data: {…}, fetchStatus: 'fetching'}data: {amount: 12}fetchStatus: "fetching"isLoading: false[[Prototype]]: Object
page-client.tsx:36 === PageClient Render END ===

// server
!!! Cache: {
  status: 'pending',
  data: undefined,
  promise: Promise {}
  promiseType: 'object'
}

@@@ Before useQuery: { cacheExists: true, cacheState: 'pending', cacheData: undefined }
### After useQuery: { isLoading: true, data: undefined, fetchStatus: 'fetching' }
=== PageClient Render END ===
```

Sometimes it was `undefined` on the server, yet the client had data.

I thought about why this might be possible.

Possible hypothesis =>
// 1. RSC (React Server Component) stage

* Render `Layout`
* Run `page.tsx` (server component)
  → `getServerQueryClient()` creates QueryClient **A**
  → `prefetchQuery()` runs (queryFn execution "1")
  → `dehydrate(QueryClient A)` = serialize data

// 2. SSR (Server Side Rendering) stage

* `ClientProvider` runs
  → `createQueryClient()` creates QueryClient **B** (new, empty cache!)
* Render `PageClient` (uses QueryClient **B**)
  → Cache is empty → `useQuery` starts fetch (queryFn execution "2")
  → `status: 'pending'`, `isLoading: true`
  → HTML: `<em>Loading...</em>`

// 3. RSC payload generation

* Next.js injects the result of queryFn **2** into the RSC payload
  → e.g., `{"amount":14}`
  This is what I actually observed (you can find this injection in the Network tab > search for `amount` under `localhost`)

```json
<script>
  self.__next_f.push([1, "...59:{\"amount\":12}..."])
</script>
```

Then the client state becomes
// 1. Initial state

* `ClientProvider`: creates QueryClient **C** (new)
* `HydrationBoundary`: receives the dehydrated state from QueryClient **A**

// 2. During hydration

* In `HydrationBoundary`'s `useMemo`
  → Finds new queries → hydrates immediately
  → Creates a Promise-like object
  → Fills it with data from the RSC payload

// 3. React Query v5 `tryResolveSync`

* Detects an already resolved Promise-like object
* Extracts data synchronously!
* `status: 'success'`, `data: { amount: 14 }`

// 4. First render

* `isLoading: false`
* DOM shows `"14"`
* React: HTML mismatch! (`<em>Loading...</em>` vs `"14"`)

---

In summary, the core points seem to be

1. Multiple `QueryClient` instances are created independently, fragmenting state across them.
2. There's a collision (or unintended interaction) between the RSC payload and React Query's `tryResolveSync` behavior.

## The Solution

To resolve this hydration mismatch, I implemented `getServerSnapshot` support in `useSyncExternalStore`

### 1. Added `getServerResult()` method to QueryObserver

This method detects hydrated data (where `dataUpdatedAt === 0`) and returns a pending state for server-side rendering

```typescript
getServerResult(): QueryObserverResult<TData, TError> {
  // If data was hydrated (dataUpdatedAt === 0), 
  // return pending state for server to match initial HTML
  if (isHydratedData) {
    return {
      ...currentResult,
      status: 'pending',
      isPending: true,
      isLoading: true,
      data: undefined,
    }
  }
  return currentResult
}
```

### 2. Updated `useBaseQuery` to use proper server snapshot

```typescript
React.useSyncExternalStore(
  subscribe,
  () => observer.getCurrentResult(),  // client snapshot
  () => observer.getServerResult(),   // server snapshot
)
```

## Testing

I've added comprehensive tests to verify the fix

```typescript
describe('SSR Hydration', () => {
  test('should demonstrate hydration mismatch issue (before fix)')
  test('getServerResult should return pending state for hydrated data')
  test('should handle fetching state during hydration')
  test('should return normal result for non-hydrated data')
  test('should handle error state correctly')
  test('should provide different snapshots for server and client')
})
```

All tests are passing

## Build & Test Results

I've successfully run the following locally
- `pnpm test` - All tests passing
- `pnpm build:all` - Build successful
- `pnpm test:types` - Type checks passing

## Related Issues

- Fixes the issue discussed in #9399 
- Implements the solution suggested by @Ephem regarding `getServerSnapshot`

## Breaking Changes

None. This is a backward-compatible fix that only affects SSR behavior.

### DIFF
--- a/packages/query-core/src/queryObserver.ts
+++ b/packages/query-core/src/queryObserver.ts
@@ -264,6 +264,34 @@ export class QueryObserver<
     return this.#currentResult
   }
 
+  getServerResult(): QueryObserverResult<TData, TError> {
+    const currentResult = this.#currentResult
+
+    if (
+      currentResult.status === 'success' &&
+      this.#currentQuery.state.dataUpdatedAt === 0
+    ) {
+      const pendingResult: QueryObserverResult<TData, TError> = {
+        ...currentResult,
+        status: 'pending',
+        isPending: true,
+        isSuccess: false,
+        isError: false,
+        isLoading: currentResult.fetchStatus === 'fetching',
+        isInitialLoading: currentResult.fetchStatus === 'fetching',
+        data: undefined,
+        error: null,
+        isLoadingError: false,
+        isRefetchError: false,
+        isPlaceholderData: false,
+      } as QueryObserverResult<TData, TError>
+
+      return pendingResult
+    }
+
+    return currentResult
+  }
+
   trackResult(
     result: QueryObserverResult<TData, TError>,
     onPropTracked?: (key: keyof QueryObserverResult) => void,
--- a/packages/query-core/src/__tests__/queryObserver.test.tsx
+++ b/packages/query-core/src/__tests__/queryObserver.test.tsx
@@ -1494,4 +1494,179 @@ describe('queryObserver', () => {
       unsubscribe2()
     })
   })
+
+  describe('SSR Hydration', () => {
+    describe('Hydration Mismatch Problem', () => {
+      test('should demonstrate hydration mismatch issue (before fix)', () => {
+        const key = queryKey()
+
+        queryClient.setQueryData(key, { amount: 10 })
+        const cache = queryClient.getQueryCache().find({ queryKey: key })
+        if (cache) {
+          cache.state.dataUpdatedAt = 0
+          cache.state.fetchStatus = 'idle'
+        }
+
+        const observer = new QueryObserver(queryClient, {
+          queryKey: key,
+          queryFn: () => ({ amount: 10 }),
+        })
+
+        const clientResult = observer.getCurrentResult()
+
+        expect(clientResult).toMatchObject({
+          status: 'success',
+          data: { amount: 10 },
+          isLoading: false,
+          isPending: false,
+        })
+      })
+    })
+
+    describe('Solution with getServerResult', () => {
+      test('getServerResult should return pending state for hydrated data', () => {
+        const key = queryKey()
+
+        queryClient.setQueryData(key, { amount: 10 })
+        const cache = queryClient.getQueryCache().find({ queryKey: key })
+        if (cache) {
+          cache.state.dataUpdatedAt = 0
+          cache.state.fetchStatus = 'idle'
+        }
+
+        const observer = new QueryObserver(queryClient, {
+          queryKey: key,
+          queryFn: () => ({ amount: 10 }),
+        })
+
+        const clientResult = observer.getCurrentResult()
+        expect(clientResult).toMatchObject({
+          status: 'success',
+          data: { amount: 10 },
+          isLoading: false,
+        })
+
+        const serverResult = observer.getServerResult()
+        expect(serverResult).toMatchObject({
+          status: 'pending',
+          data: undefined,
+          isLoading: false,
+          isPending: true,
+          isSuccess: false,
+        })
+      })
+
+      test('should handle fetching state during hydration', () => {
+        const key = queryKey()
+
+        queryClient.setQueryData(key, { amount: 10 })
+        const cache = queryClient.getQueryCache().find({ queryKey: key })
+        if (cache) {
+          cache.state.dataUpdatedAt = 0
+          cache.state.fetchStatus = 'fetching'
+        }
+
+        const observer = new QueryObserver(queryClient, {
+          queryKey: key,
+          queryFn: () => ({ amount: 10 }),
+        })
+
+        const serverResult = observer.getServerResult()
+        expect(serverResult).toMatchObject({
+          status: 'pending',
+          fetchStatus: 'fetching',
+          isLoading: true,
+          isPending: true,
+        })
+      })
+
+      test('should return normal result for non-hydrated data', () => {
+        const key = queryKey()
+
+        queryClient.setQueryData(key, { amount: 10 })
+        const cache = queryClient.getQueryCache().find({ queryKey: key })
+        if (cache) {
+          cache.state.dataUpdatedAt = Date.now()
+        }
+
+        const observer = new QueryObserver(queryClient, {
+          queryKey: key,
+          queryFn: () => ({ amount: 10 }),
+        })
+
+        const clientResult = observer.getCurrentResult()
+        const serverResult = observer.getServerResult()
+
+        expect(serverResult.status).toBe(clientResult.status)
+        expect(serverResult.data).toBe(clientResult.data)
+        expect(serverResult.isLoading).toBe(clientResult.isLoading)
+      })
+
+      test('should handle error state correctly', () => {
+        const key = queryKey()
+        const error = new Error('fetch error')
+
+        queryClient.getQueryCache().build(
+          queryClient,
+          {
+            queryKey: key,
+            queryFn: () => Promise.reject(error),
+          },
+          {
+            status: 'error',
+            error,
+            data: undefined,
+            dataUpdatedAt: 0,
+            dataUpdateCount: 0,
+            errorUpdateCount: 1,
+            errorUpdatedAt: Date.now(),
+            fetchFailureCount: 1,
+            fetchFailureReason: error,
+            fetchMeta: null,
+            fetchStatus: 'idle',
+            isInvalidated: false,
+          },
+        )
+
+        const observer = new QueryObserver(queryClient, {
+          queryKey: key,
+          queryFn: () => Promise.reject(error),
+        })
+
+        const serverResult = observer.getServerResult()
+
+        expect(serverResult).toMatchObject({
+          status: 'error',
+          error,
+          isError: true,
+        })
+      })
+    })
+
+    describe('Integration with useSyncExternalStore', () => {
+      test('should provide different snapshots for server and client', () => {
+        const key = queryKey()
+
+        queryClient.setQueryData(key, 'hydrated')
+        const cache = queryClient.getQueryCache().find({ queryKey: key })
+        if (cache) {
+          cache.state.dataUpdatedAt = 0
+        }
+
+        const observer = new QueryObserver(queryClient, {
+          queryKey: key,
+          queryFn: () => 'data',
+        })
+
+        const getSnapshot = () => observer.getCurrentResult()
+        const getServerSnapshot = () => observer.getServerResult()
+
+        const clientSnapshot = getSnapshot()
+        const serverSnapshot = getServerSnapshot()
+
+        expect(clientSnapshot.status).toBe('success')
+        expect(serverSnapshot.status).toBe('pending')
+      })
+    })
+  })
 })
--- a/packages/react-query/src/useBaseQuery.ts
+++ b/packages/react-query/src/useBaseQuery.ts
@@ -109,7 +109,7 @@ export function useBaseQuery<
       [observer, shouldSubscribe],
     ),
     () => observer.getCurrentResult(),
-    () => observer.getCurrentResult(),
+    () => observer.getServerResult(),
   )
 
   React.useEffect(() => {


packages/query-core/src/queryObserver.ts
```diff
@@ -264,6 +264,34 @@ export class QueryObserver<
...
+        isPending: true,
+        isSuccess: false,
+        isError: false,
+        isLoading: currentResult.fetchStatus === 'fetching',
+        isInitialLoading: currentResult.fetchStatus === 'fetching',
+        data: undefined,
+        error: null,
+        isLoadingError: false,
+        isRefetchError: false,
+        isPlaceholderData: false,
```
TkDodo: how did you come up with this list of fields to change? For example, why aren’t fields like `isFetched` or `isStale` being re-set?

I’m wondering how we can make sure to keep this list in-sync if we add new properties to `QueryObserverResult`...

packages/query-core/src/queryObserver.ts
```diff
@@ -264,6 +264,34 @@ export class QueryObserver<
...
   }
 
+  getServerResult(): QueryObserverResult<TData, TError> {
+    const currentResult = this.#currentResult
+
+    if (
+      currentResult.status === 'success' &&
+      this.#currentQuery.state.dataUpdatedAt === 0
+    ) {
+      const pendingResult: QueryObserverResult<TData, TError> = {
```
TkDodo: should this be of type `QueryObserverPendingResult` to make sure we’re at least always in-sync with what this type guarantees ?

TkDodo (commented):
Thanks for looking into this 🙌

packages/query-core/src/queryObserver.ts
joseph0926: Thank you for pointing out the type > `QueryObserverPendingResult`, which I had overlooked.
I agree that applying this will improve type stability.
However, since `currentResult` must be of type `QueryObserverResult`, I handled it as a union.

[ab2ddad](https://github.com/TanStack/query/pull/9572/commits/ab2ddadd2a780293d9404d7ab47c316461b34197)

packages/query-core/src/queryObserver.ts
joseph0926: Thank you for your feedback!

In conclusion, the criteria for selecting the specified fields was “only change fields that directly affect SSR hydration inconsistencies.”

I believe that intentionally maintaining the currentResult value for the remaining fields will be beneficial during subsequent maintenance processes.

However, based on coderabbitai's feedback, I realized that I had overlooked `isRefetching`, so I added it.
Since the status is fixed as `pending`, I added it to explicitly state that refetching is not possible.

[cb50d25](https://github.com/TanStack/query/pull/9572/commits/cb50d25b8b099a04e3753d18accee4d4e5e6865b)

joseph0926:
@TkDodo 
It seems like the more I edit, the more complicated it gets... What do you think?

TkDodo:
I’m not an expert here so I’ll wait for @Ephem to jump in but the way I see it, the `getServerSnapshot` should always return the value that we get from the server render; In that sense, we should probably “only” need to (conceptually) put that on a separate context and then read it from there

joseph0926:
> I’m not an expert here so I’ll wait for @Ephem to jump in but the way I see it, the `getServerSnapshot` should always return the value that we get from the server render; In that sense, we should probably “only” need to (conceptually) put that on a separate context and then read it from there

@TkDodo 

Hmm, I understand what you're saying.
The React documentation states, “The server snapshot must be the same between the client and the server, and is usually serialized and passed from the server to the client.”

My PR currently checks the status of currentResult at runtime and changes it dynamically, so there is a possibility of inconsistency.
Additionally, from a maintenance perspective, while currently checking `currentResult.status === ‘success’` may suffice, it's unclear how things will change in the future, so there's the issue of having to modify the if statement each time.

The Context approach you suggested preserves the value rather than transforming it, which aligns more closely with React's philosophy.
Furthermore, since Context “preserves” a fixed value and only “returns” it based on the environment, from a maintenance perspective, it doesn't matter if the state changes or not, so it's better.

I'm also curious about @Ephem  final opinion.

For now, I see two paths forward
1. Keep the current implementation as it solves the immediate problem
2. Refactor to use Context approach for better alignment with React's philosophy

For now, I will wait for @Ephem  response and refactor this PR using the Context method.

---

+++
I tried to implement it using the Context method, but I got stuck on the core part.  
The Context method ultimately solves the problem by “preserving” and ‘transmitting’ rather than “converting,”  
but the problem is that the “original server state” to be preserved does not exist in the first place.
1. After prefetching on the server → the cache is already in a success state  
2. During component rendering → the success state is read as-is  
3. The fact that it was “rendered as pending” is a fiction

In other words, the “actual pending state created by the server” that we want to put into Context has never actually existed, has it?

The server only knows the “success” state from start to finish, and the “pending” state we desire is merely our wish for how it should appear. Therefore, the Context approach ultimately
```typescript
// Not “preservation” but still “creation/transformation”
const serverSnapshot = {
  status: ‘pending’,  // artificially created
  data: undefined,    // hiding the actual data
}
```

This is an artificially created state being placed into the Context, not preserving the actual server state.

joseph0926:
[25fc95e](https://github.com/TanStack/query/pull/9572/commits/25fc95ed0f6bbf232e9763697bd7e9090d7c2225)

I added comprehensive test cases to verify the issues you raised and obtained the following results. 

[Test 1](https://github.com/TanStack/query/pull/9572/commits/25fc95ed0f6bbf232e9763697bd7e9090d7c2225#diff-d216de90023f217929be4053db0d25b8033d9e5141a70379a0e13896d2f31ddeR1672)
This test verified that “the masking logic only handles the ‘success’ state and omits other potential success-like states.” If new state types such as ‘idle’ are added in future versions, the current implementation will pass without change. In other words, we confirmed that problems such as “the state is ‘idle’ but data exists” could occur.

[Test 2](https://github.com/TanStack/query/pull/9572/commits/25fc95ed0f6bbf232e9763697bd7e9090d7c2225#diff-d216de90023f217929be4053db0d25b8033d9e5141a70379a0e13896d2f31ddeR1708)
The current PR implementation correctly handles field consistency. When the client is `isRefetching: true` and `isPending: false`, the server view correctly masks this as `isRefetching: false` and `isPending: true` to maintain SSR/CSR consistency.

[Test 3](https://github.com/TanStack/query/pull/9572/commits/25fc95ed0f6bbf232e9763697bd7e9090d7c2225#diff-d216de90023f217929be4053db0d25b8033d9e5141a70379a0e13896d2f31ddeR1736)
Although not directly related to the above feedback, we discovered an issue during testing.
When `refetch()` is called on hydrated data, the observer's snapshot mechanism can cause `currentResult.fetchStatus` to become an outdated value. To resolve this, we used `queryState.fetchStatus` instead.

```typescript
// Before
isLoading: currentResult.fetchStatus === ‘fetching’,

// After  
isLoading: queryState.fetchStatus === ‘fetching’,
```

This change ensures that the real-time fetch status is reflected even during concurrent operations.

TkDodo:
@Ephem said he might have time this week to look into it, so let’s wait for his feedback

joseph0926:
I’d like to share a few additional findings regarding this issue

1. Based on my testing, it occurs only in Next.js specifically when using React Server Components
2. The same behavior also appears with `useQueries`.

Below are links to reproduction code for both scenarios.

[react router](https://stackblitz.com/edit/github-7f512ght?file=app%2Froutes%2Fhome.tsx)
[useQuries Issue](https://stackblitz.com/edit/nextjs-w8dgcpja?file=app%2Fpage.tsx)

joseph0926:
af8b955

Initially, i used the spread operator to only explicitly modify the fields that needed to be changed while preserving the rest. 
However, i realized that a pending state is not simply a success state with a few modified fields - it's a fundamentally different state that requires all fields to be explicitly defined.
Through this improvement, we expect enhanced predictability and maintainability going forward (for example, type errors will be immediately visible). Additionally, since the definition of a server snapshot is to 'reconstruct the state as the server would have rendered it,' we concluded that all fields should be intentionally and explicitly set, which led to this modification.

joseph0926:
1bf795b

It appears that useQueries has the same hydration issue. I've added getServerResult() to QueriesObserver which seems to fix it, along with relevant tests. Would appreciate confirmation that this approach makes sense.

[useQuries Issue](https://stackblitz.com/edit/nextjs-w8dgcpja?file=app%2Fpage.tsx)

Icestonks:
Hey, is there any update on this? Did @Ephem ever take a look at it?

joseph0926:
> Hey, is there any update on this? Did @Ephem ever take a look at it?

I'm unsure if my PR addresses the core issue and is structurally sound, so I intended to close it and make corrections if his feedback indicated it was wrong. However, since I haven't received feedback yet, I've closed it for now.
If he provides feedback confirming the direction is correct, I'll reopen it and continue making adjustments.
Alternatively, having someone else reopen the PR could also be an option.

DeepL로 번역함 (https://dee.pl/apps)

Ephem:
Sorry this has taken so long, life has been incredibly hectic so had to step away from OSS for a bit to stay sane. 😄

Despite what I said in the issue, I'm not sure if hard-coding the pending state is the right approach here. What we'd ideally do is snapshot a immutable version of the initial payload that gets passed to `HydrationBoundary` and use that for the `getServerSnapshot`. I haven't looked deeply enough at this yet to know if that is feasible though, it gets tricky when multiple `HydrationBoundary` are involved, especially when they have the same query. This version might be good enough as a short/medium/long term fix, but I need to do some more digging to say for sure.

I've scheduled time on Friday to look deeper into this!

joseph0926:
I attempted to store immutable data and utilize it, but structural changes would be required, which seems to go beyond the scope of this PR and is beyond my capabilities, so I stopped trying.
If needed, feel free to reopen this PR; otherwise, I'll keep it closed.

Ephem:
WIll do, I really appreciate you taking time to try to fix this! it might be a good fix while we work on the bigger structural changes, I'll reopen it in that case. Even if not, I'm sure some of those tests will come in useful, so thank you. 😄

dBianchii:
Hi, I think the issue I am experiencing is this exact thing, where a prefetch on the server first SSRs a "loading" state because of query.isLoading. Then, at time of hydration the data is already available and renders the new tree with the correct loaded state. Has there been any progress on this? Any workarounds until this is fixed?

PR #9575, opened 2025-08-18
docs(react-query): recommend `defaultError = unknown` instead of `AxiosError`

leesb971204:
This PR updates the TypeScript docs for **Registering a global error**:

- **Changed the example** to recommend `defaultError = unknown` instead of `AxiosError`.  
- **Added rationale:**  
  - By default, the error type in TanStack Query is inferred as `Error`.  
  - However, if developers want to make the error type safer and more explicit, they can override the global default to `unknown`.  
  - Errors in TanStack Query are not limited to your `queryFn`; they may also come from `select`, memoized callbacks, or other user-land code.  
  - Using `AxiosError` globally can mislead developers into thinking all errors originate from Axios, which is incorrect.  

discussions in [here](https://discord.com/channels/719702312431386674/1405082270125264987)

## Summary by CodeRabbit

* **Documentation**
  * Updated TypeScript guidance to prefer using unknown for documented error types (defaultError) so call sites must explicitly narrow errors.
  * Revised examples across React, Solid, Angular and Vue docs to show error typed as unknown | null with comments demonstrating required narrowing.
  * Improved snippet formatting and added anchor markers for clarity.
  * Clarified these are documentation-only, non-runtime changes.

### DIFF
--- a/docs/framework/vue/typescript.md
+++ b/docs/framework/vue/typescript.md
@@ -89,6 +89,21 @@ if (error.value instanceof Error) {
 [//]: # 'TypingError3'
 [//]: # 'TypingError3'
 [//]: # 'RegisterErrorType'
+
+```tsx
+import '@tanstack/vue-query'
+
+declare module '@tanstack/vue-query' {
+  interface Register {
+    // Use unknown so call sites must narrow explicitly.
+    defaultError: unknown
+  }
+}
+
+const { error } = useQuery({ queryKey: ['groups'], queryFn: fetchGroups })
+//      ^? const error: unknown | null
+```
+
 [//]: # 'RegisterErrorType'
 [//]: # 'TypingMeta'
 [//]: # 'TypingMeta'


docs/framework/react/typescript.md
```diff
@@ -130,24 +130,63 @@ if (axios.isAxiosError(error)) {
 
 ### Registering a global Error
 
-TanStack Query v5 allows for a way to set a global Error type for everything, without having to specify generics on call-sides, by amending the `Register` interface. This will make sure inference still works, but the error field will be of the specified type:
+TanStack Query v5 allows for a way to set a global Error type for everything, without having to specify generics on call-sides, by amending the `Register` interface. This will make sure inference still works, but the error field will be of the specified type. We recommend `unknown` to ensure safe, explicit narrowing at call sites, by amending the `Register` interface:
```
TkDodo:
```suggestion
TanStack Query v5 allows for a way to set a global Error type for everything, without having to specify generics on call-sides, by amending the `Register` interface. This will make sure inference still works, but the error field will be of the specified type. If you want to enforce that call-sides must do explicit type-narrowing, set `defaultError` to `unknown`:
```

docs/framework/react/typescript.md
```diff
@@ -130,24 +130,63 @@ if (axios.isAxiosError(error)) {
...
 
 [//]: # 'RegisterErrorType'
 
 ```tsx
 import '@tanstack/react-query'
 
+// ✅ Recommended
```
TkDodo: If this were “recommended”, we’d have kept it as `unknown` per default instead of `Error`. `Error` will work for 99.99% of cases, so going back to `unknown` is something you can do if you want the safest possible variant because you explicitly throw things that aren’t errors (which also isn’t recommended!), but let’s please not document that this is “recommended”.

docs/framework/react/typescript.md
```diff
@@ -130,24 +130,63 @@ if (axios.isAxiosError(error)) {
...
 }
 
 const { error } = useQuery({ queryKey: ['groups'], queryFn: fetchGroups })
-//      ^? const error: AxiosError | null
+//      ^? const error: unknown | null
 ```
 
 [//]: # 'RegisterErrorType'
+
+#### Why `unknown` instead of `AxiosError`?
```
TkDodo: Please remove this whole paragraph. The reference to `AxiosError` makes no sense anymore since it was removed from the previous paragraph, and it’s just confusing, and the pattern with try/catch in the queryFn is not something I would recommend. Instead, the API layer should likely already throw the type of error you want, which is also why `axios` or `ky` are a good idea. But that’s all speculative, people can do whatever they want inside the queryFn, doesn’t even have to be data fetching. So let’s remove this please

leesb971204:
@TkDodo 

Thank you for the feedback! I’ve applied your suggestions.
Should I reflect this change in the other TypeScript docs as well (Vue, Solid, Angular)?
Interestingly, Vue Query doesn’t even mention a global Error type.

TkDodo:
yes please, you probably only need to change the examples

leesb971204:
@TkDodo 

Solid didn’t have the `[//]: # 'RegisterErrorType'` section,
Vue didn’t have any example code at all,
and for Angular, only the example code needed to be updated.

all done! 🫡

PR #9576, opened 2025-08-18
fix(core): do not reject promise right away - only when `promise` gets used

### LINKED ISSUE #8209: [solid-query] Uncaught (in promise) Error: experimental_prefetchInRender feature flag is not enabled
### Describe the bug

When using the latest version of tanstack-query you get at least in solidjs for each query this error message in the console:
```
Uncaught (in promise) Error: experimental_prefetchInRender feature flag is not enabled
```

https://codesandbox.io/p/github/BierDav/temp-tanstack-query-ssr-bug/experimental-flag-warninghttps://codesandbox.io/p/github/BierDav/temp-tanstack-query-ssr-bug/experimental-flag-warning

### Steps to reproduce

1. Just start it using `bun i && bun dev`

### Expected behavior

No error message in the console, because the `promise` wasn't even accessed

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: Windows, Linux

### Tanstack Query adapter

solid-query

### TanStack Query version

v5.59.13

### TypeScript version

_No response_

### Additional context

I have looked into the code. It is not really that something went wrong, and more about that the promise which causes this issue is evaluated even though the feature is not set.

Probably the easiest solution is to replace the promise when the flag is not enable with a `getter` function which then only throws the error, when the promise is really accessed.

TkDodo:
fixes #8209

## Summary by CodeRabbit

- New Features
  - None.

- Bug Fixes
  - Prevented premature errors when the experimental prefetch-in-render flag is disabled; errors now occur only if a query’s promise is explicitly accessed.

- Refactor
  - Switched from eager to lazy error handling to avoid initialization failures and preserve normal behavior unless the promise is used.

### DIFF
--- a/packages/query-core/src/queryObserver.ts
+++ b/packages/query-core/src/queryObserver.ts
@@ -267,9 +267,12 @@ export class QueryObserver<
       get: (target, key) => {
         this.trackProp(key as keyof QueryObserverResult)
         onPropTracked?.(key as keyof QueryObserverResult)
-        if (key === 'promise' && !this.options.experimental_prefetchInRender) {
-          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-          ;(this.#currentThenable as PendingThenable<TData>).reject?.(
+        if (
+          key === 'promise' &&
+          !this.options.experimental_prefetchInRender &&
+          this.#currentThenable.status === 'pending'
+        ) {
+          this.#currentThenable.reject(
             new Error(
               'experimental_prefetchInRender feature flag is not enabled',
             ),


packages/query-core/src/queryObserver.ts
```diff
@@ -272,6 +267,15 @@ export class QueryObserver<
       get: (target, key) => {
         this.trackProp(key as keyof QueryObserverResult)
         onPropTracked?.(key as keyof QueryObserverResult)
+        console.log('tracking prop', key)
```
TkDodo:
```suggestion
```

packages/query-core/src/queryObserver.ts
```diff
@@ -272,6 +267,14 @@ export class QueryObserver<
       get: (target, key) => {
         this.trackProp(key as keyof QueryObserverResult)
         onPropTracked?.(key as keyof QueryObserverResult)
+        if (key === 'promise' && !this.options.experimental_prefetchInRender) {
+          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+          ;(this.#currentThenable as PendingThenable<TData>).reject?.(
+            new Error(
+              'experimental_prefetchInRender feature flag is not enabled',
+            ),
+          )
```
TkDodo: @KATT the type assertion + optional chaining is necessary because we might not have a `PendingThenable` here and the generic `Thenable` doesn’t have a `reject` method :/

I don’t fully understand why that is but I’m still shipping this fix 😂

TkDodo: right, I need a status check I think: 

4a51a545

PR #9577, opened 2025-08-18
fix(core): never revert to undefined data

TkDodo:
## Summary by CodeRabbit

- **New Features**
  - Added a public CancelledError to surface cancellation outcomes.

- **Bug Fixes**
  - Cancelling an initial in-flight fetch now raises CancelledError instead of resolving with missing/undefined data; previous data is preserved when available.

- **Tests**
  - Refactored and added tests to validate cancellation behavior, error propagation, and post-cancellation state.

### DIFF
--- a/packages/query-core/src/query.ts
+++ b/packages/query-core/src/query.ts
@@ -559,10 +559,11 @@ export class Query<
           })
           // transform error into reverted state data
           // if the initial fetch was cancelled, we have no data, so we have
-          // to get into error state with the CancelledError
-          if (this.state.data !== undefined) {
-            return this.state.data
+          // to get reject with a CancelledError
+          if (this.state.data === undefined) {
+            throw error
           }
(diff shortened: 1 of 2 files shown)


packages/query-core/src/query.ts
```diff
@@ -558,7 +558,12 @@ export class Query<
             fetchStatus: 'idle' as const,
           })
           // transform error into reverted state data
-          return this.state.data!
+          // if the initial fetch was cancelled, we have no data, so we have
+          // to get reject with a CancelledError
+          if (this.state.data === undefined) {
+            throw error
+          }
+          return this.state.data
```
TkDodo, line 566: this changes makes sure we never resolve with `undefined` for imperative methods like `queryClient.fetchQuery(...)`. The bang operator `!` was a dead giveaway that this could happen 😅

Issue #9579, opened 2025-08-19
v5.85.2 introduced a breaking change while using StrictMode


pvs-chainalysis:
### Describe the bug

v5.85.2 introduced a breaking change while using `StrictMode` impacting state of `isLoading` and `isFetching`

On initial load state of `isLoading` doesn't reflect described API behaviour, API states that `isLoading` reflects the state of the first fetch, not subsequent fetches. However with `StrictMode` enabled `isLoading` reverts to `false` before the first request finishes 

Example: https://stackblitz.com/edit/tanstack-query-2c4dbawu?file=src%2Findex.tsx

Example above updates DOM as request state changes:  `req1: isLoading:false, state: success, fetching: false data: request 1, time=1755607826585`. `time=` is added when request is finished

### Your minimal, reproducible example

https://stackblitz.com/edit/tanstack-query-2c4dbawu?file=src%2Findex.tsx

### Steps to reproduce

1. On initial load observe that `isLoading: true`, `state: pending` and `fetching: false`
2. Observe `isLoading` quickly changes to `false` but `state` remains `pending`.

Before `v5.85.2` `isLoading` would remain true until first request completes. If `StrictMode` is removed `isLoading` behaves as expected

Downgrade to 5.85.0 `npm install @tanstack/react-query-devtools@5.85.0 @tanstack/react-query@5.85.0` and run the example again `npm run dev` and observe `isLoading` behaves as expected.

### Expected behavior

At first load `isLoading` and `isFetching` should reflect the `status` of the first request.

### How often does this bug happen?

Every time

### Screenshots or Videos

https://github.com/user-attachments/assets/58464f7e-c7bf-4246-a0c3-3debf3e97882

### Platform

- OS: macOS
- Browser: Chrome
- Version: 5.85.2

### Tanstack Query adapter

None

### TanStack Query version

5.85.2

TkDodo:
Slight correction: It seems the behaviour first occurred in 5.85.2

joseph0926:
This issue does not appear to be solely related to `StrictMode`, but rather occurs when using `signal` within the `queryFn` under `StrictMode`.
In fact, if you remove the `signal` from the provided code as shown below,

```tsx
queryFn: () => {
  return new Promise((resolve) => {
    setTimeout(
      () => {
        const now = Date.now();
        console.log(`fetchCompleted index=${index}, time=${now}`);
        resolve(`request ${index}, time=${now}`);
      },
      4000,
      // { signal }
    );
  });
[...]

PR #9580, opened 2025-08-20
fix(query-core): prevent state override when observer remount occurs …

joseph0926:
fixes: #9579 
comment: [#9579 (comment)](https://github.com/TanStack/query/issues/9579#issuecomment-3203591586)

## Description

This PR fixes an issue where `isLoading` incorrectly becomes `false` during an active fetch when using React StrictMode with signal destructuring in the queryFn.

## Problem

This issue does not appear to be solely related to `StrictMode`, but rather occurs when using `signal` within the `queryFn` under `StrictMode`.
In fact, if you remove the `signal` from the provided code as shown below,
```tsx
queryFn: () => {
    setTimeout(
      () => {
        const now = Date.now();
        console.log(`fetchCompleted index=${index}, time=${now}`);
        resolve(`request ${index}, time=${now}`);
      },
      4000,
      // { signal }
    );
  });
},
```
`isLoading` will correctly become `true`.
The problem arises because when the `signal` option is present and the fetch starts, destructuring the signal sets `abortSignalConsumed = true` https://github.com/TanStack/query/blob/7c464e3ded7c4764170839b4aa7f763c7d0aa647/packages/query-core/src/query.ts#L428. Due to `StrictMode`, during cleanup the following condition is met inside `removeObserver`
```ts
if (this.#abortSignalConsumed) {
  this.#retryer.cancel({ revert: true });
}
```
https://github.com/TanStack/query/blob/7c464e3ded7c4764170839b4aa7f763c7d0aa647/packages/query-core/src/query.ts#L351-L352
This triggers an asynchronous execution of `this.#retryer.cancel({ revert: true })`.
As a result, the sequence becomes: immediate re-mount → new observer added → second fetch begins → meanwhile, the asynchronous `catch` logic from the first fetch executes, which then hits the following branch
```ts
else if (error.revert) {
  this.setState({
    ...this.#revertState,
    fetchStatus: 'idle' as const,
  })
}
```
https://github.com/TanStack/query/blob/7c464e3ded7c4764170839b4aa7f763c7d0aa647/packages/query-core/src/query.ts#L555-L567
This causes the state to be overwritten with the initial values, leading to the observed issue.
The reason it did not occur in the previous version is that `onError` existed and handled the case, but now the logic is fully asynchronous and falls into `catch`, which results in the problem you are seeing.

## Solution

The fix adds a check to prevent state reversion when
1. The cancellation was triggered by observer removal (`isObserverRemoval` flag)
2. New observers have already been added (`this.observers.length > 0`)

This pattern is typical of React StrictMode's cleanup simulation where components are immediately remounted.

## Changes

- Added `isObserverRemoval` flag to `CancelOptions` and `CancelledError`
- Modified `removeObserver` to pass this flag when cancelling with revert
- Updated the catch block in `fetch` method to check for this condition before reverting state
- Added test case to verify the fix

## Testing

- [x] All existing tests pass
- [x] Added new test: `should not override fetching state when revert happens after new observer subscribes`
- [x] Manually tested with the reproduction case in StrictMode

### DIFF
--- a/packages/query-core/src/query.ts
+++ b/packages/query-core/src/query.ts
@@ -349,7 +349,7 @@ export class Query<
         // we'll let the query continue so the result can be cached
         if (this.#retryer) {
           if (this.#abortSignalConsumed) {
-            this.#retryer.cancel({ revert: true })
+            this.#retryer.cancel({ revert: true, isObserverRemoval: true })
           } else {
             this.#retryer.cancelRetry()
           }
@@ -553,16 +553,22 @@ export class Query<
           // so we hatch onto that promise
           return this.#retryer.promise
         } else if (error.revert) {
+          if (error.isObserverRemoval && this.observers.length > 0) {
+            if (this.state.data === undefined) {
+              throw error
+            }
+            return this.state.data
+          }
+
           this.setState({
             ...this.#revertState,
             fetchStatus: 'idle' as const,
           })
-          // transform error into reverted state data
-          // if the initial fetch was cancelled, we have no data, so we have
-          // to get reject with a CancelledError
+
           if (this.state.data === undefined) {
             throw error
           }
+
           return this.state.data
         }
       }
--- a/packages/query-core/src/__tests__/query.test.tsx
+++ b/packages/query-core/src/__tests__/query.test.tsx
@@ -1192,4 +1192,48 @@ describe('query', () => {
     expect(initialDataFn).toHaveBeenCalledTimes(1)
     expect(query.state.data).toBe('initial data')
   })
+
+  test('should not override fetching state when revert happens after new observer subscribes', async () => {
+    const key = queryKey()
+
+    // @ts-expect-error This field has been added for troubleshooting purposes. Disable ts error for testing.
+    const queryFn = vi.fn(async ({ signal }) => {
+      await sleep(50)
+      return 'data'
+    })
+
+    const query = new Query({
+      client: queryClient,
+      queryKey: key,
+      queryHash: hashQueryKeyByOptions(key),
+      options: { queryFn },
+    })
+
+    const observer1 = new QueryObserver(queryClient, {
+      queryKey: key,
+      queryFn,
+    })
+
+    query.addObserver(observer1)
+    const promise1 = query.fetch()
+
+    await vi.advanceTimersByTimeAsync(10)
+
+    query.removeObserver(observer1)
+
+    const observer2 = new QueryObserver(queryClient, {
+      queryKey: key,
+      queryFn,
+    })
+
+    query.addObserver(observer2)
+
+    query.fetch()
+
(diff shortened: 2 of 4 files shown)


joseph0926:
`@tanstack/solid-query-devtools:test:types`

```
src/devtoolsPanel.tsx(31,11): error TS2552: Cannot find name 'CSSStyleValue'. Did you mean 'CSSStyleRule'?
 ELIFECYCLE  Command failed with exit code 2.
ERROR: "test:types:ts50" exited with 2.
```

It seems that an error unrelated to this PR is occurring at the CI stage.

TkDodo:
Not sure about this fix. I'll have to think a bit about it but I'm on vacation now.

packages/query-core/src/query.ts
```diff
@@ -553,16 +553,25 @@ export class Query<
           // so we hatch onto that promise
           return this.#retryer.promise
         } else if (error.revert) {
+          // If cancellation was caused by observer removal and there are active observers again,
+          // do not revert to idle: a new fetch may already be in flight, and reverting would
```
TkDodo, line 557: I don’t think this is a good enough check and can easily introduce regressions again. For example, you could mount an observer that won’t trigger a fetch, like a disabled one or one that has `refetchOnMount: false`.

TkDodo:
I think I have a better fix here:
- https://github.com/TanStack/query/pull/9601

but I kept your test so thank you 🙏

packages/query-core/src/query.ts
joseph0926: Although it's already closed, I've made the fix based on your feedback.

Since `isActive()` checks “Is there an active observer?”, while `observers.length > 0` checks “Does an observer exist?”, this if-check is indeed correct.

```ts
          if (error.isObserverRemoval && this.observers.length > 0) {
            if (this.state.data === undefined) {
              throw error
            }
            return this.state.data
          }
```

// query.test.tsx
```tsx
  test('should prevent revert when disabled observer is added after removal', async () => {
    const key = queryKey()

    const queryFn = vi.fn(async ({ signal: _signal }) => {
      await sleep(50)
      return 'data'
    })

    const query = new Query({
      client: queryClient,
      queryKey: key,
      queryHash: hashQueryKeyByOptions(key),
      options: { queryFn },
    })

    const observer1 = new QueryObserver(queryClient, {
      queryKey: key,
      queryFn,
    })

    query.addObserver(observer1)
    const promise1 = query.fetch()

    query.removeObserver(observer1)

    const observer2 = new QueryObserver(queryClient, {
      queryKey: key,
      queryFn,
      enabled: false,
    })

    query.addObserver(observer2)

    await expect(promise1).rejects.toBeInstanceOf(CancelledError)

    expect(query.state.fetchStatus).toBe('fetching')
    expect(query.state.data).toBeUndefined()
  })
```

Issue #9586, opened 2025-08-23
`useQueries` doesn’t call `combine` when `StoragePersister` deserializes data


iamakulov:
### Describe the bug

Take some queries that are 1) allowed to be stale, 2) are combined into something:

```ts
const ids = [1, 2, 3];
const combinedQueries = useQueries({
  queries: ids.map((id) => ({
    queryKey: ["post", id],
    queryFn: () => Promise.resolve(id),
    staleTime: 30_000, // ← 1️⃣
  })),
  combine: i => i, // ← 2️⃣
});
```

and 3) are persisted in `localStorage`:

```ts
const persister = createAsyncStoragePersister({ storage: window.localStorage });
const queryClient = new QueryClient();

<PersistQueryClientProvider
  client={queryClient}
  persistOptions={{ persister }}
>
```

In this setup, when the persister finishes deserializing the queries, and the queries aren’t stale yet, **`useQueries` will not call `combine`** (unless `combine`’s identity changes). As a result, `combinedQueries` will remain in `isPending: true` state until `staleTime` expires.

This only reproduces with a custom `combine` function, even if it simply returns the same value.

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/wqrzzm

### Steps to reproduce

**To repro the issue:**
1. Open the codesandbox
2. Confirm the preview shows “Pending: false, false, false. Data: 1, 2, 3”
3. Reload the preview a few times
4. Confirm the preview shows “Pending: true, true, true. Data: , ,” every time

<details>

<summary>The issue disappears if you pass a different `combine` on every render</summary>

1. Unwrap `combine` on line 27 from `useCallback`
2. Reload the preview a few times
2. Confirm the preview shows “Pending: false, false, false. Data: 1, 2, 3” every time

</details>

<details>

<summary>The issue disappears if you don’t pass `combine` at all</summary>

1. Remove `combine` on line 38
2. Reload the preview a few times
2. Confirm the preview shows “Pending: false, false, false. Data: 1, 2, 3” every time

</details>

### Expected behavior

As a user, I expect to see “Pending: false, false, false. Data: 1, 2, 3” on every reload. However, I actually see “Pending: true, true, true. Data: , ,” starting with the second load.

### How often does this bug happen?

Every time

### Platform

- macOS 15.6
- Chrome 139

### Tanstack Query adapter

None

### TanStack Query version

v5.85.5

joseph0926:
[...]

PR #9592, opened 2025-08-26
fix(query-core): ensure combine re-executes after cache restoration with memoized combine

joseph0926:
Fixes: #9586 

## Fix combine function not re-executing after cache restoration with memoized combine

When using `useQueries` with a memoized `combine` function (via `useCallback`) alongside `PersistQueryClientProvider`, the UI fails to update after page refresh even though cached data exists.

### The Problem

We discovered this issue while working with React Query's persist functionality. After refreshing the page, the combine function would return stale pending state despite the cache containing valid data. The root cause was in `QueriesObserver`'s `setQueries` method.

During our debugging process, we traced through the execution flow and found that when `isRestoring` changes from `true` to `false`, `setQueries` gets called but hits an early return condition. Since the same observer instances are reused and no index changes occur, the method returns early without updating the results or notifying listeners.

The issue becomes more critical with React 19's automatic memoization, where all functions will be memoized by default.

### The Solution  

We modified the early return logic in `setQueries` to check if the actual results have changed. When observers haven't changed but the data or pending state differs, we now update `this.#result` and call `this.#notify()` to trigger the combine function re-execution.

```typescript
if (prevObservers.length === newObservers.length && !hasIndexChange) {
  const resultChanged = newResult.some((result, index) => {
    const prev = this.#result[index]
    return !prev || result.data !== prev.data || result.isPending !== prev.isPending
  })

  if (resultChanged) {
    this.#result = newResult
    this.#notify()
  }
  
  return
}
```

This ensures the UI updates correctly while maintaining the performance optimization of not recreating observers unnecessarily.

### Testing

Added test case to verify combine function works correctly with memoization and persist. All existing tests pass without modification.

### DIFF
--- a/packages/query-core/src/queriesObserver.ts
+++ b/packages/query-core/src/queriesObserver.ts
@@ -123,6 +123,20 @@ export class QueriesObserver<
       )
 
       if (prevObservers.length === newObservers.length && !hasIndexChange) {
+        const resultChanged = newResult.some((result, index) => {
+          const prev = this.#result[index]
+          return (
+            !prev ||
+            result.data !== prev.data ||
+            result.isPending !== prev.isPending
+          )
+        })
+
+        if (resultChanged) {
+          this.#result = newResult
+          this.#notify()
+        }
+
         return
       }
 
--- a/packages/react-query-persist-client/src/__tests__/use-queries-with-persist.test.tsx
+++ b/packages/react-query-persist-client/src/__tests__/use-queries-with-persist.test.tsx
@@ -1,0 +1,134 @@
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+import { render, waitFor } from '@testing-library/react'
+import * as React from 'react'
+import { QueryClient, useQueries } from '@tanstack/react-query'
+import { PersistQueryClientProvider } from '@tanstack/react-query-persist-client'
+import type {
+  PersistedClient,
+  Persister,
+} from '@tanstack/query-persist-client-core'
+import type { QueryObserverResult } from '@tanstack/react-query'
+
+describe('useQueries with persist and memoized combine', () => {
+  const storage: { [key: string]: string } = {}
+
+  beforeEach(() => {
+    Object.defineProperty(window, 'localStorage', {
+      value: {
+        getItem: (key: string) => storage[key] || null,
+        setItem: (key: string, value: string) => {
+          storage[key] = value
+        },
+        removeItem: (key: string) => {
+          delete storage[key]
+        },
+        clear: () => {
+          Object.keys(storage).forEach((key) => delete storage[key])
+        },
+      },
+      writable: true,
+    })
+  })
+
+  afterEach(() => {
+    Object.keys(storage).forEach((key) => delete storage[key])
+  })
+
+  it('should update UI when combine is memoized with persist', async () => {
+    const queryClient = new QueryClient({
+      defaultOptions: {
+        queries: {
+          staleTime: 30_000,
+          gcTime: 1000 * 60 * 60 * 24,
+        },
+      },
+    })
+
+    const persister: Persister = {
+      persistClient: (client: PersistedClient) => {
+        storage['REACT_QUERY_OFFLINE_CACHE'] = JSON.stringify(client)
+        return Promise.resolve()
+      },
+      restoreClient: async () => {
+        const stored = storage['REACT_QUERY_OFFLINE_CACHE']
+        if (stored) {
+          await new Promise((resolve) => setTimeout(resolve, 10))
+          return JSON.parse(stored) as PersistedClient
+        }
+        return undefined
+      },
+      removeClient: () => {
+        delete storage['REACT_QUERY_OFFLINE_CACHE']
+        return Promise.resolve()
+      },
+    }
+
+    const persistedData: PersistedClient = {
+      timestamp: Date.now(),
+      buster: '',
+      clientState: {
+        mutations: [],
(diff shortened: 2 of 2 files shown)


packages/query-core/src/queriesObserver.ts
```diff
@@ -123,6 +123,21 @@ export class QueriesObserver<
       )
 
       if (prevObservers.length === newObservers.length && !hasIndexChange) {
+        const resultChanged = newResult.some((result, index) => {
+          const prev = this.#result[index]
+          return (
+            !prev ||
+            result.data !== prev.data ||
+            result.isPending !== prev.isPending ||
+            result.error !== prev.error
```
TkDodo: checking for these 3 props seems a bit random. I think we’d want to use our `shallowEqualObjects` helper from the query-core to just compare `result` with `prev`. This is also what we do inside `queryObserver`:

https://github.com/TanStack/query/blob/abab082593424bd4b6a3c3e494f0cf4a8c40dcc1/packages/query-core/src/queryObserver.ts#L656-L659

packages/query-core/src/queriesObserver.ts
```diff
@@ -123,6 +123,21 @@ export class QueriesObserver<
...
+          )
+        })
+
+        if (resultChanged) {
+          this.#result = newResult
+          this.#notify()
+        }
```
TkDodo: I don’t really like the early returns here and the duplicated assignment to `#result` as well as calls to `#notify`, makes it kinda hard to read what is going on. I think we’d want something like this:

- first, check for `hasIndexChange`
  - if we have an index change, we want to compute the difference for observers and destroy / re-create them, then assign `this.#observers = newObservers`
- then, we check for `hasResultChange`
  - if that’s the case, we want to assign `this.#result = newResult`, then check for listeners, then call `this.#notify()`.

this flow ties logic better to the conditions: when indices change, we need to re-do observers and assign them; when result changes, we need to re-assign it. If anything changes, we need to `notify` (I’m assuming here that if indices change, we’ll also get a resultChange)

packages/query-core/src/queriesObserver.ts
joseph0926: I found that using shallowEqualObjects causes test failures because it compares ALL properties of QueryObserverResult, including timestamps like dataUpdatedAt. This means it detects changes even when the actual data hasn't changed. The individual property comparison (data, isPending, error) passes all tests. Should we stick with the explicit property checks, or would you prefer a different approach?

TkDodo: but if `dataUpdatedAt` changes, we need to inform observers about it, right? A component could  render `dataUpdatedAt` on the screen so it will want that information to be correct 🤔

packages/query-core/src/queriesObserver.ts
joseph0926: [999c8d2](https://github.com/TanStack/query/pull/9592/commits/999c8d2da2d28dcaddd1745a06b97585f6d22d7e)

Thank you for the detailed feedback! I've updated the code based on your suggestions.
To be honest, I was struggling to find the right approach, but the CodeRabbitAI feedback really helped clarify the path forward.

Would appreciate another review when you have time. Thanks!

packages/query-core/src/queriesObserver.ts
joseph0926: You're absolutely right. Components could render any property including dataUpdatedAt, so we need to detect all changes. I'll use shallowEqualObjects as you suggested.
The test failures appear to be because the tests expect combine not to re-run when metadata changes, but that expectation seems incorrect. When any property changes (even metadata), combine should re-execute to ensure the UI stays in sync.
Should I update the test expectations to reflect this correct behavior?

test file: packages/react-query/src/__tests__/useQueries.test.tsx > `should optimize combine if it is a stable reference`

joseph0926: [4bc297b](https://github.com/TanStack/query/pull/9592/commits/4bc297b538476c69ad2a48809193e4bf5be3c703)

I've updated the test expectations to account for the shallowEqualObjects comparison. The tests now expect combine to re-run when metadata changes occur.
However, I'm not entirely confident whether having combine re-run on metadata changes is the intended behavior from this PR... Perhaps there's a performance consideration I'm missing, or my implementation itself might be incorrect.
The tests are passing with the updated expectations, but I believe a different approach might be more appropriate. I defer to your judgment on whether this is the desired behavior.

TkDodo: yeah that test seems slightly off, and actually shows the bug: we get new metadata (`dataUpdatedAt`), which is part of `results` passed to `combine`, but we don’t run it :/

I think I was basing this test off of `select` on `useQuery`, but the difference is that `select` only gets `data` passed while `combine` gets passed _everything_.

> Should I update the test expectations to reflect this correct behavior?

yes please, and we need to test combine with a stable reference differently, likely by triggering a re-render differently, e.g. just having an `increment` that triggers an unrelated re-render, because in those cases, a stable combine shouldn’t re-run.

packages/react-query/src/__tests__/useQueries.test.tsx
TkDodo, line 1: please add another test that verifies that `combine` isn’t called when an unrelated re-render happens

joseph0926: > please add another test that verifies that `combine` isn’t called when an unrelated re-render happens


44bfedb

I have added the test.
Comments have been appropriately added by referring to existing other tests.

TkDodo (approved):

PR #9597, opened 2025-08-26
docs: mention second parameter for `cancelQueries`

Sc4ramouche:
The API page for `cancelQueries` does not mention that it accepts second optional parameter of `CancelOptions` ([docs page](https://tanstack.com/query/v5/docs/reference/QueryClient#queryclientcancelqueries)), which is indicated by the TypeScript signature.

### DIFF
--- a/docs/reference/QueryClient.md
+++ b/docs/reference/QueryClient.md
@@ -409,6 +409,7 @@ await queryClient.cancelQueries({ queryKey: ['posts'], exact: true })
 **Options**
 
 - `filters?: QueryFilters`: [Query Filters](../../framework/react/guides/filters.md#query-filters)
+- `cancelOptions?: CancelOptions`: [Cancel Options](../../framework/react/guides/filters.md#query-filters)
 
 **Returns**
 


docs/reference/QueryClient.md
```diff
@@ -409,6 +409,7 @@ await queryClient.cancelQueries({ queryKey: ['posts'], exact: true })
 **Options**
 
 - `filters?: QueryFilters`: [Query Filters](../../framework/react/guides/filters.md#query-filters)
+- `cancelOptions?: CancelOptions`: [Cancel Options](../../framework/react/guides/filters.md#cancel-options)
```
manudeli:
```suggestion
- `cancelOptions?: CancelOptions`: [Cancel Options](../../framework/react/guides/query-cancellation#XXX)
```

I think this would be a better place for `CancelOptions`. because `CancelOptions` is not about `filters`

Sc4ramouche: Thanks, moved it to query-cancellation guide.

PR #9601, opened 2025-08-30
fix(core): make sure queries revert synchronously

### LINKED ISSUE #9579: v5.85.2 introduced a breaking change while using StrictMode

TkDodo:
the switch from the callback approach to async/await with catch was necessary to be able to transform errors into revert-state data for imperative query calls; however, this was a change in behavior because catch in async/await doesn't happen immediately - it runs in the next microtask. That opened up an opportunity for a slight race condition if we re-start a fetch in-between. And who does that? Of course, React Strict Mode.

This PR moves the actual state reverting back to a callback (so it happens synchronously), while still keeping the try/catch refactoring merely to transform the promise and the usual error handling

fixes #9579

## Summary by CodeRabbit

- Refactor
  - Updated cancellation behavior: cancellations now invoke an onCancel callback with error details; revert (when requested) happens during onCancel, and canceled fetches no longer auto-reapply revert in the error path. Existing data is retained when available; otherwise, the cancellation surfaces.
  - Renamed the configuration option from abort to onCancel for retry/cancellation handling (breaking change for integrators).
- Tests
  - Added coverage for unsubscribe/resubscribe during an in-flight fetch to validate the new cancellation semantics.

### DIFF
--- a/packages/query-core/src/__tests__/query.test.tsx
+++ b/packages/query-core/src/__tests__/query.test.tsx
@@ -1192,4 +1192,54 @@ describe('query', () => {
     expect(initialDataFn).toHaveBeenCalledTimes(1)
     expect(query.state.data).toBe('initial data')
   })
+
+  test('should not override fetching state when revert happens after new observer subscribes', async () => {
+    const key = queryKey()
+    let count = 0
+
+    const queryFn = vi.fn(async ({ signal: _signal }) => {
+      // Destructure `signal` to intentionally consume it so observer-removal uses revert-cancel path
+      await sleep(50)
+      return 'data' + count++
+    })
+
+    const query = new Query({
+      client: queryClient,
+      queryKey: key,
+      queryHash: hashQueryKeyByOptions(key),
+      options: { queryFn },
+    })
+
+    const observer1 = new QueryObserver(queryClient, {
+      queryKey: key,
+      queryFn,
+    })
+
+    query.addObserver(observer1)
+    const promise1 = query.fetch()
+
+    await vi.advanceTimersByTimeAsync(10)
+
+    query.removeObserver(observer1)
+
+    const observer2 = new QueryObserver(queryClient, {
+      queryKey: key,
+      queryFn,
+    })
+
+    query.addObserver(observer2)
+
+    query.fetch()
+
+    await expect(promise1).rejects.toBeInstanceOf(CancelledError)
+    await vi.waitFor(() => expect(query.state.fetchStatus).toBe('idle'))
+
+    expect(queryFn).toHaveBeenCalledTimes(2)
+
+    expect(query.state).toMatchObject({
+      fetchStatus: 'idle',
+      status: 'success',
+      data: 'data1',
+    })
+  })
 })
--- a/packages/query-core/src/query.ts
+++ b/packages/query-core/src/query.ts
@@ -507,7 +507,15 @@ export class Query<
         | Promise<TData>
         | undefined,
       fn: context.fetchFn as () => Promise<TData>,
-      abort: abortController.abort.bind(abortController),
+      onCancel: (error) => {
+        if (error instanceof CancelledError && error.revert) {
+          this.setState({
+            ...this.#revertState,
+            fetchStatus: 'idle' as const,
+          })
+        }
+        abortController.abort()
+      },
       onFail: (failureCount, error) => {
         this.#dispatch({ type: 'failed', failureCount, error })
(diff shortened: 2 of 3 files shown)


joseph0926:
Hi @TkDodo , I was thinking about the synchronous revert approach and wondered if it might face a similar challenge to what you mentioned in my PR?
Your comment: https://github.com/TanStack/query/pull/9580#discussion_r2311893532

When onCancel executes synchronously during removeObserver, wouldn't the revert happen immediately before any new observer can be added? Here's a test case that might help illustrate this scenario

```typescript
test('should prevent revert when disabled observer is added after removal', async () => {
  const key = queryKey()

  const queryFn = vi.fn(async ({ signal: _signal }) => {
    await sleep(50)
    return 'data'
  })

  const query = new Query({
    client: queryClient,
    queryKey: key,
    queryHash: hashQueryKeyByOptions(key),
    options: { queryFn },
  })

  const observer1 = new QueryObserver(queryClient, {
    queryKey: key,
    queryFn,
  })

  query.addObserver(observer1)
  const promise1 = query.fetch()

  query.removeObserver(observer1)

  const observer2 = new QueryObserver(queryClient, {
    queryKey: key,
    queryFn,
    enabled: false,
  })

  query.addObserver(observer2)

  await expect(promise1).rejects.toBeInstanceOf(CancelledError)

  expect(query.state.fetchStatus).toBe('fetching')
  expect(query.state.data).toBeUndefined()
})
```

It seems like the synchronous execution might revert the state before the disabled observer gets added, which could lose the fetch state even though no new fetch was initiated.

Perhaps checking `this.observers.length > 0` instead of `this.isActive()` might address both concerns? Since a disabled observer still indicates the component is using the query (just not actively fetching), wouldn't it make sense to preserve the state in that case? I'm curious what you think about this approach - it might handle both the original StrictMode issue and the disabled observer scenario.

TkDodo:
> When onCancel executes synchronously during removeObserver, wouldn't the revert happen immediately before any new observer can be added?

yes, it would, but that is what we want. The flow is this:

- observer gets added, we fetch
- observer gets removed, we cancel and revert the query back into `fetchStatus: 'idle'`
- new observer gets mounted

now, the new observer can start from a “clean state” and will get us into `fetchStatus: 'fetching'` if the observer triggers a fetch, or it will keep us in idle state if its disabled.

before, we had this flow:

- observer gets added, we fetch
- observer gets removed, revert gets “scheduled” but isn’t done.
- new observer gets mounted, initiates fetch
- revert happens and sets us back to `fetchStatus: 'idle'`, which is not what we want.

---

the test case you showed is wrong because `query.addObserver` alone doesn’t trigger anything. To initiate a fetch, we need `observer.subscribe()` or `query.fetch`. So you can’t expect the query to be in `fetchStatus: 'fetching'` by just calling `query.addObserver`.

PR #9612, opened 2025-09-02
feat(query-core): add timeoutManager to allow changing setTimeout/setInterval

justjake:
Introduces new class `TimeoutManager` that is a facade over a "provider" implementation of `setTimeout`, `clearTimeout`, `setInterval`, `clearInterval`, following the pattern of `OnlineManager` and `FocusManager`.

All calls to global `setTimeout`/`setInterval` and their matching `clearTimeout`/`clearInterval` are replaced by new helper functions that use `TimeoutManager` as appropriate instead:

- `setTimeout(cb, number > 0)` -> `managedSetTimeout`
- `setTimeout(cb, 0)` -> `systemSetTimeoutZero` (to preserve the intention of running on the next system event loop tick)
- `setInterval(cb, number)` -> `managedSetInterval`

I opted to export functions instead of do direct calls on `timeoutManager.setTimeout` as this reduces bundle size impact in callers.

Let me know what you think.

### DIFF
--- a/packages/query-async-storage-persister/src/asyncThrottle.ts
+++ b/packages/query-async-storage-persister/src/asyncThrottle.ts
@@ -1,3 +1,4 @@
+import { managedSetTimeout } from '../../query-core/src/timeoutManager'
 import { noop } from './utils'
 
 interface AsyncThrottleOptions {
@@ -21,11 +22,11 @@ export function asyncThrottle<TArgs extends ReadonlyArray<unknown>>(
     if (isScheduled) return
     isScheduled = true
     while (isExecuting) {
-      await new Promise((done) => setTimeout(done, interval))
+      await new Promise((done) => managedSetTimeout(done, interval))
     }
     while (Date.now() < nextExecutionTime) {
       await new Promise((done) =>
-        setTimeout(done, nextExecutionTime - Date.now()),
+        managedSetTimeout(done, nextExecutionTime - Date.now()),
       )
     }
     isScheduled = false
--- a/packages/query-core/src/index.ts
+++ b/packages/query-core/src/index.ts
@@ -11,6 +11,15 @@ export { MutationCache } from './mutationCache'
 export type { MutationCacheNotifyEvent } from './mutationCache'
 export { MutationObserver } from './mutationObserver'
 export { notifyManager, defaultScheduler } from './notifyManager'
+export {
+  managedSetTimeout,
+  managedClearTimeout,
+  managedSetInterval,
+  managedClearInterval,
+  systemSetTimeoutZero,
+  defaultTimeoutProvider,
+  type ManagedTimerId,
+} from './timeoutManager'
 export { focusManager } from './focusManager'
 export { onlineManager } from './onlineManager'
 export {
--- a/packages/query-core/src/notifyManager.ts
+++ b/packages/query-core/src/notifyManager.ts
@@ -1,5 +1,7 @@
 // TYPES
 
+import { systemSetTimeoutZero } from './timeoutManager'
+
 type NotifyCallback = () => void
 
 type NotifyFunction = (callback: () => void) => void
@@ -10,7 +12,8 @@ type BatchCallsCallback<T extends Array<unknown>> = (...args: T) => void
 
 type ScheduleFunction = (callback: () => void) => void
 
-export const defaultScheduler: ScheduleFunction = (cb) => setTimeout(cb, 0)
+export const defaultScheduler: ScheduleFunction = (cb) =>
+  systemSetTimeoutZero(cb)
 
 export function createNotifyManager() {
   let queue: Array<NotifyCallback> = []
--- a/packages/query-core/src/timeoutManager.ts
+++ b/packages/query-core/src/timeoutManager.ts
@@ -8,16 +8,37 @@
  * Symbol.toPrimitive.
  */
 export type TimeoutProviderId = number | { [Symbol.toPrimitive]: () => number }
+export type TimeoutCallback = (_: void) => void
 
 export type TimeoutProvider = {
-  setTimeout: (callback: () => void, delay: number) => TimeoutProviderId
-  clearTimeout: (timeoutId: number | undefined) => void
+  /** Used in error messages. */
+  readonly name: string
+
+  readonly setTimeout: (
+    callback: TimeoutCallback,
+    delay: number,
+  ) => TimeoutProviderId
+  readonly clearTimeout: (timeoutId: number | undefined) => void
+
+  readonly setInterval: (
+    callback: TimeoutCallback,
+    delay: number,
+  ) => TimeoutProviderId
+  readonly clearInterval: (intervalId: number | undefined) => void
 }
 
-const defaultTimeoutProvider: TimeoutProvider = {
+export const defaultTimeoutProvider: TimeoutProvider = {
+  name: 'default',
+
   setTimeout: (callback, delay) => setTimeout(callback, delay),
   clearTimeout: (timeoutId) => clearTimeout(timeoutId),
+
+  setInterval: (callback, delay) => setInterval(callback, delay),
+  clearInterval: (intervalId) => clearInterval(intervalId),
 }
+
+/** Timeout ID returned by {@link TimeoutManager} */
+export type ManagedTimerId = number
 
 /**
  * Allows customization of how timeouts are created.
@@ -31,11 +52,13 @@ const defaultTimeoutProvider: TimeoutProvider = {
  * coalesces timeouts.
  */
 export class TimeoutManager implements TimeoutProvider {
+  public readonly name = 'TimeoutManager'
+
   #provider: TimeoutProvider = defaultTimeoutProvider
   #setTimeoutCalls = 0
 
   setTimeoutProvider(provider: TimeoutProvider): void {
-    if (this.#setTimeoutCalls > 0) {
+    if (this.#setTimeoutCalls > 0 && provider !== this.#provider) {
       // After changing providers, `clearTimeout` will not work as expected for
       // timeouts from the previous provider.
       //
@@ -48,21 +71,48 @@ export class TimeoutManager implements TimeoutProvider {
       // We could internally queue `setTimeout` calls to `TimeoutManager` until
       // some API call to set the initial provider.
       console.warn(
-        '[timeoutManager]: Provider changed after setTimeout calls were made. This might result in unexpected behavior.',
+        `[timeoutManager]: Switching to ${provider.name} provider after setTimeout calls were made with ${this.#provider.name} provider might result in unexpected behavior.`,
       )
     }
 
     this.#provider = provider
   }
 
-  setTimeout(callback: () => void, delay: number): number {
+  setTimeout(callback: TimeoutCallback, delay: number): ManagedTimerId {
     this.#setTimeoutCalls++
-    return Number(this.#provider.setTimeout(callback, delay))
+    return providerIdToNumber(
+      this.#provider,
+      this.#provider.setTimeout(callback, delay),
+    )
   }
 
-  clearTimeout(timeoutId: number | undefined): void {
+  clearTimeout(timeoutId: ManagedTimerId | undefined): void {
     this.#provider.clearTimeout(timeoutId)
   }
+
+  setInterval(callback: TimeoutCallback, delay: number): ManagedTimerId {
+    return providerIdToNumber(
+      this.#provider,
+      this.#provider.setInterval(callback, delay),
+    )
+  }
+
+  clearInterval(intervalId: ManagedTimerId | undefined): void {
+    this.#provider.clearInterval(intervalId)
+  }
+}
+
+function providerIdToNumber(
+  provider: TimeoutProvider,
+  providerId: TimeoutProviderId,
+): ManagedTimerId {
+  const numberId = Number(providerId)
+  if (isNaN(numberId)) {
+    throw new Error(
+      `TimeoutManager: could not convert ${provider.name} provider timeout ID to valid number`,
+    )
+  }
+  return numberId
 }
 
 export const timeoutManager = new TimeoutManager()
@@ -70,12 +120,43 @@ export const timeoutManager = new TimeoutManager()
 // Exporting functions that use `setTimeout` to reduce bundle size impact, since
 // method names on objects are usually not minified.
 
-/** A version of `setTimeout` that uses {@link timeoutManager} to set the timeout. */
-export function managedSetTimeout(callback: () => void, delay: number): number {
+/** A version of `setTimeout` controlled by {@link timeoutManager}. */
+export function managedSetTimeout(
+  callback: TimeoutCallback,
+  delay: number,
+): ManagedTimerId {
   return timeoutManager.setTimeout(callback, delay)
 }
 
-/** A version of `clearTimeout` that uses {@link timeoutManager} to set the timeout. */
-export function managedClearTimeout(timeoutId: number | undefined): void {
+/** A version of `clearTimeout` controlled by {@link timeoutManager}. */
+export function managedClearTimeout(
+  timeoutId: ManagedTimerId | undefined,
+): void {
   timeoutManager.clearTimeout(timeoutId)
+}
+
+/** A version of `setInterval` controlled by {@link timeoutManager}. */
+export function managedSetInterval(
+  callback: TimeoutCallback,
+  delay: number,
+): ManagedTimerId {
+  return timeoutManager.setInterval(callback, delay)
+}
+
+/** A version of `clearInterval` controlled by {@link timeoutManager}. */
+export function managedClearInterval(
+  intervalId: ManagedTimerId | undefined,
+): void {
+  timeoutManager.clearInterval(intervalId)
+}
+
+/**
+ * In many cases code wants to delay to the next event loop tick; this is not
+ * mediated by {@link timeoutManager}.
+ *
+ * This function is provided to make auditing the `tanstack/query-core` for
+ * incorrect use of system `setTimeout` easier.
+ */
+export function systemSetTimeoutZero(callback: TimeoutCallback): void {
+  setTimeout(callback, 0)
 }
--- a/packages/query-core/src/timeoutManager.ts
+++ b/packages/query-core/src/timeoutManager.ts
@@ -1,0 +1,81 @@
+/**
+ * Wrapping `setTimeout` is awkward from a typing perspective because platform
+ * typings may extend the return type of `setTimeout`. For example, NodeJS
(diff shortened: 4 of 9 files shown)


packages/query-async-storage-persister/src/asyncThrottle.ts
```diff
@@ -1,3 +1,4 @@
+import { managedSetTimeout } from '../../query-core/src/timeoutManager'
```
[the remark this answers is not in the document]
TkDodo: that’s correct, and we would need to add a dependency on:

```
"@tanstack/query-core": "workspace:*"
```

for this, which is okay.

packages/query-core/src/utils.ts
```diff
@@ -361,7 +362,9 @@ function hasObjectPrototype(o: any): boolean {
 
 export function sleep(timeout: number): Promise<void> {
   return new Promise((resolve) => {
-    setTimeout(resolve, timeout)
+    const setTimeoutFn =
+      timeout === 0 ? systemSetTimeoutZero : managedSetTimeout
```
TkDodo: why do we need to handle `setTimeout(0)` differently?

packages/query-core/src/notifyManager.ts
```diff
@@ -10,7 +12,8 @@ type BatchCallsCallback<T extends Array<unknown>> = (...args: T) => void
 
 type ScheduleFunction = (callback: () => void) => void
 
-export const defaultScheduler: ScheduleFunction = (cb) => setTimeout(cb, 0)
+export const defaultScheduler: ScheduleFunction = (cb) =>
```
TkDodo: note: we want to change this to `scheduleMicroTask` in the next major version, so maybe we should keep this outside of the managed timeouts to avoid larger changes when someone opts into their own managed timeout

packages/query-core/src/timeoutManager.ts
```diff
@@ -0,0 +1,174 @@
...
+      //
+      // We could protect against this by mixing the timeout ID bits
+      // deterministically with some per-provider bits.
+      //
+      // We could internally queue `setTimeout` calls to `TimeoutManager` until
+      // some API call to set the initial provider.
+      console.warn(
+        `[timeoutManager]: Switching to ${provider.name} provider after calls to ${this.#provider.name} provider might result in unexpected behavior.`,
+      )
+    }
```
TkDodo: let’s please put this behind an env check and do a `console.error` then, like e.g. here:

https://github.com/TanStack/query/blob/379670dd0a004df872da8494396d30eadcd71d19/packages/query-core/src/query.ts#L539-L543

packages/query-core/src/timeoutManager.ts
```diff
@@ -0,0 +1,174 @@
...
+
+  readonly setInterval: (
+    callback: TimeoutCallback,
+    delay: number,
+  ) => TimeoutProviderId
+  readonly clearInterval: (intervalId: number | undefined) => void
+}
+
+export const defaultTimeoutProvider: TimeoutProvider = {
+  name: 'default',
```
TkDodo: not sure if we need the name just for the error

packages/query-core/src/timeoutManager.ts
```diff
@@ -0,0 +1,174 @@
...
+
+  setTimeout: (callback, delay) => setTimeout(callback, delay),
```
TkDodo, line 44: could this be:

```
setTimeout: window.setTimeout
```

etc, or maybe even `setTimeout: setTimeout`, which is just `setTimeout` 😅 ?

packages/query-core/src/timeoutManager.ts
```diff
@@ -0,0 +1,174 @@
...
+  readonly clearTimeout: (timeoutId: number | undefined) => void
+
+  readonly setInterval: (
+    callback: TimeoutCallback,
+    delay: number,
+  ) => TimeoutProviderId
+  readonly clearInterval: (intervalId: number | undefined) => void
+}
+
+export const defaultTimeoutProvider: TimeoutProvider = {
```
TkDodo: do we need to `export` this?

packages/query-core/src/timeoutManager.ts
```diff
@@ -0,0 +1,174 @@
...
+function providerIdToNumber(
+  provider: TimeoutProvider,
+  providerId: TimeoutProviderId,
+): ManagedTimerId {
+  const numberId = Number(providerId)
+  if (isNaN(numberId)) {
+    throw new Error(
+      `TimeoutManager: could not convert ${provider.name} provider timeout ID to valid number`,
+    )
+  }
```
TkDodo: is this runtime coercion really necessary? I thought the problems with nodejs timings are on type-level only ? In that sense, could we not make `TimeoutManager` generic to accept e.g. `TimeoutManager<number>` or `TimeoutManager<NodeJS.Timeout>` ?

packages/query-core/src/timeoutManager.ts
```diff
@@ -0,0 +1,174 @@
...
+      this.#provider.setTimeout(callback, delay),
+    )
+  }
+
+  clearTimeout(timeoutId: ManagedTimerId | undefined): void {
+    this.#provider.clearTimeout(timeoutId)
+  }
+
+  setInterval(callback: TimeoutCallback, delay: number): ManagedTimerId {
+    this.#providerCalled = true
```
TkDodo: those assignments could also be put behind env checks if we move the console error behind the check, too.

TkDodo (commented):
bundle size impact looks okay:

before: 44.2kb (13.1kb gzip)
after: 45.5kb (13.6kb gzip)

I’m not sure if the extra functions are worth it, we should try out how this performs without them. Even if they aren’t minifyable on the call-sides, they usually gzip quite well.

packages/query-core/src/notifyManager.ts
justjake: That's why I have it calling `systemSetTimeoutZero` - no change in semantics, and the code clearly shows that it's not an oversight

packages/query-core/src/utils.ts
justjake: IDK, I was nervous about this for nebulous reasons

packages/query-core/src/timeoutManager.ts
justjake: will do 👍🏻. (i took the pattern from searching `console.warn` in the codebase, found one in QueryObserver that i copied)

packages/query-core/src/timeoutManager.ts
justjake: I saw that `notifyManager` exports `defaultScheduler` so I followed that pattern

packages/query-core/src/timeoutManager.ts
justjake: It's a bit awkward at the type level to parameterize `TimoutManager` because the default `TimoutManager` must be instantiated with a concrete type, which would be `TimeoutManager<ReturnType<typeof setTimeout>>` which would result in any provider passed to `setTimeoutProvider(provider: TimeoutProvider<T>)` having that concrete type and thus needing to accept and return NodeJS.Timeout in NodeJS.

justjake: I can forgo the runtime coercion and either cast here, or use `export type ManagedTimeoutId = number | { [Symbol.toPrimitive]: () => number }`

packages/query-core/src/timeoutManager.ts
justjake: This can lead to odd issues for callers in their tests, where they do something like `vi.spyOn(globalThis.setTimeout)`, which changes the global, but `defaultTimoutProvider` already has a hard reference to the previous `globalThis.setTimeout`. I'm not sure how compatible the hard reference is with something like `vi.useFakeTimers()` for example.

justjake: Also, doing this breaks our ability to `spyOn(globalThis, 'setTimout')` in our own tests.

packages/query-core/src/timeoutManager.ts
justjake: Does this comment mean "please remove this"?

packages/query-core/src/timeoutManager.ts
justjake: Done

packages/query-core/src/timeoutManager.ts
justjake: I'm against hard reference to `globalThis.setTimeout`

packages/query-core/src/timeoutManager.ts
justjake: Done

packages/query-core/src/utils.ts
justjake: Changed to managedSetTimeout

justjake:
@TkDodo If i was doing this in my codebase, I would add a lint rule that forbids use of `global/globalThis.setTimeout/setInterval` with an error message that suggests `managedSetTimeout/managedSetInterval`; is that something we should do here?

packages/query-core/src/timeoutManager.ts
TkDodo: yeah I think we don’t need this 👍

packages/query-core/src/timeoutManager.ts
TkDodo: oh, got it 👍

packages/query-core/src/notifyManager.ts
TkDodo: so you’re saying places where we now use `systemSetTimeoutZero` we should in the future use `scheduleMicroTask` ?

packages/query-core/src/timeoutManager.ts
TkDodo: I think we exported the `defaultScheduler` in `notifyManager` so that you can “switch back” to it. I think it’s fine to keep this private and maybe start to export it when we see a need. It’s easier to widen an API than to narrow it later.

packages/query-core/src/timeoutManager.ts
TkDodo: > or use export type ManagedTimeoutId = number | { [Symbol.toPrimitive]: () => number }

yeah let’s try that 👍

packages/query-core/src/notifyManager.ts
TkDodo: ah, I’m just now realizing that `systemSetTimeoutZero` doesn’t use the `timeoutManager` 😅 . I get it now 👍

packages/query-core/src/notifyManager.ts
```diff
@@ -10,7 +12,8 @@ type BatchCallsCallback<T extends Array<unknown>> = (...args: T) => void
...
+  systemSetTimeoutZero(cb)
```
TkDodo: that should work, right?

```
export const defaultScheduler: ScheduleFunction = systemSetTimeoutZero
```

packages/query-core/src/index.ts
```diff
@@ -11,6 +11,19 @@ export { MutationCache } from './mutationCache'
...
 export { MutationObserver } from './mutationObserver'
 export { notifyManager, defaultScheduler } from './notifyManager'
+export {
+  managedSetTimeout,
+  managedClearTimeout,
+  managedSetInterval,
+  managedClearInterval,
+  systemSetTimeoutZero,
+  defaultTimeoutProvider,
+  timeoutManager,
```
TkDodo: since the framework adapters re-export everything from the query-core, all of these now become part of the adapter APIs. I think it would be better if we just exported the `timeoutManager` and called the class methods directly, unless this has a large impact on bundle size.

justjake: 👍🏻

packages/query-core/src/notifyManager.ts
justjake: 👍🏻

justjake: Difference is that `defaultScheduler.name` will `"defaultScheduler"` -> `"systemSetTimeoutZero"` in dev, which perhaps is a tiny improvement but could potentially constitute a change in semantics visible to userspace code ¯\_(ツ)_/¯

packages/query-core/src/timeoutManager.ts
justjake: Done

packages/query-core/src/timeoutManager.ts
justjake: Done

packages/query-core/src/timeoutManager.ts
justjake: Done

justjake:
@TkDodo I noticed that `defaultTimeoutProvider` is useful in the internal TimeoutManager tests. I think if anyone wants to implement their own `TimeoutProvider`, perhaps they will want to switch between provider implementations and the default in their test cases too.

justjake:
@TkDodo can you approve CI workflow runs for my branch? I don't know if I'm passing CI

packages/query-core/src/notifyManager.ts
justjake: > so you’re saying places where we now use `systemSetTimeoutZero` we should in the future use `scheduleMicroTask` ?

I can't make that assessment - it's up to you to decide that should remain `globalThis.setTimeout(cb, 0)` and what should change to your new `scheduleMicroTask` function. 

But yeah, my intention is that changes from `setTimeout(cb, 0)` to `systemSetTimeoutZero` are syntactic only (to ease code review going forward), and remain semantically unchanged compared to before this PR.

packages/query-core/src/timeoutManager.ts
```diff
@@ -0,0 +1,139 @@
...
+ * typings may extend the return type of `setTimeout`. For example, NodeJS
+ * typings add `NodeJS.Timeout`; but a non-default `timeoutManager` may not be
+ * able to return such a type.
+ */
+export type ManagedTimerId = number | { [Symbol.toPrimitive]: () => number }
+
+/**
+ * Backend for timer functions.
+ */
+export type TimeoutProvider<TTimerId extends ManagedTimerId = ManagedTimerId> =
```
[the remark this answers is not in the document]
justjake: Read the comment

packages/query-core/src/timeoutManager.ts
```diff
@@ -0,0 +1,139 @@
...
+  // this class is unsafe internally to allow for extension.
+  #provider: TimeoutProvider<any> = defaultTimeoutProvider
+  #providerCalled = false
+
+  setTimeoutProvider<TTimerId extends ManagedTimerId>(
+    provider: TimeoutProvider<TTimerId>,
+  ): void {
+    if (provider === this.#provider) {
+      return
+    }
```
TkDodo: if the point of this early return is to just not have the console.error when called with the same provider, I think we could just move this check into the env check too:

```diff
 if (process.env.NODE_ENV !== 'production') {
-  if (this.#providerCalled) {
+  if (this.#providerCalled && provider !== this.#provider)
```

TkDodo:
I thought had to export `systemSetTimeoutZero` from `query-core` because the persister plugin uses it (841ac54)

but then it occurred to me that calling `setTimeout` there was wrong initially because we need to tie what’s happening there to our `scheduling` concept, and it only worked because the default scheduler was also using `setTimeout`.

so the correct fix (I hope) is to use `notifyManager.schedule` inside the persister plugin: 

fb22c67

TkDodo:
bundle size goes from 44.4 (13.2) to 44.9 (13.3), which is acceptable 👍

packages/query-core/src/timeoutManager.ts
TkDodo: did that here: https://github.com/TanStack/query/pull/9612/commits/09a787e2f82ac6e86f234f20f1c8a1427a7f1301

packages/query-core/src/__tests__/timeoutManager.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+      manager.clearTimeout(200)
+      expect(clearTimeoutSpy).toHaveBeenCalledWith(200)
+
+      const intervalId = manager.setInterval(callback, 300)
+      expect(setIntervalSpy).toHaveBeenCalledWith(callback, 300)
+      clearInterval(Number(intervalId))
+
+      manager.clearInterval(400)
+      expect(clearIntervalSpy).toHaveBeenCalledWith(400)
+    })
```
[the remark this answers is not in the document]
justjake: no, this has been legal in nodejs since 12

docs/reference/timeoutManager.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+## `timeoutManager.clearInterval`
+
+`clearInterval(intervalId)` can be used to cancel an interval, like the global [clearInterval function](https://developer.mozilla.org/en-US/docs/Web/API/Window/clearInterval). It should be called with an interval ID returned by `timeoutManager.setInterval`.
+
+```tsx
+import { timeoutManager } from '@tanstack/react-query'
+
+const intervalId = timeoutManager.setTimeout(
```
TkDodo:
```suggestion
const intervalId = timeoutManager.setInterval(
```

docs/reference/timeoutManager.md
```diff
@@ -0,0 +1,119 @@
...
+## `timeoutManager.setInterval`
+
+`setInterval(callback, intervalMs)` schedules a callback to be called approximately every `intervalMs`, like the global [setInterval function](https://developer.mozilla.org/en-US/docs/Web/API/Window/setInterval).
+
+Like `setTimeout`, it returns a timer ID, which may be a number or an object that can be coerced to a number via [Symbol.toPrimitive](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toPrimitive).
+
+```tsx
+import { timeoutManager } from '@tanstack/react-query'
+
+const intervalId = timeoutManager.setTimeout(
```
TkDodo:
```suggestion
const intervalId = timeoutManager.setInterval(
```

TkDodo (approved):

docs/reference/timeoutManager.md
```diff
@@ -0,0 +1,119 @@
...
+It is important to call `setTimeoutProvider` before creating a QueryClient or queries, so that the same provider is used consistently for all timers in the application, since different TimeoutProviders cannot cancel each others' timers.
+
+```tsx
+import { timeoutManager, QueryClient } from '@tanstack/react-query'
+import { CustomTimeoutProvider } from './CustomTimeoutProvider'
+
+timeoutManager.setTimeoutProvider(new CustomTimeoutProvider())
+
+export const queryClient = new QueryClient()
+```
```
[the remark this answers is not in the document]
TkDodo: @tanstack/react-query re-exports _everything_ from @tanstack/query-core

PR #9615, opened 2025-09-03
feat(query-core): add context to mutationfn

joshuaellis:
### Description

Kicking off where #9193 left off, i've added the same code modifications to add context to mutationFn, in addition i've changed type references & documentation references to the word `context` in reference to callbacks such as `onError | onSuccess | onSettled` to the word `scope` to avoid confusion – `TContext` also becomes `TScope`.

I've not changed the implementation of `MutationObserverResult` from `context` to `scope` because this would be a breaking change to the APIs available as far as I can tell – maybe this should be a follow up for v6 to create alignment across the codebase?

Finally, I was wondering if we should be passing the context to the aforementioned callbacks but then we do have the `scope` ability to for instance, you could pass the client along as seen in the changed examples...

Let me know if i've missed anything or misunderstood some of the agreements in the other PR.

### Related Issues

* closes #9193

### DIFF
--- a/docs/framework/angular/guides/mutations.md
+++ b/docs/framework/angular/guides/mutations.md
@@ -90,20 +90,20 @@ export class TodosComponent {
 ```ts
 mutation = injectMutation(() => ({
   mutationFn: addTodo,
-  onMutate: (variables) => {
+  onMutate: (variables, context) => {
     // A mutation is about to happen!
 
-    // Optionally return a context containing data to use when for example rolling back
+    // Optionally return a scope containing data to use when for example rolling back
     return { id: 1 }
   },
-  onError: (error, variables, context) => {
+  onError: (error, variables, scope) => {
     // An error happened!
-    console.log(`rolling back optimistic update with id ${context.id}`)
+    console.log(`rolling back optimistic update with id ${scope.id}`)
   },
-  onSuccess: (data, variables, context) => {
+  onSuccess: (data, variables, scope) => {
     // Boom baby!
   },
-  onSettled: (data, error, variables, context) => {
+  onSettled: (data, error, variables, scope) => {
     // Error or success... doesn't matter!
   },
 }))
@@ -130,25 +130,25 @@ mutation = injectMutation(() => ({
 ```ts
 mutation = injectMutation(() => ({
   mutationFn: addTodo,
-  onSuccess: (data, variables, context) => {
+  onSuccess: (data, variables, scope) => {
     // I will fire first
   },
-  onError: (error, variables, context) => {
+  onError: (error, variables, scope) => {
     // I will fire first
   },
-  onSettled: (data, error, variables, context) => {
+  onSettled: (data, error, variables, scope) => {
     // I will fire first
   },
 }))
 
 mutation.mutate(todo, {
-  onSuccess: (data, variables, context) => {
+  onSuccess: (data, variables, scope) => {
     // I will fire second!
   },
-  onError: (error, variables, context) => {
+  onError: (error, variables, scope) => {
     // I will fire second!
   },
-  onSettled: (data, error, variables, context) => {
+  onSettled: (data, error, variables, scope) => {
     // I will fire second!
   },
 })
@@ -161,7 +161,7 @@ mutation.mutate(todo, {
 export class Example {
   mutation = injectMutation(() => ({
     mutationFn: addTodo,
-    onSuccess: (data, variables, context) => {
+    onSuccess: (data, variables, scope) => {
       // Will be called 3 times
     },
   }))
@@ -169,7 +169,7 @@ export class Example {
   doMutations() {
     ;['Todo 1', 'Todo 2', 'Todo 3'].forEach((todo) => {
       this.mutation.mutate(todo, {
-        onSuccess: (data, variables, context) => {
+        onSuccess: (data, variables, scope) => {
           // Will execute only once, for the last mutation (Todo 3),
           // regardless which mutation resolves first
         },
@@ -214,31 +214,29 @@ const queryClient = new QueryClient()
 // Define the "addTodo" mutation
 queryClient.setMutationDefaults(['addTodo'], {
   mutationFn: addTodo,
-  onMutate: async (variables) => {
+  onMutate: async (variables, context) => {
     // Cancel current queries for the todos list
-    await queryClient.cancelQueries({ queryKey: ['todos'] })
+    await context.client.cancelQueries({ queryKey: ['todos'] })
 
     // Create optimistic todo
     const optimisticTodo = { id: uuid(), title: variables.title }
 
     // Add optimistic todo to todos list
-    queryClient.setQueryData(['todos'], (old) => [...old, optimisticTodo])
+    context.client.setQueryData(['todos'], (old) => [...old, optimisticTodo])
 
-    // Return context with the optimistic todo
-    return { optimisticTodo }
+    // Return scope with the optimistic todo
+    return { optimisticTodo, client: context.client }
   },
-  onSuccess: (result, variables, context) => {
+  onSuccess: (result, variables, scope) => {
     // Replace optimistic todo in the todos list with the result
-    queryClient.setQueryData(['todos'], (old) =>
-      old.map((todo) =>
-        todo.id === context.optimisticTodo.id ? result : todo,
-      ),
+    scope.client.setQueryData(['todos'], (old) =>
+      old.map((todo) => (todo.id === scope.optimisticTodo.id ? result : todo)),
     )
   },
-  onError: (error, variables, context) => {
+  onError: (error, variables, scope) => {
     // Remove optimistic todo from the todos list
-    queryClient.setQueryData(['todos'], (old) =>
-      old.filter((todo) => todo.id !== context.optimisticTodo.id),
+    scope.client.setQueryData(['todos'], (old) =>
+      old.filter((todo) => todo.id !== scope.optimisticTodo.id),
     )
   },
   retry: 3,
--- a/docs/framework/react/guides/mutations.md
+++ b/docs/framework/react/guides/mutations.md
@@ -143,20 +143,20 @@ const CreateTodo = () => {
 ```tsx
 useMutation({
   mutationFn: addTodo,
-  onMutate: (variables) => {
+  onMutate: (variables, context) => {
     // A mutation is about to happen!
 
-    // Optionally return a context containing data to use when for example rolling back
+    // Optionally return a scope containing data to use when for example rolling back
     return { id: 1 }
   },
-  onError: (error, variables, context) => {
+  onError: (error, variables, scope) => {
     // An error happened!
-    console.log(`rolling back optimistic update with id ${context.id}`)
+    console.log(`rolling back optimistic update with id ${scope.id}`)
   },
-  onSuccess: (data, variables, context) => {
+  onSuccess: (data, variables, scope) => {
     // Boom baby!
   },
-  onSettled: (data, error, variables, context) => {
+  onSettled: (data, error, variables, scope) => {
     // Error or success... doesn't matter!
   },
 })
@@ -189,25 +189,25 @@ You might find that you want to **trigger additional callbacks** beyond the ones
 ```tsx
 useMutation({
   mutationFn: addTodo,
-  onSuccess: (data, variables, context) => {
+  onSuccess: (data, variables, scope) => {
     // I will fire first
   },
-  onError: (error, variables, context) => {
+  onError: (error, variables, scope) => {
     // I will fire first
   },
-  onSettled: (data, error, variables, context) => {
+  onSettled: (data, error, variables, scope) => {
     // I will fire first
   },
 })
 
 mutate(todo, {
-  onSuccess: (data, variables, context) => {
+  onSuccess: (data, variables, scope) => {
     // I will fire second!
   },
-  onError: (error, variables, context) => {
+  onError: (error, variables, scope) => {
     // I will fire second!
   },
-  onSettled: (data, error, variables, context) => {
+  onSettled: (data, error, variables, scope) => {
     // I will fire second!
   },
 })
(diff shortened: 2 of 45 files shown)


packages/query-core/src/mutation.ts
```diff
@@ -200,7 +212,10 @@ export class Mutation<
           variables,
           this as Mutation<unknown, unknown, unknown, unknown>,
         )
-        const context = await this.options.onMutate?.(variables)
+        const context = await this.options.onMutate?.(
+          variables,
+          mutationFnContext,
+        )
```
TkDodo: the thing is, if we add this to `onMutate`, people will inevitably also want this for `onSuccess`, `onError` and `onSettled`

packages/query-core/src/types.ts
```diff
@@ -1091,36 +1091,44 @@ export type MutationMeta = Register extends {
...
     variables: TVariables,
-  ) => Promise<TContext | undefined> | TContext | undefined
+    context: MutationFunctionContext,
+  ) => Promise<TScope | undefined> | TScope | undefined
   onSuccess?: (
     data: TData,
     variables: TVariables,
-    context: TContext,
+    scope: TScope,
   ) => Promise<unknown> | unknown
```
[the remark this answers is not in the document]
TkDodo: I think this comment is correct. Not sure why `onSuccess` had `context` typed as non-optional, as the only way to fill it is to return something from `onMutate` ...

TkDodo (commented):
> I've not changed the implementation of MutationObserverResult from context to scope because this would be a breaking change 

we should probably deprecate `context` in `mutationState` and add a `scope` field and then remove `context` in v6

packages/query-core/src/mutation.ts
TkDodo: I think it makes sense to have them on those callbacks too, especially since with `mutationOptions`, you need the client for invalidation, but you otherwise need a hook to read it

joshuaellis: Great, i thought this API started to appear and make sense as I was working on it but didn't want to overstep the mark – I suppose realistically you'd have `context` appear before `scope` in the args list – but that would be a breaking change so maybe food for thought for v6 & we just have it added at the end?

We could make `scope` inherit context but that might be confusing & if people assert on scope not being defined then that would also break.

TkDodo: yeah it would have to be at the end. I don’t really like that many params but 🤷

joshuaellis:
Made all the amends @TkDodo so this is ready for another review when you have time.

packages/query-core/src/mutation.ts
```diff
@@ -3,30 +3,37 @@ import { Removable } from './removable'
...
   TError = DefaultError,
   TVariables = unknown,
-  TContext = unknown,
+  TScope = unknown,
 > {
-  context: TContext | undefined
+  /**
+   * @deprecated This will be removed in the next major version. Use the `scope` property instead.
+   */
+  context: TScope | undefined
```
TkDodo:
```suggestion
  /**
   * @deprecated `context` was renamed to `scope`. This will be removed in the next major version. Use the `scope` property instead.
   */
  context: TScope | undefined
```

docs/framework/react/guides/mutations.md
```diff
@@ -143,20 +143,20 @@ const CreateTodo = () => {
...
-  onMutate: (variables) => {
+  onMutate: (variables, context) => {
     // A mutation is about to happen!
 
-    // Optionally return a context containing data to use when for example rolling back
+    // Optionally return a scope containing data to use when for example rolling back
     return { id: 1 }
   },
-  onError: (error, variables, context) => {
+  onError: (error, variables, scope) => {
```
TkDodo: I think we should list `context` here as well, just for completeness


```suggestion
  onError: (error, variables, scope, context) => {
```

docs/framework/react/reference/useMutation.md
```diff
@@ -63,20 +64,20 @@ mutate(variables, {
...
   - defaults to `'online'`
   - see [Network Mode](../../guides/network-mode.md) for more information.
-- `onMutate: (variables: TVariables) => Promise<TContext | void> | TContext | void`
+- `onMutate: (variables: TVariables) => Promise<TScope | void> | TScope | void`
   - Optional
   - This function will fire before the mutation function is fired and is passed the same variables the mutation function would receive
   - Useful to perform optimistic updates to a resource in hopes that the mutation succeeds
   - The value returned from this function will be passed to both the `onError` and `onSettled` functions in the event of a mutation failure and can be useful for rolling back optimistic updates.
-- `onSuccess: (data: TData, variables: TVariables, context: TContext) => Promise<unknown> | unknown`
+- `onSuccess: (data: TData, variables: TVariables, scope: TScope) => Promise<unknown> | unknown`
```
TkDodo: we have `context: MutationFunctionContext` here at the end too, right?

joshuaellis:
@TkDodo this is ready for review, i've looked at the tests and got the CI green 👍🏼

TkDodo (commented):
oh noo, I forgot that we also have a MutationScope already 😭 

https://github.com/TanStack/query/blob/23c89080864c41161541c5bb3dfb21b4af8b0330/packages/query-core/src/types.ts#L1082-L1084

It can be passed as an argument to mutations. So, we need to find a different name again 😅

joshuaellis:
> It can be passed as an argument to mutations. So, we need to find a different name again 😅

We could append `localX` to something? alternatively it could be `state` since its scoped to the mutation?

joshuaellis:
Or `environment` maybe? but that's a common paradigm already used in development practice

TkDodo:
> > It can be passed as an argument to mutations. So, we need to find a different name again 😅
> 
> We could append `localX` to something? alternatively it could be `state` since its scoped to the mutation?

we also have a `MutationState` already. `mutateState` might work because it comes from “onMutate”, but it could still be confused with the `mutate` function. `environment` is too broad imo.

joshuaellis:
yeah `mutateState` could be confusing – would `localContext` work? its context, but local to the fn?

TkDodo:
> would localContext work?

yeah that’s not bad, I think `session` might also work. In the end, it’s just a positional argument in the list and the name of a type so it’s not that important, but still, getting this “right” would be nice :)

joshuaellis:
Do you think `session` is still too broad? maybe `sessionContext` would be more appropriate?

Also i agree, it's important especially because it's in the docs & helps developers learn 😄

joshuaellis:
alt. name `instance`?

TkDodo:
what do you think about just `onMutateResult` and `TOnMutateResult` for the type. I wouldn’t want this as a name for an object property, but for a positional argument, I think it’s fine as people can still name it however they want

joshuaellis:
I like it, explicit. Agreed, if it were an object it'd be a mouthful but positionally & for documentation wise its clear. I'll make that change.

joshuaellis:
Okay that's been done & i've rebased with latest updates – think i've caught everything...

docs/framework/angular/guides/mutations.md
```diff
@@ -90,20 +90,20 @@ export class TodosComponent {
...
 
-    // Optionally return a context containing data to use when for example rolling back
+    // Optionally return a result containing data to use when for example rolling back
     return { id: 1 }
   },
-  onError: (error, variables, context) => {
+  onError: (error, variables, onMutateResult, context) => {
     // An error happened!
-    console.log(`rolling back optimistic update with id ${context.id}`)
+    console.log(`rolling back optimistic update with id ${scope.id}`)
```
TkDodo:
```suggestion
    console.log(`rolling back optimistic update with id ${onMutateResult.id}`)
```

docs/framework/angular/guides/mutations.md
```diff
@@ -214,31 +214,31 @@ const queryClient = new QueryClient()
...
 
     // Create optimistic todo
     const optimisticTodo = { id: uuid(), title: variables.title }
 
     // Add optimistic todo to todos list
-    queryClient.setQueryData(['todos'], (old) => [...old, optimisticTodo])
+    context.client.setQueryData(['todos'], (old) => [...old, optimisticTodo])
 
-    // Return context with the optimistic todo
+    // Return scope with the optimistic todo
```
TkDodo:
```suggestion
    // Return result with the optimistic todo
```

docs/framework/react/guides/mutations.md
```diff
@@ -143,20 +143,20 @@ const CreateTodo = () => {
 ```tsx
 useMutation({
   mutationFn: addTodo,
-  onMutate: (variables) => {
+  onMutate: (variables, context) => {
     // A mutation is about to happen!
 
-    // Optionally return a context containing data to use when for example rolling back
+    // Optionally return a scope containing data to use when for example rolling back
```
TkDodo:
```suggestion
    // Optionally return a result containing data to use when for example rolling back
```

docs/framework/react/reference/useMutation.md
```diff
@@ -63,20 +64,20 @@ mutate(variables, {
   - Optional
   - defaults to `'online'`
   - see [Network Mode](../../guides/network-mode.md) for more information.
-- `onMutate: (variables: TVariables) => Promise<TContext | void> | TContext | void`
+- `onMutate: (variables: TVariables) => Promise<TScope | void> | TScope | void`
```
TkDodo:
```suggestion
- `onMutate: (variables: TVariables) => Promise<TOnMutateResult | void> | TOnMutateResult | void`
```

examples/react/optimistic-updates-cache/src/pages/index.tsx
```diff
@@ -45,18 +46,20 @@ function Example() {
       return await response.json()
     },
     // When mutate is called:
-    onMutate: async (newTodo: string) => {
+    onMutate: async (newTodo: string, context: MutationFunctionContext) => {
```
TkDodo: this should actually infer the types:


```suggestion
    onMutate: async (newTodo, context) => {
```

examples/react/optimistic-updates-cache/src/pages/index.tsx
```diff
@@ -9,6 +9,7 @@ import {
   useQueryClient,
 } from '@tanstack/react-query'
 import { ReactQueryDevtools } from '@tanstack/react-query-devtools'
+import type { MutationFunctionContext } from '@tanstack/react-query'
```
TkDodo: if inference works, we should delete this

packages/angular-query-experimental/src/inject-mutation.ts
```diff
@@ -46,16 +46,16 @@ export function injectMutation<
   TData = unknown,
   TError = DefaultError,
   TVariables = void,
-  TContext = unknown,
+  TScope = unknown,
```
TkDodo: `TScope` is outdated

packages/query-core/src/mutation.ts
```diff
@@ -3,30 +3,37 @@ import { Removable } from './removable'
...
   TVariables = unknown,
-  TContext = unknown,
+  TOnMutateResult = unknown,
 > {
-  context: TContext | undefined
+  /**
+   * @deprecated `context` was renamed to `scope`. This will be removed in the next major version. Use the `scope` property instead.
+   */
+  context: TOnMutateResult | undefined
+  scope: TOnMutateResult | undefined
```
joshuaellis: @TkDodo should we rename this variable then? or do you think this is fine, this is the only hangover of `scope` left afaik

joshuaellis:
polite bump @TkDodo – left a comment around `mutationState` and whether we should keep the name `scope`.

packages/query-core/src/mutation.ts
TkDodo: phew, okay, in this case, let’s leave it at `context` because I don’t think many people will be using it here and at least we don’t break anything.

joshuaellis: Done ✅

packages/query-core/src/__tests__/mutations.test.tsx
```diff
@@ -64,6 +68,7 @@ describe('mutations', () => {
 
     expect(mutation.getCurrentResult()).toEqual({
       context: undefined,
+      scope: undefined,
```
TkDodo: this needs to be reverted I guess?

packages/query-core/src/mutationCache.ts
```diff
@@ -14,25 +19,29 @@ interface MutationCacheConfig {
   onError?: (
     error: DefaultError,
     variables: unknown,
-    context: unknown,
+    onMutateResult: unknown,
     mutation: Mutation<unknown, unknown, unknown>,
+    mutationFnContext: MutationFunctionContext,
```
TkDodo: I think for a query, we call this:

```
context: QueryFunctionContext
```

so I think here it should be:

```
context: MutationFunctionContext
```

packages/query-core/src/mutationObserver.ts
```diff
@@ -138,25 +155,43 @@ export class MutationObserver<
...
   }
 
-  #notify(action?: Action<TData, TError, TVariables, TContext>): void {
+  #notify(action?: Action<TData, TError, TVariables, TOnMutateResult>): void {
     notifyManager.batch(() => {
       // First trigger the mutate callbacks
       if (this.#mutateOptions && this.hasListeners()) {
         const variables = this.#currentResult.variables!
-        const context = this.#currentResult.context
+        const scope = this.#currentResult.context
```
TkDodo: there’s still a `scope` here

joshuaellis: I'm gonna see the word scope in my nightmares 🤪

packages/query-core/src/mutationCache.ts
joshuaellis: thanks!

joshuaellis:
Caught those comments – thanks for your patience @TkDodo

joshuaellis:
Conflicts resolved @TkDodo 👍🏼

PR #9618, opened 2025-09-05
fix(query-core): fix combine function cache invalidation with stable reference

### LINKED ISSUE #8781: `combine` stable reference is not called with correct parameters

joseph0926:
Fixes #8781 
Closes #9207 : He had taken ownership of the fix, and his proposed direction was similar to mine. However, since there has been no response for several months and no tests seem to be included, I am opening an additional PR to help resolve the issue. If there is a rule that prevents someone else from submitting a PR once an issue has already been assigned, I will be happy to close this one.

## Description

`combine` function with stable reference not updating correctly when queries change
On the other hand, `combine` implemented as an inline function works correctly because the reference changes each time. 
> `combine !== this.#lastCombine`

## Problem

When using `useQueries` with a stable reference for the `combine` function, the combined result was not updated immediately when the queries array changed. This resulted in stale data being returned for one render cycle.

### Root Cause

The caching logic in `#combineResult` was comparing `this.#result` to determine cache invalidation, but the actual data being processed was the `input` parameter. When `getOptimisticResult` generated a new result array, `this.#result` hadn't been updated yet (happens later in `useEffect`), causing incorrect cache hits.

## Solution

Added `input.length !== this.#lastInput?.length` check to the cache invalidation logic. This ensures the cache is properly invalidated when the number of queries changes, which is the most common scenario.

### Why length comparison is sufficient

While comparing only the length might seem incomplete, it works because,
1. Most real-world cases involve changing the number of queries
2. For same-length changes, `setQueries` is called via `useEffect` which updates `this.#result` for the next render
3. This approach maintains the original performance optimization intent while fixing the bug

## Testing

- stable reference combine should update immediately when queries change
- inline combine should update immediately when queries change
- should handle dynamic query array changes correctly
- should handle same length but different queries
- should handle query order changes with same length
- should handle combine function that transforms data
- should not break when switching between stable and inline combine

All existing tests continue to pass, ensuring backward compatibility.

---

No breaking changes – this is a targeted optimization

### Alternative approaches considered

- Full input reference comparison: Would cause excessive cache misses
   - `input.length !== this.#lastInput`: This has been confirmed to significantly impact existing caching to the point of breaking existing caching tests. 
- Content-based comparison: Performance overhead
    - While it might be possible to achieve this by adding an additional flag to `#combineResult` and redesigning it, we determined that the overall impact would be too significant.
- Separate optimistic cache: Over-engineering for the use case
   - Perhaps something like this...? `#combineCache = new WeakMap<Array<QueryObserverResult>, TCombinedResult>()`
It might be possible, but I judged it to be overhead. 

## Summary by CodeRabbit

- Bug Fixes
  - useQueries combine now reliably recomputes when the number of queries changes, when query order changes, when switching between stable and inline combine implementations, and during optimistic-update scenarios so result length, counts, and data stay in sync.

- Tests
  - Added comprehensive tests for combine memoization, dynamic query arrays, transformed combine results, stability between combine implementations, and cache invalidation when query count changes.


joseph0926:
The test failure seems to be occurring in nx run @tanstack/solid-query-devtools:test:types, which is unrelated to this PR.

packages/query-core/src/queriesObserver.ts
```diff
@@ -216,10 +218,13 @@ export class QueriesObserver<
       if (
         !this.#combinedResult ||
         this.#result !== this.#lastResult ||
+        // Compare input.length to handle optimistic updates where input differs from this.#result
+        input.length !== this.#lastInput?.length ||
```
TkDodo, line 222: I don’t think a length check is what we want here. What about situations where the length is the same, but data inside it changed?

I’m honestly not sure why `input` would change, but the check for `this.#result !== this.#lastResult` would still see things as being equal. That’s what we need to find out imo

joseph0926: > I’m honestly not sure why `input` would change, but the check for `this.#result !== this.#lastResult` would still see things as being equal. That’s what we need to find out imo

When `getOptimisticResult` is called, a new input is created—but this happens at render time.
https://github.com/TanStack/query/blob/92d6b7bc86a45ed3d259d813e51a86b07fa9a7b3/packages/react-query/src/useQueries.ts#L261-L266

On the other hand, `observer.setQueries` is executed inside `useEffect`, which means it runs at commit time.

https://github.com/TanStack/query/blob/92d6b7bc86a45ed3d259d813e51a86b07fa9a7b3/packages/react-query/src/useQueries.ts#L281-L286

In other words, at the moment `getOptimisticResult` is called, the input already reflects the new value, but `this.#result` still holds the previous value.

Because of this timing difference, the following sequence occurs

First render

1. `this.#result = []`, `this.#lastResult = undefined` → `getOptimisticResult` is called → `input = []` is created
2. `#combineResult` is called → since `this.#lastResult = this.#result`, the result becomes `this.#result = []`, `this.#lastResult = []`
3. `useEffect() { setQueries ... }` runs

Second render
4. `getOptimisticResult` is called → `input = [query1]`, while `this.#result = []`, `this.#lastResult = []`
5. `#combineResult` is called → still `this.#result === this.#lastResult` (`[] === []`)
6. `useEffect() { setQueries ... }` runs → `this.#result = [query1]`

---

For this reason, a discrepancy arises between the `result` and the `input`
From a semantic perspective, it would indeed be more accurate to add a comparison such as,
`input !== this.#result && input !== this.#lastInput`

However, while this approach handles optimistic updates well, I believe it could also cause excessive cache misses in regular updates. That is why I chose the alternative solution of comparing lengths instead

Regarding your concern about edge cases—such as when the arrays have the same length but different order—I confirmed through the additional tests I wrote (`"should handle same length but different queries"` and `"should handle query order changes with same length"`) that no issues arise there. Based on that, I decided to adopt this solution as the final choice.

That said, as you pointed out, this method is ultimately more of a stopgap—it works well for optimistic updates while avoiding excessive cache misses in regular updates, but it remains a temporary measure.

---

Aside from this temporary workaround, I think a better solution would be to explicitly distinguish the case using a flag.

For example, something like this

```ts
#combineResult(
  input: Array<QueryObserverResult>,
  combine: CombineFn<TCombinedResult> | undefined,
  isOptimisticUpdate = false
): TCombinedResult {
  if (combine) {
    if (
      !this.#combinedResult ||
      this.#result !== this.#lastResult ||
      (isOptimisticUpdate && input !== this.#lastInput) ||
      combine !== this.#lastCombine
    ) {
      // ...
    }
  }
}
```

Perhaps an approach along these lines?

But come to think of it, this method would still cause additional cache misses,,

For example, the expected values for the three tests below seem likely to increase from 3 to 5.

`useQueries > should optimize combine if it is a stable reference`
`useQueries > should re-run combine if the functional reference changes`
`useQueries > should not re-run stable combine on unrelated re-render`

So my final thought is that comparing lengths seems like the best choice when not considering overall mechanism modifications. What do you think?

PR #9623, opened 2025-09-06
[v4] fix(ci): resolve v4 CI timeout by disabling Nx Cloud distributed execution

joseph0926:
CI issues related to Nx Cloud: #9584 #9334 #9310 

### Problem
The CI pipeline has been timing out after 1 hour with the error
```
NX   No new completed tasks after 3600 seconds.
```

### Maybe Root Cause
In commit eee4405f7f76996831a665d9a5ae51057f0c3ab1, the workflow was refactored with the following changes
- Removed the `agents` job that was starting 3 distributed agents
- Added `tanstack/config/.github/setup@main` for simplified setup
- Added `--parallel=3` flag to the test command

However, the environment variables still had
- `NX_CLOUD_DISTRIBUTED_EXECUTION: true` 
- `NX_CLOUD_DISTRIBUTED_EXECUTION_AGENT_COUNT: 3`

This created a configuration mismatch where
1. The main job starts the CI orchestrator with `npx nx-cloud start-ci-run`
2. The orchestrator waits for 3 agents to connect (as configured)
3. No agents actually start (since the `agents` job was removed)
4. After 1 hour of waiting, the job times out

### Solution
Set `NX_CLOUD_DISTRIBUTED_EXECUTION: false` to align with the simplified workflow structure. This ensures tests run directly in the main job.

### Note
While we cannot fully replicate the GitHub Actions environment locally, this analysis is based on
- The observed CI behavior and error messages
- Local testing confirming tests pass without distributed execution
- The clear mismatch between the workflow structure and configuration


joseph0926:
@manudeli 
I would appreciate it if you could review these changes when you have time.

---

I'm not entirely sure if this is the perfect solution, but anyway, this PR passes the NX Cloud stage with the modified pr.yml,,

manudeli (approved):
Let's try this after merging

manudeli:
@joseph0926 I think you fixed it! 👍
image

joseph0926:
> @joseph0926 I think you fixed it! 👍 

@manudeli 
That's a relief.
Thanks for testing it!

PR #9631, opened 2025-09-08
test(query-core/focusManager): remove redundant 'toHaveBeenCalledTimes' from listener test assertions

sukvvon:
## Summary by CodeRabbit

* **Tests**
  * Relaxed assertions in focus management tests to verify only the first three listener invocations (true, false, true)
  * No changes to runtime behavior or public APIs; end-user functionality remains unaffected.


TkDodo:
do you have an example where this failed in CI? I don’t think we should allow additional listener invocations 🤔

sukvvon:
> do you have an example where this failed in CI? I don’t think we should allow additional listener invocations 🤔

@TkDodo 

Could you explain more detail?

TkDodo:
@sukvvon in the PR description, you said:

> This reduces test flakiness and improves CI stability.

So I wanted to know if this test has ever been flaky in CI before? Because I don’t think that we should be getting additional calls to the subscribe callback

sukvvon:
@TkDodo 

I apologize for the confusion in the PR description. That text was auto-generated by CodeRabbitAI and I didn't review it thoroughly before submitting.

To clarify: this test has never been flaky in CI before, and our CI has been stable.

TkDodo:
Then we don't need this change ?

PR #9633, opened 2025-09-09
fix(query-core,react-query): preserve infinite query type through failed SSR hydration

### LINKED ISSUE #8825: Next.js: exception encountered in prefetchInfiniteQuery causes useSuspenseInfiniteQuery to fail
### Describe the bug

I believe there is some weird behaviour when a `prefetchInfiniteQuery` fails in a server component and `useSuspenseInfiniteQuery` retries to fetch the data on the client. The exception thrown is: `Uncaught TypeError: Cannot read properties of undefined (reading 'length')`. I have spent some time debugging into this, it feels like after the failure on the server, the client retries to fetch the data and upon a successful response, it doesn't modify the response to be in the expected `{pages: [], pageParams:[]}` format, but treats it like a normal query response.

I have also noticed that query options passed are ignored (such as retry: 0).

https://codesandbox.io/p/devbox/9lyl7g

### Steps to reproduce

The sandbox has an API endpoint that is setup to fail randomly. Verify that the hardcoded endpoint path is correct in `pokemon.ts`.

1. Restart sandbox browser until prefetch has failed
2. Observe an exception being thrown when a retry to /api request succeeds

### Expected behavior

I expect it to work fine.

None

### Screenshots or Videos

_No response_

### Platform

- OS: MacOS
- Chrome
- Dependencies:

"next": "14.2.15",
"react": "18.3.1",
"react-dom": "18.3.1"

### Tanstack Query adapter

None

### TanStack Query version

5.62.7

### TypeScript version

_No response_

### Additional context

_No response_

joseph0926:
Fixes #8825 

### Problem
When prefetchInfiniteQuery fails on the server and useSuspenseInfiniteQuery later retries on the client, the response is treated as a regular query rather than the expected infinite structure ({ pages: [], pageParams: [] }). As a result, accessors like data.pages can throw runtime errors (for example, “Cannot read properties of undefined (reading 'length')”).

### Maybe Root cause
During SSR, a failed infinite query loses its infinite-query type information at dehydration time. The hydrated instance therefore lacks the behavior that marks it as infinite. When a retry occurs on the client, the retryer can remain pending, and even after InfiniteQueryObserver attaches the proper behavior, the data is not transformed into the infinite shape. Together, these conditions cause the client recovery path to return a regular query payload instead of { pages, pageParams }.

### Solution
The fix preserves infinite-query type information across the hydration boundary and ensures the retryer can progress for hydrated infinite queries. Concretely, a new isInfiniteQuery flag is added to the dehydrate/hydrate flow so the query is recognized as infinite after hydration. In addition, when a hydrated infinite query is found with a pending retryer, the fetch logic now allows cancellation and re-start when the correct behavior is attached—even if no prior data exists—so the subsequent retry yields the properly transformed infinite structure.

### Changes
In query-core, DehydratedQuery now includes an optional isInfiniteQuery flag, which dehydrateQuery sets for infinite queries and hydrate carries onto the Query instance. The Query class maintains this via a private #isInfiniteQuery field with corresponding accessors. The fetch path has been updated so hydrated infinite queries with a stuck retryer can transition cleanly once behavior is applied. In react-query, comprehensive tests exercise SSR hydration and recovery paths to verify that the infinite data shape is preserved across the boundary.

### Tests
The new suite covers failed SSR hydration with subsequent client recovery, mixtures of infinite and regular queries, race conditions between hydration and observer attachment, and structural validation of { pages, pageParams }, including edge cases such as nested keys and multiple concurrent queries. All existing tests pass unchanged.

### Example

```tsx
// Server - fails
await queryClient.prefetchInfiniteQuery({
  queryKey: ['posts'],
  queryFn: () => throw new Error('Network error'),
  initialPageParam: 0,
  getNextPageParam: () => 1,
})

// Client - recovers with correct structure
const { data } = useSuspenseInfiniteQuery({
  queryKey: ['posts'],
  queryFn: ({ pageParam }) => fetchPosts(pageParam),
  initialPageParam: 0,
  getNextPageParam: (lastPage) => lastPage.nextCursor,
})

// data is now correctly typed as { pages: [...], pageParams: [...] }
```

#### Notes
The change is backward-compatible and adds only negligible overhead, because the new flag is relevant solely for hydrated queries under the described conditions.

### DIFF
--- a/packages/query-core/src/__tests__/hydration.test.tsx
+++ b/packages/query-core/src/__tests__/hydration.test.tsx
@@ -1399,4 +1399,207 @@ describe('dehydration and rehydration', () => {
     // error and test will fail
     await originalPromise
   })
+
+  test('should preserve infinite query type when hydrating failed promise', async () => {
+    const queryClient = new QueryClient({
+      defaultOptions: {
+        dehydrate: {
+          shouldDehydrateQuery: () => true,
+        },
+      },
+    })
+
+    const promise = queryClient
+      .prefetchInfiniteQuery({
+        queryKey: ['infinite', 'failed'],
+        queryFn: () => Promise.reject(new Error('fetch failed')),
+        initialPageParam: 0,
+        getNextPageParam: () => 1,
+        retry: false,
+      })
+      .catch(() => {})
+
+    const dehydrated = dehydrate(queryClient)
+
+    const hydrationClient = new QueryClient()
+    hydrate(hydrationClient, dehydrated)
+
+    const hydratedQuery = hydrationClient.getQueryCache().find({
+      queryKey: ['infinite', 'failed'],
+    })
+
+    expect(hydratedQuery?.isInfiniteQuery).toBe(true)
+
+    await promise
+  })
+
+  test('should mark infinite queries with isInfiniteQuery flag during dehydration', async () => {
+    const queryClient = new QueryClient({
+      defaultOptions: {
+        dehydrate: { shouldDehydrateQuery: () => true },
+      },
+    })
+
+    await queryClient.prefetchInfiniteQuery({
+      queryKey: ['infinite'],
+      queryFn: ({ pageParam = 0 }) => Promise.resolve(`page-${pageParam}`),
+      initialPageParam: 0,
+      getNextPageParam: (_lastPage: any, pages: any) => pages.length,
+      retry: false,
+    })
+
+    await queryClient.prefetchQuery({
+      queryKey: ['regular'],
+      queryFn: () => Promise.resolve('data'),
+    })
+
+    const dehydrated = dehydrate(queryClient)
+
+    const infiniteQuery = dehydrated.queries.find(
+      (q) => q.queryKey[0] === 'infinite',
+    )
+    expect(infiniteQuery?.isInfiniteQuery).toBe(true)
+
+    const regularQuery = dehydrated.queries.find(
+      (q) => q.queryKey[0] === 'regular',
+    )
+    expect(regularQuery?.isInfiniteQuery).toBeUndefined()
+  })
+
+  test('should preserve isInfiniteQuery flag through hydration', async () => {
+    const queryClient = new QueryClient({
+      defaultOptions: {
+        dehydrate: { shouldDehydrateQuery: () => true },
+      },
+    })
+
+    await queryClient
+      .prefetchInfiniteQuery({
+        queryKey: ['infinite'],
+        queryFn: () => Promise.reject(new Error('Failed')),
+        initialPageParam: 0,
+        getNextPageParam: () => 1,
+        retry: false,
+      })
+      .catch(() => {})
+
+    const dehydrated = dehydrate(queryClient)
+
+    expect(dehydrated.queries[0]?.isInfiniteQuery).toBe(true)
+
+    const newClient = new QueryClient()
+    hydrate(newClient, dehydrated)
+
+    const hydratedQuery = newClient.getQueryCache().find({
+      queryKey: ['infinite'],
+    })
+
+    expect(hydratedQuery?.isInfiniteQuery).toBe(true)
+  })
+
+  test('should handle JSON serialization of dehydrated infinite queries', async () => {
+    const queryClient = new QueryClient({
+      defaultOptions: {
+        dehydrate: { shouldDehydrateQuery: () => true },
+      },
+    })
+
+    await queryClient
+      .prefetchInfiniteQuery({
+        queryKey: ['infinite'],
+        queryFn: () => Promise.reject(new Error('Failed')),
+        initialPageParam: 0,
+        getNextPageParam: () => 1,
+        retry: false,
+      })
+      .catch(() => {})
+
+    const dehydrated = dehydrate(queryClient)
+
+    const serialized = JSON.stringify(dehydrated)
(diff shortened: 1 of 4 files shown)


TkDodo:
serializing the information about “infinite” query seems fine, though I think we probably shouldn’t do this with a flag, but maybe with a `type` like field that would allow more than just two behvaviors.

what I don’t understand is why it isn’t enough to just set `options.behavior` to `infiniteQueryBehavior` before calling `query.fetch` in the hydration, as I’ve outlined here:
- https://github.com/TanStack/query/issues/8825#issuecomment-2733549348

this _should_ be enough, and should be the solution we strive towards.

joseph0926:
> serializing the information about “infinite” query seems fine, though I think we probably shouldn’t do this with a flag, but maybe with a `type` like field that would allow more than just two behvaviors.
> 
> what I don’t understand is why it isn’t enough to just set `options.behavior` to `infiniteQueryBehavior` before calling `query.fetch` in the hydration, as I’ve outlined here:
> - https://github.com/TanStack/query/issues/8825#issuecomment-2733549348
> 
> this _should_ be enough, and should be the solution we strive towards.

You're absolutely right.
I'll follow the design you suggested and apply the solution appropriately now.
Thanks for the feedback.

TkDodo:
> You're absolutely right.
I'll follow the design you suggested and apply the solution appropriately now.

Sorry this sounds absolutely like something that claude code would say. I have nothing against using AI as an assistant but please let me know if everything you do is just coming straight from AI and this account is just a proxy for it.

joseph0926:
> > You're absolutely right.
> > I'll follow the design you suggested and apply the solution appropriately now.
> 
> Sorry this sounds absolutely like something that claude code would say. I have nothing against using AI as an assistant but please let me know if everything you do is just coming straight from AI and this account is just a proxy for it.

That's how Deepl translates.
Try translating “당신의 말이 맞습니다” into English.
Of course, AI like GPT is used, but calling it an “if everything you do is just coming straight from AI and this account is just a proxy for it” seems a bit of an overstatement.

joseph0926:
I’m not a native English speaker,
so I sometimes use translation and assistive tools.
Hearing that “everything I do is coming straight from AI” and that this account is “just a proxy for it”—especially based on my use of the phrase “You’re absolutely right”—was disappointing,
Honestly, if I'd been asked questions like "Do you code with AI?"
I would have admitted yes or no, or explained something,
but when I hear "Are you AI?" I can't say anything at all...
but I’m sorry if my wording made you uncomfortable.
I sincerely hope the project continues to improve and that remaining bugs are addressed thoroughly, whether with the help of AI tools or otherwise.
Since it seems further contributions from me may not be welcome, I’ll refrain from submitting more for now, but I’ll continue to use and support the project as a user.

But I learned a lot from you during my time contributing, and it was truly enjoyable.
Thank you.

joseph0926:
@TkDodo 
I respect your concern. and once again, I'm sorry if this caused you any discomfort.
However, no matter how much I think about it, I simply cannot figure it out, so I respectfully ask a question.
From my perspective, it's not clear how that single sentence led to the conclusion of 'proxy'.
Could you point out the exact phrasing that made you uncomfortable? I'm genuinely asking so I can make corrections, whether in this PR or elsewhere.
I apologize for taking up your time when you must be busy.
However, from my perspective, that expression was shocking, so I'm asking about it.

TkDodo:
Let me start by saying that I’m very sorry for how my comment came across. I too am not a native speaker, so I understand the troubles. Re-reading my comment now it’s clear to me that it was inappropriate.

> Honestly, if I'd been asked questions like "Do you code with AI?"

Yes, you’re right, that would’ve been the better question, and I should’ve phrased it that way.

What I found odd in this interaction is that I tried to outline in the linked issue what I think the problem is and what a potential solution could look like. Then, this PR shows up that implements a different approach, which I think would fail in the normal case of a successfully streamed infinite query because it is absolutely necessary that we attach the infinite query behavior in all cases. So when I point out pretty much the exact same thing that I said on the issue again in the PR, I’m met with “you’re absolutely right”. This is typical for AI interactions: One says something, AI does something else, you say the exact same thing again, and get an “you’re absolutely right, thanks for pointing that out. I will fix it that way” response.

That gave me the feeling that the whole conversation is being held by AI, and that I might be wasting my time here. However, I shouldn’t have written that unreflected comment, and again, I’m sorry for that.

joseph0926:
@TkDodo 
Thank you for your thoughtful reply. After reading your additional explanation, I understand the source of the misunderstanding and your concerns.

To explain the reasons behind your concerns, I did read your comment on the issue. However, I became overly focused on this specific part
> “the conceptual problem I’m having now is that hydration doesn’t know about infinite / non-infinite query. There’s no information stored in the cache that knows about this.”

Because I fixated on that line, I framed the problem as “if we just solve this piece, the whole thing will work,” and I proceeded from that assumption.

Approaching it that way led me down a different path from your proposed solution. After receiving your feedback and re-reading your comment carefully, I realized that my direction was aimed at addressing a specific symptom, while your approach both resolves that symptom and tackles the root cause.

That is why I said I would make changes in response to your feedback.

Having read your detailed explanation, I fully understand the concern and how the misunderstanding arose.

Thank you again for taking the time to explain everything so clearly. I was a bit emotional as well, I sincerely apologize for that.

Thanks again.

PR #9635, opened 2025-09-09
fix(query-core): fix combine function cache invalidation with stable reference

### LINKED ISSUE #8781: `combine` stable reference is not called with correct parameters

joseph0926:
Fixes #8781 
Related PR: #9618 

## Description

`combine` function with stable reference not updating correctly when queries change
On the other hand, `combine` implemented as an inline function works correctly because the reference changes each time. 
> `combine !== this.#lastCombine`

## Problem

When using `useQueries` with a stable reference for the `combine` function, the combined result was not updated immediately when the queries array changed. This resulted in stale data being returned for one render cycle.

### Root Cause

The caching logic in `#combineResult` was comparing `this.#result` to determine cache invalidation, but the actual data being processed was the `input` parameter. When `getOptimisticResult` generated a new result array, `this.#result` hadn't been updated yet (happens later in `useEffect`), causing incorrect cache hits.

## Solution

Added `input.length !== this.#lastInput?.length` check to the cache invalidation logic. This ensures the cache is properly invalidated when the number of queries changes, which is the most common scenario.

### Why length comparison is sufficient

While comparing only the length might seem incomplete, it works because,
1. Most real-world cases involve changing the number of queries
2. For same-length changes, `setQueries` is called via `useEffect` which updates `this.#result` for the next render
3. This approach maintains the original performance optimization intent while fixing the bug

## Testing

- stable reference combine should update immediately when queries change
- inline combine should update immediately when queries change
- should handle dynamic query array changes correctly
- should handle same length but different queries
- should handle query order changes with same length
- should handle combine function that transforms data
- should not break when switching between stable and inline combine

All existing tests continue to pass, ensuring backward compatibility.

---

No breaking changes – this is a targeted optimization

### Alternative approaches considered

- Full input reference comparison: Would cause excessive cache misses
   - `input.length !== this.#lastInput`: This has been confirmed to significantly impact existing caching to the point of breaking existing caching tests. 
- Content-based comparison: Performance overhead
    - While it might be possible to achieve this by adding an additional flag to `#combineResult` and redesigning it, we determined that the overall impact would be too significant.
- Separate optimistic cache: Over-engineering for the use case
   - Perhaps something like this...? `#combineCache = new WeakMap<Array<QueryObserverResult>, TCombinedResult>()`
It might be possible, but I judged it to be overhead. 

### Notes
This is regarding the feedback from the previous [PR](https://github.com/TanStack/query/pull/9618).
https://github.com/TanStack/query/pull/9618#discussion_r2325272070
1. Comparing lengths doesn't seem to be the main solution.
2. The crucial point is why changes to `input` aren't detected in `this.#result !== this.#lastResult`. 

My response to this 
https://github.com/TanStack/query/pull/9618#discussion_r2325366131

1. That's correct. Length comparison is a second-best solution to resolve the issue while preventing existing tests from failing.
2. The reason it wasn't caught by that conditional statement was due to timing differences (details are provided in the link).
So my conclusion was that the length comparison is a reasonable compromise—it resolves the issue and detects input changes while minimizing impact on existing functionality (the cache). I ended the discussion by asking if my reasoning was correct. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved combine behavior for multi-query results to update immediately when the number of queries changes, including during optimistic updates.
  * More reliable handling of dynamic query arrays, including additions, removals, and reorderings, ensuring combined results reflect the current set and order.

* **Tests**
  * Added comprehensive test coverage for useQueries with combine, covering stable and inline functions, changing query counts, reordering, and data alignment scenarios.

### DIFF
--- a/examples/angular/auto-refetching/package.json
+++ b/examples/angular/auto-refetching/package.json
@@ -13,7 +13,7 @@
     "@angular/compiler": "^20.0.0",
     "@angular/core": "^20.0.0",
     "@angular/platform-browser": "^20.0.0",
-    "@tanstack/angular-query-experimental": "^5.87.1",
+    "@tanstack/angular-query-experimental": "^5.87.3",
     "rxjs": "^7.8.2",
     "tslib": "^2.8.1",
     "zone.js": "0.15.0"
--- a/examples/angular/basic-persister/package.json
+++ b/examples/angular/basic-persister/package.json
@@ -13,7 +13,7 @@
     "@angular/compiler": "^20.0.0",
     "@angular/core": "^20.0.0",
     "@angular/platform-browser": "^20.0.0",
-    "@tanstack/angular-query-experimental": "^5.87.1",
+    "@tanstack/angular-query-experimental": "^5.87.3",
     "@tanstack/angular-query-persist-client": "^5.62.7",
     "@tanstack/query-async-storage-persister": "^5.87.1",
     "rxjs": "^7.8.2",
--- a/examples/angular/basic/package.json
+++ b/examples/angular/basic/package.json
@@ -13,7 +13,7 @@
     "@angular/compiler": "^20.0.0",
     "@angular/core": "^20.0.0",
     "@angular/platform-browser": "^20.0.0",
-    "@tanstack/angular-query-experimental": "^5.87.1",
+    "@tanstack/angular-query-experimental": "^5.87.3",
     "rxjs": "^7.8.2",
     "tslib": "^2.8.1",
     "zone.js": "0.15.0"
--- a/examples/angular/devtools-panel/package.json
+++ b/examples/angular/devtools-panel/package.json
@@ -14,8 +14,8 @@
     "@angular/core": "^20.0.0",
     "@angular/platform-browser": "^20.0.0",
     "@angular/router": "^20.0.0",
-    "@tanstack/angular-query-devtools-experimental": "^5.87.1",
-    "@tanstack/angular-query-experimental": "^5.87.1",
+    "@tanstack/angular-query-devtools-experimental": "^5.87.3",
+    "@tanstack/angular-query-experimental": "^5.87.3",
     "rxjs": "^7.8.2",
     "tslib": "^2.8.1",
     "zone.js": "0.15.0"
--- a/examples/angular/infinite-query-with-max-pages/package.json
+++ b/examples/angular/infinite-query-with-max-pages/package.json
@@ -13,7 +13,7 @@
     "@angular/compiler": "^20.0.0",
     "@angular/core": "^20.0.0",
     "@angular/platform-browser": "^20.0.0",
-    "@tanstack/angular-query-experimental": "^5.87.1",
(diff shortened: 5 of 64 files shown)


joseph0926:
I'm not quite sure how to respond to coderabbitai's nitpicky comments, so I'm leaving a separate comment instead. 

> Store only the input length (or compare to current result length) to reduce state and coupling

First off, Option A seems fundamentally flawed in its approach.
The reason we save it as `#inputResult` isn't just to store a number, but to track changes.
Therefore, I think this is incorrect feedback...
Option B could be in the same vein, but I tried `input.length !== this.#result.length` and confirmed it passes all tests.
However, the point at which `this.#result` is determined is `setQueries`, while `input` is determined at `getOptimisticResult`. This difference in timing makes me feel like we're back to the root cause of the issue.
But I'm not entirely sure about this. Regardless, it passes all tests as is.

> Consider adding the symmetric shrink case

I could add it if needed, but I'm not sure if this test is truly necessary... I believe the tests already added cover most edge cases, but I'm not entirely sure... Adding it is literally just copy-pasting the code provided by coderabbitai, so it's easy, but I'm not confident it's a meaningful test.
For now, I added it and ran it locally, and it passes. 

> Optional: assert combine invocation to guard perf | Optional: assert combine invocation to guard perf

It seems optional, so I'll do it if necessary.

joseph0926:
Length comparison is not a fundamental solution
I will close this PR.

PR #9640, opened 2025-09-10
README Updates

LadyBluenotes:
Summary:
-   Updated the README to improve feature overview and descriptions.
-  Refined wording for clarity and consistency across product blurbs and feature lists.
- Updated header images so it's consistent with other libraries

**Note:** The image is currently a relative link and I see the previous ones aren't so that might need to be updated.


README.md
```diff
@@ -1,52 +1,59 @@
...
-  <img alt="" src="https://img.shields.io/github/stars/TanStack/query.svg?style=social&label=Star" />
-</a><a href="https://twitter.com/tannerlinsley" target="\_parent">
-  <img alt="" src="https://img.shields.io/twitter/follow/tannerlinsley.svg?style=social&label=Follow" />
-</a> <a href="https://gitpod.io/from-referrer/">
-  <img src="https://img.shields.io/badge/Gitpod-Ready--to--Code-blue?logo=gitpod" alt="Gitpod Ready-to-Code"/>
+<div align="center">
+  <img src="./media/header_query.png" >
+</div>
+
+A data‑fetching library built to simplify fetching, caching, synchronizing, and updating server state.
```
TkDodo: I think we shouldn’t call TanStack Query a data-fetching library - I’ve previously always pointed out that it isn’t a data-fetching library, because it isn’t tied to data fetching.

I’ve always described it as an “async state manager” that works well with server state and data-fetching. On tanstack.com, it’s described as:

> Powerful asynchronous state management, server-state utilities and data fetching. Fetch, cache, update, and wrangle all forms of async data in your TS/JS, React, Vue, Solid, Svelte & Angular applications all without touching any "global state"

LadyBluenotes: Would this be on par with what you're thinking? just switching the "data-fetching" part out

```suggestion
An async state management library built to simplify fetching, caching, synchronizing, and updating server state.
```

TkDodo: Yep 👍

PR #9644, opened 2025-09-11
Fix: Solid query `useIsMutating` throwing an error and `useMutationState` infinite loop

dev-rb:
- Fixes an infinite loop in `useMutationState` caused by reading and writing to `results` in an effect.
- Fixes `useIsMutating` throwing a `reading properties of undefined` error caused by reading `isMutating` of `client` which could be undefined in the case of some SSR environments/setups. This fix also makes it consistent with `useMutationState` where the subscription is created inside of an effect

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable mutation status updates that react correctly to changing inputs.
  * Proper cleanup of subscriptions on component unmount or re-run, reducing potential leaks.
  * Prevented stale data by resubscribing when relevant reactive sources change.

* **Performance**
  * Reduced unnecessary recalculations and renders by avoiding tracked reads during comparisons.
  * More efficient equality checks to minimize redundant updates.

* **Refactor**
  * Improved reactivity management without altering the public API.


TkDodo:
please add a test that fails on `main` but works with your changes. thanks 🙏 

@ardeora FYI

daviduzumeri:
I hit both of these bugs independently while rebasing a local solid-query patch onto `5.101.0`, and can confirm they're both still present on `main` (`3c77cd0`). Here's the failing test @TkDodo asked for, covering the `useMutationState` infinite loop.

**Root cause of the loop:** when `mutate()` is called inside a tracked scope, the mutation-cache notification fires synchronously *while that scope is the active listener*. On `main`, the subscription in `useMutationState` reads `result()` in that scope, so `result` becomes a dependency of the **caller's** computation; `setResult` then re-triggers it → `mutate()` again → infinite loop. The `untrack(result)` in this PR removes that back-edge.

```tsx
// packages/solid-query/src/__tests__/useMutationState.test.tsx
it('does not feed back into a tracked scope that triggers mutations', async () => {
  const mutationKey = queryKey()
  let effectRuns = 0

  function Page() {
    useMutationState(() => ({ filters: { mutationKey } }))

    const mutation = useMutation(() => ({
      mutationKey,
      mutationFn: (n: number) => sleep(10).then(() => 'data' + n),
    }))

    // mutate() inside a tracked scope fires a cache notification synchronously
    // while this effect is the active listener. On main, useMutationState reads
    // result() in that scope -> result becomes a dep of THIS effect -> setResult
    // re-runs it -> mutate() again -> infinite loop. untrack(result) breaks it.
    createEffect(() => {
      effectRuns++
      if (effectRuns > 100) return // valve so the loop can't hang CI
      mutation.mutate(1)
    })

    return null
  }

  render(() => (
      <Page />
    </QueryClientProvider>
  ))

  expect(effectRuns).toBeLessThan(10)
})
```

Verified: **fails on `main`** (`effectRuns` hits the 100 cap — runaway loop), **passes with this PR's `untrack` change** (`effectRuns` stays at 1). Uses only imports already in the test file.

On the `useIsMutating` half: I couldn't find a client-observable difference in the jsdom suite — moving the subscription into `createEffect` only changes server-render behavior (effects don't run during SSR), which is probably why a regression test here was awkward. Happy to add an SSR-based test if you have a preferred harness for that.

If it's useful, I'm glad to open a PR that rebases this onto current `main` and adds the test above (crediting @dev-rb) — just say the word.

(Unrelated, while I was in here: filed #10907 for an SSR hang on `enabled: false` queries in `useBaseQuery`.)

PR #9650, opened 2025-09-13
feat(svelte-query): merge main branch and fix failed tests

Oungseik:
Merge the main branch into the target branch. Fix locally reference state which cause some test to fail by wrap the reference of state inside a function in [createMutation.svelte.ts](./packages/svelte-query/src/createMutation.svelte.ts)


lachlancollins:
@Oungseik oops I've made a mistake here and accidentally "merged" this PR. I've synced changes from `main` to `svelte-5-adapter`, and it seems like I messed up with the merge conflict resolution with your PR. Would you be able to open a new PR with just the mutation test fixes? Thanks!!

Oungseik:
Yes, I can.

Oungseik:
I pulled the latest changes, ran the tests and all passed. No fixes need right?

lachlancollins:
> I pulled the latest changes, ran the tests and all passed. No fixes need right?

Great, I assumed that there was something you were fixing independently of the merge, but if not then I guess we don't need any more changes!

PR #9675, opened 2025-09-22
fix(chore): report errors of mutation callbacks asynchronously

### LINKED ISSUE #9664: `isPending` remains true if error is thrown in `onError` of `mutate` and a query is invalidated on settle
### Describe the bug

The `isPending` state of a mutation does not reset to `false` but remains `true` if an error is thrown in the `onError` function of the `mutate` function, and the error did originate either within the `mutationFn` or in the `onSuccess` passed to the `useMutate` function.

Furthermore, the queryClient must be configured to invalidate all queries when a mutation settles, and a query must be present in the tree.

In this exact scenario, I could reproduce this bug. When deviating from this setting, the `isPending` state of the mutation did reset to `false` as one would expect.

https://stackblitz.com/edit/tanstack-query-bwruozre?file=src%2Findex.tsx&preset=node

### Steps to reproduce

1. Open the provided link (https://stackblitz.com/edit/tanstack-query-bwruozre?file=src%2Findex.tsx&preset=node)
2. Click on the button "Action!"

### Expected behavior

The value of `isPending` of my mutation should be `false` again - as it does if you remove line 69, which throws the error inside the onError function of the `mutate` call.

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Chrome v140.0.7339.133

### Tanstack Query adapter

react-query

### TanStack Query version

5.89.0

### TypeScript version

5.9.2

### Additional context

I found this error while trying to check what would happen if an `onSuccess` or `onError` callback would throw an error.

SimonSimCity:
Fixes #9664, by catching the error of `onSuccess`, `onError` and `onSettled` callbacks passed to the `mutate` function and reporting it on a separate execution context.

This change will catch all errors and, by passing it to `Promise.reject(e)`, move it to a new execution context, which we explicitly want to ignore (hence the `void` keyword).

By ignoring the error on the newly created execution context, this error is reported by https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event, where tools like Sentry can pick it up.

Raising of an `unhandledRejection` event is crucial to help the developer to be informed about their function misbehaving, and the surrounding try-catch will help securing this libraries code, despite of the fact that something failed in the users codebase.

Not to be confused with https://github.com/TanStack/query/pull/9676, which handles a slightly different problem.

## Summary by CodeRabbit

* **Bug Fixes**
  * Lifecycle callback exceptions (success/error/settled) no longer disrupt mutation processing or observer notifications; such errors are isolated and surfaced asynchronously to avoid app crashes.

* **Tests**
  * Added tests verifying that thrown errors from lifecycle callbacks are reported without preventing observer updates or multiple notification deliveries.


SimonSimCity:
If there's anything that would need a discussion or if you want further insight on why I wrote something the way I did, please ping me. I just want to help having this merged, because I see not having this in as a way of disguising errors in my codebase that I want to be aware of.

PR #9676, opened 2025-09-22
fix(chore): report errors of useMutation callbacks asynchronously

SimonSimCity:
Not to be confused with #9675, which handles a slightly different problem.

By catching the error of `onError` and `onSettled` callbacks passed to the `useMutation` hook and reporting it on a separate execution context, we can ensure that `onSettled` will be called even if `onError` throws an error.

For developers, by reporting the error in a separate context, we enable the ability to globally catch those callback errors by using the [`unhandledRejection` event fired on `window`](https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event).

Unchanged remains the error handling of `onSuccess`, which will (as before) call `onError` and `onSettled`.

Just one thing might remain suspicious - and this is still how it has been before: If the mutation function was successful, and the following `onSettled` throws an error, the `onError` callbacks are called, and also `onSettled` - means, `onSettled`, if it errors, is called twice. I didn't want to go into a deep refactoring here, so I left it this way - the same behaviour as before, just that its second error now raises an `unhandledRejection` event.

## Summary by CodeRabbit

- Bug Fixes
  - Mutation error handling is now more resilient: failures inside lifecycle callbacks (onSuccess/onError/onSettled) are contained, preventing additional unhandled rejections while preserving the original mutation error.

- Tests
  - Added comprehensive tests covering erroneous mutation callbacks, callback ordering, timing, and unhandled-rejection reporting; test harnesses now register and clean up temporary unhandledRejection listeners.

- Breaking Changes
  - Mutation cache constructor now accepts lifecycle callback options at creation time (public API change).

### DIFF
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -8,7 +8,7 @@
   {
     "name": "react minimal",
     "path": "packages/react-query/build/modern/index.js",
-    "limit": "9.11 kB",
+    "limit": "9.12 kB",
     "import": "{ useQuery, QueryClient, QueryClientProvider }",
     "ignore": ["react", "react-dom"]
   }


SimonSimCity:
If there's anything that would need a discussion or if you want further insight on why I wrote something the way I did, please ping me. I just want to help having this merged, because I see not having this in as a way of disguising errors in my codebase that I want to be aware of.

TkDodo:
ugh damn, the PRs didn’t have a changeset and now this isn’t being released :/

SimonSimCity:
I wasn’t aware of changesets and how to use them 😅 let me know if I should open a new PR containing the changeset files.

Issue #9680, opened 2025-09-23
`matchQuery` types don't support partial keys


JosNun:
### Describe the bug

If I type my query key as per the docs (https://tanstack.com/query/latest/docs/framework/react/typescript#typing-query-and-mutation-keys), `matchQuery` expects a full query key instead of a partial query key (even if the `exact` option of the QueryFilters is `false`.

### Your minimal, reproducible example

https://stackblitz.com/edit/vitejs-vite-49gqkj84?file=src%2Fmain.tsx,src%2FApp.tsx

### Steps to reproduce

1. Type the query keys as per the docs (https://tanstack.com/query/latest/docs/framework/react/typescript#typing-query-and-mutation-keys)
2. Try to use a partial key with `matchQuery`

### Expected behavior

If `exact` isn't true (I believe this is the default), I would expect that I would be able to use a partial query key for the QueryFilter passed to `matchQuery` instead of having to pass a complete key.

### How often does this bug happen?

Every time

### Platform

N/A

### Tanstack Query adapter

react-query

### TanStack Query version

5.90.2

### TypeScript version

5.4.5

TkDodo:
hm, I see the issue but the example is somewhat non-realistic. Even a union of all possible types isn’t really what we have in the docs, where the shown pattern would be equivalent to:

```
type QueryKeys = ['entity', ...Array<string>];
```

in that case, calling `matchQuery` would work.

the only solution I could see for your use-case would be to have complicated type gymnastics in all the places where we allow query filters, and I don’t think that’s feasible to tackle from our end, sorry.

JosNun:
Gotcha! Correct me if I'm wrong, but could you use something like below to support partial keys in the `QueryFilters` (QueryFilters.queryKey would be `Prefixes<TQueryKey>`)?

```ts
type DropLast<T extends readonly unknown[]> = T extends readonly [
  ...infer R,
  unknown
]
  ? R
  : never;

type Prefixes<T extends readonly unknown[]> = T extends readonly []
  ? readonly []
  : Prefixes<DropLast<T>> | T;

type QueryKeys = ['turtle', 'item', string] | ['fish', 'item', string];
type PartialQueryKey = Prefixes<QueryKey>;
//     ^? readonly [] | ["turtle"] | ["fish"] | ["turtle", "item"] | ["fish", "item"] | ["turtle", "item", string] | ["fish", "item", string]

```

[...]

PR #9686, opened 2025-09-25
fix(query-core): allow partial query keys in `QueryFilters`

JosNun:
## 🎯 Changes

Add a type to match tuple prefixes to match partial query keys in `QueryFilters` and `MutationFilters`

Fixes tanstack/query#9680

It's worth noting, there _may_ be a case where it makes sense to turn `QueryFilters` (and `MutationFilters` into a discriminated union on the `exact` field, so that we only allow partial query filter types in the case where `exact` is false (or unspecified), but it's possible there are cases where one would want partial query key matching even in the `exact: true` case, and I'm not sure the added constraint adds much benefit. If you'd like me to add that, let me know.

## ✅ Checklist

- [ ] I have tested this code locally with `pnpm test:pr`.

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Improved TypeScript support for filtering queries and mutations by allowing partial key prefixes, enhancing autocomplete and type safety with no runtime changes.

* **Tests**
  * Added type-level tests to validate partial key usage and ensure invalid keys are rejected by the type system.

* **Chores**
  * Prepared a patch release entry to publish the fix.

### DIFF
--- a/.changeset/wise-numbers-double.md
+++ b/.changeset/wise-numbers-double.md
@@ -1,5 +1,5 @@
 ---
-"@tanstack/query-core": patch
+'@tanstack/query-core': patch
 ---
 
 fix: allow partial query keys in `QueryFilters`


packages/query-core/src/types.ts
```diff
@@ -36,6 +36,16 @@ export type Override<TTargetA, TTargetB> = {
...
 
+export type DropLast<T extends ReadonlyArray<unknown>> = T extends readonly [
+  ...infer R,
+  unknown,
+]
+  ? R
+  : never
+
+export type TuplePrefixes<T extends ReadonlyArray<unknown>> =
+  T extends readonly [] ? readonly [] : TuplePrefixes<DropLast<T>> | T
```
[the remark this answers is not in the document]
JosNun: I tried a standalone case in the TypeScript playground that this doesn't seem to be the case (at least, on TS 5.4.5).

https://www.typescriptlang.org/play/?ts=5.4.5&ssl=11&ssc=47&pln=11&pc=48#code/KYDwDg9gTgLgBDAnmYcAiUJgDIEMDOMAPACpygzAB2AJvnAErC40RUA2iAglFLokQCuVANZUIAdyoA+aXAC8cMhWp04UZqw6I4AbQBQcOADpTASyoAzYFEYAaQ3GFjJVBwF1HAfkaOAXHBUwABuNvr6oJCwCMioJIJg7MAAChqWZiDA+KTkIJS09EwsbJw8fALO4lKyCo7Kear0GsXaeu5wPs1anG1wAfGJKWkZWUQYWHiEpDUAPkrhSChwIApKCUmpwOmZ2UXd3Lz8QqJVMtIA3EA

It also seems to infer just fine with the example it gives as what would cause the "type instantiation is excessively deep" error, so I'm inclined to think this may be a hallucination. 
image

JosNun:
@coderabbitai review

packages/query-core/src/types.ts
```diff
@@ -36,6 +36,16 @@ export type Override<TTargetA, TTargetB> = {
...
 export type NoInfer<T> = [T][T extends any ? 0 : never]
 
+export type DropLast<T extends ReadonlyArray<unknown>> = T extends readonly [
+  ...infer R,
+  unknown,
+]
+  ? R
+  : never
+
+export type TuplePrefixes<T extends ReadonlyArray<unknown>> =
```
TkDodo: all those types become part of the public interface, I don’t think we want that so please just inline them in `utils.ts`

JosNun: That's a good thought! I didn't even think of that. Updated.

PR #9720, opened 2025-10-01
fix: allow useQueries with combine property in no-unstable-deps rule

### LINKED ISSUE #9718: False positive in `@tanstack/query/no-unstable-deps` ESLint rule with combine property
### Describe the bug

Hello,
One of the key features of `combine` argument for `useQueries` (or `useSuspenseQueries`) is that its result if referentially stable.
Though the ESLint rule `@tanstack/query/no-unstable-deps` is not aware of that and warns us of result of `useQueries` being unstable.
This case should be added to the linter rule.

https://stackblitz.com/edit/tanstack-query-2ypnpvyn

### Steps to reproduce

- add `useQueries` with `combine`
- add ESLint rule `@tanstack/query/no-unstable-deps`
- use combined value in `useMemo` dependency array

### Expected behavior

Expected to have no warning in this case

Every time

### Screenshots or Videos

_No response_

### Platform

- Version [5.91.0]

### Tanstack Query adapter

None

### TanStack Query version

5.91.0

### TypeScript version

_No response_

### Additional context

_No response_

alimttth:
- Add hasCombineProperty function to detect combine property in useQueries calls
- Skip tracking useQueries with combine as unstable since combine makes result stable
- Add tests for useQueries with and without combine property
- Fixes #9718

## Checklist

- [ ] I have tested this code locally with `pnpm test:pr`.

## Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Linting now treats useQueries calls that include a combine option as stable, avoiding false-positive no-unstable-deps warnings in React hooks.

* **Tests**
  * Added tests validating useQueries scenarios with and without combine, confirming correct diagnostics for stable and unstable dependency patterns (including when the queries object is used as a hook dependency).

### DIFF
--- a/.changeset/ten-crabs-glow.md
+++ b/.changeset/ten-crabs-glow.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/eslint-plugin-query": patch
+---
+
+fix: allow useQueries with combine property in no-unstable-deps rule


TkDodo (approved):

Laubeee:
@TkDodo is there a way to also update the recommendation to use `combine` to make it a stable reference?
> The result of useQueries is not referentially stable, so don't pass it directly into the dependencies array of useCallback. Instead, destructure the return value of useQueries and pass the destructured values into the dependency array of useCallback.

PR #9737, opened 2025-10-08
chore: update eslint-plugin-react-hooks

TkDodo:
what a chore

## Summary by CodeRabbit

- Bug Fixes
  - Improved hydration reliability by syncing options after render, reducing edge-case issues during initial hydration.
  - Corrected mutation state initialization to avoid false negatives on first render, ensuring accurate state tracking.

- Chores
  - Updated React Hooks ESLint plugin to the latest stable version.
  - Aligned projects with the latest recommended React Hooks lint rules for compatibility and consistency.
  - Cleaned up lint directives and test annotations without affecting runtime behavior.

### DIFF
--- a/examples/react/algolia/eslint.config.js
+++ b/examples/react/algolia/eslint.config.js
@@ -1,17 +1,9 @@
 import { tanstackConfig } from '@tanstack/config/eslint'
 import pluginQuery from '@tanstack/eslint-plugin-query'
 import pluginReact from '@eslint-react/eslint-plugin'
-import * as reactHooks from 'eslint-plugin-react-hooks'
 
 export default [
   ...tanstackConfig,
   ...pluginQuery.configs['flat/recommended'],
   pluginReact.configs.recommended,
-  reactHooks.configs.recommended,
-  {
-    rules: {
-      'react-hooks/exhaustive-deps': 'error',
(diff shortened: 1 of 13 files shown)


packages/react-query-devtools/eslint.config.js
```diff
@@ -1,24 +1,22 @@
...
 
 import pluginReact from '@eslint-react/eslint-plugin'
-import * as reactHooks from 'eslint-plugin-react-hooks'
+import reactHooks from 'eslint-plugin-react-hooks'
 import rootConfig from './root.eslint.config.js'
 
 export default [
   ...rootConfig,
-  reactHooks.configs.recommended,
+  // @ts-expect-error wtf
```
achou11, line 9: drive-by comment as I saw your posts on Bluesky...

if you hadn't already come across it, thought it'd be helpful to point to https://github.com/facebook/react/issues/34745, which explains (some of?) the type issues with integrating the plugin's config (and the expect error overrides you needed).

Issue was closed but the fix hasn't been published yet. You may have a better time with integration once it is 😄

TkDodo: ts-expect-error should go away if the types become correct

PR #9743, opened 2025-10-10
fix: don't update dataUpdateCount when setting initial data

### LINKED ISSUE #9656: `isFetchedAfterMount` was broken in commit 1c8a921 / v5.87.1
### Describe the bug

Hello!

It appears that `isFetchedAfterMount` was broken in [5.87.1](https://github.com/TanStack/query/commit/1c8a92167769f322d0206325b98493769afb0961) through that change inadvertently updating the `dataUpdateCount`. We're seeing behavior where the `isLoading` flag is set to `false` when data is loaded from cache, but since `isFetchedAfterMount` is inadvertently set to `true` too early we can't tell if that data was loaded from the server or from the cache. In essence, I believe that in cases where you are persisting data the `isFetchedAfterMount` flag might be entirely broken.

From Claude's analysis:
## Root Cause Analysis

The commit `1c8a92167769f322d0206325b98493769afb0961` introduced logic that updates `initialData` when a Query exists without data. The issue is that when `setData` is called with `manual: true`, it increments `dataUpdateCount` from 0 to 1, even though this data wasn't actually fetched from the server but just set as initial data.

### The Problem Flow:

1. **Prefetch scenario**: A query is created via prefetching (data is undefined, status is pending)
2. **Observer mounts with initialData**: When an observer mounts with `initialData` set
3. **setOptions is called**: The new code in `query.ts:210-218` detects the query has no data and sets the initialData
4. **setData with manual:true**: This calls `setData` with `manual: true`, which increments `dataUpdateCount` from 0 to 1
5. **queryInitialState capture**: The observer captures the query state as `#currentQueryInitialState` with `dataUpdateCount: 1`
6. **isFetchedAfterMount calculation**: When calculated, it compares current `dataUpdateCount` (1) with initial `dataUpdateCount` (1), resulting in false initially, but any subsequent update makes it true prematurely

https://codesandbox.io/p/github/0xdiid/simple-is-fetched-repro/main?import=true

### Steps to reproduce

Steps are in the repro app itself, but a quick copypaste here:

1. Click "1. Prefetch Query" - Creates query with undefined data, but starts fetching with a 3 second delay
2. Click "2. Mount Observer with initialData" - Triggers the bug (if pressed before 3 seconds)
3. If you clicked 2 before 3 seconds, you should see isFetchedAfterMount be true which is not expected
4. You can now click "3. Reset" to reset the state and wait for 3 seconds after clicking 1 to see flags set "properly"

### Expected behavior

As a user, I expect `isFetchedAfterMount` to _only_ trigger when the data is fetched from outside of the persisted cache, which matches previous implementations

Every time

### Screenshots or Videos

https://github.com/user-attachments/assets/77765709-a4e5-46f6-af3b-e565fc483e2a

### Platform

- OS: mac
- Browser: chrome
- Version: 140.0

### Tanstack Query adapter

react-query

### TanStack Query version

v5.87.1

### TypeScript version

v5.9.2

### Additional context

_No response_

0xdiid:
## 🎯 Changes

This introduces a check when updating `dataUpdateCount` so that it isn't updated when setting initial data. This in turn fixes the `isFetchedAfterMount` bug discussed in #9656 

fixes #9656 

## ✅ Checklist

- [x] I have tested this code locally with `pnpm test:pr`

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed isFetchedAfterMount when initialData is applied.
  * Prevented dataUpdateCount from incrementing until a real refetch occurs after initialData.
  * Ensure state is marked success and prior fetch metadata/errors are cleared when initialData is used.

* **Tests**
  * Added tests covering prefetched queries with initialData and observer-initiated refetch behavior.

* **Chores**
  * Prepared patch release notes for the affected package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,13 +7,13 @@
 <br />
 
 <div align="center">
-<a href="https://www.npmjs.com/package/@tanstack/query-core" target="\_parent">
+<a href="https://www.npmjs.com/package/@tanstack/query-core" target="_parent">
   <img alt="" src="https://img.shields.io/npm/dm/@tanstack/query-core.svg" alt="npm downloads" />
 </a>
- <a href="https://github.com/TanStack/query/" target="\_parent">
+ <a href="https://github.com/TanStack/query/" target="_parent">
   <img alt="" src="https://img.shields.io/github/stars/TanStack/query.svg?style=social&label=Star" alt="GitHub stars" />
 </a>
-<a href="https://bundlejs.com/?q=%40tanstack%2Freact-query&config=%7B%22esbuild%22%3A%7B%22external%22%3A%5B%22react%22%2C%22react-dom%22%5D%7D%7D&badge=" target="\_parent">
+<a href="https://bundlejs.com/?q=%40tanstack%2Freact-query&config=%7B%22esbuild%22%3A%7B%22external%22%3A%5B%22react%22%2C%22react-dom%22%5D%7D%7D&badge=" target="_parent">
   <img alt="" src="https://deno.bundlejs.com/?q=@tanstack/react-query&config={%22esbuild%22:{%22external%22:[%22react%22,%22react-dom%22]}}&badge=detailed" alt="Bundle size" />
 </a>
 </div>


packages/query-core/src/query.ts
```diff
@@ -636,7 +636,9 @@ export class Query<
           const newState = {
             ...state,
             data: action.data,
-            dataUpdateCount: state.dataUpdateCount + 1,
+            dataUpdateCount: action.manual && state.data === undefined
+              ? state.dataUpdateCount  // Don't increment for initial data
+              : state.dataUpdateCount + 1,
```
TkDodo: I’m surprised that this doesn’t break any existing tests. For example, we would still want the count to increase if you call `queryClient.setQueryData()` manually while e.g. a prefetch is in-flight. The exception should likely only be for `initialData`

0xdiid: Looks like it did fail on CI, not sure why locally it didn't run those tests. I'll mess with it today as work allows (although this is blocking us from another fix so I'd like to tackle it asap)

TkDodo: I think the fix should not change code here, but instead dispatch `setState` directly with the state we want for initialData rather than calling `setData`

0xdiid: Updated!

.changeset/eight-webs-buy.md
```diff
@@ -0,0 +1,18 @@
...
+'@tanstack/svelte-query-persist-client': patch
+'@tanstack/angular-query-experimental': patch
+'@tanstack/react-query-persist-client': patch
+'@tanstack/solid-query-persist-client': patch
+'@tanstack/query-persist-client-core': patch
+'@tanstack/svelte-query': patch
+'@tanstack/react-query': patch
+'@tanstack/solid-query': patch
+'@tanstack/query-core': patch
+'@tanstack/vue-query': patch
```
TkDodo: I think it’s enough to put `query-core` here, as it will release the others automatically as they are a dependency.

packages/query-core/src/query.ts
```diff
@@ -210,9 +210,11 @@ export class Query<
...
       const defaultState = getDefaultState(this.options)
       if (defaultState.data !== undefined) {
-        this.setData(defaultState.data, {
-          updatedAt: defaultState.dataUpdatedAt,
-          manual: true,
+        this.setState({
+          data: defaultState.data,
+          dataUpdatedAt: defaultState.dataUpdatedAt,
+          status: 'success',
+          error: null,
```
TkDodo: this is the right direction, but I don’t think the fix is complete.

I think we need to spread `...defaultState` because it does a bunch of other things, like setting `errorUpdateCount: 0`.

I was just checking if that would overwrite too much. Essentially, we want everything that we set unconditionally in `setData`:

https://github.com/TanStack/query/blob/1c8a92167769f322d0206325b98493769afb0961/packages/query-core/src/query.ts#L638-L643

_except_ increasing `dataUpdateCount`.

that means we want:

- ✅ `error` set to `null`, which `getDefaultState` does
- ✅ `isInvalidated` set to `false`, which `getDefaultState` does
- ✅ `data` set to our `data`, which `getDefaultState` does
- ✅ `dataUpdatedAt` set, which `getDefaultState` does
- ✅ `status: 'success'`, also handled by `getDefaultState`

But additionally, we would lose:
- `fetchMeta` gets reset to `null`
- `fetchFailureReason` and `fetchFailureCount` get reset, too. It could be that this is just missing from the `success` action too but not sure.

So I think the best fix would be to extract the code above into a function that we can call in the reducer and here that does what we want:

```
function successState(data, dataUpdatedAt) {
  return {
    data,
    dataUpdatedAt: ataUpdatedAt ?? Date.now(),
    error: null,
    isInvalidated: false,
    status: 'success' as const,
  }
}
```

This would make sure these two places don’t diverge in the future.

0xdiid: sweet, lmk if those latest changes look a bit more aligned with this!

packages/query-core/src/query.ts
```diff
@@ -635,16 +637,13 @@ export class Query<
...
             dataUpdateCount: state.dataUpdateCount + 1,
-            dataUpdatedAt: action.dataUpdatedAt ?? Date.now(),
-            error: null,
-            isInvalidated: false,
-            status: 'success' as const,
             ...(!action.manual && {
               fetchStatus: 'idle' as const,
               fetchFailureCount: 0,
               fetchFailureReason: null,
+              fetchMeta: null,
```
TkDodo: let’s not change this here please. I think I’ll investigate if this is necessary separately

packages/query-core/src/query.ts
```diff
@@ -210,9 +210,11 @@ export class Query<
...
       if (defaultState.data !== undefined) {
-        this.setData(defaultState.data, {
-          updatedAt: defaultState.dataUpdatedAt,
-          manual: true,
+        this.setState({
+          ...successState(defaultState.data, defaultState.dataUpdatedAt),
+          fetchMeta: null,
+          fetchFailureReason: null,
+          fetchFailureCount: 0,
         })
```
TkDodo:
```
this.setState(
  successState(defaultState.data, defaultState.dataUpdatedAt),
)
```

The whole idea was to have one defined `successState` so that these two places don’t diverge. I’ll try to fix `fetchMeta` and `fetchFailureCount` separately.

0xdiid:
@TkDodo let me know if you have any more feedback here!

PR #9755, opened 2025-10-14
docs(react-query): add react background retry pausing documentation for v5

Hellol77:
## 🎯 Changes

Related to https://github.com/TanStack/query/pull/9563
Added react-query documentation on how query retries are paused in background tabs, explaining why this overrides refetchIntervalInBackground

Would love to hear your thoughts on this direction.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Documentation**
  * Added a new guide "Background Retry Behavior" explaining how refresh intervals behave when a tab is inactive and how built-in retries interact with background execution.
  * Includes guidance and a full example for disabling built-in retries and implementing a custom background retry strategy, with recommendations for controlling retry timing while a page is inactive.


TkDodo (commented):
can you add this to the `query retry` guide please:

https://tanstack.com/query/v5/docs/framework/react/guides/query-retries

also maybe trim it down a bit because I don’t think it’s such an important / relevant situation that we need to dedicate a full guide to it.

at the end of the day, if you have `refetchInterval` and `refetchIntervalInBackground` turned on, you can just turn off retries and maybe implement a custom `refetchInterval` function that refetches more often when the query is in error state to kinda imitate what retries do...

Hellol77:
@TkDodo Thanks for the feedback! I've updated the documentation as suggested.
Added a section to the query-retries guide instead of creating a separate guide. The section explains the background retry pausing behavior and provides the custom refetch strategy approach you mentioned.

PR #9764, opened 2025-10-16
ci: oidc

manudeli:


lachlancollins (changes requested):
This shouldn't have a changeset, since it doesn't actually affect the *content* of published packages. This also requires specific settings to be enabled by Tanner on NPM.

manudeli:
> This shouldn't have a changeset, since it doesn't actually affect the _content_ of published packages.

I thought adding a changeset and releasing a new version would still be fine, since switching to OIDC changes the publishing workflow itself — even if the package content doesn’t. It also helps confirm that the new OIDC-based release flow works as expected.

> This also requires specific settings to be enabled by Tanner on NPM.

If the release workflow fails, we can just rerun the changeset workflow after Tanner finishes the NPM configuration to verify everything’s working properly. Hopefully that should confirm the setup.

Issue #9798, opened 2025-10-22
A call to fetchQuery is cancelled when React useQuery is unmounted


bartlangelaan:
### Describe the bug

While integrating Tanstack Query with another library, I found `queryClient.fetchQuery()` to fail the first time during development. The promise was cancelled when some component was using the same query with useQuery - only during development, when using StrictMode.

But it can also happen in production. If an useQuery is unmounted, it cancels the query if it is the last one observing the query. Even if some other code is running `queryClient.fetchQuery()`. I would expect Tanstack Query to know that `fetchQuery` is running and not cancel the query.

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/tanstack-query-issue-9798-fetchquery-cancelled-gkt869

### Steps to reproduce

1. Open the codesandbox
2. The useQueryResult will succesfully show the amount of stargazers, but the fetchQueryResult will contain a CancelledError.

### Expected behavior

As a user, I expected the fetch not to be cancelled because fetchQuery was called. I would expect that unrendering a `useQuery` call would not in any way influence a call to `fetchQuery` anywhere in the application.

### How often does this bug happen?

Every time

### Platform

- OS: Linux, Mac
- Browser: Edge, Chrome 141.0.7390.76

### Tanstack Query adapter

react-query

### TanStack Query version

v5.90.5

bartlangelaan:
I have one additional example, which is not using strictmode. The query is cancelled after two seconds: https://codesandbox.io/p/sandbox/tanstack-query-issue-9798-fetchquery-cancelled-without-strictmode-2xcsfm

DogPawHat:
which library are you trying to integrate with and is it open source?

DogPawHat:
I took a closer look at your sandbox and made a slight change here to track when things run https://codesandbox.io/p/devbox/tanstack-query-issue-9798-fetchquery-cancelled-forked-p93k2t

basically, the abort signal will always be false while there are no observers, so if you run fetch query and use the  signal it will abort. however by default the signal isn't used and if it wasn't' using signal it would complete and populate the cache. (https://tanstack.com/query/latest/docs/framework/react/guides/query-cancellation#default-behavior)

so we'd have to change the semantics of that abort signal to make an exception for `fetchQuery` if it's imperative. I'm not sure that's what we want.

TkDodo:
[...]

PR #9799, opened 2025-10-22
Don't cancel fetchQuery when unmounting useQuery

bartlangelaan:
## 🎯 Changes

This is an attempt to solve issue #9798.

It adds an query observer when running `fetchQuery`, to make sure the query is not cancelled if all other observers are unsubscribed.

**I am unfamiliar with this codebase, and at least one test is still failing. I am unsure if this is the right direction to fix the issue. If anyone wants to take over this PR, feel free to do so.**

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where `fetchQuery` would cancel in-progress queries when other observers unsubscribed. Queries now continue to completion as expected.

### DIFF
--- a/packages/query-core/src/queryClient.ts
+++ b/packages/query-core/src/queryClient.ts
@@ -13,6 +13,7 @@ import { focusManager } from './focusManager'
 import { onlineManager } from './onlineManager'
 import { notifyManager } from './notifyManager'
 import { infiniteQueryBehavior } from './infiniteQueryBehavior'
+import { QueryObserver } from './queryObserver'
 import type {
   CancelOptions,
   DefaultError,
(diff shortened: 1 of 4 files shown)


DogPawHat:
The behavior is if there are no observers, the query is inactive and scheduled for garbage collection, but I think it is indeed unintuvitive that an imperative fetch gets canceled.

packages/query-core/src/queryClient.ts
```diff
@@ -362,11 +363,19 @@ export class QueryClient {
...
+
+    if (!isDataStale) {
+      return Promise.resolve(query.state.data as TData)
+    }
+
+
+    const observer = new QueryObserver(this,defaultedOptions)
+    query.addObserver(observer);
+
+    return query.fetch(defaultedOptions).finally(() => query.removeObserver(observer))
```
DogPawHat, line 378: I think we'd want a way to differentiate imperative fetches from observer fetches, rather then this temp observer hack.

TkDodo:
sorry I don’t think this is the way to go.

PR #9810, opened 2025-10-25
fix: support async Svelte

dummdidumm:
## 🎯 Changes

Svelte 5 now supports async await in components. One behavior there is that when a component is created as part of a boundary that has pending async work, its `$effect`s will not run until that async work is done.

The way the code is written right now means you can end up in an infinite pending state: If you do `await someQuery.promise`, the `$effect` for the subscription will never run, and therefore the `await` will never resolve. This PR fixes it.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored reliable support for async Svelte so queries behave correctly in modern Svelte environments.
  * Improved query lifecycle and subscription handling to ensure consistent result updates and proper cleanup across server and client lifecycles.

### DIFF
--- a/packages/svelte-query/src/createBaseQuery.svelte.ts
+++ b/packages/svelte-query/src/createBaseQuery.svelte.ts
@@ -1,3 +1,4 @@
+import { onDestroy } from 'svelte'
 import { useIsRestoring } from './useIsRestoring.js'
 import { useQueryClient } from './useQueryClient.js'
 import { createRawRef } from './containers.svelte.js'
@@ -8,7 +9,6 @@ import type {
   CreateBaseQueryOptions,
   CreateBaseQueryResult,
 } from './types.js'
-import { onDestroy } from 'svelte'
 
 /**
  * Base implementation for `createQuery` and `createInfiniteQuery`
--- a/packages/svelte-query/src/createBaseQuery.svelte.ts
+++ b/packages/svelte-query/src/createBaseQuery.svelte.ts
@@ -93,9 +93,13 @@ export function createBaseQuery<
     },
   )
   // ...and finally also cleanup via onDestroy because that one runs on the server whereas $effect.pre does not.
-  onDestroy(() => {
-    unsubscribe()
-  })
+  // (in a try-catch because it theoretically can be called in a non-component context - that should not happen
+  // but it would be a breaking change technically to error out here. SSR-safe because this wouldn't be called during SSR if it was not in a component)
+  try {
+    onDestroy(() => {
+      unsubscribe()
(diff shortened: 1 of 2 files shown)


dummdidumm:
@lachlancollins could you take a look? This is blocking people from using Tanstack + async Svelte.

tkvw:
This also solves an issue I have where queryFn is called twice in my test after I change the queryOptions. It looks like the effect is triggered when I change the queryOptions. 

Some debugging statements in the code (without patch applied): 

``` 
Component - queryOptions recalculated
createBaseQuery - effect triggered 
createBaseQuery - effect triggered 
queryFn called

Component - queryOptions recalculated
createBaseQuery - options changed
queryFn called
createBaseQuery - options or observer changed
createBaseQuery - effect triggered 
queryFn called
```
With the patch applied: 

```
Component - queryOptions recalculated
queryFn called

Component - queryOptions recalculated
createBaseQuery - options changed
queryFn called
createBaseQuery - options or observer changed
```

packages/svelte-query/src/createBaseQuery.svelte.ts
```diff
@@ -71,13 +72,34 @@ export function createBaseQuery<
     createResult(),
   )
 
-  $effect(() => {
-    const unsubscribe = isRestoring.current
+  // The following is convoluted but necessary:
```
VedAnt-1004, line 75: Thanks for tackling async Svelte lifecycles! A few technical considerations on this implementation:

1. **SSR Subscription Teardown**:
   Because `typeof window === 'undefined'` evaluates to false on the server, `observer.subscribe(...)` will run eagerly during SSR. Given the `try...catch` around `onDestroy`, if this runs in streaming SSR or an edge context where `onDestroy` lifecycle guarantees differ, could this risk leaking query subscriptions in long-running Node processes?

2. **Immediate Redundant Resubscription**:
   The eager `observer.subscribe()` at initialization is immediately followed by `watchChanges(..., 'pre')`, which runs its setup phase and immediately tears down (`unsubscribe()`) and recreates the subscription. Could the eager call be integrated cleanly into the watcher to prevent churn on initial component mount?

3. **Regression Test Coverage**:
   Because this touches tricky boundaries between eager SSR subscription, `$effect.pre`, and uncompleted async component teardown, adding a regression test in `packages/svelte-query/tests` (simulating an unmount before async resolution) would be invaluable to safeguard against future Svelte compiler updates.

lachlancollins (approved):
Sorry for the delay @dummdidumm , thank you for your help!

PR #9835, opened 2025-11-01
feat(query-core): add query() and   infiniteQuery() methods

DogPawHat:
## 🎯 Changes

Implements #10658

- adds `query` and `infiniteQuery` methods as replacements for `fetchQuery`/`fetchInfiniteQuery` 
- new methods respect `select`, `enabled` and `queryFn === skipToken`
- throws errors if `enabled === false` or `queryFn === skipToken` and no cached data is able to be returned.
- updates tests for QueryClient 
  - existing scenarios for `fetchQuery` fetch tests plus the new scenarios introduced (enabled, select, skipToken, static staleTime)
- adds tests for the queryOptions adaptors
- updates to the react-query and preact-query prefetch hooks (`prefetchQuery` -> `query`)
- added new methods to vue-query QueryClient subclass.
- added deprecation tags to old methods (`fetchQuery`, `prefetchQuery`, `ensureQueryData`)
- add new docs, updated prefetching guides

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Added imperative query() and infiniteQuery() methods with preserved option semantics.

* **Documentation**
  * Updated guides and examples to prefer query()/infiniteQuery(), showing migration notes and error‑handling patterns (void … .catch(noop)).

* **Types**
  * Improved execute-option typings so shared options (select, enabled, page params) narrow returned data correctly.

* **Tests**
  * Expanded type and runtime coverage validating new APIs, select/enabled behavior, and pageParam typing.


TkDodo:
> renames fetchQuery and fetchInfiniteQuery to query and infiniteQuery

sorry the idea was not to rename those, but to add new methods and deprecate the old ones. Otherwise, this would be a breaking change.

DogPawHat:
From #9135:

> Per default, this should behave like queryClient.fetchQuery does today:
> - it respects staleTime (like any good query should)
> - it returns a Promise you can await.

`fetchQuery` is still there, it's just a straight passthough for `query` now

packages/query-core/src/queryClient.ts
```diff
@@ -369,6 +364,23 @@ export class QueryClient {
...
+    options: FetchQueryOptions<
+      TQueryFnData,
+      TError,
+      TData,
+      TQueryKey,
+      TPageParam
+    >,
+  ): Promise<TData> {
+    return this.query(options)
+  }
```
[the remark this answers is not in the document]
DogPawHat: I don't want to add the deprecated tag till the docs have been merged in.

packages/react-query/src/__tests__/infiniteQueryOptions.test-d.tsx
```diff
@@ -62,6 +76,45 @@ describe('infiniteQueryOptions', () => {
...
+  })
+  it('should work when passed to infiniteQuery', async () => {
+    const options = infiniteQueryOptions({
+      queryKey: ['key'],
+      queryFn: () => Promise.resolve('string'),
+      getNextPageParam: () => 1,
+      initialPageParam: 1,
+    })
+
+    const data = await new QueryClient().infiniteQuery(options)
```
DogPawHat, line 101: Known type failures here.

as far as I know a `select` pass into a `useInfiniteQuery` can transform the `data: InfiniteData` into whatever (e.g. an Array<>) and that will end up being the data returned from the infinite query*. `infiniteQuery` should do the same I think, but we've got a big type mismatch here at the moment.

*I suspect this is not the best idea

DogPawHat: Type errors from my local VSCode:

```
Types of property 'select' are incompatible.
        Type '((data: InfiniteData<string, number>) => string[]) | undefined' is not assignable to type '((data: InfiniteData<string, number>) => string) | undefined'.
          Type '(data: InfiniteData<string, number>) => string[]' is not assignable to type '(data: InfiniteData<string, number>) => string'.
            Type 'string[]' is not assignable to type 'string'.ts(2345)
```

packages/query-core/src/__tests__/queryClient.test-d.tsx
```diff
@@ -183,6 +215,72 @@ describe('fetchInfiniteQuery', () => {
...
+        initialPageParam: 1,
+        getNextPageParam: () => 2,
+        select: (data) => ({
+          pages: data.pages.map(
+            (x) => `count: ${(x as { count: number }).count}`,
+          ),
+          pageParams: data.pageParams,
+        }),
+      },
+    ])
```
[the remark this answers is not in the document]
DogPawHat: we are adding support for `infiniteQuery` to respect select, but not changing the behavior or `fetchInfiniteQuery`

packages/query-core/src/queryClient.ts
```diff
@@ -147,20 +149,15 @@ export class QueryClient {
   ): Promise<TData> {
     const defaultedOptions = this.defaultQueryOptions(options)
     const query = this.#queryCache.build(this, defaultedOptions)
-    const cachedData = query.state.data
-
-    if (cachedData === undefined) {
-      return this.fetchQuery(options)
-    }
```
TkDodo, line 153: I get why this was removed (because `staleTime: 'static'` achieves the same thing), however, background refetching with `revalidateIfStale` was never triggered if we had no data in the cache, and right now, it’s the first thing that happens.

I guess it’s mostly fine because `query.fetch` will dedupe promises, but there is the edge-case where users pass `cancelRefetch` which would lead to to the first fetch being cancelled.

Not sure what a good way out here would be, except maybe not re-implement `ensureQueryData` with `this.query` rather than keep the current implementation ?

also since `query` now supports `select`, these functions also do transformation in case you’ve passed excess properties!

I think what we should do is leave `ensureQueryData`, `fetchQuery` and `prefetchQuery` alone for now and NOT have them call `.query` or `.infiniteQuery` under the hood.

packages/query-core/src/queryClient.ts
```diff
@@ -362,13 +361,40 @@ export class QueryClient {
...
+    const basePromise = isStale
       ? query.fetch(defaultedOptions)
-      : Promise.resolve(query.state.data as TData)
+      : Promise.resolve(query.state.data as TQueryData)
+
+    const select = defaultedOptions.select
+
+    if (select) {
+      return basePromise.then((data) => select(data))
+    }
```
TkDodo, line 395: we could just `await` the `basePromise` ?

packages/query-core/src/queryClient.ts
```diff
@@ -362,13 +361,40 @@ export class QueryClient {
...
+
+    return basePromise.then((data) => data as unknown as TData)
```
TkDodo: not sure why this is necessary :/

packages/react-query/src/__tests__/infiniteQueryOptions.test-d.tsx
TkDodo: since it works for `useInfiniteQuery`, it should also work for `client.infiniteQuery()`

packages/query-core/src/queryClient.ts
```diff
@@ -428,7 +428,7 @@ export class QueryClient {
   infiniteQuery<
     TQueryFnData,
     TError = DefaultError,
-    TData = TQueryFnData,
+    TData = InfiniteData<TQueryFnData>,
```
DogPawHat, line 451: This seems to match the options used by react query queryOptions but not InfiniteQueryObserverOptions in query-core types.ts.

DogPawHat:
@coderabbitai full review

packages/query-core/src/__tests__/queryClient.test-d.tsx
```diff
@@ -183,6 +215,72 @@ describe('fetchInfiniteQuery', () => {
...
+  })
+})
+
+describe('query', () => {
+  it('should allow passing select option', () => {
+    assertType<Parameters<QueryClient['query']>>([
+      {
+        queryKey: ['key'],
+        queryFn: () => Promise.resolve('string'),
+        select: (data) => (data as string).length,
```
TkDodo: shouldn’t string be inferred here?

```suggestion
        select: (data) => data.length,
```

packages/query-core/src/__tests__/queryClient.test-d.tsx
```diff
@@ -183,6 +215,72 @@ describe('fetchInfiniteQuery', () => {
...
+  it('should allow passing select option', () => {
+    assertType<Parameters<QueryClient['infiniteQuery']>>([
+      {
+        queryKey: ['key'],
+        queryFn: () => Promise.resolve({ count: 1 }),
+        initialPageParam: 1,
+        getNextPageParam: () => 2,
+        select: (data) => ({
+          pages: data.pages.map(
+            (x) => `count: ${(x as { count: number }).count}`,
```
TkDodo: this assertion also shouldn’t be necessary if types flow correctly

packages/query-core/src/__tests__/queryClient.test-d.tsx
```diff
@@ -227,6 +325,16 @@ describe('fully typed usage', () => {
     // Construct typed arguments
     //
 
+    const infiniteQueryOptions: InfiniteQueryExecuteOptions<TData, TError> = {
```
TkDodo: I think you need 4 type params to avoid the type assertion on the query key

packages/query-core/src/__tests__/queryClient.test-d.tsx
DogPawHat: It inferrs as unknown in this case, because I think the generic in `assertTypeOf` has already intialised `TQueryFnData` to `unknown`

DogPawHat: ended up making a stub `queryOptions` fn to make the inference work with select. but the assertType still fails comparing string to unknown.

DogPawHat:
@coderabbit full review

DogPawHat:
Pushing an intial docs pass. I've also opened a draft pr for router here https://github.com/TanStack/router/pull/7100 as the `ensureQueryData` docs live there entirly.

DogPawHat:
Deprecation tags and new tests added

DogPawHat:
@coderabbitai full review

docs/framework/angular/typescript.md
```diff
@@ -215,11 +215,13 @@ export class Component {
...
 
   someMethod() {
-    this.queryClient.prefetchQuery(this.queries.post(23))
+    this.queryClient.query(this.queries.post(23)).catch(noop)
   }
 }
 ```
 
+Because `queryClient.query` preserves `select` and `enabled`, the extracted options behave the same way in both places. The legacy `fetchQuery` and `prefetchQuery` APIs still accept those options at the type level, but they ignore `select` and `enabled` at runtime.
+
```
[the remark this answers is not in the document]
DogPawHat: Check query-core queryClient.ts again. Existing methods have been preserved.

DogPawHat:
@coderabbitai What issues are left from the last review?

DogPawHat:
@coderabbitai Do another full review please

docs/framework/react/guides/migrating-to-v5.md
```diff
@@ -62,6 +52,39 @@ queryCache.findAll(key, filters) // [!code --]
 queryCache.findAll({ queryKey, ...filters }) // [!code ++]
 ```
 
+### Imperative QueryClient methods
+
+These methods are deprecated as of Tanstack Query `INSERT_FUTURE_V5_MINOR` and will be removed in v6.
```
DogPawHat, line 57: Yeah, I know this is a janky way of doing this. Open to suggestions or just saying "these methods are now deprecated"

docs/framework/react/guides/prefetching.md
```diff
@@ -16,49 +16,65 @@ In this guide, we'll take a look at the first three, while the fourth will be co
 
 One specific use of prefetching is to avoid Request Waterfalls, for an in-depth background and explanation of those, see the [Performance & Request Waterfalls guide](./request-waterfalls.md).
 
-## prefetchQuery & prefetchInfiniteQuery
+## Using `query` to prefetch
```
DogPawHat, line 20: might need @Ephem to take a look at the prefetching guide changes and make sure it makes sense.

docs/reference/QueryClient.md
```diff
@@ -18,18 +18,14 @@ const queryClient = new QueryClient({
   },
 })
 
-await queryClient.prefetchQuery({ queryKey: ['posts'], queryFn: fetchPosts })
+await queryClient.query({ queryKey: ['posts'], queryFn: fetchPosts })
 ```
 
 Its available methods are:
 
-- [`queryClient.fetchQuery`](#queryclientfetchquery)
```
DogPawHat, line 26: Should we have a section for these deprecated methods?

docs/reference/QueryClient.md
```diff
@@ -95,19 +92,19 @@ try {
 
 **Options**
 
-The options for `fetchQuery` are exactly the same as those of [`useQuery`](../framework/react/reference/useQuery.md), except the following: `enabled, refetchInterval, refetchIntervalInBackground, refetchOnWindowFocus, refetchOnReconnect, refetchOnMount, notifyOnChangeProps, throwOnError, select, suspense, placeholderData`; which are strictly for useQuery and useInfiniteQuery. You can check the [source code](https://github.com/TanStack/query/blob/7cd2d192e6da3df0b08e334ea1cf04cd70478827/packages/query-core/src/types.ts#L119) for more clarity.
+The options for `query` are exactly the same as those of [`useQuery`](../framework/react/reference/useQuery.md), except the following: `refetchInterval, refetchIntervalInBackground, refetchOnWindowFocus, refetchOnReconnect, refetchOnMount, notifyOnChangeProps, throwOnError, suspense, placeholderData`; which are strictly for useQuery and useInfiniteQuery. You can check the [source code](https://github.com/TanStack/query/blob/7cd2d192e6da3df0b08e334ea1cf04cd70478827/packages/query-core/src/types.ts#L119) for more clarity.
```
DogPawHat, line 95: Need to update the link to the source code

packages/query-core/src/types.ts
```diff
@@ -540,13 +568,34 @@ export type EnsureInfiniteQueryDataOptions<
   revalidateIfStale?: boolean
 }
 
-type FetchInfiniteQueryPages<TQueryFnData = unknown, TPageParam = unknown> =
+type InfiniteQueryPages<TQueryFnData = unknown, TPageParam = unknown> =
   | { pages?: never }
   | {
       pages: number
       getNextPageParam: GetNextPageParamFunction<TPageParam, TQueryFnData>
```
DogPawHat, line 573: `getPreviousPageParam` is not in the types for fetchInfiniteQuery/inifiniteQuery as it turns out. Should it be added in to the types here?

packages/react-query/src/usePrefetchQuery.tsx
```diff
@@ -1,19 +1,25 @@
...
 >(
-  options: UsePrefetchQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+  options: QueryExecuteOptions<TQueryFnData, TError, TData, TQueryKey>,
   queryClient?: QueryClient,
 ) {
   const client = useQueryClient(queryClient)
 
   if (!client.getQueryState(options.queryKey)) {
-    client.prefetchQuery(options)
+    void client.query(options).then(noop).catch(noop)
```
DogPawHat, line 23: These prefetch hooks are the only direct use the adapter packages (react-query and preact-query) make of query, hense the inclusion of those in the change set.

DogPawHat: Hmm, wouldn't the check for `getQueryState` be doing the same thing as `staleTime: 'static'` here?

DogPawHat (commented):
Pushed up the docs rewrite, new tests, deprecation notices, and a few misc changes regarding things like `usePrefetchQuery`

Self-review comments below:

DogPawHat:
@coderabbitai full review

DogPawHat:
@TkDodo btw let me know if I need to start a new PR, this has a lot of commits and outdated Coderabbit comments now sorry.

TkDodo:
> @TkDodo btw let me know if I need to start a new PR, this has a lot of commits and outdated Coderabbit comments now sorry.

@DogPawHat yeah let’s do that!

> - new methods respect select, enabled and queryFn === skipToken
> - throws errors if enabled === false or queryFn === skipToken and no cached data is able to be returned.

I think respecting `select` is a good thing, just to iterate on what the behaviour for `skipToken` / `enabled` is gonna be:

since the functions respect `staleTime`, if there is fresh data, we can just return it with a Promise.resolve(), all good. But if we execute `query.fetch()`, I would expect it to throw an error “Missing queryFn”, which it already does:

https://github.com/TanStack/query/blob/6540a4126b1c087d86d64525e78f32d9920dcd31/packages/query-core/src/utils.ts#L458

This is also what happens currently with `queryClient.fetchQuery`, because the `skipToken` is stored on the query options as `queryFn`, while `enabled` is an observer level property.

```
queryClient.fetchQuery({
  queryKey: ['foo'],
  queryFn: skipToken,
});
```

So the thing I can’t quite wrap my head around is how `enabled` should work. Right now, this just runs the `queryFn`

```
queryClient.fetchQuery({
  queryKey: ['foo'],
  queryFn: () => Promise.resolve(something.foo),
  enabled: !!something,
});
```

and it will just error at runtime if `something` is `undefined`. So respecting `enabled` seems fine, but what kind of error would we throw here? The same was with `skipToken` ?

DogPawHat:
New PR here https://github.com/TanStack/query/pull/10658. The adaptors and docs I'm gonna split out into stack/parallel pr's on top of the query-core feature.

PR #9841, opened 2025-11-03
fix(solid-query): ignore error if data undefined so can redirect

brenelz:
## 🎯 Changes

When doing something like this in tanstack router for solid we get an error that the query can not be undefined. It seems like it should allow it to be undefined and even skip the error if its a redirect.

I'm not 100% this is the right fix but at least its a start.

```tsx
const redirectFn = useServerFn($redirectServerFn)
  const query = useQuery(() => ({
    queryKey: ['redirect-test-ssr'],
    queryFn: () => redirectFn(),
  }))
```

## ✅ Checklist

## 🚀 Release Impact

## Release Notes

* **Bug Fixes**
  * Improved error handling to skip errors when returned data is undefined, enhancing server-side rendering stability and data synchronization reliability.

### DIFF
--- a/.changeset/green-lands-occur.md
+++ b/.changeset/green-lands-occur.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/solid-query': patch
+---
+
+Skip errors if returned data is undefined


packages/solid-query/src/useBaseQuery.ts
```diff
@@ -159,7 +156,7 @@ export function useBaseQuery<
         const query = observer().getCurrentQuery()
         const unwrappedResult = hydratableObserverResult(query, result)
 
-        if (unwrappedResult.isError) {
+        if (result.data !== undefined && unwrappedResult.isError) {
```
brenelz, line 159: this only skips it for the server subscriber. I am not sure if there is a better heuristic here instead of just data is undefined

PR #9847, opened 2025-11-05
fix(core): only attach .then and .catch onto a promise if it gets dehydrated

TkDodo:
## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed unhandled promise rejections when dehydrating pending queries during server-side rendering.

### DIFF
--- a/.changeset/little-berries-search.md
+++ b/.changeset/little-berries-search.md
@@ -1,0 +1,5 @@
+---
+"@tanstack/query-core": patch
+---
+
+fix(core): only attach .then and .catch onto a promise if it gets dehydrated


Svish:
@TkDodo Re [request to try this out](https://github.com/TanStack/query/discussions/8138#discussioncomment-14881311), I reverted our project to before I removed the cancellation token thing, and was able to get the same errors in the log. I then installed the preview-version from this PR, and as far as I can see, it _seems_ there are no longer any errors in the log. So, from my point of view at least, this PR _seems_ to fix the issue. 👍

PR #9875, opened 2025-11-15
(fix): commented out automatic search&replace in solid docs

GEREGUR:
React docs use object syntax inside useQuery({...}), useMutation({...}) and etc., but Solid requires a function returning an object: useQuery(() => ({ ... })). The current replace rule only prepends () => but does not wrap the object in parentheses, causing Solid examples to return undefined. This PR removes the useQuery, useMutation and etc replace rules.

## ✅ Checklist

I didn’t run pnpm run test:pr because this PR does not modify any runtime or documentation build logic. Instead, it proposes a call for correction the Solid-specific documentation to fix a misunderstanding caused by the automatic React → Solid transformation. The current transformation produces invalid Solid examples, so this PR just comments out search&replace rule to call out about incorrect examples of code rather than providing a code fix that would require tests.

## 🚀 Release Impact

## Summary by CodeRabbit

* **Documentation**
  * Updated framework integration guide by modifying hook mapping references.

### DIFF
--- a/docs/framework/solid/guides/queries.md
+++ b/docs/framework/solid/guides/queries.md
@@ -5,9 +5,9 @@ ref: docs/framework/react/guides/queries.md
 replace:
   {
     '@tanstack/react-query': '@tanstack/solid-query',
-    'useMutationState[(]': 'useMutationState(() => ',
-    'useMutation[(]': 'useMutation(() => ',
-    'useQuery[(]': 'useQuery(() => ',
-    'useQueries[(]': 'useQueries(() => ',
+    <!-- 'useMutationState[(]': 'useMutationState(() => ', -->
+    <!-- 'useMutation[(]': 'useMutation(() => ', -->
+    <!-- 'useQuery[(]': 'useQuery(() => ', -->
+    <!-- 'useQueries[(]': 'useQueries(() => ', -->
   }
 ---


TkDodo:
so this code:

```
const result = useQuery(() => { queryKey: ['todos'], queryFn: fetchTodoList })
```

should become:

```
const result = useQuery(() => ({ queryKey: ['todos'], queryFn: fetchTodoList }))
```

so we’re just missing two parentheses. Can you add those instead please?

GEREGUR:
> so we’re just missing two parentheses. Can you add those instead please?

Currently, all code examples are being transformed using a simple search-and-replace rule like:
'useQuery[(]': 'useQuery(() => '
The issue is that we also need to ensure the closing parenthesis is duplicated, and this cannot be reliably solved with basic search & replace. You can try to hack something together with complex regexes, but it’s fragile — especially when dealing with multiline code examples.

My guess is proper fix would require using AST-based code transformations with a JavaScript parser, but that would mean changing the existing doc build pipeline. If you are okay with that, I can hack my way around and try to implement this

GEREGUR:
https://github.com/TanStack/tanstack.com/pull/720

i have created another pull request with proper fixes, took me a while but finally got this done

SeanCassiere:
Just left a [comment](https://github.com/TanStack/tanstack.com/pull/720#pullrequestreview-3823437596) on https://github.com/TanStack/tanstack.com/pull/720, describing why we shouldn't be doing this at the markdown processing layer.

GEREGUR:
> Just left a [comment](https://github.com/TanStack/tanstack.com/pull/720#pullrequestreview-3823437596) on [TanStack/tanstack.com#720](https://github.com/TanStack/tanstack.com/pull/720), describing why we shouldn't be doing this at the markdown processing layer.

so how i should deal with multiline examples?

SeanCassiere:
> so how i should deal with multiline examples?

I just spoke with Dominik and we decided on using block replacement instead. I wrote up a short example of using it here on your other PR - https://github.com/TanStack/tanstack.com/pull/720#issuecomment-3932447689

PR #9878, opened 2025-11-16
fix(types): do not drop readonly for partial QueryFilter matching

### LINKED ISSUE #9871: Partial QueryFilter matching breaks readonly query keys
### Describe the bug

There seems to be an incompatibility between the new functionality introduced in https://github.com/TanStack/query/pull/9686 and the idea of using `readonly` arrays as query keys, at least the way that I have been creating them. If what I am doing is not intended to be supported, feel free to WONTFIX this and I will relax my types.

I have a codegen layer over a RPC library that creates helper objects for every endpoint in the system. Frequently, there are situations where an argument may not yet be available, for example when it is needing to be fed by the result of another query. To deal with this situation without breaking the "rules of hooks", the key generation creates things that look like so:
```
export interface FooArgs { fooId: string }
export const fooVerb = "Foo";
export function fooTqk(args?: FooArgs) {
  return args ? [fooVerb, args] as const : [fooVerb] as const;
}
```
Previous to version 5.90.8, this worked fine. With the partial query matching in place, however, attempts to pass these `*Tqk` objects to `invalidateQueries` generate type errors:

```
return useMutation<FooResponse, DefaultError, FooArgs>({
  [...]
  onSuccess: ([...]) => tqc.invalidateQueries(
            {queryKey: fooTqk({fooId: [...]})}),
});
```
This results in TS2322 errors saying that `readonly ["Foo", FooArgs] | readonly ["Foo"]` cannot be assigned to `readonly [] | readonly ["Foo", FooArgs] | ["Foo"] | undefined`, because a `readonly ["Foo"]` cannot be assigned to an ordinary `["Foo"]`.

N/A

### Steps to reproduce

Pass a `readonly` array to `invalidateQueries`.

### Expected behavior

`readonly` query keys would continue to work.

Every time

### Screenshots or Videos

_No response_

### Platform

Debian sid

### Tanstack Query adapter

react-query

### TanStack Query version

5.90.8

### TypeScript version

5.9.3

### Additional context

_No response_

TkDodo:
closes #9871 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TypeScript typings for query filters: supports unions of tuple keys with different lengths and allows direct key matches as well as tuple-prefix matches.

* **Tests**
  * Added test coverage validating query key type handling for unions of varying tuple lengths.

### DIFF
--- a/.changeset/orange-boxes-visit.md
+++ b/.changeset/orange-boxes-visit.md
@@ -2,4 +2,4 @@
 "@tanstack/query-core": patch
 ---
 
-fix(types): do not drop readonly for partial QueryFilter matching
+fix(types): allow QueryFilters union with different lengths


.changeset/orange-boxes-visit.md
```diff
@@ -0,0 +1,5 @@
+---
+"@tanstack/query-core": patch
+---
+
+fix(types): do not drop readonly for partial QueryFilter matching
```
TkDodo:
```suggestion
fix(types): allow QueryFilters union with different lengths
```

PR #9886, opened 2025-11-19
fix: Prevent infinite render loops when useSuspenseQueries has duplicate queryKeys

olivierbeaulieu:
## 🎯 Changes

Fixes an infinite re-render loop in useSuspenseQueries (and useQueries) when duplicate query keys are provided introduced in [#8771](https://github.com/TanStack/query/pull/8771).

While duplicate query keys are technically unsupported and log a warning, the current behavior causes the application to crash due to an infinite render loop, which is a severe degradation.

### The issue

QueriesObserver used a Map keyed by queryHash to track observers, causing duplicate queries to overwrite each other in the map. This forced all duplicate queries to share a single observer instance. During updates, this shared observer was repeatedly reconfigured with conflicting options (like different select functions) within the same render cycle, causing its state to thrash and triggering infinite re-renders.

### The fix

Updated QueriesObserver to track a list of observers per hash (Map<string, QueryObserver[]>). This ensures each duplicate query is matched to its own distinct, stable observer instance, preventing the state conflict.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents infinite render loops when using duplicate query keys with useSuspenseQueries; improves stability when multiple queries share the same key.

* **Tests**
  * Added tests validating duplicate query keys no longer cause unbounded re-renders under Suspense and that resolution occurs as expected.

* **Chores**
  * Released a changeset bumping related packages to patch versions containing the fix.

### DIFF
--- a/packages/query-core/src/queriesObserver.ts
+++ b/packages/query-core/src/queriesObserver.ts
@@ -236,7 +236,7 @@ export class QueriesObserver<
     queries: Array<QueryObserverOptions>,
   ): Array<QueryObserverMatch> {
     const prevObserversMap = new Map<string, Array<QueryObserver>>()
-    
+
     this.#observers.forEach((observer) => {
       const key = observer.options.queryHash
       if (!key) return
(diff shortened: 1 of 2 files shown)


TkDodo:
nice work! does that mean we can actually support the same key multiple times within useQueries?

olivierbeaulieu:
@TkDodo I'm not sure if this issue still applies? https://github.com/TanStack/query/issues/4187

PR #9893, opened 2025-11-23
refactor(angular-query): require Angular v19+ and rely on component effect schedule timing

### LINKED ISSUE #9735: Angular query: Optimistic updates are not synchronous
### Describe the bug

Here's a stackblitz that repro's the issue about 20% of the time (I think it's timing related): https://stackblitz.com/edit/stackblitz-starters-hn17w1mv?file=src%2Fmain.ts,src%2Findex.html

If you click "edit" and change the "whatever" to just "1", then tab out to trigger blur, you'll see the "whatever" show up for 1 frame. I think this is happening because the optimistic update (which should set the title to "1") is not happening until the next frame. So when I set the "isEditingTitle" signal to false, it flickers the old title for 1 frame. I would expect the optimistic update to be immediate and synchronous, like how normal signals behave, so this flicker never happens.

Generally speaking, this can happen whenever you have a UX that let's the user *preview* an edit before committing it. In this case, the preview is the input text box, and commiting it is blurring. It would be even more noticeable if the input field looked exactly like the non-editing text.

What are people's thoughts on this? Is this not a use case many encounter?

https://stackblitz.com/edit/stackblitz-starters-hn17w1mv?file=src%2Fmain.ts,src%2Findex.htmlhttps://stackblitz.com/edit/stackblitz-starters-hn17w1mv?file=src%2Fmain.ts,src%2Findex.html

### Steps to reproduce

1. Click the "edit" button
2. Change "whatever" to just "w"
3. Press Tab to trigger blur and finish editing.

### Expected behavior

20% of the time, you'll see "whatever" show up for 1 frame. Desired behavior: I would expect the optimistic update to immediately take effect, and the text immediately shows "w".

Sometimes

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: Google Chrome, Version 141.0.7390.54 (Official Build) (arm64)

### Tanstack Query adapter

angular-query

### TanStack Query version

5.69.1

### TypeScript version

^5.69.1

### Additional context

_No response_

arnoud-dv:
## Improves reactivity and timing

To benefit from the more mature signal ecosystem and PendingTasks integration in recent Angular versions this drops support for Angular v16-v18. These Angular versions are also not under LTS support anymore, so we keep support for all Angular versions currently under active or LTS support: v19 to v21

## 🎯 Changes

  - Bumped peer dependency from Angular 16+ to Angular 19+
  - Use queueMicrotask instead of notifyManager.batchCalls to prevent render glitches fixes #9735 
  - Improve Angular PendingTasks integration
  - Query timing now correctly depends on framework component effect scheduling
  - use linkedSignal

  Breaking Changes

  - Requires Angular v19+
  - Component effect timing may slightly change initialization order in edge cases

Aliasing query properties on a component will result in an error:

```ts
readonly #hasPreviousPage = this.query.hasPreviousPage
```

This should be resolved once the signalProxy is replaced.

## ✅ Checklist

## 🚀 Release Impact

## Release Notes

* **Breaking Changes**
  * Angular v19+ is now required (previously v16+)

* **New Features**
  * Component effect scheduling support for Angular v19

* **Improvements**
  * Simplified query API with direct access to public query state properties
  * Enhanced query lifecycle management for better component integration


arnoud-dv:
@coderabbitai review

benjavicente:
How we can help on the PR?

I have been playing around with it in [this personal repo with PRs](https://github.com/benjavicente/query/pulls) and would love to contribute to get it out of experimental. What I have found out is:

-  **Pending Tasks:** `await fixture.whenStable()` will return immediately while `await app.whenStable()` does not, causing this issue: https://github.com/TanStack/query/issues/9910 . This might be fixed by starting the pending task [right away in queries](https://github.com/benjavicente/query/pull/1/changes/BASE..63664abb657f78abbd61ca3c416284a8adfdbc50#r2616734670) and avoiding a [race condition in mutations](https://github.com/benjavicente/query/pull/1/changes/BASE..63664abb657f78abbd61ca3c416284a8adfdbc50#r2616734670). Testing can be done by adding `expect(fixture.isStable()).toBe(false)` on any text with a component fixture where a pending state is expected.
- **`Signal Proxy`**: To keep the current API while avoiding problems when accessing the proxy, it is ideal to pass thought the proxy the array of functions it should not convert to signals. Here is a [PR with the idea](https://github.com/benjavicente/query/pull/4). I also played around chaining it to [not transform the reactive proxy getters to functions](https://github.com/benjavicente/query/pull/2) that would improve type narrowing and reduce type complexity, but I understand that a "store pattern" like solid might not be appropriate with angular's conventions 🤔

PR #9900, opened 2025-11-23
test(svelte-query-persist-client/PersistQueryClientProvider): switch to fake timers, replace 'waitFor' with 'advanceTimersByTimeAsync', and use 'sleep().then()' pattern

sukvvon:
## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Added a public onFetch prop to a test component to notify when data is fetched.
  * Default test view display changed from "undefined" to "null".

* **Tests**
  * Introduced short artificial delays in query flows to better simulate async timing.
  * Converted several tests to use fake timers and explicit timer-driven advancement for deterministic state progression.


packages/svelte-query-persist-client/tests/InitialData/InitialData.svelte
```diff
@@ -8,11 +8,7 @@
 
   const query = createQuery(() => ({
     queryKey: ['test'],
-    queryFn: async () => {
-      await sleep(5)
-      return 'fetched'
-    },
-
+    queryFn: () => sleep(10).then(() => 'fetched'),
```
TkDodo, line 11: why is this better than `await` ?

packages/svelte-query-persist-client/tests/FreshData/FreshData.svelte
```diff
@@ -10,8 +10,7 @@
 
   const query = createQuery(() => ({
     queryKey: ['test'],
-    queryFn: () => Promise.resolve('fetched'),
-
+    // queryFn is provided by queryClient.setQueryDefaults in test
```
TkDodo: I don’t see `setQueryDefaults` being called here anywhere?

sukvvon: I changed my mind and reverted to the original approach, adding the `onFetch` callback and `sleep`.

packages/svelte-query-persist-client/tests/InitialData/InitialData.svelte
sukvvon: @TkDodo 

In my opinion, `sleep().then()` is more concise than `async/await`. There's no functional difference, but it avoids the async keyword and extra braces, and makes the timer-to-callback dependency explicit in a single line. I discussed this with @manudeli.

```ts
// await: 4 lines
async () => {
  await sleep(10)
  return 'fetched'
}

// then: 1 line
() => sleep(10).then(() => 'fetched')
```

TkDodo: okay but we have over 350 instances of `await sleep(...)` and I think it’s unnecessary to refactor them “just because”. This is also only true if you have a single `await` statement, we also sometimes have:

```
mutationFn: async () => {
  results.push('start-A')
  await sleep(10)
  results.push('finish-A')
  return 'a'
}
```

which is much more readable with async/await.

If it’s all the same, I’d like to keep it as-is.

sukvvon: So the `.then()` ones from the fake timers PRs that already got merged stay put, use `.then()` for single-statement cases as I suggested, and keep `await` otherwise from now on?

TkDodo: we don’t have a lint rule for this so I don’t really care but I don’t see the value in re-writing a pattern like this in our tests. We now have a mixture of both usages anyways 🤷‍♂️

PR #9901, opened 2025-11-23
test(react-query-persist-client/PersistQueryClientProvider): convert complex 'then' callback to async/await and add delayed assertion for refetch verification

sukvvon:
## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Tests**
  * Updated tests to use async/await-style control flow and adjusted simulated timing to verify persisted hydration remains stable across a slightly extended delay, ensuring hydration and subsequent fetch behavior are consistently validated.


packages/react-query-persist-client/src/__tests__/PersistQueryClientProvider.test.tsx
```diff
@@ -41,8 +41,9 @@ const createMockErrorPersister = (
         // noop
       },
       async restoreClient() {
-        await sleep(10)
-        throw error
+        return sleep(10).then(() => {
+          throw error
+        })
```
TkDodo: same question: why? are there any functional difference ?

sukvvon: @TkDodo If we apply the pattern from https://github.com/TanStack/query/pull/9900#discussion_r2564012983,  we could write it like this:

```tsx
return sleep(10).then(() => Promise.reject(error))
```

What do you think about this approach?

PR #9906, opened 2025-11-25
feat(persist): add refetchOnRestore option to refetch queries after hydration

### LINKED ISSUE #9889: `refetchOnMount: 'always'` does not work with persister
### Describe the bug

`refetchOnMount: 'always'` does not trigger an API refetch with persister, as the first fetch for a query with persister is to load from the store and not API call.

TBA

### Steps to reproduce

1. Use `useQuery` with `refetchOnMount: 'always'`, `persister`, and `staleTime`.
2. Open browser network tab and reload the page, no API call is made.

### Expected behavior

An API call is made to refetch the data.

Every time

### Screenshots or Videos

_No response_

### Platform

*

### Tanstack Query adapter

None

### TanStack Query version

v5.90.7

### TypeScript version

_No response_

### Additional context

_No response_

yureiblack:
## Description

Fixes the issue where `refetchOnMount: 'always'` does not trigger an API refetch when using a persister. The first fetch with a persister loads from the store instead of making an API call, causing the expected refetch behavior to be skipped.

## Problem

When using `useQuery` with:
- `refetchOnMount: 'always'`
- A persister
- `staleTime` set

Reloading the page showed no API call in the network tab because hydration from the persister takes precedence over the refetch mount logic.

## Solution

Added a new `refetchOnRestore` option to `persistQueryClientRestore()` that refetches queries after successful hydration from the persister:

- `refetchOnRestore: true` (default) - Refetches only stale queries after restoration
- `refetchOnRestore: 'always'` - Always refetches all queries after restoration
- `refetchOnRestore: false` - No refetch after restoration

## Changes

- Added `refetchOnRestore?: boolean | 'always'` option to `PersistedQueryClientRestoreOptions`
- Implemented refetch logic in `persistQueryClientRestore()` after hydration completes
- Iterates through all queries and refetches based on the flag value and stale status

## Testing

- Run: `npx nx run @tanstack/query-persist-client-core:test:lib`
- All existing tests pass
- New option defaults to `true` for backward compatibility

## Related Issue

Resolves #9889

## Summary by CodeRabbit

* **New Features**
  * Added a configurable refetch-on-restore option to control how queries are refreshed after restoring persisted data (enabled by default, can be disabled or forced).
* **Style**
  * Normalized and reformatted the example TypeScript configuration and updated JSX/compilation settings for consistency.

### DIFF
--- a/examples/react/nextjs-suspense-streaming/tsconfig.json
+++ b/examples/react/nextjs-suspense-streaming/tsconfig.json
@@ -1,11 +1,7 @@
 {
   "compilerOptions": {
     "target": "ES5",
-    "lib": [
-      "dom",
-      "dom.iterable",
-      "esnext"
-    ],
+    "lib": ["dom", "dom.iterable", "esnext"],
     "allowJs": true,
     "skipLibCheck": true,
     "strict": true,
@@ -16,7 +12,7 @@
     "moduleResolution": "Bundler",
     "resolveJsonModule": true,
     "isolatedModules": true,
-    "jsx": "react-jsx",
+    "jsx": "preserve",
     "incremental": true,
     "plugins": [
       {
@@ -31,7 +27,5 @@
     ".next/types/**/*.ts",
     ".next/dev/types/**/*.ts"
   ],
-  "exclude": [
-    "node_modules"
-  ]
+  "exclude": ["node_modules"]
 }


yureiblack:
> sorry, I don’t think this is something we want:
> 
> For one, when you use `<PersistQueryClientProvider>`, this should work out-of-the-box as queries start to subscribe only after restore has finished. That means they will always be inactive, and become active afterwards. At this point, `refetchOnMount` should trigger.
> 
> Also, we expose an `onSuccess` callback already that runs after restore is finished, so you can do additional stuff there.
> 
> If you use `persistQueryClientRestore` manually, you can just `await` it and do a refetch yourself afterwards too.

Thanks for the clarification!
I understand the reasoning now.
I’ll review my usage and adjust accordingly.
Appreciate the guidance!

PR #9907, opened 2025-11-25
docs(vue): change query keys example to typescript

sybers:
## 🎯 Changes

This example is written in Javascript, and the fact that `todoId` is a `Ref` is a bit hard to identify.

I propose to migrate this example to Typescript, with an import to Vue's Ref type and make the parameter to the example composable explicit.

There are more docs that could be improved this way, if it suits you I could also make the changes to these docs :)

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Documentation**
  * Updated Vue query keys guide to TypeScript examples.
  * Added Vue Ref type annotation and import for clearer examples.
  * Example function signature now accepts a Ref<string> and demonstrates using .value when accessing the ID.
  * Improved type-safety and clarity in the documentation samples.

### DIFF
--- a/docs/framework/vue/guides/query-keys.md
+++ b/docs/framework/vue/guides/query-keys.md
@@ -7,7 +7,7 @@ ref: docs/framework/react/guides/query-keys.md
 [//]: # 'Example5'
 
 ```ts
-import type { Ref } from 'vue';
+import type { Ref } from 'vue'
 
 function useTodos(todoId: Ref<string>) {
   const queryKey = ['todos', todoId]


docs/framework/vue/guides/query-keys.md
```diff
@@ -6,8 +6,10 @@ ref: docs/framework/react/guides/query-keys.md
 
 [//]: # 'Example5'
 
-```js
-function useTodos(todoId) {
+```ts
+import type { Ref } from 'vue';
+
+function useTodos(todoId: Ref<string>) {
```
DamianOsipiuk, line 12: Well, this is not entirely true. While you might use both `plain value` or `ref` for reactivity, you might also pass `computed` or `getter` in this place.

So if we are going to change the docs to include `types` in examples, it should actually use `MaybeRefOrGetter`

sybers: Hey @DamianOsipiuk, thanks for your quick feedback!

It's fine for me to use this more precise type, but the `.value` isn't going to work anymore as the `todoId` here could be a plain function when it's a getter.

So the final example will be more complex, since we'd have to do something like `computed(() => toValue(todoId))` to handle all cases.

So,
- If the goal is to be as precise as possible in the docs, then using `MaybeRefOrGetter` is the way to go.
- If the goal is to provide a simple example, I'd vote for keeping a simple ref, just to outline that the `queryKey` can have reactive dependencies.

WDYT ?

DamianOsipiuk: Makes sense

Issue #9728, opened 2025-10-04
Query caches error result and never calls queryFn


travikk:
### Describe the bug

Hey,

So I found a very specific conditions, in which useQuery will cache an error thrown by queryFn, and then from there on out always return that error never calling queryFn again.

As you can see in the example from Stackblitz below, it only occurs when 2 things are set:

`staleTime: Infinity`

as well as this weird counter in the component itself, right after mounting.

```

  useEffect(() => {
    console.log('MyComponent MOUNT');
    setCount(1);

    return () => console.log('MyComponent UNMOUNT ***');
  }, []);
```

This obviously is a simplified use-case meant to reproduce a bug, but one can imagine in a complex app there will be a hook that triggers component re-render on component mount. 

### Your minimal, reproducible example

https://stackblitz.com/edit/vitejs-vite-zhvjhgqf?file=src%2FApp.jsx

### Steps to reproduce

1. Go to Stackblitz
2. Open browser DevTools and observe console output.
3. Look for **About to fetch! Count: N**
4. Click **Manual Retry** until the counter N reaches 3, at that point the queryFn will try to fetch from a non-existing endpoint resulting in an error thrown
5. The ErrorBoundary catches the error and displays fallback component. 

### Expected behavior

When you hit "Try Again" on the screen the expected behavior is that the queryFn is invoked again and **About to fetch! Count: 4** is printed in the console but that never happens.

From this point on, useQuery cannot escape the previously returned error and it'll forever be failing

### How often does this bug happen?

Every time

### Platform

- macOS
- Chrome 140
- "@tanstack/react-query": "^5.90.2",

### Tanstack Query adapter

react-query

### TanStack Query version

v5.90.2

### TypeScript version

NA

TkDodo:
When the error boundary resets, it’s a signal to us that we can continue re-fetching. We have to temporarily stop this from happening because of how react re-renders the component again even though we throw an error. That’s the whole reason why we have a QueryErrorResetBoundary - I wish we wouldn’t need that at all.

We also have an effect that clears that re-set:

https://github.com/TanStack/query/blob/295a5d586f40176a7452c8e4b7ff8c5f39905caf/packages/react-query/src/errorBoundaryUtils.ts#L41-L47

This runs _after_ the component has rendered once, where we expect it to have put the Query into a `pending` state already.

[...]

PR #9927, opened 2025-11-27
fix(query-core): ensure query refetches on mount/retry when status is error (#9728)

xiangnuans:
error (#9728)

## 🎯 Changes

- Modified `shouldFetchOn` in `packages/query-core/src/queryObserver.ts`:
  - Added a check for `query.state.status === 'error'`.
  - This ensures that when a component mounts or explicitly requests a refetch (e.g. `refetchOnWindowFocus`, `refetchOnMount`, or manual retry via ErrorBoundary), it will trigger a fetch if the query is in an error state, even if the data is technically considered "fresh" (not stale) due to `staleTime: Infinity`.
- Added regression test in `packages/react-query/src/__tests__/issue-9728.test.tsx` to reproduce the issue and verify the fix.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Queries now refetch on mount or retry when in an error state, even if existing data is not stale.

* **Tests**
  * Added tests validating error-boundary retry and refetch behavior to ensure data is recovered and displayed after failures.
  * Added tests covering refetch-on-window-focus behavior after background errors for different stale-time settings.

### DIFF
--- a/packages/query-core/src/__tests__/queryObserver.test.tsx
+++ b/packages/query-core/src/__tests__/queryObserver.test.tsx
@@ -1410,6 +1410,7 @@ describe('queryObserver', () => {
       queryFn,
       staleTime: 'static',
       refetchOnWindowFocus: true,
+      retry: false,
     })
 
     const unsubscribe = observer.subscribe(() => undefined)
@@ -1451,6 +1452,7 @@ describe('queryObserver', () => {
       queryFn,
       staleTime: 1000,
       refetchOnWindowFocus: true,
+      retry: false,
     })
 
     const unsubscribe = observer.subscribe(() => undefined)


packages/query-core/src/queryObserver.ts
```diff
@@ -781,7 +781,11 @@ function shouldFetchOn(
   ) {
     const value = typeof field === 'function' ? field(query) : field
 
-    return value === 'always' || (value !== false && isStale(query, options))
+    return (
+      value === 'always' ||
+      (value !== false &&
+        (isStale(query, options) || query.state.status === 'error'))
+    )
```
TkDodo: I think we can make the argument that if a query errors and has data, that this data should be considered stale. However, this check here doesn’t convey that and likely misses a couple of cases, e.g. `staleTime: 'static'`, which definitely shouldn't be refetched.

We do however have the option to mark a query as `invalidated`, and then the functions `isStale` and `isStaleByTime` on a query both check that in the correct order.

So maybe the right fix would be to just set `isInvalidated` in the reducer when we get an error:

https://github.com/TanStack/query/blob/1638c028df55648995d04431179904371a189772/packages/query-core/src/query.ts#L650

I think setting it to:

```
// flag existing data as invalidated if we get a background error
isInvalidated: true,
```

should suffice because if we get an error and have no data yet, we’re always implicitly stale

xiangnuans: Good point — I agree that semantically, “error with existing data” could be modeled as invalidated, and that would keep all stale logic centralized in `isStale` / `isStaleByTime`.

For this PR, I tried to keep the change narrowly scoped to fixing the concrete regression in #9728 via `shouldFetchOn`, without altering global invalidation semantics (which could have broader effects, e.g. with `staleTime: 'static'`).

That said, I do think invalidating on background errors could be a cleaner long-term model. Happy to explore that in a follow-up PR if that aligns with the direction you’d prefer.

TkDodo: yeah I’d prefer this direction really because I don’t think I can merge this PR as it would have the regression of queries with `staleTime: 'static'` getting refetched if they have an error even though they shouldn't. The flow is:

-  `staleTime: 'static'`
- fetch successfully
- re-fetch with an error

now this solution would then return `true` from `refetchOn`, but the solution that sets `isInvalidated` would not.

If you want, you can change the approach in this PR or file a new one.

TkDodo: would also be great to have a test for this flow ⬆️

xiangnuans: Got it — thanks for the clarification, that makes sense 👍
I agree this would be a regression for `staleTime: 'static'`, and the invalidation approach is the right abstraction.

I’ll update this PR to:

* remove the observer-level `status === 'error'` check
* instead mark queries as `isInvalidated` on background errors in the reducer
* add a regression test for the `staleTime: 'static' → success → error` flow you described

Will push an update shortly.

TkDodo: i think your fix is correct but we can simplify the code by just doing `isInvalidated: true` unconditionally. The reason is that if we have no data, the query is always considered stale:

https://github.com/TanStack/query/blob/1638c028df55648995d04431179904371a189772/packages/query-core/src/query.ts#L308-L312

so setting `isInvalidate: true` if we have no data doesn’t matter. Maybe leave a comment though :)

xiangnuans: Makes sense! Since `isStaleByTime` already treats queries with no data as stale, we can simplify this unconditionally. Updated accordingly.

packages/query-core/src/query.ts
```diff
@@ -658,6 +658,7 @@ export class Query<
             fetchFailureReason: error,
             fetchStatus: 'idle',
             status: 'error',
+            isInvalidated: true,
```
TkDodo, line 663:
```suggestion
            // flag existing data as invalidated if we get a background error
            // note that "no data" always means stale so we can set unconditionally here
            isInvalidated: true,
```

.changeset/open-keys-create.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tanstack/query-core': patch
+---
+
+Fix: Ensure queries refetch on mount or retry when in error state, even if data is not stale.
```
TkDodo:
```suggestion
Fix: Always treat existing data as stale when query goes into error state.
```

TkDodo:
your new tests don’t seem to pass

xiangnuans:
Thanks for catching that! The tests were timing out because when `refetch()` throws an error, the default retry mechanism (3 retries with exponential backoff) was being triggered. With fake timers, the retry delays weren't being advanced, causing the tests to hang.

I've fixed this by adding `retry: false` to both background error tests, which disables retries and allows them to complete immediately after the error. This is appropriate for these tests that specifically check background error behavior.

The tests are now passing ✅

PR #9959, opened 2025-12-09
fix(query-core): prevent duplicate abort event listeners in infinite queries

kamja44:
## 🎯 Changes
Fixed a memory leak in infiniteQueryBehavior.ts where abort event listeners were being registered multiple times when the signal property was accessed repeatedly during pagination. Root Cause: The addSignalProperty getter was registering a new abort event listener on every signal access without checking if a listener was already attached. This pattern differed from query.ts, which correctly handles this scenario. Solution:
- Added listenerAttached flag to prevent duplicate listener registration (line 25, 32-33)
- Added { once: true } option for automatic listener cleanup (line 39)
- Added test case to verify no duplicate listeners when signal is accessed multiple times

Files Changed:
- packages/query-core/src/infiniteQueryBehavior.ts - Fixed duplicate listener bug
- packages/query-core/src/__tests__/infiniteQueryBehavior.test.tsx - Added regression test

Impact:
- Prevents memory leaks in infinite queries with many pages
- No API changes, fully backward compatible
- Test coverage: 100% statements, 100% branches for infiniteQueryBehavior.ts

## ✅ Checklist

- [O] I have followed the steps in the [Contributing guide](https://github.com/TanStack/query/blob/main/CONTRIBUTING.md).
- [O] I have tested this code locally with `pnpm run test:pr`.

## 🚀 Release Impact

- [O] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a memory leak in infinite queries by ensuring abort listeners are not registered more than once, improving stability and memory usage during paginated loads.

* **Tests**
  * Added tests confirming abort listeners are attached only once even when the abort signal is accessed repeatedly during query execution.

* **Chores**
  * Added a changeset entry to publish the patch.

### DIFF
--- a/docs/framework/react/guides/caching.md
+++ b/docs/framework/react/guides/caching.md
@@ -27,7 +27,7 @@ Let's assume we are using the default `gcTime` of **5 minutes** and the default 
   - When the request completes successfully, the cache's data under the `['todos']` key is updated with the new data, and both instances are updated with the new data.
 - Both instances of the `useQuery({ queryKey: ['todos'], queryFn: fetchTodos })` query are unmounted and no longer in use.
   - Since there are no more active instances of this query, a garbage collection timeout is set using `gcTime` to delete and garbage collect the query (defaults to **5 minutes**).
-- Before the cache timeout has completed, another instance of `useQuery({ queryKey: ['todos'], queryFn: fetchTodos })` mounts. The query immediately returns the available cached data while the `fetchTodos` function is being run in the background. When it completes successfully, it will populate the cache with fresh data.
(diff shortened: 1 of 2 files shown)


TkDodo:
thanks, I think this was implicitly fixed when we extracted the behavior because we needed to re-use it for streamedQuery:

https://github.com/TanStack/query/blob/4a0a78afbc2432f8cb6828035965853fa98c86a0/packages/query-core/src/utils.ts#L489

It would still be great if you could contribute only the test in a separate PR 🙏

PR #9963, opened 2025-12-11
Fix/streamed query consume aware signal

razhayat:
## 🎯 Changes

Made `streamedQuery` aware of consumption of `context.signal`.
If the signal is not consumed, the chunks will keep updating the data in the background.

Before, a `streamFn` that doesn't use `context.signal` would ignore the chunks that come after the signal is aborted, even though the signal wasn't used. This causes the `query` to have partial data when the it's mounted again. And because there is cached data, the `streamFn` isn't executed again, so the data may be partial forever.
This PR makes it so chunks will not get ignored after the signal is aborted if the signal wasn't consumed.

+ Added `addConsumeAwareSignal` util and used it in `infiniteQueryBehavior` and `streamedQuery`
+ Added a test case that checks that the `streamFn` isn't aborted if the signal is not consumed.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cancellation handling for streamed queries — streams no longer prematurely abort when the abort signal isn't consumed, allowing data to progress across ticks and refetches to behave correctly.

* **New Features**
  * Stream handlers now receive a consumption-aware context exposing the abort signal so stream logic can observe cancellation and react accordingly.

* **Chores**
  * Core utilities and internal wiring updated to provide consume-aware signal support for cancellation-aware contexts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     {
       "name": "react full",
       "path": "packages/react-query/build/modern/index.js",
-      "limit": "12.02 kB",
+      "limit": "12.10 kB",
       "ignore": [
         "react",
         "react-dom"
@@ -103,7 +103,7 @@
     {
       "name": "react minimal",
       "path": "packages/react-query/build/modern/index.js",
-      "limit": "9.08 kB",
+      "limit": "9.11 kB",
       "import": "{ useQuery, QueryClient, QueryClientProvider }",
       "ignore": [
         "react",


packages/query-core/src/streamedQuery.ts
```diff
@@ -73,10 +78,25 @@ export function streamedQuery<
 
     let result = initialValue
 
-    const stream = await streamFn(context)
+    let cancelled = false
```
TkDodo: I think you need to make this type boolean so that eslint doesn’t complain

TkDodo: did that here: 46091df

PR #9968, opened 2025-12-14
fix(react-query/HydrationBoundary): prevent unnecessary refetch during hydration

### LINKED ISSUE #9610: HydrationBoundary double fetching on subsequent visits
### Describe the bug

I used `HydrationBoundary` in React Router. In `loader()`, I get the data, dehydrate it, then hydrate the client side query client. 

1. For the **first visit**, since it's a new query, it got hydrated in `React.useMemo`
2. For **subsequent visits**, it will be deferred to `React.useEffect`, however, if the cache is stale, query client will refetch before `useEffect`, which makes the server compute the data twice: one in loader, the other in the component

I looked into the source code and found that it's due to `refetchOnMount`, which is true by default. I think it would be plausible to prevent a refetch if the query in the dehydrated state is fresh.

There are some experiments in the provided code, e.g. `initialData` from [React Query meets React Router](https://tkdodo.eu/blog/react-query-meets-react-router), `initialData` + `initialDataUpdatedAt`, `HydrationBoundary`, and my custom hook use-hydrated-query which sets `refetchOnMount` to false for that query if not explicitly provided during hydration. You may navigate away and back for each experiment and see the log in console to reproduce the issue.

https://stackblitz.com/edit/vitejs-vite-yiggax3s

### Steps to reproduce

mentioned in 'Describe the bug'

### Expected behavior

If query is fresh from `loader()`, client side query client should not refetch

Every time

### Screenshots or Videos

_No response_

### Platform

Any browser

### Tanstack Query adapter

react-query

### TanStack Query version

v5.85.3

### TypeScript version

5.8.3

### Additional context

_No response_

sukvvon:
## 🎯 Changes

Fixes #9610

When using `HydrationBoundary` with existing queries in the cache, an unnecessary refetch was triggered during hydration even though fresh data was about to be hydrated from the server.

### Root Cause
- `useQuery`'s `onSubscribe` runs before `HydrationBoundary`'s `useEffect` completes hydration
- The query appears stale (before hydration), triggering an unnecessary refetch

### Solution
- Added `IsHydratingContext` (React Context) to track queries pending hydration
- Mark existing queries as pending in `useMemo` (before children render)
- Skip refetch in `queryObserver.#executeFetch` if query is pending hydration and data is fresh
- Exception: `refetchOnMount: 'always'` still triggers refetch (user explicitly wants fresh data)
- Exception: Stale data (based on `staleTime`) triggers refetch (cached markup scenario)
- Clear pending flag in `useEffect` after hydration completes

### Why React Context instead of global state?
- Supports multiple `QueryClient` instances (micro-frontends, testing)
- Proper isolation between different `HydrationBoundary` components
- Follows React patterns for state management

### Files Changed
- `packages/query-core/src/queryObserver.ts` - Check pending hydration and stale state in `#executeFetch`
- `packages/query-core/src/types.ts` - Add `_isHydrating` option type
- `packages/react-query/src/HydrationBoundary.tsx` - Mark/clear pending hydration queries with Context
- `packages/react-query/src/IsHydratingProvider.ts` - New Context provider for hydration state
- `packages/react-query/src/useBaseQuery.ts` - Set `_isHydrating` flag from Context
- `packages/react-query/src/useQueries.ts` - Set `_isHydrating` flag from Context
- `packages/react-query/src/index.ts` - Export `useIsHydrating` and `IsHydratingProvider`
- `packages/react-query/src/__tests__/HydrationBoundary.test.tsx` - Add 26 test cases
- `docs/framework/react/guides/ssr.md` - Document new behavior
- `docs/framework/react/reference/hydration.md` - Document new behavior

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Added a hydration tracking hook and provider to expose pending hydration state.

* **Bug Fixes**
  * Prevents unnecessary refetches during hydration when hydrated data is fresh.

* **Documentation**
  * Expanded hydration docs with server memory guidance and Next.js rewrite caveats; guidance on refetch behavior and staleTime.

* **Tests**
  * Extended hydration test suite with many scenarios to validate correct hydration/refetch behavior.

* **Chores**
  * Patch version bumps recorded.


tarasvarshava:
Thanks a lot for this. I hope it gets merged soon 🤞

docs/framework/react/reference/hydration.md
```diff
@@ -114,6 +114,8 @@ function App() {
 
 > Note: Only `queries` can be dehydrated with an `HydrationBoundary`.
 
+> Note: `HydrationBoundary` automatically prevents unnecessary refetching during hydration. Queries being hydrated will not trigger a refetch on mount, unless `refetchOnMount` is explicitly set to `'always'`.
```
Ephem: Thank you for this PR! I've only had time to skim it so far and it will take some time before I can dig in properly, but I wanted to provide some quick feedback here.

We also need to account for the case where there is some time in between fetching on the server, and hydrating on the client. This commonly happens when the markup is cached, in which case the hydration can happen _days_ after the fetch on the server. In this situation, we do want to render with the stale data initially (this is what the SSR pass did and we need to match it), but we also want to immediately refetch the data because it's stale.

packages/query-core/src/hydration.ts
```diff
@@ -14,6 +14,11 @@ import type { QueryClient } from './queryClient'
 import type { Query, QueryState } from './query'
 import type { Mutation, MutationState } from './mutation'
 
+// WeakSet to track queries that are pending hydration
+// Used to prevent double-fetching when HydrationBoundary defers hydration to useEffect
+export const pendingHydrationQueries: WeakSet<Query<any, any, any, any>> =
```
Ephem: I don't think we can use a global singleton to keep track of this, what if there are several `queryClient`s? In that case a query might just be about to hydrate into _one_ of these, not all of them.

Annoying as that is, I think we need to use a context for this. Note that there is some prior art with `isRestoring`.

sukvvon:
@Ephem Could you review this PR?

packages/react-query/src/HydrationBoundary.tsx
```diff
@@ -50,62 +51,92 @@ export const HydrationBoundary = ({
...
   // updating the UI.
-  const hydrationQueue: DehydratedState['queries'] | undefined =
-    React.useMemo(() => {
-      if (state) {
-        if (typeof state !== 'object') {
-          return
-        }
+  // Create a mutable ref object to hold pending query hashes
+  // This allows us to add/remove hashes without triggering re-renders
+  const hydratingQueriesRef = React.useRef<Set<string>>(new Set())
```
Ephem, line 56: I don't think this can be a ref unfortunately, it needs to be a state. It's being read in renders further down in context consumers, and it's not safe to read refs in render.

The reason is that with concurrent features, we want this to reflect the correct queries being hydrated _in this fork of the world_, we don't want to have the same state for the different concurrent realities so to speak.

packages/query-core/src/queryObserver.ts
```diff
@@ -96,7 +96,25 @@ export class QueryObserver<
...
+
+      const shouldSkipFetchForHydration =
+        this.options._isHydrating &&
+        resolvedRefetchOnMount !== 'always' &&
+        !isStale(this.#currentQuery, this.options)
+
+      if (
+        shouldFetchOnMount(this.#currentQuery, this.options) &&
+        !shouldSkipFetchForHydration
+      ) {
```
Ephem, line 117: There's also logic in `createResult` that needs to account for this. That's also used to determine what state the query _will be in_, the optimistic result that the query hooks see on first render, before this is called and the actual fetch (possibly) happens.

This gets into some other tricky territory (where we might already have other bugs tbh), which is that the first render needs to match the render on the server. If the data was not stale when server rendering, but is now stale, we need to:

* Do the first render as if it's not stale to avoid hydration mismatches
* Still trigger a refetch and update the query _after_ the first render

Ephem (commented):
I'm so sorry for the radio silence here, I've been in the middle of a long distance move and had to step away from OSS for a few months, I'm still pretty busy IRL but I'm now easing back into it.

I appreciate this contribution, it's trying to tackle something pretty gnarly. I'm not at all happy about the current state of RQ when it comes to these things.

The long term fix here is even bigger, which is to put the actual queries that are being hydrated on a context. The hooks would read from that context and kind of "patch" the current store state with the query state from the context in case it's newer. This would mean we better support concurrent features, as even if the query already exists in the store, the concurrent render would still immediately see the new data, which is not the case today. This approach would also fix a few other bugs we have.

I need to get back into the code a bit more before I can fully evaluate if this PR might be a pragmatic middle step that builds towards that "full" fix, but I think so.

sukvvon:
@Ephem First of all, my apologies for the long delay in following up here — thank you for the detailed review, it was really helpful in getting this right.

I've addressed both points:

**1. Pending hydration state via context, not a mutated ref** (your comment on `HydrationBoundary.tsx`)

You were right that a ref read in render isn't safe under concurrent features. Instead of mutating a shared `Set`, the pending set is now derived purely in render from `(client, state)`, so each render/fork carries its own immutable snapshot — no shared mutable object, nothing read from a ref during render. Consumers still read `.has(hash)`, but the value they read is a fresh per-render `Set`.

A nice consequence is that clearing is now automatic: once the deferred hydration lands and the cache's `dataUpdatedAt` matches the dehydrated data, the query is no longer classified as "newer", so it drops out of the pending set on the next render — no manual cleanup, and it stays correct for a late-mounting observer of the same key.

**2. `createResult` first render matching the server** (your comment on `queryObserver.ts`)

I reproduced the mismatch you described: when the server rendered non-stale but the client is now stale, the optimistic first render flipped to `isFetching: true`, which the server never rendered (dehydrated queries are always idle). The fix suppresses the optimistic fetch state while `_isHydrating`, so the first render matches the server regardless of staleness, and the actual refetch still fires afterwards in `onSubscribe` (unchanged) when the data is stale.

I've added tests covering both the pending-flag clearing and the first-render-not-fetching behavior. Would appreciate another look when you have time 🙏

packages/react-query/src/HydrationBoundary.tsx
```diff
@@ -50,62 +107,33 @@ export const HydrationBoundary = ({
...
+
+    return existingQueries.length > 0 ? existingQueries : undefined
+  }, [client, state])
+
+  // The pending set drives `_isHydrating` for consumers, which is read per
+  // observer at mount. It must be recomputed on every render (not memoized on
+  // `state` identity) so that once the deferred hydration lands and the cache
+  // catches up, the query is no longer classified as newer and the flag clears
+  // for any observer mounting afterwards.
+  const { pendingQueries } = classifyQueries(client, state)
```
VedAnt-1004, line 128: Nice work on tackling the hydration refetch issue—the test suite you added here is super thorough.

Caught a few things in `HydrationBoundary.tsx` that look like leftover merge artifacts:

1. **Syntax error in props interface**: It looks like `optionsRef.current = options` got pasted directly into the `HydrationBoundaryProps` interface declaration instead of the component body.
2. **Duplicate return statement**: The component has `return children` right before `return <IsHydratingProvider>`.
3. **Double cache scan**: `classifyQueries` is being called twice during render (once inside `useMemo` and again right below it for `pendingQueries`). Since both need the query classification, could we compute that once to avoid iterating over `state.queries` and querying `queryCache.get()` repeatedly?
4. **Stable empty set**: Reusing `EMPTY_HYDRATING_QUERIES` for the provider value is a nice touch to prevent unnecessary context re-renders when nothing is pending.

PR #9970, opened 2025-12-15
fix(query-core): prevent reducer from being called twice in streamedQuery

### LINKED ISSUE #9787: streamedQuery reducer being called twice
### Describe the bug

It seems the streamedQuery's reducer is being called twice, resulting in duplicated values.
Please see the relevant code:

https://github.com/TanStack/query/blob/ecdf96296b1dfcee617a5a5b9afdf2a42ad436ba/packages/query-core/src/streamedQuery.ts#L83C1-L89C38

```ts
      // don't append to the cache directly when replace-refetching
      if (!isRefetch || refetchMode !== 'replace') {
        context.client.setQueryData<TData>(context.queryKey, (prev) =>
          reducer(prev === undefined ? initialValue : prev, chunk),
        )
      }
      result = reducer(result, chunk)
```

As you can see, `reducer` gets called once in the `setQueryData` under special conditions, and then again at the end of the loop. This can mess up a stateful object in a custom reducer.

It's very simple

### Steps to reproduce

```ts
const arr = [];

queryOptions({
    queryKey: ["somekey"],
    queryFn: streamedQuery({
      streamFn: async function* (ctx) {
        const v = [1, 2, 3];
        yield* v;
      },
      reducer: (oldStream, newChunk) => {
        arr.push(newChunk);
        return [...oldStream, newChunk];
      },
      initialValue: [],
      refetchMode: "replace",
    }),
```

`arr.length == 6`

### Expected behavior

Reducer gets called once per chunk.

None

### Screenshots or Videos

_No response_

### Platform

Windows

### Tanstack Query adapter

None

### TanStack Query version

v5.90.5

### TypeScript version

_No response_

### Additional context

_No response_

HuijungYoon:
fixes #9787

fix(query-core): prevent reducer from being called twice in streamedQuery #9787

- Fix bug where reducer was called twice for each chunk (once in setQueryData, once in loop)
- Store reducer result in variable to avoid duplicate calls
- Add test case to verify reducer is only called once per chunk

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicated application of streamed updates during a "replace" refetch so displayed data remains stable and consistent.

* **Tests**
  * Added a test validating streamed refetch behavior to ensure stream chunks are applied once and overall data consistency is preserved.

* **Chores**
  * Added a changeset marking a release bump related to the bug fix.

### DIFF
--- a/.changeset/silver-peas-end.md
+++ b/.changeset/silver-peas-end.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/query-devtools': minor
+---
+
+improve a11y around screenreader + resizing
--- a/packages/query-devtools/src/Devtools.tsx
+++ b/packages/query-devtools/src/Devtools.tsx
@@ -417,6 +417,13 @@ const DraggablePanel: Component<DevtoolsPanelProps> = (props) => {
     return theme() === 'dark' ? darkStyles(css) : lightStyles(css)
   })
 
+  let closeBtnRef!: HTMLButtonElement
+
+  // Focus the close button when the panel opens for screen reader accessibility
+  onMount(() => {
+    closeBtnRef.focus()
+  })
+
   const [isResizing, setIsResizing] = createSignal(false)
 
   const position = createMemo(
@@ -480,7 +487,7 @@ const DraggablePanel: Component<DevtoolsPanelProps> = (props) => {
         setIsResizing(false)
       }
       document.removeEventListener('mousemove', runDrag, false)
-      document.removeEventListener('mouseUp', unsubscribe, false)
+      document.removeEventListener('mouseup', unsubscribe, false)
     }
 
     document.addEventListener('mousemove', runDrag, false)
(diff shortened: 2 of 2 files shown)


HuijungYoon:
- typo correction
- fix correct test case

pnpm-lock.yaml
TkDodo, line 1: please don’t commit 36k line logfile changes! You’ve likely used the wrong version of pnpm. if you didn’t change a dependency, don’t commit lock files!

TkDodo:
two existing tests are failing for streamedQuery after your change, please take a look

TkDodo:
the lockfile is again messed up. please check things locally, including all tests, before you get me back to this PR because this back and forth is a bit draining

HuijungYoon:
Hi @TkDodo,
   
   Sorry for all the back and forth. I've cleaned this up:

   pnpm-lock.yaml is now unchanged (matches upstream/main exactly)
   Only code changes included (streamedQuery.ts + test file)
   All CI tests passing

packages/query-core/src/streamedQuery.ts
TkDodo, line 1: I think this version complicates things a bit as it adds more conditions. I think the only exception to the flow we’d want is for replace-refetches, as they don’t write to the cache immediately, but only at the end. They should also always start with the initialValue. How about this:

```
const isReplaceRefetch = isRefetch && refetchMode === 'replace'

for await (const chunk of stream) {
  if (cancelled) {
    break
  }

  if (isReplaceRefetch) {
    // don't append to the cache directly when replace-refetching
    result = reducer(result, chunk)
  } else {
    context.client.setQueryData<TData>(context.queryKey, (prev) =>
      reducer(prev === undefined ? initialValue : prev, chunk),
    )
  }
}

if (isReplaceRefetch && !cancelled) {
  context.client.setQueryData<TData>(context.queryKey, result)
}

return context.client.getQueryData(context.queryKey) ?? initialValue
```

I think it achieves the same thing of only calling reducer once per branch, but makes it clear with which values we call it in which case, and also has a bit better  documentation by extracting `isReplaceRefetch` to a separate variable that we can then re-use for the finalizing. It also avoids negations in conditions which makes things easier to grasp.

HuijungYoon: That makes a lot more sense! The extracted `isReplaceRefetch` variable and the clearer if-else branches make the logic much easier to follow. I've updated the code with your suggested approach.

PR #10012, opened 2026-01-02
chore: prefer `React.use` instead of throwing promises [WIP]

juliusmarminge:
## 🎯 Changes

Migrate to use `React.use` when available instead of throwing promises. This adds support for the `suspended by` state in the React Devtools:

### Before

image

### After

`<add screenshot>`

Ref: https://discord.com/channels/719702312431386674/1446346759776370772

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/packages/react-query/src/useBaseQuery.ts
+++ b/packages/react-query/src/useBaseQuery.ts
@@ -1,7 +1,7 @@
 'use client'
 import * as React from 'react'
 
-import {  noop, notifyManager } from '@tanstack/query-core'
+import { noop, notifyManager } from '@tanstack/query-core'
 import { useQueryClient } from './QueryClientProvider'
 import { useQueryErrorResetBoundary } from './QueryErrorResetBoundary'
 import {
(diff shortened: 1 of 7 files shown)


packages/react-query/src/useBaseQuery.ts
```diff
@@ -125,7 +126,12 @@ export function useBaseQuery<
 
   // Handle suspense
   if (shouldSuspend(defaultedOptions, result)) {
-    throw fetchOptimistic(defaultedOptions, observer, errorResetBoundary)
+    const promise = fetchOptimistic(
+      defaultedOptions,
+      observer,
+      errorResetBoundary,
+    )
+    suspend(promise.finally(() => observer.updateResult()))
```
TkDodo: why isn’t this just `suspend(promise)` ? also, do we need to `return` after that ? I think `throw` usually stops execution of the rest?

TkDodo:
@juliusmarminge took me some time to get somewhere but I think I have something:

- https://github.com/TanStack/query/pull/10366

the key was to cache promises on observer level because `fetchOptimistic` returns a new promise every render

PR #10034, opened 2026-01-13
WIP: infer errors from Throws returns

shkumbinhasani:
## Summary
- infer hook error types from `Throws` return annotations when provided
- guard `any` and strip `Throws` to avoid pollution in data inference
- add type tests for auto inference and `Promise<any>` fallback

## Testing
- `pnpm --filter @tanstack/react-query test:types` (fails in `test:types:tscurrent` at `packages/react-query/src/__tests__/useQuery.test.tsx:2186`: NotifyOnChangeProps type mismatch)

## Release Notes

* **New Features**
  * Enhanced TypeScript error type inference in query and mutation operations for improved IDE autocomplete and compile-time type checking.
  * New type utilities for declaring and extracting error types from functions, enabling more precise error handling and type safety.

### DIFF
--- a/packages/react-query/src/useQuery.ts
+++ b/packages/react-query/src/useQuery.ts
@@ -3,9 +3,13 @@ import { QueryObserver } from '@tanstack/query-core'
 import { useBaseQuery } from './useBaseQuery'
 import type {
   DefaultError,
+  InferErrorFromFn,
   NoInfer,
   QueryClient,
+  QueryFunction,
   QueryKey,
+  StripThrows,
+  ThrowsFnOptions,
 } from '@tanstack/query-core'
 import type {
   DefinedUseQueryResult,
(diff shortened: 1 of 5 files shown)


packages/react-query/src/useQuery.ts
```diff
@@ -17,6 +21,54 @@ import type {
...
+  queryClient?: QueryClient,
+): DefinedUseQueryResult<NoInfer<TData>, InferErrorFromFn<TQueryFn>>
+
+export function useQuery<
+  TQueryKey extends QueryKey = QueryKey,
+  TQueryFn extends QueryFunction<any, TQueryKey> = QueryFunction<
+    any,
+    TQueryKey
+  >,
+  TQueryFnData = StripThrows<Awaited<ReturnType<TQueryFn>>>,
```
TkDodo, line 54: isn’t the big problem here that we would need to do this on _all_ functions that can produce `TQueryFnData` like `queryOptions`, `usePrefetchQuery`, the imperative methods on the `queryClient` etc ... ?

shkumbinhasani: Yeah, I think so. I only looked into it as a quick proof of concept and didn’t dig too deep, but I ended up dropping the idea because of a drawback: the returned type can’t reliably strip out the `& Throws<>` part. https://github.com/microsoft/TypeScript/issues/62985 — If you happen to have any ideas for a workaround (or a similar approach that avoids this issue), I’d love to hear them. I’d be really happy to see this come to life in some form.

PR #10037, opened 2026-01-14
fix(query-core): replaceEqualDeep max depth (v4)

TkDodo:


manudeli:
@TkDodo ci error is fixed now!
@lachlancollins Thanks for your resolving ci error

PR #10061, opened 2026-01-20
feat(eslint-plugin-query): add new rule for SSR QueryClient setup

jackleslie:
## 🎯 Changes

Resolves: https://github.com/TanStack/query/discussions/10060

New rule for ensuring that `QueryClient` isn't created in the root module scope (for Next.js applications)

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Added an ESLint rule that flags module-level QueryClient creation in Next.js projects, encouraging component-level instantiation (useState/useRef).

* **Documentation**
  * Added comprehensive guidance with multiple valid and invalid examples for Next.js usage.

* **Tests**
  * Added unit tests covering valid and invalid scenarios across Next.js file types and usage patterns.

* **Chores**
  * Added a changeset entry for a minor release.

### DIFF
--- a/packages/eslint-plugin-query/src/__tests__/no-module-query-client.test.ts
+++ b/packages/eslint-plugin-query/src/__tests__/no-module-query-client.test.ts
@@ -84,6 +84,17 @@ ruleTester.run('no-module-query-client', rule, {
       `,
     },
     {
+      name: 'QueryClient at module level in directory containing "app" as substring',
+      filename: 'myapp/components/Provider.tsx',
+      code: normalizeIndent`
+        import { QueryClient } from "@tanstack/react-query";
+
+        const queryClient = new QueryClient();
+
+        export { queryClient };
+      `,
+    },
+    {
       name: 'QueryClient in custom hook',
       filename: 'pages/custom-hook.ts',
       code: normalizeIndent`
--- a/packages/eslint-plugin-query/src/rules/no-module-query-client/no-module-query-client.rule.ts
+++ b/packages/eslint-plugin-query/src/rules/no-module-query-client/no-module-query-client.rule.ts
@@ -53,9 +53,12 @@ export const rule = createRule({
           return
         }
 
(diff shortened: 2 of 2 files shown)


packages/eslint-plugin-query/src/__tests__/no-module-query-client.test.ts
```diff
@@ -0,0 +1,218 @@
...
+    },
+    {
+      name: 'QueryClient created with React.useRef in pages directory',
+      filename: 'pages/_app.tsx',
+      code: normalizeIndent`
+        import { QueryClient } from "@tanstack/react-query";
+        import React from "react";
+
+        export default function MyApp({ Component, pageProps }) {
+          const queryClient = React.useRef(new QueryClient());
```
TkDodo, line 43: this needs to work, too:

```
const queryClient = React.useRef(null)
if (!queryClient.current) {
  queryClient.current = new QueryClient()
}
```

docs/eslint/no-module-query-client.md
```diff
@@ -0,0 +1,71 @@
...
+    <QueryClientProvider client={queryClient}>
+      <Component {...pageProps} />
+    </QueryClientProvider>
+  )
+}
+```
+
+## When Not To Use It
+
+This rule is specifically designed for Next.js applications. If you're not using Next.js or server-side rendering, this rule may not be necessary. The rule only runs on files that appear to be in a Next.js project structure (`pages/`, `app/`, `_app`, `_document`).
```
TkDodo, line 66: I think it’s too easy to hit false-positives here when someone has a client side rendered app that just happens to have a directory called `/pages` or `/app`.

Being limited / bound to `next.js` is also not ideal as there are other frameworks doing SSR.

Also, wouldn’t this fail in tests where queryClients might be created differently ? Not sure if tests can be in those directories though.

All in all, I think if we want this rule it should be opt-in and maybe needs to be able to customize locations to check ?

PR #10074, opened 2026-01-27
Fix/infinite query hydration 8825

### LINKED ISSUE #8825: Next.js: exception encountered in prefetchInfiniteQuery causes useSuspenseInfiniteQuery to fail

chosule:
Fixes #8825

> **Note**: This solution was discussed and approved by @TkDodo in [Discussion #10027](https://github.com/TanStack/query/discussions/10027)

## Problem
When `prefetchInfiniteQuery` fails on the server, `useSuspenseInfiniteQuery` on the client doesn't return the correct infinite query structure (`{ pages: [], pageParams: [] }`), causing a runtime error when accessing `data.pages`:
```
TypeError: Cannot read properties of undefined (reading 'length')
```

## Root Cause
Infinite query behavior information gets lost during hydration:

1. `infiniteQueryBehavior` is attached when `prefetchInfiniteQuery` runs on the server
2. This behavior info isn't preserved during dehydration
3. On the client, it gets restored as a regular query during hydration
4. Error occurs when data is accessed before `InfiniteQueryObserver` attaches the behavior

## Solution
Set up infinite query behavior at hydration time to guarantee correct data structure for both success and failure cases.

### 1. Added `queryType` field to `DehydratedQuery`
- Metadata to identify infinite queries
- Stores `'query' | 'infiniteQuery'` value

### 2. Auto-detect query type during dehydration
- Checks for `initialPageParam` presence to identify infinite queries
- Stores result in `queryType` field

### 3. Auto-configure behavior during hydration
- Attaches `infiniteQueryBehavior` when `queryType` is `'infiniteQuery'`

### 4. Handle failed promises
- Rejected promises aren't passed as `initialPromise`
- Allows automatic retry on the client

## Modified Files
1. `packages/query-core/src/hydration.ts`
2. `packages/query-core/src/__tests__/hydration.test.tsx`

## Tests 

**All 35 tests passing** (including 2 new tests)

New tests added:
- `should preserve queryType for infinite queries during hydration`
- `should attach infiniteQueryBehavior during hydration`

## Checklist
- [x] Added `queryType: 'query' | 'infiniteQuery'` field to `DehydratedQuery`
- [x] Auto-detect infinite queries during dehydration by checking `initialPageParam`
- [x] Auto-configure `infiniteQueryBehavior` for infinite queries during hydration
- [x] Added test cases for hydrated infinite query behavior
- [x] Guarantee correct data structure for both success and failure cases

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve infinite-query pagination and behavior during SSR hydration so hydrated clients retain pages, cursors, and correct subsequent fetch behavior.

* **Tests**
  * Added hydration tests to validate infinite query preservation, restoration, and post-hydration fetching.

* **Chores**
  * Added release metadata documenting this patch release.


packages/query-core/src/hydration.ts
```diff
@@ -239,16 +255,21 @@ export function hydrate(
           })
         }
       } else {
+        const queryOptions: any = {
+          ...client.getDefaultOptions().hydrate?.queries,
+          ...options?.defaultOptions?.queries,
+          queryKey,
+          queryHash,
+          meta,
+        }
```
TkDodo: can we keep this inlined please and just conditionally assign `behavior`:

```
behavior: queryType === 'infiniteQuery'
  ? (infiniteQueryBehavior() as QueryBehavior<unknown, unknown, unknown>)
  : undefined,
```

packages/query-core/src/hydration.ts
```diff
@@ -272,13 +293,24 @@ export function hydrate(
...
+        if (!isRejectedThenable) {
+          query
+            .fetch(undefined, {
+              // RSC transformed promises are not thenable
+              initialPromise: Promise.resolve(promise).then((resolvedData) => {
+                return deserializeData(resolvedData)
+              }),
+            })
+            // Avoid unhandled promise rejections
+            .catch(noop)
```
TkDodo: I don’t understand what this change has to do with infinite queries ?

chosule: @TkDodo  
Applied your feedback! 🙏

You're right that the rejected promise handling wasn't directly related to #8825. 
I've removed that part and inlined the behavior assignment as suggested.

Please take another look when you can!

chosule:
Hi @TkDodo, gentle ping! Let me know if there's anything else to address.

packages/query-core/src/hydration.ts
```diff
@@ -70,6 +72,11 @@ function dehydrateMutation(mutation: Mutation): DehydratedMutation {
   }
 }
 
+function isInfiniteQuery(query: Query): boolean {
+  const options = query.options as any
```
TkDodo: the `as any` seems super unnecessary


```suggestion
  const options = query.options
```

packages/query-core/src/hydration.ts
TkDodo: this is still not addressed

chosule: Removed `any` from `queryOptions` and changed the cast to `QueryBehavior<unknown, DefaultError, unknown>`. 
Is this the direction you had in mind?

packages/query-core/src/hydration.ts
chosule: Removed `as any` from `query.options`.

TkDodo:
I had a different idea today: what if we actually store the `type` on the `Query` class? In places where we now append the `infiniteQueryBehavior`, we could instead set `type: 'infiniteQuery'`.

then, we could just attach the behavior when the queryFn is run for every query that has an infinite type.

for hydration, all we would need to do is add the `type` field to the list of fields we hydrate / dehydrate then, and it should work ootb.

what do you think?

chosule:
> I had a different idea today: what if we actually store the `type` on the `Query` class? In places where we now append the `infiniteQueryBehavior`, we could instead set `type: 'infiniteQuery'`.
> 
> then, we could just attach the behavior when the queryFn is run for every query that has an infinite type.
> 
> for hydration, all we would need to do is add the `type` field to the list of fields we hydrate / dehydrate then, and it should work ootb.
> 
> what do you think?

Hi @TkDodo, thanks again for the suggestion — I tried your approach and refactored things a bit.
- Added type: 'query' | 'infiniteQuery' to Query
- Query.fetch() now applies infiniteQueryBehavior automatically when type === 'infiniteQuery'
- Callers now pass _type: 'infiniteQuery' as a hint instead of wiring behavior directly
- hydrate() restores that _type for infinite queries, so the type survives hydration
- Also added a regression test covering SSR -> dehydrate -> hydrate -> refetch

PR #10079, opened 2026-01-29
fix(react-query): retry useQuery().promise on error boundary reset

### LINKED ISSUE #9978: `experimental_prefetchInRender` does not work with `useQueryErrorResetBoundary`
### Describe the bug

I think there is something something missing with how errors are reset when using `useQuery().promise`. Which leads to the error boundary only being reset on the second try.

The problem is that you would use `useQuery().promise` differently from `useSuspenseQuery()`:

- With `useSuspenseQuery()`, you would have your `<Suspense>` and `<ErrorBoundary>` somewhere up the component tree from `useSuspenseQuery`:
  ```js
  const App = () => {
    const { reset } = useQueryErrorResetBoundary();
    return (
      <Suspense fallback={<span>Loading...</span>}>
        <ErrorBoundary 
          fallbackRender={({ resetErrorBoundary }) => <button type="button" onClick={resetErrorBoundary}>Retry</button>}
          onReset={reset}
        >
          <Page />
        </ErrorBoundary>
      </Suspense>
    );
  }

  const Page = () => {
    const { data } = useSuspenseQuery(/* ... */);

    // do something with the data
  }
  ```
- With `useQuery().promise`, you have `<Suspense>` and `<ErrorBoundary>` in the same component as `useQuery()`:
  ```js
  const App = () => <Page />;

  const Page = () => {
    const promiseData = useQuery({ experimental_prefetchInRender: true, /* ... */ }).promise;
    const { reset } = useQueryErrorResetBoundary();

    return (
      <Suspense fallback={<span>Loading...</span>}>
        <ErrorBoundary 
          fallbackRender={({ resetErrorBoundary }) => <button type="button" onClick={resetErrorBoundary}>Retry</button>}
          onReset={reset}
        >
          <MySubComponent promiseData={promiseData} />
        </ErrorBoundary>
      </Suspense>
    );
  }

  const MySubComponent = ({ promiseData }) => {
    const data = React.use(promiseData);

    /* do something with the data */
  }
  ```

The problem is that the query will not be retried if you click the retry-button in the second example. I think the problem is that the current mechanism relys on the component that contains `useQuery()` (or `useSuspenseQuery` for that matter) to re-mount. Which of course does not happen in the second example.

https://codesandbox.io/p/sandbox/admiring-bas-xgdrjg

### Steps to reproduce

1. Open Codesandbox
2. Wait for the requests to finish: Both show a retry button
3. Click retry button in first line (useSuspenseQuery())
4. Click retry button in second line (useQuery().promise

### Expected behavior

Both retry button should work the same: When you click them, the query is retried.
This, however, does only work for the `useSuspenseQuery` option. Not the `useQuery().promise`.

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browsers: any that I tried (Firefox, Safari, Edge)

### Tanstack Query adapter

react-query

### TanStack Query version

5.90.12

### TypeScript version

_No response_

### Additional context

_No response_

zaewc:
## 🎯 Changes

Fixes https://github.com/TanStack/query/issues/9978

Recreate the query promise when a retry transitions the query to fetching, ensuring useQuery().promise provides a fresh pending promise to Suspense.

Explicitly refetch errored queries on QueryErrorResetBoundary reset, even when useQuery is outside the ErrorBoundary.

Limit refetching to active and enabled queries to avoid unintended retries.

## ✅ Checklist

## 🚀 Release Impact

## Release Notes

**Bug Fixes**
* Fixed `useQuery().promise` behavior when used with `QueryErrorResetBoundary` to ensure queries retry correctly when the boundary resets, and promises are properly recreated for Suspense.
* Improved error boundary reset to explicitly refetch queries in errored state when a reset is triggered.


packages/react-query/src/QueryErrorResetBoundary.tsx
```diff
@@ -12,14 +15,22 @@ export interface QueryErrorResetBoundaryValue {
...
     reset: () => {
       isReset = true
+      void client?.refetchQueries({
+        predicate: (query) =>
+          query.state.status === 'error' &&
+          query.getObserversCount() > 0 &&
+          query.observers.some(
+            (observer) => observer.options.enabled !== false,
+          ),
+      })
```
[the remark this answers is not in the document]
zaewc: [2f9a7fe](https://github.com/TanStack/query/pull/10079/commits/2f9a7fe07b80408b858893c292e835b92376ef4b)

I've updated the logic to use `resolveEnabled` to correctly evaluate callback-based enabled options as suggested

packages/react-query/src/QueryErrorResetBoundary.tsx
```diff
@@ -12,14 +17,23 @@ export interface QueryErrorResetBoundaryValue {
...
       isReset = true
+      void client?.refetchQueries({
+        predicate: (query) =>
+          query.state.status === 'error' &&
+          query.getObserversCount() > 0 &&
+          query.observers.some(
+            (observer) =>
+              resolveEnabled(observer.options.enabled, query) !== false,
+          ),
+      })
```
TkDodo: what is this doing? When we reset a boundary, the query in question likely doesn’t have subscribed observers. Also, observerCount and enabled is checked internally by `refetchQueries`. All I can see this doing is that it might refetch unrelated queries ?

zaewc: In the `useQuery().promise` pattern, the `useQuery` hook is often called in a parent component, while the promise is consumed in a child component. when the child throws an error the child unmounts, but the parent (containing the useQuery observer) stays mounted. therefore, an observer still exists. we check this to avoid refetching garbage queries that have no active consumers.

While `refetchQueries` has some internal logic, calling it without specific filters defaults to refetching matching queries regardless of whether they are enabled or have observers. by adding the enabled check in the predicate, we prevent disabled queries (which were intentionally stopped by the developer) from being accidentally restarted by a boundary reset. this was a regression I noticed during local testing in the existing test suite: 'should not throw error if query is disabled'.

zaewc: > that it might refetch unrelated queries

you’re right that this effectively becomes a global reset within the QueryClient, refetching all errored queries that are currently active and enabled. 

This isn’t perfect, but it matches the intended behavior of `useQueryErrorResetBoundary().reset()`, which is to clear error state within the current rendering context. Since TanStack Query doesn’t have a reliable way to determine which queries are actually scoped to a specific React Error Boundary, this felt like the safest and most predictable approach for ensuring `useQuery().promise` retries correctly

That said, I’m definitely open to better ideas here if there’s a more precise or elegant way to scope this

zaewc:
@TkDodo 

after reading your feedback, I had a new idea and reworked the approach. I've updated the logic to use a registry-based tracking system with queryHash-tagged promises.

Could you please take another look ?

TkDodo:
https://github.com/TanStack/query/issues/9978#issuecomment-5326017513

PR #10137, opened 2026-02-15
fix(query-core): handle combine throwing in QueriesObserver notify

### LINKED ISSUE #10129: `useSuspenseQueries` `combine` is called with pending/error results despite types implying data is always defined
### Describe the bug

The `combine` function in `useSuspenseQueries` has types that narrow `data` to always be defined (non-`undefined`), implying it should only be called when all queries have successfully resolved. However, `combine` is actually invoked via `QueriesObserver.#combineResult` in the `#notify` path even when query results are in `pending` or `error` state (e.g., after `queryClient.resetQueries()` or when a query errors out).

This causes a runtime `TypeError` because user code in `combine` reasonably accesses `.data` without null checks — as the types suggest it is safe to do so.

https://codesandbox.io/p/devbox/wonderful-pine-74lmth?workspaceId=ws_QZPvCEAK42bpnZ7RJHDjaX

### Steps to reproduce

1. Open the reproduction link above.
2. Confirm that `loading...` is displayed on the initial load.
3. Open the console in the Preview panel.
4. In DevTools, select a query and trigger loading or trigger error. Or click the `reset query` button.
5. `Uncaught TypeError: Cannot read properties of undefined (reading 'full_name')` error is thrown.

### Expected behavior

One of the following:

1. If `combine` is intended to only receive successful results, `combine` should not be called when any query result is in a non-`success` state. The observer notification path should skip the combine optimization and let the framework render path handle suspension or error boundaries.

2. If `combine` is intended to handle all states, The type signature of `combine` in `useSuspenseQueries` should reflect that `data` can be `undefined`, matching the runtime behavior.

Always

### Screenshots or Videos

Image

### Platform

- OS: macOS
- Browser: Chrome
- Version: 144.0.7559.97

### Tanstack Query adapter

react-query

### TanStack Query version

v5.90.21

### TypeScript version

v5.8.3

### Additional context

React Version: v19.2.14

veeceey:
Fixes #10129

When `useSuspenseQueries` is used with `combine`, the types narrow `data` to always be defined. But if a query transitions back to pending/error state (e.g. via `queryClient.resetQueries()`), the `combine` function gets called with results where `data` is `undefined`, causing a runtime TypeError.

This happens because `QueriesObserver.#notify()` calls `combine` as an optimization check — to see if the combined result changed before notifying listeners. The listeners themselves receive raw results (not the combined one), so the combine call is purely for deduplication.

The fix wraps the combine call in a try/catch. If combine throws, we treat it as a change and notify listeners anyway. This lets the framework-level code (React Suspense / Error Boundary) handle the state transition properly.

Added a test that reproduces the exact scenario: subscribe to a `QueriesObserver` with a combine function that accesses `data` properties, let queries resolve, then call `resetQueries` — previously this would throw, now it correctly notifies listeners with the pending result.

## Summary by CodeRabbit

* **Bug Fixes**
  * Observers now reliably notify listeners even when a data-combining step throws during result evaluation.

* **Tests**
  * Added test validating notification behavior when queries are reset and combined-data assumptions fail.

* **Chores**
  * Added a changeset documenting the fix and bumping the patch version.

### DIFF
--- a/.changeset/brave-tigers-wave.md
+++ b/.changeset/brave-tigers-wave.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/query-core': patch
+---
+
+fix: handle combine throwing in QueriesObserver notify
--- a/packages/query-core/src/__tests__/queriesObserver.test.tsx
+++ b/packages/query-core/src/__tests__/queriesObserver.test.tsx
@@ -630,4 +630,43 @@ describe('queriesObserver', () => {
       { status: 'success', data: 3 },
     ])
   })
+
+  test('should still notify listeners when combine throws after query reset', async () => {
+    const key1 = queryKey()
+    const queryFn1 = vi.fn().mockReturnValue({ name: 'test' })
+
+    const combine = vi.fn(
+      (results: Array<QueryObserverResult>) => {
+        // This simulates a combine function that assumes data is always defined
+        // (like useSuspenseQueries types suggest)
+        return results.map((r) => (r.data as { name: string }).name)
+      },
+    )
+
+    const observer = new QueriesObserver<Array<string>>(
+      queryClient,
+      [{ queryKey: key1, queryFn: queryFn1 }],
+      { combine },
+    )
+
+    const results: Array<Array<QueryObserverResult>> = []
+    const unsubscribe = observer.subscribe((result) => {
+      results.push(result)
+    })
+
+    // Wait for queries to resolve
+    await vi.advanceTimersByTimeAsync(0)
+
+    // Reset the query - this transitions it to pending state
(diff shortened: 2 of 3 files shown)


themavik (commented):
test

themavik (commented):
QueriesObserver#notify: bare catch means any throw from #combineResult schedules listener notify (right when user combine throws under suspense) but also swallows unexpected throws inside #combineResult itself. nit: narrow the catch or rethrow after notify if you need to separate user combine bugs from internal failures.

TkDodo:
try /catch is not the correct way to address this bug

PR #10153, opened 2026-02-19
fix(query-core): decouple timers, retry and gc defaults from isServer guard

### LINKED ISSUE #10139: Tanstack Query refetch does not work in certain environments (vscode extension, chrome extension)
## Issue
This isServer check prevents the queries from running in these environments, mistakenly seeing them as servers anywhere globalThis.window is undefined.

https://github.com/TanStack/query/blob/984bfa6f15271c0ba7c750d35daff2f6fb596667/packages/query-core/src/utils.ts#L89

[Related Discussion](https://github.com/TanStack/query/discussions/4018)

### Workarounds
- Use queryClient.invalidateQueries in a `setInterval`
- [Maybe?] Put a "dummy" item into globalthis.window so that it isn't seen as a server. WARNING: This could have major unexpected side effects.

### Proposed Fix
([by @TkDodo](https://github.com/TanStack/query/discussions/4018#discussioncomment-3413849))
Decouple setInterval calls from isServer checks. None if this code runs on the server anyways, because subscriptions are only created on the client (they run in effects / useSyncExternalStore, so not on the server)

See above

### Steps to reproduce

See above

### Expected behavior

See above

Every time

### Screenshots or Videos

_No response_

### Platform

Chrome extensions and vscode extensions

### Tanstack Query adapter

vanilla

### TanStack Query version

v5.90.20

### TypeScript version

v5.8.3

### Additional context

_No response_

grzdev:
Fixes #10139

## Problem

`isServer` (`typeof window === 'undefined' || 'Deno' in globalThis`) is used to gate runtime features like:

- stale timeouts
- refetch intervals
- retry defaults
- garbage collection defaults

This incorrectly disables those features in **windowless client environments**
(e.g. VSCode extensions, Chrome extensions, Electron, Workers), which are not SSR
but do not define `window`.

As a result:
- `refetchInterval` does not run
- `staleTime` does not flip to stale
- default `retry` becomes `0`
- default `gcTime` becomes `Infinity`

## Fix

Remove `isServer` guards from runtime paths that are only created via
client-side subscriptions (effects / `useSyncExternalStore`),
so they never execute during SSR anyway.

Changes:

- `queryObserver.ts`
  - Remove `isServer` from `#updateStaleTimeout()`
  - Remove `isServer` from `#updateRefetchInterval()`

- `retryer.ts`
  - Default `retry` to `3` unconditionally instead of `isServer ? 0 : 3`

- `removable.ts`
  - Default `gcTime` to `5 * 60 * 1000` unconditionally instead of
    `isServer ? Infinity : 5 * 60 * 1000`

Not modified:
- `focusManager.ts`
- `onlineManager.ts`

Those `isServer` checks guard `window.addEventListener` usage and are correct.

## Tests

Added tests proving behavior in a `window`-undefined environment:

- refetchInterval fires when `window` is undefined
- stale timeout flips `isStale`
- retry defaults to `3`
- gcTime defaults to 5 minutes

All new tests pass.

### DIFF
--- a/packages/query-core/src/__tests__/queryObserver.test.tsx
+++ b/packages/query-core/src/__tests__/queryObserver.test.tsx
@@ -1536,6 +1536,71 @@ describe('queryObserver', () => {
     expect(observer.getCurrentResult().data).toBe('updated')
   })
 
+  test('should refetchInterval when window is undefined (non-windowed environments)', async () => {
+  const originalWindow = (globalThis as any).window
+  ;(globalThis as any).window = undefined
+
+  try {
+    const key = queryKey()
+    let count = 0
+
+    const fetchData = () => {
+      count++
+      return Promise.resolve('data')
+    }
+
+    const observer = new QueryObserver(queryClient, {
+      queryKey: key,
+      queryFn: fetchData,
+      refetchInterval: 10,
+    })
+
+    const unsubscribe = observer.subscribe(() => undefined)
+
+    // initial fetch
+    expect(count).toBe(1)
+
+    // should refetch even though isServer would be true
+    await vi.advanceTimersByTimeAsync(10)
+    expect(count).toBe(2)
+
+    unsubscribe()
+  } finally {
+    ;(globalThis as any).window = originalWindow
+  }
+  })
+
+  test('should schedule staleTimeout when window is undefined (non-windowed environments)', async () => {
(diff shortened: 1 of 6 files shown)


TkDodo:
please see:

- https://github.com/TanStack/query/issues/10139#issuecomment-3967376409

PR #10159, opened 2026-02-21
Prevent useQuery from skipping hydration

### LINKED ISSUE #10145: HydrationBoundary ignores/skips server-prefetched detail query when a parent useQuery with same key is rendered before child suspense fetch
### Describe the bug

We have an id-based route where parent/global components can see the `id` from URL and call `useQuery` for the detail resource, but suspended/SSR-safe child components use `useSuspenseQuery` for the same query key. The render of the `useQuery` causes the hydration boundary to skip hydration for the child using the `useSuspenseQuery` and triggers a guard we have to ensure we do not client fetch on the server during SSR. This also means the request is being duplicated and firing twice (in our prefetch and during SSR, which should have been skipped)

Flow:
1. Parent/route boundary prefetches the detail query on server.
2. A higher component in layout/header still renders `useQuery` with `enabled: Boolean(id)`.
3. On SSR, `HydrationBoundary` does not hydrate/replace the prefetched query as expected.
4. Child `useSuspenseQuery` then attempts a fetch with client requestor during SSR and hits the guard error.

### Repro
Server prefetch at detail layout/page:

```ts
// In app/(dashboard)/todos/[todoId]/layout.tsx or page.tsx
const queryClient = getQueryClient();
void queryClient.prefetchQuery(buildTodoQueryOptions(todoId, makeServerRequest));
return <HydrationBoundary state={dehydrate(queryClient)}>{children}</HydrationBoundary>;
```

Global/header client component that also reads URL `todoId` (lives OUTSIDE/above the above layout.tsx)

```ts
"use client";

import { useQuery, useParams } from "@tanstack/react-query";

function TodoHeaderMenu() {
  const params = useParams();
  const todoId = typeof params.todoId === "string" ? params.todoId : "";

  const { data: todo } = useQuery({
    ...buildTodoQueryOptions(todoId),
    enabled: Boolean(todoId),
  });

  return <span>{todo?.title}</span>;
}
```

Child detail component:

```tsx
function TodoPage({ todoId }: { todoId: string }) {
  const { data: todo } = useSuspenseQuery(buildTodoQueryOptions(todoId));
  return <TodoDetails todo={todo} />;
}
```

`buildTodoQueryOptions` defaults to client requestor. Client-request guard throws:

```ts
if (typeof window === "undefined") {
  throw new Error(
    "Warning, aborting SSR. You attempted to fetch from the server with clientRequest."
  );
}
```

AI disclosure: I had codex help me with the quick repro.

### Questions
1. Is this current behavior expected?
2. Should hydration logic be changed to only skip queries that are fetching (aka don't skip `useQueries` that fill the cache key on the server but do not initiate filling the value on the server)

https://codesandbox.io/p/devbox/pr7c4g

### Steps to reproduce

The home page on that will bring up the error in the next dev error popup

### Expected behavior

I would expect a `useQuery` in a higher component with the same key to not STOP the server fetch/hydration given `useQuery` would not trigger the client request guard like `useSuspenseQuery` does. `useQuery` does not start the `queryFn` on the server, so it should not stop `useSuspenseQueries` that CAN imo.

This also comes with the additional problem that because of the `useQuery`, the `useSuspenseQuery` is firing again, ruining deduplication. 

Every time

### Screenshots or Videos

_No response_

### Platform

macOS

Chromium

next: 16.1.1

### Tanstack Query adapter

react-query

### TanStack Query version

5.90.21

### TypeScript version

_No response_

### Additional context

I can technically gate the `useQuery` in the parent component with something like

```ts
const isDealQueryReady = !!(
    state &&
    state.data !== undefined &&
    state.status !== "pending" &&
    state.fetchStatus !== "fetching"
  )

//and then 
if (!isDealQueryReady) {
  return null;
}

//...

//...`useQuery`
```

and then I do not get the error but that is obnoxious.

ColemanDunn:
## 🎯 Changes

Fixes https://github.com/TanStack/query/issues/10145

Hydration is being skipped for useQueries that add to the cache above a `HydrationBoundary` but will actually not run or hydrate the queryClient. 

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Hydration now includes idle/pending queries to avoid unnecessary refetches or suspense-triggered reloads.

* **New Features**
  * Added two public APIs: defaultShouldDehydrateQuery and useSuspenseQuery.

* **Tests**
  * Added coverage for idle-query hydration behavior inside Suspense boundaries.

* **Chores**
  * Published patch updates for two TanStack libraries.


TkDodo:
@ephem can you take a look here please

@colemanDunn please fix the conflicts

ColemanDunn:
> @ColemanDunn please fix the conflicts

Done! Thank you!

ColemanDunn:
There were some new considerations and tests I added in for the new `syncData` changes in the merge conflict: https://github.com/TanStack/query/pull/10159/changes/ac2716703aa24d6fdd913b3e17d0557c45ef4a8f

packages/query-core/src/hydration.ts
```diff
@@ -290,11 +295,12 @@ export function hydrate(
...
         promise &&
-        // If the data was synchronously available, there is no need to set up
-        // a retryer and thus no reason to call fetch
-        !syncData &&
-        !existingQueryIsPending &&
-        !existingQueryIsFetching &&
+        (existingQueryIsIdleUseQuery ||
+          // If the data was synchronously available, there is no need to set up
+          // a retryer and thus no reason to call fetch
+          (!syncData &&
```
Ephem: I think this logic is wrong. If `existingQueryIsIdleUseQuery` is true, we'll always call fetch, even if data was synchronously available. This causes a regression where the query quickly flashes the loading state when it shouldn't.

Ephem (commented):
Thanks for trying to tackle this! I have to think some more about this, but I think this fix might be a bit too simplistic.

```tsx
query?.state.dataUpdatedAt === 0 &&
query.state.status === 'pending' &&
query.state.fetchStatus === 'idle'
```

This part is a heuristic for trying to determine _when it is safe to hydrate, even if the query is already in the cache_. Unfortunately, there is no such time. First, there are more situations than the reproduction itself that can cause this set of states, and it's hard to foresee all the possible side effects.

Second, imagine this case:

```tsx
<UseQueryComponent>
  <HydrationBoundary>
    <Suspense>
      <UseSuspenseQueryComponent>
```

Anything in the Suspense can hydrate later than the `<UseQueryComponent>`. By that time, the observer in the `<UseQueryComponent>` will have been set up, so if we hydrate into the cache, we are now triggering an _observed_ side effect in render, which is not safe.

The real problem here is that different subtrees _should_ be able to see different data for a single query during hydration. This is not possible with the current approach, which is why this bug has gone unfixed for so long.

The planned fix for this is to avoid hydrating in render entirely. Instead we put the data to be hydrated on a context, meaning different subtrees can see different data during hydration, and a single subtree can always see the data that it was rendered with on the server. We then make all hooks read from that hydration context when necessary. We then hydrate in a `useEffect` at the end and remove the context data after hydration.

This separates UI reads during hydration from the cache, which is also a prerequisite for better concurrent rendering support, and it also fixes some other edge cases we have.

All that said, if we can find a _safe_ quick fix for the issue at hand, I'm all for it, but I'm not certain yet this is one, and I'm not sure what it would take to feel safe since this is so complex to reason about. 😄

---

Note that a `useQuery` before the point of hydration should always be expected to sometimes start its own fetch on the client, even after a fix, imagine:

```tsx
<UseQueryComponent>
  <Suspense>
    <UnrelatedComponentThatSuspends>
      <HydrationBoundary>
        <UseSuspenseQueryComponent>
```

In this case, which might look contrived but is not uncommon in complex apps, the unrelated component makes it so the `<UseQueryComponent>` has time to start a fetch on the client before we even hit hydration. Still something we need to fix ofc, just thought I'd mention another footgun of the pattern. 😄

ColemanDunn:
Thanks for the reply @Ephem! Very well put and makes sense. 

The pattern with this shape:

```ts
<UseQueryComponent>
  <HydrationBoundary>
    <Suspense>
      <UseSuspenseQueryComponent>
```

has been something that is pretty common in projects I have worked on (think breadcrumbs that use `useQuery` and depend on something like `useParams` in a parent layout) so I appreciate being able to find a safe fix for this. I'll look into more options and write tests to ensure safety as much as we can for this change. 

Are there any specific tests that should be added to this PR that that can can ensure the fix is safe? Perhaps one for this case:

```ts

<UseQueryComponent>
  <Suspense>
    <UnrelatedComponentThatSuspends>
      <HydrationBoundary>
        <UseSuspenseQueryComponent>

```

Thanks!

packages/query-core/src/hydration.ts
ColemanDunn: ThanksI I overcorrected on that follow up commit after the merge. I have updated the code and added a test to check the loading state is in the correct state. 

I know you mentioned it is hard to foresee the side effects, but I hope there are some tests we can add to determine if we can find a safe enough fix for this.

Ephem:
@ColemanDunn Thanks for keeping a constructive attitude in the face of my pessimism! 😅

---

An aside on a possible workaround:

I've run into this case myself, where I feel I need a `useQuery` in a place above where I can prefetch said query. I ended up solving it by setting up a manual cache subscription to the store, along these lines:

```tsx
  return useSyncExternalStore(
    useCallback(
      (onStoreChange: () => void) =>
        cache.subscribe(event => {
          if (
            event.query.queryKey.length === queryKey.length &&
            event.query.queryKey.every(
              // The === check here will only work for primitives, to support complex keys one would need to deep compare
              (k: unknown, i: number) => k === queryKey[i],
            )
          ) {
            onStoreChange();
          }
        }),
      [cache, queryKey],
    ),
    () => queryClient.getQueryData(queryKey),
    () => queryClient.getQueryData(queryKey),
  );
```

Maybe that can be a workaround for you to until this is fixed?

---

Back to the PR. Thinking about this some more, the thing I worry most about is probably the "hydrate in render after a store subscription has already been set up" case. You could likely get around that by also checking explicitly for whether the query has an active listener by using `query.isActive()`, and bailing out from hydrating in render if it does. 🤔

I haven't had time to think very deeply about this, no promises it makes the PR safe enough to merge as a quick fix etc, but I think it definitely helps!

TkDodo:
@ColemanDunn please fix conflicts in this PR.
@Ephem what’s your latest take? Is this mergable / needs more work done / should be closed? Trying to do some housekeeping here 😅

Ephem:
@TkDodo Well, it's hard to say. I still think the correct fix is this:

> The planned fix for this is to avoid hydrating in render entirely. Instead we put the data to be hydrated on a context, meaning different subtrees can see different data during hydration, and a single subtree can always see the data that it was rendered with on the server

As we both know, this is tricky for a bunch of reasons though and we/I haven't made progress towards that for a long time.

In general I dislike the idea of adding more workarounds though, it has already led to pretty complex code that tends to create new bugs, creating new workarounds etc, but as I put it earlier:

> All that said, if we can find a safe quick fix for the issue at hand, I'm all for it, but I'm not certain yet this is one, and I'm not sure what it would take to feel safe since this is so complex to reason about. 😄

I think this PR still doesn't tackle this case:

> Anything in the Suspense can hydrate later than the `<UseQueryComponent>`. By that time, the observer in the `<UseQueryComponent>` will have been set up, so if we hydrate into the cache, we are now triggering an observed side effect in render, which is not safe.

@ColemanDunn This PR has been sitting with some unaddressed feedback for a while, so I'll close it for housekeeping, but feel free to re-open if you keep working on it. 😄 It's totally fine it's gone unaddressed btw, especially since I'm being vague on if addressing the comments will even get it to a mergable state, sorry about that. 😬

Either way, thank you for trying to tackle it! 🙏

PR #10186, opened 2026-02-25
feat(experimental_createQueryPersister): Add removeQueries

rklomp:
## 🎯 Changes

This adds a `removeQueries` method to the `experimental_createQueryPersister`
See also https://github.com/TanStack/query/discussions/9189

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Added ability to remove persisted queries from storage with optional query-key filters (supports exact and partial matches).

* **Documentation**
  * Added API docs describing the new query-removal utility.

* **Tests**
  * Added test coverage for query removal (filters, exact/partial matches); a duplicate test suite was unintentionally introduced.

* **Chores**
  * Bumped package version to reflect the new public API.

### DIFF
--- a/packages/query-persist-client-core/src/createPersister.ts
+++ b/packages/query-persist-client-core/src/createPersister.ts
@@ -131,13 +131,7 @@ export function experimental_createQueryPersister<TStorageValue = string>({
       try {
         const storedData = await storage.getItem(storageKey)
         if (storedData) {
-          let persistedQuery: PersistedQuery
-          try {
-            persistedQuery = await deserialize(storedData)
-          } catch {
-            await storage.removeItem(storageKey)
-            return
-          }
+          const persistedQuery = await deserialize(storedData)
 
           if (isExpiredOrBusted(persistedQuery)) {
             await storage.removeItem(storageKey)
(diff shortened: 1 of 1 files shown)


packages/query-persist-client-core/src/createPersister.ts
```diff
@@ -301,12 +313,53 @@ export function experimental_createQueryPersister<TStorageValue = string>({
...
+        if (key.startsWith(storageKeyPrefix)) {
+          if (!queryKey) {
+            await storage.removeItem(key)
+            continue
+          }
+
+          let persistedQuery: PersistedQuery
+          try {
+            persistedQuery = await deserialize(value)
+          } catch {
```
DamianOsipiuk, line 322: I see you have added `await storage.removeItem(key)` to the other method, yet not here.

rklomp: Ah good catch. I have added it there as well and also added the same logic to `retrieveQuery`. Not sure if this is really in scope of this change or we should move that to a different change though?

rklomp: I moved all changes outside the scope of this new function to a separate pull request: #10190

PR #10190, opened 2026-02-26
feat(experimental_createQueryPersister): removeItem if deserialize fails

rklomp:
## 🎯 Changes

We can remove the item from storage if deserialize fails as its not usable anymore.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience of persisted query cache: unreadable or corrupted stored entries are now removed automatically during retrieval, cleanup, and restoration to prevent failures.
  * Retrieval now aborts and clears a problematic stored entry when it cannot be decoded, avoiding stale errors.
  * Cleanup and restoration processes now skip and remove problematic entries so overall recovery is more stable and reliable.

### DIFF
--- a/.changeset/hungry-adults-count.md
+++ b/.changeset/hungry-adults-count.md
@@ -1,5 +1,5 @@
 ---
-'@tanstack/query-persist-client-core': minor
+'@tanstack/query-persist-client-core': patch
 ---
 
 removeItem if deserialize fails


.changeset/hungry-adults-count.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tanstack/query-persist-client-core': minor
```
DamianOsipiuk: This should be a `patch` change

rklomp: Changed it :)

PR #10198, opened 2026-02-28
isServer

### LINKED ISSUE #10139: Tanstack Query refetch does not work in certain environments (vscode extension, chrome extension)

TkDodo:
fixes #10139

### DIFF
--- a/packages/vue-query/src/__tests__/queryClient.test.ts
+++ b/packages/vue-query/src/__tests__/queryClient.test.ts
@@ -4,6 +4,14 @@ import { QueryClient as QueryClientOrigin } from '@tanstack/query-core'
 import { QueryClient } from '../queryClient'
 import { infiniteQueryOptions } from '../infiniteQueryOptions'
 
+type QueryClientPrototypeMethod = {
+  [TKey in keyof typeof QueryClientOrigin.prototype]: (typeof QueryClientOrigin.prototype)[TKey] extends (
+    ...args: Array<any>
+  ) => any
+    ? TKey
+    : never
+}[keyof typeof QueryClientOrigin.prototype]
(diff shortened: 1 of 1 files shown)


rootInfluence:
# GitHub Comment

The `isServer` flag helps determine the runtime environment so TanStack Query can apply appropriate behavior for server vs. client contexts (e.g., skipping certain operations on the server or handling hydration differently). You can check the current implementation in the `QueryClient` configuration or the relevant platform detection utilities to see how it's being used across the codebase.

PR #10218, opened 2026-03-04
fix(react-query): add experimental_prefetchInRender support to useQueries

### LINKED ISSUE #9988: `experimental_prefetchInRender` doesn't work with `useQueries`
### Describe the bug

```ts
const query = queryOptions({
  queryKey: ['foo'],
  queryFn: async () => [],
  experimental_prefetchInRender: true,
});

const Example: FC = () => {
  // ✅ This finishes loading.
  // const { promise } = useQuery(query);

  // ❌ This never finishes loading.
  const [{ promise }] = useQueries({ queries: [query] });

  React.use(promise);

  return <div>loaded</div>;
};
```

https://stackblitz.com/edit/tanstack-query-5j76gubf?file=src%2Findex.tsx&preset=node

### Steps to reproduce

See above.

### Expected behavior

`experimental_prefetchInRender` works with `useQueries`.

None

### Screenshots or Videos

_No response_

### Platform

N/A

### Tanstack Query adapter

None

### TanStack Query version

5.90.12

### TypeScript version

_No response_

### Additional context

_No response_

isaackaara:
## Problem

`experimental_prefetchInRender` works with `useQuery` but not with `useQueries`. When using `React.use(promise)` with `useQueries`, the component never finishes loading because the promise never resolves.

```tsx
const query = queryOptions({
  queryKey: ['foo'],
  queryFn: async () => [],
  experimental_prefetchInRender: true,
});

// ✅ Works with useQuery
const { promise } = useQuery(query);
React.use(promise);

// ❌ Never finishes loading with useQueries
const [{ promise }] = useQueries({ queries: [query] });
React.use(promise);
```

## Root Cause

`useBaseQuery` implements `experimental_prefetchInRender` by detecting when a query will fetch and either:
1. Triggering an optimistic fetch for new cache entries, or
2. Subscribing to the existing cache promise

When the promise resolves, it calls `observer.updateResult()` which finalizes the `currentThenable`.

`useQueries` was missing this logic entirely — the `experimental_prefetchInRender` option was ignored.

## Fix

Added the same prefetch-in-render logic to `useQueries`, iterating over each query's result and options to handle the `experimental_prefetchInRender` flag individually. For each query that will fetch and has the option enabled:

- New cache entries trigger `fetchOptimistic()` immediately
- Existing cache entries subscribe to the query's cache promise
- On promise resolution, `observer.updateResult()` is called to finalize the thenable

Closes #9988

## Summary by CodeRabbit

* **New Features**
  * Added client-side support for experimental prefetch-on-render optimization, improving data loading and responsiveness when rendering queries.

* **Chores**
  * Added release changeset documenting the patch and the new prefetch-on-render behavior.

### DIFF
--- a/.changeset/fix-use-queries-prefetch-in-render.md
+++ b/.changeset/fix-use-queries-prefetch-in-render.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/react-query': patch
+---
+
+Add experimental_prefetchInRender support to useQueries so that React.use(promise) resolves correctly
--- a/packages/react-query/src/useQueries.ts
+++ b/packages/react-query/src/useQueries.ts
@@ -4,6 +4,7 @@ import * as React from 'react'
 import {
   QueriesObserver,
   QueryObserver,
+  isServer,
   noop,
   notifyManager,
 } from '@tanstack/query-core'
@@ -19,6 +20,7 @@ import {
   ensureSuspenseTimers,
   fetchOptimistic,
   shouldSuspend,
+  willFetch,
 } from './suspense'
 import type {
   DefinedUseQueryResult,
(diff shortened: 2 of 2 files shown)


TkDodo (approved):
can you add a test please

TkDodo:
I’ve decided to get rid of this experimental flag.

PR #10234, opened 2026-03-05
fix(vue-query/useBaseQuery): prevent dual error propagation when 'suspense()' and error watcher both handle the same error

sukvvon:
## 🎯 Changes

When using `throwOnError: true` with `suspense()`, the same error was propagated through two paths simultaneously:
1. `suspense()` → `fetchOptimistic` failure → `reject(error)` (Promise rejection)
2. Error watcher → `state.error` change detected → `throw error` (watcher throw)

This caused an **Unhandled Rejection** because both paths fired concurrently due to Vue's async watcher system.

Added `isSuspenseFetching` flag to coordinate between `suspense()` and the error watcher. When `suspense()` is actively handling a fetch error, the error watcher skips its `throw`, ensuring the error is propagated through only one path (Promise rejection).

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate error propagation when suspense and error watchers handle the same query error.
  * Improved error handling for regular and infinite queries during suspense.

* **Tests**
  * Added coverage for error handling with and without suspense, including watcher behavior and error states.

* **Chore**
  * Added a changeset entry for a patch release.


themavik (commented):
useBaseQuery suspenseFetchCount only increments on the stale fetchOptimistic path, so the error watcher does not double-throw while that promise is in flight — matches the new useQuery/useInfiniteQuery tests. nit: useInfiniteQuery.test.ts vi.mocks ../useBaseQuery; if integration with the real module graph is covered elsewhere, a one-line note saves the next reader a detour.

sukvvon:
@DamianOsipiuk Could you also review this PR?

packages/vue-query/src/__tests__/useInfiniteQuery.test.ts
```diff
@@ -78,4 +81,73 @@ describe('useInfiniteQuery', () => {
...
+        queryFn: () =>
+          sleep(10).then(() => Promise.reject(new Error('Some error'))),
+        initialPageParam: 0,
+        getNextPageParam: () => 12,
+        retry: false,
+        throwOnError: throwOnErrorFn,
+      })
+
+      // Suppress the Unhandled Rejection caused by watcher throw in Vue 3
+      const rejectionHandler = () => {}
```
DamianOsipiuk: Should this be a `spy` with assertion at the end?

packages/vue-query/src/__tests__/useInfiniteQuery.test.ts
```diff
@@ -78,4 +81,73 @@ describe('useInfiniteQuery', () => {
...
+      await vi.advanceTimersByTimeAsync(10)
+
+      await promise
+
+      expect(rejectedError).toBeInstanceOf(Error)
+      expect((rejectedError as Error).message).toBe('Some error')
+      // throwOnError is evaluated in both suspense() and the error watcher
+      expect(throwOnErrorFn).toHaveBeenCalledTimes(2)
+      // but the error watcher should not throw when suspense is active
+      expect(query).toMatchObject({
```
DamianOsipiuk, line 172: This actually does not assert, what comment suggests.

Should there be a spy on `unhandledRejection` with assertion that it was not caled?

packages/vue-query/src/__tests__/useQuery.test.ts
```diff
@@ -476,6 +476,34 @@ describe('useQuery', () => {
...
+      useQuery({
+        queryKey: ['throwOnErrorWithoutSuspense'],
+        queryFn: () =>
+          sleep(10).then(() => Promise.reject(new Error('Some error'))),
+        retry: false,
+        throwOnError: throwOnErrorFn,
+      })
+
+      // Suppress the Unhandled Rejection caused by watcher throw in Vue 3
+      const rejectionHandler = () => {}
```
DamianOsipiuk: same for these tests

packages/vue-query/src/__tests__/useInfiniteQuery.test.ts
sukvvon: @DamianOsipiuk Good catch — switched the handler to `vi.fn()` and asserted it was called once with the rejected error in 9e2924c04.

packages/vue-query/src/__tests__/useInfiniteQuery.test.ts
sukvvon: @DamianOsipiuk You're right — the previous assertion did not match the comment's intent. In 9e2924c04 I added a `vi.fn()` spy on `unhandledRejection` and asserted `not.toHaveBeenCalled()` so the watcher's non-rethrow is directly verified.

packages/vue-query/src/__tests__/useQuery.test.ts
sukvvon: @DamianOsipiuk Applied the same spy pattern to both `useQuery` tests in 9e2924c04 (and changed `test` → `it` for consistency with the rest of the file).

sukvvon:
@DamianOsipiuk Apologies for the long delay in getting back to this — thank you for your patience.

I've switched these two "suspense not used" tests to spy-and-assert as you suggested (228f73605).

One caveat worth noting: a plain `unhandledRejection` spy only works on Vue 3 — on Vue 2.x the watcher rethrow surfaces through `console.error` instead, so asserting solely on `unhandledRejection` failed the 2.x matrix. The tests now capture both channels and assert the rejected error surfaced through either one, keeping the assertion valid across all supported Vue versions.

PR #10258, opened 2026-03-11
fix: improve exhaustive dependencies detection

Newbie012:
## 🎯 Changes

I updated the lint rule `exhaustive-deps` so it now catches dependencies used inside nested `queryFn` callbacks/control flow, and it no longer reports false positives when those dependencies are already encoded in "complex" `queryKey` expressions.

```ts
// 🔴 Before: false negative, no lint error.
// 🟢 After: reports "The following dependencies are missing in your queryKey: id".
useQuery({
  queryKey: ['todo'],
  queryFn: () => Promise.resolve().then(() => id),
})
```

```ts
// 🔴 Before: false positive, reported "The following dependencies are missing in your queryKey: cond, a, b".
// 🟢 After: valid, no exhaustive-deps error.
useQuery({
  queryKey: ['todo', cond ? a : b],
  queryFn: () => (cond ? a : b),
})
```

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed exhaustive-deps rule to properly detect dependencies used in nested queryFn callbacks and control flow constructs (then/catch, try/catch/finally).
  * Improved handling of complex queryKey expressions to reduce false positives.

* **Tests**
  * Expanded test coverage for various queryKey formations and missing dependency scenarios.


hanneswidrig:
I think there might be some further work to do on this PR, https://tanstack.com/query/v5/docs/eslint/exhaustive-deps

From my understanding of query keys, the following code should not emit an error for the `api` missing in the query key. The changes after this PR steer the user towards treating the `queryKey` like the `useEffect` dependency array which might or might not be intentional behavior.

> error  The following dependencies are missing in your queryKey: api  @tanstack/query/exhaustive-deps

```ts
useQuery({
  queryKey: ['todo', todoId],
  queryFn: () => api.getTodo(todoId),
})
```

Newbie012:
@hanneswidrig You're right.

see https://github.com/TanStack/query/pull/10295

maapteh:
This can create eternal queries because you can pass the query itself (instead of the id, which makes it generic in getTodo from operation perspective). As a generic rule in current state i guess this can make things worse.

```
const Component = ({ todoId }) => {
  const todos = useTodos()
  useQuery({
    queryKey: ['todo', todos, todoId],
    queryFn: () => todos.getTodo(todoId),
  })
}
```

This can't be the correct way (for example when my function is a hook vs simple method, the hook has loading state etc etc ... this generates eternal queries)? 

People using auto fix command will get the weirdest results when their regression tests are not on par. It all depends :)

PR #10265, opened 2026-03-13
Referential stability for observer result objects

zeorin:
## 🎯 Changes

- Add result object property change notification to mutation observers. By default, property access is tracked and used to determine which properties to notify for.
- Compare result objects to most recently returned results, if they are shallowly equal, return a reference to the previous object rather than a new object that is equal by value but not by reference.
- When returning a proxy object for property access tracking, compare the previously tracked result and the new one only by the tracked properties, and if none of those have changed, return the previous reference.

## ✅ Checklist

- [x] I have tested this code locally with `pnpm run test:pr`. (Didn't work for me, but `npx nx run @tanstack/react-query:test:lib` passes).

## 🚀 Release Impact

### DIFF
--- a/docs/framework/react/guides/render-optimizations.md
+++ b/docs/framework/react/guides/render-optimizations.md
@@ -13,7 +13,7 @@ React Query uses a technique called "structural sharing" to ensure that as many 
 
 ### referential identity
 
-The top level object returned from `useQuery`, `useInfiniteQuery`, `useMutation` and the Array returned from `useQueries` is **not referentially stable**. It will be a new reference on every render. However, the `data` properties returned from these hooks will be as stable as possible.
+The top level object returned from `useQuery`, `useInfiniteQuery`, `useMutation` and the Array returned from `useQueries` is referentially stable unless React Query has triggered a re-render.
 
 ## tracked properties
 
--- a/packages/query-core/src/mutationObserver.ts
+++ b/packages/query-core/src/mutationObserver.ts
@@ -40,6 +40,16 @@ export class MutationObserver<
   #currentMutation?: Mutation<TData, TError, TVariables, TOnMutateResult>
   #mutateOptions?: MutateOptions<TData, TError, TVariables, TOnMutateResult>
   #trackedProps = new Set<keyof MutationObserverResult>()
+  #lastTrackedResult?: MutationObserverResult<
+    TData,
+    TError,
+    TVariables,
+    TOnMutateResult
+  >
+  #resultProxyCache = new WeakMap<
+    MutationObserverResult<TData, TError, TVariables, TOnMutateResult>, // un-proxied result
+    MutationObserverResult<TData, TError, TVariables, TOnMutateResult> // proxied result
+  >()
 
   constructor(
     client: QueryClient,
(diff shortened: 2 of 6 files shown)


themavik (commented):
QueryObserver.trackResult / MutationObserver trackResult caching with WeakMap + shallowEqualObjects on #updateResult should cut a lot of proxy churn. nit: render-optimizations.md now says hook return objects are stable unless RQ triggers a re-render — worth one clause that option/key changes still produce new result objects when underlying data moves, so readers do not read stable as frozen forever.

Zelys-DFKH:
**Edit (2026-08-24):** Struck the parts of this comment that are wrong. I described a production scenario I can't substantiate and said I had tested this branch; I fetched the PR ref but never checked it out. The re-render claim is also wrong, corrected with an actual reproduction below. Leaving the original text visible rather than deleting it, since it was replied to.

---

~~Commenting because this directly addresses a problem I've hit in production.~~

~~We poll blockchain contract data every 60 seconds in a DeFi game — ~15 concurrent `useQuery` instances with `refetchInterval: 60_000`. The data almost never changes (most hero stats, inventory balances, etc. update on player actions, not on a timer).~~ The `queryFn` return values are structurally equal between polls, so `structuralSharing` correctly preserves the `data` reference. But the result object itself changes on every poll cycle because `isFetching` briefly flips `false → true → false` mid-fetch. ~~Any component that tracks *only* `data` still re-renders twice per cycle, because the outer object is new.~~

What this PR does for that case: `trackResult()` now returns the previous proxy when only `data` is tracked and `data` hasn't changed — even though the result object has a new reference. ~~That eliminates the spurious renders without any changes at the call site.~~

~~Tested by checking out the branch and running a polling scenario with property-access tracking. The proxy cache correctly reuses the old reference when `isFetching` is not in `#trackedProps`.~~

One thing worth considering: the `shallowEqualObjects` path assumes the tracked props are stable at the result level. With `select`, the `data` property is the selected value — `structuralSharing` runs on the raw query data before `select`, so if `select` returns a new object on every call (even with identical input), `shallowEqualObjects` will always return `false` and the proxy cache won't help. That's a user error but common enough that a note in the docs might be worth it. Not a blocker, just an edge case worth surfacing.

Overall the approach is clean. `WeakMap` keying on the raw result is the right call for memory safety, and `#lastTrackedResult` as the single strong reference means only one result stays live at a time. Changeset is missing, would be a patch for `@tanstack/query-core` and a minor for `@tanstack/react-query` given the new `notifyOnChangeProps` option on mutations.

TkDodo:
> But the result object itself changes on every poll cycle because isFetching briefly flips false → true → false mid-fetch. Any component that tracks only data still re-renders twice per cycle, because the outer object is new.

That shouldn’t happen. If you only use `result.data`, your component should only re-render when `data` changes, _not_ when `isFetching` changes. That’s what tracked properties are doing. If  that’s not the case, please open a separate issue with a minimal reproduction. The top-level object being a new reference on every render shouldn’t better because a render should never start if nothing observed changes.

Zelys-DFKH:
You're right, and the claim in my earlier comment was wrong. I've annotated it above rather than deleting it, since you replied to it.

I built the reproduction I should have built in March, on `@tanstack/react-query` 5.102.3: a `queryFn` returning a fresh object with identical contents each poll, so structural sharing keeps `data` referentially stable while the result object identity changes every cycle. Renders counted over five poll cycles, initial load excluded:

- component reads only `result.data`: 0 re-renders
- component reads `result.data` and `result.isFetching`: one per cycle

Tracked properties do exactly what you said. `shouldNotifyListeners` only returns true when a changed key is in the tracked set, so the new top-level reference never reaches the component. There's no issue to open.

I should also correct the rest of that comment: I described a production scenario I can't substantiate and said I had tested this branch. I fetched the PR ref and read the diff, but I never ran it. Sorry for putting you in the position of debunking something I hadn't checked.

PR #10267, opened 2026-03-13
feat(query-core): add refetchIntervalOnServer option

K-Mistele:
## Summary

- Adds a new `refetchIntervalOnServer` option to `QueryObserverOptions` that allows `refetchInterval` to remain active in server environments
- Defaults to `false`, preserving existing behavior where `refetchInterval` is disabled when `isServer` is true
- Includes two new tests validating both the default behavior and the opt-in behavior

## Motivation

`@tanstack/query-core` defines `isServer = typeof window === 'undefined' || 'Deno' in globalThis` and uses it in `QueryObserver#updateRefetchInterval` to unconditionally disable `refetchInterval` on the server. This makes sense for SSR, where polling during server-side rendering is wasteful.

However, there are legitimate server-side use cases where polling is essential. For example, long-running daemon processes and background workers that use TanStack Query (often via TanStack DB query collections) for real-time data synchronization rely on `refetchInterval` as a polling-based fallback when primary sync channels (like WebSockets or Electric SQL) are unavailable.

Currently, the only workaround is to monkey-patch the `isServer` export (as the test utils already do via `setIsServer`), which is fragile and not a supported API.

## API Design

This follows the exact same pattern as the existing `refetchIntervalInBackground` option:

```ts
  queryKey: ['sync-data'],
  queryFn: fetchData,
  refetchInterval: 5000,
  refetchIntervalOnServer: true, // NEW: opt-in to server-side polling
})
```

It can also be set as a default for all queries via `QueryClient`:

```ts
  defaultOptions: {
    queries: {
      refetchIntervalOnServer: true,
    },
  },
})
```

## Changes

| File | Change |
|------|--------|
| `packages/query-core/src/types.ts` | Add `refetchIntervalOnServer?: boolean` to `QueryObserverOptions` with JSDoc |
| `packages/query-core/src/queryObserver.ts` | Check `this.options.refetchIntervalOnServer` before skipping interval setup on server |
| `packages/query-core/src/__tests__/queryObserver.test.tsx` | Add tests for default (no polling on server) and opt-in (polling on server) behaviors |

## Test plan

- [x] Existing test suite passes (496 tests, all passing)
- [x] TypeScript compilation passes
- [x] New test: `should not refetch on server by default` -- verifies default behavior is unchanged
- [x] New test: `should refetch on server when refetchIntervalOnServer is true` -- verifies opt-in works

## Summary by CodeRabbit

* **New Features**
  * Added `refetchIntervalOnServer` configuration option to control server-side query refetch behavior. By default, refetch intervals are now disabled on servers; enable this option to allow interval-based refetches in server environments.

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -18,11 +18,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v6.0.1
-        with:
-          fetch-depth: 0
+        uses: actions/checkout@v6.0.2
       - name: Setup Tools
-        uses: tanstack/config/.github/setup@main
+        uses: TanStack/config/.github/setup@main
       - name: Fix formatting
         run: pnpm run format
       - name: Apply fixes
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,7 +1,7 @@
 name: Labeler
 
 on:
-  - pull_request_target
+  pull_request_target:
 
 permissions:
   contents: read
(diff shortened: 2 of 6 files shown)


TkDodo:
I don’t want to add an extra option for this, however, we do expose an `environmentManager` now where you can override how `isServer` is computed:

https://tanstack.com/query/latest/docs/reference/environmentManager

PR #10272, opened 2026-03-14
refactor: solid v2 pre-release for solid-query v6

brenelz:
## 🎯 Changes

Update solid-query to support v2 of Solid

### DIFF
--- a/packages/solid-query-persist-client/package.json
+++ b/packages/solid-query-persist-client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tanstack/solid-query-persist-client",
-  "version": "5.90.26",
+  "version": "5.90.28",
   "description": "Solid.js bindings to work with persisters in TanStack/solid-query",
   "author": "tannerlinsley",
   "license": "MIT",
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -1,0 +1,32 @@
+{
+  "mode": "pre",
+  "tag": "alpha",
+  "initialVersions": {
+    "@tanstack/angular-query-experimental": "5.90.28",
+    "@tanstack/angular-query-persist-client": "5.62.32",
+    "@tanstack/eslint-plugin-query": "5.91.5",
+    "@tanstack/preact-query": "5.94.2",
+    "@tanstack/preact-query-devtools": "5.92.0",
+    "@tanstack/preact-query-persist-client": "5.92.3",
+    "@tanstack/query-async-storage-persister": "5.90.27",
+    "@tanstack/query-broadcast-client-experimental": "5.90.23",
+    "@tanstack/query-core": "5.91.2",
+    "@tanstack/query-devtools": "5.93.0",
+    "@tanstack/query-persist-client-core": "5.92.4",
+    "@tanstack/query-sync-storage-persister": "5.90.27",
+    "@tanstack/query-test-utils": "0.0.0",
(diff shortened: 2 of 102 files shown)


2wheeh:
Hi, I opened a small PR against your branch https://github.com/brenelz/query/pull/1, would you be open to this?

packages/solid-query/src/useQueries.ts
```diff
@@ -222,122 +215,63 @@ export function useQueries<
...
-      return new Proxy(s, handler(index))
-    })
-
-  const [proxyState, setProxyState] = createStore(getProxies())
-  createRenderEffect(() => setProxyState(getProxies()))
+  // Force read of trackedDefaultedQueries to ensure it runs
+  void trackedDefaultedQueries
 
-  return proxyState as TCombinedResult
+  return state as unknown as TCombinedResult
```
birkskyum, line 276: Are we able to keep the type safety?

packages/solid-query/src/__tests__/useMutation.test.tsx
```diff
@@ -659,92 +673,77 @@ describe('useMutation', () => {
...
-    })
-    expect(states[1]).toMatchObject({
-      isPending: true,
-      isPaused: false,
-      failureCount: 0,
-      failureReason: null,
-    })
-    expect(states[2]).toMatchObject({
-      isPending: true,
-      isPaused: false,
```
birkskyum: are these test changes safe? always bit worried when fixtures get modified

packages/solid-query/src/__tests__/useMutation.test.tsx
```diff
@@ -589,16 +598,17 @@ describe('useMutation', () => {
       mutationKey: undefined,
     })
 
-    onlineMock.mockRestore()
-    window.dispatchEvent(new Event('online'))
```
birkskyum: can we make solid v2 also run using events?

packages/solid-query-persist-client/package.json
```diff
@@ -66,16 +66,20 @@
...
-    "solid-js": "^1.9.7",
+    "solid-js": "2.0.0-beta.2",
     "tsup-preset-solid": "^2.2.0",
-    "vite-plugin-solid": "^2.11.6"
+    "vite-plugin-solid": "3.0.0-next.2"
   },
   "peerDependencies": {
     "@tanstack/solid-query": "workspace:^",
-    "solid-js": "^1.6.0"
+    "solid-js": "^1.6.0 || >=2.0.0-beta.0 <2.0.0"
```
birkskyum: that's an unusual range - I think we can make it allow version ^2.0.0 as well, and not allow 1.x since it would break at this version.

packages/solid-query/src/__tests__/useMutation.test.tsx
brenelz: yeah im not 100% sure... some needed to change to pass

packages/solid-query/src/__tests__/mutationOptions.test.tsx
```diff
@@ -420,6 +423,7 @@ describe('mutationOptions', () => {
       </QueryClientProvider>
     ))
 
+    await Promise.resolve()
```
2wheeh, line 426: flush() ?

brenelz: I think this works too

PR #10321, opened 2026-03-23
fix(query-devtools): sanitize invalid browser locales

### LINKED ISSUE #10286: @tanstack/query-devtools: invalid navigator.language values like "undefined" cause a RangeError at runtime
### Describe the bug

`@tanstack/query-devtools` throws at runtime when `navigator.language` is an invalid string such as `"undefined"`.

The published package still derives the locale from:

```js
(navigator.language || navigator.userLanguage) || "en-US"
```

If a browser/runtime reports `navigator.language === "undefined"` instead of `undefined`, that value reaches locale-sensitive logic and causes a `RangeError`.

I reproduced this with a minimal React app that mounts `ReactQueryDevtools` after forcing:

```js
navigator.language = "undefined"
navigator.languages = ["undefined"]
navigator.userLanguage = "undefined"
```

https://github.com/jakvbs/query-devtools-locale-repro

### Steps to reproduce

1. Install the repro dependencies
2. Run the app
3. Open it in Firefox or Chromium
4. The app forces `navigator.language` to the string `"undefined"` before mounting `ReactQueryDevtools`
5. Observe a runtime `RangeError`

### Expected behavior

Devtools should sanitize invalid locale values and fall back safely to `"en-US"` instead of throwing.

Every time

### Screenshots or Videos

Firefox:

```text
RangeError: invalid language tag: "undefined"
    at mount/dispose< (.../@tanstack_react-query-devtools.js:111:14)
```

Chromium:

```text
RangeError: Incorrect locale information provided
    at getDefaultLocale (.../chunk-*.js:2496:16)
```

The wording differs by browser, but the root cause is the same invalid locale path.

### Platform

- OS: Linux
- Browsers:
  - Firefox 146.0
  - Chromium (Playwright-installed Chromium)

### TanStack Query adapter

react-query

### TanStack Query version

Repro app:
- `@tanstack/react-query`: `5.90.20`
- `@tanstack/react-query-devtools`: `5.91.3`

I also inspected published `@tanstack/query-devtools` tarballs for:
- `5.91.1`
- `5.92.0`
- `5.93.0`

All three still contain:

```js
function getDefaultLocale() {
  let locale = typeof navigator !== "undefined" &&
    (navigator.language || navigator.userLanguage) || "en-US";
```

### TypeScript version

Not relevant for this runtime bug

### Additional context

This reproduces in both Firefox and Chromium.

One thing that made this confusing is that a quick search in the current source tree does not obviously show `getDefaultLocale`, because that logic comes from bundled dependency code in the published `@tanstack/query-devtools` artifact rather than a handwritten source file in this repo.

The devtools source also has direct locale-sensitive `toLocaleString()` call sites, so the invalid locale can surface through more than one runtime path.

jakvbs:
## What
- sanitize invalid browser locale values before devtools formatting uses them
- wrap the query devtools UI in a Kobalte `I18nProvider` with the sanitized locale and route mutation date formatting through the sanitized locale helpers
- add regression tests for invalid `navigator.language` values like `"undefined"`
- add a patch changeset for `@tanstack/query-devtools`

## Why
- `@tanstack/query-devtools` can currently throw at runtime when the browser reports an invalid locale string, for example `navigator.language === "undefined"`
- the invalid locale can surface through bundled dependency locale resolution as well as devtools' own locale-sensitive date formatting paths

## How tested
- `npx nx run @tanstack/query-devtools:test:eslint`
- `npx nx run @tanstack/query-devtools:test:types`
- `npx nx run @tanstack/query-devtools:test:lib`
- `npx nx run @tanstack/query-devtools:build`
- `npx nx run @tanstack/query-devtools:test:build`
- manual repro against the linked issue reproduction app using Playwright-managed Chromium `145.0.7632.6` and Firefox `146.0.1`; verified the unfixed published packages throw and the patched local packages no longer do

Fixes #10286.

## Summary by CodeRabbit

* **New Features**
  * Devtools now uses an i18n-aware locale provider so timestamps display according to the current locale.
  * Date/time displays in mutation and query panels use consistent locale-aware formatting.

* **Bug Fixes**
  * Invalid or unsupported browser locales are sanitized and now safely fall back to English (US).

* **Tests**
  * Added tests covering locale selection, sanitization, and date/time formatting.

### DIFF
--- a/.changeset/safe-languages-wash.md
+++ b/.changeset/safe-languages-wash.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/query-devtools': patch
+---
+
+Sanitize invalid browser locale values before query devtools use them for locale-sensitive formatting.
--- a/packages/query-devtools/src/Devtools.tsx
+++ b/packages/query-devtools/src/Devtools.tsx
@@ -15,12 +15,13 @@ import { clsx as cx } from 'clsx'
 import { TransitionGroup } from 'solid-transition-group'
 import { Key } from '@solid-primitives/keyed'
 import { createResizeObserver } from '@solid-primitives/resize-observer'
-import { DropdownMenu, RadioGroup } from '@kobalte/core'
+import { DropdownMenu, RadioGroup, useLocale } from '@kobalte/core'
 import { Portal } from 'solid-js/web'
 import { tokens } from './theme'
 import {
   convertRemToPixels,
   displayValue,
+  formatDateTime,
   getMutationStatusColor,
   getQueryStatusColor,
   getQueryStatusColorByLabel,
@@ -675,6 +676,7 @@ export const ContentView: Component<ContentViewProps> = (props) => {
   setupQueryCacheSubscription()
   setupMutationCacheSubscription()
   let containerRef!: HTMLDivElement
+  const locale = useLocale()
   const theme = useTheme()
   const css = useQueryDevtoolsContext().shadowDOMTarget
     ? goober.css.bind({ target: useQueryDevtoolsContext().shadowDOMTarget })
@@ -778,7 +780,7 @@ export const ContentView: Component<ContentViewProps> = (props) => {
                 item.options.mutationKey
                   ? JSON.stringify(item.options.mutationKey) + ' - '
                   : ''
-              }${new Date(item.state.submittedAt).toLocaleString()}`
+              }${formatDateTime(item.state.submittedAt, locale.locale())}`
               return rankItem(value, props.localStore.mutationFilter || '')
                 .passed
             })
@@ -1501,6 +1503,7 @@ const QueryRow: Component<{ query: Query }> = (props) => {
 }
 
 const MutationRow: Component<{ mutation: Mutation }> = (props) => {
+  const locale = useLocale()
   const theme = useTheme()
   const css = useQueryDevtoolsContext().shadowDOMTarget
     ? goober.css.bind({ target: useQueryDevtoolsContext().shadowDOMTarget })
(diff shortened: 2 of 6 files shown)


afurm:
Nice fix. One observation on `sanitizeLocale` in `utils.tsx`: `Intl.DateTimeFormat.supportedLocalesOf()` does not throw on invalid locale tags - it returns an empty array. So the `try/catch` around it is effectively dead code in any environment that supports `Intl.DateTimeFormat.supportedLocalesOf`. The `|| fallback` on the result already handles the invalid-locale case correctly.

The `try/catch` would only matter if you wanted to guard against environments where `Intl.DateTimeFormat.supportedLocalesOf` itself is undefined, but at that point the whole devtools locale feature would be broken anyway. Worth removing it to keep the code honest, or adding a comment explaining what it guards against.

jakvbs:
> Nice fix. One observation on `sanitizeLocale` in `utils.tsx`: `Intl.DateTimeFormat.supportedLocalesOf()` does not throw on invalid locale tags - it returns an empty array. So the `try/catch` around it is effectively dead code in any environment that supports `Intl.DateTimeFormat.supportedLocalesOf`. The `|| fallback` on the result already handles the invalid-locale case correctly.
> 
> The `try/catch` would only matter if you wanted to guard against environments where `Intl.DateTimeFormat.supportedLocalesOf` itself is undefined, but at that point the whole devtools locale feature would be broken anyway. Worth removing it to keep the code honest, or adding a comment explaining what it guards against.

I don’t think the try/catch is dead code for this repro. `Intl.DateTimeFormat.supportedLocalesOf(['undefined'])` throws a `RangeError` here, rather than returning `[]`. Returning an empty array is the behavior for well-formed but unsupported locales, whereas `"undefined"` is a syntactically invalid locale tag. That is exactly what the repro sets `navigator.language` to, so the `catch` is handling a real failure mode in this fix.

PR #10335, opened 2026-03-25
fix: infer pageParam from computed queryFn with conditional skipToken

danieljayasurya-e:
## 🎯 Changes

This PR fixes a Vue Query type inference issue in useInfiniteQuery when queryFn is provided through a computed wrapper,  
especially in conditional skipToken cases.

Previously, TypeScript could lose contextual typing for the computed queryFn, causing pageParam to fall back to unknown  even when initialPageParam clearly established its type. This change makes the queryFn option preserve contextual typing  in that shape by explicitly typing the property and widening the skipToken branch to a generic symbol for inference      
purposes.

The PR also adds a regression type test covering computed(() => enabled ? ({ pageParam }) => ... : skipToken) to ensure  
pageParam and lastPageParam remain strongly typed.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Improved TypeScript typing for useInfiniteQuery so query functions provided via computed/refs or skip tokens infer pageParam more accurately.
* **Tests**
  * Added type-level tests validating inference for conditional query functions, skip-token scenarios, and numeric pageParam progression in getNextPageParam.


packages/vue-query/src/useInfiniteQuery.ts
```diff
@@ -48,6 +53,14 @@ export type UseInfiniteQueryOptions<
             TPageParam
           >[Property]
         >
+      : Property extends 'queryFn'
+        ? MaybeRefOrGetter<
+            QueryFunction<
+              TQueryFnData,
+              DeepUnwrapRef<TQueryKey>,
+              TPageParam
+            > | SkipTokenForUseInfiniteQuery
```
TkDodo: should we just use the type `SkipToken` here?

https://github.com/TanStack/query/blob/6fa901b97a22a80d0fca3f6ed86237ff0cbdd13b/packages/query-core/src/utils.ts#L427

danieljayasurya-e: Thanks for the suggestion.
Using SkipToken directly here makes sense. I’ll switch to using **SkipToken** from core

danieljayasurya-e: @TkDodo  I've updated the implementation to use SkipToken directly and removed the extra type.

TkDodo (approved):

TkDodo:
test are failing, seems related

peterbud:
I have looked at this PR, and the test failure is real. 

IMHO this doesn’t seem fixable through `UseInfiniteQueryOptions` alone because `computed()` is typed before `initialPageParam` inference completes. The test passes if `computed` is explicitly typed with `QueryFunction<..., number> | SkipToken`, but without that it won't infer `pageParam`

PR #10348, opened 2026-03-28
fix(query-devtools): guard invalid default locales

### LINKED ISSUE #10286: @tanstack/query-devtools: invalid navigator.language values like "undefined" cause a RangeError at runtime

raashish1601:
## Summary
- canonicalize browser-reported locales before formatting query devtools mutation timestamps
- fall back to en-US when navigator locale values are invalid
- add focused locale helper regression coverage and a patch changeset

## Testing
- corepack pnpm dlx vitest run packages/query-devtools/src/__tests__/locale.test.ts

Closes #10286

### DIFF
--- a/.changeset/olive-carpets-rhyme.md
+++ b/.changeset/olive-carpets-rhyme.md
@@ -1,0 +1,7 @@
+---
+'@tanstack/query-devtools': patch
+---
+
+Guard devtools mutation timestamp formatting against invalid browser locale
+values by canonicalizing navigator locales before calling
+`Date.prototype.toLocaleString()`.
--- a/packages/query-devtools/src/Devtools.tsx
+++ b/packages/query-devtools/src/Devtools.tsx
@@ -18,6 +18,7 @@ import { createResizeObserver } from '@solid-primitives/resize-observer'
 import { DropdownMenu, RadioGroup } from '@kobalte/core'
 import { Portal } from 'solid-js/web'
 import { tokens } from './theme'
+import { formatDateTime } from './locale'
 import {
   convertRemToPixels,
   displayValue,
@@ -778,7 +779,7 @@ export const ContentView: Component<ContentViewProps> = (props) => {
                 item.options.mutationKey
                   ? JSON.stringify(item.options.mutationKey) + ' - '
                   : ''
-              }${new Date(item.state.submittedAt).toLocaleString()}`
+              }${formatDateTime(item.state.submittedAt)}`
               return rankItem(value, props.localStore.mutationFilter || '')
                 .passed
             })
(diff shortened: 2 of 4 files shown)


TkDodo:
closing as of:

- https://github.com/TanStack/query/issues/10286#issuecomment-4366300808

PR #10444, opened 2026-04-10
fix(query-core): fix hydration bugs for already resolved promises

### LINKED ISSUE #9642: Race condition causing hydration error when streaming with server components
### Describe the bug

I'm following the guidance [here](https://tanstack.com/query/v5/docs/framework/react/guides/advanced-ssr#streaming-with-server-components) to implement streaming with server components.

I'm prefetching data like so (note, no await):

  ```
const streamingQueryClient = getQueryClient();
  streamingQueryClient.prefetchQuery({
    queryKey: ['roles-client-streaming'],
    queryFn: () => getRoles(),
  });
```

And inside a client component:

```
export default function ClientComponentWithStreaming() {
  const { data, isFetching, status, isStale, fetchStatus } = useSuspenseQuery({
    queryKey: ['roles-client-streaming'],
    queryFn: getRolesClient,
  });

  console.log('data', data);
  console.log('isFetching', isFetching);
  console.log('status', status);
  console.log('isStale', isStale);
  console.log('fetchStatus', fetchStatus);

  return (
    <>
      <ul className="list-disc pl-5 mt-4">
        {data &&
          data!.map((item: any) => (
            <li key={item.id} className="mb-2">
              {item.name}
            </li>
          ))}
      </ul>
      {isFetching && <LoadingPanel />}
    </>
  );
}

```

The issue i'm seeing is that intermittently, isFetching and isStale are true on first client render. This causes a hydration error as it tries to render the LoadingPanel.

In my actual app, this is not intermittent and reproducible on every page load.

During server side rendering, the console logs are as follows:

```
data Array(1)
isFetching false
status success
isStale false
fetchStatus idle
```

So i would not expect the first client paint to be in a fetching state. However, when the hydration error occurs, the logs in the browser on the client show as:

```
data [{…}]
isFetching true
status success
isStale true
fetchStatus fetching
```

Under what conditions would the first client render be in a fetching state if the data is already prefetched via a promise?

https://github.com/brettpostin/tanstack-query-hydration-issue

### Steps to reproduce

1. Pull and run repo
2. Keep refreshing until you hit the hydration error

### Expected behavior

I would expect the first client paint to not be in a fetching state if the data is present in the cache from a prefetch.

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: Windows
- Browser: Edge
- TanStack Query Version: 5.87.4

### Tanstack Query adapter

react-query

### TanStack Query version

5.87.4

### TypeScript version

_No response_

### Additional context

_No response_

Ephem:
## 🎯 Changes

Fixes https://github.com/TanStack/query/issues/9642

This PR fixes and adds tests for two different bugs that would both cause queries to incorrectly jump into the fetching/pending state on hydration.

This would happen for queries that were prefetched without awaiting, like so: `void queryClient.prefetchQuery(...)` and then dehydrated, but where the promise would resolve before the hydration would happen. The query state in the hydration would be `status: 'pending'`, since that's the status on the server at the moment of hydration, but when hydrating we should put them into the `'successful'` state since we already have data.

First bug is that we did this for _new_ queries, but not when queries already existed in the cache.

The second bug is that for both cases, we also still called `query.fetch` with an `initialPromise`, which would briefly emit a fetching state. We did this to set up a retryer and I even wrote a comment in the code that this was necessary even for queries that already had data, but I've since reconsidered. The retryer is optional, so if the query is already in a successful state, I don't _think_ we need to do that for either new or existing queries and removing it shows no failing tests. This is a controversial part of the PR though and I'm happy for pushback.

If we need to keep the data in sync with the promise in the retryer, we already have a problem with that when people call `query.setData`, since we don't do anything with the retryer in those situations either.

Note that the promise we return from `useQuery` etc is different, that comes from the observer, not directly from the query.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Hydration no longer leaves queries backed by already-resolved or synchronously-resolved promises briefly in a pending/fetching state; such queries resolve to success immediately.

* **Tests**
  * Added coverage for rehydration with synchronously-resolved promises to ensure correct cache status transitions and prevent regressions.

* **Chores**
  * Added a patch release changeset documenting the hydration fix.


TkDodo (approved):

afurm:
Good regression tests for this hydration edge case. A few observations:

**On the `syncData` check in `fetch`:** The new condition that prevents setting up a retryer when data is synchronously available makes sense — if the promise already resolved, we do not need retry logic. But does this interact with `cacheTime` or `staleTime` options? If a query is re-hydrated with sync data, goes stale, and then the user triggers a refetch, will the retryer be set up correctly then?

**On preserving `fetchStatus` during re-hydration:** The logic that preserves `fetchStatus` if the existing query is actively fetching is correct. But what if the existing query was fetching stale data and the dehydrated data is fresher? Would that cause a brief flash of incorrect fetch status before the next update?

**On the test setup:** The pattern to override `promise.then` to simulate synchronous resolution is clever but quite invasive. Have you considered using a microtask or `Promise.resolve().then()` instead to simulate the synchronous-thenable case more naturally without mutating the promise?

Ephem:
@afurm 

> will the retryer be set up correctly then

Yes. The logic whether to revalidate lies elsewhere, if that elsewhere triggers a fetch, that will set up the retryer.

> But what if the existing query was fetching stale data and the dehydrated data is fresher?

If a fetch is ongoing on the client, it's very likely to fetch fresher data than what was in a dehydrated query, so this is a tradeoff I'm willing to live with. 😄

> Have you considered using a microtask or Promise.resolve().then() instead

The whole point is that it needs to be fully synchronous, so no.

Issue #10463, opened 2026-04-12
useMutationState loses type inference in select callback due to non-generic Mutation type


mukeshdhadhariya:
### Describe the bug

`useMutationState` does not preserve type information inside the `select` callback. The `mutation` parameter is typed as a non-generic `Mutation`, causing `mutation.state.data` to be inferred as `unknown`.

### Reproduction

```ts
const result = useMutationState({
  select: (mutation) => mutation.state.data
})
```

### Your minimal, reproducible example

```
import type { Mutation } from '@tanstack/query-core'  type MutationStateOptions<TResult = unknown> = {   select?: (mutation: Mutation) => TResult }  function test() {   const options: MutationStateOptions = {     select: (mutation) => {       return mutation.state.data     },   }    const result = options.select?.({} as any)    // ❌ inferred as unknown   return result }
```

### Steps to reproduce

1. Use useMutationState with a select callback
2. Access mutation.state.data inside select
3. Observe the inferred type of the result

### Expected behavior

The type of mutation.state.data should be correctly inferred based on the mutation's generic type (e.g., number, string, etc.), and the returned result should reflect that type (e.g., number[]).

### How often does this bug happen?

None

### Platform

For all

### Tanstack Query adapter

None

### TanStack Query version

@tanstack/react-query v5 (latest)

### Additional context

The type of mutation.state.data is inferred as unknown because Mutation is used without generics. As a result, the final returned value from useMutationState is typed as unknown[].

TkDodo:
`useMutationState` can and will match multiple mutations, so there is no way we can infer the type to something here.

PR #10464, opened 2026-04-12
fix(react-query): preserve Mutation generics in useMutationState select callback

mukeshdhadhariya:
## Summary

#10463 issue

`useMutationState` currently loses type information in the `select` callback because `Mutation` is used without generics. This causes values like `mutation.state.data` to be inferred as `unknown`.

## Changes

- Made `MutationStateOptions` generic
- Propagated generics to `Mutation` in the `select` callback
- Updated `getResult` and `useMutationState` to support generics

## Example

Before:

const result = useMutationState({
  select: (mutation) => mutation.state.data
})
// result: unknown[]

After:

const result = useMutationState<number>({
  select: (mutation) => mutation.state.data
})
// result: number[]

## Impact

- Improves TypeScript type inference
- No runtime behavior changes
- Fully backward compatible

## Motivation

This ensures that type information from core `Mutation` is preserved in the React adapter, improving developer experience when using `select`.

## Summary by CodeRabbit

* **Refactor**
  * Enhanced type inference in mutation state handling for improved developer experience and IDE support. Mutation state callbacks now receive fully typed mutation objects, enabling better autocomplete and type checking in TypeScript projects.

### DIFF
--- a/packages/react-query/src/useMutationState.ts
+++ b/packages/react-query/src/useMutationState.ts
@@ -22,14 +22,34 @@ export function useIsMutating(
   ).length
 }
 
-type MutationStateOptions<TResult = MutationState> = {
+type MutationStateOptions<
+  TData = unknown,
+  TError = unknown,
+  TVariables = unknown,
+  TContext = unknown,
+  TResult = MutationState
+> = {
   filters?: MutationFilters
-  select?: (mutation: Mutation) => TResult
+  select?: (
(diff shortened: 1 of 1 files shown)


TkDodo:
> after
```
const result = useMutationState({
select: (mutation) => mutation.state.data
})
// result: number[]
```

not true. it’s still `unknown[]`  because there is now way to infer from. Also, there’s still a type-error in your implementation. I don’t think this is possible or desirable.

PR #10476, opened 2026-04-13
fix(query-core): propagate AbortSignal reason in infiniteQueryBehavior fetchPage cancellation

### LINKED ISSUE #10475: bug: infiniteQueryBehavior fetchPage rejects with undefined when cancelled instead of propagating AbortSignal reason
### Describe the Bug

In `packages/query-core/src/infiniteQueryBehavior.ts`, the internal `fetchPage` helper function calls `Promise.reject()` **without any argument** (i.e., rejects with `undefined`) when the query has been cancelled via the AbortSignal:

```ts
const fetchPage = async (
  data: InfiniteData<unknown>,
  param: unknown,
  previous?: boolean,
): Promise<InfiniteData<unknown>> => {
  if (cancelled) {
    return Promise.reject()  // rejects with undefined, not a proper error
  }
  // ...
}
```

The `cancelled` flag is set to `true` by the `AbortSignal` abort event listener (via `addConsumeAwareSignal`). When `abortController.abort()` is called in `query.ts` (inside `onCancel`), the signal fires and `cancelled` becomes `true`. On the next `fetchPage` invocation within the multi-page fetch loop, the rejection propagates `undefined` as the error reason rather than a meaningful `AbortError` (i.e., `context.signal.reason`).

### Impact

1. **Custom `retry` functions** receive `undefined` as the error argument instead of an `AbortError`, breaking any logic that inspects the error type.
2. **`onError` / `onSettled` callbacks** in `QueryCache` config receive `undefined` as the error.
3. **Error boundaries** using `throwOnError` may receive `undefined` instead of a real error, leading to confusing error messages.

### Steps to Reproduce

```ts
  queryCache: new QueryCache({
    onError: (error) => {
      console.log("error is:", error)          // logs: "error is: undefined"
      console.log(error instanceof DOMException) // false (should be true)
    }
  })
})

client.fetchInfiniteQuery({
  queryKey: ["items"],
  queryFn: async ({ pageParam }) => {
    await new Promise(r => setTimeout(r, 200))
    return { items: [pageParam], nextCursor: pageParam + 1 }
  },
  getNextPageParam: (lastPage) => lastPage.nextCursor,
  initialPageParam: 0,
  pages: 3,
})

// Cancel while fetching page 2 (after ~250ms, first page done)
setTimeout(() => client.cancelQueries({ queryKey: ["items"] }), 250)
```

### Expected Behavior

`Promise.reject()` should propagate the actual abort reason:

```ts
if (cancelled) {
  return Promise.reject(context.signal.reason)
  // context.signal.reason is a DOMException{ name: "AbortError" }
  // when abortController.abort() is called without an explicit reason
}
```

This matches the Web standard: [`AbortSignal.reason`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/reason) is automatically set to a `DOMException` with name `"AbortError"` when `AbortController.abort()` is called without an argument.

### Environment

- `@tanstack/query-core`: v5.x (latest)
- Affects: `useInfiniteQuery`, `useSuspenseInfiniteQuery`, `prefetchInfiniteQuery`

kuishou68:
## Summary

Fixes #10475

### Problem

In `infiniteQueryBehavior.ts`, the internal `fetchPage` helper function calls `Promise.reject()` **without any argument** when a query cancellation is detected via the `cancelled` flag (which is set by the AbortSignal abort event listener):

```ts
// Before (bug)
const fetchPage = async (data, param, previous) => {
  if (cancelled) {
    return Promise.reject()  // rejects with undefined!
  }
  // ...
}
```

When `query.cancel()` → `abortController.abort()` is called, the abort signal fires and sets `cancelled = true`. On the next iteration of the multi-page fetch loop, `fetchPage` rejects with `undefined` instead of the proper `AbortError` that is set as `context.signal.reason`.

### Fix

Propagate `context.signal.reason` as the rejection value, which is a `DOMException` with name `"AbortError"` when `AbortController.abort()` is called without an explicit reason (per the [AbortSignal spec](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/reason)):

```ts
// After (fix)
const fetchPage = async (data, param, previous) => {
  if (cancelled) {
    return Promise.reject(context.signal.reason)  // proper AbortError DOMException
  }
  // ...
}
```

### Impact

This fixes the following issues that occur when an infinite query is cancelled mid-pagination:

- Custom `retry` functions receiving `undefined` instead of an `AbortError`
- `QueryCache.onError` callbacks receiving `undefined` as the error argument
- Error boundaries via `throwOnError` showing `"undefined"` errors

### Testing

The fix is minimal and targeted. The change ensures that when the AbortSignal fires during a multi-page infinite query fetch, the proper `DOMException{ name: "AbortError" }` is propagated as the rejection reason, consistent with the Web AbortController specification.

## Summary by CodeRabbit

* **Bug Fixes**
  * Cancellation now surfaces the underlying abort reason when an in-flight operation is cancelled, improving error diagnostics and debugging.

* **Tests**
  * Added a unit test to validate that cancellation propagates the abort reason and that fetch behavior correctly rejects with that reason.


TkDodo (commented):
this looks good, but can you please add the steps-to-reproduce from your issue as a test case? Ideally, that test fails on main but works with your PR. Thanks.

kuishou68:
Thanks for the review, @TkDodo! I'll add a test case that reproduces the issue from #10475.

The test will demonstrate that when a query is aborted, the retry function and error handlers receive the proper `AbortError` reason (i.e., `context.signal.reason`) instead of `undefined`. I'll push the test shortly — it will fail on `main` and pass with this PR's change.

PR #10484, opened 2026-04-13
fix(eslint-plugin-query): use TypeFlags from typescript instead of hardcoded values

### LINKED ISSUE #10461: [eslint-plugin-query] `no-void-query-fn` false positive on enum returns with TypeScript 6
### Describe the bug

The `no-void-query-fn` rule  incorrectly reports `queryFn must return a non-undefined value` when a `queryFn` returns a TypeScript enum member. This only occurs when using **TypeScript 6.0**.

On investigating, looks like we're implicitly relying on these values for [`TypeFlags`](https://github.com/TanStack/query/blob/24de260ad26626a6d5420587f97714f6b94ffa2a/packages/eslint-plugin-query/src/rules/no-void-query-fn/no-void-query-fn.rule.ts#L8-L12) which should be internal to TS. 

```ts
const TypeFlags = {
  Void: 16384,
  Undefined: 32768,
} as const

function isIllegalReturn(checker: TypeChecker, type: Type): boolean {
  const awaited = checker.getAwaitedType(type)

  if (!awaited) return false

  if (awaited.isUnion()) {
    return awaited.types.some((t) => isIllegalReturn(checker, t))
  }

  return awaited.flags & (TypeFlags.Void | TypeFlags.Undefined) ? true : false
}
```

The underlying values for `TypeFlags` were reorderd in this change: https://github.com/microsoft/TypeScript/pull/63084. This ends up actually testing for `UniqueESSymbol | EnumLiteral` instead of `Void | Undefined` as intended.

### Steps to reproduce

On TypeScript 6, the below triggers the lint violation:

```ts
import { useQuery } from '@tanstack/react-query'

enum ExampleEnum {
  A,
}

export function useExampleEnum() {
  return useQuery({
    queryKey: ['exampleEnum'],
    queryFn: () => {
      return ExampleEnum.A
    },
  })
}
```

### Expected behavior

Returning an enum value from `queryFn` should not trigger `no-void-query-fn`.

Every time

### Screenshots or Videos

_No response_

### Platform

all

### Tanstack Query adapter

None

### TanStack Query version

v5.99.0

### TypeScript version

v6.0.2

### Additional context

Quick fix addressing this: https://github.com/TanStack/query/pull/10460

mvtandas:
## Summary

Fixes #10461

The `no-void-query-fn` rule used hardcoded values for `TypeFlags.Void` (16384) and `TypeFlags.Undefined` (32768):

```ts
const TypeFlags = {
  Void: 16384,
  Undefined: 32768,
} as const
```

These internal TypeScript flag values were [reordered in TypeScript 6](https://github.com/microsoft/TypeScript/pull/63084), causing the bitwise check to test for `UniqueESSymbol | EnumLiteral` instead of `Void | Undefined`. This resulted in false positives when `queryFn` returns an enum member.

## Fix

Import `TypeFlags` directly from the `typescript` package (already listed as a peer dependency) instead of hardcoding the numeric values. This ensures correct behavior across all supported TypeScript versions (5.x and 6.x).

## Test

All 34 existing `no-void-query-fn` tests pass.

## Summary by CodeRabbit

* **Refactor**
  * Updated internal implementation in eslint-plugin-query to improve code maintainability with no visible user-facing changes.

### DIFF
--- a/packages/eslint-plugin-query/src/rules/no-void-query-fn/no-void-query-fn.rule.ts
+++ b/packages/eslint-plugin-query/src/rules/no-void-query-fn/no-void-query-fn.rule.ts
@@ -1,14 +1,10 @@
 import { ESLintUtils } from '@typescript-eslint/utils'
+import { TypeFlags } from 'typescript'
 import { ASTUtils } from '../../utils/ast-utils'
 import { detectTanstackQueryImports } from '../../utils/detect-react-query-imports'
 import { getDocsUrl } from '../../utils/get-docs-url'
 import type { ParserServicesWithTypeInformation } from '@typescript-eslint/utils'
 import type { ExtraRuleDocs } from '../../types'
-
-const TypeFlags = {
-  Void: 16384,
-  Undefined: 32768,
-} as const
 
 export const name = 'no-void-query-fn'
 


mvtandas:
Friendly ping — this is a one-line fix (removes hardcoded TypeFlags values in favor of importing from typescript). All existing tests pass. Happy to address any feedback!

TkDodo:
we’re gonna go with #10460 which also solves this issue.

PR #10516, opened 2026-04-19
chore: upgrade to solid v2 beta 7

brenelz:


nidhishgajjar:
## Summary
This PR upgrades Solid.js dependencies from v2.0.0-beta.4/5 to v2.0.0-beta.7 and vite-plugin-solid from 3.0.0-next.4 to 3.0.0-next.5 across all Solid-related packages and examples. The primary code changes adapt to Solid v2's stricter ownership model and auto-disposal behavior for memos, requiring use of `createSignal(fn)` instead of `createMemo` in certain contexts and wrapping store writes with `runWithOwner(null, ...)` to escape owned scopes.

## Architecture
The changes affect the Solid query integration layer across:
- `packages/solid-query/src/useBaseQuery.ts` - Core query hook implementation
- `packages/solid-query/src/useIsFetching.ts` - Fetching status tracking
- `packages/solid-query/src/useIsMutating.ts` - Mutation status tracking  
- `packages/solid-query/src/useMutationState.ts` - Mutation state management
- `packages/solid-query/src/useQueries.ts` - Multiple query management

The changes adapt to Solid v2's ownership model where:
1. Memos that aren't reactively read get auto-disposed, cascading to cleanup handlers
2. Signal/store writes within owned scopes throw errors
3. Signal option naming changed from `pureWrite` to `ownedWrite`

## Issues

**[warning]** packages/solid-query/src/useBaseQuery.ts:155-160 — The change from `createMemo` to `createSignal` for `client`, `defaultedOptions`, and `trackedDefaultedOptions` is semantically different. `createMemo` caches values and only recomputes when dependencies change, while `createSignal(fn)` with a function will create a new signal on each access. This could cause unnecessary recomputation or loss of memoization benefits. Verify this doesn't impact performance negatively by adding benchmarking or profiling.

**[warning]** packages/solid-query/src/useBaseQuery.ts:241-289 — The queryResource restructuring removes the `ssrSource: 'client'` option from the resource creation. This option was previously passed to `createMemo`. Ensure SSR behavior is preserved without this option, particularly for streaming scenarios. Add explicit SSR tests to verify hydration still works correctly.

**[warning]** packages/solid-query/src/useBaseQuery.ts:253-260 — The error handling logic is restructured. Previously, `reject` was called outside of `runWithOwner`, now it's potentially inside or removed. Verify error propagation still works correctly, especially for thrown errors during SSR. The early return after `reject` in the new code could skip important cleanup.

**[warning]** packages/solid-query/src/useBaseQuery.ts:195-202 — The `observer.setOptions` call is now wrapped in `runWithOwner(null, ...)`. While this prevents throwing, it means the observer update happens outside the component's ownership scope. This could affect tracking and cleanup if the observer relies on owner context. Verify observer lifecycle management remains correct.

**[suggestion]** packages/solid-query/src/useBaseQuery.ts:155-160 — Consider adding comments explaining why `createSignal(fn)` is used instead of `createMemo`, referencing the specific Solid v2 behavior that necessitates this change. The current comment is good but could reference Solid v2 documentation or issues for future maintainers.

**[suggestion]** packages/solid-query/src/useQueries.ts:248-268 — Similar `runWithOwner` usage as useBaseQuery. Ensure the batched updates in `subscribeToObserver` still work correctly when executed outside the owner scope. The `queueMicrotask` pattern has been removed - verify this doesn't cause timing issues with multiple rapid updates.

**[suggestion]** All changed files — The PR changes significant core logic but doesn't show corresponding test updates. Ensure comprehensive tests cover:
- SSR hydration scenarios
- Error handling and propagation
- Query lifecycle (mount, update, unmount)
- Multiple rapid option changes
- Streaming with deferStream option
- Cleanup on unmount during loading state

**[suggestion]** packages/solid-query/src/useBaseQuery.ts:290 — The comment about NotReadyError and Loading boundary has been simplified. Consider preserving more context about why the try-catch was needed and whether the new implementation handles the same edge cases.

## Cross-file impact
- **Examples**: All Solid examples updated to use beta.7 - users testing with these examples will need Solid v2 beta installed
- **pnpm-lock.yaml**: Massive lockfile update - all consumers will get the new dependency versions
- **Type compatibility**: The peer dependency in `packages/solid-query/package.json` shows `">=2.0.0-beta.0 <3.0.0"` which is correct for supporting v2 betas
- **Backward compatibility**: These changes break Solid v1.x compatibility due to the `runWithOwner` API and `ownedWrite` option. Ensure this is intentional and documented in migration guide

## Assessment
**request-changes** — The changes appear technically sound for adapting to Solid v2's ownership model, but require verification that performance characteristics aren't degraded by replacing `createMemo` with `createSignal`, and that SSR/error handling behavior is preserved. Missing test coverage for the significant refactoring is concerning.

---
_Automated review by [code-review-agent](https://github.com/AWLSEN/code-review-agent). May contain mistakes. Ignore or rebut as you see fit._

birkskyum (approved):

PR #10522, opened 2026-04-19
chore(tsconfig.json): narrow 'include' pattern to prevent TS6053 race conditions from tsup bundled config files

sukvvon:
## 🎯 Changes

Narrow the `include` pattern in all `packages/*/tsconfig.json` files from `*.config.*` to explicit `*.config.ts` and `*.config.js`.

### Why

tsup bundles `tsup.config.ts` into a temporary ESM file named `tsup.config.bundled_<random>.mjs` before loading it. This temporary file exists on disk for a few milliseconds during tsup builds.

All package tsconfigs use `"include": ["*.config.*", ...]`, which is too broad and matches this temporary file. When Nx runs multiple tasks in parallel (e.g., `build` and `test:types`), the TypeScript compiler may enumerate the temporary file, but by the time it tries to read it, tsup has already deleted it. This results in intermittent CI failures such as:

```
error TS6053: File '/.../packages/query-devtools/tsup.config.bundled_<random>.mjs' not found.
```

### Fix

Instead of relying on `exclude` to filter out tsup's temporary files, narrow the `include` pattern itself to only match config files with the extensions actually used in the repo: `.ts` and `.js`. This way, any temporary file with an unexpected name (e.g., `*.bundled_<random>.mjs`) is naturally excluded.

This approach is:
- More robust — automatically handles other build tools that produce similar temporary files (e.g., `tsdown`)
- More explicit — declares exactly which config extensions are valid
- Simpler — no need for an additional `exclude` entry

If `.mjs` or `.cjs` config files are added in the future, they can be added to the `include` array as a deliberate, scoped change.

### Real CI failures

- https://github.com/TanStack/query/actions/runs/24626867439 (first attempt failed, required re-run to pass)
- https://github.com/TanStack/query/actions/runs/24631499262 (same issue after merging \`main\`)
- https://github.com/TanStack/query/actions/runs/25282211546 (#10629 hit the same issue)

### Scope

Applied to all 26 packages using the `*.config.*` include pattern, for consistency.

## ✅ Checklist

- [x] I have tested this code locally with \`pnpm run test:pr\`.

## 🚀 Release Impact

## Summary by CodeRabbit

* **Chores**
  * Standardized TypeScript project includes across packages to explicitly list supported config file extensions (TS/JS and common module variants) and reformatted include blocks. This narrows which config files are considered and improves build consistency, reducing accidental file pickup and making compilation inputs more predictable.


sukvvon:
Confirming this fix is needed — PR #10629 just hit the exact same TS6053 error in CI: https://github.com/TanStack/query/actions/runs/25282211546/job/74121119451

```
error TS6053: File '/.../packages/query-devtools/tsup.config.bundled_kw8jt83pcno.mjs' not found.
  Matched by include pattern '*.config.*' in '/.../packages/query-devtools/tsconfig.json'
```

PR #10530, opened 2026-04-20
fix(vue-query): allow computed ref as queryKey property in queryOptions

### LINKED ISSUE #10525: [vue-query] `queryOptions` rejects `computed` ref as `queryKey` since v5.98.0
### Describe the bug

Since `@tanstack/vue-query` **v5.98.0** (PR #10452), passing a `ComputedRef` (or any reactive value) as `queryKey` to `queryOptions` no longer type-checks. This is the same class of regression as #10458 (which only fixed the `enabled` property), but it affects `queryKey` too — and the fix PR #10465 did not address it.

The symptom is a large `TS2769: No overload matches this call` error. The most revealing part of the error message is:

```
Object literal may only specify known properties, and 'queryKey' does not exist in type
'() => UndefinedInitialQueryOptions_alias_1<unknown, Error, unknown, readonly unknown[]>'.
```

TypeScript picks the getter-style overload (options: () => ...) because `computed(...)` isn't assignable to the plain `QueryKey` position on the object-style overload. This is exactly the same root cause identified in CodeRabbit's comment on #10452: https://github.com/TanStack/query/pull/10452#discussion_r3068236648

The downstream cascade also produces misleading errors about `staleTime` being `StaleTimeFunction<...>` vs `MaybeRefDeep<StaleTimeFunction<...>>`, and `queryKey` missing `dataTagSymbol` / `dataTagErrorSymbol` brand properties.

Minimal code that fails on 5.98.0, 5.99.0, 5.99.1, and 5.99.2:

```ts
import { queryOptions } from "@tanstack/vue-query"
import { computed, type MaybeRefOrGetter, toValue } from "vue"

export const fooQueryOptions = (id: MaybeRefOrGetter<string | null>) =>
    queryOptions({
        // ❌ TS2769 since 5.98.0 — works on 5.97.0
        queryKey: computed(() => ["foo", toValue(id)] as const),
        queryFn: async () => {
            const idValue = toValue(id)
            return idValue ? { id: idValue } : null
        },
    })
```

StackBlitz repro: https://stackblitz.com/edit/tanstack-query-dd82rg7b?file=src%2Fmain.ts (adapt from #10458's repro — replace `enabled: computed(...)` with `queryKey: computed(() => [...])`).

### Steps to reproduce

1. Install `@tanstack/vue-query@5.98.0` (or any version up to `5.99.2`).
2. Write a `queryOptions({ queryKey: computed(() => [...]), queryFn: ... })` call.
3. Run `tsc --noEmit` (or `vue-tsc`).
4. TypeScript reports TS2769 "No overload matches this call" with the message quoted above.

### Expected behavior

`queryOptions` should accept a `ComputedRef` / reactive value for `queryKey`, same as 5.97.0 did and same as what the documentation shows:

https://tanstack.com/query/latest/docs/framework/vue/guides/dependent-queries

The fix should mirror what #10465 did for `enabled` — adjust the input type of `queryOptions` so `queryKey` accepts `MaybeRefDeep<QueryKey>` / `ComputedRef<QueryKey>` in addition to plain `QueryKey`.

Every time.

### Screenshots, code snippets, OS, etc.

**Affected versions:** 5.98.0, 5.99.0, 5.99.1, 5.99.2 (latest).
**Last working version:** 5.97.0.

Verified locally in a project with ~80 `queryOptions` usages: pinning to `5.97.0` produces zero type errors; upgrading to any of 5.98.0 – 5.99.2 yields 83 TS2769 errors, all originating from the same pattern.

TypeScript version: 5.8+
Vue version: 3.5+

KimHyeongRae0:
## 🎯 Changes

Closes #10525.

Fixes a regression from #10452 where the `queryOptions` helper in `@tanstack/vue-query` stopped accepting `computed` refs, `Ref` values, or other reactive values for the `queryKey` property. The earlier fix in #10465 only covered the `enabled` property, so user code like

```ts
queryOptions({
  queryKey: computed(() => ['foo', toValue(id)] as const),
  queryFn: async () => /* ... */,
})
```

was emitting `TS2769: No overload matches this call` on every version since `5.98.0`.

This PR narrows the fix to the `queryKey` property only (mirroring how #10465 special-cased `enabled`), so downstream usages like `queryClient.fetchQuery(options)` — which depend on non-reactive core types for other properties — continue to work without regressions.

Type tests added for both `computed(...)` and `ref(...)` as `queryKey` in `packages/vue-query/src/__tests__/queryOptions.test-d.ts`.

## ✅ Checklist

- [x] I have tested this code locally with `npx nx run @tanstack/vue-query:test:types` and the full `pnpm --filter @tanstack/vue-query test` suite (291 tests passed, no type errors).

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored support for reactive and getter forms for query keys (works with refs, computed values, and getter functions as well as plain arrays).

* **Tests**
  * Added type tests covering reactive and getter forms of query keys.

* **Documentation**
  * Added a release note documenting the fix and clarifying its scope.


KimHyeongRae0:
Thanks for the suggestion. I'm going to leave the assertion as `expectTypeOf(options.queryKey).not.toBeUndefined()` for now — it mirrors the style of the neighboring `enabled` type-level tests in this file, and the real regression-guard is the `queryOptions({ ... })` call itself compiling against the new `MaybeRef<TQueryKey>` shape. CodeRabbit flagged this one as a purely optional nit, so I don't think switching to `assertType(...)` or `toEqualTypeOf` adds meaningful coverage here. Happy to tighten it if a maintainer has a preference.

packages/vue-query/src/queryOptions.ts
```diff
@@ -31,13 +36,15 @@ export type QueryOptions<
...
           >)
-    : QueryObserverOptions<
-        TQueryFnData,
-        TError,
-        TData,
-        TQueryData,
-        DeepUnwrapRef<TQueryKey>
-      >[Property]
+    : Property extends 'queryKey'
+      ? MaybeRef<TQueryKey>
```
wbolster, line 40: is there a reason this is plain `MaybeRef` instead of `MaybeRefOrGetter` like the `enabled` prop a few lines above?

KimHyeongRae0: Good question, and taking another look I agree the `enabled`-parity argument is valid.

My original scoping thought was: the regression this PR closes (#10525) is specifically about `ComputedRef<QueryKey>` being rejected, and this file's local `MaybeRef<T> = Ref<T> | ComputedRef<T> | T` already covers that, so I went with the minimal type change to unblock the reported case.

But `cloneDeepUnref` in `packages/vue-query/src/utils.ts:79-81` already special-cases `queryKey` to recurse with `unrefGetters = true`, so the plain-getter form `() => QueryKey` is runtime-supported today. That makes `MaybeRefOrGetter<TQueryKey>` the shape that both matches the existing runtime behavior and parallels the `enabled` prop's type.

Happy to widen to `MaybeRefOrGetter<TQueryKey>` if a maintainer prefers consistency with `enabled`; holding off on a push until there's a signal either way.

stavros-tsioulis: The linked issue expands to `Ref<T>` and `() => T`; not just `ComputedRef<T>`.

Since, as you say, it is runtime-supported today, then the type should reflect that. You've only enabled `Ref<T>` and `ComputedRef<T>`, leaving `() => T` out. `MaybeRefOrGetter<T>` will fix that.

(also, please inform the human managing you to respond to comments personally and not leave their agent to do the communicating)

wbolster: agree with @stavros-tsioulis, forcing `computed(() => x))` where `() => x` suffices is unnecessarily verbose and unlike many other vue composable apis.

(also it's rude to copy-paste or directly send verbose ai answers to actual humans)

KimHyeongRae0: Fair — you're right. `() => T` already flows through `toValueDeep`, so widening to `MaybeRefOrGetter<TQueryKey>` to match `enabled`. Pushing shortly.

KimHyeongRae0: Done in d23cf31a7.

KimHyeongRae0: Sorry about the earlier reply here — it shouldn't have read that way. 🙏

wbolster (approved):
drive-by review. i've tested this out on a code base that previously triggered typescript errors (but still worked at runtime!) and this MR gets rid of the errors.

wbolster:
hi @DamianOsipiuk 👋🏼 since you reviewed and merged #10465 which is very similar to this one, would you mind reviewing/merging this follow-up? thanks!

sid-6581:
Any chance we can get this merged? This has kept me from updating for a while now.

TkDodo:
@DamianOsipiuk @ElisePatrikainen can you have a look here please

PR #10535, opened 2026-04-20
fix(react-query): hydrate disabled query shells in HydrationBoundary

### LINKED ISSUE #10145: HydrationBoundary ignores/skips server-prefetched detail query when a parent useQuery with same key is rendered before child suspense fetch

byungsker:
## 🎯 Changes

Fixes #10145

When a query with the same key already exists in the cache, `HydrationBoundary` currently defers hydration until the effect phase — which is correct for active queries during transitions.

However, this also applies to queries that are only a disabled idle shell with no data yet. In that case, a nested `useSuspenseQuery` can mount before the prefetched data is applied and unnecessarily suspend or refetch.

This PR makes `HydrationBoundary` hydrate existing queries immediately during render when they are `idle`, disabled, and have no data — treating them as effectively empty shells rather than active queries worth preserving.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `HydrationBoundary` to correctly handle hydration of disabled query shells, ensuring they are properly hydrated during the render phase instead of being deferred to after rendering.

### DIFF
--- a/.changeset/fix-hydration-boundary-disabled-shell.md
+++ b/.changeset/fix-hydration-boundary-disabled-shell.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/react-query': patch
+---
+
+Fix `HydrationBoundary` deferring hydration for disabled query shells
--- a/packages/query-core/src/__tests__/hydration.test.tsx
+++ b/packages/query-core/src/__tests__/hydration.test.tsx
@@ -2,6 +2,7 @@ import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest'
 import { queryKey, sleep } from '@tanstack/query-test-utils'
 import { QueryClient } from '../queryClient'
 import { QueryCache } from '../queryCache'
+import { QueryObserver } from '../queryObserver'
 import { dehydrate, hydrate } from '../hydration'
 import { MutationCache } from '../mutationCache'
 import { executeMutation, mockOnlineManagerIsOnline } from './utils'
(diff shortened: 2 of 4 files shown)


byungsker:
Closing this as a duplicate of #10159, which was already opened by the issue author. Apologies for the noise!

PR #10573, opened 2026-04-25
fix(vue-query): preserve branded types in MaybeRefDeep

### LINKED ISSUE #9920: Branded queryKey in queryOptions causes TS error (TS2769) in useQuery
### Describe the bug

I’m encountering a TypeScript overload mismatch (**TS2769**) when calling `useQuery()` with the result of a `queryOptions(...)` function whose returned object contains a `queryKey` with a branded type. The error occurs specifically when `useQuery()` receives the output of `queryOptions(...)` containing a branded value in `queryKey`.

Passing the same options to `queryClient.fetchQuery(queryOptions(queryKey: brandedKey, queryFn))` — works fine.
Passing the branded key inline to useQuery (i.e. `useQuery({ queryKey: brandedKey, queryFn}))` — works fine.

https://codesandbox.io/p/devbox/tanstack-query-ts-error-2769-forked-spc8jh?file=%2Fsrc%2FPost.vue

### Steps to reproduce

**1.** Define a branded type in your code:

```
type PostId = string & { readonly __brand: "PostId" };
```
**2.** Create a helper function that returns `queryOptions` where the queryKey includes a value using this branded type:

```
const getPostQueryOptions = (
  params: MaybeRefOrGetter<{ postId: PostId }>
) => {
  return queryOptions({
    queryKey: ["post", params],
    queryFn: () => fetcher(toValue(params).postId),
  });
};
```

**3.** Call `useQuery()` using the result of this helper:

```
useQuery(getPostQueryOptions({ postId }));
```

**4.** Observe the TypeScript error:
  Overload 1 of 3, '(options: DefinedInitialQueryOptions<Post, Error, Post, MaybeRefDeep<string | (() => { postId: PostId; }) | { postId: PostId; }>[]>, queryClient?: QueryClient | undefined): UseQueryDefinedReturnType<...>', gave the following error....  ts(2769)

### Expected behavior

`useQuery()` should correctly accept a `QueryOptions` object whose `queryKey` contains a branded (nominal) type.

In other words:
Passing the result of `queryOptions(...)` or any helper function that returns a `QueryOptions` object should not trigger a TypeScript overload error, even if the `queryKey` includes branded values.

Every time

### Screenshots or Videos

Image

Image

### Platform

- OS [Windows]
- Browser [Chrome]

### Tanstack Query adapter

vue-query

### TanStack Query version

v5.85.3

### TypeScript version

v5.9.2

### Additional context

_No response_

byungsker:
## 🎯 Changes

This PR fixes a TypeScript type inference issue where branded types (e.g. `string & { __brand: 'PostId' }`) in `queryKey` tuples cause TS2769 errors when passed from `queryOptions` to `useQuery`.

The root cause was `MaybeRefDeep<T>` recursively mapping branded primitive types as objects, which destroyed the branded type information needed for `DataTag`-based overload resolution.

The fix makes only primitive values terminal in `MaybeRefDeep` (`T extends Primitive`), so branded primitive types pass through unchanged while ordinary objects continue to recursively accept refs.

Regression coverage includes direct and nested branded query keys, plus an ordinary object with a `__brand` property and a nested `ref`.

Fixes #9920

## Review follow-up

This PR incorporates the feedback from @peterbud:

- Replaced the broad `T extends { __brand: infer _ }` terminal condition with `T extends Primitive`, so an ordinary object with a `__brand` property is still recursively unwrapped.
- Added a focused type regression test for `{ __brand: 'post', postId: ref('123') }`.
- The implementation is in [`7d972a00e6c8`](https://github.com/TanStack/query/commit/7d972a00e6c8c929a3611f071050548a3d6b9831), and the focused regression-test follow-up is in [`de74b2e77`](https://github.com/TanStack/query/commit/de74b2e77).
- The nested-ref regression now tests `MaybeRefDeep` directly instead of coupling it to a separate `DataTag`/`useQuery` inference shape.

## ✅ Checklist

## 🚀 Release Impact

## Verification

- `pnpm nx run @tanstack/vue-query:test:types --skip-nx-cache` — passed.
- `pnpm nx run @tanstack/vue-query:test:lib --skip-nx-cache` — passed for Vue 2, Vue 2.7, and Vue 3: 28 test files, 319 passed, 6 skipped, no type errors.
- `pnpm nx run @tanstack/vue-query:test:eslint --skip-nx-cache` — passed with 0 errors; 5 existing cspell warnings.
- `pnpm exec prettier --check packages/vue-query/src/types.ts packages/vue-query/src/__tests__/queryOptions.test-d.ts .changeset/branded-types-fix.md` — passed.
- `pnpm run test:pr` was also run against current `main`; the changed Vue Query targets passed, while the full run stopped at `vue-declaration-emit:test:types` (`@babel/types` unavailable locally) and `root:test:knip` (generated `.tsup` declaration outputs). Neither failure involves files changed by this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed TypeScript inference issues when using branded types, such as branded strings, inside `queryKey` tuples passed through `queryOptions` and `useQuery`.
  * Preserved branded primitive types during deep reactive type mapping, including branded values nested in objects and references.
  * Improved type accuracy for query results involving branded identifiers.
* **Tests**
  * Added TypeScript declaration tests covering direct and nested branded values, query options, reactive references, and inferred query results.


peterbud:
Hi @byungsker , I quickly looked at this and the fix preserves branded primitive types, but it also treats any object with a `__brand` property as a terminal value. That regresses deep ref support for valid structured query keys such as `{ __brand: 'post', postId: ref('123') }`, where `postId` is no longer accepted. I know it might be a rare case, but still...

I think narrowing the terminal condition to branded primitive intersections (`T extends Primitive`), or use a brand detection approach that does not exempt normal objects from recursive unwrapping would be a better solution. Also adding a regression type test for an object containing both `__brand` and a nested ref would be beneficial.

byungsker:
Good catch. The `__brand` check matched regular objects, so nested refs could be left untouched.

I changed it to `T extends Primitive` and added a type test for an object with `__brand` and a nested `ref`. The type tests pass. I'll push the fix.

PR #10580, opened 2026-04-25
docs(vue-query): explain why reactive() is needed for result narrowing

### LINKED ISSUE #9244: narrowing query still has `data` as possible undefined
### Describe the bug

Hello,
I'm using query to fetch some data coming from the server using Vue, the code looks like this:
```ts
const { error, data, status } = useQuery({
    queryKey: ['units', page],
    queryFn: () => axios.get<PaginatedResponse<Unit>>('/api/units?page=' + page.value),
    select: (response) => response.data,
    placeholderData: keepPreviousData,
})
```

per [docs](https://tanstack.com/query/latest/docs/framework/vue/typescript#type-narrowing), if `status` == `success` or `isSuccess` is true, we should have `data` defined with type from the calling `queryFn`. However it appears by [`@tanstack/query-core` type definition](https://github.com/TanStack/query/blob/37eda0d7d9884fef95a6c349b8be15c6846fc07d/packages/query-core/src/query.ts#L114) that this is not the case:

```ts
interface SuccessAction<TData> {
  data: TData | undefined
  type: 'success'
  dataUpdatedAt?: number
  manual?: boolean
}
```

Is this wanted due to uncertainty of the data being returned? If that so, shouldn't this be responsibility of the type definition of the called `queryFn`?

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgVwM4FMCKz1QJ5wC+cAZlBCHAOQACMAhgHaoMDGA1gPQBuOAtAEcc+KgFgAUBImsIzeEgAm9BgBo4wVAGVkrVulSoicALwoM2XHgAUCCXHtwhlgGKMAXHCsBKEwD44AArkIBroAHRQ+hAANtzoVgCsXip2Dk74ANLoeB4A2lQw+jBUALoShF5S4sAknhrauvqoPrbiDnBKDBKcnPYAegD85RJAA

### Steps to reproduce

Simply check in typescript playground, `data` is of type `Ref<number, number> | Ref<undefined, undefined>` even when resolved and inside a `if` check for `success`

### Expected behavior

Per doc, this shouldn't be allowed

Every time

### Screenshots or Videos

_No response_

### Platform

Issue is platform agnostic

### Tanstack Query adapter

vue-query

### TanStack Query version

v5.76.0

### TypeScript version

5.8.3

### Additional context

_No response_

ousamabenyounes:
## Summary

Follow-up to the discussion in #9244 (does not close it).

`docs/framework/vue/typescript.md` shows the type-narrowing example wrapped in `reactive()` but never says why, so the natural reading is that destructuring `useQuery()` directly and then checking `isSuccess` narrows `data` as well. It does not — destructuring yields independent refs, and TypeScript cannot carry a narrowing from `isSuccess` to `data`.

This PR says that explicitly and shows the alternative when `reactive()` is not an option (`data.value !== undefined`), plus two type tests pinning both shapes:

- the non-reactive result exposes `Ref<T> | Ref<undefined>` (the shape already documented just above the example);
- narrowing on `data.value` refines the ref itself to `Ref<T>`.

Docs and tests only, no runtime or type change, so no changeset.

## Note on the previous revision

This branch previously carried a change to `UseBaseQueryReturnType` that made the mapped type explicitly distributive. That was a no-op: `{ [K in keyof TResult]: ... }` is homomorphic and already distributes over the union, and the two formulations are mutually assignable for `QueryObserverResult`, `DefinedQueryObserverResult` and `InfiniteQueryObserverResult`. The type tests added here pass with or without it, so I dropped it instead of shipping a diff with no effect.

The underlying request in #9244 needs the core result union to carry `Ref` in its signature, as noted in that thread. That is breaking for every adapter and out of scope here.

## Summary by CodeRabbit

* **Documentation**
  * Clarified how to use `reactive()` with `useQuery()` for improved TypeScript narrowing.
  * Documented the correct guard for narrowing data when destructuring query results.

* **Tests**
  * Added type coverage confirming accurate handling and narrowing of query data refs without `reactive()`.

### DIFF
--- a/.changeset/broadcast-client-transaction-stuck.md
+++ b/.changeset/broadcast-client-transaction-stuck.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/query-broadcast-client-experimental': patch
+---
+
+fix(broadcast-client): recover from errors thrown while applying an incoming cross-tab message instead of permanently disabling this tab's own broadcasts
--- a/.changeset/cold-islands-move.md
+++ b/.changeset/cold-islands-move.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/vue-query': minor
+---
+
+add new imperitive methods to QueryClient proxy
--- a/.changeset/cute-cloths-switch.md
+++ b/.changeset/cute-cloths-switch.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/preact-query': minor
+---
+
+preact-query: update usePrefetchQuery and usePrefetchInfiniteQuery to use queryClient.query and queryClient.infiniteQuery
--- a/.changeset/dull-masks-melt.md
+++ b/.changeset/dull-masks-melt.md
@@ -1,0 +1,6 @@
+---
+'@tanstack/preact-query': patch
+'@tanstack/react-query': patch
+---
+
+revert: remove NoInfer from useQuery return types
--- a/.changeset/tidy-lions-cache.md
+++ b/.changeset/tidy-lions-cache.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/query-core': patch
+---
+
+Memoize falsy `combine` results when the function and query results are unchanged.
--- a/.changeset/tidy-pandas-wait.md
+++ b/.changeset/tidy-pandas-wait.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/react-query': patch
+---
+
+fix(react-query): don't show optimistic fetching for unsubscribed useQueries
--- a/.changeset/vue-query-narrow-result-type.md
+++ b/.changeset/vue-query-narrow-result-type.md
@@ -1,0 +1,7 @@
+---
+'@tanstack/vue-query': patch
+---
+
+fix(vue-query): preserve discriminated union narrowing in `UseBaseQueryReturnType`
+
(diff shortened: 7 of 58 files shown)


ousamabenyounes:
Thanks @coderabbitai — addressed in `b69f079`. Lifted `suspense` outside the distributive conditional and typed it as `Promise<TResult>`, so awaiting `suspense()` now preserves the parameterized observer shape (`DefinedQueryObserverResult` keeps non-undefined `data`, `InfiniteQueryObserverResult` keeps `fetchNextPage` / `hasNextPage` / `data.pages`). Aligns with the runtime returned by `observer.fetchOptimistic()`.

ousamabenyounes:
Rebased onto `main` and rescoped in `b56dd6c`.

The `UseBaseQueryReturnType` change I had here was a no-op — the mapped type is homomorphic, so it already distributes over the result union, and the two formulations are mutually assignable for `QueryObserverResult`, `DefinedQueryObserverResult` and `InfiniteQueryObserverResult`. The type tests I added pass without it, so I dropped it along with the changeset and the `Fixes #9244` claim.

What is left is docs plus two type tests: the narrowing section now says why `reactive()` is required and shows the `data.value !== undefined` alternative. The actual ask in #9244 needs `Ref` inside the core result union, which is out of scope here.

peterbud:
I have reviewed this PR, and seems to be OK and accurately explaining why `reactive()` is needed in this case.

TkDodo (approved):

PR #10583, opened 2026-04-25
fix query option declaration emit

### LINKED ISSUE #8453: TypeScript errors when using exported queryOptions factory in @tanstack/react-query 5.62.8
### Describe the bug

When using the factory `queryOptions` and exporting the result, TypeScript errors appear:

```
Exported variable 'shopsQuery' has or is using name 'dataTagErrorSymbol' from external module "/node_modules/@tanstack/query-core/build/legacy/hydration-DiTAi-4H" but cannot be named. (ts4023)

Exported variable 'shopsQuery' has or is using name 'dataTagSymbol' from external module "/node_modules/@tanstack/query-core/build/legacy/hydration-DiTAi-4H" but cannot be named. (ts4023)
````

These errors occur in every instance where I export the result of `queryOptions`. Removing the export resolves the issue, but that isn't an acceptable workaround.

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgRwK4FMoE8DyYbAQB2AznAL5wBmUEIcARAAIwCGpbAxgNYD0U6VpxgBaNJiwMAsAChZnYiXgKoA4QEUM2PAUVwAvCi258hUgAoEsuEYkBpdFgBccANoByFWpjuAugBpZcgBKWTCZdAAPSFg4BQ44YCIAN1YAG2AAE00JHTMyQ3FtU0VLa1tsB2c3dyTUjMy-QJkQoA

### Steps to reproduce

1. Install the latest version of @tanstack/react-query (v5.62.8).
2. Create and export a variable using `queryOptions` factory, e.g., `export const shopsQuery = queryOptions(...);`.
3. Observe TypeScript errors as described.

### Expected behavior

The `queryOptions` factory should allow exporting without triggering TypeScript errors.

Every time

### Screenshots or Videos

_No response_

### Platform

- OS [Windows, Linux]

### Tanstack Query adapter

react-query

### TanStack Query version

v5.62.8

### TypeScript version

v5.7.2

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by artemxknpv**
fixes #8453

This fixes TS4023 errors when a consumer exports an inferred `queryOptions` or `infiniteQueryOptions` value while emitting declarations.

The issue is that these helpers returned anonymous intersections like:

```ts
Options & { queryKey: DataTag<...> }
```

When TypeScript tried to emit a consumer `.d.ts`, it had to expand the tagged `queryKey` type and could end up referencing the internal `dataTagSymbol` / `dataTagErrorSymbol` unique symbols directly.

This patch keeps the same runtime behavior and the same tagged `queryKey` inference, but gives those return types exported names instead. That lets declaration emit reference the helper result type instead of expanding the internal tag fields.

The patch covers the adapters that had the same tagged return shape. Svelte only changes `queryOptions`, because its `infiniteQueryOptions` does not add a tagged `queryKey` return type.

I kept the committed regression tests small: they force declaration emit to name exported option values. I also verified the overload branches separately with a temporary consumer project that imports from package roots.

Verification:

- consumer-style declaration emit repro using package root imports for all touched adapters
- TypeScript 5.4, 5.5, 5.6, 5.7, 5.8, 5.9, 6.0 RC, and current
- `test:types:ts54` through `test:types:ts60` for react, preact, vue, solid, angular experimental
- `test:types` for svelte
- `test:eslint`
- `prettier --check`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed TypeScript declaration generation for query and infinite query options across Angular, Preact, React, Solid, Svelte, and Vue integrations, preventing internal type details from appearing in consumer `.d.ts` files.

* **Tests**
  * Added regression coverage for exported query option types and declaration-generation stability.


artemxknpv:
@TkDodo hi, quick check: is this worth a look?

it keeps runtime behavior the same and only names the helper return types so d.ts emit doesn't leak the internal DataTag symbols.

artemxknpv:
@TkDodo hi, sorry for another ping. is this still worth a review?

packages/vue-query/src/infiniteQueryOptions.ts
TkDodo, line 1: I tested this and the fix is fine in general, however, it seems to be enough for TypeScript to only name the DataTag itself:

```
type QueryKeyWithDataTag<
  TQueryFnData = unknown,
  TError = DefaultError,
> = {
  queryKey: DataTag<TQueryKey, TQueryFnData, TError>
}
```

which we don’t need to export and we can add to all return types instead of the inline type:

```diff
-): DefinedInitialDataOptions<TQueryFnData, TError, TData, TQueryKey> & {
-  queryKey: DataTag<TQueryKey, TQueryFnData, TError>
-}
+): DefinedInitialDataOptions<TQueryFnData, TError, TData, TQueryKey> &
+  QueryKeyWithDataTag<TQueryKey, TQueryFnData, TError>
```

would be good if we could use this approach please. Might even be enough to have this type `QueryKeyWithDataTag` once in the query core.

artemxknpv: updated in a34a95c. moved QueryKeyWithDataTag to query-core and switched the overload returns to use it.

Vue still needs local named aliases in queryOptions.ts for declaration emit, but they are not re-exported from the package entry.

TkDodo:
tackling this in https://github.com/TanStack/query/pull/11224

will keep your attribution @artemxknpv 🙏

PR #10592, opened 2026-04-26
fix(solid-query): don't trigger Suspense when data is already cached

### LINKED ISSUE #9955: Solid Suspense is triggered on data access after checking isSuccess when loaded by ensureQueryData
### Describe the bug

Hi there, I'm running into a weird issue where a suspense is being triggered even after checking that `isSuccess` is true, but only if the data was loaded from an `ensureQueryData` call (I've only tested this from a createFileRoute loader, maybe that's somehow related?).

It seems like somehow the Solid resource ends up behind the query. I'm not very familiar with the TanStack or Solid code, but in digging through the debugger in Chrome I found this suspect situation. The getter on the query identifies that the data exists (and while I'm here I can see the `isSuccess` set to true) and tries to load the latest data from the Solid resource:

Image

But the Solid resource doesn't believe the fetcher was resolved yet, leading to it triggering a read which adds to the suspense context instead of just returning the value:

Image

https://stackblitz.com/edit/tanstack-router-lwma9xac?file=src%2Froutes%2Fposts.tsx

### Steps to reproduce

1. Visit the posts page (/posts) in the sandbox, and observe that the `wasMounted` signal remains false.
2. Uncomment the loader section (in src/routes/posts.tsx), and observe that the Suspense fallback is mounted.

### Expected behavior

I expect that the query `data` getter would behave the same no matter how the data was initially loaded.

Every time

### Screenshots or Videos

Image

### Platform

- OS: macOS
- Browser: Safari 26.1, Chrome 142.0

### Tanstack Query adapter

solid-query

### TanStack Query version

5.90.12

### TypeScript version

5.9.3

### Additional context

I realize this seems like a tiny bug, and in almost any case I don't think it'd even be perceptible, but unfortunately the suspense is a significant problem for me because it breaks the iframe that Square's web payments SDK generates.

ousamabenyounes:
## Summary

Fixes #9955.

When data was preloaded via `ensureQueryData` (e.g. from a TanStack Router loader), accessing `state.data` in a Suspense boundary would re-trigger the Suspense fallback even though the data was already in cache and `isSuccess` was `true`.

## Root cause

In `useBaseQuery.ts`, the proxy `data` getter reads `queryResource.latest?.data` whenever `state.data !== undefined`. Per Solid's docs, `.latest` falls back to a `read()`-equivalent (and therefore triggers Suspense) when the resource has no prior resolved value.

In the preloaded-cache case, the resource's fetcher synchronously calls `resolve(...)` because `!observerResult.isLoading`, but the resource doesn't transition out of the initial `pending` state until the next microtask. Reading `.latest` inside that microtask gap suspends the component even though `state.data` already holds the cached value.

## Fix

If the store already has `data` and no fetch is in-flight (`!state.isFetching`), return `state.data` directly without going through `queryResource`. `state.isFetching` is read via `untrack` so it doesn't widen the data subscriber's reactive deps (preserves existing reconcile / memo-churn tests).

When a fetch is in-flight, the existing `queryResource.latest?.data` path is preserved so background-refresh and re-mount semantics are unchanged.

## Verification

- [x] Added regression test in `packages/solid-query/src/__tests__/suspense.test.tsx` that pre-populates the cache via `queryClient.ensureQueryData(...)`, renders inside `<Suspense>`, and asserts the fallback never mounts. Test fails on `main` (Suspense fires), passes with the fix.
- [x] `pnpm exec nx run @tanstack/solid-query:test:lib` — all 310 tests pass.
- [x] `pnpm exec nx run @tanstack/solid-query:test:types` — all TS versions pass.
- [x] `pnpm exec nx run @tanstack/solid-query:test:eslint` — clean.
- [x] Changeset added (`patch` for `@tanstack/solid-query`).

## Test plan

- [x] Regression test reproduces #9955 and now passes.
- [x] All existing solid-query tests still pass (suspense, refetch-on-remount, reconcile referential equality, memo-churn).

Generated by Ora Studio
Vibe coded by ousamabenyounes

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unnecessary Suspense activation when query data is already available (including data preloaded into cache), so components render immediately without showing loading fallbacks.

* **Tests**
  * Added regression tests to ensure Suspense fallbacks do not mount when data is cached or preloaded.

* **Chores**
  * Added a changeset patch note documenting the fix.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/query/pull/10592)

---
_Generated by Ora Studio (with Claude Code)_
_Vibe coded by Ben Younes Ousama_


ousamabenyounes:
Addressed CodeRabbit nitpick in e815276e7: spied on `Page`'s `queryFn` with `vi.fn()` and added `expect(pageQueryFn).not.toHaveBeenCalled()` to assert the background refetch never ran.

PR #10593, opened 2026-04-26
fix(query-core): use built-in NoInfer for generic indexed-access types

### LINKED ISSUE #9937: NoInfer around TData causes TS error with object type
### Describe the bug

We recently updated `@tanstack/react-query` to version 5.90.11 and it introduced type errors, we believe related to the change in #8683

We have a map from an enum value to an interface so that we can have generic data loaders with the correct data type returned (simplified example in the attached TS Playground link) and then use that elsewhere to make other functions generic.

We are now getting this TS error:

```
Argument of type 'NoInfer<DataTypeToEntity[DT]>' is not assignable to parameter of type 'DataTypeToEntity[DT]'.
  Type '[DataTypeToEntity[DT]]' is not assignable to type 'DataTypeToEntity'.
```

I created a very simple reproduction of the issue with `NoInfer` that also shows the issue not happening with the built-in `NoInfer` from TS 5.4 - https://www.typescriptlang.org/play/?#code/KYOwrgtgBAIghgFzgFQJ4AdhQN4CgpQCCAxsQPZggJQC8UA5HKRVfQDT5QAKATmQCZhi1OvXR9Bw+rgC+uXAEsqwHgDMmWEuUrU8BEHAjAAXFADOCHkoDmAblnylCFeuJZeAobs7l+J85Y29nK4wAAe6GQ81AgYWPBIaJjIZACiVAqxtDicANoJKHEAdFosCAC6pqU69gT5iIWYRR6SFaYtXg64APTdUMgAFgpmUMNQCANYsZhQqnzQAEoAirjTWAByZACSIKoqAEwAPMgAfNm5yOUXUOHOIPwjcCCoUAD8UAAMUKYgwABuKnK9lw5BAFnGwHBdEOMGQNzCdwesAaSWAJwAFPwUXFTLC2FAsUhTJsdnseDDsck0hlYvlLicAJS0M56KCg8FwMxmBTWX78XGU4ApdIITKoOnlbKEuDBeTsmKQhD7bIwuG3UBIgqojHS1G45D46XE7a7A4UxJxYU08Ww8qM5k5AjyqCc7m84D85EWqkisUSqUNYJAA

Also doesn't seem to happen with a simpler `type NoInfer<T> = [T] extends [T] ? T : never` approach.

I know this is very much an edge-case but I'm struggling to find a workaround other than casting or passing generics into `useQuery`, neither of which are ideal.

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgVwM4FMCKz1QJ5wC+cAZlBCHAOQACMAhgHaoMDGA1gPRTr2swBaAI458VANwAoSZ05wAyqDAAbYCWDoAJnHSNkIVHAgk4MABbo4m+g1N4w6Q+csBBAAoBJOKmRhIsVAA6SV19OAARG3oAFXtLBEk4OBdWVghkRngAXmo+NIyYKgAaRLg3ck1kfjgcqjAKqsLJQmlZOGiAdwhvJVV1LTgeVEhmSxg4oOlgTNwSPldU9MzEUsZ6EHQALm8YKGmAcykWyWmYWfmyhuqEpLTNLZ29xkPm1rkAWXowJwsrKLsHKZus5Bo4RhgAZZ0AAPBz8LTBGH+eDjQGRBixBzRCAAUUywHGNRWSQA2uiYnFAil8pkALrbalLGBSUnkzHoQLlCCVfj0y7cxqvGRyRTgPoabT7XS4YCsODKCD0e5QUxmGygmDIKDMOBMIwAIwAVuhqh0zBAIajLMBDPr6BhtBBGKrrYwwMh4GB6FB1ugzlBgmlmPApTByUT7XhGHKADzhaI6aFnRiaQxsuIAPgAFNYMXFtvGAJTbLkgG3oONRdnYvEwAl4MnRWkZmotm7eDoE1hmOA5qtxQvEpJwVj2yzphxUxYFTalYcarXOpBrDbbKiMgpUIi6tP9rG4-HjRu0lnD0cQiccrk8mCz+cL7WIEfch51ACMW+I9oie-QNcPDbxiepQtMcbQuCgdaqIS5jqmqPyWKgvpwFKjAynKVpOMCjgomquGWAAbvQyg4COer6pY9qoMA+xoZogRwJIQYsChfoADL0BRyhEnGCYwsmqY-nmDjZrmFIOAW0RFH8DAFr+-51keQGDlkbalDwmqPlQZwsBIQqnOcrCWNEOFct8vGJgJu7CegalJGJ7KSUK4FwCAEAcM+4BOrouHqmgjgoBg2C4AQnbmC6rFoXscoKkquDBCQGT8MATrtDhFn8bogmXtmSAOXERDbCZLBmagvEZoO7bMfAeX-MQOT+cF+BZu2SQiCFADS6B4NsJLaThVC0iU87tfgABijDbFmKktqG5J9jZhYgYWUjqX6i72f8AD8rEwBxXELeJ6DSWJg7bHoyjKM0QA

### Steps to reproduce

The error is shown at the bottom of the TS Playground link, where we attempt to assign data to a type that should match fine.

### Expected behavior

The type should be fine because `NoInfer<SomeType>` should be able to be assigned to `SomeType`.

None

### Screenshots or Videos

_No response_

### Platform

`@tanstack/react-query`: v5.90.11
TypeScript: 5.7.2 (our codebase) / 5.9.3 (TS Playground)

### Tanstack Query adapter

None

### TanStack Query version

5.90.11

### TypeScript version

5.9.3

### Additional context

_No response_

ousamabenyounes:
## Summary

Fixes #9937.

## Root cause

The public `NoInfer<T>` exported from `@tanstack/query-core` is defined as the historical pre-TS-5.4 trick:

```ts
export type NoInfer<T> = [T][T extends any ? 0 : never]
```

When `T` is a generic indexed-access type (e.g. `DataTypeToEntity[DT]` inside a function generic over `DT extends keyof DataTypeToEntity`), TypeScript fails to recognize that `NoInfer<DataTypeToEntity[DT]>` and `DataTypeToEntity[DT]` are the same:

```
Argument of type 'NoInfer<DataTypeToEntity[DT]>' is not assignable to parameter of type 'DataTypeToEntity[DT]'.
  Type '[DataTypeToEntity[DT]]' is not assignable to type 'DataTypeToEntity'.
```

The package already requires TypeScript ≥ 5.4 (lowest CI target is `typescript@5.4`), so the built-in intrinsic `NoInfer` is available everywhere it ships.

## Fix

- Add a tiny `packages/query-core/src/noInfer.ts` that re-exports the built-in `NoInfer` under an `IntrinsicNoInfer<T>` alias, sidestepping the self-shadowing problem when delegating from a same-named export.
- Change `export type NoInfer<T>` in `packages/query-core/src/types.ts` to `IntrinsicNoInfer<T>`.
- Update `packages/vue-query/src/queryClient.ts`: `super.setQueryData<TData>(...)` forwards the `TData` generic explicitly so the stricter built-in `NoInfer<TData>` return type stays satisfied (previously the `unknown` returned from the unparameterized call was passing only because the loose tuple-index trick masked it).

The public `NoInfer` export keeps the same identity, so no consumer-visible API change.

## Verification

Added type test in `packages/react-query/src/__tests__/useQuery.test-d.tsx` (`generic indexed access TData`) that mirrors the issue's repro: `useQuery` inside a function generic over `<TDataType extends DataType>` whose data is then forwarded to `getLabel(_, data: DataTypeToEntity[TDataType])`. Without the fix it errors with the exact `[DataTypeToEntity[DT]]` is not assignable message; with the fix it type-checks.

Local check matrix (all green):

- `pnpm exec nx run @tanstack/query-core:test:lib`
- `pnpm exec nx run @tanstack/query-core:test:types` (TS 5.4–6.0)
- `pnpm exec nx run @tanstack/query-core:test:eslint`
- `pnpm exec nx run @tanstack/react-query:test:lib`
- `pnpm exec nx run @tanstack/react-query:test:types` (TS 5.4–6.0)
- `pnpm exec nx run @tanstack/react-query:test:eslint`
- `pnpm exec nx run @tanstack/vue-query:test:lib`
- `pnpm exec nx run @tanstack/vue-query:test:types` (TS 5.4–6.0)
- `pnpm exec nx run @tanstack/vue-query:test:eslint`
- `pnpm exec nx run @tanstack/{preact,solid,svelte,angular-query-experimental,query-persist-client-core}:test:types`

## Test plan

- [x] Unit + type tests pass on all supported TS versions
- [x] Existing `TData should depend from only arguments, not the result` ts-expect-error test (ensures `NoInfer` still blocks reverse inference) keeps erroring
- [x] Changeset added (`patch` for `@tanstack/query-core` and `@tanstack/vue-query`)

Generated by Ora Studio
Vibe coded by ousamabenyounes

## Summary by CodeRabbit

* **Bug Fixes**
  * Replaced a custom type handling with TypeScript’s built-in NoInfer to fix assignability in generic/indexed-access contexts.

* **Tests**
  * Added TypeScript-only regression tests to verify generic inference and indexed-access behavior.

* **Chores**
  * Published a patch release and added a release note documenting the fix.

---
_Generated by Ora Studio (with Claude Code)_
_Vibe coded by Ben Younes Ousama_


ousamabenyounes:
Thanks @coderabbitai — addressed in d942386ce. Lifted `describe('generic indexed access TData', ...)` out of `describe('initialData', ...)` so it now sits at `describe('useQuery', ...)` level. Test output groups it correctly. Verified with `pnpm exec nx run @tanstack/react-query:test:types`, `test:lib` and `test:eslint` — all green.

ousamabenyounes:
Thanks @coderabbitai — addressed the optional nit in 24c74b9ba. Tried strengthening with both `expectTypeOf(data).toEqualTypeOf<DataTypeToEntity[TDataType] | undefined>()` and `.toMatchTypeOf<...>()` — both fail to compile because vitest's type-relation checks reduce to `Extends<NoInfer<T>, T>` which TS leaves unresolved under the generic `TDataType extends DataType` constraint.

Per your suggestion ("the function's existence already guards the regression"), kept the implicit assertion (`getLabel(props.dataType, data)` must compile) and added a short comment documenting the intent so future readers don't mistake it for a no-op.

Type tests + lint all green locally.

packages/query-core/src/types.ts
```diff
@@ -34,7 +35,7 @@ export type Override<TTargetA, TTargetB> = {
     : TTargetA[AKey]
 }
 
-export type NoInfer<T> = [T][T extends any ? 0 : never]
+export type NoInfer<T> = IntrinsicNoInfer<T>
```
TkDodo: I’m also fine with just removing this type altogether? I know it is exported but people shouldn’t really use our `NoInfer` type if it exists intrinsically anyways!

ousamabenyounes:
@TkDodo done — dropped the public `NoInfer<T>` re-export entirely in 011141dc0 and let consumers + internals use TypeScript's intrinsic `NoInfer` directly. Also removed the `./noInfer.ts` shim since it's no longer needed. Type tests + lint + lib tests pass on TS 5.4–6.0 across query-core, react-query, vue-query, preact-query, solid-query, svelte-query, angular-query-experimental, query-persist-client-core.

TkDodo (approved):

PR #10609, opened 2026-04-28
feat(angular-query): add devtools theme option

grzdev:
## 🎯 Changes

Add `theme` option support to Angular floating devtools.

React, Vue, and Solid devtools already expose the shared `theme?: Theme` option. This adds the same option to Angular floating devtools and updates an existing devtools instance with `setTheme(theme)` when the option changes.

This keeps Angular devtools aligned with the shared query devtools API and sibling framework bindings.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Added theme option support to the Angular floating devtools, enabling light, dark, or system theme selection for the devtools panel. The theme can be applied dynamically to existing devtools instances and is respected on initial load; when unspecified it defaults to the system preference.

### DIFF
--- a/packages/angular-query-experimental/src/devtools/with-devtools.ts
+++ b/packages/angular-query-experimental/src/devtools/with-devtools.ts
@@ -126,6 +126,7 @@ export const withDevtools: WithDevtools = (
                 errorTypes,
                 buttonPosition,
                 initialIsOpen,
+                theme,
               } = devtoolsOptions()
 
               if (!shouldLoadTools) {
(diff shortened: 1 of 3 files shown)


packages/angular-query-experimental/src/devtools/with-devtools.ts
```diff
@@ -142,6 +143,7 @@ export const withDevtools: WithDevtools = (
                 buttonPosition && devtools.setButtonPosition(buttonPosition)
                 typeof initialIsOpen === 'boolean' &&
                   devtools.setInitialIsOpen(initialIsOpen)
+                if (theme) devtools.setTheme(theme)
```
TkDodo: should probably keep the same pattern here:


```suggestion
                theme && devtools.setTheme(theme)
```

grzdev: updated, thanks 👍

PR #10610, opened 2026-04-29
fix(hydration): set dataUpdatedAt when pending query resolves before hydration

### LINKED ISSUE #10603: dataUpdatedAt incorrect for streamed queries that resolve before hydration
### Describe the bug

I broke it in this PR: https://github.com/TanStack/query/pull/10444

`dataUpdatedAt` for `!hasData` defaults to 0. Previously when hydrating a streamed query that had time to resolve before the time of hydration (that is, pending at dehydration, success at hydration), we still called `query.fetch` just to set up a retryer. This part was not necessary, but removing it also means we no longer update the `dataUpdatedAt` to a correct value, so it stays 0.

The hydration path needs to set this when the query was successful. Since we don't know the time it resolved on the server, we should fallback to `Date.now()` which is what the `fetch` was doing before.

We should also verify that no other values are faulty after the `query.fetch` removal.

I plan to fix this asap.

(Update: I haven't verified all details a 100% yet, but I'm pretty confident this was the mechanism that broke it)

N/A

### Steps to reproduce

N/A

### Expected behavior

`dataUpdatedAt` should get set.

None

### Screenshots or Videos

_No response_

### Platform

N/A

### Tanstack Query adapter

None

### TanStack Query version

5.100.1

### TypeScript version

_No response_

### Additional context

_No response_

DoriAlagem:
## Problem

When a query is dehydrated while still pending (streamed) but then resolves *before* hydration runs on the client, the hydration code transitions the query's status from `'pending'` to `'success'` — but it never updates `dataUpdatedAt`. It stays at `0`.

This was introduced in #10444, which removed the `query.fetch()` call that previously set `dataUpdatedAt` as a side effect of resolving the promise.

Steps to reproduce: dehydrate a pending query (e.g. from a React Server Component), resolve it before sending the response, then hydrate on the client. `query.state.dataUpdatedAt` will be `0` instead of a real timestamp.

## Fix

When the pending→success transition is applied (both in the `setState` path for existing queries and in the `queryCache.build` path for new queries), add `dataUpdatedAt: dehydratedAt ?? Date.now()`.

`dehydratedAt` is the right value here — it records when the server serialized the resolved data, which is when the data was actually "updated". Falling back to `Date.now()` handles older dehydration payloads that don't include `dehydratedAt`.

Fixes #10603

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed hydration for streamed/promise queries that resolved while in a pending state, ensuring the hydrated query now records an updated data timestamp for correct freshness tracking.

* **Tests**
  * Added hydration test coverage for both new and pre-seeded cache entries to verify that once such pending queries resolve, `dataUpdatedAt` is set to a valid (non-zero) value.


TkDodo:
please add a test case that covers this scenario - something that fails on `main` but works on your PR.

DoriAlagem:
Added two test cases to `hydration.test.tsx` that verify `dataUpdatedAt > 0` after hydrating a pre-resolved streamed query — one for a fresh cache entry (the `queryCache.build` path) and one for an existing entry (the `query.setState` path). Both fail on `main` and pass with this PR.

packages/query-core/src/hydration.ts
```diff
@@ -253,6 +253,7 @@ export function hydrate(
             ...(state.status === 'pending' &&
               data !== undefined && {
                 status: 'success' as const,
+                dataUpdatedAt: dehydratedAt ?? Date.now(),
```
Ephem, line 256: We will only ever hit this case if `hasNewerSyncData` is `true`. That's only true if we _do_ have `dehydratedAt`, so we should never hit the `?? Date.now()` fallback here.

I still like that it's there though, it's good defensive coding if the logic above ever changes! 👍

---

The existing behavior is that if we try to hydrate a query with the older dehydration format that does not have `dehydratedAt`:

* If there's no existing query, we still hydrate
* If there is an existing query, since we're missing `dehydratedAt`, we can't really determine if the dehydrated query is actually newer, so we just ignore it

This PR doesn't change that, just wanted to spell it out explicitly since this code has become pretty hard to read. 😅

Ephem (approved):
This fix looks good to me, thanks a lot for providing it!

Sorry for taking some time to get to this, and for introducing the bug in the first place. 😓

PR #10644, opened 2026-05-05
fix(devtools): fix crash with custom queryKeyHashFn and async persistence

### LINKED ISSUE #6958: Devtools + async persistence + custom queryKeyHashFn doesn't work
### Describe the bug

When using both persistence and a custom `queryKeyHashFn`, the DevTools will crash on initial load with error:
```typescript
TypeError: Cannot read properties of undefined (reading 'fetchStatus')
    at getQueryStatusColor
```

Busting the cache on each load "fixes" it. And so does removing the custom `queryKeyHashFn`.

Just upgraded from v4, and this worked fine then. I also tried with a createSyncStoragePersister, and that works completely fine as well.

TBD

### Steps to reproduce

1. Wrap any app in PersistQueryClientProvider with an indexdb persister example from the docs
2. Add a custom `queryKeyHashFn` to queryClient `defaultOptions.queries.queryKeyHashFn`. E.g. return `(queryKey)=>hashKey(queryKey) + "_test"`
3. Reload the page

It works when there's no cache being loaded. It doesn't when something is loaded from the cache.

### Expected behavior

Devtools should support a custom queryKeyHashFn with async persistence.

Every time

### Screenshots or Videos

_No response_

### Platform

Macos, latest chrome

### Tanstack Query adapter

react-query

### TanStack Query version

5.22.2

### TypeScript version

_No response_

### Additional context

_No response_

lihan3238:
## Summary

When using async persistence (e.g. IndexedDB) combined with a custom `queryKeyHashFn`, the DevTools crash on initial load with:
```
TypeError: Cannot read properties of undefined (reading 'fetchStatus')
```

The root cause is that `queryCache.find({ queryKey })` internally uses the default `hashQueryKey` to compute the hash and match queries. When a custom `queryKeyHashFn` is configured, this computed hash won't match the query's actual hash in the cache, so `find()` returns `undefined`.

## Fix

Replaced `queryCache.find({ queryKey: props.query.queryKey })` calls with `queryCache.get(props.query.queryHash)` in the DevTools query detail component. The `get()` method looks up queries directly by their already-computed hash string, which correctly finds the query regardless of which hash function was used.

The `props.query.queryHash` is already available in this component since each query entry in the DevTools list carries its `queryHash`.

Closes #6958

## Test plan

- [x] Verified the fix targets the correct `queryCache().find()` calls that use `queryKey` matching
- [ ] Manual testing with a custom `queryKeyHashFn` and async persistence (IndexedDB) recommended

## Summary by CodeRabbit

* **Refactor**
  * Optimized query resolution in the DevTools to improve performance and responsiveness when retrieving and displaying query state information.

### DIFF
--- a/packages/query-devtools/src/Devtools.tsx
+++ b/packages/query-devtools/src/Devtools.tsx
@@ -1384,54 +1384,35 @@ const QueryRow: Component<{ query: Query }> = (props) => {
   const t = (light: string, dark: string) => (theme() === 'dark' ? dark : light)
 
   const queryState = createSubscribeToQueryCacheBatcher(
-    (queryCache) =>
-      queryCache().find({
-        queryKey: props.query.queryKey,
-      })?.state,
+    (queryCache) => queryCache().get(props.query.queryHash)?.state,
     true,
     (e) => e.query.queryHash === props.query.queryHash,
   )
 
   const isDisabled = createSubscribeToQueryCacheBatcher(
     (queryCache) =>
-      queryCache()
-        .find({
-          queryKey: props.query.queryKey,
-        })
-        ?.isDisabled() ?? false,
+      queryCache().get(props.query.queryHash)?.isDisabled() ?? false,
     true,
     (e) => e.query.queryHash === props.query.queryHash,
   )
 
   const isStatic = createSubscribeToQueryCacheBatcher(
     (queryCache) =>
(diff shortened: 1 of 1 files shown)


ifxnas:
Can this be merged? I'm experiencing this now

PR #10668, opened 2026-05-08
React: update usePrefetchQuery to use new methods, plus react adaptor tests and docs

DogPawHat:
## 🎯 Changes
Updates the react usePrefetchQuery hook to use the new query and infiniteQuery methods. Adds docs and extra tests for the hooks and impertive methods as well.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

- **New Features**
  - Updated prefetch hooks to use the latest query execution behavior.
  - Improved type support for transformed query data and infinite-query page parameters.
  - Prefetch hooks now explicitly return no value.

- **Bug Fixes**
  - Prevented unhandled promise rejections during prefetch operations.
  - Added validation to reject unsupported `skipToken` usage in infinite-query prefetching.

- **Documentation**
  - Updated prefetch hook references to reflect the latest query APIs.


packages/react-query/src/usePrefetchQuery.tsx
```diff
@@ -1,19 +1,25 @@
...
 >(
-  options: UsePrefetchQueryOptions<TQueryFnData, TError, TData, TQueryKey>,
+  options: QueryExecuteOptions<TQueryFnData, TError, TData, TQueryKey>,
   queryClient?: QueryClient,
 ) {
   const client = useQueryClient(queryClient)
 
   if (!client.getQueryState(options.queryKey)) {
-    client.prefetchQuery(options)
+    void client.query(options).then(noop).catch(noop)
```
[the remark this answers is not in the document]
TkDodo: wasn’t this the same before @coderabbitai ?

packages/react-query/src/__tests__/usePrefetchQuery.test-d.tsx
TkDodo, line 1: we have this test in every framework adapter it seems. I think we should keep that behaviour for the hook 🤔 . As opposed to the imperative function, we can’t guard it behind an if statement in user-land ....

DogPawHat: Ok so the usePrefetchQuery hooks should not allow skipToken then

TkDodo: I’d say so, yes

packages/react-query/src/usePrefetchInfiniteQuery.tsx
```diff
@@ -1,30 +1,33 @@
...
     TPageParam
   >,
   queryClient?: QueryClient,
-) {
+): void {
   const client = useQueryClient(queryClient)
 
   if (!client.getQueryState(options.queryKey)) {
-    client.prefetchInfiniteQuery(options)
+    void client.infiniteQuery(options).then(noop).catch(noop)
```
TkDodo: same question here: since we `void`, the `.then(noop)` seems unnecessary?


```suggestion
    void client.infiniteQuery(options).catch(noop)
```

packages/react-query/src/usePrefetchQuery.tsx
```diff
@@ -1,19 +1,28 @@
...
+    TQueryKey
+  >,
   queryClient?: QueryClient,
-) {
+): void {
   const client = useQueryClient(queryClient)
 
   if (!client.getQueryState(options.queryKey)) {
-    client.prefetchQuery(options)
+    void client.query(options).then(noop).catch(noop)
```
TkDodo:
```suggestion
    void client.query(options).catch(noop)
```

packages/react-query/src/usePrefetchInfiniteQuery.tsx
DogPawHat: changed

TkDodo (approved):

PR #10696, opened 2026-05-11
fix(svelte-query): reuse QueriesObserver to fix createQueries empty-init bug

### LINKED ISSUE #10681: [svelte-query] `createQueries` becomes permanently inactive when initialized with an empty `queries` array
### Describe the bug

When `createQueries` is called with an accessor whose `queries` array is empty at first render, the resulting query slots stay `pending` / `data: undefined` forever — even after the input array is later updated to a non-empty value. Subsequent reference updates do not revive any slot.

The same component works correctly if the initial `queries` array contains at least one entry.

https://svelte.dev/playground/15fedd431f424b7f9d1aac624ac8426a?version=5.55.5

### Steps to reproduce

1. Open the playground.
2. Section (1) starts with `[0]` and resolves to `success` after ~300ms.
3. Section (2) starts with `[]` and switches to `[0]` inside `onMount`. It stays `pending` indefinitely.
4. Click `Add page` under section (2). Nothing changes — the instance is permanently inactive. (In Section (1), the same operation results in a normal change.)

### Expected behavior

Section (2) should behave like section (1) once the queries array has at least one entry.

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: MacOS
- Browser: anything
- Svelte: `5.55.4`

### Tanstack Query adapter

svelte-query

### TanStack Query version

v6.1.18

### TypeScript version

v6.0.3

### Additional context

_No response_

mixelburg:
## What

Fixes `createQueries` becoming permanently inactive when initialized with an empty `queries` array.

Closes #10681.

## Why

The current implementation recreates `QueriesObserver` on every query change via `$derived`. This breaks the observer/subscription lifecycle: when `createQueries` starts with `[]`, the observer is created with no queries, and when the array later becomes non-empty, a new observer is created but the subscription doesn't properly pick up the new queries.

## How

Adopts the same pattern as `createBaseQuery`:

1. **Create observer once** with `$state` instead of `$derived`
2. **Recreate only on client change** via `watchChanges`
3. **Call `setQueries`** when resolved options change via `watchChanges`

This matches the React implementation where the observer is created once in `useState` and updated via `setQueries` in a `useEffect`.

## Testing

The existing svelte-query tests still pass. The fix can be verified by using `createQueries` with an initially empty array that is later populated — previously this stayed pending forever; now it resolves correctly.

## Release Notes

* **Refactor**
  * Improved internal query observation mechanism to ensure more consistent and reliable query state updates. Enhanced observer lifecycle management for better query synchronization across state changes.

### DIFF
--- a/packages/svelte-query/src/createQueries.svelte.ts
+++ b/packages/svelte-query/src/createQueries.svelte.ts
@@ -2,6 +2,7 @@ import { QueriesObserver } from '@tanstack/query-core'
 import { useIsRestoring } from './useIsRestoring.js'
 import { createRawRef } from './containers.svelte.js'
 import { useQueryClient } from './useQueryClient.js'
+import { watchChanges } from './utils.svelte.js'
 import type {
   Accessor,
   CreateQueryOptions,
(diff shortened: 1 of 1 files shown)


mixelburg:
Closing due to inactivity — over 5 weeks with no review or feedback. Happy to reopen if there's interest.

PR #10699, opened 2026-05-12
chore(pnpm): update pnpm to v11

harry-whorlow:
## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "type": "git",
     "url": "git+https://github.com/TanStack/query.git"
   },
-  "packageManager": "pnpm@10.24.0",
+  "packageManager": "pnpm@11.1.0",
   "type": "module",
   "scripts": {
     "clean": "pnpm --filter \"./packages/**\" run clean",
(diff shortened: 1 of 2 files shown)


pnpm-lock.yaml
```diff
@@ -4,25 +4,6 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
-overrides:
```
TkDodo, line 7: hmm any idea where the overrides went @harry-whorlow ?

harry-whorlow: Thats, strange... when updating to v11 it forces you to reinstall, hence the changes to the lock file. 

I noticed there's no overrides in the root pnpm workspace, they're located in the package.json, perhaps that's related.

harry-whorlow: ah... it appears my [suspicions were correct](https://github.com/pnpm/pnpm/issues/11536)

TkDodo: So the overrides need to move from package.json into the pnpm config file!

harry-whorlow: yeah, I added it in [fix: relocate overides to workspace](https://github.com/TanStack/query/pull/10699/commits/979701626183a4d5ea680e6f5c549c0853403fd8).

I'm just fixing the Vue devtools build, was this only broken in my pr?

harry-whorlow:
Waiting on https://github.com/TanStack/query/pull/10700 so the ci runs again.

package.json
```diff
@@ -5,7 +5,7 @@
     "type": "git",
     "url": "git+https://github.com/TanStack/query.git"
   },
-  "packageManager": "pnpm@10.24.0",
+  "packageManager": "pnpm@11.1.0",
```
DamianOsipiuk, line 8: For better security we should use `packageManager` with `sha`

Something like:
```
  "packageManager": "pnpm@11.1.1+sha512.d1fdf5f73c617b64fa1a56a81c3c8dfe0e966e33a6010aa256b517ae77be21d93e05affc0de1a83b0e4f29d569f68b446ae8f068cd7247c0bb3df0fb4d7bdf9a"
```

PR #10720, opened 2026-05-17
fix(query-core): prevent optimistic fetching when unsubscribedFix/subscribed isfetching

### LINKED ISSUE #10718: isFetching becomes true when subscribed=false even though no fetch is running
### Describe the bug

When using subscribed to disable queries on unfocused screens (as recommended in the React Native docs), useQuery can report:
`isFetching === true`

even though:
- no network request is executed
- fetchStatus remains "idle"
- the query cache state does not change

This creates an inconsistent state where the hook reports fetching activity that is not actually happening.

The issue appears when:
`enabled: true
subscribed: false`

This is especially problematic for React Native apps using the officially documented pattern for disabling updates on out-of-focus screens.

https://codesandbox.io/p/sandbox/gz3rj5

### Steps to reproduce

1. Run the example
2. Wait until the initial query finishes
3. Toggle subscribed from true to false
4. Observe the returned query state

### Expected behavior

When:
`subscribed === false`

and no request is running:
`isFetching === false
fetchStatus === "idle"`

The hook state should remain consistent with the query cache state.

Every time

### Screenshots or Videos

https://github.com/user-attachments/assets/19b8b400-a3cb-4ec5-b117-d537f6c5542f

### Platform

React Native / React
Chrome 130.0.0.0 (macOS)

### Tanstack Query adapter

react-query

### TanStack Query version

v5.100.10

### TypeScript version

_No response_

### Additional context

_No response_

Prasadzoman:
## Problem

When using the `subscribed` option (common React Native pattern for unfocused screens), `isFetching` could become `true` even though:

* no fetch was running
* `queryFn` was not called
* no observer was attached
* no query state changed

This resulted in an inconsistent state where:

```ts
fetchStatus === 'fetching'
isFetching === true
```

while the query was actually idle.

Fixes #10718.

---

## Root Cause

`QueryObserver.createResult` optimistically injected `fetchState(...)` whenever:

```ts
!mounted && shouldFetchOnMount(...)
```

However, `!mounted` was true both for:

* initial optimistic renders before subscription
* observers that were previously subscribed and later unsubscribed (`subscribed: false`)

This caused optimistic fetching state to appear even after unsubscribing.

---

## Fix

Adds tracking for whether the observer has ever been subscribed via `#everSubscribed`.

Optimistic fetch state is now only applied for genuine pre-subscription optimistic renders.

Additionally:

* `updateResult` now accepts an optimistic flag
* `getOptimisticResult` passes the correct optimistic state
* `#notify` skips cache notifications when no listeners exist

---

## Tests

Added regression coverage for:

* `isFetching`
* `fetchStatus`
* `subscribed: false`
* toggling subscription state

All existing tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fetch status reporting for unsubscribed queries, now correctly reporting `fetchStatus: 'idle'` and `isFetching: false` when appropriate.

* **Refactor**
  * Enhanced internal query observer logic to optimize subscription tracking and result creation behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/query/pull/10720?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/query-core/src/queryObserver.ts
+++ b/packages/query-core/src/queryObserver.ts
@@ -67,7 +67,7 @@ export class QueryObserver<
   #refetchIntervalId?: ManagedTimerId
   #currentRefetchInterval?: number | false
   #trackedProps = new Set<keyof QueryObserverResult>()
-
+  #everSubscribed = false 
   constructor(
     client: QueryClient,
     public options: QueryObserverOptions<
@@ -95,7 +95,7 @@ export class QueryObserver<
   protected onSubscribe(): void {
     if (this.listeners.size === 1) {
       this.#currentQuery.addObserver(this)
-
+      this.#everSubscribed = true
       if (shouldFetchOnMount(this.#currentQuery, this.options)) {
         this.#executeFetch()
       } else {
@@ -176,7 +176,7 @@ export class QueryObserver<
     }
 
     const mounted = this.hasListeners()
-
+   
     // Fetch if there are subscribers
     if (
       mounted &&
@@ -191,7 +191,7 @@ export class QueryObserver<
     }
 
     // Update result
-    this.updateResult()
+    this.updateResult(mounted ? undefined : false)
 
     // Update stale interval if needed
     if (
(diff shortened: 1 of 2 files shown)


TkDodo:
issue is already fixed

PR #10765, opened 2026-05-23
fix(query-core): snapshot resetQueries match set before mutating state

### LINKED ISSUE #10705: resetQueries method does not trigger refetch on errored queries
### Describe the bug

Hello !

I encountered a behavior of `resetQueries` that i think is not an intended behavior. In my usecase, i tried to reset all my errored queries with this code:

```typescript
void queryClient.resetQueries({
  predicate: (query) => query.state.status === "error",
});
```

and what i found out is the queries were reset, but they never got refetched, and stayed stuck in their initial `pending` status.

The reason is in the source code:

```typescript
resetQueries<TTaggedQueryKey extends QueryKey = QueryKey>(
  filters?: QueryFilters<TTaggedQueryKey>,
  options?: ResetOptions,
): Promise<void> {
  const queryCache = this.#queryCache

  return notifyManager.batch(() => {
    queryCache.findAll(filters).forEach((query) => {
      query.reset()
    })
    return this.refetchQueries(
      {
        type: 'active',
        ...filters,
      },
      options,
    )
  })
}
``` 

The initial step of resetting queries leads to their status being changed to `pending` (or `success` if they have initial data), so when we get to the refetch step, the queries we intended to reset + refetch, are not errored anymore. Applying the "only errored queries" filter on the, now changed, queries means we don't refetch any queries.

This also impacts successful queries (if they don't have initial data), as well as any query attribute that is susceptible to mutate during the operation.

The same pattern is found in `invalidateQueries` but since the only field it mutates - as far as I understand - is `isInvalidated` the impact is much smaller.
I added it in my reproduction app, but as one can get imagine it's pretty wonky to filter on that attribute.

https://stackblitz.com/github/alZyad/reset-errored-queries-no-refetch/tree/stackblitz?file=src%2FResetQueriesReproduction.jsx

### Steps to reproduce

1- reset errored queries by filtering on their status
2- queries which are initially errored stay pending forever

### Expected behavior

queries which are initially errored are reset to pending status, but also get refetched.

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: MacOS
- Browser: Arc
- Version: 1.146.0 (80402)
- Chromium Engine Version 148.0.7778.97

### Tanstack Query adapter

None

### TanStack Query version

5.99.2

### TypeScript version

_No response_

### Additional context

I opened a PR with a fix suggestion in case my analysis of the issue is correct !
https://github.com/TanStack/query/pull/10704

spokodev:
Closes #10705.

## What was broken

`queryClient.resetQueries({ predicate: q => q.state.status === 'error' })` reset the matched queries but never re-fetched them — they stayed stuck in `pending` forever.

## Root cause

`resetQueries` ran the caller's filter twice:

1. First, to pick the queries to reset.
2. Second, inside the `refetchQueries({ type: 'active', ...filters })` call, to pick the queries to re-fetch.

`query.reset()` mutates each query's state — for example, the status flips from `"error"` to `"pending"`. When `filters` contains a state-based predicate like `q => q.state.status === 'error'`, the second pass evaluates against the already-mutated state and the matched set collapses, so the re-fetch step silently does nothing.

The original issue author traced this exactly: <https://github.com/TanStack/query/issues/10705>

## Fix

Snapshot the matched queries _before_ calling `reset()`, then refetch the same set by identity (`matchedQueries.includes(query)`), still scoped to `type: 'active'`. Query references are stable across `reset()`, so the identity predicate is a faithful replacement for the caller's filter without the state-mutation hazard.

```ts
return notifyManager.batch(() => {
  const matchedQueries = queryCache.findAll(filters)
  matchedQueries.forEach((query) => {
    query.reset()
  })
  return this.refetchQueries(
    {
      type: 'active',
      predicate: (query) => matchedQueries.includes(query),
    },
    options,
  )
})
```

## Test

Added a regression test in `queryClient.test.tsx` covering the exact scenario from the issue: two active queries (one errored, one successful), call `resetQueries({ predicate: status === 'error' })`, assert the errored query's `queryFn` was called once and the successful one's wasn't.

- Fails on `main` (errored `queryFn` called `0` times — bug behaviour).
- Passes with the fix.

Full `queryClient.test.tsx` suite: 108 passed (one new test added). Lint clean on changed files.

## Note on scope

The same two-pass pattern exists in `invalidateQueries`, but `invalidate()` only flips `isInvalidated` (which the caller's predicate is unlikely to key on) and the `refetchType: 'none'` short-circuit covers the obvious case. Leaving that alone to keep this PR scoped to the reported bug.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed query reset with predicates to correctly refetch only matching queries, preventing unintended query refetches due to state mutations.

* **Tests**
  * Added regression test for predicate-based query reset and refetch behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/query/pull/10765?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/query-core/src/__tests__/queryClient.test.tsx
+++ b/packages/query-core/src/__tests__/queryClient.test.tsx
@@ -1692,6 +1692,77 @@ describe('queryClient', () => {
       expect(queryFn2).toHaveBeenCalledTimes(0)
       expect(didSkipTokenRun).toBe(false)
     })
+
+    it('should refetch queries selected via a state-based predicate', async () => {
+      // Regression test for #10705 — when `resetQueries` is called with a
+      // predicate keyed on `query.state.status`, reset() mutates the status
+      // before refetchQueries re-evaluates the predicate, so the matched set
+      // collapses and no refetch happens.
+      const erroredKey = queryKey()
+      const successKey = queryKey()
+      const erroredQueryFn = vi
+        .fn<(...args: Array<unknown>) => Promise<string>>()
+        .mockRejectedValue(new Error('boom'))
+      const successQueryFn = vi
+        .fn<(...args: Array<unknown>) => string>()
+        .mockReturnValue('ok')
+
+      // Seed the queries directly (fetchQuery returns a Promise — works under
+      // vi.useFakeTimers because no setTimeout is involved).
+      await queryClient
+        .fetchQuery({
+          queryKey: erroredKey,
+          queryFn: erroredQueryFn,
+          retry: false,
+        })
+        .catch(() => undefined)
+      await queryClient.fetchQuery({
+        queryKey: successKey,
+        queryFn: successQueryFn,
+      })
+      expect(queryClient.getQueryState(erroredKey)?.status).toBe('error')
+      expect(queryClient.getQueryState(successKey)?.status).toBe('success')
+
+      // Mount observers so the queries are "active" — refetch only runs for
+      // queries with observers. The observers may also auto-refetch on mount
+      // because the data is stale; the test re-resets the mocks AFTER that
+      // settles so the only thing the assertion counts is the resetQueries
+      // refetch under test.
+      const erroredObserver = new QueryObserver(queryClient, {
+        queryKey: erroredKey,
+        queryFn: erroredQueryFn,
+        retry: false,
+      })
+      const successObserver = new QueryObserver(queryClient, {
+        queryKey: successKey,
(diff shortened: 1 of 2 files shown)


spokodev:
Quick check-in on this since 2026-05-23. Would a quick look be possible when convenient? Ready to address review feedback if anything surfaces. Thanks!

PR #10773, opened 2026-05-24
feat(query-core): pass failureCount to QueryCache onError callback

### LINKED ISSUE #10713: Enhancement: add retryAttempt to query error callback context
## Feature request

When a query fails with retries enabled, the error callback receives the final error but does not know which retry attempt it occurred on. This makes it difficult to implement differentiated error handling (e.g., show a different message after 3 retries vs first failure).

## Proposed solution

Include `retryAttempt: number` in the context passed to the error callback, indicating which retry attempt caused the failure.

n-satoshi061:
## Summary

Closes #10713.

Adds a third `failureCount` parameter to `QueryCacheConfig.onError`, indicating how many retry attempts occurred before the final failure:

- `0` — the initial attempt failed (no retries happened)
- `n` — `n` retries were exhausted before giving up

This makes it easy to implement differentiated error handling without having to derive the count from `query.state.fetchFailureCount`:

```ts
new QueryCache({
  onError: (error, query, failureCount) => {
    if (failureCount === 0) {
      toast.error('Request failed')
    } else {
      toast.error(`Request failed after ${failureCount} retries`)
    }
  },
})
```

The value is consistent with the `failureCount` convention already used in `ShouldRetryFunction` and `RetryDelayFunction`.

## Test plan

- [x] Updated existing `onError` test to assert `failureCount = 0` when no retries occur
- [x] Added new test asserting `failureCount` equals the configured `retry` count when retries are exhausted
- [x] All `@tanstack/query-core` tests pass (527 tests)
- [x] All `@tanstack/react-query` tests pass (542 tests)

## Release Notes

* **New Features**
  * Error callbacks now receive a `failureCount` parameter indicating the number of retry attempts before the final error occurred. A value of 0 means no retries were attempted. This enables conditional error handling logic based on retry behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/query/pull/10773?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/feat-query-cache-on-error-failure-count.md
+++ b/.changeset/feat-query-cache-on-error-failure-count.md
@@ -1,0 +1,21 @@
+---
+"@tanstack/query-core": minor
+---
+
+feat(query-core): pass `failureCount` to `QueryCache` `onError` callback
+
+The `onError` callback in `QueryCacheConfig` now receives a third argument `failureCount: number` indicating how many retry attempts occurred before the final failure (0 means no retries happened).
+
+This allows differentiated error handling based on the retry count:
+
+```ts
+new QueryCache({
+  onError: (error, query, failureCount) => {
+    if (failureCount === 0) {
+      toast.error('Request failed')
+    } else {
+      toast.error(`Request failed after ${failureCount} retries`)
+    }
+  },
+})
+```
--- a/packages/query-core/src/__tests__/queryCache.test.tsx
+++ b/packages/query-core/src/__tests__/queryCache.test.tsx
@@ -325,11 +325,31 @@ describe('queryCache', () => {
       })
       await vi.advanceTimersByTimeAsync(100)
       const query = testCache.find({ queryKey: key })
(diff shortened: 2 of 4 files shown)


n-satoshi061:
I'm aware that `query.state.fetchFailureCount` already exposes this value, so let me explain the motivation.

The driving use case from #10713 is differentiated error handling — showing a different message on first failure vs. after retries are exhausted. While `query.state.fetchFailureCount` technically makes this possible, surfacing it as a direct callback argument has a couple of ergonomic advantages:

1. **Consistency with existing API** — `ShouldRetryFunction` and `RetryDelayFunction` already receive `failureCount` as a direct argument. `QueryCache.onError` being the odd one out that requires drilling into `query.state` feels like an inconsistency.
2. **Discoverability** — the value is immediately visible in the function signature without knowing the shape of `query.state`.

That said, I understand the concern about growing the API surface. Would you consider this worth adding, or do you feel the existing `query.state.fetchFailureCount` is sufficient and this should stay closed?

PR #10777, opened 2026-05-24
fix(query-core): shallowEqualObjects ignores key-set differences when counts match

francisjohnjohnston-web:
## What
`shallowEqualObjects` now returns `false` when two objects have the same number of keys but different key *names* (e.g. `{ a: 1, b: undefined }` vs `{ a: 1, c: undefined }`).

## Why
The function only iterated over `a`'s keys and compared `a[key] !== b[key]`. A key present on `a` but missing from `b` reads as `undefined` on `b`; if `a`'s value for that key is also `undefined` (or otherwise equal), the objects compared as equal despite having different key sets — violating the documented "shallow compare objects" contract. Real call site: `QueryObserver.setOptions` gates the `observerOptionsUpdated` notification with this comparison, so swapping one defaulted option key for another between renders could silently suppress that notification.

## Safety
Pure utility change; an additive own-property guard. No existing test asserted the old (incorrect) behaviour. Result-object comparisons (which use fixed-shape objects) are unaffected, so there is no behavioural or performance regression for the common path.

## Testing
Added `should return 'false' for same-length objects with different keys` to `utils.test.tsx`: it fails on the current code (`expected true to deeply equal false`) and passes after the fix. Full query-core runtime suite green (**480/480**), typecheck clean.

## Impact
Tighter, contract-correct shallow equality — restores correct change-detection for option updates that swap keys without changing key count.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed object comparison logic to correctly identify differences when objects contain different properties, improving accuracy of equality checks.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/query/pull/10777?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/query-core/src/utils.ts
+++ b/packages/query-core/src/utils.ts
@@ -330,7 +330,7 @@ export function shallowEqualObjects<T extends Record<string, any>>(
   }
 
   for (const key in a) {
-    if (a[key] !== b[key]) {
+    if (!hasOwn.call(b, key) || a[key] !== b[key]) {
       return false
     }
   }
--- a/packages/query-core/src/__tests__/utils.test.tsx
+++ b/packages/query-core/src/__tests__/utils.test.tsx
@@ -59,6 +59,17 @@ describe('core/utils', () => {
     it('should return false if b is undefined', () => {
       expect(shallowEqualObjects({ a: 1 }, undefined)).toEqual(false)
     })
+
+    it('should return `false` for same-length objects with different keys', () => {
+      // Both objects have the same number of keys, but the key sets differ.
+      // `b` is missing on the second object and `c` is missing on the first,
+      // so they are not shallowly equal even though both differing values
+      // happen to be `undefined`.
+      expect(
+        shallowEqualObjects({ a: 1, b: undefined }, { a: 1, c: undefined }),
+      ).toEqual(false)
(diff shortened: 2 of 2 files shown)


packages/query-core/src/utils.ts
```diff
@@ -330,7 +330,7 @@ export function shallowEqualObjects<T extends Record<string, any>>(
   }
 
   for (const key in a) {
-    if (a[key] !== b[key]) {
+    if (!hasOwn.call(b, key) || a[key] !== b[key]) {
```
TkDodo, line 333: `for...in` also iterates over enumerable inherited properties. With this change, an object that has such a property can compare as unequal to itself:

```ts
const value = Object.create({ inherited: 1 })

shallowEqualObjects(value, value) // false
```

Could we iterate over Object.keys(a) instead? This would match the length check:

```
for (const key of Object.keys(a)) {
  if (!hasOwn.call(b, key) || a[key] !== b[key]) {
    return false
  }
```

francisjohnjohnston-web: Good catch — you’re right. Using "for...in" here would include inherited enumerable properties, so combining it with the own-property check can make an object compare unequal to itself.

I’ll switch the loop to "Object.keys(a)" so it matches the existing length check and only compares own enumerable properties, and I’ll add a regression test covering the inherited-property case. Thanks for spotting this.

PR #10791, opened 2026-05-25
feat(lit-query): add render method to query controllers

EskiMojo14:
## 🎯 Changes

see #10711 - adds a render method based on the query status, based on the [Tasks API](https://lit.dev/docs/data/task/#rendering-tasks) 

```ts
  render() {
    return this.repo.render({
      pending: () => html`Loading...`
      error: (query) => html`Error: ${query.error.message}`
      success: (query) => html`
        <h1>${query.data.name}</h1>
        <p>${query.data.description}</p>
        <strong>${query.data.stargazers_count} stars</strong>
      `
    })
  }
```

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Added a typed `render(...)` method to query, mutation, and infinite-query controllers for `pending`, `error`, and `success` states.
  * Added shared, type-safe rendering utilities.
* **Documentation**
  * Added Lit Query rendering guides, examples, and API references.
* **Tests**
  * Added coverage for state-based rendering and missing-renderer behavior.
* **Chores**
  * Added a minor release changeset entry.


packages/lit-query/src/tests/infinite-and-options.test.ts
```diff
@@ -506,6 +506,128 @@ describe('createInfiniteQueryController', () => {
...
+
+    const successUi = infinite.render({
+      pending: () => 'pending-ui',
+      success: (result) => `success-${result.data?.pages.join(',')}`,
+      error: () => 'error-ui',
+    })
+    expect(successUi).toBe('success-0')
+
+    infinite.destroy()
+  })
```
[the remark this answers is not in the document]
EskiMojo14: there seems to  be a mix of whether tests do this or not - i think ideally it'd be cleaner to register the cleanup immediately after setup, using `onTestFinished`

packages/lit-query/src/createQueryController.ts
```diff
@@ -374,6 +385,11 @@ export function createQueryController<
       refetch: controller.refetch,
       suspense: controller.suspense,
       destroy: () => controller.destroy(),
+      render: <
+        TRenderers extends ResultRenderers<QueryObserverResult<TData, TError>>,
+      >(
+        renderers: TRenderers,
+      ) => renderResult(controller.current, renderers),
```
chughgaurav: Could this use `controller.readCurrent()` like the accessor above? `controller.current` only returns the last result pushed by the observer subscription, so observer changes that do not notify the controller (for example, `notifyOnChangeProps: []`) leave `render()` stale while `query()` returns the current result. The same issue exists in `createInfiniteQueryController`; using `renderResult(controller.readCurrent(), renderers)` at both call sites keeps the APIs consistent.

packages/lit-query/src/index.ts
```diff
@@ -70,3 +70,5 @@ export type {
   QueryControllerOptions,
   QueryControllerResult,
 } from './types.js'
+
+export { renderResult } from './render.js'
```
chughgaurav, line 75: Could we also export `ResultRenderers` and `RendererResult` from the package entry point? They appear in the signatures of the public accessors and `renderResult`, but consumers cannot import or name them from `@tanstack/lit-query`.

docs/framework/lit/guides/mutations.md
```diff
@@ -164,3 +164,21 @@ private readonly favoriteMutation = createMutationController(
...
+For convenience, the mutation accessor also includes a `render` method, based on the [Task API](https://lit.dev/docs/data/task/#rendering-tasks). It takes an object of renderers for each status, and returns the output of the matching renderer:
+
+```ts
+render() {
+  return html`
+    <button @click=${() => this.addTodo.mutate({ title: 'New todo' })}>
+      Add Todo
+    </button>
+    ${this.addTodo.render({
+      pending: ({ isIdle }) => isIdle ? nothing : html`<p>Adding...</p>`,
```
chughgaurav: `pending` is already narrowed to the pending variant, where `isIdle` is always false. The idle state is handled by omitting `idle`, so this should render “Adding...” directly (or add an explicit `idle: () => nothing` branch).

packages/lit-query/src/render.ts
```diff
@@ -0,0 +1,49 @@
...
+ *     const query = this.todos()
+ *     return renderResult(query, {
+ *       pending: () => html`Loading...`,
+ *       error: ({ error }) => html`Error: ${error.message}`,
+ *       success: ({ data }) => html`<ul>${data.map((todo) => html`<li>${todo.title}</li>`)}</ul>`,
+ *     })
+ *   }
+ * }
+ */
+export function renderResult<
```
chughgaurav, line 44: Could you regenerate and check in the Lit API reference for this public API? The PR currently has no `docs/framework/lit/reference/functions/renderResult.md` or reference-index entry, and the generated accessor type pages have not been updated for their new `render` members.

packages/lit-query/src/createQueryController.ts
EskiMojo14: updated - looks like that change happened after/around the same time I opened this originally.

chughgaurav (approved):

PR #10808, opened 2026-05-26
feat(query-core): add opt-in `ignoreUndefinedInKeys` filter option

webdevelopersrinu:
## Summary

Adds an opt-in `ignoreUndefinedInKeys` option to `QueryFilters` and `MutationFilters` to resolve the long-standing inconsistency between `hashKey` (strips `undefined` via `JSON.stringify`) and `partialMatchKey` (keeps `undefined`).

Refs #3741

## Background

When a query key contains an object property with value `undefined`, the cache stores it under a hash that excludes that property — but `invalidateQueries` cannot match it back when a sibling cache entry has a concrete value for the same property. See #3741 and the closed #4616 for full history.

@TkDodo previously raised the concern that there is no clean non-breaking fix. This PR addresses that by making the new behavior **fully opt-in** (default `false` → zero behavior change for existing users).

## API

\`\`\`ts
queryClient.invalidateQueries({
  queryKey: [{ entity: 'todos', filter: undefined }],
  ignoreUndefinedInKeys: true,
})
\`\`\`

Also available on `MutationFilters`.

## Design notes

- **Plain objects only.** `undefined` inside arrays is preserved.
- **Recursive.** Nested objects strip `undefined` at every level.
- Exported `PartialMatchKeyOptions` type for downstream adapter use.
- No new option on `QueryClient` defaults — kept the surface small. Happy to add a global default in a follow-up.

## Tests

- 4 unit tests in `utils.test.tsx` (broken-case, concrete-value-differ, array preservation, nested recursion).
- 1 end-to-end test in `queryClient.test.tsx` exercising `invalidateQueries`.

All 531 query-core tests pass locally.

## Summary by CodeRabbit

* **New Features**
  * Added ignoreUndefinedInKeys for query and mutation filters. When enabled, undefined properties in keys are treated as absent during partial matching (applies recursively to nested objects); default behavior remains unchanged.

* **Tests**
  * Added regression tests covering invalidate/match behavior with undefined values, nested objects, arrays, and the ignoreUndefinedInKeys toggle.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TanStack/query/pull/10808?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/query-core/src/__tests__/queryClient.test.tsx
+++ b/packages/query-core/src/__tests__/queryClient.test.tsx
@@ -1600,6 +1600,53 @@ describe('queryClient', () => {
       expect(queryFn).toHaveBeenCalledTimes(1)
       unsubscribe()
     })
+
+    // Regression test for https://github.com/TanStack/query/issues/3741
+    it('should match keys with `undefined` properties when ignoreUndefinedInKeys is set', async () => {
+      const queryFnAll = vi.fn().mockReturnValue('all')
+      const queryFnFiltered = vi.fn().mockReturnValue('filtered')
+
+      await queryClient.fetchQuery({
+        queryKey: [{ entity: 'todos', scope: 'list' }],
+        queryFn: queryFnAll,
+      })
+      await queryClient.fetchQuery({
+        queryKey: [{ entity: 'todos', scope: 'list', filter: { done: true } }],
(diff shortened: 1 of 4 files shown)


TkDodo:
we’ll address this differently

PR #10892, opened 2026-06-06
fix(svelte-query): handle shrinking createRawRef array by multiple entries (#10341)

### LINKED ISSUE #10341: [svelte-query] createQueries crashes when several items are removed from the array simultaneously
### Describe the bug

When using `createQueries`, changes to the array where two or more items are removed simultaneously causes a crash with the following error:
```
TypeError: can't delete property '3': proxy deleteProperty handler returned false

  in $effect.pre
  in FilterDrawer.svelte
  in ListPage.svelte
  in index.svelte
  in AnchorDecorator.svelte
  in ScrollDecorator.svelte
  in DecoratorWrapper.svelte
  in RenderFragment.svelte
  in ComposeFragments.svelte
  in Noop.svelte
  in AnchorDecorator.svelte
  in ScrollDecorator.svelte
  in DecoratorWrapper.svelte
  in RenderFragment.svelte
  in ComposeFragments.svelte
  in AnchorDecorator.svelte
  in Router.svelte
  in QueryClientProvider.svelte
  in App.svelte
```

https://svelte.dev/playground/2b60862d9c564cd1b211738f60e6b729?version=5.55.0

### Steps to reproduce

1. Go to the linked example
2. Try adding one or several elements to the array using the buttons, notice that it works
3. Try removing a single element using the button, this also works
4. Then try removing two items using the buttons, the page will now crash. Using `slice()` or `splice()` makes no difference. (the console in svelte playground isn't great, you'll have to open the console in the browser to see the error proper)

### Expected behavior

I expect useQueries to update its result with fewer queries made, but instead it crashes.

Every time

### Screenshots or Videos

https://github.com/user-attachments/assets/0f752c4f-e60c-4023-90b4-47b3f1589b23

### Platform

- OS: Windows
- Browser: Firefox 148.0.2

### Tanstack Query adapter

svelte-query

### TanStack Query version

v6.1.10

### TypeScript version

v5.9.3

### Additional context

_No response_

ousamabenyounes:
## Summary

Fix #10341

`createQueries` crashed with `TypeError: 'deleteProperty' on proxy: trap returned falsish for property 'N'` whenever two or more items were removed from its reactive array in a single update. The crash bubbled up from the `$effect.pre` in `createQueries.svelte.ts`, but the root cause sits in `createRawRef` (`packages/svelte-query/src/containers.svelte.ts`).

### Root cause

`createRawRef.update()` iterates `keysToRemove` in ascending index order. For arrays, the `deleteProperty` trap decrements `target.length` on every successful delete — so after the first deletion the next stale index is no longer `in target`, the trap returns `false`, and `delete` throws in strict mode (ES modules). The trap was also returning `false` when the property was already absent, which is not how `delete` is meant to behave on a missing prop.

### Fix

Two small, related changes in `packages/svelte-query/src/containers.svelte.ts`:

1. Sort `keysToRemove` numerically descending when `newValue` is an array, so every delete acts on the current tail and `length--` stays valid.
2. Return `true` from `deleteProperty` for missing props — the spec semantic for `delete` on an absent property is a no-op, not a strict-mode throw. This makes the trap robust against any other path that might also try to delete a stale index.

Both are needed: (1) is the correctness fix; (2) is defense in depth that makes the trap behave like a normal object trap.

### Verification

- Added regression test `should handle shrinking an array by more than one entry at once` in `packages/svelte-query/tests/containers.svelte.test.ts` — shrinks `[1, 2, 3, 4, 5]` → `[1, 2]` → `[1, 2, 3, 4]` → `[]`.
- Test fails on `main` with the exact symptom from the issue:
  `TypeError: 'deleteProperty' on proxy: trap returned falsish for property '4'` at `src/containers.svelte.ts:93`.
- Test passes with the patch applied.
- Full `@tanstack/svelte-query:test:lib` — **22 files, 164 tests, all pass.**
- `test:eslint` and `test:types` clean.

### Changeset

`@tanstack/svelte-query`: `patch` — runtime bug fix, no public API change.

Generated by Ora Studio
Vibe coded by ousamabenyounes

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed crashes that could occur when multiple items were removed from a reactive array in a single update.
  * Improved stability when dynamic query collections shrink, including when all items are removed.
  * Ensured array truncation and removals are handled reliably during rapid updates.
* **Tests**
  * Added regression coverage for shrinking reactive arrays by multiple elements, including reducing them to empty.


risalfajar:
Please merge this, this bug is making tanstack query svelte unusable

Michael-Obele (approved):
It's all good

PR #10895, opened 2026-06-06
fix(query-core): set dataUpdatedAt for streamed queries that resolve before hydration (#10603)

### LINKED ISSUE #10603: dataUpdatedAt incorrect for streamed queries that resolve before hydration

ousamabenyounes:
Fixes #10603

## Summary

When a dehydrated pending query's promise resolves synchronously during hydration (e.g., a React streamed query that completed before the client hydrated), the query correctly transitions to `success` but `dataUpdatedAt` remained `0`.

This happened because PR #10444 removed the unnecessary `query.fetch()` call for already-resolved promises — which was correct for avoiding unnecessary fetch setup, but it also removed the side effect that set `dataUpdatedAt`.

## Fix

The fix sets `dataUpdatedAt` to `Date.now()` in the hydration path when:
- The dehydrated query was pending (`status: 'pending'`)
- It resolved to have data (`data !== undefined`)
- `dataUpdatedAt` is `0` (its default for pending queries that never completed on the server)

This applies to both:
1. **New queries** — when a query is built from the dehydrated state (the `else` branch)
2. **Existing queries** — when updating a query already in the cache (the `if` branch)

## Testing

Added two test cases:
- New query: verifies `dataUpdatedAt` > 0 after hydrating a resolved pending query
- Existing query: verifies same for queries already in the cache

## Summary by CodeRabbit

* **Bug Fixes**
  * Query timestamp metadata is now correctly updated during streamed query rehydration with synchronously resolved data.

* **Tests**
  * Added test coverage for streamed query rehydration scenarios.

### DIFF
--- a/.changeset/streamed-query-dataupdatedat.md
+++ b/.changeset/streamed-query-dataupdatedat.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/query-core': patch
+---
+
+Fix `dataUpdatedAt` not being updated for streamed queries that resolve before hydration. When a dehydrated pending query's promise resolves synchronously during hydration, the query correctly transitions to `success` but `dataUpdatedAt` remained `0`. Now it is set to `Date.now()` in this case.
--- a/packages/query-core/src/__tests__/hydration.test.tsx
+++ b/packages/query-core/src/__tests__/hydration.test.tsx
@@ -1804,4 +1804,102 @@ describe('dehydration and rehydration', () => {
     clientQueryClient.clear()
     serverQueryClient.clear()
   })
+
+  // #10603 — dataUpdatedAt should not be 0 when a streamed query resolves before hydration
+  it('should set dataUpdatedAt when hydrating a pending query that resolved before hydration (new query)', async () => {
+    const key = queryKey()
+    // --- server ---
+    const serverQueryClient = new QueryClient({
+      defaultOptions: {
+        dehydrate: { shouldDehydrateQuery: () => true },
+      },
+    })
+
+    let resolvePrefetch: undefined | ((value?: unknown) => void)
+    const prefetchPromise = new Promise((res) => {
+      resolvePrefetch = res
+    })
+    void serverQueryClient.prefetchQuery({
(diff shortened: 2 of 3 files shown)


harublank0-0:
just for the info, there seems to be another PR for this issue @ https://github.com/TanStack/query/pull/10610/changes
In this PR, `state.dataUpdatedAt === 0` has not been checked while you've checked it.

ousamabenyounes:
Closing as a duplicate of #10610, which predates this by several weeks — apologies for the overlap, I didn't spot the existing PR when picking up the issue.

For what it's worth I also think #10610 is the better fix: it sets `dataUpdatedAt: dehydratedAt ?? Date.now()`, using the server's dehydration timestamp (when the data was actually fresh) rather than the client's hydration time as this PR does — that's more accurate when the promise resolved well before hydration. The extra `dataUpdatedAt === 0` guard here is also redundant, since the branch is already gated on `status === 'pending'`.

PR #10921, opened 2026-06-11
fix: retry remounted queries after error reset

### LINKED ISSUE #2712: Errored queries caught by ErrorBoundary are not retried on mount
**Describe the bug**
I'm using Suspense and ErrorBoundaries and I'm seeing unexpected behavior when queries are erroring. If a query errors I'm displaying an ErrorBoundary for that part of the UI. Now if I navigate to another page using react-router-dom and navigate back to the page, the ErrorBoundary is still there and the query is not retried. I know that normally you would have to reset the error boundary while you are on the same page to trigger a re-run of the failed queries, but in this case I would've assumed that the queries are retried when I visit the page again, i.e. the components are mounted again. 

**To Reproduce**
1. Visit the [Code Sandbox](https://codesandbox.io/s/compassionate-beaver-fs9t4?file=/src/App.tsx)
2. Open the console
3. Wait until the error occurs and the fallback of the ErrorBoundary is shown: "An error occured!" (You need to click off of react-error-overlay to see it)
4. Press the "Hide" button to unmount the ErrorBoundary
5. Press the "Show" button to render it again
6. Query is not re-executed and the error is immediately shown again

**Expected behavior**
The query is re-executed after the component mounts, i.e. honoring the `retryOnMount` setting.

 - Version: 3.24.4

pupuking723:
## 🎯 Changes

Fixes #2712.

React Query currently stores error boundary reset state as a single boolean. When an errored query is unmounted, the fallback can call `reset()`, but another query mounted before the original query remounts can clear that reset state. The original query then remounts with `retryOnMount: false` and stays in the error boundary instead of retrying.

This PR tracks reset generations per query instance so a remounted errored query can still observe that a reset happened after it last mounted, even if another query has already cleared the global reset flag.

It also adds a regression test for the route/page switch case where:

- the first query errors into an error boundary
- the fallback calls `reset()` on unmount
- another query mounts and clears the reset state
- the original query remounts and retries successfully

## ✅ Checklist

`pnpm run test:pr` was attempted with `NX_DAEMON=false`, but Nx failed locally before running project targets with `Failed to start plugin worker`.

Targeted verification completed locally:

- `npx pnpm@11.1.0 --filter @tanstack/react-query exec vitest run src/__tests__/QueryResetErrorBoundary.test.tsx`
- `npx pnpm@11.1.0 --filter @tanstack/react-query run test:types:tscurrent`
- `npx pnpm@11.1.0 --filter @tanstack/react-query run test:eslint`
- `npx pnpm@11.1.0 exec prettier --experimental-cli --check .changeset/error-boundary-remount-retry.md packages/react-query/src/QueryErrorResetBoundary.tsx packages/react-query/src/errorBoundaryUtils.ts packages/react-query/src/__tests__/QueryResetErrorBoundary.test.tsx`
- `git diff --check`

`test:eslint` completed with existing warnings in unrelated test files.

## 🚀 Release Impact

## Release Notes

* **New Features**
  * Added an optional `getResetCount` function to the Query Error Reset Boundary for tracking how many times the boundary has been reset.

* **Bug Fixes**
  * Improved error boundary reset behavior for remounted queries: retry/reset handling is now preserved correctly even if another query clears the reset state first.

* **Tests**
  * Added coverage for reset behavior across unmount/remount and query switching scenarios.

### DIFF
--- a/.changeset/error-boundary-remount-retry.md
+++ b/.changeset/error-boundary-remount-retry.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/react-query': patch
+---
+
+Preserve error boundary reset retries for remounted queries when another query clears the reset state first.
--- a/packages/react-query/src/QueryErrorResetBoundary.tsx
+++ b/packages/react-query/src/QueryErrorResetBoundary.tsx
@@ -5,21 +5,29 @@ import * as React from 'react'
 export type QueryErrorResetFunction = () => void
 export type QueryErrorIsResetFunction = () => boolean
 export type QueryErrorClearResetFunction = () => void
+export type QueryErrorResetCountFunction = () => number
 
 export interface QueryErrorResetBoundaryValue {
   clearReset: QueryErrorClearResetFunction
+  getResetCount?: QueryErrorResetCountFunction
   isReset: QueryErrorIsResetFunction
   reset: QueryErrorResetFunction
 }
 
 function createValue(): QueryErrorResetBoundaryValue {
   let isReset = false
+  let resetCount = 0
+
   return {
     clearReset: () => {
       isReset = false
     },
+    getResetCount: () => {
+      return resetCount
+    },
     reset: () => {
       isReset = true
+      resetCount++
     },
     isReset: () => {
       return isReset
--- a/packages/react-query/src/__tests__/QueryResetErrorBoundary.test.tsx
+++ b/packages/react-query/src/__tests__/QueryResetErrorBoundary.test.tsx
@@ -9,6 +9,7 @@ import {
   QueryErrorResetBoundary,
   useQueries,
   useQuery,
+  useQueryErrorResetBoundary,
   useSuspenseQueries,
   useSuspenseQuery,
 } from '..'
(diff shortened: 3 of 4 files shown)


ousamabenyounes:
Approach looks right — monotonic `resetCount` + per-`Query` last-observed count, gating `retryOnMount` only on a newer generation. Handles the sibling-consume case from #2712 cleanly, and ships with a changeset + regression test.

Two things to confirm before merge:

1. **Suspense loop guard**: looks safe since `isResetForQuery` bumps the stored count as soon as it returns `true` (one-shot per generation), but worth confirming the existing suspense-loop tests still pass — that's the regression risk.
2. **`WeakMap` keyed by `Query`**: if the query is gc'd (`gcTime`) between unmount and remount, the lookup misses and the retry won't fire. Probably fine, but worth a code comment so it's intentional.

pupuking723:
Thanks, I checked both points.

The WeakMap behavior is intentional here. The helper comment now says reset generations are tracked for the live Query instance, so if that query is collected, a later Query starts fresh instead of inheriting retry state from the old lifecycle.

I also reran the related error-boundary/suspense coverage locally:

```sh
npx pnpm@11.1.0 --filter @tanstack/react-query test:lib --run src/__tests__/QueryResetErrorBoundary.test.tsx src/__tests__/useSuspenseQuery.test.tsx src/__tests__/usePrefetchQuery.test.tsx src/__tests__/usePrefetchInfiniteQuery.test.tsx src/__tests__/useSuspenseQueries.test.tsx
```

Result: 5 test files passed, 70 tests passed, no type errors.

PR #10923, opened 2026-06-12
fix(solid-query): don't hang async SSR when a disabled query is rendered

### LINKED ISSUE #10907: Solid Query: useQuery({ enabled: false }) hangs renderToStringAsync (SSR) on 5.101.0
### Describe the bug

A `useQuery` with `enabled: false` never settles its underlying `createResource` during SSR, so `renderToStringAsync` (and SolidStart's async/stream SSR) **hangs** as soon as the disabled query's `.data` is read during render.

On the server, solid-query forces `experimental_prefetchInRender = true` ([`useBaseQuery`](https://github.com/TanStack/query/blob/main/packages/solid-query/src/useBaseQuery.ts)), so the observer result carries a `.promise`. For a **disabled** query that promise never resolves, and the resource fetcher returns it (via `hydratableObserverResult`) instead of short-circuiting — so the SSR render awaits a promise that can never settle.

Reproduces on `@tanstack/solid-query@5.101.0` (latest) with `solid-js@1.9.13`.

Related, both closed and neither matching this exact symptom (single `useQuery`, SSR render hang): #6256, #7167.

Standalone Node ESM — no framework needed. `solid-js/web` resolves to its server build under Node's default `node` export condition, so `isServer === true`.

```js
// repro.mjs  ->  node repro.mjs
import { renderToStringAsync, createComponent, Suspense } from 'solid-js/web'
import { QueryClient, QueryClientProvider, useQuery } from '@tanstack/solid-query'

const client = new QueryClient()

function Inner() {
  const q = useQuery(() => ({
    queryKey: ['disabled'],
    queryFn: () => Promise.resolve('data'), // never runs (disabled)
    enabled: false,
  }))
  return () => `data=${JSON.stringify(q.data)}` // reading data drives the resource
}

const App = () =>
  createComponent(QueryClientProvider, {
    client,
    get children() {
      return createComponent(Suspense, {
        get children() {
          return createComponent(Inner, {})
        },
      })
    },
  })

// The race timeout is ONLY so the script exits; renderToStringAsync never settles.
const html = await Promise.race([
  renderToStringAsync(() => createComponent(App, {})),
  new Promise((_, r) =>
    setTimeout(() => r(new Error('renderToStringAsync did not settle (5s)')), 5000),
  ),
])
console.log(html)
```

### Steps to reproduce

1. `npm i @tanstack/solid-query@5.101.0 solid-js@1.9.13`
2. `node repro.mjs`
3. Observe it never resolves — the 5s race rejects with `renderToStringAsync did not settle`.

### Expected behavior

A disabled query must not suspend SSR; the resource should resolve immediately with the idle state.

### Possible fix

Short-circuiting `enabled === false` in the `useBaseQuery` resource fetcher — returning the synchronous result with the dangling `.promise` stripped — fixes it. Verified locally against 5.101.0:

```js
// inside the createResource fetcher, before `return new Promise(...)`
const obs = observer()
if (obs.options.enabled === false) {
  const { promise, ...rest } = hydratableObserverResult(obs.getCurrentQuery(), observerResult)
  return rest
}
```

- unpatched: hangs (5s timeout above)
- with the short-circuit: resolves in ~25ms with the correct idle state

Happy to open a PR if this approach looks right — wanted to confirm the direction first, since it touches the core query path and `enabled` can also be reactive (this handles the literal `false` case).

### Environment

`@tanstack/solid-query@5.101.0`, `solid-js@1.9.13`, Node 24.

naveentehrpariya:
## 🎯 Changes

Resolves #10907

A `useQuery` with `enabled: false` hangs `renderToStringAsync` (and SolidStart's async/stream SSR) as soon as the query is rendered.

**Root cause:** on the server, `useBaseQuery` forces `experimental_prefetchInRender = true`, so every observer result carries a `.promise`. `hydratableObserverResult()` strips `refetch` because functions can't be serialized, but it predates `experimental_prefetchInRender` and doesn't strip `promise`. The promise therefore ends up in the resolved resource value, and Solid's SSR serializer (seroval) awaits embedded promises — for a disabled query that promise never settles, so serialization (and the render) hangs forever.

I verified this mechanism empirically against the published `@tanstack/solid-query@5.101.0` build using the standalone repro from #10907: the resource fetcher actually resolves immediately (`isLoading` is `false` for a disabled query) — the hang happens afterwards, during serialization of the resolved value. Stripping `promise` from `hydratableObserverResult()` makes `renderToStringAsync` settle with the expected markup (`data=undefined`) and a clean hydration script.

**Fix:** strip `promise` from the hydratable result alongside `refetch`. The client rebuilds the result (including its own `promise`) from the hydrated query state, so nothing is lost.

Also adds the first SSR-path test for solid-query (`isServer` mocked to `true`), asserting the resolved result of a disabled query carries neither `refetch` nor `promise` — it fails without the fix.

## ✅ Checklist

- [x] I have tested this code locally with `pnpm run test:pr`. (solid-query package: 320 tests + typecheck pass, eslint clean)

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented server-side rendering from hanging by ensuring non-serializable promise-like values aren’t included in SSR hydration data for disabled queries.

* **Tests**
  * Added server-side tests verifying disabled and enabled queries render expected data and that non-serializable fields are excluded from serialized query state.

### DIFF
--- a/.changeset/solid-query-ssr-disabled-hang.md
+++ b/.changeset/solid-query-ssr-disabled-hang.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/solid-query': patch
+---
+
+Fix `renderToStringAsync` hanging during SSR when a query is disabled: strip the unserializable `experimental_prefetchInRender` promise from the hydratable observer result, since a disabled query's promise never settles and the SSR serializer would await it forever.
--- a/packages/solid-query/src/__tests__/ssr.test.tsx
+++ b/packages/solid-query/src/__tests__/ssr.test.tsx
@@ -42,4 +42,31 @@ describe('useQuery on the server', () => {
     expect(state!.refetch).toBeUndefined()
     expect(state!.promise).toBeUndefined()
   })
+
+  it('resolves an enabled query with data and without unserializable fields', async () => {
+    const client = new QueryClient()
+    let state: UseQueryResult<string> | undefined
+
+    function Page() {
+      const query = useQuery(() => ({
+        queryKey: ['enabled-ssr'],
+        queryFn: () => Promise.resolve('server data'),
+      }))
+      state = query
+      return <div>data: {String(query.data)}</div>
+    }
+
+    const rendered = render(() => (
+      <QueryClientProvider client={client}>
+        <Page />
+      </QueryClientProvider>
+    ))
+
+    await waitFor(() => rendered.getByText('data: server data'))
+
+    expect(state!.data).toBe('server data')
+    expect(state!.isSuccess).toBe(true)
+    expect(state!.refetch).toBeUndefined()
+    expect(state!.promise).toBeUndefined()
+  })
 })
--- a/packages/solid-query/src/__tests__/ssr.test.tsx
+++ b/packages/solid-query/src/__tests__/ssr.test.tsx
@@ -1,0 +1,45 @@
+import { describe, expect, it, vi } from 'vitest'
+import { render, waitFor } from '@solidjs/testing-library'
+import { QueryClient, QueryClientProvider, useQuery } from '..'
+import type * as SolidWeb from 'solid-js/web'
+import type { UseQueryResult } from '..'
+
+// Force the server code path: on the server `useBaseQuery` enables
+// `experimental_prefetchInRender` and resolves the resource with a
+// serializable `hydratableObserverResult()`.
+vi.mock('solid-js/web', async (importOriginal) => {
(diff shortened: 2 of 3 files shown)


naveentehrpariya:
Added a complementary test for enabled queries on the server — it asserts the query resolves with data and that the serialized result still carries neither `refetch` nor `promise`.

TkDodo:
experimental_prefetchInRender has been removed

PR #10962, opened 2026-06-16
fix(react-query): retry fetch on remount when error boundary has not been reset

EduardF1:
## Summary

- Fixes a bug where `useQuery` with `throwOnError` would refuse to retry on remount even after the error boundary had unmounted and remounted the failed observer — effectively leaving the component stuck in an error state with no recovery path unless the user explicitly reset the boundary.
- The fix updates `ensurePreventErrorBoundaryRetry` in `errorBoundaryUtils.ts`: retry is only suppressed when the boundary has not been reset **and** the query is either using `suspense` or still has live observers. On a fresh mount (zero observers), the retry suppression is lifted so the remounted component can attempt a new fetch.
- Adds two test cases covering the remount retry scenario, including an edge case with `staleTime: Infinity`.

## Files changed

- `packages/react-query/src/errorBoundaryUtils.ts` — guard loosened to allow remount retry
- `packages/react-query/src/__tests__/QueryResetErrorBoundary.test.tsx` — two new/updated test cases

## Test plan

- [x] `pnpm run test:lib -- --run QueryResetErrorBoundary` in `packages/react-query` — all 14 tests pass
- [ ] Full CI (`pnpm test`) — to be run by CI

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error boundary query recovery to intelligently retry queries on remount based on Suspense configuration and active query observers, providing more predictable error recovery behavior.

### DIFF
--- a/packages/react-query/src/__tests__/QueryResetErrorBoundary.test.tsx
+++ b/packages/react-query/src/__tests__/QueryResetErrorBoundary.test.tsx
@@ -294,7 +294,7 @@ describe('QueryErrorResetBoundary', () => {
       consoleMock.mockRestore()
     })
 
-    it('should not retry fetch if the reset error boundary has not been reset', async () => {
+    it('should retry fetch on remount even if the reset error boundary has not been reset', async () => {
       const consoleMock = vi
         .spyOn(console, 'error')
         .mockImplementation(() => undefined)
@@ -302,12 +302,14 @@ describe('QueryErrorResetBoundary', () => {
       const key = queryKey()
 
       let succeed = false
+      let fetchCount = 0
 
       function Page() {
         const { data } = useQuery({
           queryKey: key,
           queryFn: () =>
             sleep(10).then(() => {
+              fetchCount++
               if (!succeed) throw new Error('Error')
               return 'data'
             }),
@@ -350,7 +352,8 @@ describe('QueryErrorResetBoundary', () => {
 
       fireEvent.click(rendered.getByText('retry'))
       await vi.advanceTimersByTimeAsync(11)
-      expect(rendered.getByText('error boundary')).toBeInTheDocument()
+      expect(rendered.getByText('data')).toBeInTheDocument()
+      expect(fetchCount).toBe(2)
 
       consoleMock.mockRestore()
     })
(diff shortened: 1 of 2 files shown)


nahrinoda (commented):
The fix direction is correct. `ensurePreventErrorBoundaryRetry` is called in `useBaseQuery` (line 87) BEFORE the observer is created (line 95-101), so `query?.getObserversCount()` reflects only *other* active observers — on a fresh remount with no siblings, it's 0, allowing retry. The suspense carve-out makes sense: suspense queries must go through the reset flow because React's Suspense/ErrorBoundary lifecycle manages retries differently.

One question on the second test change (line 483): the `fireEvent.click` was moved UP before `advanceTimersByTimeAsync`, and the intermediate retry step (old lines 494-499) was removed entirely. This collapses two user interactions into one. Can you confirm this test still validates the "after a previous reset" scenario — i.e., that a *second* error→retry cycle (where the boundary is NOT reset) still recovers? As written, it looks like the test now only covers one error→reset→retry cycle followed by a single non-reset retry, which is less coverage than the original three-step flow.

EduardF1:
@nahrinoda yes — the reworked test still covers the second error → retry cycle without the error boundary being reset. The sequence remains: initial error, boundary reset enables the first remount retry, that retry still returns an error, then the next remount happens with no additional boundary reset and verifies we still retry again and can reach data. I left the code unchanged because the coverage is preserved.

EduardF1:
I re-checked the questioned test flow and kept it unchanged. The click before \\dvanceTimersByTimeAsync(11)\\ is what triggers the remount/retry; the following assertion still observes the intermediate second error boundary state after a real retry with \\shouldReset = true\\. The final click then uses \\shouldReset = false\\ and verifies the new behavior: a fresh mount retries successfully without another QueryErrorResetBoundary reset. Targeted validation: \\pnpm --filter @tanstack/react-query test:lib -- src/__tests__/QueryResetErrorBoundary.test.tsx\\ passes (14 tests).

nahrinoda:
Thanks for the explanation. I see — the old intermediate step was testing the previous behavior where retry was blocked without reset. Since the fix changes that behavior, the test correctly reflects the new flow. LGTM on the test changes.

EduardF1:
Thanks @nahrinoda — appreciate you looking it over. The updated test reflects the new retry-on-remount behavior, so I believe this is ready. Happy to adjust anything if a maintainer has further feedback.

nahrinoda (approved):

nahrinoda (approved):

nahrinoda (approved):

TkDodo:
> would refuse to retry on remount even after the error boundary had unmounted and remounted the failed observer

the recommendation is to reset the error boundary when it unmounts in user land

PR #10969, opened 2026-06-17
fix(query-core): addToEnd caps result to max when items already exceeds max

JSap0914:
## Bug

`addToEnd(items, item, max)` uses `slice(1)` which only ever removes one element from the front. When `items.length` is already greater than `max`, the returned array still exceeds `max`.

This matters in practice for infinite queries with a dynamic `maxPages` option: if the user reduces `maxPages` between fetches (e.g. via React state), `addToEnd` receives a `pages` array that is longer than the new `max`. The result silently accumulates more pages than the configured limit.

### Before (broken)

```ts
addToEnd([1, 2, 3, 4, 5], 6, 3)
// → [2, 3, 4, 5, 6]  (length 5 — exceeds max of 3)
```

### After (fix)

```ts
addToEnd([1, 2, 3, 4, 5], 6, 3)
// → [4, 5, 6]  (length 3 — correctly capped)
```

## Fix

Replace `newItems.slice(1)` with `newItems.slice(-max)` in `addToEnd`.

All four existing `addToEnd` tests are unaffected: they all use `items.length === max`, where both expressions produce the same result.

`addToStart` has the symmetric issue (`slice(0, -1)`), but an existing test already asserts the current behavior, so that change warrants a separate discussion.

## Verification

```
npx nx run @tanstack/query-core:test:lib -- --reporter=verbose --testNamePattern="addToEnd"
```

All 5 tests pass (4 pre-existing + 1 new regression test).

> AI-assisted contribution.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the `addToEnd` function to correctly limit array length when items already exceed the maximum capacity, ensuring proper array truncation and maintaining size constraints.

* **Tests**
  * Added test coverage for the `addToEnd` function to verify correct behavior when handling oversized arrays.

### DIFF
--- a/packages/query-core/src/utils.ts
+++ b/packages/query-core/src/utils.ts
@@ -417,7 +417,7 @@ export function keepPreviousData<T>(
 
 export function addToEnd<T>(items: Array<T>, item: T, max = 0): Array<T> {
   const newItems = [...items, item]
-  return max && newItems.length > max ? newItems.slice(1) : newItems
+  return max && newItems.length > max ? newItems.slice(-max) : newItems
 }
 
 export function addToStart<T>(items: Array<T>, item: T, max = 0): Array<T> {
--- a/.changeset/fix-add-to-end-max-cap.md
+++ b/.changeset/fix-add-to-end-max-cap.md
@@ -1,0 +1,12 @@
+---
+"@tanstack/query-core": patch
+---
+
+fix(query-core): addToEnd now correctly caps result length when items already exceeds max
+
+`addToEnd(items, item, max)` previously used `slice(1)` which only removed one element from
+the front of the array. When `items.length` was already greater than `max` (e.g. when
+`maxPages` is reduced at runtime), the returned array could still exceed `max`.
+
+Replaced `slice(1)` with `slice(-max)` so the result is always capped to the last `max`
+elements regardless of the input length.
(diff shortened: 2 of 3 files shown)


nahrinoda (commented):
The `addToEnd` fix is correct — `slice(-max)` properly caps the result regardless of input length.

One observation on the deferred `addToStart` issue: the PR notes it's left out because "an existing test already asserts the current behavior." But that test is actually pinning the bug:

```ts
it('should respect the max argument', () => {
    const items = [1, 2, 3]
    const item = 4
    const max = 2
    const newItems = addToStart(items, item, max)
    expect(newItems).toEqual([4, 1, 2])  // length 3 — exceeds max of 2
})
```

With `max=2`, the result should be `[4, 1]` (2 elements), not `[4, 1, 2]` (3 elements). The test is asserting broken behavior, so it shouldn't be a reason to defer the fix. The symmetric fix would be `slice(0, max)` instead of `slice(0, -1)`.

Since both functions are used together in `infiniteQueryBehavior.ts` (`addTo = previous ? addToStart : addToEnd`), fixing them together avoids a state where forward pagination correctly caps pages but backward pagination doesn't.

packages/query-core/src/utils.ts
```diff
@@ -417,7 +417,7 @@ export function keepPreviousData<T>(
 
 export function addToEnd<T>(items: Array<T>, item: T, max = 0): Array<T> {
   const newItems = [...items, item]
-  return max && newItems.length > max ? newItems.slice(1) : newItems
+  return max && newItems.length > max ? newItems.slice(-max) : newItems
 }
 
 export function addToStart<T>(items: Array<T>, item: T, max = 0): Array<T> {
```
TkDodo, line 423: `addToStart` needs this fix too pls

PR #10972, opened 2026-06-19
fix(preact-query): allow retryOnMount when throwOnError is function

dfedoryshchev:
## Summary

The preact-query adapter never evaluates a function-form `throwOnError`. `ensurePreventErrorBoundaryRetry` treats any `throwOnError` as truthy and disables `retryOnMount`, so a query whose `throwOnError` function returns `false` is wrongly prevented from retrying on mount.

This brings preact-query in line with react-query, where the same issue was fixed in #9338: when `throwOnError` is a function, it is called against the actual query error and only suppresses retry-on-mount when it returns `true`.

- Pass the resolved `query` into `ensurePreventErrorBoundaryRetry` from `useBaseQuery` and `useQueries`.
- Evaluate the function-form `throwOnError` via `shouldThrowError` before deciding to disable `retryOnMount`.
- Added tests covering `throwOnError: () => false` (retries on mount) and `() => true` (does not).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue in Preact Query where queries would not retry on component mount when error-throwing behavior was defined as a dynamic function.

### DIFF
--- a/.changeset/preact-query-throwonerror-function.md
+++ b/.changeset/preact-query-throwonerror-function.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/preact-query': patch
+---
+
+fix(preact-query): allow retryOnMount when throwOnError is function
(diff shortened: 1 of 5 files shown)


dfedoryshchev:
Gentle bump on this one. The checks are green and the branch still merges cleanly. Happy to rebase onto the current `main` if that makes it easier to take.

PR #10973, opened 2026-06-21
Query mutation options in core

benjavicente:
## 🎯 Changes

Adds `queryOptions` and `mutationOptions` to the core library.

This was requested at https://github.com/TanStack/query/discussions/10735 and https://github.com/TanStack/query/discussions/9258 discussions. Having those function in core is useful for defining shared query options for a framework agnostic layer, or for server side usage. Specially for using query keys for type-safe interactions with the core imperative API.

Most of the frameworks accept the shared interface from core without changes. The exception is Vue, that require a small change in the interface since `MaybeRefOrGetter`wasn't compatible. I added runtime test for Vue to demonstrate that it works with the core query options, with the same shape of `() => coreOptions` as Solid/Angular/Lit/Svelte.

About docs: There is a lot of outdated docs, so I excluded the doc generation in the PR. I didn't add docs mentioning that core has this new functions.

Pending questions:

- Is `CoreQueryOptions` and `CoreMutationOptions` ok names for the interface the new `queryOptions` and `mutationOptions` requires? The core library already has `QueryOptions` and `MutationOptions`. The new `CoreQueryOptions` and `CoreMutationOptions` are the equivalent of `QueryOptions` and `MutationOptions` exported by each individual adapter. The new types can't replace the base options alredy defined in core.
- Should `CoreQueryOptions` and `CoreMutationOptions`  be excluded from the re-export from core? How? Each adapter exports `queryOptions` and `mutationOptions`, so a consumer can't import the implementation defined in core if the consumer hasn't installed `@tanstack/query-core` directly, but they can consume `CoreQueryOptions` and `CoreMutationOptions`.
- I'm not happy on how the `useQuery` interface of Vue had to be extended, there is probably something that can be done there to simplify the types.

## ✅ Checklist

## 🚀 Release Impact

## Release Notes

* **New Features**
  * Added framework-agnostic `queryOptions` and `mutationOptions` helpers for reusable query and mutation configurations.
  * Improved type inference when sharing these options across supported frameworks, including query data, mutation data, variables, and mutation inputs.

### DIFF
--- a/packages/query-core/src/index.ts
+++ b/packages/query-core/src/index.ts
@@ -41,6 +41,10 @@ export {
 export type { MutationFilters, QueryFilters, SkipToken, Updater } from './utils'
 
 export { streamedQuery as experimental_streamedQuery } from './streamedQuery'
+export { queryOptions } from './queryOptions'
+export type { CoreQueryOptions } from './queryOptions'
+export { mutationOptions } from './mutationOptions'
+export type { CoreMutationOptions } from './mutationOptions'
 
 // Types
 export type {
--- a/packages/query-core/src/index.ts
+++ b/packages/query-core/src/index.ts
@@ -42,6 +42,10 @@ export {
 export type { MutationFilters, QueryFilters, SkipToken, Updater } from './utils'
 
 export { streamedQuery as experimental_streamedQuery } from './streamedQuery'
+export { queryOptions } from './queryOptions'
+export type { CoreQueryOptions } from './queryOptions'
+export { mutationOptions } from './mutationOptions'
+export type { CoreMutationOptions } from './mutationOptions'
 
 // Types
 export type {
--- a/packages/react-query/src/__tests__/queryOptions.test-d.tsx
+++ b/packages/react-query/src/__tests__/queryOptions.test-d.tsx
@@ -2,6 +2,7 @@ import { assertType, describe, expectTypeOf, it } from 'vitest'
 import {
   QueriesObserver,
   QueryClient,
+  queryOptions as coreQueryOptions,
   dataTagSymbol,
   skipToken,
 } from '@tanstack/query-core'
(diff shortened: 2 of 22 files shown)


packages/query-core/src/index.ts
```diff
@@ -41,6 +41,10 @@ export {
 export type { MutationFilters, QueryFilters, SkipToken, Updater } from './utils'
 
 export { streamedQuery as experimental_streamedQuery } from './streamedQuery'
+export { queryOptions } from './queryOptions'
+export type { CoreQueryOptions } from './queryOptions'
+export { mutationOptions } from './mutationOptions'
+export type { CoreMutationOptions } from './mutationOptions'
```
TkDodo, line 48: since adapters re-export the core and then export their own things, they won’t have access to the `queryOptions` from the core, right? is that on purpose?

packages/react-query/src/__tests__/queryOptions.test-d.tsx
```diff
@@ -48,6 +49,17 @@ describe('queryOptions', () => {
...
+  it('should work when core queryOptions are passed to useQuery', () => {
+    const options = coreQueryOptions({
+      queryKey: queryKey(),
+      queryFn: () => Promise.resolve({ id: '1', title: 'Do Laundry' }),
+    })
+
+    const { data } = useQuery(options)
+    expectTypeOf(data).toEqualTypeOf<
+      { id: string; title: string } | undefined
+    >()
```
TkDodo, line 68: with this, what’s the difference between `useQuery(queryOptions)` and `useQuery(coreQueryOption)` ?

in other words, if they are compatible, why would e.g. the react adapter need a separate `queryOptions` and not just re-export and let users use the one from the core?

packages/query-core/src/index.ts
benjavicente: Yes. I don't think someone wanting to define shared query options for multiple frameworks would import from a framework specific package.

packages/react-query/src/__tests__/queryOptions.test-d.tsx
benjavicente: `queryOptions` from each library could extend the `coreQueryOption`, like with `deferStream` in Solid or `RefOrGetter` pattern in Vue.

benjavicente:
I avoided changing the types of the adapters to reuse the new types in core to keep the changes of this PR small (only test changes for most adapters). The code query options could be something that adapters completely reuse.

TkDodo:
> The code query options could be something that adapters completely reuse.

I guess you mean re-use on type-level, because at runtime, they all don’t do anything?

benjavicente:
> > The code query options could be something that adapters completely reuse.
> 
> I guess you mean re-use on type-level, because at runtime, they all don’t do anything?

Yes

TkDodo:
okay, let’s do it 🔥 . can you fix the conflicts ?

PR #10980, opened 2026-06-24
fix(query-broadcast-client-experimental): catch postMessage rejections to prevent unhandled DataCloneErrors

### LINKED ISSUE #10542: [query-broadcast-client-experimental] postMessage failures surface as unhandled DataCloneError rejections
### Describe the bug

`broadcastQueryClient` calls `channel.postMessage(...)` three times (on `updated`, `removed`, `added`) without `.catch()`-ing the returned promise — see [`src/index.ts#L39-L62`](https://github.com/TanStack/query/blob/main/packages/query-broadcast-client-experimental/src/index.ts#L39-L62).

When a query's `state.data` (or `state.error`, or `queryKey`) contains a value the structured-clone algorithm can't serialize — `ReadableStream`, `Response`, `Request`, `File`, functions, framework proxies (Vue reactive, MobX, etc.) — `broadcast-channel` rejects internally and the rejection becomes an `unhandledrejection`:

```
DataCloneError: Failed to execute 'postMessage' on 'BroadcastChannel':
  A ReadableStream could not be cloned because it was not transferred.
    at postMessage (broadcast-channel/.../methods/native.js:24)
```

The stack points into `node_modules` with no indication of which query is at fault, so it's effectively unactionable when it shows up in Sentry / Datadog. Same root cause as #8356 (Vue reactive proxy in Nuxt) which was closed via a user-side `toRaw()` workaround — but that fix doesn't generalize: any query whose data happens to contain a non-cloneable value silently kills cross-tab sync for that update *and* spams error trackers.

Bug location: https://github.com/TanStack/query/blob/main/packages/query-broadcast-client-experimental/src/index.ts#L39-L62

Inline repro (5 lines of vanilla TS, no framework — happy to wire up a Stackblitz if needed but I think it'd just add ceremony around this snippet):

```ts
import { QueryClient } from '@tanstack/query-core'
import { broadcastQueryClient } from '@tanstack/query-broadcast-client-experimental'

broadcastQueryClient({ queryClient, broadcastChannel: 'demo' })

window.addEventListener('unhandledrejection', (e) =>
  console.error('caught unhandled:', e.reason),
)

// any non-cloneable value triggers it. ReadableStream is the most common in
// the wild (Response.body, fetch streaming, AI SDK, etc.):
queryClient.setQueryData(['stream'], new Response('hi').body)
```

### Steps to reproduce

1. Set up `broadcastQueryClient` against any `QueryClient`.
2. Cache any value containing a `ReadableStream` / `Response` / `File` / function / proxy.
3. Observe an unhandled `DataCloneError` rejection in the console.

### Expected behavior

`broadcastQueryClient` should not surface clone failures as unhandled rejections. Cross-tab sync of one query should not crash silently, nor should it pollute consumers' error trackers with stacks that point into `node_modules` and don't identify the offending `queryHash`.

Every time

### Platform

- OS: any (Chrome on macOS in our case)
- Browser: any with `BroadcastChannel` support
- Version: n/a

### Tanstack Query adapter

vanilla (the broadcast client itself)

### TanStack Query version

`@tanstack/query-broadcast-client-experimental@5.99.0` (current latest), `@tanstack/query-core@5.x`

### TypeScript version

5.x

### Additional context

Two possible fixes — happy to PR either flavor:

**1. Minimal — just \`.catch()\` and warn in dev:**

\`\`\`ts
const safePost = (message: BroadcastMessage) => {
  channel.postMessage(message).catch((error: unknown) => {
      console.warn(
        \`[broadcastQueryClient] failed to broadcast \"\${message.type}\" for queryHash \"\${message.queryHash}\":\`,
        error,
      )
    }
  })
}
\`\`\`

…then replace the three \`channel.postMessage({...})\` call sites with \`safePost({...})\`.

**2. Hookable — same as (1) plus an opt-in \`onBroadcastError\` so consumers can pipe to Sentry/Datadog/etc. without forking:**

\`\`\`ts
interface BroadcastQueryClientOptions {
  queryClient: QueryClient
  broadcastChannel?: string
  options?: BroadcastChannelOptions
  onBroadcastError?: (error: unknown, message: BroadcastMessage) => void
}
\`\`\`

Tests: extend \`src/__tests__/index.test.ts\` with a case that calls \`setQueryData\` with a \`ReadableStream\` and asserts no \`unhandledrejection\` fires (and, for flavor 2, that \`onBroadcastError\` is invoked with the offending \`queryHash\`).

tsushanth:
Fixes #10542

## Problem

`broadcastQueryClient` subscribes to the query cache and calls `channel.postMessage()` on three paths (`updated`, `removed`, `added`). The returned promise was never caught, so any value the structured-clone algorithm cannot serialize — `ReadableStream`, `Response`, `File`, functions, Vue reactive proxies, MobX observables, etc. — causes a `DataCloneError` that surfaces as an unhandled rejection:

```
DataCloneError: Failed to execute 'postMessage' on 'BroadcastChannel':
  A ReadableStream could not be cloned because it was not transferred.
    at postMessage (broadcast-channel/.../methods/native.js:24)
```

The stack points into `node_modules` with no indication of which query triggered it, making the error effectively unactionable in Sentry / Datadog. Cross-tab sync for that update is silently dropped while the error tracker accumulates noise.

## Solution

Introduce a `safePost` helper that wraps all three `channel.postMessage()` call sites in `.catch()`. The default behavior is a `console.warn` in non-production builds that includes the event `type` and offending `queryHash`, making the failure locally actionable without polluting production error trackers.

Callers who want to route errors to their own monitoring can supply an `onBroadcastError` callback:

```ts
broadcastQueryClient({
  queryClient,
  onBroadcastError: (error, message) => {
    Sentry.captureException(error, {
      extra: { queryHash: message.queryHash, type: message.type },
    })
  },
})
```

`onBroadcastError` receives the raw error and the attempted `BroadcastMessage` (including `type` and `queryHash`), giving enough context to filter noise or tie the failure back to a specific query.

## Changes

- `src/index.ts` — adds `BroadcastMessage` type, `onBroadcastError` option, and `safePost` helper; replaces the three bare `channel.postMessage()` calls
- `src/__tests__/index.test.ts` — adds two new tests: one asserting no `unhandledrejection` fires when postMessage rejects, and one asserting `onBroadcastError` is called with the correct error and `queryHash`

All four tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when broadcast updates fail, reducing the chance of unhandled promise rejections.
  * Added a callback option to report broadcast errors and included clearer warning behavior in non-production environments.
  * Strengthened test coverage for broadcast failure scenarios to ensure errors are handled safely.

### DIFF
--- a/packages/query-broadcast-client-experimental/src/__tests__/index.test.ts
+++ b/packages/query-broadcast-client-experimental/src/__tests__/index.test.ts
@@ -1,5 +1,5 @@
 import { QueryClient } from '@tanstack/query-core'
-import { beforeEach, describe, expect, it } from 'vitest'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { broadcastQueryClient } from '..'
 import type { QueryCache } from '@tanstack/query-core'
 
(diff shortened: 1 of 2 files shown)


tsushanth:
Closing — this has been open without activity long enough that it's more useful to withdraw and let a fresh PR reflect current `main`/`master` than to keep pinging for review on stale state. Happy to reopen or resubmit if there's still interest.

Issue #10924, opened 2026-06-12
Suspended query continues to suspend until queryFn promise is fullfiled even when the query data is updated


MrBr:
### Describe the bug

Suspense query keeps suspending when data exists. Originally, I noticed the problem in `streamedQuery` in combination with `useSuspenseQuery`. The issue reflects in such way that suspense query keeps suspending even after the first stream chunk is received. 

I'd like to note that the aforementioned behaviour could be useful in some cases, but that's a separate issue I guess.

### Your minimal, reproducible example

https://codesandbox.io/p/sandbox/cranky-benz-xvt2n2

### Steps to reproduce

1. Go to https://codesandbox.io/p/sandbox/cranky-benz-xvt2n2?file=%2Fsrc%2FApp.js
2. Click the button
3. Loading text remains even though the query client has the data for the key

### Expected behavior

I'd expect data to show.

### How often does this bug happen?

Every time

### Platform

- Every platform

### Tanstack Query adapter

None

### TanStack Query version

5.101.0

### Additional context

An idea about solution. Maybe wrapping suspense queryFn result with additional promise could provide more control?

Not sure what is the right way to handle the ongoing promise when the data is set programatically? Should this behaviour be controlled by user through options?

nijuse:
@MrBr  There is a way to manually update the cache with a button.
You can keep using useSuspenseQuery, but first cancel the fetch promise to release Suspense.
```
<button
  onClick={async () => {
    await queryClient.cancelQueries({ queryKey })
    queryClient.setQueryData(queryKey, "MANUAL OK")
  }}
>
  Set query state
</button>
```

MrBr:
@nijuse I've demonstrated the root cause of the issue but the actual issue occurs in `streamedQuery`. Since `useSuspenseQuery` will continue to suspend until streamed query Async generator is "done" it makes it unusable with suspense. In most cases (I'd say), you want to show whatever you have in the cache instead of suspending.

cc @TkDodo, we talked about this in Amsterdam on React Summit. Do you have an idea how to tackle this?

TkDodo:
i have an attempted fix here:

- https://github.com/TanStack/query/pull/10994

TkDodo:
@MrBr can you maybe try out the preview build if that fixes your real issue with streaming too?

```
npm i https://pkg.pr.new/@tanstack/react-query@10994
```

MrBr:
@TkDodo I tested it and it seems something isn't working right. Few things I've noted:

[...]

PR #10994, opened 2026-06-27
fix: resolve suspense queries when data gets into the cache by other means

TkDodo:
fixes
- #10924

## Summary by CodeRabbit

* **Bug Fixes**
  * Query promises now remain synchronized with cached data when `setQueryData` occurs during an in-flight fetch.
  * Suspense queries resolve immediately from newly available cached data, then update when the original request completes.
  * Improved error boundary retry suppression to match the current “throw on error” behavior.
* **Tests**
  * Added coverage for in-flight promise resolution after cache updates (React and Preact Suspense).
  * Updated an existing Suspense/error boundary test flow for clearer, stable assertions.

### DIFF
--- a/.changeset/silly-tires-brake.md
+++ b/.changeset/silly-tires-brake.md
@@ -1,6 +1,6 @@
 ---
 '@tanstack/react-query': patch
-'@tanstack/peact-query': patch
+'@tanstack/preact-query': patch
 '@tanstack/query-core': patch
 ---
 
--- a/.changeset/silly-tires-brake.md
+++ b/.changeset/silly-tires-brake.md
@@ -1,5 +1,6 @@
 ---
 '@tanstack/react-query': patch
+'@tanstack/peact-query': patch
 '@tanstack/query-core': patch
 ---
 
--- a/.changeset/silly-tires-brake.md
+++ b/.changeset/silly-tires-brake.md
@@ -1,0 +1,6 @@
+---
+'@tanstack/react-query': patch
+'@tanstack/query-core': patch
+---
+
+fix(react): make sure suspense queries can resolve if data gets into the cache in other ways by tracking a promise from the query directly instead of tying it to the fetch that triggered it.
--- a/packages/preact-query/src/__tests__/useSuspenseQuery.test.tsx
+++ b/packages/preact-query/src/__tests__/useSuspenseQuery.test.tsx
@@ -176,6 +176,45 @@ describe('useSuspenseQuery', () => {
     expect(queryFn).toHaveBeenCalledTimes(1)
   })
 
+  it('should resolve suspense with cached data when data is set while fetching', async () => {
+    const key = queryKey()
+
+    function Page() {
+      const state = useSuspenseQuery({
+        queryKey: key,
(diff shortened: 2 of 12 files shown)


.changeset/silly-tires-brake.md
```diff
@@ -0,0 +1,7 @@
+---
+'@tanstack/react-query': patch
+'@tanstack/peact-query': patch
```
TkDodo:
```suggestion
'@tanstack/preact-query': patch
```

gonzoblasco:
I've been comparing the two approaches for fixing #10924 (this PR and #11036). Sharing some observations:

---

**Architecture comparison**

This PR introduces an `updateThenable` function in `thenable.ts` that externalizes promise resolution based on query state changes. It's a more structural fix — the query itself tracks a thenable that can be resolved by cache updates, not just the observer's `fetchOptimistic`.

#11036 takes a localized approach — `Promise.race` inside `fetchOptimistic` only, using the existing cache subscribe mechanism.

---

**Trade-offs of this PR**

**Pros:**
- More structurally sound — the thenable is tied to the query, so all observers benefit
- Handles the `useQueries` path (modifies `useQueries.ts` too)
- The `updateThenable` logic correctly handles status transitions (pending → fulfilled → rejected) and recreates thenables when data changes

**Concerns:**

1. **Surface area**: 11 files touched, including `hydration.ts`, `query.ts`, `queryObserver.ts`, `useBaseQuery.ts` (both react and preact), `useQueries.ts`, and `errorBoundaryUtils.ts`. This is a lot of core surface for a suspense edge case. The maintainer will need to verify each interaction carefully.

2. **`errorBoundaryUtils.ts` signature change**: `ensurePreventErrorBoundaryRetry` now takes a `query` parameter. This changes an API used by framework adapters — solid, vue, lit adapters would need updating too, but they're not touched in this PR. This could break non-react adapters silently.

3. **Bundle size**: +0.13% on react full (12.14 KB). Small, but #11036 adds zero to the bundle since it's internal to `queryObserver.ts`.

4. **The `thenable.ts` addition**: `updateThenable` adds 40 lines of promise management logic with status-based branching. This is powerful but introduces a new concept that needs to be maintained and understood by future contributors.

---

**Suggestion**

The core insight — that the query's thenable should be resolvable by cache updates, not just fetch completion — is correct and more fundamental than #11036's approach. But the implementation could be scoped tighter:

- Could `updateThenable` be used without changing `errorBoundaryUtils.ts`'s signature? That would reduce the adapter-breaking change.
- The `useQueries.ts` change is valuable — #11036 doesn't handle that path. But is it necessary for the initial fix, or could it be a follow-up?

If the maintainer prefers the structural approach, this PR is the right direction, but I'd recommend narrowing the scope to reduce the blast radius. If they prefer minimal changes, #11036 is the safer path with the caveat of the floating promise I noted there.

TkDodo:
we’re gonna go with https://github.com/TanStack/query/pull/11036 instead

gonzoblasco:
Makes sense, #11036 keeps the blast radius to fetchOptimistic and avoids the adapter-breaking change. Glad the comparison was useful. Thanks for the follow-up.

PR #11003, opened 2026-06-29
docs(svelte): replace deprecated isLoading with isPending in overview

poetryofcode:
## Summary

The Svelte overview example used `query.isLoading` which is a deprecated alias in TanStack Query v5. The correct property is `query.isPending`.

- All other framework overviews (React, Vue, Solid, Angular) already use `isPending`
- `isLoading` was deprecated in v5 and is now just an alias for `isPending && isFetching`

## Changes

- `docs/framework/svelte/overview.md`: Changed `{#if query.isLoading}` → `{#if query.isPending}`

## Test plan
- [ ] Verify `isPending` matches the API reference docs
- [ ] Check other framework overview files use the same pattern

## Summary by CodeRabbit

* **Documentation**
  * Updated the Svelte example to use the current loading-state check in the conditional rendering example.


sukvvon:
This change was already applied in #10990, so this PR is now redundant and I'll close it.

One clarification for the record: `isLoading` isn't deprecated. It's a derived flag (`isPending && isFetching`), so it's perfectly valid to keep using it. The one that's deprecated is `isInitialLoading`, which is an alias of `isLoading`. In this particular example `isPending` is still the better fit, since the snippet only needs to reflect the initial pending state.

PR #11005, opened 2026-06-29
docs: replace deprecated isLoading with isPending in disabled queries guides

DakshSinghDhami:
Updates the `disabling-queries` guides across React, Vue, Solid, and Angular frameworks to use `isPending` instead of the deprecated `isLoading` flag, which was replaced in TanStack Query v5.

`isLoading` is still available as an alias but will be removed in the next major version.

## Summary by CodeRabbit

* **Documentation**
  * Updated query disabling examples across Angular, React, Solid, and Vue to use the newer pending state in loading conditions.
  * Aligned the displayed “Loading...” behavior with the updated state handling in all examples.
  * Clarified the initial disabled-query state in the React guide.

### DIFF
--- a/docs/framework/angular/guides/disabling-queries.md
+++ b/docs/framework/angular/guides/disabling-queries.md
@@ -22,14 +22,14 @@ replace: { 'useQuery': 'injectQuery' }
     } @else {
       @if (query.isError()) {
         <span>Error: {{ query.error().message }}</span>
-      } @else if (query.isLoading()) {
+      } @else if (query.isPending()) {
         <span>Loading...</span>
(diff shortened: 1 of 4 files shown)


sukvvon:
In these examples the query is `enabled: false`, so it's `isPending: true` but `fetchStatus: 'idle'` → `isLoading` (`isPending && isFetching`) is `false`, which is what lets it fall through to **"Not ready ..."**. Switching to `isPending` would show "Loading..." even though nothing is fetching. Also, `isLoading` isn't deprecated in v5 — the deprecated alias is `isInitialLoading`.

So `isLoading` is intentional here, and I'll close this PR.

PR #11013, opened 2026-07-01
test(query-core/utils): add tests for 'undefined' key properties and 'isValidTimeout' validation

whizzkid1452:
## 🎯 Changes

Adds tests for `query-core` utils behavior that was executed but never asserted:

- `hashKey` / `partialMatchKey`: query key objects with `undefined`-valued properties are treated the same as objects without those properties.
- `isValidTimeout`: accepts finite non-negative numbers and rejects `-1`, `NaN`, `Infinity`, non-numbers and `undefined`. This function gates `staleTime`, `gcTime` and `refetchInterval`, but had no direct test coverage.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/docs/community-resources.md
+++ b/docs/community-resources.md
@@ -186,7 +186,7 @@ others:
       description: 'VS Code extension for TanStack Query (React Query): visualize query keys, cache invalidation/refetch flows, and file impact graph',
     },
     {
-      title: 'TanStack Query Visualizer',
+      title: 'Tanstack Query Visualizer',
       url: 'https://tanstack-query-visualizer.sofi.coop/',
       description: 'An interactive sandbox that visualizes the relationship between mutations and query keys.',
     },
--- a/docs/community-resources.md
+++ b/docs/community-resources.md
@@ -186,7 +186,7 @@ others:
       description: 'VS Code extension for TanStack Query (React Query): visualize query keys, cache invalidation/refetch flows, and file impact graph',
     },
     {
-      title: 'Tanstack Query Visualizer',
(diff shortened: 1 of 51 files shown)


docs/community-resources.md
```diff
@@ -186,7 +186,7 @@ others:
       description: 'VS Code extension for TanStack Query (React Query): visualize query keys, cache invalidation/refetch flows, and file impact graph',
     },
     {
-      title: 'Tanstack Query Visualizer',
+      title: 'TanStack Query Visualizer',
```
sukvvon: This capitalization fix is correct, but it's unrelated to the query-core test coverage this PR is about, so let's keep it out of here.

```suggestion
      title: 'Tanstack Query Visualizer',
```

packages/query-core/src/utils.ts
```diff
@@ -105,7 +105,7 @@ export function functionalUpdate<TInput, TOutput>(
 }
 
 export function isValidTimeout(value: unknown): value is number {
-  return typeof value === 'number' && value >= 0 && value !== Infinity
+  return typeof value === 'number' && Number.isFinite(value) && value >= 0
```
sukvvon: Both versions behave identically — `NaN` and `-Infinity` are already rejected by `value >= 0`, so this doesn't fix anything, and the new tests pass either way. It's not worth touching query-core for a stylistic change, and this PR is about test coverage, so let's leave `isValidTimeout` untouched.

```suggestion
  return typeof value === 'number' && value >= 0 && value !== Infinity
```

sukvvon (approved):

PR #11017, opened 2026-07-02
fix(solid-query): restore observer subscription after SSR hydration

brenelz:
## Problem

`useQuery` components hydrated from SSR are **permanently non-reactive**: the DOM renders the server snapshot correctly, but `setQueryData`, refetches, and invalidations never update it. With a loader that does not await the query this also surfaced as hydration errors.

During hydration, Solid restores `useBaseQuery`'s resource memo from the value the server serialized, so the compute's wiring never runs on the client:

- **sync-serialized values** (query settled on the server before the shell flushed — the common case): the compute is skipped entirely
- **async-serialized values** (query streamed as a pending promise): the compute is replayed under a constructor-less `MockPromise` stub (`subFetch`/`readHydratedValue`) that silently drops the Promise executor's side effects

Either way, the `QueryObserver` subscription that lives inside the resource Promise executor is never established — the observer ends up with zero subscribers and cache updates never reach the component's store.

## Fix

Recovery runs on **unowned `setTimeout` polls** after hydration has fully completed, in two stages (`packages/solid-query/src/useBaseQuery.ts`):

1. `recoverAfterHydration()` — for all components created by the hydration walk: waits for `sharedConfig.done` (boundary hydration is asynchronous; subscribing earlier delivers an immediate `updateResult()` notification whose store write re-renders the component mid-walk and breaks DOM claiming), then re-establishes the observer subscription — unless the executor ran natively and wired itself — plus a guarded store catch-up that only writes when the result actually changed.
2. `finishPendingRecovery()` — streamed-pending queries only (`hydratedAsPending && !computeWired`): waits for the hydrated resource to settle, seeds the cache from the snapshot's `hydrationData` when nothing fresher exists (standalone Solid SSR without a router integration), then `refresh(queryResource)` re-runs the compute under the native `Promise` for the full normal wiring.

### Why no owned computations

Two earlier approaches using a recovery `createRenderEffect` regressed hydration itself: a **client-only** owned computation created during the hydration walk shifts the hydration key (`getNextChildId`) of every subsequent computation in the component, so JSX claiming fails — "Hydration completed with N unclaimed server-rendered node(s)" warnings and duplicated inert DOM. Both recovery stages therefore run on unowned timers guarded by the component's `disposed` flag.

## Verification

- `@tanstack/solid-query` unit tests: 322/322 passing; `eslint` and `tsc` clean
- TanStack Router monorepo e2e (solid-start SSR + streaming, built against this branch):
  - `query-integration` suite 8/8, including a new liveness regression test: visit a route whose loader kicks off `prefetchQuery` without awaiting, then `setQueryData` after hydration must re-render the region
  - `solid-query-layout-suspense` (repro for TanStack/router#7195) passing
  - `basic-solid-query` / `basic-solid-query-file-based` (CSR) 3/3 each
- Dev-mode probes show no unclaimed-node warnings and live post-hydration updates

### DIFF
--- a/.changeset/solid-query-hydration-observer-recovery.md
+++ b/.changeset/solid-query-hydration-observer-recovery.md
@@ -1,0 +1,25 @@
+---
+'@tanstack/solid-query': patch
+---
+
+fix(solid-query): restore observer subscription and store sync after SSR hydration
+
+During hydration, Solid restores `useBaseQuery`'s resource memo from the
+value the server serialized, so the compute's wiring never runs on the
+client: sync-serialized values (query settled before the shell flushed)
+skip the compute entirely, and async-serialized values (query streamed as
+a pending promise) replay it under a constructor-less `MockPromise` stub
+that drops the Promise executor's side effects. Either way the
+`QueryObserver` subscription was never established, leaving hydrated
+components permanently inert — `setQueryData`, refetches, and
+invalidations never reached the DOM.
+
+Recovery now runs on unowned timers after hydration fully completes
+(`sharedConfig.done`, which accounts for asynchronously resumed Loading
+boundaries): the observer subscription is re-established with a guarded
+store catch-up, and streamed-pending queries additionally seed the cache
+from the serialized snapshot and re-run the compute under the native
+`Promise`. Owned computations are deliberately avoided — a client-only
+render effect created during the walk shifts the hydration keys of every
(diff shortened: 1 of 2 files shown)


brenelz:
[solid-query-v6-pre](https://github.com/TanStack/query/tree/solid-query-v6-pre) is undergoing a rewrite so will need to re-review this after

PR #11036, opened 2026-07-08
fix(query-core): resolve suspense when query data is set programmatically

### LINKED ISSUE #10924: Suspended query continues to suspend until queryFn promise is fullfiled even when the query data is updated

AmariahAK:
## PR Description

```markdown
When `useSuspenseQuery` is used with `streamedQuery` or `setQueryData`,
the Suspense boundary stays stuck on the original fetch promise even
after data arrives in the cache. This happens because `fetchOptimistic`
in `queryObserver.ts` returned `query.fetch().then(...)` — a promise
that only resolves when the `queryFn` completes, ignoring cache updates
from other sources.

**Fix:** Modified `fetchOptimistic` to use `Promise.race` between the
fetch promise and a query cache subscriber. The subscriber listens for
`'updated'` events on the same query hash and resolves immediately when
`query.state.data` becomes defined — whether via `setQueryData`, a
streamed chunk, or any other mechanism. The fetch continues running
in the background; only the suspense block is released early.

**Why `Promise.race` over alternatives:**
- Resolving the fetch promise early would abort the ongoing fetch
  (breaks streamed queries' abort signal — the approach attempted in
  preview PR #10994)
- Switching to an observer-based throw in `useBaseQuery` would add
  undue complexity
- `Promise.race` with the existing cache notification system is a
  20-line diff that touches nothing else

**Files changed:**

| File | Change |
|------|--------|
| `packages/query-core/src/queryObserver.ts` | +20/−1 in `fetchOptimistic` |
| `packages/react-query/src/__tests__/useSuspenseQuery.test.tsx` | +102 lines (3 new tests) |
| `.changeset/resolve-suspense-setquerydata.md` | changeset (patch) |

**New tests:**
1. `setQueryData` called while fetch is in-flight → suspense releases
2. `streamedQuery` first chunk → suspense releases
3. `setQueryData` before component mounts → suspense never shows

## ✅ Checklist

- [x] I have followed the steps in the
  [Contributing guide](https://github.com/TanStack/query/blob/main/CONTRIBUTING.md).

```
@tanstack/query-core:test:lib    →  546 passed, 0 failures
@tanstack/react-query:test:lib   →  567 passed, 0 failures
```

## 🚀 Release Impact

- [x] This change affects published code, and I have generated a
  [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).

Closes #10924

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved React Suspense release behavior for optimistic and streaming queries, resolving boundaries when cache data becomes available through programmatic updates.
  * Suspense now releases after the first available streamed chunk.
  * Preloaded cached data avoids initial suspension; setting cached data to `undefined` no longer triggers an incorrect release.
* **Tests**
  * Added coverage for cache updates during fetching, streamed first-chunk release, preloaded-cache mounting, and the `undefined` scenario.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### CODE REVIEW

**Comment by gonzoblasco** [comment]
Thanks for this PR — the `Promise.race` approach is elegantly minimal compared to alternatives. I've been reviewing the implementation and have a few observations:

---

**1. Floating promise when fetch wins the race**

When the fetch promise resolves first, `.finally(() => unsubscribe())` correctly removes the cache subscriber, but the second promise (the subscriber one) remains in a permanently pending state. This is benign in practice (the listener is removed, so no leaks via the `Set`), but some linting rules (`@typescript-eslint/no-floating-promises`) and environments flag this pattern.

A cleaner approach would be to make the subscriber promise self-resolving on cleanup:

```ts
let resolveEarly: ((result: QueryObserverResult<TData, TError>) => void) | undefined

const cachePromise = new Promise<QueryObserverResult<TData, TError>>((resolve) => {
  resolveEarly = resolve
  unsubscribe = this.#client.getQueryCache().subscribe((event) => {
    if (
      event.type === 'updated' &&
      event.query.queryHash === query.queryHash &&
      query.state.data !== undefined
    ) {
      unsubscribe()
      resolve(this.createResult(query, defaultedOptions))
    }
  })
})

return Promise.race([
  query.fetch().then(() => {
    return this.createResult(query, defaultedOptions)
  }).finally(() => {
    unsubscribe()
    // Resolve the floating promise to avoid it pending forever
    if (resolveEarly) {
      // Subscriber never fired — resolve with current result so the promise settles
      // This value is ignored by Promise.race since the fetch branch already won
      resolveEarly(this.createResult(query, defaultedOptions))
    }
  }),
  cachePromise,
])
```

This ensures both promises always settle, regardless of which wins.

---

**2. No cleanup if the component unmounts mid-fetch**

If the consuming component unmounts while `fetchOptimistic` is in-flight, the cache subscriber remains active until the fetch promise settles (via `.finally`). During that window, any `setQueryData` call on the same `queryHash` will trigger `createResult` and resolve the promise — but nobody is listening anymore. Not a correctness bug, but unnecessary work.

Consider adding an AbortSignal-aware cleanup, or at minimum documenting that `fetchOptimistic` assumes the caller will remain mounted through resolution.

---

**3. Multiple observers on the same query hash**

The cache subscriber listens to `'updated'` events filtered by `queryHash`. If two `useSuspenseQuery` instances share the same key and one receives `setQueryData`, both subscribers fire. This is actually correct behavior (both should unsuspend), but worth noting in a test case to prevent regressions if the filter logic changes later.

---

**4. Test coverage consideration**

The three tests are solid. One edge case that might be worth adding: what happens when `setQueryData` is called with `undefined`? The current check `query.state.data !== undefined` guards against this, but an explicit test would document the intent.

---

Overall, this is the right approach — much less invasive than #10994 and uses the existing notification system well. The main actionable item is settling the floating promise in point 1.

AmariahAK:
@gonzoblasco  hey, apologies for seeing this late
So ive done the things you suggested
So could you re-take a look when you get time?

gonzoblasco:
@AmariahAK thanks for the quick turnaround. I took a look at the latest commit.

**The floating promise fix looks good** — the `resolveEarly` pattern in `.then()` ensures both branches of `Promise.race` always settle. The `setQueryData(undefined)` test is a nice addition that documents the guard correctly.

**One thing still open:** the unmount cleanup concern. If the component unmounts while `fetchOptimistic` is in-flight, the cache subscriber stays registered until the fetch promise settles (via `.finally`). During that window, any `updated` event for the same queryHash would call `resolve(this.createResult(...))` on a stale observer, which is harmless in practice (React won't commit the result) but the subscriber itself leaks in the `QueryCache.subscribers` Set.

A simple guard would be to check if the observer is still mounted before resolving:

```ts
unsubscribe = this.#client.getQueryCache().subscribe((event) => {
  if (
    event.type === 'updated' &&
    event.query.queryHash === query.queryHash &&
    query.state.data !== undefined
  ) {
    unsubscribe()
    resolve(this.createResult(query, defaultedOptions))
  }
})
```

Could add a `this.#isMounted` check or use an AbortSignal from the observer's lifecycle. But honestly, for a patch release this is probably fine — the leak is bounded (until fetch completes) and the impact is minimal.

**Re: #10994** — Dominik's approach is more structural but has a larger blast radius. I think both PRs solve the core issue, and whichever the maintainers prefer is fine. Yours is the safer incremental fix.

AmariahAK:
@gonzoblasco thanks for the re-review. On the unmount cleanup concern — I looked into adding a liveness guard (`hasListeners()` check before `resolve`) but it breaks during normal suspense: since `useSyncExternalStore` only subscribes in the commit phase and `fetchOptimistic` throws during render, the observer has zero listeners at the point the subscriber fires, even for a fully alive component. Any check tied to the listener count would block the cache subscriber from ever resolving.

I agree with your assessment that this is acceptable for a patch release , the subscriber entry lives only until `query.fetch()` settles via `.finally()`, and any stale `resolve()` on a destroyed observer is harmless (React ignores it). Happy to revisit with an AbortSignal-based approach in a follow-up if the maintainers want it, but the cost/benefit doesn't justify holding this fix.

All other points are resolved — floating promise settled, `undefined` test added, 546/567 tests passing. Ready for another look when you have time.

gonzoblasco:
@AmariahAK just re-reviewed the latest diff. Everything looks good:

- ✅ Floating promise settled via `resolveEarly` in the `.then()` branch
- ✅ `setQueryData(undefined)` test documents the guard correctly
- ✅ `.finally(() => unsubscribe())` ensures cleanup
- ✅ Tests pass (546/567)

Your reasoning on the `hasListeners()` guard makes sense — during Suspense the observer has zero listeners at that point, so any listener-based check would block resolution. Agreed that the bounded subscriber lifetime (until fetch settles) is acceptable for a patch release.

This LGTM. Ready for maintainer review. 🚀

packages/query-core/src/queryObserver.ts
```diff
@@ -317,7 +317,44 @@ export class QueryObserver<
...
+    return Promise.race([
+      query
+        .fetch()
+        .then(() => {
+          const result = this.createResult(query, defaultedOptions)
+          // Settle the subscriber promise so both branches always settle.
+          // This value is ignored by Promise.race since the fetch branch already won.
+          if (resolveEarly) {
+            resolveEarly(result)
+          }
```
TkDodo:
```suggestion
          resolveEarly?.(result)
```

TkDodo (approved):

PR #11061, opened 2026-07-13
fix(query-core): don't reject imperative fetch with CancelledError when a superseding fetch is started

### LINKED ISSUE #8060: `invalidateQueries` inconsistently cancels `fetchQuery`
### Describe the bug

`fetchQuery` may throw an `CancelledError { silent: true, revalidate: undefined }` under rare conditions:
- the query is stale (otherwise the fetch won't run).
- the invalidation happens at least one tick after the fetch started.
- **the query is in an `active` state. For example because a component with `useQuery` is still mounted.**

In the real world, this might happen in combination with `react-router`, when the user submits a form and then navigates to a new (or the same) page. In my case, we were updating a search param to close a modal after a form submission, which then triggered the loader function. We're also using a mutation observer[^1] to trigger invalidations, which is why the invalidation only happened after the navigation began.

I fixed the issue by using `ensureQueryData` instead of `fetchQuery`. This is the better choice anyways.

Still, this seems like a footgun since it is difficult to catch while testing. Further, if the intended behavior for `fetchQuery` is to throw a `CancelledError` on invalidation, it should *always* do that and not only when a revalidation is triggered.

[^1]: inspired by https://tkdodo.eu/blog/automatic-query-invalidation-after-mutations

https://codesandbox.io/p/sandbox/react-query-cancellederror-on-invalidation-d4g259

### Steps to reproduce

See steps above

### Expected behavior

Option A: The returned promise should _consistently_ reject, regardless of whether an unrelated observer exists.
Option B: The returned promise should _consistently_ resolve with the data returned from this fetch
Option C: The QueryClient should treat pending promises as observers, revalidate automatically and always resolve with the result from the latest revalidation.

Every time

### Screenshots or Videos

_No response_

### Platform

*

### Tanstack Query adapter

react-query

### TanStack Query version

5.50.1

### TypeScript version

_No response_

### Additional context

_No response_

xianjianlf2:
## Problem

When a fetch is silently cancelled because a new fetch supersedes it (e.g.
`invalidateQueries` refetching an active observer with `cancelRefetch: true`),
callers that joined the in-flight fetch via the early-return in `Query.fetch`
received the raw retryer promise, which rejects with a silent `CancelledError`.

The async `fetch` body already piggybacks onto the superseding fetch on silent
cancellation, but the early-return path bypassed it, so an imperative
`fetchQuery` could reject with a `CancelledError` even though the query
continued fetching.

## Fix

Route both the early-return and the catch-block piggyback through a shared
helper that follows the chain of superseding fetches until one settles, while
still rethrowing when the silent cancellation isn't from a superseding fetch
(e.g. query destroyed).

## Testing

Added a regression test in `packages/query-core/src/__tests__/queryClient.test.tsx`
covering the scenario where an imperative fetch is superseded by an
`invalidateQueries`-triggered refetch, asserting the imperative call resolves
with the fresh data instead of rejecting with a `CancelledError`.

## Checklist

- [x] Added regression coverage for the superseding-fetch path
- [x] Preserved the existing silent-cancellation behavior for destroyed queries
- [x] Kept the change scoped to `query-core`

## Release Impact

Low risk bug fix in `query-core`. The change only corrects how imperative fetches
follow a superseding refetch; it should not affect callers outside the failed
`CancelledError` path.

Closes #8060

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed query fetching when an in-progress request is invalidated and refetched, ensuring the latest result is used.
  * Prevented silently cancelled fetches from failing incorrectly when a newer request supersedes them.
  * Preserved existing cancellation behavior when a query is explicitly destroyed.

* **Tests**
  * Added regression coverage for fetch resolution during invalidation while an active query is refetching.


xianjianlf2:
Addressed the assertion-strengthening review in 2c8a3ec. I used `toBe(3)` rather than `toBe(2)` after verifying the sequence locally: 1 is the priming fetch, 2 is the observer's background refetch, and the invalidate-triggered superseding fetch resolves to 3.\n\nLocal check:\n- `PATH=/Users/even/.nvm/versions/node/v24.18.0/bin:/Users/even/.nvm/versions/node/v20.20.2/bin:/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/opt/pmk/env/global/bin:/Library/Apple/usr/bin:/Users/even/.codex/tmp/arg0/codex-arg0VOKUrl:/Users/even/.cache/codex-runtimes/codex-primary-runtime/dependencies/bin/override:/Users/even/.bun/bin:/Users/even/.local/bin:/Users/even/.nvm/versions/node/v20.20.2/bin:/Users/even/.cargo/bin:/Users/even/.cache/codex-runtimes/codex-primary-runtime/dependencies/bin/fallback:/Applications/ChatGPT.app/Contents/Resources pnpm --dir packages/query-core test:lib src/__tests__/queryClient.test.tsx`

packages/query-core/src/query.ts
```diff
@@ -397,6 +397,27 @@ export class Query<
     }
   }
 
+  // When a fetch is silently cancelled, a new fetch has been started to
+  // supersede it. Any caller holding the superseded promise should follow
+  // that new fetch instead of receiving the internal `CancelledError`. This
+  // walks the chain of superseding fetches until one settles for real. When
+  // the silent cancellation does not come from a superseding fetch (e.g. the
+  // query was destroyed), `this.#retryer` still points at the settled promise,
+  // so we rethrow instead of looping.
```
TkDodo:
```suggestion
  // A silent cancellation may be caused by a superseding fetch. If the
  // current retryer is different from the cancelled promise, follow it.
  // Otherwise (for example, when the query is destroyed), preserve the
  // cancellation. Repeat this for a chain of superseding fetches.
```

TkDodo (approved):

TkDodo (changes requested):
I forked the initial reproduction from the issue and used the preview build to check it, but it seems to still throw the same unexpected CancelledError:

https://codesandbox.io/p/devbox/react-query-cancellederror-on-invalidation-forked-ymg4jw?workspaceId=ws_XiBCv7MgHSszKyXHXG5ray

PR #11068, opened 2026-07-16
fix(query-core): stop orphaned streamedQuery chunks from corrupting the cache

a-y-ibrahim:
## 🎯 Changes

When a `streamedQuery`-backed query is refetched while a previous stream is still flowing, the old fetch is cancelled but its `streamFn` keeps running unless it explicitly reads `context.signal`. Many `streamFn`s can't do that (e.g. a third-party SDK with no `AbortSignal` support), so the orphaned stream kept calling `setQueryData` after the new fetch had already reset the cache, interleaving stale chunks into what should be fresh data.

This tracks the current stream per `Query` independently of signal consumption, so an orphaned stream notices it has been superseded and stops writing, regardless of whether its `streamFn` ever touches the signal.

Reproduced and fixed with a new test (`should not let an orphaned stream write to the cache after a refetch when the signal is not consumed`) that fails on current `main` and passes with this change.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed cases where “orphaned” streamed query chunks could overwrite newer cached results after a refetch.
  * Ensured refetches in replace mode don’t commit stale superseded stream results.
* **Tests**
  * Added refetch supersession coverage to verify only the latest streamed output is written to the cache after restarts/cancellations.

### DIFF
--- a/packages/query-core/src/streamedQuery.ts
+++ b/packages/query-core/src/streamedQuery.ts
@@ -108,7 +108,7 @@ export function streamedQuery<
       activeStreamPerQuery.set(query, streamToken)
     }
     const isSuperseded = () =>
-      query !== undefined && activeStreamPerQuery.get(query) !== streamToken
+      !!query && activeStreamPerQuery.get(query) !== streamToken
 
     const stream = await streamFn(streamFnContext)
 
--- a/packages/query-core/src/streamedQuery.ts
+++ b/packages/query-core/src/streamedQuery.ts
@@ -1,10 +1,22 @@
 import { addConsumeAwareSignal, addToEnd } from './utils'
+import type { Query } from './query'
 import type {
   OmitKeyof,
   QueryFunction,
   QueryFunctionContext,
   QueryKey,
 } from './types'
+
+// Tracks, per Query, which streamedQuery() invocation is the current one.
+// A refetch cancels the previous fetch's AbortSignal, but that only stops an
+// in-flight stream if its streamFn actually reads (consumes) context.signal.
+// Many streamFns can't do that (e.g. a third-party SDK that doesn't accept
+// an AbortSignal), so this provides an unconditional way for an orphaned
+// stream to notice it has been superseded and stop writing to the cache,
+// independent of whether the signal was ever consumed. Keyed by the Query
+// instance (not the query key) so entries are naturally released once the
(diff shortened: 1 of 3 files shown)


a-y-ibrahim:
Addressed the CodeRabbit nitpick in 9497dfad3: added a dedicated test covering the refetchMode 'replace' finalize guard (a superseded stream's accumulated result must not overwrite the current one), verified it fails without the isSuperseded() check on that path and passes with it.

packages/query-core/src/streamedQuery.ts
```diff
@@ -1,11 +1,23 @@
...
 
+// Tracks, per Query, which streamedQuery() invocation is the current one.
+// A refetch cancels the previous fetch's AbortSignal, but that only stops an
+// in-flight stream if its streamFn actually reads (consumes) context.signal.
+// Many streamFns can't do that (e.g. a third-party SDK that doesn't accept
+// an AbortSignal), so this provides an unconditional way for an orphaned
+// stream to notice it has been superseded and stop writing to the cache,
+// independent of whether the signal was ever consumed. Keyed by the Query
+// instance (not the query key) so entries are naturally released once the
+// query itself is garbage collected.
```
TkDodo, line 18: this is a noisy comment we likely don’t need

packages/query-core/src/streamedQuery.ts
```diff
@@ -91,12 +103,19 @@ export function streamedQuery<
       () => (cancelled = true),
     )
 
+    const streamToken = Symbol()
+    if (query) {
+      activeStreamPerQuery.set(query, streamToken)
+    }
+    const isSuperseded = () =>
+      !!query && activeStreamPerQuery.get(query) !== streamToken
+
```
TkDodo, line 112: let’s refactor this towards:

```
const isStreamInactive = () =>
  cancelled ||
  (!!query && activeStreamPerQuery.get(query) !== streamToken)
```

and then use this as the only condition, basically binding it to the `cancelled` check:

```
if (isStreamInactive()) {
  break
}
```

and

```
if (isReplaceRefetch && !isStreamInactive()) {
```

TkDodo (changes requested):

PR #11072, opened 2026-07-17
fix: replace experimentalDts with dts to fix TS2883 with nodenext module resolution

lokesh-7977:
## 🎯 Changes

Replace `experimentalDts: true` with `dts: true` in all tsup configs to fix
TS2883 errors when using `module: "nodenext"` with declaration emit.

`experimentalDts` generates a shared `_tsup-dts-rollup.cjs` file that TypeScript
cannot reference portably from `.d.cts` re-exports under nodenext resolution.
Switching to `dts` generates per-entry declaration files, eliminating the
problematic internal path.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Build Improvements**
  * Updated package builds to use the standard TypeScript declaration generation setting.
  * Ensured modern and legacy builds continue producing type declaration files consistently.


TkDodo (approved):

TkDodo:
sadly this fails the build now. we have plans to move away from `tsup` in the near future.

lokesh-7977:
Got it. Since the `TS2883` issue is still open, would it make sense to address it as part of the planned `tsup` migration? If so, is there already a preferred replacement/build setup I could work with?

PR #11073, opened 2026-07-17
perf: use stable-hash for hashing keys

43081j:
Using `JSON.stringify` is much slower when we only need it to create a stable value hash. Instead, we can use `stable-hash`.

This is added as a devDependency so it ends up in the bundle rather than being a production dependency.

On my machine, some bench results:

| Task name | Latency avg (ns) | Latency med (ns) | Throughput avg (ops/s) | Throughput med (ops/s) | Samples |
| -- | -- | -- | -- | -- | -- |
| 'hashKey dev' | '29.05 ± 0.10%' | '41.00 ± 1.00' | '27211238 ± 0.01%' | '24390244 ± 580720' | 17211565 |
| 'hashKey prod' | '942.42 ± 3.32%' | '916.00 ± 1.00' | '1096146 ± 0.01%' | '1091703 ± 1191' | 530549 |

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved query-key hashing for complex and non-serializable values.
  * Updated error and warning messages to display readable query keys instead of internal hash values.
  * Improved hydration consistency by recalculating query hashes and honoring custom hashing options.
  * Enhanced diagnostics for missing query functions and structural-sharing issues.
  * Updated devtools to display and filter queries by their readable keys.

* **Tests**
  * Expanded coverage for hydration, custom query-key hashing, query errors, infinite queries, and utility behavior.


packages/query-core/src/__tests__/query.test.tsx
```diff
@@ -861,10 +861,9 @@ describe('query', () => {
     const unsubscribe = observer.subscribe(() => undefined)
 
     await vi.advanceTimersByTimeAsync(10)
-    const query = queryCache.find({ queryKey: key })!
     expect(observer.getCurrentResult()).toMatchObject({
       status: 'error',
-      error: new Error(`Missing queryFn: '${query.queryHash}'`),
+      error: new Error(`Missing queryFn: '${JSON.stringify(key)}'`),
```
DogPawHat, line 946: Shouldn't `describeKey` be used here?

43081j: This is just so we're not inadvertently also testing describeKey. Ideally this wouldn't even use stringify, it'd be a snapshot or something

packages/query-core/package.json
```diff
@@ -59,6 +59,7 @@
   ],
   "devDependencies": {
     "@tanstack/query-test-utils": "workspace:*",
-    "npm-run-all2": "^5.0.0"
+    "npm-run-all2": "^5.0.0",
+    "stable-hash": "^0.0.6"
```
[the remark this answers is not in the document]
43081j: We use tsup, so this actually causes it to be bundled into the published code rather than being a runtime dependency. 

The reason for that was to avoid introducing an extra dep for something we might eventually roll our own of. 

Happy to move it to prod dependencies so it doesn't get bundled though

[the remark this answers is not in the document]
43081j: Typescript isn't published to npm. The bundled output is - which won't contain imports to stable hash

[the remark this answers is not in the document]
43081j: Ugh. That export is the culprit if anything. Typescript shouldn't be published to npm since no regular engines (e.g. node) will import it. They will only load js. 

It's strange that we have a typescript export for react native. I suspect the solution is to fix that, not anything to do with this pr

packages/query-core/src/utils.ts
```diff
@@ -230,16 +231,22 @@ export function hashQueryKeyByOptions<TQueryKey extends QueryKey = QueryKey>(
...
-      ? Object.keys(val)
-          .sort()
-          .reduce((result, key) => {
-            result[key] = val[key]
-            return result
-          }, {} as any)
-      : val,
-  )
+  return stableHash(queryKey)
+}
```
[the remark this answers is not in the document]
43081j: @coderabbitai this is solved by the fact the query's key is recomputed during hydration rather than taking the one we stored

packages/query-core/src/utils.ts
```diff
@@ -1,3 +1,4 @@
+import stableHash from 'stable-hash'
```
TkDodo, line 1: @43081j did we not talk about going the direction of not using the dependency because it also does too much for us, but rather inline the relevant parts ?

43081j: currently it gets inlined by tsup, but there was some mention of also just hand rolling it or at least vendoring it into the repo itself.

i'm happy to do either. do you have any preference? we probably do use most or all of what's in `stable-hash` for what it's worth.

TkDodo: can you address the lint failure please 🙏

PR #11098, opened 2026-07-21
fix(vue-query): export queryOptions declaration types

### LINKED ISSUE #11042: vue-query: `queryOptions()`'s real return type is unexported; the public name with the same identifier points at a different, legacy type
### Describe the bug

Since the 5.98.0 rewrite ("fix(vue-query): fix type of `queryOptions` to allow plain properties or getters"), `queryOptions()` is typed against a **new** `QueryOptions`-based shape. In the published rollup declaration file (`build/modern/_tsup-dts-rollup.d.ts`) this shape exists as `UndefinedInitialQueryOptions_alias_1` / `DefinedInitialQueryOptions_alias_1` — but:

- `_alias_1` is **not exported** from the package root (`build/modern/index.d.ts`).
- The public names `UndefinedInitialQueryOptions` / `DefinedInitialQueryOptions` are re-exported from the root, but they still point at the **old**, `UseQueryOptions`-based, `MaybeRef<...>`-union type. Two genuinely different types share one exported name; only the wrong one is reachable.
- Supporting types referenced by the new signature are unexported too: vue-query's `ShallowOption`, its local `MaybeRefOrGetter` alias, and query-core's `QueryBehavior`, `RetryValue`, `RetryDelayValue`, plus a non-exported `unique symbol` (`dataTagErrorSymbol`).

Any library that exports a function returning `queryOptions(...)` with an **inferred** return type — the exact pattern [`@kubb/plugin-vue-query`](https://kubb.dev/) generates for every endpoint — hits one of two consequences when its declarations are emitted:

1. **Hard declaration-emit errors** (`TS2527`/`TS2883`) — reproduced here with both `tsdown`/`rolldown-plugin-dts` and plain `tsc`.
2. **Silently WRONG declarations** — the natural fix for (1) is to annotate the return type with the public `UndefinedInitialQueryOptions`. That compiles with zero errors, but it is the **wrong type**: a consumer spreading the factory's result into another `queryOptions()` call to add `select` fails to type-check, and the `DataTag` on `queryKey` is lost.

Both are reproduced below, from the same root cause: the name collision.

https://github.com/DesselBane/repro-vue-query-query-options

### Steps to reproduce

See repo README.md

### Expected behavior

No TS2527 error and/or correct type exported so that downstream select works

Every time

### Screenshots or Videos

_No response_

### Platform

| Package | Version |
| --- | --- |
| `@tanstack/vue-query` | `5.101.2` (also confirmed broken at `5.98.0`; `5.97.0` is clean — see [Regression window](#regression-window)) |
| `@tanstack/query-core` | `5.101.2` (transitive) |
| `typescript` | `6.0.3` |
| `vue` | `3.5.31` |
| `tsdown` | `0.21.4` (`rolldown-plugin-dts@0.22.5`, `rolldown@1.0.0-rc.9`) |
| Node | `v24.0.0` |
| OS | Windows 11 |

### Tanstack Query adapter

vue-query

### TanStack Query version

5.98.0 (and 5.101.2)

### TypeScript version

6.0.3

### Additional context

Diagnosis and reproduction were created with the help of CLAUDE. I checked the repo and as far as I can tell this is all correct.

ousamabenyounes:
## 🎯 Changes

- Export reusable `queryOptions` declaration types from Vue Query.
- Expose the matching Query Core declaration types needed by the Vue exports.
- Preserve direct, `ref`, `computed`, and getter `queryKey` shapes while retaining `DataTag` metadata.
- Keep the previous `useQuery` initial-data helper types available as `UndefinedInitialUseQueryOptions` and `DefinedInitialUseQueryOptions`.

Fixes #11042

## ✅ Checklist

Targeted local verification:

- `pnpm exec nx run @tanstack/vue-query:test:types --skip-nx-cache`
- `pnpm exec nx run @tanstack/vue-query:build --skip-nx-cache`

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Expanded the public TypeScript API for Vue Query and core packages with additional query behavior, retry configuration, and query option helper types.
  * Improved `queryOptions` type inference for query keys, data tags, and `select` callbacks, including portable factory patterns.
* **Tests**
  * Strengthened type-level coverage and added declaration-emit checks for reliable compile-time behavior.
* **Release**
  * Published patch releases for the core and Vue Query packages.

### DIFF
--- a/.changeset/calm-vue-query-options.md
+++ b/.changeset/calm-vue-query-options.md
@@ -1,0 +1,6 @@
+---
+'@tanstack/query-core': patch
+'@tanstack/vue-query': patch
+---
+
+Export query option declaration helper types for portable Vue queryOptions factories.
--- a/.changeset/calm-vue-query-options.md
+++ b/.changeset/calm-vue-query-options.md
@@ -4,3 +4,5 @@
 ---
 
 Export query option declaration helper types for portable Vue queryOptions factories.
+The previous `useQuery`-specific initial-data helper types remain available as
+`UndefinedInitialUseQueryOptions` and `DefinedInitialUseQueryOptions`.
--- a/packages/query-core/src/index.ts
+++ b/packages/query-core/src/index.ts
@@ -20,6 +20,8 @@ export type { QueryCacheNotifyEvent } from './queryCache'
 export { QueryClient } from './queryClient'
 export { QueryObserver } from './queryObserver'
 export { CancelledError, isCancelledError } from './retryer'
+export type { RetryDelayValue, RetryValue } from './retryer'
+export type { QueryBehavior } from './query'
 export {
   timeoutManager,
   type ManagedTimerId,
--- a/packages/vue-query/src/__tests__/queryOptions.test-d.ts
+++ b/packages/vue-query/src/__tests__/queryOptions.test-d.ts
@@ -5,6 +5,8 @@ import { queryKey } from '@tanstack/query-test-utils'
 import { QueryClient } from '../queryClient'
 import { queryOptions } from '../queryOptions'
 import { useQuery } from '../useQuery'
+import type { ComputedRef, Ref } from 'vue-demi'
+import type { DataTag } from '@tanstack/query-core'
 import type { QueryOptionsDataTag, UndefinedInitialQueryOptions } from '..'
 
 describe('queryOptions', () => {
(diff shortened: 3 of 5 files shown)


ousamabenyounes:
Rebased onto `main` and resolved the conflict by reducing scope: #11224 already merged the declaration-emit fix (`*WithDataTag`), so I dropped the competing `QueryOptionsDataTag` design and the `useQuery` export renames (they conflicted irreconcilably and re-introduced the name collision). What remains is the part #11224 left undone — re-exporting the helper types #11042 lists: `ShallowOption`/`MaybeRefOrGetter` (vue-query) and `QueryBehavior`/`RetryValue`/`RetryDelayValue` (query-core).

Verified locally: new `publicTypeExports.test-d.ts` in both packages goes RED (`has no exported member`) without the re-exports and GREEN with them; `test:types` (TS 5.7–7.0), `test:eslint`, and prettier all pass on the diff. Branch is now mergeable. Open: awaiting CodeRabbit re-review and CI on the new SHA.

PR #11111, opened 2026-07-23
fix(solid-query): subscribe to new observer when queryClient accessor switches

### LINKED ISSUE #11106: solid-query: switching the queryClient accessor strands the new client's cache (subscription stays on the old observer)
### Describe the bug

When the `queryClient` accessor passed to `useQuery` (second argument) switches to a different `QueryClient`, the hook never actually moves to the new client:

- the refetch triggered by the switch lands in the **previous** client's cache (its `dataUpdateCount` increments),
- the entry created in the **new** client's cache stays `status: 'pending'` / `fetchStatus: 'idle'` forever and never receives data,
- the hook keeps rendering results coming from the old observer, so the UI shows the old client's data while claiming the new client is active.

The root cause is an ordering issue in `useBaseQuery`. `createClientSubscriber` captures the observer eagerly when called ([`useBaseQuery.ts#L180-L181`](https://github.com/TanStack/query/blob/86bb8a6fb2c7f15c74ff50afba053d778e6edc23/packages/solid-query/src/useBaseQuery.ts#L180-L181)):

```ts
const createClientSubscriber = () => {
  const obs = observer() // reads the signal NOW
  return obs.subscribe(/* ... */)
}
```

but in the client-change computed it is called **before** `setObserver(newObserver)` ([`useBaseQuery.ts#L317-L332`](https://github.com/TanStack/query/blob/86bb8a6fb2c7f15c74ff50afba053d778e6edc23/packages/solid-query/src/useBaseQuery.ts#L317-L332)):

```ts
createComputed(
  on(client, (c) => {
    if (unsubscribe) {
      unsubscribe()
    }
    const newObserver = new Observer(c, defaultedOptions())
    unsubscribe = createClientSubscriber() // observer() still returns the OLD observer
    setObserver(newObserver)
  }, { defer: true }),
)
```

So the fresh subscription attaches to the **old** observer. Re-subscribing to the old (now stale) observer triggers its `onSubscribe` → a refetch against the old client's cache — which is where the second `queryFn` call comes from. The new observer ends up with zero subscribers, and a `QueryObserver` without subscribers never fetches, so the new client's cache entry is stranded.

Swapping the two lines (`setObserver(newObserver)` first, then `unsubscribe = createClientSubscriber()`) makes the new client's cache receive the data, with no regressions in the rest of the `solid-query` suite.

The existing test for this scenario (`useQuery` → "should refetch query when queryClient changes") passes only because it asserts `toBeDefined()` on the new client's cache entry — which is satisfied by the stranded pending entry. Tightening it to the exact cached value exposes the bug (see the repro branch below); this is the site I intentionally left loose in #11105.

https://stackblitz.com/github/reizam/query/tree/repro/solid-client-switch-observer/examples/solid/client-switch-repro

(Alternatively, the same branch tightens the existing unit test into a failing repro: [`useQuery.test.tsx#L5755-L5760`](https://github.com/reizam/query/blob/bad88ace3ad9a821964cfdb5a40594afd8306adf/packages/solid-query/src/__tests__/useQuery.test.tsx#L5755-L5760) — run `pnpm vitest run useQuery.test.tsx -t "should refetch query when queryClient changes"` in `packages/solid-query`.)

### Steps to reproduce

1. Open the StackBlitz example. The query fetches once through client 1 (both cache snapshots are rendered live).
2. Click "Switch to client 2".
3. Observe:
   - `queryFn` runs a second time, but client 1's cache receives the result (`dataUpdateCount: 2`),
   - client 2's cache entry stays `status=pending fetchStatus=idle data=undefined` indefinitely,
   - the `useQuery` result renders "result of fetch #2" — data owned by client 1 — while the active client is client 2.

### Expected behavior

After the `queryClient` accessor switches, the query should run against the new client: its cache entry should receive the data, and the hook should be subscribed to the new observer.

Every time

### Screenshots or Videos

State rendered by the repro after the switch:

```
client 1 cache: status=success fetchStatus=idle data="result of fetch #2" dataUpdateCount=2
client 2 cache: status=pending fetchStatus=idle data=undefined dataUpdateCount=0
```

### Platform

- OS: macOS
- Browser: Chrome (also reproduced in the vitest/jsdom suite)

### Tanstack Query adapter

solid-query

### TanStack Query version

v5.101.4

### TypeScript version

v5.9.3

### Additional context

Fix ready (the reorder described above) plus the tightened assertion as a regression test. Happy to send a PR.

rohanpatel2002:
## Changes
Fixes #11106.
When the `queryClient` accessor passed to `useQuery` / `useBaseQuery` switched, `createClientSubscriber()` ran before `setObserver(newObserver)`, so the subscription re-attached to the old observer. Refetches landed in the previous client's cache and the new client's entry stayed `pending`/`idle`.
- Reorder: `setObserver(newObserver)` then `createClientSubscriber()`
- Tighten `"should refetch query when queryClient changes"` to assert `queryClient2` cache `state.data === 'data'`
- Patch changeset for `@tanstack/solid-query`
## Checklist
- [x] Followed the Contributing guide
- [x] Tested locally (`vitest` for the client-switch case / solid-query suite)
- [x] Changeset included (published code)
## Release Impact
- [x] Affects published code — changeset included

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed query updates not being observed after switching the active query client in Solid Query.

### DIFF
--- a/.changeset/solid-client-switch-observer.md
+++ b/.changeset/solid-client-switch-observer.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/solid-query': patch
+---
+
+fix: subscribe to the new observer when the queryClient accessor switches


brenelz:
Only has a changeset

PR #11113, opened 2026-07-24
fix(solid-query): address suspense detachment in solid query

MaNaN1803:
Resolves an issue in Solid Query where suspense detachment causes queries to fail or hang.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved client-side subscription handling to prevent redundant refetches when results haven’t changed.
  * Updated behavior to reuse fresh results directly when not in loading/error states, while still refetching when appropriate.
  * Enhanced observer result update logic for more reliable and efficient data refreshes.


brenelz:
Closing as a duplicate

PR #11119, opened 2026-07-26
fix(query-core): keep retrying while unfocused when refetchIntervalInBackground is true

### LINKED ISSUE #8353: Polling Stops with refetchIntervalInBackground and Retry When Tab Is Inactive

makoto-developer:
Fixes #8353

### Problem

With `retry` and `refetchIntervalInBackground: true`, a failed background refetch does not retry while the tab is inactive — the retryer pauses until the page regains focus, which defeats the purpose of `refetchIntervalInBackground`.

### Cause

`Retryer.canContinue` gates continuation on `focusManager.isFocused()`:

```ts
const canContinue = () =>
  focusManager.isFocused() &&
  (config.networkMode === 'always' || onlineManager.isOnline()) &&
  config.canRun()
```

So while the tab is unfocused, a paused retry never continues.

### Fix

This follows the direction agreed by the maintainers in the earlier discussion on #9563 (checking the `refetchIntervalInBackground` prop additionally in the retryer, rather than removing the focus check):

- Thread `refetchIntervalInBackground` into the retryer config (`query.ts` → `retryer.ts`).
- `canContinue` now allows continuation while unfocused when `refetchIntervalInBackground === true`.

Default behavior (retries pause while the tab is unfocused) is unchanged when the option is unset.

### Testing

Added `should keep retrying while unfocused when refetchIntervalInBackground is true` to `query.test.tsx`. It fails on `main` (`expected undefined to be 'data3'`) and passes with this change. Full `@tanstack/query-core` suite: 555 tests pass, no type errors; prettier and eslint clean. Changeset included (patch).

---

- [x] I have followed the steps in the [contributing guide](https://tanstack.com/query/latest/docs/framework/react/community/contributing).
- [x] This change affects published code, and I have generated a changeset.

## Summary by CodeRabbit

* **Bug Fixes**
  * Failed query retries now continue while the tab is in the background when background refetching is enabled.
  * Existing focus and network conditions continue to be respected when background refetching is disabled.
  * Background refetching settings now consistently apply to retry behavior.

* **Tests**
  * Added coverage for retry behavior while the page is unfocused and hidden.


TkDodo (approved):

TkDodo:
we actually decided on not doing this:

- https://github.com/TanStack/query/pull/9563#discussion_r2313651946

Issue #5341, opened 2023-04-28
Mutations can not be started from an effect in Reacts StrictMode


jakst:
### Describe the bug

When firing a mutation from a useEffect on component creation the mutation get's stuck in loading state, despite the underlying request finishing. This only seems to happen in Reacts StrictMode.

If you naively do this you won't notice the bug, because this effect will fire twice in StrictMode and therefore "unstuck" the mutation when the request is sent the second time.
```typescript
useEffect(() => {
  if (isIdle) mutate()
}, [isIdle, mutate])
```
That's why the codesandbox example has a ref that keeps track on if the mutation has fired, to ensure it only runs once.

_The reason we're doing this in the first place is because we need to fire several mutations and keep track of the loading states individually, so we use components to create several mutation instances and fire the mutations on effect instead of in the original event handler._

### Your minimal, reproducible example

https://codesandbox.io/s/mutation-with-suspense-onsuccess-forked-lkmo6v?file=/src/Test.tsx

### Steps to reproduce

1. Go to attached codesandbox
2. Observe that the two loading states differ (reload if you have to)

### Expected behavior

I expect react-query's loading state to have the same value as the manually coded loading state

### How often does this bug happen?

Every time

### Platform

- OS: MacOS
- Browser: Chrome

### Tanstack Query adapter

react-query

### TanStack Query version

v4.28

### TypeScript version

v5.0.4

TkDodo:
the problem is that your `hasFired` ref workaround seems to screw things up. working sandbox: 
https://codesandbox.io/s/mutation-with-suspense-onsuccess-forked-2gekg2?file=/src/Test.tsx

If you don't want mutations to fire twice, either don't put them in a useEffect (they should likely run in event handlers anyways), or turn off strict mode

jakst:
> the problem is that your `hasFired` ref workaround seems to screw things up. working sandbox: 
> https://codesandbox.io/s/mutation-with-suspense-onsuccess-forked-2gekg2?file=/src/Test.tsx

Yeah but your modified example **only works** because it's firing twice. The first call gets stuck on loading, and then the second one unstucks it. Do you mean that you do not consider this a bug?

> If you don't want mutations to fire twice, either don't put them in a useEffect (they should likely run in event handlers anyways), or turn off strict mode

Yeah I would definitely prefer to run it in the event handler, but for our use case there is no way to do this with react-query AFAICT.

[...]

PR #11120, opened 2026-07-26
fix(query-core): reconnect MutationObserver on resubscribe (StrictMode fix)

leonardoag2026-beep:
Look, this is a pr related to this thread https://github.com/TanStack/query/issues/5341.

In the arguments, I saw "that mutations shouldn't have side-effects", "Strict mode tell us we shouldnt use mutate inside use effects", I do understand that "Only on dev happens, but on production it works", I do understand that "useMutationState solves the business" and I also get that "You don't want to waste your time having a workaround to trick StrictMode".

but hey, listen to me. In regards of you not wanting to waste your time, its fine, because this pr solves the business. In regard to using useMutationState, I argue that on a huge codebase, in my opinion it's pretty messy to call useMutationState, because first, isPending should just work regardless of whether you used useEffect or not. isPending should still work reliably on react, without this useMutationState, which in my opinion its only useful for tricking the StrictMode.

I know, that only on dev mode, this issue happens, but listen to me broyos, if i'm developing, i'm gonna test on develop mode, i'm not gonna build to prod everytime I need to make it work.

In regards to "mutations shouldn't have side effects", I argue that Tanstack Query Started as React Query, Tanstack is react's bitch, Bow to it (hey if you didn't laugh I meant no harm, just trying to be funny guy, you dont need to bow if you think the code gets messy).

Now about where the fix lives: onSubscribe() is on query-core, called by Subscribable.subscribe(). It fires for ALL frameworks (React, Solid, Vue, Svelte...). But the scenario that breaks (subscribe → unsubscribe → subscribe on the same tick) is exclusive to React's StrictMode. For every other framework:
- If #currentMutation is undefined → onSubscribe does nothing
- If #currentMutation exists → addObserver reconnects (harmless, just adds it back)

So the fix on query-core is safe for everyone. Moving it to react-query wouldn't work because MutationObserver lives on query-core and onSubscribe is a protected method called from inside the class.

But if you still think the change should live only on react-query, I can figure out a way. At least react query would be react's bitch and bow to it, right?

---

## Changes

Adds `onSubscribe()` to `MutationObserver` in `query-core`. When a new listener subscribes and there is an existing `#currentMutation`, the observer is reconnected to it via `addObserver(this)`, its result is refreshed via `#updateResult()`, and subscribers are notified via `#notify()`. This mirrors the existing pattern in `QueryObserver.onSubscribe()`.

## Checklist

- [x] Tests pass (58/58, including mutationObserver, mutations, mutationCache)
- [x] Appropriate type coverage — no new types added
- [x] No breaking changes — only adds a hook that's a no-op when `#currentMutation` is undefined

## Release Impact

- [x] **Patch** — bug fix, no breaking changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed mutation status handling when listeners unsubscribe and resubscribe during an in-progress mutation.
  * Mutation observers now reconnect correctly and report the eventual success state and result data instead of remaining stuck in a pending state.

* **Tests**
  * Added coverage for unsubscribe and resubscribe behavior during active mutations.


MILLERMARRU (commented):
This brings `MutationObserver` in line with a pattern `QueryObserver` already relies on, its own `onUnsubscribe` already calls `this.#currentMutation?.removeObserver(this)` when the last listener goes away, but there was no matching `onSubscribe` to reattach on the way back in. `QueryObserver.onSubscribe` does the equivalent `this.#currentQuery.addObserver(this)` on resubscribe, so this isn't new behavior for the codebase, it's `MutationObserver` catching up to something the query side has always done. That symmetry gap is exactly what StrictMode's mount, unmount, remount cycle exposes, if the same observer instance survives across that cycle but never reattaches to its `#currentMutation`, it's stuck showing whatever result was frozen at the moment it got orphaned.

One difference from `QueryObserver.onSubscribe` worth noting, that one only runs its reattach logic when `this.listeners.size === 1`, this one runs unconditionally on every `subscribe()` call since `Subscribable.subscribe()` calls `onSubscribe()` on every call, not just the first. Checked `Mutation.addObserver` and it already guards against duplicates itself (`if (!this.#observers.includes(observer))`), so a second subscribe on an already-attached observer can't double-register it. Given `useMutation()` observers aren't typically shared across multiple simultaneous consumers the way query observers are, this probably never matters in practice, but it's a real difference from the pattern this PR is otherwise mirroring, might be worth the same `size === 1` guard just for consistency with how the sibling class reads.

leonardoag2026-beep (commented):
That's right, we need a guard with listeners.size === 1. onSubscribe is called on every subscribe(), so it's not just the first time. Without this, an observer already connected would have to reconnect, which would be a waste of time. Pushed b849bce with the guard. Now MutationObserver behaves in this respect the same way as QueryObserver.

I also added regression testing to StrictMode. It unsubscribes while the mutation is still in progress, and then resubscribes to see if the observer finally succeeds. First, I tried to run it on the main branch, made sure it really failed, and after making the corrections, it passed. So it's not just a happy pass, it actually catches the bug.

TkDodo:
superseded by
- #11172

PR #11125, opened 2026-07-26
fix(eslint-plugin-query): use Object.hasOwn to avoid Object.prototype false positives in no-unstable-deps

### LINKED ISSUE #11118: eslint-plugin-query: no-unstable-deps incorrectly treats inherited Object.prototype properties as tracked custom hooks
### Describe the bug

`@tanstack/eslint-plugin-query`'s `no-unstable-deps` rule performs property lookups on `trackedCustomHooks` using:

```ts
trackedCustomHooks[callExpression.callee.name]
```

Since `trackedCustomHooks` is a plain object, inherited `Object.prototype` properties such as `toString`, `constructor`, and `valueOf` are returned instead of `undefined`.

As a result, unrelated functions with these names can be treated as tracked TanStack Query hooks, causing false positives from `no-unstable-deps`.

https://stackblitz.com/edit/vitejs-vite-mwhrnhuz?file=src%2FApp.tsx

### Steps to reproduce

1. Enable the `@tanstack/query/no-unstable-deps` rule.
2. Import a function whose name matches an inherited `Object.prototype` property, for example `toString` from Lodash.
3. Use it to derive a value and include that value in a React hook dependency array.

```ts
import { toString } from 'lodash'

const sessionId = toString(router.query.sessionId)

useCallback(() => {
  console.log(sessionId)
}, [sessionId])
```

4. Run ESLint.

## Expected behaviour

The rule should only treat explicitly tracked custom hooks as tracked hooks.

Inherited `Object.prototype` properties should not be considered, so the example above should not produce a `no-unstable-deps` warning.

Use an own-property check before reading from `trackedCustomHooks`, for example:

```ts
return Object.hasOwn(trackedCustomHooks, calleeName)
  ? trackedCustomHooks[calleeName]
  : undefined
```

Everytime

### Screenshots or Videos

_No response_

### Platform

ESLint: v10.x
@tanstack/eslint-plugin-query: 5.101.4

### Tanstack Query adapter

None

### TanStack Query version

5.101.4

### TypeScript version

_No response_

### Additional context

I've opened a PR with a proposed fix and regression tests:

**PR**: <https://github.com/TanStack/query/pull/11117>

mixelburg:
The trackedVariables record lookup in the no-unstable-deps rule used bracket access (`trackedVariables[name] !== undefined`) which matches inherited Object.prototype properties like toString, constructor, and valueOf. This caused false positives when those names appeared in dependency arrays alongside TanStack Query hooks.

Fixes #11118

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dependency validation in the ESLint rule, including identifiers that match built-in object property names such as `toString`.
  * Improved handling of `combine` properties in `useQueries`, including quoted property names and computed properties.
  * Updated dependency checks for more reliable query callback validation.
* **Tests**
  * Added coverage for local built-in-style identifiers and `combine` property edge cases.

### DIFF
--- a/packages/eslint-plugin-query/src/__tests__/no-unstable-deps.test.ts
+++ b/packages/eslint-plugin-query/src/__tests__/no-unstable-deps.test.ts
@@ -103,6 +103,22 @@ const baseTestCases = {
       ])
       .concat([
         {
+          name: `should pass when variable named toString (Object.prototype property) is used as dependency with ${reactHookAlias}`,
+          code: `
+            ${reactHookImport}
+            import { useQuery } from "@tanstack/react-query";
+
+            function Component() {
+              const { refetch } = useQuery({ queryFn: (value: string) => value });
+              const toString = () => refetch();
+              const callback = ${reactHookInvocation}(() => { toString() }, [toString]);
+              return;
+            }
+          `,
+        },
+      ])
+      .concat([
+        {
           name: `should pass when useQuery is imported from non-TanStack source and used with ${reactHookAlias}`,
           code: `
             ${reactHookImport}
--- a/packages/eslint-plugin-query/src/__tests__/no-unstable-deps.test.ts
+++ b/packages/eslint-plugin-query/src/__tests__/no-unstable-deps.test.ts
@@ -84,6 +84,26 @@ const baseTestCases = {
       ])
       .concat([
         {
+          name: `should pass when useQueries with quoted combine string key is passed to ${reactHookAlias} as dependency`,
+          code: `
+            ${reactHookImport}
+            import { useQueries } from "@tanstack/react-query";
+
+            function Component() {
+              const queries = useQueries({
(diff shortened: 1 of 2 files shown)


mixelburg:
Rebased on latest main ✅ — merge conflict resolved.

kprakhar:
Thanks for working on this! I also opened a PR addressing this issue earlier: https://github.com/TanStack/query/pull/11117

mixelburg:
Fixed the `combine` detection per CodeRabbit review:
- Added `!prop.computed` to avoid matching `{[combine]: fn}` computed keys as combine properties
- Added support for `Literal` key type so `{"combine": fn}` quoted string keys are recognized
- Previously only matched unquoted `Identifier` keys like `{combine: fn}`

The `hasCombineProperty` helper now correctly handles all three key forms.

PR #11127, opened 2026-07-27
fix(svelte-query): activate createQueries initialized with an empty array

### LINKED ISSUE #10681: [svelte-query] `createQueries` becomes permanently inactive when initialized with an empty `queries` array

renechoi:
## 🎯 Changes

Fixes #10681.

`createQueries` stayed permanently inactive when it was initialized with an empty `queries` array. Adding entries later never revived any slot, while the same component worked if the initial array had at least one entry.

The results object comes from `createRawRef`, which turns each key into a `$state.raw` field as that key appears. Two reads have nothing to subscribe to:

- an absent index (`results[0]` while the ref is still empty) has no field yet
- `length` is a plain array property on the proxy target

So a consumer that observes the result while it is empty registers no reactive dependency at all, and never re-runs once the queries are added. The data does arrive on the observer, which is why reading `results[0].data` directly returns the right value while the rendered output stays pending forever.

This adds a key version that `update()` bumps whenever the key set changes, and tracks it from the proxy `get` trap for absent keys, hidden keys, and array `length`. `createBaseQuery` passes an object whose shape is stable, so it never bumps the version and its behavior is unchanged.

For reference, #10696 targeted this issue by restructuring the `QueriesObserver` lifecycle and was closed by its author without review. The test below shows the observer itself is already doing the right thing on `main`, so the remaining defect is in how the results ref is tracked.

Heads up that the open #10892 adds a line to the same `update()` function for an unrelated defect (#10341). The two changes are independent, but they are adjacent enough to conflict textually if both land.

## ✅ Verification

Added a regression test that starts `createQueries` with `queries: []`, then adds one query and asserts that the tracked results reach `success`. It fails on `main` with `expected [] to match object [ { data: 'data1', status: 'success' } ]`, and reverting only `containers.svelte.ts` on this branch reproduces the same failure. In that failing state the preceding `expect(result[0]?.data).toBe('data1')` still passes, which is what isolates the problem to reactivity rather than fetching.

`nx affected --targets=test:sherif,test:knip,test:docs,test:eslint,test:lib,test:types,test:build,build` passes locally across the 13 affected projects. `@tanstack/svelte-query` is 23 files / 169 tests, and `@tanstack/svelte-query-devtools` and `@tanstack/svelte-query-persist-client` are green as well.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where queries initialized with an empty list could remain inactive after queries were added.
  * Improved reactive updates when object properties are added or removed, ensuring dependent views refresh correctly.

* **Tests**
  * Added coverage for dynamically adding queries after initialization and confirming successful results.


Michael-Obele (approved):
Nice, fix looks good to me.

PR #11128, opened 2026-07-28
fix(query-core): ignore a retained thenable callback invoked after settling

sobol-sudo:
`thenable.ts` is used by `hydration`, `retryer` and `queryObserver`, but it has no test file of its own, so its behaviour is only covered indirectly. This adds a unit suite for both exports, and fixes a defect the suite surfaced.

### Fix

`pendingThenable` deletes `resolve`/`reject` on settle, with the comment that this avoids calling them twice. A caller that kept a reference to one of them beforehand bypasses that guard, and `finalize` then applies to an already settled promise:

```ts
const thenable = pendingThenable<string>()
const retainedReject = thenable.reject

thenable.resolve('data')
retainedReject(new Error('late'))

// before: status 'rejected', value 'data', reason Error: late — while the promise resolves to 'data'
// after:  status 'fulfilled', value 'data', no reason
```

The underlying promise correctly ignores the late call, so only the advertised state was wrong. That matters because React reads `.status`/`.value`/`.reason` to unwrap promises optimistically, so a thenable claiming `rejected` while actually fulfilled can surface an error that never happened. `finalize` now returns early once the thenable is no longer pending.

### Tests

**`pendingThenable`** — starts out `pending` with `resolve`/`reject` attached; exposes `value`/`reason` and drops the pending props once settled; ignores a retained callback invoked after settling; does not surface an unhandled rejection when nobody awaits it.

**`tryResolveSync`** — returns `undefined` for a native promise, including an already resolved one, since `then` always defers to a microtask; returns the data for a thenable that calls back synchronously; supports a synchronous thenable whose `then()` returns something without a `catch` method, which is the case the `?.catch` guard exists for; returns `undefined` when a synchronous thenable resolves with `undefined`; returns `undefined` for a rejected promise without leaving the rejection unhandled.

The retained-callback test fails on `main` with `expected 'rejected' to be 'fulfilled'` and passes with the change. Locally `query-core` is 565 passed / 25 files, no type errors, ESLint clean.

Patch changeset included since this changes published behaviour.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed thenable behavior so retained settlement callbacks can’t overwrite state after settlement.

* **Tests**
  * Added coverage for pending and settled states, settlement data, repeated settlement prevention, synchronous thenable resolution, rejected and undefined outcomes, and suppression of unhandled rejection events.


MILLERMARRU (commented):
Checked React's own `use()` reference docs on the promise-cache pattern this file is implementing (the file's top comment already links straight to React's internal types for it), and it confirms exactly why this matters beyond just internal consistency. React's documented pattern sets `promise.status`/`promise.value`/`promise.reason` once a promise settles so `use()` can read them synchronously on a later render without re-suspending. That means `status` disagreeing with what actually settled isn't just a cosmetic inconsistency, it's the exact property `use()` (or anything else following this convention) trusts to decide whether to return a value, throw, or suspend again.

The actual bug makes sense once you see that deleting `resolve`/`reject` off the object only stops *future* calls to `thenable.resolve(...)`/`thenable.reject(...)`, a caller that already grabbed a reference to the function before it got deleted still holds a valid closure and can call it later regardless. The underlying native `Promise`'s own resolve/reject silently no-ops on a second call per spec, so that half was always safe, the actual gap was `finalize` running unconditionally and overwriting the custom `status`/`value`/`reason` properties a second time even after the object had already settled and possibly already been read by a consumer.

Putting the pending-status check at the very top of `finalize`, before touching anything, is the right place for it, since it makes the whole function a no-op for a late-arriving retained callback rather than trying to guard each individual assignment.

The test that actually exercises this (`should ignore a retained settlement callback invoked after it settled`) grabs `reject` before resolving, which is exactly the shape of the real bug, holding a reference across the settlement boundary, not just calling reject twice through the current property.

TkDodo (approved):

PR #11130, opened 2026-07-29
fix(react-query): keep unsubscribed useQueries idle

ShiroKSH:
## Summary

- Keep `useQueries` results aligned with the cache when `subscribed: false`.
- Add a regression test for the initial unsubscribed render.
- Add a patch changeset for `@tanstack/react-query`.

## Root cause

`useQueries` stopped subscriptions for `subscribed: false`, but still assigned `_optimisticResults: 'optimistic'` to every child observer. An unmounted observer then synthesized `isFetching: true` and `fetchStatus: 'fetching'` even though it did not run the query function and the cache remained idle.

This completes the `useQueries` portion of the fix proposed in #10739. PR #10759 corrected `useBaseQuery`, but did not change `useQueries`.

## Validation

- `vitest src/__tests__/useQueries.test.tsx` — 23 passing tests, including the new regression.
- `test:types:tscurrent`
- targeted ESLint and Prettier checks
- Confirmed red-to-green: the new test fails on the pre-fix implementation with `isFetching: true` / `fetchStatus: fetching`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `useQueries` so unsubscribed queries no longer display optimistic fetching states.
  - Unsubscribed queries now correctly remain idle and avoid initiating query functions.

- **Tests**
  - Added coverage verifying fetching status, query execution, and observer behavior for unsubscribed queries.

- **Release**
  - Included a patch release update for `@tanstack/react-query`.

### DIFF
--- a/.changeset/tidy-pandas-wait.md
+++ b/.changeset/tidy-pandas-wait.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/react-query': patch
+---
+
+fix(react-query): don't show optimistic fetching for unsubscribed useQueries
--- a/packages/react-query/src/__tests__/useQueries.test.tsx
+++ b/packages/react-query/src/__tests__/useQueries.test.tsx
@@ -63,6 +63,34 @@ describe('useQueries', () => {
     expect(results[2]).toMatchObject([{ data: 1 }, { data: 2 }])
   })
 
+  it('should not optimistically show fetching when unsubscribed', () => {
+    const key = queryKey()
+    const queryFn = vi.fn(() => Promise.resolve('data'))
+
+    function Page() {
+      const [query] = useQueries({
+        queries: [{ queryKey: key, queryFn }],
+        subscribed: false,
+      })
+
+      return (
+        <div>
+          <span>isFetching: {String(query.isFetching)}</span>
+          <span>fetchStatus: {query.fetchStatus}</span>
+        </div>
+      )
+    }
+
+    const rendered = renderWithClient(queryClient, <Page />)
+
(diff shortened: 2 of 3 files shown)


MILLERMARRU (commented):
Checked how `_optimisticResults` actually gets consumed in `queryObserver.ts` to confirm setting it to `undefined` is the right lever here, not just a workaround. The whole "optimistically set result in fetching state" block in `getOptimisticResult` is gated by `if (options._optimisticResults)`, so making it `undefined` when `subscribed` is `false` skips that block entirely rather than trying to override its output afterward, the query's reported state stays whatever it genuinely is instead of getting bumped into a fabricated fetching appearance for something that was never going to fetch in the first place, since `shouldSubscribe` already gates the actual observer subscription on the same `subscribed` flag.

While I was in there I checked whether this same gap exists in the sibling adapters, since `_optimisticResults` and the `subscribed` option show up in basically every framework package here. `preact-query/src/useQueries.ts` still has the unconditional version, `defaultedOptions._optimisticResults = isRestoring ? 'isRestoring' : 'optimistic'`, no `subscribed` check at all, same exact bug this PR fixes for react-query. Not something to fix in this PR, just flagging since preact-query tends to mirror react-query closely and this looks like it'd need the identical treatment.

Test is well targeted, it asserts `queryFn` was never called, the query has zero observers, and both `isFetching`/`fetchStatus` read as idle, which covers the actual bug (a misleading loading UI for a query that was never subscribed) rather than just checking one symptom of it.

Hardanish-Singh (approved):

TkDodo (approved):

PR #11150, opened 2026-08-07
perf(devtools): stop the panel freezing on caches with thousands of queries

### LINKED ISSUE #11149: Devtools panel freezes the page for minutes when the query cache holds thousands of observed queries
### Describe the bug

Opening the devtools against a `QueryClient` holding a few thousand queries blocks the main thread for minutes. The tab stops responding entirely — no scrolling, no clicks, no animation — and stays that way until the panel finishes mounting.

The panel renders one row per cached query and per cached mutation with no windowing, and each row registers several cache subscriptions, so the work done on open scales with the size of the cache rather than with what is visible. Several subscriptions also walk the whole cache on every cache event: the details panes resolve their selected item by scanning until they find a match, and the status badges count the entire cache once per status. Nothing throttles that, and each `setQueryData` emits two cache events, so an application that is still loading data keeps the panel re-doing all of it.

https://codesandbox.io/p/sandbox/brave-microservice-8kltf4

### Steps to reproduce

1. Open the example and let the list build for a few seconds.
2. Press **Open the devtools**.
3. Try to scroll the page. Nothing responds — the spinner stops and scrolling and clicks do nothing. The page stays frozen for minutes.
4. When it recovers, the readout under the spinner reports the longest unbroken freeze.

### Expected behavior

Opening the devtools should not block the main thread for a noticeable period, and keeping it open should not make the application unresponsive.

Every time

### Screenshots or Videos

_No response_

### Platform

OS: Windows 11
Browser: Edge
Version: 151.0.4129.59

### Tanstack Query adapter

react-query

### TanStack Query version

v5.101.4

### TypeScript version

v5.9.2

### Additional context

_No response_

cloudfluffy:
## 🎯 Changes

Closes #11149.

Opening the devtools against a `QueryClient` holding thousands of queries freezes the page, and it stays frozen while the cache keeps changing. The panel renders one row per cached query and per cached mutation with no windowing, and several of its subscriptions walk the whole cache on every cache event, so both the mount cost and the per-event cost grow with the size of the cache.

There were four separate causes.

---

**Every query got its own row, all at once.** The panel mounted one row per cached query and per cached mutation, and each row subscribed to the cache to keep itself current — thousands of DOM nodes, and thousands of subscriptions for the cache to notify on every change.

Both lists are now virtualized. `VirtualList` (`Devtools.tsx:680`) mounts only the rows inside the scroll viewport, plus a small overscan. This needs a known row height, so rows are now a fixed height derived from `QUERY_ROW_HEIGHT_MULTIPLIER` (`constants.ts:20`), which drives both the row CSS and the window arithmetic.

Selecting a query and then scrolling its row out of view would unmount the row and tear down the details pane's subscription, so `pinnedKey` (`Devtools.tsx:684`) keeps the selected row mounted.

---

**Parts of the panel searched the whole cache to find one query.** `QueryDetails` (`Devtools.tsx:1981`) and `MutationDetails` (`2509`) resolved their selected item by scanning the cache until they found a match, and the status badges — `QueryStatusCount` (`1781`) and `MutationStatusCount` (`1826`) — counted the entire cache five separate times, once per status. Every cache event re-ran all of that.

The details panes now look their item up by key, and the badges tally all five statuses in a single pass.

`MutationRow` (`1677`) had a variant of the same problem: it already receives its mutation as a prop, but three of its subscriptions scanned the whole cache to find it by `mutationId` on every mutation-cache event. They now read `props.mutation.state` directly.

---

**The panel recalculated on every change, even when changes arrived in bursts.** Nothing throttled the two subscribers that read the entire cache, and a single `setQueryData` fires two cache events, so every update in a burst cost two full passes.

Those two now coalesce (`setupQueryCacheSubscription`, `Devtools.tsx:2787`): a subscriber that has been idle runs immediately, and anything arriving while a window is open folds into one trailing pass. The window is `max(COALESCE_WINDOW_MS, lastPassMs * COALESCE_PASS_MULTIPLE)` — widening it with the cost of the last pass bounds the share of the main thread the panel can take as the cache grows.

Two smaller changes ride along: the query list keeps its array identity when its contents haven't changed (`sameAsPrevious`, `871-872`), and `setupMutationCacheSubscription` (`2890`) now batches its fan-out the way the query side already did.

There is also a fix here that isn't about performance. The subscriber registries are shared by every panel on the page, but a panel's teardown cleared them entirely instead of removing its own entries. With two panels mounted — during the picture-in-picture transition, or the standalone panel alongside the floating one — tearing one down froze the other.

---

**The panel's stylesheet was rebuilt for every component.** Around sixty CSS-in-JS templates were recompiled from scratch each time a row mounted or a node in the JSON explorer expanded.

`createStylesCache` (`utils.tsx:345`) now compiles a stylesheet once per theme and shares it, and `cssForTarget` (`utils.tsx:332`) gives it a stable cache key by binding `goober`'s `css` once per shadow root instead of once per component.

---

### Behaviour change

A single change to a quiet cache still updates the panel immediately, exactly as before. The difference only shows during a stream of changes: the query list and the status badges then update once per coalescing window instead of once per event.

## ✅ Checklist

- [x] I have followed the steps in the [[Contributing guide](https://github.com/TanStack/query/blob/main/CONTRIBUTING.md)](https://github.com/TanStack/query/blob/main/CONTRIBUTING.md).

## 🚀 Release Impact

- [x] This change affects published code, and I have generated a [[changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md)](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).

## Summary by CodeRabbit

* **Performance**
  * Improved Devtools responsiveness when browsing large query and mutation caches.
  * Added virtualized lists to reduce rendering overhead.
  * Optimized cache updates, filtering, status tracking, and panel lifecycle handling.

* **Usability**
  * Added query-row truncation with full-key tooltips.
  * Improved scrolling, selection retention, resizing, and light/dark theme styling.
  * Preserved active cache updates when multiple Devtools panels are open.


packages/query-devtools/src/Devtools.tsx
```diff
@@ -3372,14 +3640,16 @@ const stylesFactory = (
...
         font-family:
           ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas,
           'Liberation Mono', 'Courier New', monospace;
         border-bottom: 1px solid ${t(colors.gray[300], colors.darkGray[400])};
         text-align: left;
-        text-overflow: clip;
-        word-break: break-word;
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
```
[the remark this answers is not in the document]
cloudfluffy: Good catch, thanks — fixed in 24b3631.

Rather than adding an inner element, `.tsqd-query-hash` is now a block container with a line height equal to its `min-height`, which keeps the text vertically centred without adding a node per row:

```diff
- display: flex;
- align-items: center;
+ display: block;
+ line-height: ${tokens.size[6]};
```

Checked in Chromium against the built panel — the same row renders `..."region":"west-eu` before and `..."region":"west-…` after, with identical row geometry.

cloudfluffy:
@TkDodo sorry to ping directly, but this one needs a maintainer to get moving.

This is my first PR to the repo, so the workflows are still waiting on approval before CI can run.

It is a large diff, so happy to break it into smaller PRs if that would make it easier to review.

Issue: #11149

packages/query-devtools/src/Devtools.tsx
```diff
@@ -671,14 +667,154 @@ const DraggablePanel: Component<DevtoolsPanelProps> = (props) => {
...
+ * Windowed list renderer for the query and mutation panes.
+ *
+ * Only the rows intersecting the scroll viewport (plus `OVERSCAN` on each side)
+ * are mounted, so the DOM node count and the per-row cache subscriptions stay
+ * bounded regardless of how many entries the cache holds. Rows have a fixed
+ * `rowHeight`, so offsets are computed arithmetically without measuring the DOM.
+ * The currently selected row is always kept mounted so its subscriptions and
+ * focus survive scrolling and live re-sorts.
+ */
+function VirtualList<T>(props: {
```
blindbeat, line 680: Pr seem to handroll virtualization, while Tanstack packages already have [in-house virtualization library ](https://tanstack.com/virtual/latest). It may be a good idea to utilize it

packages/query-devtools/src/Devtools.tsx
```diff
@@ -805,6 +963,30 @@ export const ContentView: Component<ContentViewProps> = (props) => {
...
+      }
+    }
+    readRowFontSize()
+    const view = containerRef.ownerDocument.defaultView
+    view?.addEventListener('focus', readRowFontSize)
+    onCleanup(() => view?.removeEventListener('focus', readRowFontSize))
+  })
+  const rowHeight = createMemo(
+    () => rowFontSize() * QUERY_ROW_HEIGHT_MULTIPLIER,
+  )
```
blindbeat, line 989: Possibly, this part may also be owned by Tanstack Virtual: https://tanstack.com/virtual/latest/docs/pretext

blindbeat (changes requested):
HI @cloudfluffy!

I had a look on PR and though of a couple of possible simplifications. Do you think it can work? 

Also, PR seem to do multiple improvements to performance, which is awesome, but also it may be better to separate them into multiple separate PRs, so they can be tested and verified in isolation. As now it's harder to build the full picture of the changes. I would be happy to have a look at them and review!

TkDodo:
I agree. we should be using tanstack/virtual for virtualization, and smaller, more focussed PRs have a better chance of getting merged in general.

PR #11161, opened 2026-08-09
fix(query-core): clear stale select error when observer switches to a query without data

### LINKED ISSUE #11160: Stale `select` error from a previous query leaks into the next query's pending result
## Describe the bug

When a `select` function throws, the error is stored on the observer (`#selectError`) and surfaced on every result. It is only ever cleared inside the select block itself — which is skipped entirely while `data` is `undefined` (`queryObserver.ts`, `createResult`). So after switching the observer to a different `queryKey` (or after `resetQueries`), the new pending query reports `status: 'error'` carrying the *previous* query's select error. With `throwOnError` or suspense, the stale error is thrown into the error boundary while the new query is still fetching.

## Minimal reproducible example

```ts
  queryKey: ['user', 'a'],
  queryFn: fetchUser,
  select: (d) => d.profile.name, // throws for user "a"
})
observer.subscribe(() => {})
// select throws -> status 'error' (correct)

observer.setOptions({
  queryKey: ['user', 'b'], // different, healthy query
  queryFn: fetchUser,
  select: (d) => d.profile.name,
})
observer.getCurrentResult().status // 'error' with user "a"'s error — expected 'pending'
```

Equivalent in React: `useQuery({ queryKey: ['user', id], ... })` where `select` throws for id "a", then setting id to "b".

Second repro without a key change: after the select error, call `queryClient.resetQueries({ queryKey })` — the reset pending state still carries the old select error.

## Expected behavior

While the new (or reset) query is fetching and has no data, the result should be `{ status: 'pending', error: null }`. A select error belongs to the data it was computed from; once that data is gone, it should not leak into the next result.

## Platform

- `@tanstack/query-core` (all adapters affected), reproduced on `main`

## Additional context

The cause is that `createResult` only clears `#selectError` on a successful select run inside the `options.select && data !== undefined` block; the `if (this.#selectError)` check below it then fires unconditionally. Related but distinct from #11011 (which fixes `isPlaceholderData` when select throws on placeholder data).

I have a fix with regression tests ready and will open a PR.

hamed-bavar:
Fixes #11160

## Problem

When a `select` function throws, the error is stored on the observer (`#selectError`) and surfaced on every subsequent result. It is only cleared inside the select block itself, which is skipped entirely while `data` is `undefined`. Switching the observer to a different `queryKey`, or resetting the query, therefore leaves the old error in place: the new pending query reports `status: 'error'` with the previous query's select error, and with `throwOnError` / suspense the stale error is thrown into the error boundary while the new query is still fetching.

## Solution

In `createResult`, clear the stored select error whenever a result is built without data to select:

```ts
} else if (data === undefined) {
  // a stored select error belongs to previously selected data; once that
  // data is gone (query switch or reset), it must not leak into this result
  this.#selectError = null
}
```

A select error can only be produced by running `select` on some data — if the result being built has no data (fresh query after a key switch, or a reset query), that data is gone and the error cannot describe the current result.

The "keep the error while the same data and selector are in place" memoization is untouched: in that case the query's own `data` is defined (a select error implies the queryFn succeeded), so that path still goes through the select block as before. If the selector genuinely throws again for new data, the error is re-created — only stale errors are dropped.

Clearing is done at result-creation time rather than on query switch (`#updateQuery`) because `resetQueries` replaces the state of the *same* Query instance — a switch-based clear would miss it.

## Tests

Two regression tests in `queryObserver.test.tsx`, written first and failing on `main` with the exact buggy behavior (`status: 'error'` where `'pending'` is expected):

- select throws on query A → `setOptions` to key B → result is `pending` with `error: null`, then succeeds
- select throws → `resetQueries` → same expectation

Full `query-core` (509) and `react-query` (417) suites pass.

Related but distinct from #11011, which handles `isPlaceholderData` when select throws on placeholder data; the two changes touch different branches and compose.

## Summary by CodeRabbit

- **Bug Fixes**
  - Cleared stale selection errors when switching to a query without data.
  - Cleared previous selection errors when resetting a query before a successful refetch.
  - Prevented errors from earlier query results from appearing in subsequent query states.
  - Correctly reset placeholder-data status when transitioning between query states.

- **Tests**
  - Added regression coverage for query switching, placeholder data, memoized results, and reset scenarios.

## Update

After the review discussion below, this PR also resets `isPlaceholderData` in the select-error block. That is the same one-line fix as #11011 by @chatman-media (credit to that PR); folding it in aligns the runtime with the error-result types in `types.ts` (which declare `isPlaceholderData: false`) and closes the memoized-placeholderData leak found in review, so this PR no longer depends on #11011's merge order. A regression test covers the full leak chain.


MILLERMARRU (commented):
Traced `#selectError` through `createResult` to confirm this. It's an instance field that only gets touched inside the `options.select && data !== undefined && !skipSelect` block, so when the observer switches to a query whose current data is `undefined` (a fresh key, or right after a reset), that block is skipped entirely and whatever `#selectError` was left over from the previous query keeps sitting there. Then the unconditional `if (this.#selectError)` a few lines down turns that stale error into `status: 'error'` on a query that hasn't even started fetching yet. The fix closes exactly that hole by clearing it in the one case the original condition didn't cover.

One combination I don't see a test for: `skipSelect` gets set to `true` in the placeholderData memoization path (when `prevResult.isPlaceholderData` and the same `placeholderData` option is being reused), and in that path `data` ends up defined again (it's the placeholder), so neither branch of the `if/else if` runs. If a query switch lands on that exact memoized-placeholder path while a `#selectError` from the prior query is still set, I think it could still leak through. Narrow edge case, probably not worth blocking on, but might be worth a third branch or a comment confirming it's out of scope.

hamed-bavar:
Good catch. I spent some time trying to reproduce this, and you're right that the memoized-placeholder path is the one exit my `if`/`else if` doesn't cover: `skipSelect` is true, `data` holds the reused placeholder, and a leftover `#selectError` would go straight into the `if (this.#selectError)` block.

But I couldn't reach that state without going through the bug #11011 fixes. The memoization only runs when `prevResult.isPlaceholderData` is true. Any result built while `#selectError` is set is an error result, and error results only keep `isPlaceholderData: true` because of the bug #11011 removes. `#selectError` is never touched outside `createResult`, so if the previous result wasn't an error, the stored error was already null.

The one repro I found on current `main`:

1. selector succeeds on query A, so `#selectResult` is populated
2. switch to query B with the same `placeholderData` reference; the selector throws on B's placeholder, and that error result wrongly keeps `isPlaceholderData: true`
3. switch to query C: the memoization sees `prevResult.isPlaceholderData`, reuses A's stale result, and B's error lands on C

With #11011 applied step 3 is unreachable, so a clearing branch here turns into dead code the moment that merges. My preference is to leave this branch alone and let #11011 close the hole

MILLERMARRU:
Traced this against current `main` line by line and your chain is right, `isPlaceholderData` only ever gets set at the one spot in the placeholder block, never reset in the `if (this.#selectError)` block, and `#selectError` is only ever written inside `createResult`'s own select try/catch. So yes, the only way to reach `prevResult.isPlaceholderData === true` on a result that's actually an error is exactly the state #11011 fixes. Also checked `types.ts`: both `QueryObserverLoadingErrorResult` and `QueryObserverRefetchErrorResult` hard-code `isPlaceholderData: false`, so what's happening on current `main` isn't just an edge case, it's the runtime violating a combination the type system itself declares impossible. That's a stronger argument for #11011 landing than just this thread's repro.

Where I'd push back a little: "leave this branch alone and let #11011 close the hole" assumes #11011 actually lands, and it's been open since June 30 with zero reviews. This PR is moving, that one isn't. If this merges and releases before #11011 does (which looks likely at this rate), the gap you found is a known, filed, reachable bug shipping in a release, on purpose, because it's waiting on a dependency that has no timeline.

Given that, I'd just fold #11011's one-line fix into this PR instead of leaving the two coupled:

```ts
if (this.#selectError) {
  error = this.#selectError
  data = this.#selectResult
  errorUpdatedAt = Date.now()
  status = 'error'
  isPlaceholderData = false
}
```

That's not dead code the way a guard on the memoization branch would be, it's the actual fix for the type-contract violation, and it happens to also close the memoization hole as a side effect, since `prevResult.isPlaceholderData` can never observe `true` on an error result once this lands. #11161 stops depending on #11011 merging at all, either order works, and if #11011 does land later the same line collides in a rebase, which is a one-second fix, not a real cost. Feels like the more complete version of what this PR is already doing rather than a separate concern.

hamed-bavar:
Fair enough. The types.ts angle settles it for me: if the error result variants declare `isPlaceholderData: false`, the runtime shouldn't be able to produce anything else, and waiting on a PR that's had no reviews since June to restore that wasn't a great bet.

I folded the line into the error block here, plus a regression test for the exact chain from this thread (stable `placeholderData` reference; selector succeeds on A, throws on B, works on C; before the change the memoization handed A's stale result to C along with B's error).

Credit where due: that line is #11011 by @chatman-media, which also comes with its own test for the placeholder case. If that lands first I'll rebase; the collision is trivial in either order.

TkDodo:
https://github.com/TanStack/query/pull/11011 was just merged

TkDodo (approved):

PR #11163, opened 2026-08-09
fix(query-core): release the retryer once a fetch settles

codebytere-ant:
## 🎯 Changes

`Query.fetch()` leaves `#retryer` set after the fetch settles. The retryer's resolved promise still holds the raw result of that fetch, so once a query has refetched it holds its data twice: the structurally shared objects in `state.data`, plus the fresh result the promise resolved with. It also means `reset()` and `removeQueries()` don't free the data while anything still points at the query.

This clears `#retryer` in the existing `finally` once the fetch has settled. The identity check is there so a fetch started synchronously from a cache `onSuccess`/`onSettled` callback keeps its own retryer. The other readers of `#retryer` are already behind `?.` or a `fetchStatus` check, so the only visible change is that `query.promise` is `undefined` after a fetch has settled instead of the settled promise.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Released query retry resources after fetches settle, queries reset, or queries are removed.
  * Prevented settled promises from retaining unnecessary fetch results.
  * Ensured callback-triggered refetches complete correctly after an initial fetch.

* **Release**
  * Included in a patch release of the query core package.

### DIFF
--- a/.changeset/release-settled-retryer.md
+++ b/.changeset/release-settled-retryer.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/query-core': patch
+---
+
+Release a query's retryer once its fetch settles, so the settled promise no longer keeps that fetch's raw result in memory alongside the structurally shared `state.data` (or after the query is reset or removed).
--- a/packages/query-core/src/__tests__/query.test.tsx
+++ b/packages/query-core/src/__tests__/query.test.tsx
@@ -7,6 +7,7 @@ import {
 import {
   CancelledError,
   Query,
+  QueryCache,
   QueryClient,
   QueryObserver,
   dehydrate,
@@ -14,12 +15,7 @@ import {
 } from '..'
 import { hashQueryKeyByOptions } from '../utils'
 import { mockOnlineManagerIsOnline, setIsServer } from './utils'
-import type {
-  QueryCache,
-  QueryFunctionContext,
-  QueryKey,
-  QueryObserverResult,
-} from '..'
+import type { QueryFunctionContext, QueryKey, QueryObserverResult } from '..'
 
 describe('query', () => {
   let queryClient: QueryClient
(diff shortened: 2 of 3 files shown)


MILLERMARRU (commented):
Went through every remaining read of `#retryer` on current `main` to check the "already behind `?.` or a `fetchStatus` check" claim, and it holds. The one that took real tracing was the `fetch()` entry-point block gated by `this.state.fetchStatus !== 'idle'`, since that's a separate piece of state from `#retryer` itself, updated via `#dispatch` rather than alongside the field. Once a fetch settles, `fetchStatus` is already back to `idle` by the time a new `fetch()` call runs, so the whole `this.#retryer?.status()`/`else if (this.#retryer)` block is skipped outright, it never gets a chance to observe `#retryer` as `undefined` in a way that would matter.

The one case I wasn't sure about from reading alone: the silent-cancellation catch branch (`return this.#retryer.promise`, deliberately still reading the field rather than the local `retryer` const, since it's meant to piggyback on whatever *new* fetch triggered the cancel) reads `this.#retryer` without an `?.`. If the new fetch's retryer could settle and get cleared before the old fetch's catch block runs, that'd throw instead of returning a promise. Pulled `retryer.ts` and `thenable.ts` as-is and built a standalone repro of the real `Query.fetch()` control flow around them to check the microtask ordering rather than reason it out by hand:

```
applyFix=false [{"status":"fulfilled","value":"fresh-data"},{"status":"fulfilled","value":"fresh-data"}]
applyFix=true [{"status":"fulfilled","value":"fresh-data"},{"status":"fulfilled","value":"fresh-data"}]
```

No difference, and it's structurally guaranteed rather than a fluke: `cancel()` synchronously rejects the old thenable, which schedules the old fetch's `await` continuation as a microtask immediately, before the new retryer's `run()` even gets called. Clearing `#retryer` only happens once the new fetch's *own* `await retryer.start()` resumes and its `finally` runs, which needs its own settle-then-await-resume hop after that, so the old fetch's catch always gets there first.

Nice fix, the identity check in the new `finally` is the right guard for the synchronous-refetch-from-onSuccess case.

TkDodo (approved):

PR #11174, opened 2026-08-12
chore: merge main into solid-query-v6-pre, upgrade to solid v2 rc

brenelz:
Brings `main` into the v6 prerelease branch and moves the branch onto the Solid 2.0 RC.

Two commits: the Solid RC upgrade, then the `main` merge.

> Note: this includes the commit from #11173. If this lands, that PR can be closed.

## Solid RC, with carets

`solid-js`, `@solidjs/web`, `@solidjs/signals` and `babel-preset-solid` go from the pinned `2.0.0-beta.33` to `^2.0.0-rc.0` across the three solid packages, the `solid-vite` integration and the solid examples — also flipping these from exact pins to carets so the rc line is tracked as it moves. `packages/query-devtools` is untouched; it is still on Solid 1.x.

`@tanstack/solid-query`'s peer floor moves to `>=2.0.0-rc.0 <3.0.0`. The deliberately broader `>=2.0.0-beta.0 <3.0.0` floors on the other two are left alone, matching how previous bumps handled them.

## `vite-plugin-solid` → `@solidjs/vite-plugin`

Renamed at `^3.0.0-next.27` — the version published under both names, so a clean cut-over. Covers every manifest, every vite config and the hydration test fixture. `packages/query-devtools` stays on `vite-plugin-solid@^2.11.6`; the scoped package has no 2.x line. The `-i vite-plugin-solid` sherif ignore is dropped, since the dependency no longer resolves to more than one version.

## Prerelease tag → `rc`, landing on `6.0.0-rc.0`

`.changeset/pre.json` moves to `"tag": "rc"`. No workflow change needed — `solid-query-v6-pre` still matches the `*-pre` glob in `release.yml`, which sets `prerelease=true` and passes no `--tag`, so `changeset publish` takes the dist-tag from pre.json.

Changesets derives the pre number as `(prerelease[1] ?? -1) + 1`, so switching the tag alone would have continued the beta counter into `6.0.0-rc.9`. To land on `6.0.0-rc.0`, the solid packages carry the sentinel version `6.0.0-rc` — valid semver with no numeric prerelease part, so the pre number resolves to `0` while `semver.inc("6.0.0-rc", "patch")` still yields `6.0.0`. That sentinel only lives here until the next `ci: Version Packages` commit rewrites it; nothing publishes from it, since the release action runs `version` before `publish`.

Verified by running the real `changeset version` against a throwaway copy: all three solid packages land on `6.0.0-rc.0`, no other package is touched, and the changelog entry lands directly above `## 6.0.0-beta.8`.

## Merge conflict resolutions

All conflicts were in the solid packages, examples and workspace config:

- **`examples/solid/*/package.json`** — kept the v6 / Solid 2 side, but took main's newer `^5.101.4` for the 5.x packages that are not part of the v6 line (`query-async-storage-persister`, `eslint-plugin-query`)
- **`packages/solid-query*/package.json`** — kept the `6.0.0-rc` sentinel over main's `5.101.4`
- **`packages/solid-query*/CHANGELOG.md`** — kept both histories, `6.0.0-*` above main's new `5.101.1`–`5.101.4` entries, leaving them in descending version order ahead of the shared `5.101.0` tail
- **`solid-query/useMutation.test.tsx`** — kept our rewrite of the retry test. The surrounding assertions require `mutationKey` and `data2`, which only our `mutationFn` produces; main's variant is the Solid 1 version of the test that this branch had already replaced
- **`solid-query-devtools/devtools.test.tsx`** — kept main's five new prop-change forwarding tests from #11021, ported to this branch's Solid 2 pattern by flushing after each signal write, and kept our `async` declaration of the unmount test, whose body awaits. Also merged the two `solid-js` import statements the merge left duplicated
- **`solid-query-devtools/devtoolsPanel.test.tsx`** — kept our `flush()` and took main's stronger assertion from #11030 that the default `onClose` is a no-op
- **`pnpm-workspace.yaml`** — took main's expanded `trustPolicyExclude` and kept this branch's `minimumReleaseAgeExclude` block
- **`pnpm-lock.yaml`** — regenerated from the merged manifests

## Known peer warning

`pnpm install` reports an unmet peer for `@solidjs/vite-plugin@3.0.0-next.27`, which declares `solid-js` / `@solidjs/web` as `>=2.0.0-beta.32 <2.0.0-experimental.0`. Prerelease identifiers sort alphabetically, so the real ordering is `beta.32 < experimental.0 < rc.0` — the upper bound excludes the RC by accident. Metadata only, and warning only (no `strict-peer-dependencies` here, and pnpm defaults it to `false`). It should clear when Solid ships a plugin release with a corrected bound.

## Verification

On the merged tree, for all three solid packages:

- `test:lib` — 369 tests pass, no type errors (devtools goes 28 → 33 with main's ported tests)
- `test:types`, `test:build`, `test:eslint` — pass
- `sherif` and `knip` — clean
- `integrations/solid-vite` and `examples/solid/basic` still build

No source changes were needed for the RC itself.


brenelz:
@birkskyum we could probably merge this even if we dont do a release

PR #11181, opened 2026-08-13
fix(scripts): make the docs link check work on Windows

### LINKED ISSUE #11177: pnpm test:docs reports 27 broken links on Windows because verify-links.ts assumes forward slashes
### Describe the bug

`pnpm test:docs` fails on Windows with 27 broken links that are all fine. Every one of them is a documentation link pointing at an example.

`scripts/verify-links.ts` builds the target path with `path.resolve`, which uses the platform separator, and then matches that path against patterns written with forward slashes:

https://github.com/TanStack/query/blob/main/scripts/verify-links.ts#L56-L68

```ts
const isExample = absPath.includes('/examples/')

if (isExample) {
  absPath = absPath.replace(
    /\/docs\/framework\/([^/]+)\/examples\//,
    '/examples/$1/',
  )
  exists = existsSync(absPath) && statSync(absPath).isDirectory()
}
```

On Windows `absPath` is `C:\Users\...\docs\framework\react\examples\simple`, so `includes('/examples/')` is false. The example branch never runs, the path is never rewritten to `examples/react/simple`, and the link is looked up as a markdown file that does not exist.

CI does not catch this because it runs on Linux.

Clone the repository on Windows and run the script directly:

```
pnpm install
node scripts/verify-links.ts
```

### Steps to reproduce

1. Clone the repository on Windows.
2. Run `pnpm install`.
3. Run `node scripts/verify-links.ts` (or `pnpm test:docs`).

Output:

```
Found 494 markdown files

❌ Found 27 broken links:
docs/framework/lit/quick-start.md
  link:      ./examples/basic
  resolved:  C:\Users\...\query\docs\framework\lit\examples\basic.md
  why:       Not found
...
```

The resolved path in the report shows the problem directly: it still points into `docs/framework/lit/examples/` instead of `examples/lit/`.

### Expected behavior

The check should pass on Windows just as it does on CI, so contributors can run `pnpm test:docs` locally before pushing. CONTRIBUTING.md points people at these commands.

Every time

### Screenshots or Videos

_No response_

### Platform

- OS: Windows 11
- Node: 24.11.1
- pnpm: 11.9.0

### Tanstack Query adapter

None

### TanStack Query version

current `main`

### TypeScript version

5.9.3

### Additional context

CONTRIBUTING.md does note that the repository uses symlink based configuration and recommends WSL, but this script does not need symlinks and works fine once the separators are normalized. Fix and PR coming.

yogesh968:
Fixes #11177

### The problem

`relativeLinkExists` builds the target path with `path.resolve`, which uses the platform separator, and then matches that path against patterns written with forward slashes:

```ts
const isExample = absPath.includes('/examples/')

if (isExample) {
  absPath = absPath.replace(
    /\/docs\/framework\/([^/]+)\/examples\//,
    '/examples/$1/',
  )
  exists = existsSync(absPath) && statSync(absPath).isDirectory()
}
```

On Windows the resolved path uses backslashes, so neither pattern matches. Every documentation link pointing at an example skips the rewrite and gets looked up as a markdown file under `docs/framework/<name>/examples/` instead of a directory under `examples/<name>/`.

`pnpm test:docs` then reports 27 broken links that are all fine. CI does not catch it because it runs on Linux, and the reported `resolved:` path shows the problem directly.

### The change

Normalize the separators once, before the two comparisons. Node accepts forward slashes on Windows, so the path stays usable for `existsSync` and `statSync`, and the value printed in the error report stays readable.

The `docsRoot` containment check above it still runs on the platform native path, so it is unaffected.

### Tests

`node scripts/verify-links.ts` on Windows goes from 27 reported broken links to "No broken links found". Behaviour on Linux is unchanged, since `split(sep).join('/')` is a no-op there.

No changeset, this is repository tooling.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved link verification consistency across operating systems.
  * Ensured example paths are detected reliably regardless of platform-specific path separators.

### DIFF
--- a/scripts/verify-links.ts
+++ b/scripts/verify-links.ts
@@ -1,5 +1,5 @@
 import { existsSync, readFileSync, statSync } from 'node:fs'
-import { extname, resolve } from 'node:path'
+import { extname, resolve, sep } from 'node:path'
 import { glob } from 'tinyglobby'
 // @ts-ignore Could not find a declaration file for module 'markdown-link-extractor'.
 import markdownLinkExtractor from 'markdown-link-extractor'
@@ -52,6 +52,11 @@ function relativeLinkExists(link: string, file: string): boolean {
     })
     return false
   }
+
+  // `resolve` uses the platform separator, but the patterns below are written
+  // with forward slashes, so match against a normalized copy of the path.
+  // Node accepts forward slashes on Windows too, so the result stays usable.
+  absPath = absPath.split(sep).join('/')
 
   // Check if this is an example path
   const isExample = absPath.includes('/examples/')


TkDodo (approved):

yogesh968:
thnku very much <>

PR #11186, opened 2026-08-14
fix(solid-query): attach 'useQueries' through the provider's hydration channel

### LINKED ISSUE #11185: solid-query v6: useQueries doesn't attach to the new hydration channel from #11168
#11168 (merged today on `solid-query-v6-pre`, targeting `@tanstack/solid-query@6.0.0-beta.x`) adds a provider-owned hydration channel so `QueryClientProvider` primes the cache from the server-dehydrated stream and `useBaseQuery` waits for its own entry to be primed (`whenQueryPrimed`) before attaching its observer, instead of doing a mount refetch on data that's about to arrive.

`useQueries` doesn't go through any of that. I read the current `packages/solid-query/src/useQueries.ts` on `solid-query-v6-pre`: it builds its own `QueriesObserver` directly —

```ts
const observer = untrack(
  () => new QueriesObserver<TCombinedResult>(client(), defaultedQueries(), ...),
)
...
createEffect(() => {
  if (!isRestoring()) {
    unsubscribe = observer.subscribe((result) => { ... })
  }
})
```

— with no reference to `hydrationChannel`, `whenQueryPrimed`, or the `executorRan` replay-detection flag `useBaseQuery` added (confirmed with `rg "whenQueryPrimed|executorRan|hydrat"` over the file: zero matches). The observer is constructed and its optimistic result read synchronously on mount, so if this runs before the provider's channel has applied `hydrate()` for these query keys, `useQueries` sees a cold cache and behaves exactly like the pre-#11168 `useBaseQuery` did: no priming coordination, so a query whose data is already in flight from the SSR stream can still trigger a mount refetch even though it's well within `staleTime`.

This isn't a guess — the PR body says so explicitly under "Coexistence with TanStack Start": *"`useQueries` does not go through this path (it did not serialize hydration data before either); with the channel in place its queries would be primed too as soon as they settle on the server — client attach for it could be covered in a follow-up."*

Since this is still pre-release (`solid-query-v6-pre`, not yet on `main`), flagging it now while the hydration channel work is fresh in mind seems more useful than after the branch merges and beta ships. Happy to help verify a fix against the same `hydration.test.tsx` SSR/hydration fixture #11168 added if that's useful.

lazerg:
## 🎯 Changes

`useQueries` never got the hydration coordination that #11168 added to `useBaseQuery`. It builds its own `QueriesObserver` and subscribes from an effect, so a component that hydrates before its dehydrated entries have arrived on the SSR stream attaches to a cold cache and refetches on mount, even for data that is well inside `staleTime`.

The subscription now goes through the same coordinator. With a provider in place, `useQueries` attaches once every one of its queries has been primed, or once the channel reports itself done. One `QueriesObserver` covers all of the queries, so it can only attach when the last one is ready. `QueryClientProvider` closes the channel right away on a fresh client mount, detected with the same `Promise` executor probe `useBaseQuery` uses, so an app without SSR never waits.

For the test, the streaming fixture gains a `useQueries` in the shell whose query settles after the slow boundary, so its entry only rides the final flush. Before the change it refetches at shell hydration.

Fixes #11185

## ✅ Checklist

- [x] I have tested this code locally.

Local verification: `vitest run` on `solid-query` (330 tests), `solid-query-persist-client` and `solid-query-devtools`, all green with no type errors; eslint and prettier clean.

## 🚀 Release Impact

### DIFF
--- a/.changeset/solid-usequeries-hydration-channel.md
+++ b/.changeset/solid-usequeries-hydration-channel.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/solid-query': patch
+---
+
+fix: attach `useQueries` through the provider's hydration channel, so a hydrated `useQueries` waits for its entries to be primed instead of refetching data that is still streaming in from the server.
--- a/packages/solid-query/src/QueryClientProvider.tsx
+++ b/packages/solid-query/src/QueryClientProvider.tsx
@@ -69,10 +69,21 @@ export const QueryClientProvider = (
   // query-core hydrate() (newer-wins) and unblocks `useBaseQuery`
   // subscribers waiting on their query's entry.
   //
-  // Client, fresh mount: the compute returns undefined and the effect
-  // never fires.
+  // Client, fresh mount: the compute runs for real and returns undefined,
+  // so the channel is closed immediately and nothing waits on it.
+  const replayProbe = { executorRan: false }
   const [channelValue] = createSignal<DehydrationChannelYield | undefined>(
-    () => (isServer ? createServerDehydrationChannel(props.client) : undefined),
+    () => {
+      if (isServer) return createServerDehydrationChannel(props.client)
+      // Replay detection, as in useBaseQuery: a real Promise runs its
+      // executor synchronously, the hydration mock does not, so the
+      // executor running means this compute was not replayed from a
+      // serialized channel.
+      void new Promise<void>(() => {
+        replayProbe.executorRan = true
+      })
+      return undefined
+    },
   )
   const coordinator = isServer
     ? null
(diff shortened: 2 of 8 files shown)


MILLERMARRU (commented):
I checked this against the pre-PR `QueryClientProvider.tsx` on the `solid-query-v6-pre` branch and the existing code comment literally says the effect never fires for a fresh client mount. That's fine as long as nothing waits on the channel, but this PR makes `useQueries` wait on it via `coordinator.whenQueryPrimed`. Without the `replayProbe` addition, a fresh client-only mount (no SSR at all) would leave every `useQueries` call stuck waiting for entries that will never arrive. The executor-runs-synchronously trick for distinguishing a real fresh-mount Promise from a replayed hydration signal matches the same pattern already used by `useBaseQuery`, so it's consistent with the existing convention rather than a new one. The added `StreamApp` fixture query that settles last but hydrates in the shell is a good regression test for the actual race being fixed.

brenelz:
[solid-query-v6-pre](https://github.com/TanStack/query/tree/solid-query-v6-pre) going under a rewrite so will need to revisit this after

lazerg:
Got it, thanks for the heads up. I'll revisit this once the solid-query-v6-pre rewrite lands. No rush on my end.

brenelz:
Are you able to revisit and see if this pr still relevant

brenelz:
It also has conflicts

brenelz:
I feel this probably needs to be redone off the new native 2.0 rewrite. THe hydration changed quite a bit

Issue #11102, opened 2026-07-22
[vue-query] queryClient.fetchQuery passes getter options through without unwrapping


byungsker:
### Describe the bug

`QueryClient.fetchQuery` in `@tanstack/vue-query` has an overload that accepts getter-function options, but the runtime implementation currently passes that top-level function through to query-core instead of evaluating it first.

The overload allows this shape:

```ts
queryClient.fetchQuery(() => ({
  queryKey: ["foo", ref("bar")],
  queryFn,
}))
```

But the implementation calls `cloneDeepUnref(options)` directly. `cloneDeepUnref` unwraps refs and nested query-key getters, but it does not call a top-level function value. As a result, query-core receives the function itself instead of the resolved `FetchQueryOptions` object.

### Your minimal, reproducible example

Focused regression test in the linked PR: https://github.com/TanStack/query/pull/11103/files#diff-3d6cc18a7ff7e8e5afb08c0e3cba8d9f3e08fb9c6f6c59f1f5f58dc12f3f8b8e

Minimal code shape:

```ts
import { QueryClient } from "@tanstack/vue-query"
import { ref } from "vue"

const queryClient = new QueryClient()
const queryFn = () => Promise.resolve("mock")

queryClient.fetchQuery(() => ({
  queryKey: ["foo", ref("bar")],
  queryFn,
}))
```

### Steps to reproduce

1. Create a Vue Query `QueryClient`.
2. Call `queryClient.fetchQuery` with a getter function returning `FetchQueryOptions`.
3. Observe that the getter is not evaluated before delegating to query-core.

A focused test shows the current behavior clearly:

```ts
queryClient.fetchQuery(() => ({
  queryKey: queryKeyRef,
  queryFn: fn,
}))

expect(QueryClientOrigin.prototype.fetchQuery).toHaveBeenCalledWith({
  queryKey: queryKeyUnref,
  queryFn: fn,
})
```

Currently the received argument is `[Function anonymous]`.

### Expected behavior

`queryClient.fetchQuery(() => options)` should evaluate the top-level getter, then deep-unref the returned options object before passing it to query-core.

### How often does this bug happen?

Every time

### Platform

- OS: macOS
- Browser: N/A
- Runtime: Node.js v24.13.0
- Adapter: vue-query
- Vue runtimes tested: Vue 2 / Vue 2.7 / Vue 3

### Tanstack Query adapter

vue-query

### TanStack Query version

[...]

PR #11208, opened 2026-08-17
Vue: add new impertive methods to vue proxy QueryClient

DogPawHat:
## 🎯 Changes

Proxy the `queryClient.query` and `queryClient.infiniteQuery` methods to the vue-query proxy query client. Adds tests, docs and updates the usePrefetchQuery hooks. 

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **New Features**
  * Added reactive `query` and `infiniteQuery` methods to the Vue Query client.
  * Added support for Vue refs, selectors, pagination, and improved type inference.
  * Updated prefetch composables to use the new query execution APIs.

* **Documentation**
  * Updated Vue prefetching and SSR guides with current query APIs.
  * Documented static stale-time behavior and refreshed Nuxt examples.

* **Deprecations**
  * Added migration guidance for older query and prefetch methods.


packages/vue-query/src/queryClient.ts
```diff
@@ -231,6 +236,78 @@ export class QueryClient extends QC {
...
+          QueryExecuteOptions<
+            TQueryFnData,
+            TError,
+            TData,
+            TQueryData,
+            TQueryKey,
+            TPageParam
+          >
+        >),
+  ): Promise<TData>
```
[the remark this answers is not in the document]
DogPawHat: This behavior was copied from fetchQuery. that's being fixed in https://github.com/TanStack/query/pull/11103 supposedly, I don't know enough about Vue's reactivity system to say it fixes it (other then multiple AI systems flagging it)

packages/vue-query/src/queryClient.ts
```diff
@@ -231,6 +236,69 @@ export class QueryClient extends QC {
...
+    options: MaybeRefDeep<
+      QueryExecuteOptions<
+        TQueryFnData,
+        TError,
+        TData,
+        TQueryData,
+        TQueryKey,
+        TPageParam
+      >
+    >,
```
DogPawHat, line 275: I've decided not to bring over the `(() => MaybeRefDeep<QueryExecuteOptions>)` type that's on `fetchQuery`, `fetchInfiniteQuery` and the other old methods as there's a good bit of evidence that there not  implemented properly. Specifically, per https://github.com/TanStack/query/issues/11102 (along with numerious other AI systems yapping at me), `fetchQuery` won't resolve a queryOptions getter and just pass a function down to query-core where it breaks.

I think in an imperative setting, if you have a getter like this:
```
const options = queryOptions(() => ({
  queryKey: ['foo']
  ...otherStuff
}))
```
Simply doing `queryClient.query(options())` is an ok workaround, as were doing the execution outside of Vue's reactivity engine anyway. I don't think we want to block finishing off https://github.com/TanStack/query/discussions/9135 on resolving #11102 .

DogPawHat:
@coderabbitai full review again

packages/vue-query/src/queryClient.ts
```diff
@@ -362,6 +502,9 @@ export class QueryClient extends QC {
     return super.fetchInfiniteQuery(cloneDeepUnref(options))
   }
 
+  /**
+   * @deprecated use void queryClient.infiniteQuery(options)instead. You can swallow errors with `.catch(noop)`. This method will be removed in the next major version.
```
TkDodo:
```suggestion
   * @deprecated use void queryClient.infiniteQuery(options) instead. You can swallow errors with `.catch(noop)`. This method will be removed in the next major version.
```

packages/vue-query/src/usePrefetchQuery.ts
```diff
@@ -59,11 +82,12 @@ export function usePrefetchQuery<
       TQueryFnData,
       TError,
       TData,
+      TQueryData,
       TQueryKey
     > = cloneDeepUnref(resolvedOptions)
 
     if (!client.getQueryState(clonedOptions.queryKey)) {
-      void client.prefetchQuery(clonedOptions)
+      void client.query(clonedOptions).then(noop).catch(noop)
```
TkDodo: since we `void` anyways, why is the `.then` necessary?


```suggestion
      void client.query(clonedOptions).catch(noop)
```

DogPawHat: good catch. removed all of them where we use `void`

PR #11218, opened 2026-08-18
fix(query-core): release the retryer once a mutation settles

### LINKED ISSUE #11216: Mutation.execute() never releases #retryer after settling, unlike Query.fetch() after #11163
## Describe the bug

`Mutation.execute()` never clears `this.#retryer` after the mutation settles, unlike `Query.fetch()`, which just got this exact fix in #11163 (merged today). This means a settled `Mutation` instance keeps its `Retryer` (and everything closed over by it: the resolved/rejected result, `variables`, `mutationFnContext`) reachable for as long as the `Mutation` instance itself is retained by `MutationCache`.

## Your minimal, reproducible example

`packages/query-core/src/mutation.ts`, `execute()`:

```ts
} finally {
  this.#mutationCache.runNext(this)
}
```

Compare with `packages/query-core/src/query.ts`'s `fetch()` after #11163:

```ts
} finally {
  // The settled retryer's promise would otherwise pin this fetch's raw
  // result (a second copy after structural sharing) for the query's lifetime
  if (this.#retryer === retryer) {
    this.#retryer = undefined
  }
  this.scheduleGc()
}
```

`Mutation`'s `finally` block only calls `runNext`, it never sets `this.#retryer = undefined`.

## Steps to reproduce

1. Look at `#11163`'s changeset/PR description: `Query.fetch()` left `#retryer` set after a fetch settled, and the retryer's resolved promise held the raw fetch result, so a query that had refetched held its data twice: once via structural sharing in `state.data`, and again through the settled retryer's closure. Same shape, same class, different method.
2. `Mutation.execute()` builds `this.#retryer = createRetryer({...})` the same way `Query.fetch()` does, `await`s `this.#retryer.start()` inside `try`, and has a `finally` block — but that `finally` only runs `this.#mutationCache.runNext(this)`, it never releases `this.#retryer`.
3. `MutationCache` keeps completed `Mutation` instances around until `gcTime` elapses (or indefinitely while an observer/dehydration still references them, per `Removable`/`scheduleGc`), so for that whole window the settled retryer (and whatever it closed over) stays reachable through `mutation.#retryer`, not just through `mutation.state.data`/`variables`.

## Expected behavior

Same treatment `Query.fetch()` just got: once the retryer's promise settles, release the reference (guarded by an identity check like `if (this.#retryer === retryer)`, since `execute()` can in principle be called again on the same instance via `continue()`).

## Additional context

Not filing a PR myself, wanted to flag this mirrors #11163 closely enough that whoever wrote that fix for `Query` is probably the right person to confirm the same identity-check subtlety applies to `Mutation` (I don't have full confidence on interaction with `continue()` re-entering `execute()` on the same instance).

iamshahid1997:
Fixes #11216

## 🎯 Changes

`Mutation.execute()` never cleared `#retryer` after the mutation settled, so the settled retryer's promise kept that mutation's result, `variables` and `mutationFnContext` reachable for as long as the `MutationCache` retained the `Mutation` instance. This mirrors what `Query.fetch()` had before #11163.

This clears `#retryer` in the existing `finally` once the execution has settled, with the same identity check used for queries so that a mutation re-executed from a cache `onSuccess`/`onSettled` callback keeps its own retryer.

### The `continue()` interaction

The issue flagged uncertainty about `continue()` re-entering `execute()` on the same instance, and it is a real hazard: releasing the retryer alone makes `continue()` on a settled mutation fall through to `this.execute(this.state.variables!)` and run the `mutationFn` a second time. The added test `should not re-execute a settled mutation when it is continued` fails with only the `finally` change applied.

Both internal callers (`MutationCache.runNext` and `resumePausedMutations`) filter on `state.isPaused`, so neither reaches a settled mutation — but `continue()` is public and reachable directly. The fallback is therefore gated on the mutation still being `pending`, which is the same condition `execute()` already uses to detect a restored (dehydrated) mutation, so the intended hydration path is unchanged.

## ✅ Checklist

## 🚀 Release Impact

## Tests

Two tests in `mutations.test.tsx`:

1. `should not re-execute a settled mutation when it is continued` — guards the `continue()` behavior change described above
2. `should still continue a restored paused mutation that has no retryer` — covers the dehydration path the fallback exists for

Full `query-core` (629) and `react-query` (577) suites pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Settled mutations no longer restart when continued.
  - Retained mutation-cache entries no longer keep completed execution results, variables, or context in memory.
  - Restored paused mutations without a retryer can still continue and complete successfully.
  - Mutation continuation now resolves reliably without reusing completed execution state.

- **Tests**
  - Added coverage for settled mutation continuation, retryer cleanup, and restored paused mutation behavior.


TkDodo (changes requested):
all the tests here also pass on `main`, which means they don’t test what they claim to fix. please add better tests that actually show that this code fixes something!

iamshahid1997:
You're right — those tests only guarded the `continue()` gating, they didn't demonstrate the fix. Pushed two that fail on `main`.

`continue()` is the only reader of `#retryer` outside `execute()`. While a settled retryer is still referenced, `continue()` hands back that retryer's promise — which resolves with the raw result it closed over, or rejects with its error. Once released, `continue()` on a settled mutation resolves with `undefined`, so `await expect(mutation.continue()).resolves.toBeUndefined()` fails on `main` for both the success and error cases. Same observable surface #11163 used for queries (`query.promise` being `undefined` after a fetch settles) — `Mutation` just doesn't expose the retryer any other way.

I kept the two original tests since the change does need that `continue()` gating covered, but they're guards rather than proof — happy to drop them if you'd prefer.

TkDodo (approved):

PR #11219, opened 2026-08-18
fix(query-core): release the mutation retryer once execute settles

### LINKED ISSUE #11216: Mutation.execute() never releases #retryer after settling, unlike Query.fetch() after #11163

Gujiassh:
## Summary

Fixes #11216.

`Mutation.execute()` kept `this.#retryer` after the mutation settled, while `Query.fetch()` already clears it after #11163. The settled retryer promise retains that mutation's variables and result for as long as `MutationCache` keeps the instance.

This change mirrors the query cleanup: capture the local retryer, clear `this.#retryer` in `finally` when it still owns that instance, then run the next queued mutation.

## Validation

- `pnpm --filter @tanstack/query-core exec vitest run src/__tests__/mutations.test.tsx` — 25 passed

## Summary by CodeRabbit

- **Bug Fixes**
  - Settled mutations now release retry-related resources after execution completes.
  - Mutation variables and results are no longer retained unnecessarily for the lifetime of the mutation cache.
  - Continuing a settled mutation starts a fresh execution as expected.

### DIFF
--- a/.changeset/release-mutation-settled-retryer.md
+++ b/.changeset/release-mutation-settled-retryer.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/query-core': patch
+---
+
+Release a mutation's retryer once execute settles, so the settled promise no longer keeps that mutation's variables and result reachable for the MutationCache lifetime.
--- a/packages/query-core/src/__tests__/mutations.test.tsx
+++ b/packages/query-core/src/__tests__/mutations.test.tsx
@@ -1166,6 +1166,30 @@ describe('mutations', () => {
     })
   })
 
+  it('should release the retryer once its mutation has settled', async () => {
+    let count = 0
+    const observer = new MutationObserver(queryClient, {
+      mutationFn: () => {
+        count += 1
+        return sleep(10).then(() => 'data')
+      },
+    })
+
+    const mutatePromise = observer.mutate()
+    await vi.advanceTimersByTimeAsync(10)
+    await expect(mutatePromise).resolves.toBe('data')
+    expect(count).toBe(1)
+
+    const mutation = queryClient.getMutationCache().getAll()[0]!
+    // With the retryer cleared, continue() falls through to a fresh execute().
+    // If the settled retryer were retained, continue() would return it and skip
+    // the mutation function.
+    const continued = mutation.continue()
+    await vi.advanceTimersByTimeAsync(10)
+    await expect(continued).resolves.toBe('data')
+    expect(count).toBe(2)
+  })
+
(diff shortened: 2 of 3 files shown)


TkDodo:
duplicate of #11218

PR #11231, opened 2026-08-19
fix(query-core): notify sibling observers when one unsubscribes during dispatch

contactjawad:
## Summary

`Query.#dispatch` can skip a still-subscribed observer when another observer on the same query unsubscribes during notification, leaving it with a stale `pending` result after the query has already resolved.

## Root cause

`removeObserver` mutates the observer array in place (`this.observers.splice(index, 1)`). `#dispatch` iterates that same live array with `this.observers.forEach(...)` and calls `onQueryUpdate()` synchronously inside the open `notifyManager` batch. If a listener unsubscribes a same-query observer during that loop, an element is spliced out mid-iteration and `forEach` skips the next observer.

## Fix

Iterate over a snapshot (`[...this.observers]`) so an observer unsubscribing during notification can't cause a sibling to be skipped.

## Test

Added a regression test in `queryObserver.test.tsx`: two observers subscribe to the same query; observer A unsubscribes itself when it first sees `success`, and the test asserts observer B still receives the resolved `success` data. It fails on `main` (`expected 'pending' to be 'success'`) and passes with this change. A changeset is included.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed query updates being skipped for some observers when another observer unsubscribes during notification.
  - Ensured all still-subscribed observers receive successful query results reliably, including when subscriptions change during an update.

- **Tests**
  - Added regression coverage for observer notifications during subscription changes and asynchronous query completion.

### DIFF
--- a/.changeset/fix-query-dispatch-observer-skip.md
+++ b/.changeset/fix-query-dispatch-observer-skip.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/query-core': patch
+---
+
+Fix an observer being skipped during query notification when another observer on the same query unsubscribes mid-dispatch. `Query.#dispatch` iterated the live `observers` array while `onQueryUpdate()` could splice it, so a still-subscribed sibling could be left with a stale `pending` result after the query resolved. The notification now iterates over a snapshot of the observers.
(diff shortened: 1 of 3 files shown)


TkDodo (approved):

TkDodo:
merging this one:
- https://github.com/TanStack/query/pull/11234

PR #11239, opened 2026-08-20
fix(solid-query): support combine functions returning any object shape

### LINKED ISSUE #7522: [Bug?]: Result type of combine works not as expected

wasaybuilds:
Fixes #7522

### Problem

`useQueries` (`createQueries`) put whatever `combine` returned into the `createStore` that backs the hook, but everything built around that store — the per-query `createResource`s, the `data` proxies, and the `setState(index, ...)` calls in the observer subscription — is keyed by query index and assumes the store holds the results array.

As a result, a `combine` function that returned any other shape:

1. failed to type check, because `TCombinedResult` was constrained to `QueriesResults<T>` (the React adapter has no such constraint), and
2. threw `TypeError: state.map is not a function` at runtime once the constraint was worked around.

@TkDodo [pinpointed the type constraint](https://github.com/TanStack/query/issues/7522#issuecomment-2755142799) and noted that simply removing it runs into `createStore`'s `T extends object` requirement plus follow-up errors in `createMemo` — those come from the store being asked to hold the combined value in the first place.

### Fix

Keep the store holding the **raw** results (so the resources, proxies and index-keyed updates keep working exactly as before) and derive the combined value from the tracked proxies on top of it, which is how the React adapter separates the two.

- The store is now seeded from `getOptimisticResult(...)[0]` (raw results) instead of `[1]()` (combined result).
- When `combine` is set, the result is a `createMemo` over the result proxies, exposed through a small proxy so consumers stay subscribed to just the properties they read.
- The proxy target is `[]` or `{}` depending on what `combine` returns, so `Array.isArray`, `JSON.stringify`, spreading and `Object.keys` keep behaving as they did for combine functions that return the results array.
- `TCombinedResult` is now constrained to `object` rather than `QueriesResults<T>`. Solid needs an object to stay reactive, so a combined primitive is reported as a type error instead of crashing at runtime.

The `@ts-expect-error` on `setState(index, ...)` is gone as well — with the store typed as the raw results, that index write type checks on its own.

### Tests

- `useQueries.test.tsx`: a combine function returning `{ data, pending }` renders and updates as the queries resolve (this is the failing case from the issue), and a combine function returning the results array stays array-like.
- `useQueries.test-d.tsx`: `useQueries` with a combine function returning `{ data: boolean; pending: boolean }` infers exactly that type.

`pnpm test` for `@tanstack/solid-query` passes (338 tests, type tests included).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `useQueries` and `createQueries` so `combine` functions can return custom object shapes, not only result arrays.
  * Prevented type-checking failures and runtime errors when using custom combined results.
  * Preserved array-like behavior when the combined result is the original query results array.

### DIFF
--- a/.changeset/solid-query-combine-result-shape.md
+++ b/.changeset/solid-query-combine-result-shape.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/solid-query': patch
+---
+
+Fix `useQueries`/`createQueries` rejecting a `combine` function that returns a shape other than the results array, which previously failed to type check and threw `state.map is not a function` at runtime.
--- a/packages/solid-query/src/__tests__/useQueries.test-d.tsx
+++ b/packages/solid-query/src/__tests__/useQueries.test-d.tsx
@@ -296,6 +296,27 @@ describe('useQueries', () => {
     }))
   })
 
+  it('should allow combine to return a shape other than the results array', () => {
+    const result = useQueries(() => ({
+      queries: [
+        {
+          queryKey: queryKey(),
+          queryFn: () => Promise.resolve(1),
+        },
+        {
+          queryKey: queryKey(),
+          queryFn: () => Promise.resolve(2),
+        },
+      ],
+      combine: (results) => ({
+        data: results.every((queryResult) => queryResult.data),
+        pending: results.some((queryResult) => queryResult.isPending),
+      }),
+    }))
+
+    expectTypeOf(result).toEqualTypeOf<{ data: boolean; pending: boolean }>()
+  })
+
   describe('type parameters', () => {
     it('should handle type parameter - tuple of tuples', () => {
       const key1 = queryKey()
--- a/packages/solid-query/src/__tests__/useQueries.test.tsx
+++ b/packages/solid-query/src/__tests__/useQueries.test.tsx
@@ -159,6 +159,92 @@ describe('useQueries', () => {
     expect(rendered.getByText('data: custom client')).toBeInTheDocument()
   })
 
+  it('should support a combine function that returns a shape other than the results array', async () => {
+    const key1 = queryKey()
+    const key2 = queryKey()
+
+    function Page() {
+      const result = useQueries(() => ({
+        queries: [
+          {
+            queryKey: key1,
+            queryFn: () => sleep(10).then(() => 1),
+          },
+          {
+            queryKey: key2,
+            queryFn: () => sleep(20).then(() => 2),
+          },
+        ],
+        combine: (results) => ({
+          data: results.map((queryResult) => queryResult.data),
+          pending: results.some((queryResult) => queryResult.isPending),
+        }),
+      }))
+
+      return (
+        <div>
+          <div data-testid="data">{JSON.stringify(result.data)}</div>
+          <div data-testid="pending">{String(result.pending)}</div>
+        </div>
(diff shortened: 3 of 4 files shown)


wasaybuilds:
Thanks, the `TypeError` was real — pushed 7cdc713.

**Result category drift (fixed).** With an array target, `length` stays a non-configurable own property, so once `combine` returned a non-array the `ownKeys` trap no longer reported it and `ownKeys`/`Object.keys`/spreading threw. The traps now report the properties the target owns itself with the target's own flags (and `has` reports them as present), so no trap can violate an invariant when the shape changes. Covered by `should not throw when the shape returned by combine changes`, which fails with the previous traps at exactly `'ownKeys' on proxy: trap result did not include 'length'`.

**`combine` presence.** Removing `combine` after the fact is handled — the memo falls back to the results array. Adding it later cannot change what the component already returned, since a Solid component hands out the value once rather than re-rendering; that case also returned the raw results array before this PR, so it is not a new behaviour. I left it as is (with a comment) rather than routing the no-`combine` path through the proxy too, to keep the default path byte-for-byte unchanged.

One remaining rough edge worth flagging: if the category does drift, `Array.isArray` and `JSON.stringify` still reflect the target picked at creation time. That is inherent to a fixed proxy target, and it degrades instead of throwing now.

wasaybuilds:
One more argument for this being a bug rather than a missing feature: the Solid `useQueries` reference is a `ref:` of the React page, so the whole **Combine** section is published on the Solid docs with the package name and the accessor rewritten:

https://github.com/TanStack/query/blob/main/docs/framework/solid/reference/useQueries.md

That section documents

```tsx
combine: (results) => {
  return {
    data: results.map((result) => result.data),
    pending: results.some((result) => result.isPending),
  }
}
```

and states that "`combinedQueries` will be an object with a `data` and a `pending` property". On `main` that exact snippet does not type check in Solid and throws `state.map is not a function` at runtime — it is the case the first new test covers. So no docs change is needed here; the fix makes the adapter match the API its own docs already describe.

PR #11246, opened 2026-08-21
test: transition tests

TkDodo:
## Summary by CodeRabbit

* **Tests**
  * Added coverage for React transitions with suspense-enabled queries.
  * Verified previously displayed data remains visible while new data loads.
  * Confirmed results update correctly after asynchronous loading completes.

### DIFF
--- a/packages/react-query/src/__tests__/transition.test.tsx
+++ b/packages/react-query/src/__tests__/transition.test.tsx
@@ -1,0 +1,114 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import * as React from 'react'
+import { act, fireEvent } from '@testing-library/react'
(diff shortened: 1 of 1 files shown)


packages/react-query/src/__tests__/transition.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+    expect(rendered.getByText('data: test0')).toBeInTheDocument()
+
+    await act(() =>
+      fireEvent.click(rendered.getByRole('button', { name: 'increment' })),
+    )
+
+    // the deferred key is still the old one, so the old data stays on screen
+    // while the new query is in flight. `toBeVisible` matters here: on a
+    // fallback React keeps the old subtree mounted but hides it with
+    // `display: none`, so `toBeInTheDocument` would also pass.
```
TkDodo:
```suggestion
```

PR #11253, opened 2026-08-22
perf(query-core): skip no-op hydration callbacks and export dehydrateQuery

schiller-manuel:
## Summary by CodeRabbit

* **New Features**
  * Exported `dehydrateQuery` for direct use.
  * Made data serialization and deserialization optional during hydration.
  * Improved hydration performance by avoiding unnecessary transformations.
  * Added flexible error-redaction handling with secure defaults.

* **Bug Fixes**
  * Preserved data correctly when no serialization or deserialization callback is provided.

* **Chores**
  * Prepared a patch release for `@tanstack/query-core`.

### DIFF
--- a/packages/query-core/src/__tests__/hydration.test.tsx
+++ b/packages/query-core/src/__tests__/hydration.test.tsx
@@ -22,6 +22,8 @@ describe('dehydration and rehydration', () => {
       const queryClient = new QueryClient()
       queryClient.setQueryData(key, 'data')
       const query = queryClient.getQueryCache().find({ queryKey: key })!
+      const dehydratedAt = new Date('2024-01-01T00:00:00.000Z')
+      vi.setSystemTime(dehydratedAt)
 
       const dehydrated = dehydrateQuery(query)
 
@@ -30,7 +32,7 @@ describe('dehydration and rehydration', () => {
         queryKey: key,
         state: query.state,
       })
-      expect(dehydrated.dehydratedAt).toBe(Date.now())
+      expect(dehydrated.dehydratedAt).toBe(dehydratedAt.getTime())
       expect(dehydrated.state).not.toBe(query.state)
       expect(dehydrated.promise).toBeUndefined()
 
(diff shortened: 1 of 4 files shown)


packages/query-core/src/hydration.ts
```diff
@@ -89,51 +86,57 @@ function dehydrateMutation(mutation: Mutation): DehydratedMutation {
...
+
+  return promise
+}
+
 // Most config is not dehydrated but instead meant to configure again when
 // consuming the de/rehydrated data, typically with useQuery on the client.
 // Sometimes it might make sense to prefetch data on the server and include
 // in the html-payload, but not consume it on the initial render.
-function dehydrateQuery(
+export function dehydrateQuery(
```
TkDodo, line 121: if you want this export to be available, we need to add it to `src/index.ts`

schiller-manuel: on it

packages/query-core/src/hydration.ts
```diff
@@ -89,51 +86,57 @@ function dehydrateMutation(mutation: Mutation): DehydratedMutation {
   }
 }
 
+function dehydratePromise(
```
TkDodo, line 89: was moving this to a separate function part of the perf improvement? it’s only used once and passing all the params seems like it’s the culprit for the bundle size increase 🤔

packages/query-core/src/__tests__/hydration.test.tsx
```diff
@@ -15,6 +16,68 @@ describe('dehydration and rehydration', () => {
...
+      const query = queryClient.getQueryCache().find({ queryKey: key })!
+
+      const dehydrated = dehydrateQuery(query)
+
+      expect(dehydrated).toMatchObject({
+        queryHash: query.queryHash,
+        queryKey: key,
+        state: query.state,
+      })
+      expect(dehydrated.dehydratedAt).toBe(Date.now())
```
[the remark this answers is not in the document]
TkDodo: this sounds legit ⬆️

packages/query-core/src/hydration.ts
schiller-manuel: we don't need to instantiate a new function for each dehydrate call, yes it's a perf optimization. will optimize bundlesize , I forgot this is used on the client as well

TkDodo (approved):

PR #11260, opened 2026-08-23
fix(query-core): accept partial dehydrated state

schiller-manuel:
## Summary by CodeRabbit

- **New Features**
  - `hydrate` now supports partial dehydrated state objects, allowing mutation or query collections to be omitted.
  - Empty mutation and query collections continue to be supported during hydration.
  - Improved type definitions provide clearer guidance when supplying dehydrated state.

- **Bug Fixes**
  - Hydration now handles incomplete state payloads more reliably without requiring unused collections.

### DIFF
--- a/.changeset/soft-rivers-hydrate.md
+++ b/.changeset/soft-rivers-hydrate.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/query-core': patch
+---
+
+Type hydration input as a partial dehydrated state so omitted mutation and query collections are supported.
--- a/packages/query-core/src/__tests__/hydration.test.tsx
+++ b/packages/query-core/src/__tests__/hydration.test.tsx
@@ -676,22 +676,13 @@ describe('dehydration and rehydration', () => {
     consoleMock.mockRestore()
   })
 
-  it('should not hydrate if the hydratedState is null or is not an object', () => {
-    const queryCache = new QueryCache()
-    const queryClient = new QueryClient({ queryCache })
-
-    expect(() => hydrate(queryClient, null)).not.toThrow()
-    expect(() => hydrate(queryClient, 'invalid')).not.toThrow()
-
-    queryClient.clear()
-  })
-
   it('should support hydratedState with undefined queries and mutations', () => {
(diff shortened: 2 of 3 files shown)


TkDodo (approved):

peterbud:
Hi @schiller-manuel and @TkDodo  - this change of removing then `null` and `undefined` checks from the `hydrate` function is now causing at least in vue-query a major problems in CSR pages, where the incoming `dehydratedState` is null, and hydrate just throws error, when the plugin is set up as the documentation describes here:

https://tanstack.com/query/latest/docs/framework/vue/guides/ssr#using-nuxt-js

```ts
 if (import.meta.client) {
    hydrate(queryClient, vueQueryState.value)
  }
``` 

In CSR, `vueQueryState` is null, there is no checks anymore in vue-query `hydrate`, and it tries to access `dehydratedState.mutations` leading to a throw.

So as I see @coderabbitai was right above.

Can you advise how we can avoid the regression?

TkDodo:
> Hi @schiller-manuel and @TkDodo - this change of removing then `null` and `undefined` checks from the `hydrate` function is now causing at least in vue-query a major problems in CSR pages, where the incoming `dehydratedState` is null, and hydrate just throws error, when the plugin is set up as the documentation describes here:
> 
> https://tanstack.com/query/latest/docs/framework/vue/guides/ssr#using-nuxt-js
> 
> ```ts
>  if (import.meta.client) {
>     hydrate(queryClient, vueQueryState.value)
>   }
> ```
> 
> In CSR, `vueQueryState` is null, there is no checks anymore in vue-query `hydrate`, and it tries to access `dehydratedState.mutations` leading to a throw.
> 
> So as I see @coderabbitai was right above.
> 
> Can you advise how we can avoid the regression?

I think the docs were / are wrong and should’ve always been:

```ts
if (import.meta.client && vueQueryState.value !== null) {
  hydrate(queryClient, vueQueryState.value)
}
```

You should be getting a type error now because it’s not allowed to pass `null` to `hydrate`

I’d appreciate a PR to update the docs 🙏

PR #11266, opened 2026-08-23
docs(query-core): document cache subscription events

spaansba:
## Motivation

I wanted to detect when a paused mutation resumed. While looking into it (thanks opus), I found that this is possible through an `updated` event with a `continue` action.

The cache subscription docs currently show very little about the available event types, actions, and their payloads. I am not sure whether tables are the best way to present this information, or whether these action details are intended to be part of the documented public API in the first place. But I could really have used them when looking into this. 

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Documentation**
  * Expanded `mutationCache.subscribe` documentation with event types, event-specific properties, mutation action types, and paused-mutation resume examples.
  * Updated the callback example to require a mutation cache event.
  * Expanded `queryCache.subscribe` documentation with cache and observer events, action types, and paused-query resume examples.


TkDodo:
we are working on generating the reference doc from jsdocs, so we’re not accepting any manual reference docs changes atm.

PR #11277, opened 2026-08-24
fix(query-core): align partialMatchKey undefined handling with hashKey (#3741)

### LINKED ISSUE #3741: Query keys with object property set to undefined are not considered equal to missing property during invalidation

adityachaudhary99:
Closes #3741

hashKey drops undefined-valued object properties (JSON.stringify semantics), but partialMatchKey compared them literally - so invalidating with ['todos', {status: undefined}] failed to match queries keyed like ['todos', {status: 'open'}] even though they hash identically to ['todos', {}].

partialMatchKey now skips filter-side undefined props during object matching, aligning it exactly with documented hashing behavior. Arrays and concrete-value negatives unchanged; direction already covered by merged #11013 keeps passing. Note: open perf PR #11073 touches hashKey functions in the same file - different hunks, low conflict risk.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved partial query-key matching when filters include properties set to `undefined`.
  - Query invalidation now correctly matches cached queries with corresponding concrete values while excluding unrelated query shapes.
  - Matching behavior is consistent for nested filter objects.

### DIFF
--- a/.changeset/partial-match-key-undefined.md
+++ b/.changeset/partial-match-key-undefined.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/query-core': patch
+---
+
+fix(query-core): ignore `undefined` properties in partial-match filter keys so that `partialMatchKey` is consistent with the default `hashKey`, which drops them. Keys like `['todos', { status: undefined }]` now partially match queries with a concrete value for that property, e.g. `['todos', { status: 'open' }]`
--- a/packages/query-core/src/__tests__/queryClient.test.tsx
+++ b/packages/query-core/src/__tests__/queryClient.test.tsx
@@ -2547,6 +2547,30 @@ describe('queryClient', () => {
       expect(queryFn).toHaveBeenCalledTimes(1)
       unsubscribe()
     })
+
+    it('should match queries when a filter key property is undefined but the query key has a concrete value', async () => {
+      const key = queryKey()
+      await queryClient.fetchQuery({
+        queryKey: [key, { status: 'open' }],
+        queryFn: () => Promise.resolve('open'),
+      })
+      await queryClient.fetchQuery({
+        queryKey: [key, { assignee: 'a' }],
+        queryFn: () => Promise.resolve('assigned'),
+      })
+
+      await queryClient.invalidateQueries({
+        queryKey: [key, { status: undefined }],
+        refetchType: 'none',
+      })
+
+      expect(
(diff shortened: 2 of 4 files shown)


MILLERMARRU (commented):
Traced `partialMatchKey` and the change is consistent with `hashKey`: `hashKey` serializes through `JSON.stringify` with a sorted replacer, so `{ status: undefined }` hashes identically to `{}`, and skipping `undefined` filter-side props during object matching lines the two up. The directional asymmetry (filter-side `undefined` ignored, query-key-side `undefined` still strict) is the right choice for invalidation filters and the three added util tests pin it down well.

Two things worth a note in the code or tests:

1. The skip only applies inside the object branch. An `undefined` element inside a filter *array* (`invalidateQueries({ queryKey: [key, [1, undefined]] })`) still compares strictly. That is arguably correct since `JSON.stringify` turns array holes/undefined into `null` rather than dropping them, but it is an easy thing for a caller to expect symmetry on, so a one-line test asserting the array case stays strict would document the boundary.

2. A nested filter object whose values are all `undefined` (`[key, { filters: { status: undefined } }]`) now recurses into an empty key set and returns `true` for any query whose `filters` is any non-null object. That is the logical consequence of aligning with `hashKey` (`{filters:{status:undefined}}` hashes as `{filters:{}}`), but it is a broader match than the flat case and not covered by a test.

`#11073` touching `hashKey` in the same file is a separate hunk, agreed, low conflict risk.

PR #11292, opened 2026-08-25
feat(solid-query): type useQuery/useInfiniteQuery data as non-nullable

brenelz:
## Summary

In the Solid 2.0 port, reading `data` on a `useQuery`/`useInfiniteQuery` result is backed by an async resource: while the query is loading, the component suspends into the nearest `<Loading>` boundary, so by the time `data` is actually read during render the value has settled. The types didn't reflect that — `data` was still `TData | undefined`, forcing guards that the runtime makes unnecessary.

This PR makes `data` non-nullable on the public result types:

- Adds a distributive `NonNullableData<TResult, TData>` wrapper in `types.ts` and applies it to `UseBaseQueryResult` (and therefore `UseQueryResult`) and `UseInfiniteQueryResult`. Each status variant of the result union keeps its other discriminants (`status`, `error`, ...); only `data` is narrowed.
- `useQueries` results are a plain reactive store with no resource backing — reads never suspend — so its `data` stays honestly nullable via a local alias.

No runtime changes.

## Test updates

- Type tests updated to the new non-nullable expectations (`useQuery.test-d`, `useInfiniteQuery.test-d`, `queryOptions.test-d`, `infiniteQueryOptions.test-d`). The maybe-undefined `initialData` function case keeps `| undefined`, since there the undefined comes from `TData` itself through the defined-initialData overload (matching React Query).
- `useQueries.test-d` now asserts against the raw query-core `QueryObserverResult`, since `useQueries` keeps nullable data.
- Runtime tests that deliberately observe pending/error states read `data` before the value exists at runtime; those reads are widened through a new `pendingData` test helper in `__tests__/utils.tsx`, which documents the mismatch instead of tripping `no-unnecessary-condition` or losing runtime-necessary guards.

## Verification

- `tsc --build` clean
- `eslint ./src` clean
- `vitest run`: 22 files, 329 tests passed

### DIFF
--- a/packages/solid-query-persist-client/src/__tests__/PersistQueryClientProvider.test.tsx
+++ b/packages/solid-query-persist-client/src/__tests__/PersistQueryClientProvider.test.tsx
@@ -364,7 +364,7 @@ describe('PersistQueryClientProvider', () => {
 
       return (
         <div>
-          <h1>data: {state.data ?? 'null'}</h1>
+          <h1>data: {state.data}</h1>
           <h2>fetchStatus: {state.fetchStatus}</h2>
         </div>
       )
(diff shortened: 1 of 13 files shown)


brenelz:
Superceeded by this pr
https://github.com/TanStack/query/pull/11308

PR #11305, opened 2026-08-26
fix(react-query): propagate falsy errors to the error boundary

### LINKED ISSUE #11304: useQueries and useSuspenseQueries don't throw falsy errors to the error boundary
### Describe the bug

When a `queryFn` rejects with a falsy value (`Promise.reject()`, `null`, `''`, `0`), `useQuery` throws it to the nearest error boundary, but `useQueries` and `useSuspenseQueries` do not.

`useSuspenseQueries` is the worst case: it renders with `data === undefined`, even though its types declare `data` as always defined — so user code dereferences `undefined` and fails with an unrelated `TypeError`.

`useQueries` guards its throw with `if (firstSingleResultWhichShouldThrow?.error)`. The optional chain is needed to narrow `Array.prototype.find`'s `T | undefined`, but the truthiness check on `.error` also skips the throw when the error *value* is falsy. `useBaseQuery` has no equivalent check.

https://stackblitz.com/edit/github-xu81gzmb?file=src%2Findex.tsx

### Steps to reproduce

1. Open the StackBlitz link.
2. Each of the three hooks rejects with `Promise.reject()` (i.e. `undefined`) inside its own `ErrorBoundary`.
3. Only the `useQuery` boundary renders. The `useQueries` and `useSuspenseQueries` boundaries do not.

### Expected behavior

All three hooks should throw to their error boundary, as `useQuery` does.

Given identical options and an identical rejection:

| hook | truthy error | falsy error |
| --- | --- | --- |
| `useQuery` + `throwOnError` | boundary ✅ | boundary ✅ |
| `useQueries` + `throwOnError` | boundary ✅ | never throws ❌ |
| `useSuspenseQueries` | boundary ✅ | renders `data === undefined` ❌ |

Every time

### Platform

- OS: macOS
- Browser: any (reproduces in the StackBlitz sandbox)

### Tanstack Query adapter

react-query

### TanStack Query version

v5.102.5

### TypeScript version

v5.8.3

### Additional context

Happy to open a PR — I have a one-line fix and regression tests locally.

alex-js-ltd:
Fixes #11304

## 🎯 Changes

A `queryFn` that rejects with a falsy value (`Promise.reject()`, `null`, `''`, `0`)
reaches the error boundary from `useQuery`, but not from `useQueries` or
`useSuspenseQueries`.

`useQueries` guards its render-phase throw with a truthiness check on the error
*value*:

```diff
-if (firstSingleResultWhichShouldThrow?.error) {
+if (firstSingleResultWhichShouldThrow) {
   throw firstSingleResultWhichShouldThrow.error
 }
```

The optional chain is needed to narrow `Array.prototype.find`'s `T | undefined`,
but `?.error` also skips the throw when the error itself is falsy.
`getHasError()` has already decided the result should throw — it gates on
`result.isError` — so re-checking the value second-guesses it. `useBaseQuery`
has no equivalent check.

`useSuspenseQueries` delegates to `useQueries`, so it was affected too.

### Tests

`useQuery` — reference behaviour, passes before and after:

```tsx
useQuery({
  queryKey: key,
  queryFn: () => Promise.reject(), // rejects with `undefined`
  retry: false,
  throwOnError: true,
})
// ✅ error boundary renders
```

`useQueries` — failed before the fix:

```tsx
useQueries({
  queries: [
    {
      queryKey: key,
      queryFn: () => Promise.reject(),
      retry: false,
      throwOnError: true,
    },
  ],
})
// ❌ was: rendered normally, boundary never appeared
```

`useSuspenseQueries` — failed before the fix:

```tsx
const [query] = useSuspenseQueries({
  queries: [
    { queryKey: key, queryFn: () => sleep(10).then(() => Promise.reject()), retry: false },
  ],
})
return <div>data: {String(query.data)}</div>
// ❌ was: rendered `data: undefined`, though `data` is typed as always defined
```

Reproduction: https://stackblitz.com/edit/github-xu81gzmb?file=src%2Findex.tsx

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed error boundary handling for `useQuery`, `useQueries`, and `useSuspenseQueries` when queries reject with falsy error values.
  * Prevented suspense queries from suspending indefinitely in these cases.

* **Tests**
  * Added regression coverage for falsy query errors and error boundary behavior.

### DIFF
--- a/.changeset/olive-donuts-shave.md
+++ b/.changeset/olive-donuts-shave.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/react-query': patch
+---
+
+fix(react-query): throw falsy errors from `useQueries` and `useSuspenseQueries` to the error boundary
--- a/packages/react-query/src/__tests__/useQueries.test.tsx
+++ b/packages/react-query/src/__tests__/useQueries.test.tsx
@@ -273,6 +273,39 @@ describe('useQueries', () => {
     consoleMock.mockRestore()
   })
 
+  it("should throw error if in one of queries' queryFn rejects with a falsy error and throwOnError is in use", async () => {
+    const consoleMock = vi
+      .spyOn(console, 'error')
+      .mockImplementation(() => undefined)
+    const key = queryKey()
+
+    function Page() {
+      useQueries({
+        queries: [
+          {
+            queryKey: key,
+            queryFn: () => Promise.reject(),
+            retry: false,
+            throwOnError: true,
+          },
+        ],
+      })
+
+      return null
+    }
+
+    const rendered = renderWithClient(
+      queryClient,
+      <ErrorBoundary fallbackRender={() => <div>error boundary</div>}>
+        <Page />
+      </ErrorBoundary>,
+    )
+
+    await vi.advanceTimersByTimeAsync(0)
+    expect(rendered.getByText('error boundary')).toBeInTheDocument()
+    consoleMock.mockRestore()
+  })
+
   it('should use provided custom queryClient', async () => {
     const key = queryKey()
     const queryFn = async () => {
--- a/packages/react-query/src/__tests__/useQuery.test.tsx
+++ b/packages/react-query/src/__tests__/useQuery.test.tsx
@@ -2773,6 +2773,35 @@ describe('useQuery', () => {
     consoleMock.mockRestore()
   })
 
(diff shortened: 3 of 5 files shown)


TkDodo (approved):
thank you

TkDodo:
but also: `error` should be of type `Error`. I know you can reject with _anything_ but you probably shouldn’T.

PR #11308, opened 2026-08-26
feat(solid-query): rewrite the adapter onto Solid 2.0's native async model

ryansolid:
Target branch: `solid-query-v6-pre` (`@tanstack/solid-query@6.x`, Solid 2)

> **Status: ready to merge.** All dependencies are published (solid-js
> 2.0.0-rc.4 carries the hydration fix this leans on), CI is green against
> the released packages, and the fullstack template has been re-validated
> end-to-end (build, streaming SSR, hydration, single-flight mutations in
> a real browser) on rc.4.

## What this is

A ground-up rewrite of the Solid Query adapter onto Solid 2.0's native
async model. The v6-pre adapter still mirrors the React adapter's
architecture — observer notifications driving a store, tracking proxies,
manually-thrown `NotReadyError`s to mimic suspension — which fights the
framework: Solid 2 components don't re-render, its reactive graph already
knows how to suspend on a pending promise, hold previous values through
refetches, route rejections to error boundaries, and serialize settled
async values for streaming hydration. The rewrite hands those jobs to the
framework and keeps TanStack Query for what the framework does not do:
the cache, key identity, staleness/GC policy, retry/dedupe/cancellation,
devtools, and the ecosystem contract.

The result deletes the notification/store layer entirely:
`createStore`/`reconcile`, tracking proxies, `notifyManager` batching, and
observer-notification coupling are gone, replaced by
memo/projection/signal/action/optimistic primitives. `QueryObserver` and
`MutationObserver` survive only as lifecycle/policy engines (GC pinning,
mount-fetch policy, refetch intervals, focus/reconnect) with noop
listeners.

### Reads (`useBaseQuery`)

One memo is the data node: it returns either a settled value or the live
fetch promise, and the engine does the rest — first loads suspend into
`<Loading>`, refetches hold the previous committed value, rejections
surface to `<Errored>`, and on the server the settled value serializes
into the hydration payload.

A framing note: this one `useQuery` subsumes what React ships as three
surfaces — `useQuery`, the `useSuspenseQuery` family, and the
experimental `promise` field for `use()` interop. React's suspension is
component-scoped (the hook throws, the whole component unwinds), so its
suspense variant must forbid `enabled` and `placeholderData` and exists
as a parallel hook family. Solid's suspension is read-scoped: `q.data`
suspends while `q.status` beside it does not, so disabled queries,
placeholders, and suspense coexist in a single hook with the
`useSuspenseQuery` data contract (`data: TData`, non-optional).

- Subscription model: one version signal per hook, bumped by cache events
  for the query's hash (committed OR latest-computed hash — they diverge
  during a held transition, and settle events for a new key land before
  the hold commits). Every derived read pulls fresh state from the cache
  through it.
- Pull model: an enabled pending-idle query is fetched by the tracked
  read itself (the router `query()` model). This is what revives queries
  whose enabling flips while parked under a suspended boundary, where
  effects cannot run. `q.fetch` dedupes against policy-driven fetches;
  observer options are synced before pulling so the deferred options
  effect sees a no-op diff.
- The data node IS a single auto-reconciling store projection, identical
  on both sides (the Solid Router "feed `query()` into `createProjection`"
  shape): the derive asks the cache for the answer and the engine owns
  everything else — the server runs it, waits, and serializes the settled
  value (`deferStream` holds the flush); pending reads suspend into
  `<Loading>`; a hydrating client adopts the serialized value with
  default ('server') semantics — server truth owns the DOM for the whole
  hydration window, and cache writes that land mid-stream commit when the
  stream closes via Solid's hydration-end divergence takeover (requires
  solid-js >= 2.0.0-rc.4, where a latched node's mid-stream divergence is
  deferred rather than lost). Deep reads are fine-grained (a
  component reading `data[0].name` re-runs only when that leaf changes),
  and every landing — refetch, invalidation, placeholder upgrade —
  reconciles into the existing proxy graph instead of replacing it, so
  item identity survives across fetches. The identity key is the
  `reconcile` option (default `'id'`, function or `null` for positional).
  Primitives ride a wrapper leaf. A per-hook `primed` gate keeps the
  pull-model fetch from racing hydration: until the hook's hash-keyed
  registry entry has primed the cache (or was found absent), the read
  parks instead of fetching data the stream already carries.
- Scalar metadata is a `createProjection` reconciled per-field from query
  state — a component reading only `isFetching` re-runs on fetchStatus
  flips, not every state transition.
- `isFetching` is the OR of two channels: committed `fetchStatus` (first
  loads, imperative reads) and the engine's pending probe (refetches
  inside held transitions, where committed state cannot flip mid-hold by
  design).
- Stable promise identity across recomputes (`chainOnce`), so mid-flight
  version bumps don't hand the engine a new pending promise per recompute.
- A reactive `queryClient` accessor is honored mid-life: the version
  subscription and observer are re-pointed when the client changes.
- `refetch()` syncs the observer to the latest computed options at call
  time — under a held transition the setOptions render effect is
  deferred, and an imperative refetch must target what the UI is
  currently asking for.

### SSR + streaming hydration

Cache transfer is content-addressed — the Solid Router `query()` pattern,
one hydration story, no adapter-owned channel. On the server the provider
serializes every query the request touches into Solid's hydration
registry under `sq:<queryHash>` (`{ data, t: dataUpdatedAt }`, raw
pre-select data). Serialization happens at fetch-DISPATCH time, while the
request's serialization context is live, by handing seroval the fetch
promise; the payload streams whenever the fetch settles, and entries the
data nodes also serialized are emitted once (seroval cross-reference
dedupe). On the client each hook looks its own hash up in the registry,
reconstructs the cache entry through query-core `hydrate()` — staleness
policy intact — and only then attaches its observer. Settled entries
prime synchronously at setup; streamed entries prime as their chunk
lands, per query, so early-flushed components go live while later
boundaries still stream.

Content addressing means coverage is the cache's, not the rendered
tree's: a query prefetched in a loader and never read by any component
still transfers, and — like the router — the registry outlives the
hydration window, so a component mounted long after (lazy route,
post-hydration navigation) adopts the server's payload instead of
refetching. Entries are consumed one-shot. External `hydrate()` hosts
(the TanStack Start pattern) coexist: re-priming the same entry is silent
under newer-wins.

Contracts pinned by the SSR fixture tests:

- **Settled-only serialization**: boundaries guarantee only settled state
  serializes, and every exposed field honors it. Server meta reads of an
  unsettled enabled query tie themselves to the data node and suspend;
  `isFetching`/`isPending` can never serialize `true`; disabled-pending
  serializes as-is ('pending' IS its settled truth).
- **Hydration id parity**: every reactive node a hook creates exists on
  both server and client (some inert), because hydration id assignment is
  positional.
- **Fetch gates**: no fetch is ever started inside the hydration window
  or while a persister is restoring; values come from adoption or node
  priming, and the post-attach observer issues any genuinely-needed fetch
  after the window closes.
- **Cross-cache aggregates** (`useIsFetching`, `useIsMutating`,
  `useMutationState`): serialize their empty value (a hydrating client
  can only observe 0/[] at claim time) and stay latched through the
  hydration window, going live from the first post-hydration effect.

### Mutations (`useMutation`)

Rides Solid core's `action` primitive: one transaction across the async
gap, optimistic overlays with automatic rollback, and atomic settle.
Lifecycle callbacks are stripped from core's `Mutation.execute` and
re-run inside the transaction post-yield, preserving query-core's
callback-rejection semantics (success-path callback failures fail the
mutation; error-path failures report unhandled without displacing the
original error). Invalidation-triggered refetches hold the settle
transition until fresh data lands — mutation settle and cache refresh
commit together.

`mutate` returns the mutation promise; `mutateAsync` is removed (there is
no longer a fire-and-forget vs promise split to justify two entry
points).

### The rest

- `useQueries`: N base-query layers under `repeat` (positional identity),
  combined through a memo; the queries array is reactive in length and
  content.
- `useInfiniteQuery`: same read layer with `_type: 'infinite'` stamped so
  reactive key changes keep `infiniteQueryBehavior`; pagers are memos over
  the same entry.
- Counting hooks share one internal helper (`createCacheAggregate`):
  a durable signal + optimistic override (cache events firing inside an
  action transaction stay observable mid-flight; out-of-transaction
  writes stay durable) with the hydration latch described above.

## Surface changes (breaking)

- `data` is non-optional on query results: reads suspend (or throw to the
  boundary) instead of ever returning `undefined`. The `<Show
  when={q.data}>` loading-guard idiom is dead code now — JSX reads reach
  parity with plain Solid async reads.
- `mutateAsync` removed; `mutate` returns a promise.
- `reconcile` option re-shaped: it is now the reconciliation KEY for the
  data store (`string | (item => any) | null`, default `'id'`) — Solid 2's
  projections auto-reconcile their returns, so the v5 mechanism (manual
  `reconcile()` merges into an adapter-owned store, with a
  `structuredClone` seed hack) is core behavior now. The v5 function form
  `(oldData, newData) => newData` is gone; core `structuralSharing` still
  covers cache-level referential stability underneath.
- `suspense` option removed: suspension is the model, and boundaries are
  the control point.
- The `create*` runtime aliases (`createQuery`, `createMutation`,
  `createQueries`, `createInfiniteQuery`, `createIsFetching`,
  `createIsMutating`, `createMutationState`) and the `Create*` type
  aliases are removed, completing the deprecation from #8950 — `use*` is
  the only naming, matching the docs.
- `isInitialLoading` removed (deprecated alias of `isLoading` in core;
  its getter was byte-identical).
- `deferStream` is now actually implemented — on the v6 line it was
  declared in the types but wired to nothing. It passes straight through
  to Solid 2's per-computation `deferStream` option on the data node: the
  server holds the stream flush until the query resolves instead of
  letting the surrounding boundary render its fallback (engine behavior,
  covered by solid core's stream tests; ignored on the client).
- `placeholderData` keeps React semantics (client-progressive, no server
  fetch, `isPlaceholderData` provenance flag, masked 'success' status)
  and composes with SSR: the placeholder serializes, the client shows it
  through hydration and fetches for real after its window closes.

## Intended divergences (documented in-tests)

- A cancelled first load cannot stay idle while an enabled query is
  actively read: the pull model re-demands the value (`it.skip` with
  rationale in `useQuery.test.tsx`). Unmount, disable, or remove the
  query to stop it.
- Mid-flight mutation state (`isPaused`, `failureCount`) inside an action
  transaction holds its pre-flight face until settle — held updates
  commit atomically. The mutation cache remains imperatively inspectable
  mid-flight.

## Verification

- Full package suite green, vitest typecheck and eslint clean, `tsc
  --build` clean. Suites were ported test-by-test with per-file porting
  notes; tests asserting store/observer mechanics (render counts,
  notification batching) were dropped, semantics tests were kept or
  rewritten to DOM-level assertions.
- New semantics suites for reads, mutations, queries-array, and infinite
  queries; SSR fixture harness (real vite server/client builds, jsdom
  `hydrate()` against real server HTML + streamed payload) covering
  string SSR, streaming with held-open boundaries, fully-buffered late
  hydration, prefetched-never-rendered transfer with post-hydration
  adoption, external-`hydrate()` coexistence (TanStack Start pattern),
  and the settled-only serialization contracts.
- Template acceptance: the Solid 2.0 fullstack TanStack template
  (solid-js rc.0 through rc.4, production SSR server) with this adapter
  swapped in — suite green, hydration clean, and single-flight mutations
  verified end-to-end in a browser (login/rename/logout flows, mutation
  responses refreshing multiple queries with zero invalidation calls and
  zero extra fetches).
- Size: comment-stripped source is ~1,770 lines — smaller than the React
  adapter's core (~1,850), which additionally ships its suspense hook
  surface and Next.js-only streaming SSR as separate packages; this
  adapter's streaming hydration is included and framework-agnostic. The
  primitive inventory is strictly simpler.

## Notes

- Requires solid-js >= 2.0.0-rc.4 (peer dep updated) for one core piece
  this PR leans on: the hydration-end divergence takeover (a latched
  node's mid-stream cache write commits when the stream closes instead of
  being lost). rc.4 is published and the full suite runs green against it.
  The adapter performs no server reactive writes, per rc.3's deprecation.
- The devtools package is untouched; a follow-up could surface the
  action-transaction view.
- `useQueries` rows ride the same per-hook node priming automatically
  (each row is a full base-query layer).


brenelz:
Looks like some tests failed. The main query one passed though. Just some surrounding ones
https://cloud.nx.app/runs/DjhRqLThLP?utm_source=pull-request&utm_medium=comment

brenelz:
I tested this in the fullstack-tanstack template and seems to function better

ryansolid:
This is now ready to merge: solid-js 2.0.0-rc.4 shipped with the hydration-end divergence takeover this PR relies on, the peer range points at it, and CI is green against the published packages (no locally-built deps anywhere). Also re-validated the Solid 2.0 fullstack TanStack template end-to-end on rc.4 + the latest vite plugin — build, streaming SSR with the content-addressed cache payloads, clean hydration, and single-flight mutations (login/rename/logout) in a real browser.

brenelz (approved):

brenelz:
Used the pkg.new link to verify Tanstack router all still passing.

image

birkskyum (approved):

PR #11324, opened 2026-08-29
fix(solid-query): untrack the client read in useMutation's cache subscription

brenelz:
## Problem

Any component that calls `useMutation` triggers Solid's `STRICT_READ_UNTRACKED` dev diagnostic on mount:

```
[STRICT_READ_UNTRACKED] Reactive value read directly in <Page> will not update. Move it into a tracking scope (JSX, a memo, or an effect's compute function).
```

The hook sets up its mutation-cache subscription in the component body with a bare `client()` read — `client` is a `createMemo`, so this is a reactive read outside any tracking scope.

## Fix

Wrap the read in `untrack`, matching how `useBaseQuery` guards the same body-scope read (`untrack(client)`). The subscription is intentionally a one-time setup, so a snapshot read is the correct semantics.

The other cache-reading hooks (`useIsMutating`, `useMutationState`, `useIsFetching`) are unaffected: `createCacheAggregate` already wraps their `subscribe`/`read` callbacks in `untrack` internally.

## Test

Added a regression test that renders a bare `useMutation` component and asserts no `STRICT_READ_UNTRACKED` warning is emitted. It fails without the fix and passes with it. Full solid-query suite passes (26 files, 345 tests, no type errors).

### DIFF
--- a/packages/solid-query/src/__tests__/useMutation.test.tsx
+++ b/packages/solid-query/src/__tests__/useMutation.test.tsx
@@ -105,6 +105,30 @@ describe('useMutation', () => {
     consoleMock.mockRestore()
   })
 
+  it('should not emit a strict-read diagnostic on mount', () => {
+    const consoleMock = vi
+      .spyOn(console, 'warn')
+      .mockImplementation(() => undefined)
+
+    function Page() {
+      const mutation = useMutation(() => ({
+        mutationFn: () => Promise.resolve('mutation'),
+      }))
+
+      return <button disabled={mutation.isPending}>mutate</button>
+    }
+
+    renderWithClient(queryClient, () => <Page />)
+
+    expect(
+      consoleMock.mock.calls.filter((args) =>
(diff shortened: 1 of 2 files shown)


brenelz:
@ryansolid does this make sense?

ryansolid:
Good catch — the diagnostic is real, and your read of the semantics (one-time setup snapshot, matching `useBaseQuery`) is fair. But digging into it, the warning was pointing at something structural: that subscription only does anything between mutate and settle (`activeMutation` is null otherwise), yet it reads the client at mount and stays attached for the hook's whole life. It also snapshots a different client than `mutate` builds against, so a `QueryClient` swap could strand the listener on the old cache.

I put up #11325 which removes the mount-time read instead of untracking it — the subscription becomes per-flight, created next to the `build()` call on the same client, torn down at settle. Your regression test carries over verbatim (it's what pins the fix), plus a listener-lifecycle test. Closing this in favor of that one, but the find and the test are yours — thanks!

Issue #8345, opened 2024-11-24
Suspense causes application to hang in SSR context when query is disabled


genu:
### Describe the bug

When a query is disabled, calling `suspense()` in SSR context causes hang.

### Your minimal, reproducible example

https://stackblitz.com/edit/tanstack-query-5ck7qt?file=app.vue
### Steps to reproduce

```ts
const fetcher = async () =>
  await fetch('https://jsonplaceholder.typicode.com/posts').then((response) =>
    response.json()
  );

const { data, suspense } = useQuery({
  queryKey: ['test'],
  queryFn: fetcher,
  enabled: false,
});

onServerPrefetch(async () => {
  await suspense(); // <-- Causes hang
});
```

### Expected behavior

When query is disabled, I expect `suspense()` not to hang the application

### How often does this bug happen?

Every time

### Platform

Universal

### Tanstack Query adapter

vue-query

### TanStack Query version

5.61.3

TkDodo:
I’m not sure - I think with this, you’re awaiting a promise that never resolves. But what would you expect to happen?

- We can’t resolve with `undefined`, because that doesn’t match the shape of what `fetcher` returns
- We could throw an error, but that would trigger retries and get the query into error state. Not sure if that’s better

DamianOsipiuk:
@genu This is expected behavior as mentioned multiple times before - https://github.com/TanStack/query/issues/7609

If we would want to change this it would be a breaking change, so the earliest we could change it is the next `major`.

However, current behavior can work in two cases, when:
- you want to actually wait for `enabled` to become `true` on the server
- skip `suspense` if it's disabled, by conditionally calling `suspense`

genu:
Thanks for the explanation @TkDodo and @DamianOsipiuk It makes sense.

The issue though is that it isn't consistent with the non-SSR behavior. If we move the suspense outside of the `onServerPrefetch` it doesn't cause a hang.

At the very least maybe indicate with a console warning that query is disabled? 🤔 The application simply hanging due to this makes problem obscure.

PR #11349, opened 2026-08-31
fix(solid-query): commit the idle read of a disabled query on the server

### LINKED ISSUE #11348: solid-query 6.0.0-rc.1: `useQuery({ enabled: false })` hangs SSR again (regression of #10907)
### Describe the bug

A `useQuery` with `enabled: false`, whose `.data` is read during render, stops an SSR render from
ever finishing at `@tanstack/solid-query` 6.0.0-rc.1. Nothing is emitted at all — the render is
stuck before any output, not stalled mid-stream.

This is the same user-visible failure as #10907, which was closed with
"experimental_prefetchInRender has been removed". That did fix the v5 path; the v6 implementation
is different, so I am filing separately rather than commenting there.

**Where it comes from in rc.1.** In
[packages/solid-query/src/useBaseQuery.ts](https://github.com/TanStack/query/blob/2222f61f9b718e41d53a2eabc8b81999b248c04e/packages/solid-query/src/useBaseQuery.ts),
`computeData()` — the derive of the single `createProjection` that is the data node — ends:

```ts
if (state.data !== undefined) return wrap(state.data)

// Pending-idle: nothing in flight, nothing cached. …
if (isEnabled()) {
	…
	return chainOnce(q.fetch(opts as any), select, wrap)
}
return NEVER
```

with
[`const NEVER: Promise<never> = new Promise(noop)`](https://github.com/TanStack/query/blob/2222f61f9b718e41d53a2eabc8b81999b248c04e/packages/solid-query/src/useBaseQuery.ts#L38).
So a disabled query with no cached data hands the `createProjection` a promise that can never
settle.

Per its docblock
([L30-L38](https://github.com/TanStack/query/blob/2222f61f9b718e41d53a2eabc8b81999b248c04e/packages/solid-query/src/useBaseQuery.ts#L30-L38))
that parking is intentional client behaviour: it suspends the reader into the nearest `<Loading>`
until the query starts fetching, "at which point the version bump re-runs the compute". On the
server there is no later — the render has to finish, and nothing will enable the query or write
the cache before it does. The three earlier `return NEVER` guards are each gated
([`!isServer && hydrating`](https://github.com/TanStack/query/blob/2222f61f9b718e41d53a2eabc8b81999b248c04e/packages/solid-query/src/useBaseQuery.ts#L497),
[`!primed()`](https://github.com/TanStack/query/blob/2222f61f9b718e41d53a2eabc8b81999b248c04e/packages/solid-query/src/useBaseQuery.ts#L509),
[`isRestoring()`](https://github.com/TanStack/query/blob/2222f61f9b718e41d53a2eabc8b81999b248c04e/packages/solid-query/src/useBaseQuery.ts#L519))
and `observer.setOptions` is gated `!isServer`, but the final disabled
[`return NEVER`](https://github.com/TanStack/query/blob/2222f61f9b718e41d53a2eabc8b81999b248c04e/packages/solid-query/src/useBaseQuery.ts#L534)
has no `isServer` branch. rc.0's `useBaseQuery` carries several `if (isServer)` special-cases by
contrast.

I have not proposed a patch — what a server read of a disabled query should commit, and where the
guard belongs, looks like your call.

(Permalinks are pinned to the `@tanstack/solid-query@6.0.0-rc.1` tag, commit `2222f61f9b71`.)

<https://github.com/TylerRick/solid-query-disabled-query-ssr-hang-repro>

A git repo rather than a sandbox, because the bug only appears in a server render — it is four
files, `pnpm install` and one `node` command. Pure Solid SSR: `renderToStream`, one
`QueryClientProvider`, one `useQuery`. No router, no solid-start, no metaframework, no `<Loading>`
boundary.

```tsx
import { QueryClient, QueryClientProvider, useQuery } from '@tanstack/solid-query';

const client = new QueryClient();

function Inner() {
	const q = useQuery(() => ({
		queryKey: ['disabled'],
		queryFn: async () => 'data',
		enabled: false,
	}));
	return <div id="out">data: {String(q.data)}</div>;
}

export default function App() {
	return (
		<QueryClientProvider client={client}>
			<Inner />
		</QueryClientProvider>
	);
}
```

### Steps to reproduce

1. `git clone https://github.com/TylerRick/solid-query-disabled-query-ssr-hang-repro && cd solid-query-disabled-query-ssr-hang-repro`
2. `pnpm install`
3. `ENABLED=0 node render.mjs` — renders the component above through `@solidjs/web`'s
   `renderToStream` and reports whether the stream finishes, with an 8 second cap.
4. It reports `HANG — stream did not finish in 8s (bytes so far: 0)`.
5. `ENABLED=1 node render.mjs` — the identical code with the query enabled — reports
   `COMPLETED — 347 bytes`.
6. For the version control, set `@tanstack/solid-query` to `6.0.0-rc.0` in `package.json` (leaving
   the `@tanstack/query-core` override at 5.101.4), `pnpm install`, and repeat step 3: it
   completes.

### Expected behavior

A disabled query has nothing to wait for, so reading `.data` during SSR should expose the idle
result and allow rendering to finish. rc.0 completes with `data: undefined`; rc.1 also completes
when the query is enabled.

Instead the render never finishes and no bytes are produced. In a metaframework the effect is that
the page never responds at all.

Every time

### Platform

- OS: Linux (x64)
- Node: 24.19.0
- Browser: n/a — this is a server render

### Tanstack Query adapter

solid-query

### TanStack Query version

dependency, pinned in every row below via a direct dependency and a `pnpm.overrides` entry, so the
query-core bump that normally rides along with rc.1 is held constant).

### TypeScript version

n/a for this bug — the reproduction's `.tsx` is transpiled by vite and never typechecked.

### Additional context

The three-cell result, all with query-core held at 5.101.4:

| solid-query | `enabled` | result                                    |
| ----------- | --------- | ----------------------------------------- |
| 6.0.0-rc.1  | `false`   | **hang** — stream never finishes, 0 bytes |
| 6.0.0-rc.1  | `true`    | completes                                 |
| 6.0.0-rc.0  | `false`   | completes                                 |

Full environment: `solid-js` / `@solidjs/web` 2.0.0-rc.4, `@solidjs/vite-plugin` 3.0.0-next.35
with `@solidjs/compiler` / `@solidjs/babel-plugin` 2.0.0-rc.4, vite 8.2.1.

Found upgrading an app to Solid 2.0.0-rc.4, where every page hung: a hook in the root chrome gates
its query with `enabled` on data only some routes provide, so on every other route that query is
disabled. Through TanStack Start the visible effect is that the SSR-query integration's
`queryStream` never closes, because `router.serverSsr.onRenderFinished(...)` is never reached —
but as the reproduction shows, the router is not needed and the defect is not there.

Possibly related, both closed: #10907 (the v5 original) and #8345 (`<Suspense>` hanging in SSR
with a disabled query).

Two PRs on the v5 line strip the never-resolving promise out of `hydratableObserverResult`: #10923
(closed unmerged) and #10104 (still open — its description matches this failure mode closely,
including the `enabled: false` case). Neither approach transfers to v6 as-is: rc.1's `useBaseQuery`
has no `hydratableObserverResult` and no `createResource`, and the equivalent path is the `NEVER`
returned from `computeData` into `createProjection`.

lazerg:
## 🎯 Changes

A `useQuery` with `enabled: false` and nothing cached stops an SSR render from ever finishing: no bytes are emitted at all.

In `computeData`, the pending-idle path for a disabled query returns `NEVER`, a promise that never settles. That is the right client behavior — the reader parks in the nearest `<Loading>` until an enabling change, a refetch or a cache write revives it. The server has no later: the render has to finish, and nothing will enable the query or write the cache before it does.

So the server now commits the idle read instead, which is the settled SSR truth for a disabled query and the same contract the meta channel already honors for one. Resolving with `undefined` was rejected for the v5 line in #8345 because it does not match the shape `queryFn` returns, and that objection stands on the client, where suspending costs nothing and the query may still start. On the server the alternative is not suspension but a page that never responds, so the type docblock now carries the exception. The client path is untouched.

The hydration fixture grows a disabled query in a boundary of its own, so the regression test covers both sides: the server emits the idle face, and the hydrating client adopts it without fetching. Without the fix the fixture's SSR entry never settles and the whole suite fails.

Fixes #11348

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/disabled-query-server-read.md
+++ b/.changeset/disabled-query-server-read.md
@@ -1,0 +1,12 @@
+---
+'@tanstack/solid-query': patch
+---
+
+fix: commit the idle read of a disabled query on the server. A query
+disabled with nothing cached parked its reader on a promise that never
+settles, which is the intended client behavior — an enabling change, a
+refetch or a cache write revives it. The server has no later: the render
+has to finish, and nothing will enable the query or write the cache
+before it does, so the render stalled forever and emitted nothing at all.
+The server now commits the idle read, which is the settled SSR truth for
+a disabled query and the same contract its meta channel already honors.
--- a/packages/solid-query/src/__tests__/fixtures/hydration/App.tsx
+++ b/packages/solid-query/src/__tests__/fixtures/hydration/App.tsx
@@ -21,6 +21,7 @@ export interface FetchCounts {
   stale: number
   placeholder: number
   prefetched: number
+  disabled: number
 }
 
 export interface AppProps {
@@ -100,6 +101,29 @@ function Queries(props: AppProps) {
   )
 }
 
+/** Reads the data of a query that is disabled with nothing cached. On the
+ * server that read has nothing to wait on and no later in which to get one —
+ * the render has to finish, so the idle read is its settled SSR truth and the
+ * document is emitted. (On the client the same read parks the reader in this
+ * boundary until something starts the query, which is why it lives in a
+ * boundary of its own here: the rest of the app hydrates around it.) */
+function DisabledConsumer(props: AppProps) {
+  const disabled = useQuery(() => ({
+    queryKey: ['disabled'],
+    queryFn: async () => {
+      props.counts.disabled++
+      await sleep(5)
+      return `disabled-${props.source}`
+    },
+    enabled: false,
+  }))
+  return (
+    <span id="disabled">
+      {String(disabled.data)}|{disabled.status}|{String(disabled.isEnabled)}
+    </span>
+  )
+}
+
 /** Never rendered on the server — mounted by tests after hydration. Its
  * query was prefetched (and only prefetched) during SSR; the hash-keyed
  * registry entry must satisfy it with zero client fetches. */
@@ -138,6 +162,9 @@ export function App(props: AppProps) {
         <Show when={props.lateMount?.()}>
           <LateConsumer {...props} />
         </Show>
+      </Loading>
+      <Loading fallback={<div>loading</div>}>
+        <DisabledConsumer {...props} />
       </Loading>
     </QueryClientProvider>
   )
--- a/packages/solid-query/src/__tests__/fixtures/hydration/entry-client.tsx
+++ b/packages/solid-query/src/__tests__/fixtures/hydration/entry-client.tsx
@@ -19,6 +19,7 @@ export function createApp() {
     stale: 0,
     placeholder: 0,
     prefetched: 0,
+    disabled: 0,
   }
   const [lateMount, setLateMount] = createSignal(false)
   return {
--- a/packages/solid-query/src/__tests__/fixtures/hydration/entry-server.tsx
+++ b/packages/solid-query/src/__tests__/fixtures/hydration/entry-server.tsx
@@ -15,6 +15,7 @@ const counts: FetchCounts = {
   stale: 0,
   placeholder: 0,
   prefetched: 0,
+  disabled: 0,
 }
 
 // Fully-settled single-string render. Collected through pipe() rather than
--- a/packages/solid-query/src/__tests__/hydration-utils.ts
+++ b/packages/solid-query/src/__tests__/hydration-utils.ts
@@ -36,6 +36,7 @@ export interface ServerReport {
       stale: number
       placeholder: number
       prefetched: number
+      disabled: number
     }
     queries: Array<QuerySnapshot>
   }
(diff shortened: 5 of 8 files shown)


brenelz:
Can you retest this after the rewrite?

brenelz:
I believe this is a duplicate of this pr. Please open a new pr if that is not the case

https://github.com/TanStack/query/pull/11449

Issue #11351, opened 2026-09-01
solid-query 6.0.0-rc.1: a tracked read through `createMemo(() => query.data)` re-runs on refetch but reads the previous value, and is never notified again


TylerRick:
### Describe the bug

A tracked computation that reads query data THROUGH a `createMemo(() => query.data)` indirection
re-runs after a refetch — but reads the value from before the refetch, and is never notified
again, so its last observation stays stale permanently. At 6.0.0-rc.0 the same computation reads
the new value.

Narrowing, all measured at rc.1:

- An effect reading `query.data` directly (no memo) observes the new value — the memo indirection
  is the trigger.
- The same value change committed synchronously via `queryClient.setQueryData()` is observed
  through the memo too — the failure needs the refetch path, where the data projection's derive
  returns a promise and the commit is asynchronous.
- The memo itself never recomputes across the refetch (consistent with the data node being one
  stable store root reconciled in place), and after everything settles the memo's cached value is
  `===` a direct `query.data` read — at which point an UNTRACKED read through the memo returns the
  NEW value. So the stale read happens at notify time: the effect re-runs before the new value is
  observable through the memo, and no second notification arrives after it commits.
- Two pure-solid `createProjection` mimics of the rc.1 data-node shape (a promise-returning
  derive, a memo over the root leaf, an effect reading through the memo — with and without the
  derive reading its own previous committed value) both pass on solid-js 2.0.0-rc.4, so this does
  not look like a generic solid-js memo-over-projection defect, at least not in the documented
  shape.

A memo over `query.data` is the natural shape of any wrapper hook that gates or narrows a query's
data before handing it to consumers, so this bites app code that never spells
`createMemo(() => query.data)` explicitly.

### Your minimal, reproducible example

<https://github.com/TylerRick/solid-query-stale-read-through-memo-repro>

A git repo rather than a sandbox because the repro is a vitest run — four files, `pnpm install`
and one command.

```tsx
function Probe() {
	const query = useQuery(() => ({
		queryKey: key,
		queryFn: async () => ({ ...serverValue }),
	}));
	const data = createMemo(() => query.data);
	createEffect(
		() => data()?.video1,
		(video1) => {
			captured = video1;
		},
	);
	return <div />;
}
// serverValue starts { video1: false }; after mount+settle:
// serverValue = { video1: true }; await queryClient.refetchQueries({ queryKey: key }); settle
// rc.0: effect re-runs, captured === true
[...]

PR #11357, opened 2026-09-01
test(solid-query): pin the stale read through a memo over `query.data` (#11351) 🤖🤖🤖

mgarcialeniolabs:
## 🎯 Changes

Adds a reproduction for #11351 to `useQuery-semantics.test.tsx`, in the `data store face` block.

A tracked computation that reaches a leaf **through** a `createMemo(() => query.data)` indirection
is not re-notified after a refetch, so its last observation stays the superseded value permanently.
A wrapper hook that gates or narrows `query.data` before handing it to consumers is exactly this
shape, so it reaches app code that never writes the memo explicitly.

The test is added as `it.skip`, following the existing convention in `useQuery.test.tsx` for a
documented divergence, because **the fix is not in this repo** — see below. Un-skipped it fails on
the last assertion only, *after* `expect(rendered.getByText('flag: true'))` has already passed: the
commit lands and the DOM swaps, and only the reader behind the memo is starved.

### Why no fix here

The data node is not at fault. `computeData` runs exactly the expected four times across mount and
refetch, `chainOnce` produces one chained promise per fetch with no identity churn, the commit
lands, an untracked read through the very same memo returns the new value, and neither the memo nor
the data node is left `isPending`. Only the notification to one reader shape is lost.

Narrowing across reader shapes (all against the real hook):

| Reader shape | Result |
| --- | --- |
| effect reads `query.data.flag` | pass |
| memo reads `query.data.flag`, effect reads the memo | pass |
| effect captures `query.data` via `untrack`, then reads `.flag` off it | pass |
| effect depends on a suspended memo over an **unrelated** async source, reads `query.data.flag` | pass |
| **memo reads `query.data`, effect reads `memo().flag`** | **fail** |
| memo reads `query.data`, effect reads **both** `query.data` and `memo().flag` | **fail** |

Row 6 is the one that rules out a missing subscription: the effect reads the projection directly
*and* depends on the memo, and is still not notified. Row 4 rules out "any suspended memo
dependency" — the memo has to be a consumer of *this* projection. A memo created after the first
settle is fine; only one that suspended on its first run is affected.

The behaviour reproduces with no TanStack code involved, so I filed it upstream:
**solidjs/solid#3181**. `2.0.0-rc.4` is the newest published `solid-js` / `@solidjs/signals`, so
there is no version bump that fixes it.

The workarounds I tried and rejected, in case they come up in review:

- Making the committed value's identity change per commit so the memo propagates — that is
  reconcile-off, and discards the fine-grained leaf tracking pinned by `'tracks deep reads at the
  leaf — unrelated changes do not re-run them'`.
- Returning a stable re-reading facade from `get data()` — works only if the getter drops its
  suspending read, which would stop `query.data` suspending into `<Loading>` and break primitive
  and array data.
- Not returning the promise on refetch, to avoid the hold — deletes the SWR hold semantics covered
  by `transition.test.tsx`.
- Not superseding the in-flight answer with a synchronous value at settle (a `landed` flag on
  `chainOnce`, returning the unlanded chained promise instead of `wrap(state.data)`) — **tried and
  reverted**: it does not fix the reproduction and it breaks `'keeps stale data and exposes error
  state when a refetch fails'`.

Happy to drop the skipped test and just leave the narrowing on the issue instead, if you'd rather
not carry it.

> Note on base branch: this targets `solid-query-v6-pre`, since the rc.1 code the issue is about
> does not exist on `main`.

## ✅ Checklist

Verified locally from the repo root:

```
pnpm nx run @tanstack/solid-query:test:lib      # 26 files, 346 passed, 2 skipped, no type errors
pnpm nx run @tanstack/solid-query:test:eslint   # clean
```

Baseline on `solid-query-v6-pre` before this change is 346 passed / 1 skipped, so the only delta is
the added skipped test.

## 🚀 Release Impact

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/solid-query/src/__tests__/useQuery-semantics.test.tsx
+++ b/packages/solid-query/src/__tests__/useQuery-semantics.test.tsx
@@ -6,7 +6,13 @@
 // re-pointed separately.
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { fireEvent } from '@solidjs/testing-library'
-import { Errored, Loading, createSignal } from 'solid-js'
+import {
+  Errored,
+  Loading,
+  createEffect,
+  createMemo,
+  createSignal,
+} from 'solid-js'
 import { queryKey, sleep } from '@tanstack/query-test-utils'
 import { QueryCache, QueryClient, useQuery } from '..'
 import { renderWithClient } from './utils'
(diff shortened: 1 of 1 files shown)


brenelz (approved):

brenelz:
Looks like the issue was fixed on the Solid side.

PR #11360, opened 2026-09-01
fix(solid-query): don't re-create and refetch a removed query

### LINKED ISSUE #11350: solid-query 6.0.0-rc.1: `removeQueries()` on an actively observed query is immediately re-created and refetched
### Describe the bug

Calling `queryClient.removeQueries()` while a `useQuery` observer is mounted removes the entry and
then immediately brings it back: the cache emits `removed` followed by `added`, the observer is
torn down and re-attached, and the recreated entry starts a fresh fetch of the key that was just
removed. At 6.0.0-rc.0 the removal sticks (entry count 0, no further events, no fetch).

The practical consequence: a caller that removes entries to make them unreadable — for example an
identity/login boundary that must guarantee the previous user's data cannot be read — cannot make
the removal last. The entry is re-created immediately, and the automatic refetch repopulates the
removed key, under whatever key and queryFn the still-mounted observer holds.

Provenance controls in the repro pin down what does and does not happen at rc.1:

- The old value is NOT retained: with the post-removal refetch parked on a never-resolving
  promise, the re-added entry sits `'pending'`/`'fetching'` and `getQueryData` stays `undefined`.
- The refetch is real: a queryFn returning `'v1'` on the first call and `'v2'` afterwards ends
  with `getQueryData === 'v2'` and a fetch count of 2, with `staleTime` ruling out an ordinary
  mount refetch.

If an active observer re-creating and refetching a removed query is intended in the rc.1 model, is
there a supported way to make removal stick — or at least be observable — while observers are
mounted?

<https://github.com/TylerRick/solid-query-remove-queries-recreated-repro>

A git repo rather than a sandbox because the repro is a vitest run — three files, `pnpm install`
and one command.

```tsx
const queryClient = new QueryClient({ defaultOptions: { queries: { staleTime: 60_000 } } });
const key = ['probe'];
let fetches = 0;
const queryFn = async () => (++fetches === 1 ? 'v1' : 'v2');

await queryClient.prefetchQuery({ queryKey: key, queryFn });

function Reader() {
	const query = useQuery(() => ({ queryKey: key, queryFn }));
	return <div>{String(query.data)}</div>;
}
// render <QueryClientProvider client={queryClient}><Reader /></QueryClientProvider>, settle

await queryClient.cancelQueries();
queryClient.removeQueries();
// settle 30ms
// rc.0: cache events […, 'removed'], 0 entries, getQueryData undefined, fetches === 1
// rc.1: […, 'removed', 'added', 'observerRemoved', 'observerAdded', 'updated', …], 1 entry,
//       getQueryData 'v2', fetches === 2
```

### Steps to reproduce

1. Prefetch a key, mount one `useQuery` observing it (staleTime high enough that the mount does
   not refetch).
2. `await queryClient.cancelQueries(); queryClient.removeQueries();`
3. Wait a tick and inspect `queryClient.getQueryCache().getAll()`, `getQueryData`, and the fetch
   count.

### Expected behavior

`removeQueries()` leaves the cache without the entry, as at 6.0.0-rc.0 — or, if recreation by an
active observer is intended, at least does not silently refetch a key the caller just removed, and
gives the caller some way to observe/prevent the resurrection.

Every time

### Platform

- OS: Linux (x64)
- Node: 24.19.0
- jsdom (vitest) — client-side only, no SSR

### Tanstack Query adapter

solid-query

### TanStack Query version

dependency, pinned via a direct dependency and a `pnpm.overrides` entry, so the query-core bump
that normally rides along with rc.1 is held constant). Clean at `@tanstack/solid-query` 6.0.0-rc.0
with the same query-core pin.

### TypeScript version

n/a — the reproduction's `.tsx` is transpiled by vitest/vite-plugin-solid and never typechecked.

### Additional context

solid-js / @solidjs/web 2.0.0-rc.4, @solidjs/testing-library 1.0.0-beta.2, vitest 3.

Found upgrading an app from rc.0: an identity-change boundary that cancels and removes all
queries, then gates UI on the removal being observable, never sees the entries gone.

theRizwan:
Fixes #11350.

Targets `solid-query-v6-pre`, since the affected read layer arrived with the 2.0-native rewrite (#11308) and the regression is between `6.0.0-rc.0` and `6.0.0-rc.1`. One housekeeping note that follows from that base: GitHub only honours a closing keyword when the PR targets the default branch, so the `Fixes` above registers as a plain cross-reference and #11350 will need closing by hand on merge.

## The bug

Calling `queryClient.removeQueries()` while a `useQuery` observer is mounted removes the entry and then immediately brings it back, and the recreated entry starts a fresh fetch of the key that was just removed. A caller that removes entries to make them unreadable, for example an identity boundary that must guarantee the previous user's data cannot be read, cannot make the removal last.

## Cause

`useBaseQueryLayer` has one version signal per hook, bumped by `onCacheEvent` for any cache event on its hash. `'removed'` is such an event, and the recompute it triggers ran through `query()`:

```ts
const query = () => {
  version()
  const c = client()
  syncClient(c)
  return c.getQueryCache().build(c, defaultedOptions() as any) as any
}
```

`queryCache.build()` creates the entry when the cache does not hold it, so the hook's own removal notification made it rebuild what the caller had just deleted. The resurrection is not passive: the `'added'` entry also re-points the still-live observer, and its mount-fetch policy then refetches and repopulates the key. Hence the reported event sequence, `removed, added, observerRemoved, observerAdded, updated`.

This contradicts the documented contract for the method, that `removeQueries` "removes matching queries from the cache instead of refetching them".

## Fix

`query()` keeps the entry it last read and falls back to it when the cache no longer holds that hash, building only when the hash is genuinely new. A removal now leaves the cache empty and fires no fetch, and the mounted reader holds its last value until options change or a real entry returns through `setQueryData`, a refetch or a later mount. `cache.get()` is consulted first on every read, so a live entry is never shadowed by the held-over one. The fallback is dropped on a client swap in `syncClient`, as it is only meaningful for the cache it was read from.

Because `useInfiniteQuery` and `useQueries` both go through this layer, the one change covers all three hooks.

## Verification

The reporter's scenario, run against the react adapter on this branch to establish the reference, and against solid-query before and after the change:

| | react-query | solid-query rc.1 | solid-query, patched |
|---|---|---|---|
| cache events after removal | `removed` | `removed, added, observerRemoved, observerAdded, ...` | `removed` |
| cache entries | 0 | 1 | 0 |
| `getQueryData` | `undefined` | `'v2'` | `undefined` |
| queryFn calls | 1 | 2 | 1 |
| mounted reader shows | `v1` | `v2` | `v1` |

The patched adapter matches react-query exactly, and matches `6.0.0-rc.0` as reported.

Two regression tests in `useQuery-semantics.test.tsx`. Both fail on the unpatched branch, the first on the event sequence and the second on the entry count:

- `leaves a removed query out of the cache while an observer is mounted`, which asserts the cache event list is exactly `['removed']` with no `'added'` behind it, an empty cache, an unchanged fetch count, and that the reader holds its last value rather than suspending over a key that no longer exists.
- `picks up a real entry written after a removal`, which guards the fallback against becoming a stale read: a `setQueryData` after the removal rebuilds the entry, and the reader must track the new instance.

Full `@tanstack/solid-query` suite passes, 26 files and 348 tests, along with `@tanstack/solid-query-devtools` and `@tanstack/solid-query-persist-client`. `test:eslint` and `tsc --build` are clean.

### DIFF
--- a/packages/solid-query/src/useBaseQuery.ts
+++ b/packages/solid-query/src/useBaseQuery.ts
@@ -216,6 +216,27 @@ export function useBaseQueryLayer<
    * swap re-attaches the rebuilt observer iff its predecessor was attached. */
   let shouldAttach = false
   let activeClient = untrack(client)
+  /**
+   * The entry this hook last read at its current hash, and the fallback for
+   * `query()` when the cache no longer holds that hash. Removal is the case
+   * that needs it: `removeQueries()` (or `clear()`) fires 'removed' for a
+   * hash this hook is mounted on, which bumps `version` and re-runs every
+   * derived read, and a plain `build()` there would re-create the entry the
+   * caller just removed. That resurrection is not passive, since the 'added'
+   * entry also re-points the still-live observer, whose mount-fetch policy
+   * then refetches and repopulates the key, contrary to `removeQueries`
+   * being documented to remove entries instead of refetching them.
+   *
+   * The removed instance keeps its final state, so reads hold their last
+   * value until options change or a real entry returns (`setQueryData`, a
+   * refetch, a later mount). That is what the other adapters do: a React
+   * observer keeps rendering the removed query's last result while the
+   * cache stays empty.
+   *
+   * Cleared on a client swap, below: the fallback is only meaningful for the
+   * cache it was read from.
+   */
+  let lastQuery: Query<TQueryFnData, TError, TQueryData, TQueryKey> | undefined
 
   const attach = () => {
     if (!disposed && !observerSub && !untrack(isRestoring)) {
@@ -236,6 +257,7 @@ export function useBaseQueryLayer<
   const syncClient = (c: QueryClient) => {
     if (isServer || c === activeClient) return
     activeClient = c
+    lastQuery = undefined
     cacheSub?.()
     cacheSub = c.getQueryCache().subscribe(onCacheEvent)
     observerSub?.()
@@ -365,7 +387,14 @@ export function useBaseQueryLayer<
     version()
     const c = client()
     syncClient(c)
-    return c.getQueryCache().build(c, defaultedOptions() as any) as any
+    const cache = c.getQueryCache()
+    const opts = defaultedOptions()
+    const existing = cache.get<TQueryFnData, TError, TQueryData, TQueryKey>(
+      opts.queryHash,
+    )
+    if (existing) return (lastQuery = existing)
+    if (lastQuery?.queryHash === opts.queryHash) return lastQuery
+    return (lastQuery = cache.build(c, opts as any) as any)
   }
 
   const isEnabled = () => {
--- a/.changeset/solid-removed-query-not-rebuilt.md
+++ b/.changeset/solid-removed-query-not-rebuilt.md
@@ -1,0 +1,18 @@
+---
+'@tanstack/solid-query': patch
+---
+
+fix: stop a mounted observer from re-creating and refetching a removed
+query. The read layer bumps its per-hook version signal on every cache
+event for its hash, `removed` included, and the recompute that followed
+called `queryCache.build()`, which put the entry the caller had just
+deleted straight back. The resurrection was not passive: the rebuilt
+entry also re-pointed the still-live observer, whose mount-fetch policy
+then refetched and repopulated the key, so `removeQueries()` (and
+`clear()`) could not be made to stick while any hook observed the key.
+`query()` now reuses the entry it last read when the cache no longer
+holds that hash, and only builds when the hash is genuinely new, so a
+removal leaves the cache empty and fires no fetch, while the mounted
+reader holds its last value until options change or a real entry returns
+through `setQueryData`, a refetch or a later mount. This is the behavior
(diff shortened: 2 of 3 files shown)


packages/solid-query/src/useBaseQuery.ts
```diff
@@ -365,7 +387,14 @@ export function useBaseQueryLayer<
...
     syncClient(c)
-    return c.getQueryCache().build(c, defaultedOptions() as any) as any
+    const cache = c.getQueryCache()
+    const opts = defaultedOptions()
+    const existing = cache.get<TQueryFnData, TError, TQueryData, TQueryKey>(
+      opts.queryHash,
+    )
+    if (existing) return (lastQuery = existing)
+    if (lastQuery?.queryHash === opts.queryHash) return lastQuery
+    return (lastQuery = cache.build(c, opts as any) as any)
```
Hardanish-Singh, line 397: [nit]: avoid `any` type

brenelz (approved):

PR #11361, opened 2026-09-01
fix(solid-query): keep removed queries out of cache

### LINKED ISSUE #11350: solid-query 6.0.0-rc.1: `removeQueries()` on an actively observed query is immediately re-created and refetched

nickshiro:
## 🎯 Changes

FIX #11350. 

Prevents an actively observed query from being recreated immediately after the `removeQueries()` function removes it from the cache. `TQuery` stores the removed query as a read-only tombstone for fixed read operations. This prevents a situation where reactive cache updates call `cache.build()` for the same query hash and cause unexpected re-fetch. An explicit re-fetch, a write to the cache, a change to the query key, or change to the `QueryClient` can result in the query being recreated as expected.

Added a regression test covering both behaviors:
- removing observed query keeps it out of the cache without another fetch.
- explicitly refetching recreates the query and updates its data.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/solid-query-removed-query-tombstone.md
+++ b/.changeset/solid-query-removed-query-tombstone.md
@@ -1,0 +1,6 @@
+---
+'@tanstack/solid-query': patch
+---
+
+Keep actively observed queries removed until an explicit refetch, cache write,
+or query change recreates them.
--- a/packages/solid-query/src/__tests__/useQuery-semantics.test.tsx
+++ b/packages/solid-query/src/__tests__/useQuery-semantics.test.tsx
@@ -102,6 +102,55 @@ describe('useQuery 2.0 read semantics', () => {
     expect(rendered.getByText('data2')).toBeInTheDocument()
   })
 
+  it('does not recreate a removed query while it is observed', async () => {
+    const key = queryKey()
+    const events: Array<string> = []
+    let fetches = 0
+    let refetch!: () => Promise<unknown>
+    const queryFn = () => Promise.resolve(`data${++fetches}`)
+
+    await queryClient.prefetchQuery({
+      queryKey: key,
+      queryFn,
+      staleTime: Infinity,
+    })
+
+    function Page() {
+      const state = useQuery(() => ({
+        queryKey: key,
+        queryFn,
+        staleTime: Infinity,
+      }))
+      refetch = state.refetch
+      return <span>{state.data}</span>
+    }
+
+    const rendered = renderWithClient(queryClient, () => <Page />)
+    expect(rendered.getByText('data1')).toBeInTheDocument()
+
+    const unsubscribe = queryCache.subscribe((event) => {
+      events.push(event.type)
+    })
+
+    queryClient.removeQueries({ queryKey: key })
+    await vi.advanceTimersByTimeAsync(0)
+
+    expect(queryCache.find({ queryKey: key })).toBeUndefined()
+    expect(queryClient.getQueryData(key)).toBeUndefined()
+    expect(fetches).toBe(1)
+    expect(events).toEqual(['removed'])
+    expect(rendered.getByText('data1')).toBeInTheDocument()
+
+    await refetch()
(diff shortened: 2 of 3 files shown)


brenelz:
This looks like mostly a duplicate of https://github.com/TanStack/query/pull/11360

PR #11363, opened 2026-09-02
fix: make Algolia load more keyboard accessible

chuhuangvio-itch:
## Bug
The More action was a clickable div and could only be activated with a pointer.

## Fix
The action is a native button with focus-visible styling, so keyboard activation follows browser button behavior.

## Validation
prettier --check on both changed files and git diff --check passed. Full workspace build remains blocked by an upstream tsdown config-loader dependency error (tsx/esm/api), unrelated to this change.

Reviewed and reproduced by Codex.

## Summary by CodeRabbit

* **Accessibility**
  * Improved keyboard accessibility for the “more” pagination control by using a button with a visible focus indicator.
  * Preserved the existing pagination behavior and label.
* **Style**
  * Updated the control styling to match its previous appearance while removing default button formatting.

### DIFF
--- a/examples/react/algolia/src/SearchResults.tsx
+++ b/examples/react/algolia/src/SearchResults.tsx
@@ -59,9 +59,13 @@ export default function SearchResults({ query = '' }: SearchResultsProps) {
           )}
         </div>
         {hasNextPage && (
-          <div className="search-more" onClick={() => fetchNextPage()}>
+          <button
+            type="button"
+            className="search-more"
+            onClick={() => fetchNextPage()}
+          >
             more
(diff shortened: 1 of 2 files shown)


kvnloo:
a11y review (exact-head check on `2196d2e0ac`): Load more control change from `div` → `button type="button"` with CSS reset and `:focus-visible` looks correct for keyboard activation.

One follow-up to consider: ensure the control also has an accessible name (visible text or `aria-label`) so screen-reader users get the same “Load more” affordance as sighted users.

No competing PR from me — KEEP this head.

PR #11385, opened 2026-09-03
fix(solid-query): prevent lost updates when mounting cached queries

### LINKED ISSUE #11384: Solid Query v6 loses derived updates when mounting with warm stale cache
### Describe the bug

In 6.0.0-rc.1+, mounting a consumer for a warm, stale query starts an immediate refetch.

`query.data` contains the item, but an effect-derived signal remains empty. `rc.0` passes; the repro does not use Router.

https://stackblitz.com/~/github.com/2wheeh/solid-query-cached-mount-repro

### Steps to reproduce

1. Wait until `query.data` shows `Example item`.
2. Click `Mount`.
3. Observe that `query.data` contains the item while `Mounted consumer output` is `[]`.

### Expected behavior

`Mounted consumer output` should contain the same item as `query.data`.

Every time

### Screenshots or Videos

_No response_

### Platform

Reproduces on macOS and StackBlitz WebContainers; appears to be platform-neutral.

### Tanstack Query adapter

solid-query

### TanStack Query version

v6.0.0-rc.2

### TypeScript version

5.9.2

### Additional context

https://github.com/2wheeh/solid-query-cached-mount-repro/blob/main/README.md

2wheeh:
fixes #11384 

## 🎯 Changes

Fixes a Solid Query timing issue where effect-derived updates could be lost when a cached query refetched during mount without a `<Loading>` boundary.

Observer attachment is deferred until the conditional subtree is ready, while the optimistic fetch status remains available synchronously. A regression test covers the derived update and verifies that the mount refetch occurs.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/warm-caches-notify.md
+++ b/.changeset/warm-caches-notify.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/solid-query': patch
+---
+
+fix: prevent cached query updates from being lost during mount.
--- a/packages/solid-query/src/__tests__/useQuery-semantics.test.tsx
+++ b/packages/solid-query/src/__tests__/useQuery-semantics.test.tsx
@@ -6,7 +6,14 @@
 // re-pointed separately.
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { fireEvent } from '@solidjs/testing-library'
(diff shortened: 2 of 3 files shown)


2wheeh:
fixed from the upstream

Issue #9457, opened 2025-07-18
Next.js cacheComponents (dynamicIO) dehydration error


pluscubed:
### Describe the bug

In Next.js 15 with [cacheComponents](https://nextjs.org/blog/next-15-4#:~:text=Cache%20Components%20(beta)%3A%20Consolidates%20experimental%20caching%20features%20(Dynamic%20IO%2C%20use%20cache%2C%20and%20Partial%20Prerendering)%20into%20a%20unified%20cacheComponents%20flag%2C%20simplifying%20performance%20optimization%20strategies.%20The%20Next.js%2016%20blog%20post%20will%20detail%20our%20vision.) (formerly dynamicIO) turned on, dehydration causes an error:

```
Error: Route "/" used `Date.now()` instead of using `performance` or without explicitly calling `await connection()` beforehand. See more info here: https://nextjs.org/docs/messages/next-prerender-current-time
    at Home (app/page.tsx:15:42)
    at Home (<anonymous>)
  13 |     <main>
  14 |       <h1>Pokemon Info</h1>
> 15 |       <HydrationBoundary state={dehydrate(queryClient)}>
     |                                          ^
  16 |         <PokemonInfo />
  17 |       </HydrationBoundary>
  18 |     </main>
```

I think it's this Date.now() usage here:

https://github.com/TanStack/query/blob/3bcaa417ac302a93d17c4287d518121bb77d1b26/packages/query-core/src/hydration.ts#L82

I believe this issue is similar to #8277.

### Your minimal, reproducible example

https://codesandbox.io/p/devbox/blazing-moon-kyqpg5

### Steps to reproduce

I took the example at https://tanstack.com/query/latest/docs/framework/react/examples/nextjs-app-prefetching, changed to Next.js canary and enabled cacheComponents

### Expected behavior

No error

### How often does this bug happen?

Every time

### Platform

All platforms

### Tanstack Query adapter

react-query

### TanStack Query version

5.83.0

mdj-uk:
With `cacheComponents`/`dynamicIO` we need to choose whether the server component kicking off the prefetching is dynamic or static.

If we want it to be dynamic (i.e. we do the prefetching every time the page renders on the server at runtime, rather than once at build time) then it works fine if we
1. add `await connection()` to `Home`, to force it to render dynamically. (NB: this isn't necessary if we're already calling dynamic APIs such as `headers()` or `cookies()`, or we `await` the prefetch)
[...]

PR #11395, opened 2026-09-04
feat(query-core): allow overriding dehydration timestamp 🤖🤖🤖

aurorascharff:
## 🎯 Changes

Next.js Cache Components cannot prerender `dehydrate()` because `dehydrateQuery()` reads `Date.now()`.

The [Next.js client-side fetching guide](https://nextjs.org/docs/app/guides/client-side-data-fetching/tanstack-query#build-a-prerenderable-hydration-state) works around this by caching a timestamp with the same tags as the data, then manually mapping the query cache into a `DehydratedState`:

```ts
const updatedAt = await getHydrationUpdatedAt(tags)

queryClient.setQueryData(queryKey, data, { updatedAt })

return {
  mutations: [],
  queries: queryClient.getQueryCache().getAll().map((query) => ({
    dehydratedAt: updatedAt,
    queryHash: query.queryHash,
    queryKey: query.queryKey,
    state: query.state,
  })),
}
```

This PR adds `dehydratedAt` to `DehydrateOptions`, so applications can use the normal dehydration logic instead:

```ts
return dehydrate(queryClient, { dehydratedAt: updatedAt })
```

The caller is still responsible for keeping the timestamp in the same cache lifecycle as the data. When omitted, `dehydratedAt` continues to default to `Date.now()`.

Includes tests, documentation, and a query-core changeset. Related to #9457.

## ✅ Checklist

- [x] I have followed the steps in the Contributing Guide
- [x] I have tested the code changes locally with `pnpm run test:pr`, or does not apply
- [x] I fully understand how the code works, including any changes generated by or with the assistance of AI

Local verification passed for query-core tests, its TypeScript matrix, lint, package build, formatting, and documentation links. The only remaining repository-wide failure was a Next.js example whose build-time PokeAPI request is unavailable in the restricted environment.

## 🚀 Release Impact

- [x] This change affects code that is released to users, and I have added a changeset
- [ ] This is a documentation, CI, or development-only change, and no changeset is needed

### DIFF
--- a/.changeset/calm-plums-hydrate.md
+++ b/.changeset/calm-plums-hydrate.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/query-core': minor
+---
+
+Allow callers to provide the timestamp recorded on dehydrated queries.
--- a/docs/framework/react/guides/advanced-ssr.md
+++ b/docs/framework/react/guides/advanced-ssr.md
@@ -432,6 +432,17 @@ export function getQueryClient() {
 
 > Note: This works in NextJs and Server Components because React can serialize Promises over the wire when you pass them down to Client Components.
 
+By default, `dehydrate` records the current time on each dehydrated query. If your framework provides a stable timestamp for a cached data snapshot, pass it as `dehydratedAt` so hydration uses the same timestamp without reading the current time again:
+
+```tsx
+const { data, updatedAt } = await getCachedPostsSnapshot()
+const queryClient = new QueryClient()
+
+queryClient.setQueryData(['posts'], data, { updatedAt })
+
+const dehydratedState = dehydrate(queryClient, { dehydratedAt })
+```
+
 Then, all we need to do is provide a `HydrationBoundary`, but we don't need to `await` prefetches anymore:
 
 ```tsx
--- a/packages/query-core/src/__tests__/hydration.test.tsx
+++ b/packages/query-core/src/__tests__/hydration.test.tsx
@@ -40,6 +40,22 @@ describe('dehydration and rehydration', () => {
       queryClient.clear()
     })
 
+    it('should use a provided dehydration timestamp', () => {
+      const key = queryKey()
+      const queryClient = new QueryClient()
+      queryClient.setQueryData(key, 'data')
+      const query = queryClient.getQueryCache().find({ queryKey: key })!
+      const dateNowSpy = vi.spyOn(Date, 'now')
+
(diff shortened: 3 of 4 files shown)


Ephem:
Hi and thank you for posting this! That guide outlines the issues very well, as well as the current pain points of trying to use RQ with Cache Components. This is something we definitely want to support better, and for the full story to be really good, I think we need to do a bigger lift here.

To share some historical context, `dehydratedAt` was always a shortcut. It was introduced when hydrating promises/pending queries was still a very new thing, and it let us avoid taking on too much complexity at the time. The time the _dehydration happened_ isn't really that interesting, what we really want/need to correctly determine if a query should be hydrated or not is the same `updatedAt` each query already has.

The issue is that we don't have that available _at the time of dehydration_ and the promise itself currently only carries the data. This means that when de/rehydrating queries, the `updatedAt` is just wrong. To fix this, we'd need to:

* Wrap the promises so they also include the real `updatedAt` from the server
* Rewrite the hydration so the decision to hydrate a pending query happens _after it resolves_, not up front when it's received

If we do this we wont need a `dehydratedAt` at all.

This has some interesting overlaps with the Cache Components issues. _If a cached query function could include it's own `updatedAt`, resolved from within the cached function_, we could get rid of a lot of the ceremony. The nice thing about this approach is that it also helps the case `await queryClient.prefetchQuery(...)` which I bet also throws the current time pre-render error today?

To adopt an example from the guide, what if this _just worked_?

```tsx
export async function getProduct(id: string): Promise<Product> {
  'use cache'
  cacheLife('max')
  cacheTag(`product:${id}`)
 
  const product = await db.product.findUnique({ where: { id } })
  if (!product) throw new Error('Product not found')
  
  /* -- This is the new part --
    Essentially returns:
      {
        updatedAt: Date.now(),
        data
      }
  */
  return decorateResult(product)
}
```

We'd make it so this is also how it works internally. If you don't decorate explicitly because you use Cache Components, we still wrap the promise from the `queryFn` so the resolved promise includes the `updatedAt`, so when we dehydrate it, it always includes a timestamp as well. We'd always unwrap the promise to just data before users get a hold of it, so this is all an internal detail.

---

I'm not sure where that leaves this PR. On the one hand, it's a quick way to make the story a bit nicer while we work on figuring out the big picture. On the other hand, today `dehydratedAt` is kind of an internal detail and this PR makes it public, which is something we probably don't want long term. Will need some more time to think about this. 😄

Would love any input or feedback you have on the above as well!

PR #11407, opened 2026-09-05
test(query-core): expand notifyManager unit tests 🤖🤖🤖

pengzh1:
## 🎯 Changes

Expands the unit test coverage for `notifyManager` in `packages/query-core`. The existing suite covered the happy paths of batching and custom `notifyFn`/`batchNotifyFn`/scheduler configuration; this PR adds behavioral coverage for the gaps that were not exercised:

- **Empty batches**: a `batch()` call that schedules nothing does not invoke the scheduler at all (the flush no-op path).
- **Scheduling outside of a batch**: each `schedule()` notifies its callback individually and does not route through `batchNotifyFn`, pinning the two distinct code paths of `schedule`.
- **Nested batches**: ending an inner batch does not flush; queued callbacks are dispatched only after the outermost batch ends.
- **Return value**: `batch()` passes the callback's return value through to the caller.
- **Error recovery**: after a batch throws, the transaction counter is released, so a subsequent `schedule()` is dispatched immediately instead of being queued forever.
- **Default scheduler semantics**: notifications scheduled with the default scheduler run on a macrotask (`setTimeout(0)`) — draining microtasks alone does not flush them.

Key assertions were verified against intentionally broken implementations (empty-batch flush guard, nested-batch flush guard) and go red, so they test behavior rather than implementation details.

Test-only change: no production source is modified.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

- **Tests**
  - Added coverage for notification batching, including nested batches and empty batches.
  - Verified scheduler behavior for notifications made inside and outside batches.
  - Confirmed batched callbacks return their results and recover correctly after errors.
  - Added coverage ensuring default notifications run on a later task rather than during microtask processing.

### DIFF
--- a/packages/query-core/src/__tests__/notifyManager.test.tsx
+++ b/packages/query-core/src/__tests__/notifyManager.test.tsx
@@ -130,4 +130,98 @@ describe('notifyManager', () => {
 
     expect(callbackSpy).toHaveBeenCalledWith(1, 'test')
   })
+
+  it('should not schedule anything when a batch is empty', () => {
+    const notifyManagerTest = createNotifyManager()
+    const schedulerSpy = vi.fn((cb) => queueMicrotask(cb))
+    notifyManagerTest.setScheduler(schedulerSpy)
+
+    notifyManagerTest.batch(() => {})
+
+    expect(schedulerSpy).not.toHaveBeenCalled()
+  })
+
+  it('should notify scheduled callbacks individually outside of a batch', async () => {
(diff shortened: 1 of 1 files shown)


pengzh1:
Closing for now to keep my open set focused — no review came through in a long while. The branch is preserved, and I'm happy to re-submit the notifyManager coverage if there's appetite later. Thanks!

PR #11437, opened 2026-09-06
ref(HydrationBoundary): remove checks that are guarded by types

TkDodo:
## Summary by CodeRabbit

- **Changes**
  - Updated `HydrationBoundary` for React Query and Preact Query so the `state` property is required and must use the supported dehydrated-state format.
  - Hydration now processes query state directly while preserving deferred hydration for newer existing queries.
  - Invalid or incomplete hydration state values are no longer accepted.

- **Release**
  - Included in a patch release for the React Query and Preact Query packages.


TkDodo:
this will have to wait until v6

Issue #11371, opened 2026-09-02
Docs deleted


djmitche:
### Describe the bug

Old URLs like https://tanstack.com/query/latest/docs/framework/react/reference/useQuery now redirect to the docs homepage.

Searching for `useQuery` also goes to the homepage.

Navigating manually to the `useQuery` page brings me to [this URL](https://tanstack.com/query/latest/docs/framework/react/reference/functions/useQuery) but it seems to be a vaguely HTML-ified version of typescript alphabet-soup. I'm trying to find the option to disable reload on focus/blur, which was pretty clear in the old docs and not findable in these docs.

I'm sure there were reasons to switch to a new docs system, but perhaps consider reverting until the new docs are at least at parity in quality with the old?

### Your minimal, reproducible example

https://tanstack.com/query/latest/docs/framework/react/reference/functions/useQuery#see

### Steps to reproduce

https://tanstack.com/query/latest/docs/framework/react/reference/functions/useQuery#see

### Expected behavior

Usable documentation

### How often does this bug happen?

None

### Platform

Web

### Tanstack Query adapter

None

### TanStack Query version

latest

TkDodo:
Reference docs are now being generated from jsdoc code comments. @sukvvon can you please look into why the old links don't work anymore

baverman:
To add. I can't find useQuery options reference as well. I'm ok to follow TS interface mambo-jambo (however it's suboptimal) but there are no links either.

TkDodo:
> To add. I can't find useQuery options reference as well. I'm ok to follow TS interface mambo-jambo (however it's suboptimal) but there are no links either.

if you want to help, please a file a PR to make the jsdocs better that power the API docs: https://github.com/TanStack/query/blob/d5c0d1e475b8717be81f48541aa16eadf2d6feb3/packages/react-query/src/useQuery.ts#L50

sukvvon:
Understood. This process is currently in progress. I plan to make the document better. I would appreciate it if you could wait just a few days.

sukvvon:
@djmitche This should be fixed now — a caching/redirect issue on the docs site has been resolved. I verified `docs/framework/react/reference/useQuery` correctly redirects to the new page. Could you confirm it works on your end?

djmitche:
Confirmed - the redirect now works for me!

mwg-ofx:
This URL isn't redirecting to the correct page anymore. Is it supposed to?

[...]

PR #11438, opened 2026-09-06
docs(query-core): add JSDoc across the package and expose it in generated reference docs

sukvvon:
## 🎯 Changes

`query-core`'s public API (options like `staleTime`/`gcTime`, and classes like `QueryClient`, `Query`, `QueryObserver`, ...) previously had little to no JSDoc. On top of that, every adapter's TypeDoc config (`scripts/generate-docs.ts`) explicitly excluded `query-core` from reference-doc generation via `exclude: ['./packages/query-core/**/*']`. Together, this meant generated reference docs for framework-level hooks/functions (e.g. `useQuery`'s `options` parameter) couldn't render the underlying `query-core`-derived property details — see #11371 and Discussion #11404, where users reported not being able to tell what options exist just from the generated reference.

Rather than creating a separate "Core reference" doc hub, this PR documents `query-core` in place and lets each adapter's existing generated reference pull those docs in automatically, since every adapter's options types are ultimately derived from `query-core`.

### 1. JSDoc added across `packages/query-core/src`

18 of the 20 `packages/query-core/src/*.ts` files touched by this PR, ~1700 lines of JSDoc added (`environmentManager.ts` and `index.ts` are the other 2, covered in sections 7 and 8 below):

| File | What was documented |
| --- | --- |
| `types.ts` | `QueryOptions` (`retry`, `retryDelay`, `networkMode`, `queryFn`, `persister`, `queryHash`, `queryKey`, `queryKeyHashFn`, `initialData`, `initialDataUpdatedAt`) and `QueryObserverOptions` (`staleTime`, `gcTime`, `refetchOnMount`, `refetchOnWindowFocus`, `refetchOnReconnect`, `select`) |
| `queryClient.ts` | `QueryClient` class + `query`, `infiniteQuery`, `fetchQuery`/`fetchInfiniteQuery` (deprecated, superseded by `query`/`infiniteQuery`), `getQueriesData`, `setQueryData`, `setQueriesData`, `getQueryState`, `removeQueries`, `resetQueries`, `cancelQueries`, `invalidateQueries`, `refetchQueries`, `resumePausedMutations`, `getQueryCache`, `getMutationCache`, `get/setDefaultOptions`, `get/setQueryDefaults`, `get/setMutationDefaults`, `clear`, `isFetching`, `isMutating`, `mount`, `unmount`, `defaultQueryOptions`, `defaultMutationOptions` |
| `queryCache.ts` | `QueryCache` class, `QueryCacheConfig`, `QueryCacheNotifyEvent`, `build`, `remove`, `get`, `find`, `findAll`, `clear`, `getAll` |
| `mutationCache.ts` | `MutationCache` class, `MutationCacheConfig`, `MutationCacheNotifyEvent`, `getAll`, `clear`, `find`, `findAll` |
| `query.ts` | `QueryState`, `Query` class, `meta`/`promise` getters, `cancel`, `destroy`, `reset`, `isActive`, `isDisabled`, `isFetched`, `isStatic`, `isStale`, `isStaleByTime`, `invalidate`, `fetch`, `getObserversCount`, `setState` |
| `queryObserver.ts` | `QueryObserver` class, `setOptions`, `getCurrentResult`, `getCurrentQuery`, `refetch`, `fetchOptimistic`, `destroy`, `shouldFetchOnReconnect`, `shouldFetchOnWindowFocus`, `getOptimisticResult`, `trackResult`, `trackProp`, `updateResult` |
| `queriesObserver.ts` | `QueriesObserver` class, `QueriesObserverOptions`, `destroy`, `setQueries`, `getCurrentResult`, `getQueries`, `getObservers`, `getOptimisticResult` |
| `infiniteQueryObserver.ts` | `InfiniteQueryObserver` class, `setOptions`, `fetchNextPage`, `fetchPreviousPage`, `getOptimisticResult` |
| `mutation.ts` | `MutationState`, `Mutation` class, `meta` getter, `continue`, `execute` |
| `mutationObserver.ts` | `MutationObserver` class, `mutate`, `getCurrentResult`, `reset`, `setOptions` |
| `focusManager.ts`, `onlineManager.ts`, `timeoutManager.ts`, `notifyManager.ts`, `utils.ts`, `hydration.ts`, `retryer.ts`, `streamedQuery.ts` | Full JSDoc, based on the previously-hand-written (now superseded) `docs/reference/*.md` and current source behavior |

Descriptions were written against, and in a few cases corrected relative to, the legacy `docs/reference/*.md` prose where it no longer matched current code — e.g. `staleTime: 'static'` does *not* disable `refetchInterval` (only stops staleness-based refetches), `cancelQueries` resolves `Promise<void>` with `revert: true` as the default, and `setQueryData`'s updater short-circuits when it returns `undefined`.

58 `@example` blocks and 16 `@see` cross-references (e.g. `getQueryDefaults` ↔ `setQueryDefaults`, `isStale` ↔ `isStaleByTime`) were added on top of the prose.

### 2. `@internal` tagging

29 members with no caller outside `query-core` are tagged `@internal`, which excludes them from generated docs (their `.d.ts` output and runtime behavior are unaffected — `@internal` is TypeDoc-only). Tagging was done by auditing actual external call sites in the monorepo (all 7 framework adapters, `query-broadcast-client-experimental`, `query-persist-client-core`, `query-devtools`) rather than by guessing from the method's name — several methods that look internal by naming convention are in fact called directly by adapters or plugins and are documented instead, e.g.:

- `QueryClient.mount`/`unmount` — called by every adapter's `QueryClientProvider`-equivalent
- `Query.setState` — called by `query-broadcast-client-experimental`, `query-persist-client-core`, devtools
- `QueryCache.build`/`get` — called by plugins that need to get-or-create a `Query` directly
- `QueryObserver.getOptimisticResult`/`trackResult`/`trackProp`/`updateResult` — called by every adapter's `useBaseQuery`-equivalent
- `QueriesObserver.getOptimisticResult`, `InfiniteQueryObserver.getOptimisticResult` — called by `useQueries`/infinite-query adapter code

### 3. `scripts/generate-docs.ts`: remove `query-core` exclusion

Removed `exclude: ['./packages/query-core/**/*']` from all 7 adapter configs (`angular-query-experimental`, `svelte-query`, `solid-query`, `vue-query`, `react-query`, `preact-query`, `lit-query`), so `query-core` symbols referenced by adapter option types (e.g. `staleTime` on `UseQueryOptions`) now render inline instead of being excluded. `query-core`-derived symbols (e.g. `QueryClient`, `focusManager`) are now duplicated across each adapter's generated reference — this is intentional per-adapter self-containment, matching how `query-core` types were already inlined/duplicated for adapters that didn't have the exclude misconfigured.

Also fixes `generatePackageReferenceDocs` passing `exclude: pkg.exclude` verbatim to TypeDoc, which broke when `pkg.exclude` was `undefined` ("The 'exclude' option must be set to an array of strings"); changed to `...(pkg.exclude && { exclude: pkg.exclude })`.

### 4. tsconfig: drop `query-core` project reference

Removed the `query-core` TypeScript project reference from `angular-query-experimental`, `svelte-query`, `solid-query`, `vue-query`, and `lit-query`'s `tsconfig.json`. With the reference in place, TypeDoc/tsc resolved `query-core` types through its `dist-ts` build output rather than its `.ts` source, which meant generated docs couldn't produce GitHub source links for `query-core` symbols (they rendered as plain text instead of linked, e.g. `useInfiniteQuery.ts:70`-style links). `angular-query-experimental` keeps its `query-devtools` reference (needed for JSX/TSX compilation), which was initially assumed to reintroduce the same `dist-ts` resolution for angular's `query-core` pages — but `query-devtools`'s own `tsconfig.json` carried an unnecessary `query-core` reference of its own (transitively routing angular → query-devtools → query-core through `dist-ts`). Removed that reference too, so all 7 frameworks now consistently produce linked `.ts` source paths for `query-core` symbols.

This does not affect build output or dependency ordering: nx's project graph is derived from each package's actual `package.json` `dependencies` (e.g. `"@tanstack/query-core": "workspace:*"`), not from `tsconfig.json` `references`. Verified with `nx run-many --target=compile,test:types,test:eslint,build,test:lib --projects=vue-query,angular-query-experimental,svelte-query,solid-query,lit-query,query-devtools,angular-query-persist-client --skip-nx-cache` against a clean `dist-ts`, all green.

### 5. Regenerated docs

Ran `pnpm run generate-docs` to regenerate all reference docs under `docs/framework/*/reference/` reflecting the above (net ~1150 files touched, mostly new `query-core`-derived pages plus updated cross-links/source-line numbers in existing adapter pages).

### 6. `docs/config.json`: add sidebar entries for the newly generated pages

Each framework's TypeDoc-generated reference pages are also manually listed in `docs/config.json` to appear in the docs site sidebar. Added an entry for every newly generated `query-core`-derived page (classes, functions, interfaces, types, variables) under each framework's existing `API Reference` section, grouped and alphabetized consistently with each framework's existing category convention (`Functions` / `Classes` / `Interfaces` / `Types` / `Variables`), without disturbing pre-existing entries or lit's special `Context` grouping.

### 7. Migrate remaining prose from `docs/reference/*.md` into JSDoc

`docs/reference/*.md` (12 files: `QueryClient`, `QueryCache`, `MutationCache`, `QueryObserver`, `InfiniteQueryObserver`, `QueriesObserver`, `focusManager`, `onlineManager`, `notifyManager`, `timeoutManager`, `environmentManager`, `streamedQuery`) are legacy hand-written pages that predate TypeDoc and are not generated — see section 6 in this PR's changes for why they still exist and aren't removed here. Comparing their content against the JSDoc from section 1 turned up a few gaps, filled in this batch:

- `QueryClient.query()`'s excluded-option list is now derived from the type relationship (`QueryExecuteOptions extends QueryOptions`, whereas `QueryObserverOptions extends QueryOptions` adds the reactive-only fields) rather than copied from the legacy doc's list — which, on inspection, was missing `retryOnMount` and pinned to a 2-year-old commit permalink.
- `InvalidateQueryFilters.refetchType`'s four values (`active`/`inactive`/`all`/`none`) and `ResultOptions.throwOnError` now have JSDoc; both were previously undocumented in `types.ts`.
- `QueryClientConfig` and `DefaultOptions` fields now have JSDoc.
- `QueryClient.getQueriesData()` now documents why its `TQueryFnData` generic defaults to `unknown` (matched queries can hold different shapes; a more specific type is an unchecked convenience, not inferred).
- `QueryCacheConfig` now notes its callbacks are fire-and-forget, unlike `MutationCacheConfig`'s (verified against `query.ts`: `QueryCache`'s `onSuccess`/`onError`/`onSettled` calls aren't awaited, `MutationCache`'s are).
- `QueryCache`/`MutationCache` now have an `@example` for `subscribe` (inherited from `Subscribable`, which isn't itself exported, so it can't carry its own JSDoc/page).
- `TimeoutProvider` now documents the short-lived-vs-coalesced timer tradeoff that motivates a custom provider.
- `notifyManager.setBatchNotifyFunction` now has a framework-neutral description (dropped a stray "React Query" reference from this framework-agnostic package) and the solid-query `batch` integration example from the legacy doc.
- `environmentManager` now documents its override/restore use case with a corrected example (the legacy doc's `setIsServer(() => isServer)` referred to the boolean `isServer` export from `utils.ts`, not the differently-named function in `environmentManager.ts` — using the latter's shape here would have set up infinite recursion).
- `streamedQuery`'s `@param queryFn` was renamed to `@param streamFn` to match the actual parameter name, and its options object types now carry field-level JSDoc too (though TypeDoc doesn't currently render those on this destructured-parameter signature — a pre-existing TypeDoc limitation, not something this PR fixes).

Not migrated, and not planned to be: relative doc links (`../framework/react/guides/...`), `redirect_from` frontmatter, the commit-pinned source permalink, the TkDodo "Inside React Query" blog link, and mentions of adapter-specific hooks (`useIsFetching`, `useIsMutating`) or "options are the same as `useQuery`" — none of these translate into a framework-neutral package whose JSDoc is rendered into seven different framework trees.

### 8. Delete `docs/reference/*.md`, now that its content has been migrated

With the prose from section 7 migrated into JSDoc, all 12 legacy `docs/reference/*.md` pages (`QueryClient`, `QueryCache`, `MutationCache`, `QueryObserver`, `InfiniteQueryObserver`, `QueriesObserver`, `focusManager`, `onlineManager`, `notifyManager`, `timeoutManager`, `environmentManager`, `streamedQuery`) are deleted, along with the framework-agnostic `API Reference` `children` in `docs/config.json` that linked to them.

- **Redirects**: the 9 pages that had `redirect_from` frontmatter (pointing at old `framework/react/reference/<Name>` URLs) had those redirects re-registered, via `generate-docs.ts`'s existing `redirectFrom` mechanism, onto the new generated page that carries the equivalent content — plus the bare `/reference/<Name>` URL from the old sidebar for all 12 pages (`streamedQuery`, `environmentManager`, and `timeoutManager` didn't have a `redirect_from` before, since they were added after the flat-URL era, but did have a `/reference/<Name>` sidebar URL). #11371 tracked these old URLs falling through to the docs homepage instead of redirecting; `TanStack/tanstack.com#1217` (merged) fixed the underlying cause — a GitHub webhook 10s timeout was cutting off cache invalidation/CDN purge before `redirect_from` changes could take effect — and I verified in-browser that already-live equivalents (e.g. `docs/framework/react/reference/useQuery` → `.../functions/useQuery`) now redirect correctly. Only react ever had these legacy pages/URLs; the other 6 frameworks are unaffected.
- **In-repo links**: 12 relative markdown links from guides/plugins docs (across react, vue, preact) into the deleted pages were repointed to the new generated pages, with anchors updated to match TypeDoc's heading-derived slugs (e.g. `#queryclientclear` → `#clear`), verified against the equivalent already-live generated page on tanstack.com.
- **Content-loss check**: `focusManager`, `onlineManager`, and `timeoutManager` are exported as *instances* of the `FocusManager`/`OnlineManager`/`TimeoutManager` classes, and those classes weren't themselves exported from `query-core`'s entry point — so TypeDoc could only render a 3-line stub for the instance (no method docs), which would have silently dropped the `setEventListener`/`setFocused`/`isFocused`-style documentation the legacy pages carried. Fixed by adding `export type { FocusManager, OnlineManager, TimeoutManager }` to `query-core`'s `index.ts` (type-only, no runtime change, but new published surface — see the changeset). TypeDoc now generates a full `interfaces/<Name>.md` page for each with all methods and examples, and the redirects above point there instead of at the stub. `environmentManager` and `notifyManager` didn't need this — both are plain object literals, so TypeDoc already inlines their methods on the instance's own page.

This closes out the `docs/reference` legacy-doc-removal work: options/API descriptions for `query-core` now live solely in JSDoc, surfaced through each framework's generated reference.

### 9. Fix `query-devtools`'s own unnecessary `query-core` project reference

Found while investigating why `angular-query-experimental` (the one framework whose docs still resolved through `dist-ts` after section 4) hadn't been fixed by the exclusion changes above: `packages/query-devtools/tsconfig.json` carried its own `query-core` project reference that `angular-query-experimental` transitively pulled in via its (required, JSX-compilation-only) `query-devtools` reference. Removed it — same `@tanstack/custom-condition` reasoning as section 4 — so all 7 frameworks now consistently resolve `query-core` symbols through `.ts` source. Verified `nx run-many --target=compile,test:types,test:eslint,build,test:lib --projects=preact-query-devtools,react-query-devtools,solid-query-devtools,svelte-query-devtools,vue-query-devtools --skip-nx-cache` against a clean `dist-ts`, all green.

### 10. Document previously-missing default values, and normalize "Defaults to" spacing

A pass over `types.ts` (and `queriesObserver.ts`/`utils.ts`) looking for options that already had a description but no stated default, or that had one immediately abutting the description with no line break (inconsistent with most of the file):

- `QueryOptions`/`MutationOptions`: `retryDelay` (defaults to exponential backoff capped at 30s — was previously only described via example, no stated default), `networkMode` (added a describing sentence; it previously jumped straight to "Defaults to `'online'`" with none)
- `MutationOptions.retry`/`networkMode`/`gcTime`: same "Defaults to ..." treatment as their `QueryOptions` counterparts, but the actual defaults differ and weren't copy-pasted — mutation `retry` defaults to `0` vs. query's client/server `3`/`0` split (`retryer.ts` vs. `mutation.ts`); `networkMode`/`gcTime` match `QueryOptions`
- `QueriesObserverOptions.combine`: defaults to returning the raw per-query result array unchanged
- `QueryFilters.type`: defaults to `'all'` (confirmed via the `type = 'all'` destructure in `utils.ts`)
- `MutationObserverOptions.throwOnError`: had no JSDoc at all, unlike `QueryObserverOptions.throwOnError`'s full description — added, default `false` confirmed via `shouldThrowError`'s `!!throwOnError` fallback
- `notifyOnChangeProps`: reworded its prose-only default explanation into the same "Defaults to ..." shape as its thirteen sibling fields
- Blank-line-before-"Defaults to" normalized across ~15 fields that had the description and default running together with no line break

No behavior changes; this is documentation-only. Verified with `pnpm run generate-docs` that only the intended sentences render as new content, with no deleted descriptions.

## ✅ Checklist

## 🚀 Release Impact

- [x] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md) (`@tanstack/query-core` minor, for the `FocusManager`/`OnlineManager`/`TimeoutManager` type exports in section 8 — everything else in this PR is docs-only).


sukvvon:
A few non-obvious spots worth a closer look, since this diff is large:

- [`scripts/generate-docs.ts#L166`](https://github.com/TanStack/query/blob/4a387380cb43cbf487ecc6fbdc15fa9c42e7f1d0/scripts/generate-docs.ts#L166) — `...(pkg.exclude && { exclude: pkg.exclude })` instead of `exclude: pkg.exclude`. This is what makes removing every adapter's `exclude` in this PR possible: passing `exclude: undefined` to TypeDoc throws `"The 'exclude' option must be set to an array of strings"`, so the key has to be omitted entirely when a package doesn't set it.

- [`scripts/generate-docs.ts#L22-25`](https://github.com/TanStack/query/blob/4a387380cb43cbf487ecc6fbdc15fa9c42e7f1d0/scripts/generate-docs.ts#L22-L25) and the `redirectFrom` blocks below it — this is the mechanism used to keep the 12 deleted `docs/reference/*.md` URLs (and their `redirect_from`) alive by re-registering them on the new generated pages. Only react ever had these legacy pages, so only react's config carries these entries.

- [`packages/query-core/src/index.ts#L4,L19,L30`](https://github.com/TanStack/query/blob/4a387380cb43cbf487ecc6fbdc15fa9c42e7f1d0/packages/query-core/src/index.ts#L4) — new `export type { FocusManager }` / `OnlineManager` / `TimeoutManager`. These are the only functional (non-docs) change in this PR — see the changeset. Needed because `focusManager`/`onlineManager`/`timeoutManager` are exported as *instances*, and without the class type also exported, TypeDoc couldn't render their methods at all (just a 3-line stub) — which would have silently dropped documentation the deleted legacy pages used to carry.

- [`packages/query-core/src/environmentManager.ts#L19-26`](https://github.com/TanStack/query/blob/4a387380cb43cbf487ecc6fbdc15fa9c42e7f1d0/packages/query-core/src/environmentManager.ts#L19-L26) — the `@example`'s "restore the default behavior" line is `setIsServer(() => isServer)`, not `setIsServer(() => isServer())`. The package's exported `isServer` is the boolean constant from `utils.ts`, not the differently-named function in this file — calling it would be a type error, and copy-pasting the shape of the *other* `isServer` here would set up infinite recursion. Worth double-checking if you're familiar with this corner.

- [`packages/query-core/src/streamedQuery.ts#L57-58`](https://github.com/TanStack/query/blob/4a387380cb43cbf487ecc6fbdc15fa9c42e7f1d0/packages/query-core/src/streamedQuery.ts#L57) — `@param queryFn` was renamed to `@param streamFn` to match the actual parameter name (pure accuracy fix, first noticed because the mismatch is why TypeDoc wasn't attaching this description to the rendered docs at all).

Also flagging: this PR does not fix the tanstack.com redirect handling itself (#11371) — the `redirect_from` entries added/moved here are inert until that's resolved on the tanstack.com side.

sukvvon:
Follow-up on my [earlier comment](https://github.com/TanStack/query/pull/11438#issuecomment-5561333123) — that one was written against `4a387380c`, before the branch was updated to current `main` and before the `redirectFrom` fix in the latest commit. Links below are pinned to the current HEAD (`e3d1a93e8`). A few more spots worth flagging:

- [`packages/query-core/src/queryClient.ts#L536-539`](https://github.com/TanStack/query/blob/e3d1a93e8baa9a66b2234b25ecc8a9ad9099aacf/packages/query-core/src/queryClient.ts#L536-L539) — `query()`'s excluded-option list. The legacy `docs/reference/QueryClient.md` this replaces had a similar list but was missing `retryOnMount`, and was otherwise unverified prose from ~2 years ago. This one is derived directly from the type relationship (`QueryExecuteOptions extends QueryOptions`, `QueryObserverOptions extends QueryOptions` adds the reactive-only fields) rather than copied, so it should be trustworthy — but it's exactly the kind of claim worth someone else double-checking against `types.ts`.

- `@internal` tagging — 29 members across the package (e.g. [`queryCache.ts#L180`](https://github.com/TanStack/query/blob/e3d1a93e8baa9a66b2234b25ecc8a9ad9099aacf/packages/query-core/src/queryCache.ts#L180)) are tagged `@internal`, which hides them from generated docs but doesn't change `.d.ts` output or runtime behavior. The criterion used was "does any package outside `query-core` actually call this" (checked across all 7 adapters + `query-broadcast-client-experimental` + `query-persist-client-core` + `query-devtools`), not naming convention — several methods that read as internal (`QueryClient.mount`, `Query.setState`, `QueryCache.build`) are called externally and got full docs instead. If anyone knows of an external call site I missed for something currently tagged `@internal`, that'd be useful to catch here rather than after merge.

- [`docs/config.json#L972`](https://github.com/TanStack/query/blob/e3d1a93e8baa9a66b2234b25ecc8a9ad9099aacf/docs/config.json#L972) (`API Reference` section, each framework) — this PR adds ~130 sidebar entries per framework (906 total) for every newly generated `query-core`-derived page, including low-level types like `interfaces/CancelOptions` and internal-ish result-state interfaces. The previous sidebar was curated, not exhaustive. Flagging in case the volume itself is unwanted, independent of whether the underlying docs are correct — happy to trim to just the "primary" symbols (classes + top-level option/result types) if that's preferred, since it's a mechanical, easily-revertible pass.

- 12 in-repo relative links from guides/plugins (react, vue, preact) into the now-deleted `docs/reference/*.md` pages were repointed to the new generated pages, e.g. [`docs/framework/react/guides/mutations.md#L62`](https://github.com/TanStack/query/blob/e3d1a93e8baa9a66b2234b25ecc8a9ad9099aacf/docs/framework/react/guides/mutations.md#L62). Anchors like `#invalidatequeries` were derived from TypeDoc's `### invalidateQueries()` headings and verified against the equivalent already-live page on tanstack.com (e.g. `reference/classes/QueryClient` renders `id="setquerydata"` for that method) — but that verification was against a *different* page than the ones being linked to here, so it's inference, not a direct check of these exact URLs once deployed.

- [`scripts/generate-docs.ts#L312-320`](https://github.com/TanStack/query/blob/e3d1a93e8baa9a66b2234b25ecc8a9ad9099aacf/scripts/generate-docs.ts#L312-L320) — the `redirectFrom` entries for `focusManager`/`onlineManager`/`timeoutManager` point at `interfaces/<Name>`, not `variables/<name>`. That's deliberate: those three are exported as *instances* of a class, and the class itself wasn't exported from `query-core` before this PR, so TypeDoc could only render a stub for the instance page (no methods). The redirect goes to the interface page (now exported — see the `index.ts` change and changeset) because that's where the actual method docs the old pages covered ended up.

sukvvon:
A few non-obvious spots worth a closer look, since this diff is large (pinned to current HEAD, `3ffe31428`):

- [`scripts/generate-docs.ts#L166`](https://github.com/TanStack/query/blob/3ffe314281cdd31e424325809827ec1c6485fca5/scripts/generate-docs.ts#L166) — `...(pkg.exclude && { exclude: pkg.exclude })` instead of `exclude: pkg.exclude`. This is what makes removing every adapter's `exclude` in this PR possible: passing `exclude: undefined` to TypeDoc throws `"The 'exclude' option must be set to an array of strings"`, so the key has to be omitted entirely when a package doesn't set it.

- [`scripts/generate-docs.ts#L22-25`](https://github.com/TanStack/query/blob/3ffe314281cdd31e424325809827ec1c6485fca5/scripts/generate-docs.ts#L22-L25) and the `redirectFrom` blocks below it (e.g. [`L312-320`](https://github.com/TanStack/query/blob/3ffe314281cdd31e424325809827ec1c6485fca5/scripts/generate-docs.ts#L312-L320)) — this is the mechanism used to keep the 12 deleted `docs/reference/*.md` URLs (and their `redirect_from`) alive by re-registering them on the new generated pages. Only react ever had these legacy pages, so only react's config carries these entries. Note `interfaces/FocusManager`/`OnlineManager`/`TimeoutManager` redirect targets (not `variables/<name>`) — deliberate, see the next point.

- [`packages/query-core/src/index.ts#L4,L19,L30`](https://github.com/TanStack/query/blob/3ffe314281cdd31e424325809827ec1c6485fca5/packages/query-core/src/index.ts#L4) — new `export type { FocusManager }` / `OnlineManager` / `TimeoutManager`. These are the only functional (non-docs) change in this PR — see the changeset. `focusManager`/`onlineManager`/`timeoutManager` are exported as *instances*, and without the class type also exported, TypeDoc couldn't render their methods at all (just a 3-line stub) — which would have silently dropped documentation the deleted legacy pages used to carry. That's also why the redirects above point at `interfaces/<Name>` rather than `variables/<name>`: that's where the actual method docs ended up.

- [`packages/query-core/src/environmentManager.ts#L19-26`](https://github.com/TanStack/query/blob/3ffe314281cdd31e424325809827ec1c6485fca5/packages/query-core/src/environmentManager.ts#L19-L26) — the `@example`'s "restore the default behavior" line is `setIsServer(() => isServer)`, not `setIsServer(() => isServer())`. The package's exported `isServer` is the boolean constant from `utils.ts`, not the differently-named function in this file — calling it would be a type error, and copy-pasting the shape of the *other* `isServer` here would set up infinite recursion. Worth double-checking if you're familiar with this corner.

- [`packages/query-core/src/streamedQuery.ts#L57-58`](https://github.com/TanStack/query/blob/3ffe314281cdd31e424325809827ec1c6485fca5/packages/query-core/src/streamedQuery.ts#L57) — `@param queryFn` was renamed to `@param streamFn` to match the actual parameter name (pure accuracy fix, first noticed because the mismatch is why TypeDoc wasn't attaching this description to the rendered docs at all).

- [`packages/query-core/src/queryClient.ts#L536-539`](https://github.com/TanStack/query/blob/3ffe314281cdd31e424325809827ec1c6485fca5/packages/query-core/src/queryClient.ts#L536-L539) — `query()`'s excluded-option list. The legacy `docs/reference/QueryClient.md` this replaces had a similar list but was missing `retryOnMount`, and was otherwise unverified prose from ~2 years ago. This one is derived directly from the type relationship (`QueryExecuteOptions extends QueryOptions`, `QueryObserverOptions extends QueryOptions` adds the reactive-only fields) rather than copied, so it should be trustworthy — but it's exactly the kind of claim worth someone else double-checking against `types.ts`.

- `@internal` tagging — 29 members across the package (e.g. [`queryCache.ts#L180`](https://github.com/TanStack/query/blob/3ffe314281cdd31e424325809827ec1c6485fca5/packages/query-core/src/queryCache.ts#L180)) are tagged `@internal`, which hides them from generated docs but doesn't change `.d.ts` output or runtime behavior. The criterion used was "does any package outside `query-core` actually call this" (checked across all 7 adapters + `query-broadcast-client-experimental` + `query-persist-client-core` + `query-devtools`), not naming convention — several methods that read as internal (`QueryClient.mount`, `Query.setState`, `QueryCache.build`) are called externally and got full docs instead. If anyone knows of an external call site I missed for something currently tagged `@internal`, that'd be useful to catch here rather than after merge.

- [`docs/config.json#L972`](https://github.com/TanStack/query/blob/3ffe314281cdd31e424325809827ec1c6485fca5/docs/config.json#L972) (`API Reference` section, each framework) — this PR adds ~130 sidebar entries per framework (906 total) for every newly generated `query-core`-derived page, including low-level types like `interfaces/CancelOptions` and internal-ish result-state interfaces. The previous sidebar was curated, not exhaustive. Flagging in case the volume itself is unwanted, independent of whether the underlying docs are correct — happy to trim to just the "primary" symbols (classes + top-level option/result types) if that's preferred, since it's a mechanical, easily-revertible pass.

- 12 in-repo relative links from guides/plugins (react, vue, preact) into the now-deleted `docs/reference/*.md` pages were repointed to the new generated pages, e.g. [`docs/framework/react/guides/mutations.md#L62`](https://github.com/TanStack/query/blob/3ffe314281cdd31e424325809827ec1c6485fca5/docs/framework/react/guides/mutations.md#L62). Anchors like `#invalidatequeries` were derived from TypeDoc's `### invalidateQueries()` headings and verified against the equivalent already-live page on tanstack.com (e.g. `reference/classes/QueryClient` renders `id="setquerydata"` for that method) — but that verification was against a *different* page than the ones being linked to here, so it's inference, not a direct check of these exact URLs once deployed.

- [`packages/query-devtools/tsconfig.json`](https://github.com/TanStack/query/blob/3ffe314281cdd31e424325809827ec1c6485fca5/packages/query-devtools/tsconfig.json) (its `query-core` reference was removed) — `angular-query-experimental` was the only framework still rendering unlinked, `dist-ts`-resolved `query-core` docs (e.g. `Defined in: packages/query-core/dist-ts/src/queryClient.d.ts:24` instead of a linked `.ts` source line) even after section 4 dropped its own `query-core` reference. Root cause: `angular-query-experimental` still needs its `query-devtools` reference for JSX/TSX compilation, but `query-devtools`'s own `tsconfig.json` carried an unrelated `query-core` project reference, transitively routing angular's `query-core` symbols through `dist-ts` anyway. Removed that reference (same `@tanstack/custom-condition` reasoning as the other 5 packages in section 4) — all 7 frameworks now produce linked `.ts` source paths consistently.

- [`packages/query-core/src/types.ts`](https://github.com/TanStack/query/blob/3ffe314281cdd31e424325809827ec1c6485fca5/packages/query-core/src/types.ts) — `MutationOptions.retry`/`retryDelay`/`networkMode`/`gcTime` gained the same "Defaults to ..." documentation `QueryOptions` already had for the equivalent fields, but the defaults themselves differ and aren't just copy-pasted: mutation `retry` defaults to `0` (vs. query's `3`/`0` client/server split — see [`retryer.ts#L196`](https://github.com/TanStack/query/blob/3ffe314281cdd31e424325809827ec1c6485fca5/packages/query-core/src/retryer.ts#L196) vs [`mutation.ts#L310`](https://github.com/TanStack/query/blob/3ffe314281cdd31e424325809827ec1c6485fca5/packages/query-core/src/mutation.ts#L310)), `networkMode`/`gcTime` are the same as `QueryOptions`. Also added missing defaults to `QueriesObserverOptions.combine` (falls back to returning the raw per-query result array), `QueryFilters.type` (defaults to `'all'`, confirmed via the `type = 'all'` destructure in [`utils.ts#L180`](https://github.com/TanStack/query/blob/3ffe314281cdd31e424325809827ec1c6485fca5/packages/query-core/src/utils.ts#L180)), and `MutationObserverOptions.throwOnError` (had no JSDoc at all despite `QueryObserverOptions.throwOnError` having a full description — defaults to `false`, confirmed via `shouldThrowError`'s `!!throwOnError` fallback in `utils.ts`).

Update: #11371 (old reference URLs falling through to the docs homepage instead of redirecting) is fixed by `TanStack/tanstack.com#1217` (merged) — the root cause was a GitHub webhook 10s timeout cutting off cache invalidation/CDN purge before `redirect_from` changes took effect. Verified in-browser that already-live equivalents redirect correctly now, so the `redirect_from` entries this PR adds/moves should take effect once merged, rather than staying inert as I originally flagged.

Note for anyone mid-review: the latest commit (`3ffe31428`) is spacing/default-value documentation normalization only (consistent blank line before every "Defaults to ..." line, plus a few previously-undocumented static defaults like the one above) — no behavior changes, no new claims beyond what's already described here.

packages/query-core/src/environmentManager.ts
```diff
@@ -10,7 +10,21 @@ let isServerFn: IsServerValue = () => defaultIsServer
...
+ * extension workers), where the default detection would give the wrong answer.
+ *
+ * @example
+ * ```ts
+ * import { environmentManager, isServer } from '@tanstack/query-core'
+ *
+ * environmentManager.setIsServer(() => typeof window === 'undefined' && !('chrome' in globalThis))
+ *
+ * // Restore the default behavior:
+ * environmentManager.setIsServer(() => isServer)
```
TkDodo: can we cleanup / fix this example please?

1. `isServer` import is deprecated and shouldn’t be used
2. `’chrome' in globalThis` seems like a weird example?
3. restoring should probably capture the previous version of `environmentManager.isServer()` but I don’t think we need restoring as an example.

sukvvon: @TkDodo Fixed in 2031b93: dropped the deprecated `isServer` import, replaced `'chrome' in globalThis` with a `false` constant, and removed the restore snippet.

TkDodo (approved):

PR #11456, opened 2026-09-09
fix(solid-query): untrack the one-shot client and options reads on mount

### LINKED ISSUE #11358: solid-query: STRICT_READ_UNTRACKED on Solid 2 — client()/options() read in component body (useMutation, useBaseQuery)
### Describe the bug

With Solid 2 (`solid-js@2.0.0-rc.4`) and `@tanstack/solid-query@6.0.0-rc.1`, every component that calls `useMutation` (and `useQuery`) logs a reactivity diagnostic on mount:

```
[STRICT_READ_UNTRACKED] Reactive value read directly in <UnitCard> will not update.
Move it into a tracking scope (JSX, a memo, or an effect's compute function).
[STRICT_READ_UNTRACKED] repair guide: node_modules/solid-js/skills/reactivity-diagnostics/SKILL.md
```

Solid 2 ships these diagnostics in dev builds, and they are meant to be actionable: the console is expected to stay clean. Right now a single screen with a handful of mutations produces dozens of them, which drowns out the diagnostics that *are* about application code.

### Cause

The hooks read memos directly in the component body, outside any tracking scope.

`packages/solid-query/src/useMutation.ts`:

```ts
const client = createMemo(() => resolveClient())

const observer = new MutationObserver(client(), options())   // ← read in component body
```

`packages/solid-query/src/useBaseQuery.ts` does the same when it creates its observer:

```ts
const initialOptions = defaultedOptions()

const [observer, setObserver] = createSignal(
  new Observer(client(), defaultedOptions()),               // ← read in component body
)
```

Both reads are intentional one-shot initialisations, so the value genuinely does not need to be tracked — which is exactly the case Solid's `untrack` exists for. Under Solid 1 this was silent; Solid 2 reports it.

Any component calling `useMutation` reproduces it. The official Solid 2 fullstack template does it out of the box:

https://github.com/solidjs/templates/tree/main/solid-v2/fullstack-tanstack

`src/routes/users.$id.tsx` calls `useMutation`; open the page with a dev build of `solid-js@2` and the diagnostic appears on every mount.

### Steps to reproduce

1. Use `solid-js@2.0.0-rc.4` with `@tanstack/solid-query@6.0.0-rc.1`.
2. Render any component that calls `useMutation` (or `useQuery`).
3. Watch the browser console on mount.

### Expected behavior

No reactivity diagnostics from library code — the initial reads of `client()` / `options()` are untracked by design and should say so.

### Suggested fix

Wrap the initialisation reads in `untrack(...)`, or move observer creation into a `createRenderEffect` / memo. A related Solid-specific fix was already made in #10445 (`useQueryClient` running inside `createMemo`).

### Platform

- `@tanstack/solid-query` 6.0.0-rc.1 (latest `rc` on npm as of 2026-09-01)
- `solid-js` 2.0.0-rc.4
- Chromium 141, macOS

### Note

Not filing a PR because I did not want to guess which of the two shapes (`untrack` vs. deferred observer creation) you prefer for v6 — happy to send one either way.

mnkprs:
Fixes #11358.

Every solid-query hook seeds an observer or a signal from `client()` and its options accessor while the hook itself is running. Later changes reach those observers through `setOptions`/`setQueries` and the computations around them, so those initial reads are one-shot by design — but they were still made inside whatever tracking scope called the hook. On Solid 2 that makes each hook call log a `[STRICT_READ_UNTRACKED]` diagnostic on mount, so a screen with a handful of queries and mutations fills the console with library noise and drowns out the diagnostics that are about application code.

This wraps those reads in `untrack`, which is the case it exists for. Affected:

- `useBaseQuery` — `initialOptions`, the `Observer` construction, and the first `getOptimisticResult`
- `useMutation` — the `MutationObserver` construction
- `useQueries` — the `QueriesObserver`, the initial store contents, the seeding `batch`, and the initial `getProxies()`
- `useIsFetching` / `useIsMutating` / `useMutationState` — the signal seeds

### Why this is not a behaviour change for components

Solid's `createComponent` already wraps component bodies in `untrack`, so for the way these hooks are actually called the reads were untracked before this PR too. I checked this by running the same scenarios against the merge base and against this branch — client swaps and option changes behave identically in a component either way.

The behaviour that does change is a hook called *directly* inside a `createComputed`/`createMemo`. There, the tracked reads used to tear the hook down and rebuild it wholesale whenever the client or the options changed. That accidental rebuild is what the new tests pin down as gone.

### Tests

`src/__tests__/untrackedReads.test.tsx` calls each hook inside a computation that reads nothing itself, then invalidates the client and the options the hook read, and asserts the computation does not re-run. This makes the leak observable on Solid 1 as well, rather than only as a Solid 2 diagnostic. All six fail on `main` (3 runs instead of 1) and pass here.

### Note for reviewers

While verifying the above I found three pre-existing bugs that this PR deliberately does **not** touch, since they are out of scope for the ticket and all reproduce on `main` today in ordinary component usage:

1. `useBaseQuery`'s `on(client, …)` handler calls `createClientSubscriber()` — which reads the `observer` signal — before `setObserver(newObserver)`, so it resubscribes to the outgoing observer and later updates to the new client's cache never arrive.
2. `useMutation` and `useQueries` have no client-change handling at all; swapping the client leaves them on the old one.
3. `useMutationState` never recomputes when its filters change — the solid twin of #11272.

I have fixes and regression tests ready for (1) and (3) and will send them as separate PRs if you'd like them.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/solid-query-untracked-initial-reads.md
+++ b/.changeset/solid-query-untracked-initial-reads.md
@@ -1,0 +1,7 @@
+---
+'@tanstack/solid-query': patch
+---
+
+fix(solid-query): untrack the one-shot client and options reads the hooks make on mount
+
+`useQuery`, `useQueries`, `useMutation`, `useIsFetching`, `useIsMutating` and `useMutationState` each read their `client` memo and their options accessor directly while seeding an observer or a signal. Later changes reach those observers through `setOptions`/`setQueries` and the effects around them, so the initial reads are one-shot by design — but they were still made in a tracking scope. On Solid 2 that makes every hook call log a `[STRICT_READ_UNTRACKED]` diagnostic on mount, and a hook called from inside a computation had that computation re-run and its observer rebuilt whenever the client or the options changed. The reads are now untracked.
--- a/packages/solid-query/src/__tests__/untrackedReads.test.tsx
+++ b/packages/solid-query/src/__tests__/untrackedReads.test.tsx
@@ -1,0 +1,172 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { createComputed, createRoot, createSignal } from 'solid-js'
+import { queryKey, sleep } from '@tanstack/query-test-utils'
+import {
+  QueryClient,
+  useIsFetching,
+  useIsMutating,
+  useMutation,
+  useMutationState,
+  useQueries,
+  useQuery,
+} from '..'
+
+/**
+ * Every hook seeds its state from the current client and options before handing
+ * later changes to an observer. Those initial reads are one-shot by design, so
+ * they must not register as dependencies of whatever is running when the hook is
+ * called — Solid's `STRICT_READ_UNTRACKED` diagnostics report them, and a caller
+ * inside a tracking scope gets its computation re-run and its hook rebuilt.
+ *
+ * Each test calls a hook inside a computation that reads nothing itself, then
+ * invalidates the sources the hook read. The computation must not re-run.
+ */
+function countRunsOf(run: () => void) {
+  let runs = 0
+  const dispose = createRoot((disposeRoot) => {
+    createComputed(() => {
+      runs++
+      run()
+    })
+    return disposeRoot
+  })
+  return { runs: () => runs, dispose }
+}
+
+describe('untracked reads', () => {
+  let queryClient: QueryClient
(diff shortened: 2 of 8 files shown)


brenelz:
I don't think just untracking it is the best solution. This should also target the v6 branch

PR #11470, opened 2026-09-11
fix(query-core): clear mutations before notifying subscribers

chungyo:
## 🎯 Changes

`queryClient.clear()` empties the mutation cache, but `useMutationState()` can keep showing successful results and `useIsMutating()` can keep reporting pending mutations. `MutationCache.clear()` notifies synchronous subscribers before removing entries, so they read the old cache and see no change.

Clear the mutations and scopes before emitting removal events. Snapshot the removed mutations first so each receives its event and mutations added by a subscriber are preserved.

Adds four regression tests covering the cache seen by subscribers, mutations added during notification, and the two React hooks. All four failed before the fix.

Validation: all 682 query-core and 585 react-query tests pass, along with their TypeScript checks (5.6, 5.7, 5.8, 5.9, current, and 7.0), lint, builds, and package validation. The full `pnpm run test:pr --parallel=3` check also passed for 98 affected projects and their dependency tasks, including the other adapters and examples. Existing lint warnings are in unchanged files; the run also emitted non-fatal warnings about generated example tsconfig files.

Implemented and verified with AI assistance.

## ✅ Checklist

## 🚀 Release Impact

### DIFF
--- a/.changeset/clear-mutation-cache-before-notifying.md
+++ b/.changeset/clear-mutation-cache-before-notifying.md
@@ -1,0 +1,5 @@
+---
+'@tanstack/query-core': patch
+---
+
+Clear the mutation cache before notifying subscribers so `useMutationState` and `useIsMutating` update correctly after `queryClient.clear()`.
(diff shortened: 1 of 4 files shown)


sheraz4196 (approved):
The fix correctly clears the mutation cache before emitting removal events, while snapshotting the original mutations so subscriber-added mutations are preserved. The regression coverage also validates both the cache-level behavior and the React hook behavior.

PR #11502, opened 2026-09-15
fix(query-core): reject longer partial array keys 

gwagjiug:
## 🎯 Changes

`partialMatchKey` previously iterated over every element in the filter key without first checking whether the candidate array was long enough. A filter such as `['todos', undefined]` therefore matched the shorter `['todos']` key because both missing and explicit array elements evaluate to `undefined`.

- Return `false` when an array filter is longer than the candidate array.
- Add a regression test for a trailing `undefined` filter element.
- Add a patch changeset for `@tanstack/query-core`.

Validation: `pnpm run test:pr` with Node 24.16.0.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed partial query-key matching so longer filters with trailing `undefined` values no longer match shorter query keys.
  - Improved query matching accuracy when the filter contains more array elements than the target key.
  - Prevented incorrect matches caused by differences in query-key length.

- **Tests**
  - Added coverage for preventing incorrect matches between query keys of different lengths, including keys with trailing `undefined` values.


TkDodo (approved):

packages/query-core/src/__tests__/utils.test.tsx
```diff
@@ -144,6 +144,12 @@ describe('core/utils', () => {
       expect(partialMatchKey(a, b)).toBe(true)
     })
 
+    it('should return `false` if array b is longer than array a', () => {
```
sheraz4196: Small test naming suggestion: the regression is specifically about a longer filter with a trailing `undefined` value, rather than array length alone. Could we make the test name reflect that, e.g. `should return false if a is shorter and b has a trailing undefined`? That makes the reason for the regression clearer.

sheraz4196 (changes requested):

packages/query-core/src/__tests__/utils.test.tsx
gwagjiug: @sheraz4196 

Good point. I’ve updated the test name to `should return false if a is shorter and b has a trailing undefined` so it describes the exact regression rather than array length in general. Thanks for the suggestion!

sheraz4196 (approved):

PR #11515, opened 2026-09-16
fix(query-core): preserve symbol properties in replaceEqualDeep

### LINKED ISSUE #11514: Structural sharing returns incorrectly mutated data on refetch when the data has a `[Symbol.iterator]` property
### Describe the bug

When all of these conditions are true:

1. A query's `queryFn` returns a plain object.
2. That object is iterable via having a `[Symbol.iterator]` property.
3. That object has at least one other property that is a function.
4. Structural sharing is enabled for that query.
5. A refetch occurs for that query (e.g. via window focus, mount, interval, etc).

Then the data returned by the query following the refetch will be missing its `[Symbol.iterator]` property.

https://stackblitz.com/edit/tanstack-query-i4ows4sf?file=src%2Findex.tsx

### Steps to reproduce

1. Load the example app
2. See that the query is fetched and the data is displayed properly
3. Wait 5 seconds for an automatic refetch to happen
4. See that the returned data is now missing its [Symbol.iterator] key

### Expected behavior

The returned data after refetch should have all of the same properties as before the refetch.

Every time

### Screenshots or Videos

_No response_

### Platform

Occurs on the latest version of both Chrome and Firefox.

### Tanstack Query adapter

_No response_

### TanStack Query version

5.102.8

### TypeScript version

_No response_

### Additional context

_No response_

lx3133584:
## 🎯 Changes

Preserve enumerable symbol properties (such as `[Symbol.iterator]`) on objects during `replaceEqualDeep`.

Previously, `replaceEqualDeep` only inspected `Object.keys(o)`, omitting symbol properties when constructing the new shared copy if values differed. This caused objects with `[Symbol.iterator]` to lose their iterator property across refetches.

Fixes #11514

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed deep value replacement to preserve enumerable symbol-keyed properties, including `Symbol.iterator`.
  - Improved structural sharing comparisons so symbol properties are handled alongside standard object properties.

- **Tests**
  - Added coverage confirming symbol properties are preserved while changed regular properties are correctly replaced.

- **Chores**
  - Documented a patch release for the query core package.


TkDodo:
see:
- https://github.com/TanStack/query/issues/11514#issuecomment-5704029805

Issue #9981, opened 2025-12-17
`isSuccess()` not triggering `effect()` during testing


JPtenBerge:
### Describe the bug

I have this `AppComponent`, with an `effect()` that's dependent on the `isSuccess()` of a query:

```ts
@Component({
	selector: 'app-root',
	templateUrl: './app.html',
	imports: [ReactiveFormsModule],
	styleUrl: './app.css',
	changeDetection: ChangeDetectionStrategy.OnPush,
})
export class App {
	dataService = inject(DataService);

	formField = new FormControl<number>(42);

	query = injectQuery(() => ({
		queryKey: ['query'],
		queryFn: () => {
			console.log('in query', this.dataService.give());
			return this.dataService.give();
		},
	}));

	constructor() {
		effect(() => {
			console.log('in effect:', this.query.isSuccess());
			if (!this.query.isSuccess()) return;

			console.log('setting form value');
			this.formField.setValue(9999);
		});
	}
}
```

And I'm trying to test this, either through the classic `TestBed` or new Vitest features combined with the Testing Library:

```ts
describe('App', () => {
	let dataServiceMock: Mocked<DataService>;

	beforeEach(() => {
		dataServiceMock = { give: vi.fn().mockReturnValue(Promise.resolve([30, 40, 50])) } as Mocked<DataService>;
	});

	it('should work using the testing library', async () => {
		let sut = await render(App, {
			providers: [provideQueryClient(new QueryClient()), { provide: DataService, useValue: dataServiceMock }],
		});

		await sut.rerender();
		sut.detectChanges();
		await sut.fixture.whenStable();
		await sut.rerender();
		sut.detectChanges();
		await sut.fixture.whenStable();
		await sut.rerender();
		sut.detectChanges();
		await sut.fixture.whenStable();
		await sut.rerender();
		sut.detectChanges();
		await sut.fixture.whenStable();
		await sut.rerender();
		TestBed.tick();
		TestBed.tick();
		TestBed.tick();
		sut.fixture.detectChanges();
		await sut.fixture.whenStable();
		sut.detectChanges();
		await sut.fixture.whenStable();
		await sut.rerender();
		await expect(page.getByRole('textbox')).toHaveValue(9999);
		expect(sut.fixture.componentInstance.formField.value).toBe(9999);
	});

	it('should work using TestBed', async () => {
		TestBed.configureTestingModule({
			imports: [App],
[...]

PR #11516, opened 2026-09-16
RC of Angular Adapter

benjavicente:
## 🎯 Changes

Creates a prerelease of the Angular Adapter. It follow the same API, but has changes in the implementation to better support some edge cases present in the experimental version (better pending tasks support #9981, immediate updates #9735, invalidate the delayed subscription #9020, improved signal aliasing #7488, and remove shape that would easelly allowed accidentally sharing a query client between requests in SSR [[Discord]](https://discord.com/channels/719702312431386674/1486644228707717152/1486644228707717152)).

This changes originated from PR #9893, that would also have introduced breaking changes, but has since then deviated a lot to resolve other reported issues while improving the internal shape of the adapter. The older Angular version supported is updated to Angular 20.1, [matching the long term support](https://angular.dev/reference/releases#actively-supported-versions). The update includes some additional features:

- Build in SSR support, that hydrates queries automatically, matching how Angular HTTP client works.
- `toResource` helper, to integrate queries into new Angular APIs like SIgnal Forms
- `injectQueries` support
- Dynamic filtering for `injectIsFetching` and `injectIsMutating`
- Persist client, that works with SSR

It removed deprecated APIs of the experimental package, imrpoves some consistency issues (like the missing dynamic filters and options that accepted 2 different shapes), and moved the devtools to a separate package matching other adapters. 

Made as a prerelease targeting the release as `@tanstack/angular-query@5.0.0-rc` (similar on how the Solid 2.0 adapter), to match the version of the other adapters and how it is displayed in the docs, while not publishing a stable package yet. The experimental adapter isn't changed since it has been too "stable".

AI was used to help to write code, but it was heavelly reviewed.

## ✅ Checklist

- [X] I have tested code changes locally with `pnpm run test:pr`, or these tests do not apply to this pull request.
- [X] I fully understand the code in this pull request, including any code generated with AI assistance.

## 🚀 Release Impact


.changeset/angular-stable-release.md
```diff
@@ -0,0 +1,7 @@
+---
+'@tanstack/angular-query': major
```
TkDodo, line 2: have you set the libs to `4.0` and the changeset to major so that we get a 5.x then? I’m not sure if that will work 🤷

benjavicente: Yes 😅 

It doesn't seem to be a check for verifying `4.0` exists (or any existing published version). Audited that plan with Astra, it had an issue with the versioning group fixed in the new commit, it only seems to consider those groups versions to create the prerelease.

TkDodo (approved):

TkDodo (approved):

PR #11517, opened 2026-09-16
fix(solid-query): sweep X-Revalidate without a slice; reserve '*' for all

ryansolid:
## What

The single-flight consumer in `QueryClientProvider` assumed every delivery carried a dehydrated slice, so a response that declared `X-Revalidate` keys but folded nothing for the query cache — no collector registered, a redirect leaving the app — threw on `payload.queries`. The sweep now treats a missing payload as covering nothing and sweeps the declared scope in full.

The header's three states are three scopes:

| `X-Revalidate` | query cache |
| --- | --- |
| absent (bare `redirect()` / `reload()`) | untouched — a host default the cache does not presume |
| empty (`revalidate: []`) | nothing; an empty key is not a prefix that matches everything |
| named keys | prefix-matched queries, minus what the payload covered |
| `*` (`revalidate: '*'`) | every query, minus what the payload covered |

`*` is the host-independent spelling of "all" (cf. `Clear-Site-Data: "*"`); Solid Router keeps its own default of revalidating everything when the header is absent, and this consumer does not.

## Context

Pairs with the `@tanstack/solid-router` consumer that applies redirects and reloads route data from the same responses, and with a pending `@solidjs/web` transport change that delivers metadata-bearing mutation responses to every registered consumer whether or not data was folded — which is when `data` arrives `undefined` here.

## Tests

`flightData.test.tsx`: sliceless sweep of the declared scope, `'*'` with and without a slice (covered hashes exempt), empty declaration, metadata-free response. Full `@tanstack/solid-query` suite green (363).

— *Claude via Cursor*


brenelz:
## Review

**Intent:** Consumer used to assume every delivery had a dehydrated slice → throw on `payload.queries` when only metadata arrived. Missing payload = cover nothing; sweep declared scope. `*` = all (minus covered hashes). Absent header = **leave query cache alone** (router’s default “reload everything” is router-only).

**Verdict:** Small, correct, paired with TanStack/router#8461. Tests cover sliceless sweep, `*` with/without slice, empty declaration, metadata-free. **No blockers.**

### Nits
- Intentional asymmetry with the router on “header absent” is the right call — keep that table in the PR body.
- Empty key skip (`if (!key) continue`) prevents `revalidate: []` from matching everything — good; don’t “simplify” that away.

### What’s strong
- Clear three-state header model (absent / empty / named / `*`).
- Payload-covered hashes stay exempt from the sweep so single-flight savings aren’t spent twice.

birkskyum (approved):

PR #11347, opened 2026-08-31
Simplifed query methods/internal tests new lint

DogPawHat:
## 🎯 Changes

Prevents outdated methods (`fetchQuery` et al) being used via a lint rule.

## ✅ Checklist

- [x] I have followed the steps in the [Contributing guide](https://github.com/TanStack/query/blob/main/CONTRIBUTING.md).
- [x] I have tested code changes locally with `pnpm run test:pr`, or these tests do not apply to this pull request.
- [x] I fully understand the code in this pull request, including any code generated with AI assistance.

## 🚀 Release Impact

- [ ] This change affects published code, and I have generated a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).
- [x] This change is docs/CI/dev-only (no release).

## Summary by CodeRabbit

- **New Features**
  - Added linting guidance that flags direct use of deprecated imperative query-fetching methods and recommends the newer query APIs.

- **Tests**
  - Updated type and integration tests to accommodate the linting guidance while preserving existing coverage.
  - Added validation that infinite-query options cannot be used with standard query methods.
  - Added coverage for conditional Vue query functions using `skipToken`.
  - Corrected a test description and minor spelling error.

- **Chores**
  - Relaxed selected lint rules for codemod fixture files.

PR #11522, opened 2026-09-17
ci(*): generate reference docs in autofix and fail on TypeDoc conversion errors

sukvvon:
## 🎯 Changes

Runs `pnpm run generate-docs` in `autofix.ci` so regenerated reference docs are committed automatically. Two supporting fixes make that safe.

Reference docs under `docs/framework/*/reference` are generated from source by TypeDoc, but nothing keeps them in sync: the script runs in no workflow, and until #11523 `CONTRIBUTING.md` did not mention it. Drift accumulated twice in two days:

| PR | Pages regenerated | Caused by |
| --- | --- | --- |
| #11520 | 42 | #11347, #11502, #11508 |
| #11521 | 28 | #11512 |

None of the causing PRs touched JSDoc. #11512 was a four-line bug fix in `utils.ts`; that was enough to shift every symbol below it and break `Defined in:` links on 28 pages across all seven frameworks. An author fixing a bug has no signal that reference docs need regenerating, so this cannot reasonably be caught by review alone. `queryCache.ts` has since drifted again in #11524, which this PR will pick up on merge.

### Only on pushes

Following the review and [`TanStack/form`](https://github.com/TanStack/form/blob/main/.github/workflows/autofix.yml), generation is gated behind `github.event_name == 'push'`, with a `workflow_dispatch` input for manual runs. Regenerating on every pull request would add large, unrelated diffs.

The job also skips pushes of its own commit, matching on the `commit-message` the action is configured with — the two need to stay in sync, noted in a comment above the condition. A manual `workflow_dispatch` carries no `head_commit` and so is never skipped, which leaves a way to force a regeneration when the branch tip is already an autofix commit.

### `generate-docs` now builds the project it references

`angular-query-experimental` reaches `@tanstack/query-devtools` through a TypeScript project reference, so it consumes that package's emitted `.d.ts` rather than its source — the source is solid-js JSX and cannot be compiled under Angular's tsconfig. It is the only one of the seven documented packages with a `references` entry.

`dist-ts` is gitignored, so it is absent on a fresh clone and in CI, and TypeDoc then reports `TS6305` and converts nothing. Rather than adding a build step to the workflow, `generate-docs` builds it itself:

```json
"generate-docs": "tsc -b packages/query-devtools/tsconfig.json && node scripts/generate-docs.ts"
```

That keeps local runs and CI identical — `pnpm run generate-docs` works on a fresh clone with nothing extra to remember.

### Failing loudly instead of emptying the output

`generatePackageReferenceDocs` empties `outputDir` before conversion, then guards the rest with `if (project)`. A failed conversion therefore left the directory empty and exited 0 — indistinguishable from "every page was intentionally deleted".

The first push on this branch demonstrated it: the run went green and autofix committed the deletion of all 203 Angular pages. The script now throws instead, so CI stops rather than committing a destructive result. The follow-up push proved the fix end to end — autofix regenerated and restored all 203 pages.

Verified locally by deleting `packages/query-devtools/dist-ts`: `pnpm run generate-docs` rebuilds it, reports no `TS6305`, and leaves `docs/` unchanged.

## ✅ Checklist

## 🚀 Release Impact

## Summary by CodeRabbit

- **Documentation**
  - Improved reference documentation generation by building the required package first.
  - Documentation generation now reports clear errors when project conversion fails, helping prevent incomplete output.

- **Chores**
  - Added an option to manually trigger documentation generation through the automation workflow.
  - Automated fixes now avoid rerunning against commits they already created.


.github/workflows/autofix.yml
TkDodo, line 1: this should only run on main, not on every PR. It will make PRs quite large for no good reason otherwise. See how form does it:

https://github.com/TanStack/form/blob/e066b93afaa49f708843672a54ec1f53ec92490e/.github/workflows/autofix.yml#L37

sukvvon: @TkDodo Matched `form`'s approach in d2c7cecb9: gated the step behind `github.event_name == 'push'` and added the `workflow_dispatch` input for manual runs.

TkDodo (approved):

PR #11534, opened 2026-09-18
docs(query-core): inline option type aliases in generated reference docs

sukvvon:
## 🎯 Changes

Stacked on #11533.

Option types in the reference tables are printed as alias names, so the values an option actually accepts are one click away:

```
staleTime?    StaleTimeFunction<TQueryFnData, TError, TQueryFnData, TQueryKey>
status        QueryStatus
```

This adds `/** @inline */` to 17 type aliases in `query-core`, which makes TypeDoc print what they resolve to:

```diff
- staleTime?    StaleTimeFunction<TQueryFnData, TError, TQueryFnData, TQueryKey>
+ staleTime?    number | "static" | (query: Query<TQueryFnData, TError, TQueryFnData, TQueryKey>) => StaleTime

- status        QueryStatus
+ status        "error" | "pending" | "success"

- fetchStatus   FetchStatus
+ fetchStatus   "fetching" | "paused" | "idle"

- enabled?      QueryBooleanOption<...>
+ enabled?      false | true | (query: Query<...>) => boolean

- getNextPageParam   GetNextPageParamFunction<TPageParam, TQueryFnData>
+ getNextPageParam   (lastPage: TQueryFnData, allPages: TQueryFnData[], lastPageParam: TPageParam, allPageParams: TPageParam[]) => TPageParam | null | undefined
```

The aliases: `NetworkMode`, `QueryStatus`, `FetchStatus`, `MutationStatus`, `NotifyEventType`, `NotifyOnChangeProps`, `StaleTime`, `StaleTimeFunction`, `QueryBooleanOption`, `ThrowOnError`, `QueryPersister`, `QueryFunction`, `MutationFunction`, `QueryKeyHashFunction`, `InitialDataFunction`, `PlaceholderDataFunction`, `GetNextPageParamFunction`, `GetPreviousPageParamFunction`, `RetryValue`, `RetryDelayValue`.

`RetryValue` and `RetryDelayValue` resolve to `ShouldRetryFunction` and `RetryDelayFunction`, which are not exported and so have no page to link to. Those two are inlined as well, so `retry` renders as `(failureCount: number, error: TError) => boolean` rather than a name the reader cannot follow.

Only `query-core` is touched, and the change reaches all seven adapters because they inherit these options.

### How these were chosen

Every type alias that appears in the generated docs was checked against three rules:

1. **Alias pages with prose are excluded.** Inlining removes the link to the alias page, so any JSDoc only reachable there would be orphaned. They carry no prose of their own — only a type declaration and type parameters.
2. **Aliases that collapse to `object` are excluded** — `MutationFunctionContext`, `MutationScope`, `TimeoutProvider` and others would render as the single word `object`.
3. **Cell size was measured for each one.** The only alias that grows a row substantially is `NotifyOnChangeProps`: it is `keyof InfiniteQueryObserverResult`, so inlining it lists every key of the result type and takes the widest row from 923 to ~2,000 chars across 30 rows. It is included here at @TkDodo's request; happy to drop just that one if it reads too wide.

Rejected for other reasons: `OmitKeyof`, `DefaultedQueryObserverOptions`, `InferDataFromTag` and 10 similar ones pass the size check but lose information — `OmitKeyof<UseQueryOptions<...>, "queryFn">` becomes `object` — and they also strip the `redirect_from` frontmatter. `MutateFunction` only turns into another alias (`MutateFunctionRest`). `DefaultError` is unaffected, since it appears as a type-parameter default rather than in a type position.

No adapter package has a candidate: their aliases either carry JSDoc or name return types (`UseQueryResult`), which are concepts rather than shorthand and read worse when expanded.

### Verification

Compared the 1,415 generated pages against #11533:

- property entries lost: 0
- descriptions lost: 0
- broken relative links: 0
- all 17 alias pages still generated, so existing links keep working
- longest table row unchanged at 924 chars
- type column length: mean 39.3 → 42.0 chars, median unchanged at 9
- `pnpm --filter @tanstack/query-core run test:types` passes on TS 5.6 through 7.0

The source change is 17 comment lines; no type or runtime behaviour changes.

## ✅ Checklist

## 🚀 Release Impact


TkDodo:
please also inline:

- NotifyOnChangeProps
- ThrowOnError
- StaleTimeFunction
- QueryBooleanOption
- RetryValue
- RetryDelayValue
- InitialDataFunction

TkDodo (approved):

TkDodo (approved):

PR #11538, opened 2026-09-18
docs(*): promote default values from prose to '@defaultValue' tags

sukvvon:
## 🎯 Changes

Default values were written as prose at the end of a description, so the `Default value` column in the generated tables was almost always empty — it showed `undefined` for options that do have a default:

```
| enabled? | QueryBooleanOption<...> | undefined | ...Accepts a boolean or function that returns a boolean. Defaults to `true`. |
```

This promotes those sentences to `@defaultValue` tags, which moves the value into the column and removes it from the description:

```
| enabled? | false | true | (query: Query<...>) => boolean | true | ...Accepts a boolean or function that returns a boolean. |
```

44 tags added across `query-core`, `solid-query`, `query-persist-client-core` and the five devtools packages. The result across all generated pages:

| | Before | After |
| --- | --- | --- |
| `Default value` cells | 272 | 1,673 |
| carrying a real value | 10 (3.7%) | **517 (30.9%)** |

Values now shown: `true` (212), `false` (159), `'online'` (72), `0` (46), `'all'` (21), `'active'` (7).

Nothing is lost — the sentence moves from the description into the column, so `enabled` reads `true` instead of `undefined`, which was closer to wrong than uninformative.

### What was deliberately left as prose

A default only becomes a tag when the literal is the *whole* default. These stay in the description, because a column showing half of a conditional default is worse than one showing nothing:

- `retry` — `3` on the client and `0` on the server
- `gcTime` — `5 * 60 * 1000`, or `Infinity` during SSR
- `refetchOnReconnect` — `true` unless `networkMode` is `'always'`
- `retryDelay`, `combine`, `streamedQuery`'s `refetchMode`/`reducer` — the default is a described behaviour, not a literal

`notifyOnChangeProps` is also skipped: its default *is* `undefined`, and an untagged cell already renders as `undefined`, so tagging it would change nothing on the page.

### Devtools packages

16 of the tags are in devtools packages, which generate no reference pages. They are included because `@defaultValue` is standard JSDoc that editors surface on hover, and because leaving them as prose would mean the same pattern is written two different ways across the repo.

### A tradeoff worth flagging

Filling the column turns 125 three-column tables into four-column ones, and the `Description` column gets narrower as a result — on a 1728px viewport it renders at 228px and wraps to a few words per line. That comes from the docs site, not this PR: the content container is capped at `max-w-[768px]` and the sponsor sidebar takes the rest, so the page had 768px to divide before any of this. Worth fixing on the site side, but the column now carries 517 correct values where it previously showed `undefined`.

### Verification

Compared the 1,415 generated pages against `main`:

- property entries lost: 0
- descriptions lost: 0 (the `Defaults to …` sentences move into the column)
- broken relative links: 0
- longest table row unchanged
- `pnpm --filter @tanstack/query-core run test:types` passes on TS 5.6 through 7.0

The source change is comments only; no type or runtime behaviour changes.

## ✅ Checklist

## 🚀 Release Impact


sukvvon:
Draft until #11534 lands — this branch is stacked on it, so the diff here only makes sense once that one is in.

TkDodo (approved):

